PR #128, opened 2025-04-05
Mkhazraee/best effort reg

mkhazraee:
Agent/UCX: regiterMem does best effort across backends, and ucx errors out on second gpu
 * UCX below version 1.19 can only have a single worker per GPU, so if it's instantiated instead of UCX-MO, and memory from a second GPU gets registered, it would error out. Previously there were asserts to do this check, now it returns error.
 * registerMem in Agent was was all or nothing, now it tries to and returns error if even not a single backend could register.
 * Now user can instantiate both UCX and UCX-MO. For the first GPU, UCX will accept the registrations, for more only UCX-MO will accept. During time of transfer, if they belong to that GPU, UCX will be used (if it was instantiated first), otherwise UCX-MO.
 * This behavior will be deprecated when 1.19 becomes available, where for UCX below version 1.19 we use UCX-MO, and for after we use base UCX without the error checks added in this PR.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -34,9 +34,11 @@ class nixlUcxCudaCtx {
 public:
 #ifdef HAVE_CUDA
     CUcontext pthrCudaCtx;
+    int myDevId;
 
     nixlUcxCudaCtx() {
         pthrCudaCtx = NULL;
+        myDevId = -1;
     }
 #endif
     void cudaResetCtxPtr();
@@ -81,6 +83,13 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
 
     was_updated = false;
 
+    if (expected_dev == -1)
+        return -1;
+
+    // incorrect dev id from first registration
+    if (myDevId != -1 && expected_dev != myDevId)
+        return -1;
+
     ret = cudaQueryAddr(address, is_dev, dev, ctx);
     if (ret) {
         return ret;
@@ -107,6 +116,7 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
 
     pthrCudaCtx = ctx;
     was_updated = true;
+    myDevId = expected_dev;
 
     return 0;
 }
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -91,7 +91,6 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
         return 0;
     }
 
-    assert(dev == expected_dev);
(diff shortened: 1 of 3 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -91,7 +91,6 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
         return 0;
     }
 
-    assert(dev == expected_dev);
     if (dev != expected_dev) {
         /* TODO: proper error codes */
         return -1;
```
mkhazraee, line 105: @artpol84 There is only one assert left on line 101, can we change that to error code too? That way we can remove cassert from this file too.

tstamler:
I'd like to see the title updated, a PR description, and at least one basic test that this acts as expected, maybe in ucx_backend_mo_test

tstamler (approved):

src/plugins/ucx/ucx_backend.cpp
tstamler: I've addressed this in my commit

tstamler:
I've added a test for the correct error to be returned in ucx_backend_test, and verified that it works on Eos.

tstamler:
I've added devId check, run on Eos, and verified with prints that the UCX MO backend is selected (because our NB API currently doesn't return backend selection its hard to formalize a test).

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -82,6 +83,13 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
 
     was_updated = false;
 
+    if (expected_dev == -1)
+        return -1;
```
artpol84, line 88: Why not NIXL Error code?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -91,24 +99,24 @@ int nixlUcxCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &was_
...
     }
 
     if (pthrCudaCtx) {
-        // Context was already set previously
-        assert(pthrCudaCtx == ctx);
+        // Context was already set previously, and does not match new context
         if (pthrCudaCtx != ctx) {
-            // Fatal error
-            abort();
+            /* TODO: proper error codes */
```
artpol84: And todo log output

artpol84 (approved):

src/plugins/ucx/ucx_backend.cpp
mkhazraee: I didn't know how exactly we have to do it, as there seems to be some overloading or so. We can do it in a separate PR.

src/plugins/ucx/ucx_backend.cpp
mkhazraee: Done

PR #129, opened 2025-04-06
core: refactor registerMem

artpol84:
Avoid duplicated code on the error path


mkhazraee:
Hi Artem, that code is gone in #128, so closing this. Also I was borderline about goto statement, this is the only case that I'm kind of ok with, but since it was a single replication I didn't go for it.

PR #133, opened 2025-04-08
clang-format: Add .clang-format

etoledano:
Add .clang-format to provide unified formatting style to NIXL's CPP source files (*.cpp, *.h). This should later be used to enforce checks during CI/CD.

This format is propogated from Dynamo's at:
https://github.com/ai-dynamo/dynamo/blob/main/.clang-format

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,40 @@
+---
+BasedOnStyle: Google
+
+IndentWidth: 4
+ColumnLimit: 120
+ContinuationIndentWidth: 8
+UseTab: Never
+MaxEmptyLinesToKeep: 2
+
+SortIncludes: true
+CompactNamespaces: true
+ReflowComments: true
+
+DerivePointerAlignment: false
+PointerAlignment: Left
+
+AllowShortIfStatementsOnASingleLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+
+AccessModifierOffset: -4
+
+AlwaysBreakAfterReturnType: TopLevelDefinitions
+AlignAfterOpenBracket: AlwaysBreak
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: true
+  AfterNamespace: false
+  AfterStruct: false
+  AfterUnion: false
+  BeforeCatch: true
+
+BinPackArguments: true
+BinPackParameters: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+
+IndentCaseLabels: true
--- a/src/api/cpp/backend/backend_aux.h
+++ b/src/api/cpp/backend/backend_aux.h
@@ -19,9 +19,10 @@
 
 #include <mutex>
 #include <string>
-#include "nixl_types.h"
-#include "nixl_descriptors.h"
+
 #include "common/nixl_time.h"
+#include "nixl_descriptors.h"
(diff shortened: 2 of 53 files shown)


artpol84:
@etoledano this seems to be a complete rewrite of the code style we used previously.
IMO it's counter-productive. E.g. we definitely want to stick to 4-space tabs for better readability.

I suggest having an internal meeting to agree on the code  style first.

.clang-format
```diff
@@ -0,0 +1,38 @@
+---
+BasedOnStyle: Google
+
+IndentWidth: 2
```
artpol84:
```suggestion
IndentWidth: 4
```

.clang-format
```diff
@@ -0,0 +1,38 @@
...
+ColumnLimit: 120
+ContinuationIndentWidth: 4
```
artpol84: ?
```suggestion
ContinuationIndentWidth: 8
```

src/api/cpp/backend/backend_aux.h
```diff
@@ -19,20 +19,21 @@
...
 // level direction or so.
 typedef std::vector<std::pair<std::string, std::string>> notif_list_t;
 
 
 class nixlBackendOptionalArgs {
-    public:
-        // During postXfer, user might ask for a notification if supported
-        nixl_blob_t notifMsg;
-        bool        hasNotif = false;
+ public:
```
artpol84: I personally prefer `public` and etc. specifiers to be aligned with the `class` and not "one space away"
```suggestion
public:
```

src/api/cpp/backend/backend_engine.h
```diff
@@ -19,169 +19,148 @@
...
+  virtual nixl_status_t unloadMD(nixlBackendMD* input) = 0;
+
+  // Preparing a request, which populates the async handle as desired
+  virtual nixl_status_t prepXfer(
+      const nixl_xfer_op_t& operation, const nixl_meta_dlist_t& local, const nixl_meta_dlist_t& remote,
+      const std::string& remote_agent, nixlBackendReqH*& handle, const nixl_opt_b_args_t* opt_args = nullptr) = 0;
+
+  // Posting a request, which completes the async handle creation and posts it
+  virtual nixl_status_t postXfer(
+      const nixl_xfer_op_t& operation, const nixl_meta_dlist_t& local, const nixl_meta_dlist_t& remote,
```
artpol84: IMO we had a better alignment here.
Is there a way to force custom indentation rules for specific places?

.clang-format
etoledano: Done.

src/api/cpp/backend/backend_aux.h
etoledano: Done.

.clang-format
etoledano: Done.

src/api/cpp/backend/backend_engine.h
etoledano: Support was [recently added](https://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20240819/611583.html) using BPPS_AlwaysOnePerLine.

I'm less of a fan of this approach, since:

1. Using namespace to categorize types helps recycling names and keep them short
2. API description should result in a short variable name
3. Common practice is to gather function arguments list when there are too many of them, or when two (or more) of them are of the same type.

Thus, using a new line for very short text would result in a longer (visually speaking) code to look at, and does less sense when inducing to follow common guidelines. Nevertheless, this style is a visual preference which we can consider.

PR #134, opened 2025-04-08
DOCA plugin: enable RDMA GPU communications in NIXL

e-ago:
Creation of a new NIXL plugin with DOCA libraries ([GPUNetIO](https://docs.nvidia.com/doca/sdk/doca+gpunetio/index.html) and RDMA) to enable GPU initiated communications in NIXL


.gitlab/build.sh
```diff
@@ -86,11 +86,16 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/v1.18.0" | tar xz
         ldconfig \
 )
 
-export LIBRARY_PATH=$LIBRARY_PATH:/usr/local/cuda/lib64
-export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:${INSTALL_DIR}/lib
+wget https://www.mellanox.com/downloads/DOCA/DOCA_v2.10.0/host/doca-host_2.10.0-093000-25.01-ubuntu2404_amd64.deb
+dpkg -i doca-host_2.10.0-093000-25.01-ubuntu2404_amd64.deb
+apt-get update
+apt-get -y install doca-all doca-sdk-gpunetio libdoca-sdk-gpunetio-dev
```
tstamler, line 93: do we need both the wget package and the apt packages? its ok if we do, just feels a little odd

meson.build
```diff
@@ -40,6 +40,24 @@ else
     include_directories : include_directories(cuda_inc_path))
 endif
 
+add_languages('CUDA')
+cuda = import('unstable-cuda')
+nvcc = meson.get_compiler('cuda')
```
tstamler: will this succeed if CUDA is not installed? as of right now, CUDA is an optional dependency to NIXL, so what we'd want is to just surround all this with an if statement for the cuda_dep variable for the case where CUDA is not installed. Then we would not build DOCA plugin by default if CUDA isn't present.

meson.build
```diff
@@ -58,6 +76,10 @@ if get_option('disable_gds_backend')
     add_project_arguments('-DDISABLE_GDS_BACKEND', language: 'cpp')
 endif
 
+if get_option('disable_doca_backend')
```
tstamler: I'm sorry you seem to have gone off of the GDS backend, I should have warned you beforehand but the build situation for it is a little messy. We'd prefer to not even have the disable_gds_backend flag and plan to remove it when someone has time. 

basically, I'd prefer not to have a disable_doca_backend flag and just have it not try to build the DOCA plugin if the dependencies aren't met

pyproject.toml
```diff
@@ -32,4 +32,4 @@ authors = [
 ]
 
 [tool.meson-python.args]
-setup = ['-Ddisable_gds_backend=true', '-Dinstall_headers=false']
+setup = ['-Ddisable_gds_backend=true', '-Dinstall_headers=false'] #-Ddisable_doca_backend=false
```
tstamler: no need, we're already also removing disable_gds_backend from this file

src/api/cpp/nixl_types.h
```diff
@@ -134,6 +135,16 @@ class nixlAgentOptionalArgs {
          * @var makeXferReq boolean to skip merging consecutive descriptors, used in makeXferReq.
          */
         bool skipDescMerge = false;
+
+        /**
+         * @var Backend GPU device ID to use
+         */
+        uint32_t devId;
```
tstamler: this shouldn't be necessary, this should be specified in nixl descriptor

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+void nixlDocaEngine::_requestFini(void *request)
+{
+	/* Finalize request */
+	nixlDocaBckndReq *req = (nixlDocaBckndReq*)request;
+	req->~nixlDocaBckndReq();
+}
+
+/****************************************
+ * Constructor/Destructor
+*****************************************/
```
tstamler: try to move this comment back to the right place, and maybe relocate this helper function below

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+	result = doca_gpu_destroy(gdevs[0].second);
+	if (result != DOCA_SUCCESS)
+		DOCA_LOG_ERR("Failed to close DOCA GPU device: %s", doca_error_get_descr(result));
+}
+
+/****************************************
+ * Connection management
+*****************************************/
+
+nixl_status_t nixlDocaEngine::checkConn(const std::string &remote_agent) {
```
tstamler: these were mainly helper functions for the UCX backend to do connection management, if you're not making use of them you don't have to include them. I'll mention again when I get to the header file, but this backend_engine class specifies the only functions you actually need to implement for plugins: 

https://github.com/ai-dynamo/nixl/blob/main/src/api/cpp/backend/backend_engine.h

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+	//TODO: eventually std::byte?
+	char* addr = new char[size];
+
+	std::cout << "loadRemoteConnInfo " << remote_agent << "\n";
+	if(remoteConnMap.find(remote_agent) != remoteConnMap.end()) {
+		return NIXL_ERR_INVALID_PARAM;
+	}
+
+	nixlSerDes::_stringToBytes((void*) addr, remote_conn_info, size);
+	result = doca_rdma_connect(rdma, addr, size, connection);
```
tstamler: I'm sure I can figure it out, but is this like some doca RDMA CM?

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+	/* Already connected to remote QP at loadRemoteConnInfo time */
+	return NIXL_SUCCESS;
+}
+
+nixl_status_t nixlDocaEngine::disconnect(const std::string &remote_agent) {
+	return NIXL_SUCCESS;
+}
+
+nixl_status_t nixlDocaEngine::loadRemoteConnInfo (const std::string &remote_agent,
+												 const std::string &remote_conn_info)
```
tstamler: spacing on some of these is coming out weird, so just double check to make sure there's no tabs and the parameters are lined up properly

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+}
+
+// To be cleaned up
+nixl_status_t nixlDocaEngine::loadRemoteMD (const nixlBlobDesc &input,
+										   const nixl_mem_t &nixl_mem,
+										   const std::string &remote_agent,
+										   nixlBackendMD* &output)
+{
+	return internalMDHelper(input.metaInfo, remote_agent, output);
+	return NIXL_SUCCESS;
```
tstamler: double return?

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+	doca_kernel_write(0, rdma_gpu, nullptr, 0);
+	doca_kernel_read(0, rdma_gpu, nullptr, 0);
+
+	xferRingPos = 0;
+	firstXferRingPos = 0;
+}
+
+nixl_mem_list_t nixlDocaEngine::getSupportedMems () const {
+	nixl_mem_list_t mems;
+	mems.push_back(DRAM_SEG);
```
tstamler, line 474: it seems like the plugin only supports GPU-to-GPU transfers, so let's remove DRAM_SEG from this list

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+{
+	uint32_t pos;
+	nixlDocaBckndReq *treq = new nixlDocaBckndReq;
+	nixlDocaPrivateMetadata *lmd;
+	nixlDocaPublicMetadata *rmd;
+	uint32_t lcnt = (uint32_t)local.descCount();
+	uint32_t rcnt = (uint32_t)remote.descCount();
+
+	treq->stream = (cudaStream_t)opt_args->customParam;
+	treq->devId = (uint32_t)opt_args->devId;
```
tstamler: should mention here again, it doesn't need to be an opt_arg, the device ID should come from the descriptors. If you want to support the option for all the local descriptors to be able to be launched on different GPUs, you'll need to parse the device id for each desc. Otherwise, just verify that all the descs in the list have the same dev_id.

src/plugins/doca/doca_backend.cpp
```diff
@@ -0,0 +1,702 @@
...
+			return NIXL_ERR_BACKEND;
+	}
+	return NIXL_SUCCESS;
+}
+
+nixl_status_t nixlDocaEngine::releaseReqH(nixlBackendReqH* handle)
+{
+	firstXferRingPos = xferRingPos & (DOCA_XFER_REQ_MAX - 1);
+
+	return NIXL_SUCCESS;
```
tstamler, line 944: do we not need to free anything here?

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+        std::atomic<uint32_t> xferRingPos;
+        uint32_t firstXferRingPos;
+        // Map of agent name to saved nixlDocaConnection info
+        std::unordered_map<std::string, nixlDocaConnection,
+                           std::hash<std::string>, strEqual> remoteConnMap;
+
+        class nixlDocaBckndReq : public nixlLinkElem<nixlDocaBckndReq>, public nixlBackendReqH {
+            private:
+            public:
+                cudaStream_t stream;
```
tstamler, line 166: is this just a handle that doesn't need to be freed?

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+#include "common/nixl_time.h"
+#include "common/list_elem.h"
+
+#define DOCA_DEVINFO_IBDEV_NAME_SIZE 64
+#define RDMA_RECV_QUEUE_SIZE 2048
+#define RDMA_SEND_QUEUE_SIZE 2048
+#define DOCA_XFER_REQ_SIZE 512
+#define DOCA_XFER_REQ_MAX 16
+
+typedef enum {CONN_CHECK, NOTIF_STR, DISCONNECT} ucx_cb_op_t;
```
tstamler: don't need this here

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+
+typedef void * nixlDocaReq;
+
+struct nixl_ucx_am_hdr {
+    ucx_cb_op_t op;
+};
```
tstamler: don't need

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+        std::string remoteAgent;
+        // rdma qp
+        // nixlDocaEp ep;
+        volatile bool connected;
+
+    public:
+        // Extra information required for UCX connections
+
+    friend class nixlDocaEngine;
+};
```
tstamler, line 85: does this do anything right now? seems to just have a name

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+        struct doca_rdma_connection *connection;
+        struct docaXferReqGpu *xferReqRingGpu;
+        struct docaXferReqGpu *xferReqRingCpu;
+        std::atomic<uint32_t> xferRingPos;
+        uint32_t firstXferRingPos;
+        // Map of agent name to saved nixlDocaConnection info
+        std::unordered_map<std::string, nixlDocaConnection,
+                           std::hash<std::string>, strEqual> remoteConnMap;
+
+        class nixlDocaBckndReq : public nixlLinkElem<nixlDocaBckndReq>, public nixlBackendReqH {
```
tstamler, line 163: does this need to be a nixlLinkElem or is this copied over?

src/plugins/doca/doca_backend.h
```diff
@@ -0,0 +1,246 @@
...
+
+        nixl_status_t checkXfer (nixlBackendReqH* handle);
+        nixl_status_t releaseReqH(nixlBackendReqH* handle);
+
+        int progress();
+
+        nixl_status_t getNotifs(notif_list_t &notif_list);
+        nixl_status_t genNotif(const std::string &remote_agent, const std::string &msg);
+
+        //public function for UCX worker to mark connections as connected
```
tstamler: these functions can almost certainly be removed

src/plugins/doca/kernel_read.cu
```diff
@@ -0,0 +1,96 @@
...
+        if (result != DOCA_SUCCESS)
+            printf("Error %d doca_gpu_dev_rdma_wait_all\n", result);
+
+        // printf(">>>>>>> CUDA rdma read kernel pos %d num %d completed %d ops\n", pos, xferReqRing[pos].num, num_ops);
+
+        xferReqRing[pos].num = 0;
+        __threadfence_system();
+    }
+
+}
```
tstamler, line 65: CUDA kernels are mostly new to me but this seems simple enough

src/plugins/doca/meson.build
```diff
@@ -0,0 +1,57 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+doca_path = get_option('doca_path')
+doca_inc_path = doca_path + '/include'
+
+doca_dep0 = dependency('doca-common')
+doca_dep1 = dependency('doca-rdma')
+doca_dep2 = dependency('doca-gpunetio')
```
tstamler, line 21: does this need to be three different dependencies?

test/nixl/nixl_doca_cuda_test.cu
```diff
@@ -0,0 +1,355 @@
...
+        std::cout << " Desc List from Target to Initiator\n";
+        dram_for_doca.print();
+
+        /** Sending both metadata strings together */
+        assert(serdes->addStr("AgentMD", metadata) == NIXL_SUCCESS);
+        assert(dram_for_doca.trim().serialize(serdes) == NIXL_SUCCESS);
+
+        std::cout << " Serialize Metadata to string and Send to Initiator\n";
+        std::cout << " \t -- To be handled by runtime - currently sent via a TCP Stream\n";
+        sendToInitiator(initiator_ip, initiator_port, serdes->exportStr());
```
tstamler, line 268: I think nixl_test is out of date, but getLocalMD should already do all this for you: collect the agent connection and registered memory metadata and return a string you can send

test/nixl/nixl_doca_cuda_test.cu
```diff
@@ -0,0 +1,355 @@
...
+        std::cout << " End Control Path metadata exchanges \n";
+
+        std::string rrstr = recvFromTarget(initiator_port);
```
tstamler: this is what we talked about in the meeting, I see the problem but I'm not sure the best way to handle it. Not sure if the backend itself should be passing this OOB or not. Will discuss with team

test/unit/plugins/doca/common.h
```diff
@@ -0,0 +1,55 @@
...
+#include <cstring>
+#include "stream/metadata_stream.h"
+#include "serdes/serdes.h"
+
+#define NUM_TRANSFERS 1
+#define SIZE 1024
+
+#if __cplusplus
+extern "C" {
+#endif
```
tstamler, line 45: just curious, is there a reason to extern "C" here?

test/nixl/nixl_doca_cuda_test.cu
```diff
@@ -0,0 +1,355 @@
...
+
+static void checkCudaError(cudaError_t result, const char *message) {
+	if (result != cudaSuccess) {
+		std::cerr << message << " (Error code: " << result << " - "
+				   << cudaGetErrorString(result) << ")" << std::endl;
+		exit(EXIT_FAILURE);
+	}
+}
+
+__global__ void target_kernel(uintptr_t addr, size_t size)
```
tstamler, line 67: can we seperate out the CUDA code so that this test can be separate .cu and .cpp files?

test/unit/plugins/doca/nixl_doca_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+	metadata_initiator_connect = serdes_initiator_connect->getStr("AgentMD");
+	assert (metadata_initiator_connect != "");
+	agent_initiator.loadRemoteMD(metadata_initiator_connect, name);
+
+	/* Fake initiator -> target socket send/recv */
+	std::cout << "Connect Target\n";
+	serdes_target_connect->importStr(serdes_initiator->exportStr());
+	metadata_target_connect = serdes_target_connect->getStr("AgentMD");
+	assert (metadata_target_connect != "");
+	agent_target.loadRemoteMD(metadata_target_connect, name);
```
tstamler, line 147: I'm getting a little lost here, I think all this serdes usage shouldn't be necessary. In general, users shouldn't need to use serdes at all.

test/unit/plugins/meson.build
```diff
@@ -21,3 +21,9 @@ disable_gds_backend = get_option('disable_gds_backend')
 if not disable_gds_backend
     subdir('cuda_gds')
 endif
+
+message('almost in doca build test')
```
tstamler: let's remove this

.gitlab/build.sh
e-ago: wget download the .deb package and apt installs it

src/plugins/doca/doca_backend.cpp
e-ago: you can enable RDMA CM it if you want, I'm not using it in the code for now

src/plugins/doca/doca_backend.cpp
e-ago: Not really, I can issue an RDMA read/write from GPU with data on CPU memory

src/plugins/doca/doca_backend.cpp
e-ago: no as this is just an atomic counter I'm using to access the `xferReqRingGpu` list of objects in a ring way. What must be freed is the `xferReqRingGpu` but this is done `~nixlDocaEngine`

src/plugins/doca/doca_backend.h
e-ago: Correct, the stream is created by the application and passed to the backend so it's application responsibility to destroy it

src/plugins/doca/doca_backend.h
e-ago: without it, I get compilation errors when filling the remoteConnMap in `loadRemoteConnInfo`:
```
../src/plugins/doca/doca_backend.cpp: In member function ‘virtual nixl_status_t nixlDocaEngine::loadRemoteConnInfo(const string&, const string&)’:
../src/plugins/doca/doca_backend.cpp:329:14: error: ‘std::string nixlDocaConnection::remoteAgent’ is private within this context
  329 |         conn.remoteAgent = remote_agent;
      |              ^~~~~~~~~~~
In file included from ../src/plugins/doca/doca_backend.cpp:17:
../src/plugins/doca/doca_backend.h:68:21: note: declared private here
   68 |         std::string remoteAgent;
      |                     ^~~~~~~~~~~
../src/plugins/doca/doca_backend.cpp:330:14: error: ‘volatile bool nixlDocaConnection::connected’ is private within this context
  330 |         conn.connected = true;
      |              ^~~~~~~~~
In file included from ../src/plugins/doca/doca_backend.cpp:17:
../src/plugins/doca/doca_backend.h:71:23: note: declared private here
   71 |         volatile bool connected;
   ```

src/plugins/doca/doca_backend.h
e-ago: I copied from UCX plugin, it's used in the `internalMDHelper` when a new rkey is received from remote peer

src/plugins/doca/kernel_read.cu
e-ago: removed, don't need it as it's at the end of the kernel

src/plugins/doca/meson.build
e-ago: Yes, each doca library has a separate dependency so you can clearly stated what you need the your project

test/unit/plugins/doca/nixl_doca_test.cpp
e-ago: This is a local single peer, two engines test and I wanted to reproduce all the steps as in the 2 peers remote test. I agree serdes is not actually needed here

tstamler (commented):
Looks much better, going to wait for @mkhazraee to review

tstamler (approved):

PR #138, opened 2025-04-09
fix buggy logics in check_remote_xfer_done

jokerwyt:
two bugs:
1. str and bytes mismatch
2. `message` is removed from notifs below so it should be `msg` instead of `lookup_msg`

### DIFF
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -316,8 +316,8 @@ class nixl_agent:
 
         if remote_agent_name in self.notifs:
             for msg in self.notifs[remote_agent_name]:
-                if lookup_msg in msg:
-                    message = lookup_msg
+                if str(lookup_msg) in str(msg):
+                    message = msg
                     found = True
                     break
         if message:


tstamler:
NIXL notifications are meant to be bytes, not strings, because they do not always have data that can be encoded to a text string. Additionally, we want to use lookup_msg because the input msg does not need to match the actual message in the notification map (see how we are using "in" to look for it). We support this if the user wants to search by a specific tag, and then return the whole notification to avoid confusion.

This function is demonstrated/tested in our NIXL Python API example: 
https://github.com/ai-dynamo/nixl/blob/main/examples/python/nixl_api_example.py

jokerwyt:
@tstamler I really see exceptions reported in my use case. Please take a closer look.
```python
    def check_remote_xfer_done(self, remote_agent_name: str, lookup_msg: bytes) -> bool:
        self.notifs = self.agent.getNotifs(self.notifs)  # Adds new notifs
        found = False
        message = None

        if remote_agent_name in self.notifs:
            for msg in self.notifs[remote_agent_name]:
                # they are even not the same type, and python will raise a exception for this.
                if lookup_msg in msg:
                    message = lookup_msg
                    found = True
                    break
        if message:
            # This will obviously cause exception since message is LOOKUP_MSG which is not in self.notifs.
            self.notifs[remote_agent_name].remove(message) 
        return found

```

jokerwyt:
@tstamler 
1. I think we should check the type of user arguments since python type signature is not forced.
2. There is a bug in removing logic.

PR #145, opened 2025-04-10
cuda_gds: cuFileBatch pool and reposting postXfer/getXferStatus for NIXL + GDS plugin

vvenkates27:
1. CuFile Batches can be precreated to remove them from data path 
2. Fix PrepXfer/PostXfer usage to support multiple iterations of post/prep
3. Provide a more elaborate example for postXfer/getXferStatus 
4. Additional code cleanup for ease of readability.
5. Added a README for cufile.json configuration required for compatibility mode
6. Warn and continue on cuFileBufRegister Failures as cuFile can handle this internally.


tstamler:
@vvenkates27 can you verify that the changes here are correct? it seems like it might have been based on an old branch because a lot of these changes don't make sense. I've moved this back to a draft for now for verification, but also because we might do a larger PR to make backend transfer handles repostable.

vvenkates27:
> @vvenkates27 can you verify that the changes here are correct? it seems like it might have been based on an old branch because a lot of these changes don't make sense. I've moved this back to a draft for now for verification, but also because we might do a larger PR to make backend transfer handles repostable.

I rebased and updated it after the minor patch was merged to master.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -266,195 +159,29 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
...
-    }
-
-    cuFileDriverClose();
-}
-
-// Add helper function to create batches that can be called from both prepXfer and postXfer
-nixl_status_t nixlGdsEngine::createBatches(const nixl_xfer_op_t &operation,
-                                           const nixl_meta_dlist_t &local,
-                                           const nixl_meta_dlist_t &remote,
-                                           nixlGdsBackendReqH* gds_handle) {
```
tstamler, line 432: I'm confused why Github is saying this whole section is being deleted and readded later, but maybe just a weird diff generation...

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -509,34 +246,163 @@ nixl_status_t nixlGdsEngine::createBatches(const nixl_xfer_op_t &operation,
...
+    return status;
+}
+
+nixl_status_t nixlGdsEngine::releaseReqH(nixlBackendReqH* handle)
+{
+
+    nixlGdsBackendReqH *gds_handle = (nixlGdsBackendReqH *) handle;
+
+    delete gds_handle;
+    gds_handle = nullptr;
```
tstamler, line 402: setting to nullptr here won't do anything

src/plugins/cuda_gds/gds_backend.h
```diff
@@ -38,54 +38,19 @@ class nixlGdsMetadata : public nixlBackendMD {
...
-        nixl_status_t cancelBatch();
-        void destroyBatch();
-        void reset();
-        unsigned int getMaxReqs() const { return max_reqs; }
+struct GdsTransferRequestH {
+    void*           addr;
+    size_t          size;
+    size_t          file_offset;
+    CUfileHandle_t  fh;
+    CUfileOpcode_t  op;
```
tstamler: not sure why we're going back from a class to a struct

src/plugins/cuda_gds/gds_backend.h
```diff
@@ -38,54 +38,19 @@ class nixlGdsMetadata : public nixlBackendMD {
...
-        nixlGdsBackendReqH() {}
-        ~nixlGdsBackendReqH() {
-            for (auto obj : batch_io_list)
-                delete obj;
-        }
+struct nixlGdsBackendReqH : public nixlBackendReqH {
+    nixlGdsBackendReqH() : needs_prep(true) {}
+    std::vector<GdsTransferRequestH> request_list;
+    std::vector<nixlGdsIOBatch*> batch_io_list;
+    bool needs_prep;
```
tstamler: same as above, why struct

tstamler:
@vvenkates27 please refer to #144 or check the slack channel for how to re-enable CI running by default, or how to manually trigger CI

src/plugins/cuda_gds/gds_backend.cpp
vvenkates27: I had reorganized code, the utils code is in gds_utils files

src/plugins/cuda_gds/gds_backend.cpp
vvenkates27: Just to ensure delete doesn't cause double free.. It expects to be nullptr.

src/plugins/cuda_gds/gds_backend.cpp
tstamler: Yes, I saw the utils code change, but these are just nixlGdsEngine functions being moved around. If its part of a general reorganization of that file, that's ok, but it makes it harder to review the PR.

src/plugins/cuda_gds/gds_backend.cpp
tstamler: gds_handle is a local variable, setting it to nullptr here doesn't change the nixlXferReqH

src/plugins/cuda_gds/gds_backend.cpp
vvenkates27: createBatches is no longer used, its code that is removed.

src/plugins/cuda_gds/gds_backend.h
vvenkates27: made them classes now..

src/plugins/cuda_gds/gds_backend.h
vvenkates27: Made them classes now.

src/core/nixl_agent.cpp
```diff
@@ -751,10 +751,11 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
     }
 
     // We CAN repost a previous request that is completed
-    if (req_hndl->status == NIXL_SUCCESS && req_hndl->backendHandle)
+    if (req_hndl->status == NIXL_SUCCESS && req_hndl->backendHandle &&
```
mkhazraee: Let's do 3 lines for if statement, each part between && is long, and in a comment mention this if statement is going to be removed when UCX backends also support this.

src/plugins/cuda_gds/gds_plugin.cpp
```diff
@@ -20,7 +20,7 @@
 
 // Plugin version information
 static const char* PLUGIN_NAME = "GDS";
-static const char* PLUGIN_VERSION = "0.1.0";
+static const char* PLUGIN_VERSION = "0.1.1";
```
mkhazraee, line 23: @tstamler should we update other plugins too?

src/core/nixl_agent.cpp
vvenkates27: Done

src/plugins/cuda_gds/gds_backend.cpp
vvenkates27: Removed

src/plugins/cuda_gds/gds_plugin.cpp
vvenkates27: Note for @tstamler  not relevant to this patch.

tstamler (approved):

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
```diff
@@ -543,43 +588,77 @@ int main(int argc, char *argv[])
...
+        if (ret != NIXL_SUCCESS) {
+            std::cerr << "Failed to create read transfer request" << std::endl;
             goto cleanup;
         }
+        std::cout << "Read transfer request created." << std::endl;
 
-        while (status == NIXL_IN_PROG) {
-            status = agent.getXferStatus(treq);
+        // Now do the iterations
+        for (unsigned int iter = 0; iter < iterations; iter++) {
```
aranadive, line 624: The loops for read/write are similar, can you refactor?

aranadive (approved):
LG. Modulo the refactor comment.

mkhazraee (approved):

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
vvenkates27: Since its a test its ok to keep them separate?

aranadive: I only meant that in both WRITE READ case you are only doing xfer(req) iterations times, which could be made common. But its okay to do it later.

vvenkates27: I see, I agree Will add a TODO for later. Thanks

src/plugins/cuda_gds/gds_backend.cpp
KiranModukuri, line 1: we need to handle the case where buffer registration fails.  This is optimization path and we should be able to handle without buffer registration

KiranModukuri (approved):
Approved 
please add README for NIXL specific JSON changes and addressing the buffer register failure comment.

src/plugins/cuda_gds/gds_backend.cpp
vvenkates27: Done. I have removed the fail on buffer registration - and added a warning instead of failure.

vvenkates27:
> Approved please add README for NIXL specific JSON changes and addressing the buffer register failure comment.

Added README and addressed buffer registration failure comment.

PR #151, opened 2025-04-14
Implement multi-threading support

w1ldptr:
Current NIXL implementation sidesteps any considerations of executing in multi-threaded environment by leaving it up to the user to ensure that access to nixlAgent instance is synchronized. Extend this by adding a strict synchronization mode while also preserving existing mode but making it explicit.

To continue using NIXL with exactly the same performance set configuration variable disable_mt=true which statically removes all locking at build time.

When NIXL is compiled with multi-threading enabled still allow dynamic configurations via nixlAgentConfig->syncMode field:

- When set to NIXL_SYNC_NONE no locking is performed and behavior is similar as before this change with the small difference that dynamic check is performed every time the lock would be taken with strict model.

- When set to NIXL_SYNC_STRICT all nixlAgent API methods are fully synchronized with internal exclusive lock.

In order to allow dynamic configuration at runtime the lock implemented as a thin BasicLockable-compatible wrapper around std::mutex that accepts sync_mode parameter to its constructor and skip locking/unlocking the mutex when synchronization mode is not strict. To allow static configuration at build time the lock is compiled out and the macro that is used to create a lock guard for the lock is set to no-op.

### DIFF
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -35,6 +35,9 @@ class nixlAgentConfig {
         bool     useListenThread;
         /** @var Port for listener thread to use */
         int      listenPort;
+        /** @var synchronization mode for multi-threaded environment execution */
+        nixl_thread_sync_t syncMode;
+
 
     public:
 
@@ -62,15 +65,18 @@ class nixlAgentConfig {
          * @param port               specify port for listener thread to listen on
          * @param pthr_delay_us      Optional delay for pthread in us
          * @param pthr_delay_us      Optional delay for listener thread in us
+         * @param sync_mode          Thread synchronization mode
          */
         nixlAgentConfig (const bool use_prog_thread,
                          const bool use_listen_thread=false,
                          const int port=0,
                          const uint64_t pthr_delay_us=0,
-                         const uint64_t lthr_delay_us = 100000) :
+                         const uint64_t lthr_delay_us = 100000,
+                         nixl_thread_sync_t sync_mode=nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT) :
                          useProgThread(use_prog_thread),
                          useListenThread(use_listen_thread),
                          listenPort(port),
+                         syncMode(sync_mode),
                          pthrDelay(pthr_delay_us),
                          lthrDelay(lthr_delay_us) { }
 
@@ -87,6 +93,7 @@ class nixlAgentConfig {
         ~nixlAgentConfig () = default;
 
     friend class nixlAgent;
+    friend class nixlAgentData;
 };
 
 #endif
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -63,6 +63,16 @@ typedef enum {
 } nixl_status_t;
 
 /**
+ * @enum nixl_thread_sync_t
+ * @brief An enumeration of supported synchronization modes for NIXL
+ */
+enum class nixl_thread_sync_t {
+    NIXL_THREAD_SYNC_NONE,
+    NIXL_THREAD_SYNC_STRICT,
+    NIXL_THREAD_SYNC_DEFAULT = NIXL_THREAD_SYNC_NONE,
+};
+
+/**
  * @namespace nixlEnumStrings
  * @brief     This namespace to get string representation
  *            of different enums
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -20,6 +20,7 @@
 #include "common/str_tools.h"
 #include "mem_section.h"
 #include "stream/metadata_stream.h"
+#include "sync.h"
 
 constexpr int default_comm_port = 8888;
 
@@ -42,6 +43,7 @@ class nixlAgentData {
     private:
         std::string     name;
         nixlAgentConfig config;
+        nixlLock        lock;
 
         // some handle that can be used to instantiate an object from the lib
         std::map<std::string, void*> backendLibs;
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -60,7 +60,8 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
 /*** nixlAgentData constructor/destructor, as part of nixlAgent's ***/
 nixlAgentData::nixlAgentData(const std::string &name,
                              const nixlAgentConfig &cfg) :
-                                   name(name), config(cfg) {
+                                   name(name), config(cfg), lock(cfg.syncMode)
+{
         memorySection = new nixlLocalSection();
 }
 
@@ -114,6 +115,7 @@ nixlAgent::~nixlAgent() {
 
 nixl_status_t
 nixlAgent::getAvailPlugins (std::vector<nixl_backend_t> &plugins) {
+    NIXL_LOCK_GUARD(data->lock);
     auto& plugin_manager = nixlPluginManager::getInstance();
     plugins = plugin_manager.getLoadedPluginNames();
     return NIXL_SUCCESS;
@@ -125,6 +127,8 @@ nixlAgent::getPluginParams (const nixl_backend_t &type,
                             nixl_b_params_t &params) const {
 
     // TODO: unify to uppercase/lowercase and do ltrim/rtrim for type
+
+    NIXL_LOCK_GUARD(data->lock);
 
     // First try to get options from a loaded plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
@@ -160,6 +164,7 @@ nixlAgent::getBackendParams (const nixlBackendH* backend,
     if (!backend)
         return NIXL_ERR_INVALID_PARAM;
 
+    NIXL_LOCK_GUARD(data->lock);
     mems   = backend->engine->getSupportedMems();
     params = backend->engine->getCustomParams();
     return NIXL_SUCCESS;
@@ -177,6 +182,7 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     std::string           str;
     backend_list_t*       backend_list;
 
+    NIXL_LOCK_GUARD(data->lock);
     // Registering same type of backend is not supported, unlikely and prob error
     if (data->backendEngines.count(type)!=0)
         return NIXL_ERR_INVALID_PARAM;
@@ -257,6 +263,7 @@ nixlAgent::registerMem(const nixl_reg_dlist_t &descs,
     nixl_status_t   ret;
     unsigned int    count = 0;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (!extra_params || extra_params->backends.size() == 0) {
         backend_list = &data->memToBackend[descs.getType()];
         if (backend_list->empty())
@@ -309,6 +316,7 @@ nixlAgent::deregisterMem(const nixl_reg_dlist_t &descs,
     backend_set_t     backend_set;
     nixl_status_t     ret, bad_ret=NIXL_SUCCESS;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (!extra_params || extra_params->backends.size() == 0) {
         backend_set_t* avail_backends;
         avail_backends = data->memorySection->queryBackends(
@@ -340,6 +348,7 @@ nixlAgent::makeConnection(const std::string &remote_agent,
     std::set<nixl_backend_t> backend_set;
     int count = 0;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (data->remoteBackends.count(remote_agent) == 0)
         return NIXL_ERR_NOT_FOUND;
 
@@ -384,6 +393,7 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
     int            count = 0;
     bool           init_side = (agent_name == NIXL_INIT_AGENT);
 
+    NIXL_LOCK_GUARD(data->lock);
     // When central KV is supported, still it should return error,
     // just we can add a call to fetchRemoteMD for next time
     if (!init_side && (data->remoteSections.count(agent_name) == 0))
@@ -469,6 +479,7 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
     if ((!local_side->isLocal) || (remote_side->isLocal))
         return NIXL_ERR_INVALID_PARAM;
 
+    NIXL_LOCK_GUARD(data->lock);
     // The remote was invalidated in between prepXferDlist and this call
     if (data->remoteSections.count(remote_side->remoteAgent) == 0) {
         delete req_hndl;
@@ -621,6 +632,7 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
 
     req_hndl = nullptr;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (data->remoteSections.count(remote_agent) == 0)
         return NIXL_ERR_NOT_FOUND;
 
@@ -736,6 +748,7 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
     if (!req_hndl)
         return NIXL_ERR_INVALID_PARAM;
 
+    NIXL_LOCK_GUARD(data->lock);
     // Check if the remote was invalidated before post/repost
     if (data->remoteSections.count(req_hndl->remoteAgent) == 0) {
         delete req_hndl;
@@ -790,6 +803,7 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
 nixl_status_t
 nixlAgent::getXferStatus (nixlXferReqH *req_hndl) {
 
+    NIXL_LOCK_GUARD(data->lock);
     // If the status is done, no need to recheck.
     if (req_hndl->status != NIXL_SUCCESS) {
         // Check if the remote was invalidated before completion
@@ -808,6 +822,7 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) {
 nixl_status_t
 nixlAgent::queryXferBackend(const nixlXferReqH* req_hndl,
                             nixlBackendH* &backend) const {
+    NIXL_LOCK_GUARD(data->lock);
     backend = data->backendHandles[req_hndl->engine->getType()];
     return NIXL_SUCCESS;
 }
@@ -815,6 +830,7 @@ nixlAgent::queryXferBackend(const nixlXferReqH* req_hndl,
 nixl_status_t
 nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) {
 
+    NIXL_LOCK_GUARD(data->lock);
     //attempt to cancel request
     if(req_hndl->status == NIXL_IN_PROG) {
         req_hndl->status = req_hndl->engine->checkXfer(
@@ -839,6 +855,7 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) {
 
 nixl_status_t
 nixlAgent::releasedDlistH (nixlDlistH* dlist_hndl) const {
+    NIXL_LOCK_GUARD(data->lock);
     delete dlist_hndl;
     return NIXL_SUCCESS;
 }
@@ -850,6 +867,7 @@ nixlAgent::getNotifs(nixl_notifs_t &notif_map,
     nixl_status_t   ret, bad_ret=NIXL_SUCCESS;
     backend_list_t* backend_list;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (!extra_params || extra_params->backends.size() == 0) {
         backend_list = &data->notifEngines;
         if (backend_list->empty())
@@ -900,6 +918,7 @@ nixlAgent::genNotif(const std::string &remote_agent,
     nixlBackendEngine* backend = nullptr;
     backend_list_t*    backend_list;
 
+    NIXL_LOCK_GUARD(data->lock);
     if (!extra_params || extra_params->backends.size() == 0) {
         backend_list = &data->notifEngines;
         if (backend_list->empty())
@@ -935,10 +954,13 @@ nixlAgent::genNotif(const std::string &remote_agent,
 
 nixl_status_t
 nixlAgent::getLocalMD (nixl_blob_t &str) const {
-    // data->connMD was populated when the backend was created
-    size_t conn_cnt = data->connMD.size();
+    size_t conn_cnt;
     nixl_backend_t nixl_backend;
     nixl_status_t ret;
+
+    NIXL_LOCK_GUARD(data->lock);
+    // data->connMD was populated when the backend was created
+    conn_cnt = data->connMD.size();
 
     if (conn_cnt == 0) // Error, no backend supports remote
         return NIXL_ERR_INVALID_PARAM;
@@ -981,6 +1003,8 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
     backend_list_t tmp_list;
     backend_list_t *backend_list;
     nixl_status_t ret;
+
+    NIXL_LOCK_GUARD(data->lock);
 
     if (!extra_params || extra_params->backends.size() == 0) {
         if (descs.descCount() != 0) {
@@ -1111,6 +1135,7 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
     nixlBackendEngine* eng;
     nixl_status_t ret;
 
+    NIXL_LOCK_GUARD(data->lock);
     ret = sd.importStr(remote_metadata);
     if(ret)
         return ret;
@@ -1188,6 +1213,8 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
 
 nixl_status_t
 nixlAgent::invalidateRemoteMD(const std::string &remote_agent) {
+    NIXL_LOCK_GUARD(data->lock);
+
     if (remote_agent == data->name)
         return NIXL_ERR_INVALID_PARAM;
 
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -115,7 +115,6 @@ nixlAgent::~nixlAgent() {
 
 nixl_status_t
 nixlAgent::getAvailPlugins (std::vector<nixl_backend_t> &plugins) {
-    NIXL_LOCK_GUARD(data->lock);
     auto& plugin_manager = nixlPluginManager::getInstance();
     plugins = plugin_manager.getLoadedPluginNames();
     return NIXL_SUCCESS;
@@ -127,8 +126,6 @@ nixlAgent::getPluginParams (const nixl_backend_t &type,
                             nixl_b_params_t &params) const {
 
     // TODO: unify to uppercase/lowercase and do ltrim/rtrim for type
-
-    NIXL_LOCK_GUARD(data->lock);
 
     // First try to get options from a loaded plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
@@ -146,6 +143,8 @@ nixlAgent::getPluginParams (const nixl_backend_t &type,
     if (plugin_handle) {
         params = plugin_handle->getBackendOptions();
         mems   = plugin_handle->getBackendMems();
+
+        NIXL_LOCK_GUARD(data->lock);
 
         // We don't keep the plugin loaded if we didn't have it before
         if (data->backendEngines.count(type) == 0) {
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -49,27 +49,27 @@ nixlPluginHandle::~nixlPluginHandle() {
     }
 }
 
-nixlBackendEngine* nixlPluginHandle::createEngine(const nixlBackendInitParams* init_params) {
+nixlBackendEngine* nixlPluginHandle::createEngine(const nixlBackendInitParams* init_params) const {
     if (plugin_ && plugin_->create_engine) {
         return plugin_->create_engine(init_params);
     }
     return nullptr;
 }
 
-void nixlPluginHandle::destroyEngine(nixlBackendEngine* engine) {
+void nixlPluginHandle::destroyEngine(nixlBackendEngine* engine) const {
     if (plugin_ && plugin_->destroy_engine && engine) {
         plugin_->destroy_engine(engine);
     }
 }
 
-const char* nixlPluginHandle::getName() {
+const char* nixlPluginHandle::getName() const {
     if (plugin_ && plugin_->get_plugin_name) {
         return plugin_->get_plugin_name();
     }
     return "unknown";
 }
 
-const char* nixlPluginHandle::getVersion() {
+const char* nixlPluginHandle::getVersion() const {
     if (plugin_ && plugin_->get_plugin_version) {
         return plugin_->get_plugin_version();
     }
@@ -113,7 +113,7 @@ std::map<nixl_backend_t, std::string> loadPluginList(const std::string& filename
     return plugins;
 }
 
-std::shared_ptr<nixlPluginHandle> nixlPluginManager::loadPluginFromPath(const std::string& plugin_path) {
+std::shared_ptr<const nixlPluginHandle> nixlPluginManager::loadPluginFromPath(const std::string& plugin_path) {
     // Open the plugin file
     void* handle = dlopen(plugin_path.c_str(), RTLD_NOW | RTLD_LOCAL);
     if (!handle) {
@@ -150,7 +150,7 @@ std::shared_ptr<nixlPluginHandle> nixlPluginManager::loadPluginFromPath(const st
     }
 
     // Create and store the plugin handle
-    auto plugin_handle = std::make_shared<nixlPluginHandle>(handle, plugin);
+    auto plugin_handle = std::make_shared<const nixlPluginHandle>(handle, plugin);
 
     return plugin_handle;
 }
@@ -224,7 +224,7 @@ void nixlPluginManager::addPluginDirectory(const std::string& directory) {
     discoverPluginsFromDir(directory);
 }
 
-std::shared_ptr<nixlPluginHandle> nixlPluginManager::loadPlugin(const std::string& plugin_name) {
+std::shared_ptr<const nixlPluginHandle> nixlPluginManager::loadPlugin(const std::string& plugin_name) {
     // Check if the plugin is already loaded
     // Static Plugins are preloaded so return handle
     auto it = loaded_plugins_.find(plugin_name);
@@ -301,7 +301,7 @@ void nixlPluginManager::unloadPlugin(const nixl_backend_t& plugin_name) {
     loaded_plugins_.erase(plugin_name);
 }
 
-std::shared_ptr<nixlPluginHandle> nixlPluginManager::getPlugin(const nixl_backend_t& plugin_name) {
+std::shared_ptr<const nixlPluginHandle> nixlPluginManager::getPlugin(const nixl_backend_t& plugin_name) {
     auto it = loaded_plugins_.find(plugin_name);
     if (it != loaded_plugins_.end()) {
         return it->second;
@@ -309,7 +309,7 @@ std::shared_ptr<nixlPluginHandle> nixlPluginManager::getPlugin(const nixl_backen
     return nullptr;
 }
 
-nixl_b_params_t nixlPluginHandle::getBackendOptions() {
+nixl_b_params_t nixlPluginHandle::getBackendOptions() const {
     nixl_b_params_t params;
     if (plugin_ && plugin_->get_backend_options) {
         return plugin_->get_backend_options();
@@ -317,7 +317,7 @@ nixl_b_params_t nixlPluginHandle::getBackendOptions() {
     return params; // Return empty params if not implemented
 }
 
-nixl_mem_list_t nixlPluginHandle::getBackendMems() {
+nixl_mem_list_t nixlPluginHandle::getBackendMems() const {
     nixl_mem_list_t mems;
     if (plugin_ && plugin_->get_backend_mems) {
         return plugin_->get_backend_mems();
@@ -346,7 +346,7 @@ void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginC
     nixlBackendPlugin* plugin = info.createFunc();
     if (plugin) {
         // Register the loaded plugin
-        auto plugin_handle = std::make_shared<nixlPluginHandle>(nullptr, plugin);
+        auto plugin_handle = std::make_shared<const nixlPluginHandle>(nullptr, plugin);
         loaded_plugins_[name] = plugin_handle;
     }
 }
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -27,7 +27,8 @@
 #include <iostream>
 #include <string>
 #include <map>
-#include <mutex>
+
+using lock_guard = const std::lock_guard<std::mutex>;
 
 // pluginHandle implementation
 nixlPluginHandle::nixlPluginHandle(void* handle, nixlBackendPlugin* plugin)
@@ -159,6 +160,8 @@ std::shared_ptr<const nixlPluginHandle> nixlPluginManager::loadPluginFromPath(co
 void nixlPluginManager::loadPluginsFromList(const std::string& filename) {
     auto plugins = loadPluginList(filename);
 
+    lock_guard lg(lock);
+
     for (const auto& pair : plugins) {
         const std::string& name = pair.first;
         const std::string& path = pair.second;
@@ -209,20 +212,27 @@ void nixlPluginManager::addPluginDirectory(const std::string& directory) {
         return;
     }
 
-    // Check if directory is already in the list
-    for (const auto& dir : plugin_dirs_) {
-        if (dir == directory) {
-            std::cout << "WARNING: Plugin directory already registered: " << directory << std::endl;
-            return;
+    {
+        lock_guard lg(lock);
+
+        // Check if directory is already in the list
+        for (const auto& dir : plugin_dirs_) {
+            if (dir == directory) {
+                std::cout << "WARNING: Plugin directory already registered: " << directory << std::endl;
+                return;
+            }
         }
+
+        // Prioritize the new directory by inserting it at the beginning
+        plugin_dirs_.insert(plugin_dirs_.begin(), directory);
     }
 
-    // Prioritize the new directory by inserting it at the beginning
-    plugin_dirs_.insert(plugin_dirs_.begin(), directory);
     discoverPluginsFromDir(directory);
 }
 
 std::shared_ptr<const nixlPluginHandle> nixlPluginManager::loadPlugin(const std::string& plugin_name) {
+    lock_guard lg(lock);
+
     // Check if the plugin is already loaded
     // Static Plugins are preloaded so return handle
     auto it = loaded_plugins_.find(plugin_name);
@@ -296,10 +306,15 @@ void nixlPluginManager::unloadPlugin(const nixl_backend_t& plugin_name) {
             return;
         }
     }
+
+    lock_guard lg(lock);
+
     loaded_plugins_.erase(plugin_name);
 }
 
 std::shared_ptr<const nixlPluginHandle> nixlPluginManager::getPlugin(const nixl_backend_t& plugin_name) {
+    lock_guard lg(lock);
+
     auto it = loaded_plugins_.find(plugin_name);
     if (it != loaded_plugins_.end()) {
         return it->second;
@@ -324,6 +339,8 @@ nixl_mem_list_t nixlPluginHandle::getBackendMems() const {
 }
 
 std::vector<nixl_backend_t> nixlPluginManager::getLoadedPluginNames() {
+    lock_guard lg(lock);
+
     std::vector<nixl_backend_t> names;
     for (const auto& pair : loaded_plugins_) {
         names.push_back(pair.first);
@@ -332,6 +349,8 @@ std::vector<nixl_backend_t> nixlPluginManager::getLoadedPluginNames() {
 }
 
 void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginCreatorFunc creator) {
+    lock_guard lg(lock);
+
     nixlStaticPluginInfo info;
     info.name = name;
     info.createFunc = creator;
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -333,9 +333,6 @@ std::vector<nixl_backend_t> nixlPluginManager::getLoadedPluginNames() {
     return names;
 }
 
-// Static Plugin Helpers
-std::vector<nixlStaticPluginInfo> nixlPluginManager::static_plugins_;
-
 void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginCreatorFunc creator) {
     nixlStaticPluginInfo info;
     info.name = name;
@@ -351,7 +348,7 @@ void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginC
     }
 }
 
-std::vector<nixlStaticPluginInfo>& nixlPluginManager::getStaticPlugins() {
+const std::vector<nixlStaticPluginInfo>& nixlPluginManager::getStaticPlugins() {
     return static_plugins_;
 }
 
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -27,6 +27,7 @@
 #include <iostream>
 #include <string>
 #include <map>
+#include <mutex>
 
 // pluginHandle implementation
 nixlPluginHandle::nixlPluginHandle(void* handle, nixlBackendPlugin* plugin)
@@ -184,17 +185,14 @@ nixlPluginManager::nixlPluginManager() {
         plugin_dirs_.insert(plugin_dirs_.begin(), plugin_dir);  // Insert at the beginning for priority
         discoverPluginsFromDir(plugin_dir);
     }
+
+    registerBuiltinPlugins();
 }
 
 nixlPluginManager& nixlPluginManager::getInstance() {
+    // Meyers singleton initialization is safe in multi-threaded environment.
+    // Consult standard [stmt.dcl] chapter for details.
     static nixlPluginManager instance;
-
-    // Only register built-in plugins once
-    static bool registered = false;
-    if (!registered) {
-        instance.registerBuiltinPlugins();
-        registered = true;
-    }
 
     return instance;
 }
--- a/src/core/plugin_manager.h
+++ b/src/core/plugin_manager.h
@@ -67,6 +67,7 @@ private:
     std::vector<nixlStaticPluginInfo> static_plugins_;
 
     void registerBuiltinPlugins();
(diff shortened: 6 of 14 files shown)


meson_options.txt
```diff
@@ -21,3 +21,4 @@ option('cudapath_inc', type: 'string', value: '', description: 'Include path for
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
 option('static_plugins', type: 'string', value: '', description: 'Plugins to be built-in, comma-separated')
+option('disable_mt', type : 'boolean', value : false, description : 'Disable multi-threading support')
```
tstamler: let's make this `value: true` until we have a version we know has good performance

src/core/nixl_agent.cpp
```diff
@@ -61,7 +61,11 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
 /*** nixlAgentData constructor/destructor, as part of nixlAgent's ***/
 nixlAgentData::nixlAgentData(const std::string &name,
                              const nixlAgentConfig &cfg) :
-                                   name(name), config(cfg) {
+                                   name(name), config(cfg)
+#ifndef DISABLE_MT
+                                   , lock(cfg.syncMode)
+#endif
```
tstamler: I don't like the styling here, can we just move the lock initialization into the constructor body?

tstamler (commented):
The initial implementation is mostly fine, but we definitely will want a new test to demonstrate that it actually works

aranadive:
@w1ldptr is there any advantage to having a sync mode as opposed to always enabling locking with disable_mt=false? Having the lock/unlock be noops with MT enabled and SYNC_MODE_NONE, seems odd to me. From a performance perspective, just being able to measure with and without locking seems clearer. If we drop the sync mode altogether, it would make the code cleaner too.

src/core/sync.h
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        void unlock() {
+            if (syncMode == NIXL_SYNC_STRICT) {
+                m.unlock();
+            }
+        }
+};
+
+#define NIXL_LOCK_GUARD(lock) const std::lock_guard<nixlLock> lock_guard##__FILE__##__LINE__ (lock)
```
etoledano: Can we make this a MACRO in some utils.h?
```
#define CONCAT(a, b) CONCAT_0(a, b)
#define CONCAT_0(a, b) a ## b
```

src/api/cpp/nixl_params.h
```diff
@@ -43,15 +43,20 @@ class nixlAgentConfig {
...
+
         /**
          * @brief  Agent configuration constructor. Important configs such as
          *         useProgThread must be given and can't be changed.
          * @param use_prog_thread  flag to determine use of progress thread
          * @param pthr_delay_us    Optional delay for pthread in us
+         * @param sync_mode        Synchronization mode
          */
-        nixlAgentConfig(const bool use_prog_thread, const uint64_t pthr_delay_us=0) {
+        nixlAgentConfig(const bool use_prog_thread, const uint64_t pthr_delay_us=0, const nixl_sync_t sync_mode=NIXL_SYNC_DEFAULT) {
```
etoledano: `nixl_sync_t` is primitive, why do you need to make it const?

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,24 @@ typedef enum {
     NIXL_ERR_NOT_SUPPORTED = -9
 } nixl_status_t;
 
+/**
+ * @enum nixl_sync_t
+ * @brief An enumeration of supported synchronization modes for NIXL
+ */
+typedef enum {
```
etoledano: 1. Can we make it safer using `enum class` ?
2. If yes for 1, no need to typedef I assume

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,24 @@ typedef enum {
...
+    NIXL_SYNC_NONE,
+#ifndef DISABLE_MT
+    NIXL_SYNC_STRICT,
+#endif
+} nixl_sync_t;
+
+#ifdef DISABLE_MT
+#define NIXL_SYNC_DEFAULT NIXL_SYNC_NONE
```
etoledano: it will probably preferred using `inline static constexpr` if possible.
Since DISABLE_MT is a boolean option, suggesting to use it to not need the extra definition
```
enum class nixlSyncMode {
 Strict = 0,
 None = 1,
 Default = DISABLE_MT,
};

```

Or keep the variable with
```
enum class nixlSyncMode {
 Strict = 0,
 None = 1,
};

inline static constexpr nixlSyncMode default_sync_mode = static_cast<nixlSyncMode>(ENABLE_MT);

```

If we don't need DISABLE_MT flag eventually, we shouldn't probably pick a default.

src/core/nixl_agent.cpp
```diff
@@ -100,6 +104,7 @@ nixlAgent::~nixlAgent() {
 
 nixl_status_t
 nixlAgent::getAvailPlugins (std::vector<nixl_backend_t> &plugins) {
+    NIXL_LOCK_GUARD(data->lock);
```
etoledano: If I understood the roles  split correctly, agent data contains the actual data that agent uses and agent is a wrapper for API, in this case why should we lock in agent and not in agent data?

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,24 @@ typedef enum {
...
 } nixl_status_t;
 
+/**
+ * @enum nixl_sync_t
+ * @brief An enumeration of supported synchronization modes for NIXL
+ */
+typedef enum {
+    NIXL_SYNC_NONE,
+#ifndef DISABLE_MT
+    NIXL_SYNC_STRICT,
```
etoledano: Reaching to minimize the amount of `#ifndef` around the code, can we allow this value and do some static assert?

src/api/cpp/nixl_params.h
```diff
@@ -43,15 +43,20 @@ class nixlAgentConfig {
...
             this->useProgThread = use_prog_thread;
             this->pthrDelay     = pthr_delay_us;
+            this->syncMode      = sync_mode;
```
etoledano: I am not sure I understand why/how we allow this to be dynamic without being exposed to a deadlock?

src/core/nixl_agent.cpp
```diff
@@ -853,6 +873,7 @@ nixlAgent::getNotifs(nixl_notifs_t &notif_map,
     nixl_status_t   ret, bad_ret=NIXL_SUCCESS;
     backend_list_t* backend_list;
 
+    NIXL_LOCK_GUARD(data->lock);
```
yosefe, line 869: getNotifs function is blocking since UCX backend calls progress in a while loop. IMO it can be deadlock-prone to hold a lock while doing blocking progress.

src/core/sync.h
```diff
@@ -0,0 +1,56 @@
...
+#else
+
+#include "nixl_params.h"
+
+#include <mutex>
+
+class nixlLock {
+    private:
+        nixl_sync_t syncMode;
+        std::mutex m;
```
yosefe: consider moving to absl::SpinLock (when added) since it doesn't have futex syscall overhead

meson.build
```diff
@@ -58,6 +58,10 @@ if get_option('disable_gds_backend')
     add_project_arguments('-DDISABLE_GDS_BACKEND', language: 'cpp')
 endif
 
+if get_option('disable_mt')
```
yosefe: consider making it a "positive" parameter  (enable_mt)

src/core/agent_data.h
```diff
@@ -19,13 +19,17 @@
...
+#include "sync.h"
 
 typedef std::vector<nixlBackendEngine*> backend_list_t;
 
 class nixlAgentData {
     private:
         std::string     name;
         nixlAgentConfig config;
+#ifndef DISABLE_MT
+        nixlLock        lock;
```
etoledano, line 46: I think we can keep this lock regardless of the compilation flag to not create different conditions depending on an optional argument, not sure why do we need DISABLE_MT.

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,24 @@ typedef enum {
...
+#endif
+} nixl_sync_t;
+
+#ifdef DISABLE_MT
```
yosefe: given that we probably want to ship thread-safe library eventually, i'd make only NIXL_LOCK_GUARD macro depend on DISABLE_MT (or ENABLE_MT) build config. Having readable code and stable API/ABI is IMO more important than avoiding the overheads of std::mutex member in nixl agent and extra parameters for slow-path functions.

meson_options.txt
etoledano: Why do we need both a flag and a configuration for agent?

w1ldptr:
> The initial implementation is mostly fine, but we definitely will want a new test to demonstrate that it actually works

Hey, I marked it as draft in order to discuss just discuss C++ approach to the issue with Eylon first. This draft is quite rough and have issues (including not dealing with nixlPluginManager::getInstance() singleton safely). Sorry for the confusion.

src/core/nixl_agent.cpp
w1ldptr: In C++ all non-POD class members are initialized in initializer list or with their default constructors, if omitted from the list. So we can't initialize them in the conscructor body, we can only assign a new value to an already initialized instance. Since std::mutex doesn't implement assignment operator for obvious reason nixlLock also has it implicitly deleted.

I'll fix the issue by moving ifdefs from here to sync.h and make them noop when MT is disabled.

src/api/cpp/nixl_params.h
w1ldptr: For consistency with other parameters here.

src/api/cpp/nixl_types.h
w1ldptr: This can be passed dynamically by the user on agent initialization. How do you static assert at runtime?

src/api/cpp/nixl_params.h
w1ldptr: What is the deadlock scenario you have in mind?

src/core/agent_data.h
w1ldptr: Yes, that is what I'm doing to address Timothy's comment. Will move all ifdefs to sync.h by making nixlLock compile to zero-sized class with noop functions when MT is disabled.

src/core/nixl_agent.cpp
w1ldptr: Thanks for catching this. I'll take a look into UCX engine implementation.

src/core/sync.h
w1ldptr: With this one-big-lock approach we have huge sections of code being synchronized including ones with memory allocations and file system access. This seems like an poor fit for spinlocks that tend to be used for small sections of code without syscalls, if at all. IMO it is better to leave spinlocks as a future optimization opportunity for some more permissive synchronization model.

src/core/sync.h
w1ldptr: Files like "util.h" tend to quickly become poorly maintained dumpsters of random code pieces. Can add "macro.h" in src/utils/common I guess?

src/api/cpp/nixl_types.h
w1ldptr: We sure can! I implemented it like this only for consistency with existing enums in nixl_types.h.

src/api/cpp/nixl_types.h
w1ldptr: Yep, good idea.

src/api/cpp/nixl_types.h
etoledano: Right, missed that part where its dynamically changed.

src/api/cpp/nixl_params.h
etoledano: Basically a configuration change from Strict to None while the mutex is locked and some other thread is held, can result in not releasing it, assuming the plan was to propagate the sync configuration to the lock itself.

src/api/cpp/nixl_params.h
etoledano: Not sure why those are const too, but I don't think this justify

src/api/cpp/nixl_params.h
w1ldptr: I don't believe any of the config values are dynamically propagated after objects that use them have been created. E.g. useProgThread and pthrDelay are not propagated to engine instance after its creation in createBackend(). Same here - nixlLock locally copies sync_mode value during initialization and don't change it during object's lifetime, so I don't see how the scenario you described could be possible.

src/core/nixl_agent.cpp
w1ldptr: nixlAgentData just holds all the data that would normally be stored in nixlAgent but doesn't implement any kind of logic or provide accessors/mutators.

w1ldptr:
> @w1ldptr is there any advantage to having a sync mode as opposed to always enabling locking with disable_mt=false? Having the lock/unlock be noops with MT enabled and SYNC_MODE_NONE, seems odd to me. From a performance perspective, just being able to measure with and without locking seems clearer. If we drop the sync mode altogether, it would make the code cleaner too.

So, having locking toggle in both build and runtime options was something that @artpol84 insisted on. I don't have any strong preferences regarding this.

meson_options.txt
w1ldptr: See my reply to @aranadive

src/api/cpp/nixl_types.h
w1ldptr: See my reply to @aranadive regarding this.

Please note that it is not a given that we can have a clear split between fast and slow paths and/or avoid all locking on the fast path.

src/core/sync.h
etoledano: Unfortunately `macro.h` doesn't sound too resilient :)
I was actually going to create this MACRO anyway to handle `absl::StatusOr`, I assume this one definitely belongs to `util.h`, we would just need to maintain good granularity when adding stuff for auxiliary.

meson.build
artpol84: Agree

artpol84:
@aranadive @tstamler @yosefe @etoledano 

My reasoning for the dynamic selection was the following:
In the future, we see more "relaxed" MT modes coming that would allow to squeeze more performance if the app is ready to provide certain guarantees.
Still, the strict MT mode that we implement here will have value for apps that don't want to worry about these guarantees.

Given that we need a way to dynamically choose between different MT modes, it makes sense to also provide a no-MT mode as it comes for free, and I can see the use of it in case several agents might be created with one being MT-aware and another not.

I'm still carefully going through the PR and will try to comment in-place too.

src/api/cpp/nixl_params.h
etoledano: I would expect to see some enforcement this isn't being changed dynamically, either by using templates values or by having the member `const`

etoledano:
> @aranadive @tstamler @yosefe @etoledano
> 
> My reasoning for the dynamic selection was the following: In the future, we see more "relaxed" MT modes coming that would allow to squeeze more performance if the app is ready to provide certain guarantees. Still, the strict MT mode that we implement here will have value for apps that don't want to worry about these guarantees.
> 
> Given that we need a way to dynamically choose between different MT modes, it makes sense to also provide a no-MT mode as it comes for free, and I can see the use of it in case several agents might be created with one being MT-aware and another not.
> 
> I'm still carefully going through the PR and will try to comment in-place too.

Seems that if we could implement it such that when its turned off the overhead will be zero (looks very possible), MT is yet another knob in the configuration. The downside for having features depending on compilation flags is huge starting from dragging #ifdef around the code and maintain several different compiled versions, which also makes testing much more complex.

artpol84:
> Seems that if we could implement it such that when its turned off the overhead will be zero (looks very possible), MT is yet another knob in the configuration. The downside for having features depending on compilation flags is huge starting from dragging #ifdef around the code and maintain several different compiled versions, which also makes testing much more complex.

Well ... ideally we shouldn't have any `ifdef`'s in the code other than locking code.
Everything is macros. In this case, it's best to have that option IMO.
I believe that UCX has that and I'd have NIXL to support it too.

aranadive:
> > Seems that if we could implement it such that when its turned off the overhead will be zero (looks very possible), MT is yet another knob in the configuration. The downside for having features depending on compilation flags is huge starting from dragging #ifdef around the code and maintain several different compiled versions, which also makes testing much more complex.
> 
> Well ... ideally we shouldn't have any `ifdef`'s in the code other than locking code. Everything is macros. In this case, it's best to have that option IMO. I believe that UCX has that and I'd have NIXL to support it too.

My general opinion here is to always be thread-safe. By having a no-MT configuration, we could be opening NIXL to data races which maybe hard to debug/diagnose if the user doesnt know what they are doing. I think we should only be MT-safe by default. If the performance impact is minimal, why should the user care between MT/non-MT mode in NIXL? If they really want non-MT mode, they can recompile.

src/core/nixl_agent.cpp
w1ldptr: @yosefe could you describe the deadlock scenario in more details? My understanding of UCX implementation is very superficial but after a quick glance at ucp_worker_progress() I couldn't come up with any potential deadlocking scenario when calling it with external lock taken.

w1ldptr:
@aranadive @artpol84 @etoledano @yosefe 

Regarding the toggles and defaults: I ran nixlbench on the cluster and there is not difference in results for single-threaded execution with synchronization enabled and disabled (compile time). Which makes sense considering the price of taking uncontested futex is negligible since it doesn't involve a syscall.

IMO this makes benefits of having all those toggles to disable something that doesn't seem to measurably affect performance a bit unclear.

yosefe:
> @aranadive @artpol84 @etoledano @yosefe
> 
> Regarding the toggles and defaults: I ran nixlbench on the cluster and there is not difference in results for single-threaded execution with synchronization enabled and disabled (compile time). Which makes sense considering the price of taking uncontested futex is negligible since it doesn't involve a syscall.
> 
> IMO this makes benefits of having all those toggles to disable something that doesn't seem to measurably affect performance a bit unclear.

I think we can keep it as runtime flag (so the overhead is 2 branches when not used), to simpily code and distribution process, WDYT?

tstamler:
> > @aranadive @artpol84 @etoledano @yosefe
> > Regarding the toggles and defaults: I ran nixlbench on the cluster and there is not difference in results for single-threaded execution with synchronization enabled and disabled (compile time). Which makes sense considering the price of taking uncontested futex is negligible since it doesn't involve a syscall.
> > IMO this makes benefits of having all those toggles to disable something that doesn't seem to measurably affect performance a bit unclear.
> 
> I think we can keep it as runtime flag (so the overhead is 2 branches when not used), to simpily code and distribution process, WDYT?

Just bumping this so we can get it closed

w1ldptr:
> The initial implementation is mostly fine, but we definitely will want a new test to demonstrate that it actually works

Added a test suite for mt. Also, my changes to nixlbench that allow multi-threaded execution have been merged recently.

w1ldptr:
> > > @aranadive @artpol84 @etoledano @yosefe
> > > Regarding the toggles and defaults: I ran nixlbench on the cluster and there is not difference in results for single-threaded execution with synchronization enabled and disabled (compile time). Which makes sense considering the price of taking uncontested futex is negligible since it doesn't involve a syscall.
> > > IMO this makes benefits of having all those toggles to disable something that doesn't seem to measurably affect performance a bit unclear.
> > 
> > 
> > I think we can keep it as runtime flag (so the overhead is 2 branches when not used), to simpily code and distribution process, WDYT?
> 
> Just bumping this so we can get it closed

Since there has been no answers for almost a week I assume there is no further disagreement on approach suggested by @yosefe

src/core/agent_data.h
w1ldptr: DISABLE_MT has been removed.

src/api/cpp/nixl_params.h
w1ldptr: Marking member 'const' breaks some existing code by implicitly deleting operator= for the class. I made the field 'private' instead which precludes user from changing it dynamically after the class has been initialized.

src/api/cpp/nixl_types.h
w1ldptr: DISABLE_MT has been removed, setting the default to 'Strict'.

src/api/cpp/nixl_types.h
w1ldptr: done

src/core/sync.h
w1ldptr: Implemented util.h with the contents you suggested.

meson_options.txt
w1ldptr: removed the option

src/core/nixl_agent.cpp
w1ldptr: Removed the ifndef.

src/core/nixl_agent.cpp
w1ldptr: Executed millions of iteration of nixlbench with progress thread enabled and disabled without encountering any deadlocks.

meson.build
w1ldptr: Removed the parameter altogether as you suggested in another thread.

src/api/cpp/nixl_params.h
w1ldptr: removed 'const'

artpol84:
While I'd prefer to have an option to disable MT completely, I am OK with having it on always.

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,16 @@ typedef enum {
...
 } nixl_status_t;
 
+/**
+ * @enum nixl_sync_t
+ * @brief An enumeration of supported synchronization modes for NIXL
+ */
+enum class nixl_sync_t {
+    NIXL_SYNC_NONE,
+    NIXL_SYNC_STRICT,
+    NIXL_SYNC_DEFAULT = NIXL_SYNC_STRICT,
```
artpol84: Probably NONE should be the default.

src/api/cpp/nixl_types.h
```diff
@@ -62,6 +62,16 @@ typedef enum {
     NIXL_ERR_NOT_SUPPORTED = -9
 } nixl_status_t;
 
+/**
+ * @enum nixl_sync_t
+ * @brief An enumeration of supported synchronization modes for NIXL
+ */
+enum class nixl_sync_t {
+    NIXL_SYNC_NONE,
```
artpol84: I don't like "SYNC" and would prefer "MT" prefix as IMO it's more straightforward for the end user.
The user doesn't do any synchronization, it's hidden inside NIXL. What's important for the user is
whether or not (and under which conditions) multiple threads can use NIXL.

If we implement the target mode where each user thread will have its own UCX worker, no "synchronization" per se will be needed. However, it's not going to be "NONE" mode, it will be "RELAXED" (or something better).

src/api/cpp/nixl_params.h
```diff
@@ -62,15 +65,18 @@ class nixlAgentConfig {
...
          * @param pthr_delay_us      Optional delay for pthread in us
          * @param pthr_delay_us      Optional delay for listener thread in us
+         * @param sync_mode          Synchronization mode
          */
         nixlAgentConfig (const bool use_prog_thread,
                          const bool use_listen_thread=false,
                          const int port=0,
                          const uint64_t pthr_delay_us=0,
-                         const uint64_t lthr_delay_us = 100000) :
+                         const uint64_t lthr_delay_us = 100000,
```
artpol84, line 74: Unrelated to this PR, but just noticed a typo in this parameter, should be `pthr`

src/core/nixl_agent.cpp
artpol84: I think that we resolved this question during the API discussion.
@yosefe  was confused by the fact that we are calling progress in the loop.
However, there the loop is active only while there is some IB work to do. Once nothing is observed - the loop is exited making getNotifs async and non-blocking.

src/core/sync.h
artpol84: IMO, the spin-lock can be an option, but I'd recommend using mutex as we might run into the overhead caused by the core being unnecessarily held by a spin lock.
Unlike typical HPC scenarios, I'm not sure if NIXL threads will always have a dedicated core.

src/utils/common/util.h
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef UTIL_H
+#define UTIL_H
+
+#define CONCAT(a, b) CONCAT_0(a, b)
```
artpol84, line 20: Can you remind me why we can't directly do `a ## b` here?

src/core/plugin_manager.h
```diff
@@ -37,12 +38,12 @@ class nixlPluginHandle {
     nixlPluginHandle(void* handle, nixlBackendPlugin* plugin);
```
artpol84, line 44: I suggest adding a comment with the summary of the class usage WRT MT.
E.g.
```
This class represents a NIXL plugin and is used to create plugin instances.
* nixlPluginHandle attributes are modified only in the constructor and destructor.
* They remain unchanged during normal operation, e.g., query operations and plugin instance creation
This allows using it in multi-threading environments without lock protection.
```

src/core/nixl_plugin_manager.cpp
```diff
@@ -187,14 +191,10 @@ nixlPluginManager::nixlPluginManager() {
 }
 
 nixlPluginManager& nixlPluginManager::getInstance() {
+    static std::once_flag registered;
     static nixlPluginManager instance;
```
artpol84, line 198: While call_once protects built-in plugin initialization, two threads entering this function simultaneously will cause both to call the instance constructor.
Should we move all the plugin discovery logic to the lambda expression in `call_once` and explicitly delete the constructor?

As `loadPlugin()` is thread-safe, the existing version is likely correct. However, I believe doing everything in the context of call_once will be more elegant.
If we know that only one instance of the original initialization is called, we can do the initial plugin load without any lock (via lock-free helper functions). And later use the same helper functions in the thread-safe versions.

src/core/nixl_agent.cpp
```diff
@@ -114,6 +115,7 @@ nixlAgent::~nixlAgent() {
 
 nixl_status_t
 nixlAgent::getAvailPlugins (std::vector<nixl_backend_t> &plugins) {
+    NIXL_LOCK_GUARD(data->lock);
```
artpol84: Do we need locking here? Plugin manager locking seems sufficient.

src/core/nixl_agent.cpp
```diff
@@ -126,6 +128,8 @@ nixlAgent::getPluginParams (const nixl_backend_t &type,
 
     // TODO: unify to uppercase/lowercase and do ltrim/rtrim for type
 
+    NIXL_LOCK_GUARD(data->lock);
```
artpol84: Same here - it seems that plugin manager locking is enough.

It might be required later in this function when backend engines are accessed.

src/api/cpp/nixl_types.h
w1ldptr: Don't you think "MT" would be a misnomer here? We don't actually control any "multi-threading" with this enum (like creating an internal thread pool and spawning bunch of threads to speed up processing or something), all we do is to provide _synchronization_ in our API so the _user_ could call it in multi-threaded environment safely. And multiple user threads can already use NIXL, user would just have to synchronize access to it by themselves. Same as when sharing an instance of any other class that doesn't have internal synchronization (like std container, for example).

I don't have strong opinion either was just my 2 cents...

PS I also don't understand how any kind of UCX mode would alleviate the need to synchronize the agent since UCX is neither the only engine nor can it magically protect data in nixlAgentData from concurrent modification. But that would be a subject to discuss during the review of such changes (which could involve renaming of any enum values that no longer correctly represent the API behavior).

src/core/nixl_plugin_manager.cpp
w1ldptr: Isn't this just a typical Meyers singleton which is thread safe in C++11 and later? Am I missing some crucial detail here?

https://stackoverflow.com/questions/1661529/is-meyers-implementation-of-the-singleton-pattern-thread-safe

src/utils/common/util.h
w1ldptr: I'm not an expert in C pre-processor but AFAIK two stage approach is used here so that when CONCAT() arguments are macros themselves they would be evaluated first before the concatenation.

src/core/nixl_agent.cpp
w1ldptr: Your are correct - locking is redundant here.

src/core/nixl_agent.cpp
w1ldptr: Yep, this one is also redundant.

src/api/cpp/nixl_types.h
w1ldptr: Ack

src/core/plugin_manager.h
w1ldptr: Sure, makes sense.

src/core/nixl_plugin_manager.cpp
```diff
@@ -150,14 +152,16 @@ std::shared_ptr<nixlPluginHandle> nixlPluginManager::loadPluginFromPath(const st
...
-    auto plugin_handle = std::make_shared<nixlPluginHandle>(handle, plugin);
+    auto plugin_handle = std::make_shared<const nixlPluginHandle>(handle, plugin);
 
     return plugin_handle;
 }
 
 void nixlPluginManager::loadPluginsFromList(const std::string& filename) {
     auto plugins = loadPluginList(filename);
 
+    lock_guard lg(lock);
```
etoledano, line 163: Why does `nixlPluginManager` behaves different than `nixlAgent` for locking?

test/gtest/mocks/mock_dram_engine.h
```diff
@@ -19,19 +19,35 @@
...
+  MockDramBackendEngine(const nixlBackendInitParams *init_params) : nixlBackendEngine(init_params), sharedState(1) {}
   ~MockDramBackendEngine();
 
-  bool supportsRemote() const override;
-  bool supportsLocal() const override;
-  bool supportsNotif() const override;
-  bool supportsProgTh() const override;
-  nixl_mem_list_t getSupportedMems() const override;
+  bool supportsRemote() const override {
+    assert(sharedState > 0);
```
etoledano, line 32: Can we adopt `absl::CHECK/PCHECK/QCHECK/DCHECK` (its a MACRO, blocks are just to emphasize).

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "plugin_manager.h"
+#include <thread>
+
+namespace gtest {
+namespace mt {
+
+class Mt : public testing::Test {
+protected:
+    // Helper method to create an agent with default config
+    static nixlAgent createAgent() {
```
etoledano: Why is that static?

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+        nixlDescList<nixlBasicDesc> dst_list(DRAM_SEG);
+
+        // Create basic descriptor for transfer
+        nixlBasicDesc basic_desc(0, 1024, 0);
+        src_list.addDesc(basic_desc);
+        dst_list.addDesc(basic_desc);
+
+        nixl_status_t status = agent.createXferReq(op, src_list, dst_list, "test_agent", xfer_req, &extra_params);
+        EXPECT_EQ(status, NIXL_SUCCESS);
+        EXPECT_NE(xfer_req, nullptr);
```
etoledano: `EXPECT_EQ` doesn't allow this?
```
EXPECT_EQ(agent.createXferReq(op, src_list, dst_list, "test_agent", xfer_req, &extra_params), NIXL_SUCCESS);
```

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    // Helper method to perform a transfer operation
+    static void performTransfer(nixlAgent& agent, const nixl_opt_args_t& extra_params) {
+        nixlXferReqH* xfer_req = nullptr;
+        nixl_xfer_op_t op = NIXL_WRITE;
+        nixlDescList<nixlBasicDesc> src_list(DRAM_SEG);
+        nixlDescList<nixlBasicDesc> dst_list(DRAM_SEG);
+
+        // Create basic descriptor for transfer
+        nixlBasicDesc basic_desc(0, 1024, 0);
```
etoledano: Can this magic numbers be in an unnamed namespace?

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+        desc_list.addDesc(blob);
+
+        nixl_status_t status = agent.registerMem(desc_list, &extra_params);
+        EXPECT_EQ(status, NIXL_SUCCESS);
+    }
+
+    // Helper method to perform a transfer operation
+    static void performTransfer(nixlAgent& agent, const nixl_opt_args_t& extra_params) {
+        nixlXferReqH* xfer_req = nullptr;
+        nixl_xfer_op_t op = NIXL_WRITE;
```
etoledano: Not needed seems like

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+class Mt : public testing::Test {
+protected:
+    // Helper method to create an agent with default config
+    static nixlAgent createAgent() {
+        nixlAgentConfig cfg(false);
+        return nixlAgent("test_agent", cfg);
+    }
+
+    // Helper method to create a mock DRAM backend
+    static nixlBackendH* createMockDramBackend(nixlAgent& agent) {
```
etoledano: Including testing symbols in utilities function requires verbosity for the least, perhaps this?
`verifyMockDramBackendCreation`
(all around)

test/gtest/mocks/mock_dram_engine.h
```diff
@@ -66,6 +87,9 @@ class MockDramBackendEngine : public nixlBackendEngine {
   nixl_status_t genNotif(const std::string &remote_agent,
                          const std::string &msg) override;
   int progress() override;
+
+private:
```
etoledano, line 91: I see you declared `private` before `public` somewhere else, let's keep it consistent.

test/gtest/mocks/mock_dram_engine.h
```diff
@@ -66,6 +87,9 @@ class MockDramBackendEngine : public nixlBackendEngine {
...
+  int sharedState;
```
etoledano, line 94: description?

src/core/nixl_agent.cpp
```diff
@@ -956,11 +975,14 @@ nixlAgent::genNotif(const std::string &remote_agent,
...
 nixl_status_t
 nixlAgent::getLocalMD (nixl_blob_t &str) const {
-    // data->connMD was populated when the backend was created
-    size_t conn_cnt = data->connMD.size();
+    size_t conn_cnt;
     nixl_backend_t nixl_backend;
     nixl_status_t ret;
 
+    NIXL_LOCK_GUARD(data->lock);
+    // data->connMD was populated when the backend was created
```
etoledano, line 961: How this access is different than any `data->` one that a comment justified?

src/core/nixl_plugin_manager.cpp
```diff
@@ -150,14 +152,16 @@ std::shared_ptr<nixlPluginHandle> nixlPluginManager::loadPluginFromPath(const st
     }
 
     // Create and store the plugin handle
-    auto plugin_handle = std::make_shared<nixlPluginHandle>(handle, plugin);
+    auto plugin_handle = std::make_shared<const nixlPluginHandle>(handle, plugin);
```
etoledano, line 155: `const auto&` is a bit more verbose I assume.

src/core/plugin_manager.h
```diff
@@ -56,11 +57,10 @@ struct nixlStaticPluginInfo {
...
 class nixlPluginManager {
 private:
-    std::map<nixl_backend_t, std::shared_ptr<nixlPluginHandle>> loaded_plugins_;
+    std::map<nixl_backend_t, std::shared_ptr<const nixlPluginHandle>> loaded_plugins_;
     std::vector<std::string> plugin_dirs_;
-
-    // Static Plugins
-    static std::vector<nixlStaticPluginInfo> static_plugins_;
+    std::vector<nixlStaticPluginInfo> static_plugins_;
+    std::mutex lock;
```
etoledano, line 69: All around,
Can we take advantage of [absl::mutex](https://abseil.io/docs/cpp/guides/synchronization#abslmutex-and-stdmutex)? deadlock detection sounds good.

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    auto transfer_sequence = []() {
+        nixlAgent agent = createAgent();
+        nixlBackendH* backend = createMockDramBackend(agent);
+        nixl_opt_args_t extra_params = createExtraParams(backend);
+
+        registerMemory(agent, extra_params);
+        performTransfer(agent, extra_params);
+    };
+
+    // Create and run two threads executing the transfer sequence
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    std::thread t1(transfer_sequence);
+    std::thread t2(transfer_sequence);
+
+    // Wait for both threads to complete
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+        status = agent.getXferStatus(xfer_req);
+        EXPECT_EQ(status, NIXL_SUCCESS);
+
+        status = agent.releaseXferReq(xfer_req);
+        EXPECT_EQ(status, NIXL_SUCCESS);
+    }
+};
+
+TEST_F(Mt, ConcurrentTransfersWithPerThreadAgent) {
+    // Function to execute transfer sequence with per-thread agent and memory registration
```
etoledano: `Function` -> `Test`
And before the test makes more sense I guess?

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    nixlBackendH* backend = createMockDramBackend(agent);
+    nixl_opt_args_t extra_params = createExtraParams(backend);
+
+    // Function to execute transfer sequence with per-thread memory registration
+    auto transfer_sequence = [&agent, &extra_params]() {
+        registerMemory(agent, extra_params);
+        performTransfer(agent, extra_params);
+    };
+
+    // Create and run two threads executing the transfer sequence
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    std::thread t1(transfer_sequence);
+    std::thread t2(transfer_sequence);
+
+    // Wait for both threads to complete
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    nixl_opt_args_t extra_params = createExtraParams(backend);
+
+    registerMemory(agent, extra_params);
+
+    // Function to execute transfer sequence
+    auto transfer_sequence = [&agent, &extra_params]() {
+        performTransfer(agent, extra_params);
+    };
+
+    // Create and run two threads executing the transfer sequence
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+    std::thread t1(transfer_sequence);
+    std::thread t2(transfer_sequence);
+
+    // Wait for both threads to complete
```
etoledano: [Don’t say in comments what can be clearly stated in code](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+ */
+#include <gtest/gtest.h>
+#include "nixl.h"
+#include "plugin_manager.h"
+#include <thread>
+
+namespace gtest {
+namespace mt {
+
+class Mt : public testing::Test {
```
etoledano: Can we add `Fixture` to the class name? seems like a common thing for these.
Also, `MultiThreading` doesn't seem too long to justify `mt`

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+        performTransfer(agent, extra_params);
+    };
+
+    // Create and run two threads executing the transfer sequence
+    std::thread t1(transfer_sequence);
+    std::thread t2(transfer_sequence);
+
+    // Wait for both threads to complete
+    t1.join();
+    t2.join();
```
etoledano: I would assume to see some verification of `sharedState == num_threads`, no?

test/gtest/mt.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
```
etoledano: Same here, I don't think a test in gtest should be written to only rely on TSAN, it needs to have a verification stage

In this case, does it makes sense to increase the number of threads?

test/gtest/mt.cpp
w1ldptr: No, why? I'm not verifying any kind of white box nixlAgent behavior (how many times it calls dram mock and/or which functions) to check for that, such tests would be very non-specific and broad. These are to trigger Thread sanitizer splats in case of unsafe modification of shared state.

test/gtest/mt.cpp
w1ldptr: It doesn't really make sense. Thread sanitizer doesn't need actual race condition to occur because it tracks accesses to memory and verifies which locks were held at that moment. Doesn't matter whether it is 2 or 2000000 threads, 1 or 1000000 iterations....

test/gtest/mt.cpp
w1ldptr: Ack

test/gtest/mt.cpp
w1ldptr: This is just arbitrary numbers that are irrelevant to the functionality being validated. Wouldn't it be complete over-engineering to refactor them out into some constexprs contained in anonymous namespace and then bickering over their naming for several more days? 😁

test/gtest/mt.cpp
w1ldptr: Yep, agree.

test/gtest/mt.cpp
w1ldptr: ditto

src/core/nixl_plugin_manager.cpp
w1ldptr: Because it is a singleton. User might want to initialize multiple instances of nixlAgent with different synchronization parameters (Artem gave me an example of such case when it is done indirectly by using several libraries that both use NIXL internally) but since the nixlPluginManager is only accessible as a global singleton instance we need to ensure it is always safely shared.

test/gtest/mocks/mock_dram_engine.h
w1ldptr: Currently I found 317 mentions of string "assert" in the library. I suggest dedicated standalone refactoring PR for this may be more appropriate instead of just starting using another macros for checks in a few random places.

test/gtest/mt.cpp
w1ldptr: I guess it does. I just liked having the expects on different lines because it reads better in the cases when there are multiple expects for a single agent function call (like this one), so I put all expects on another line for consistency. Don't have a strong opinion either way.

test/gtest/mt.cpp
etoledano: If the case is that TSAN is the verifier, then this test shouldn't exist if TSAN doesn't.
If this test is a standalone, it should verify something, currently it doesn't.

test/gtest/mt.cpp
etoledano: If they don't matter it should be clear, otherwise their enchantments are assumed to overpower the reader's mind :)

src/core/nixl_plugin_manager.cpp
etoledano: Maybe the question should have been "why are we making the internal class `nixlPluginManager` thread safe?"
A proper use of it in `nixlAgent` shouldn't break NIXL's sustainability, no?

test/gtest/mt.cpp
w1ldptr: This is exactly the case - as you can see in meson file the test is not included when the project is compiled without tsan.

src/core/nixl_plugin_manager.cpp
w1ldptr: Again, there could be multiple nixlAgents sharing nixlPluginManager (because it is a singleton). The only other viable option is to refactor nixlAgent to use static lock that is always enabled which has the following drawbacks:

1. The lock has to be always enabled because otherwise nixlPluginManager singleton usage would race with any other nixlAgent in the application (even indirectly included by some dependency).
2. All nixlAgent instances would be globally synchronized which would reduce performance by precluding any kind of parallelism between them.

test/gtest/mt.cpp
w1ldptr: Well, it is used further down as a parameter, but I can just pass NIXL_WRITE directly.

test/gtest/mt.cpp
w1ldptr: ack

test/gtest/mocks/mock_dram_engine.h
w1ldptr: Preferred order is: public, protected, private? Right? I'll go over the patch series to verify it is maintained.

test/gtest/mocks/mock_dram_engine.h
w1ldptr: will add

src/core/nixl_agent.cpp
w1ldptr: Look at the full diff - I moved access to data from conn_cnt declaration at the beginning of the function to here together with existing comment.

src/core/nixl_plugin_manager.cpp
w1ldptr: It is not the same. Here we have a mutable shared_ptr (we can assign it with new object or whatever) that we hold by-value (we created it with make_shared and we own it) that points to a const instance of nixlPluginHandle (so we can't call non-const methods of that instance via operator-> of shared_ptr).

test/gtest/mt.cpp
w1ldptr: Removed the comment altogether since it appears superfluous considering ConcurrentTransfersWithPerThreadAgent already explains it as it is.

src/core/nixl_plugin_manager.cpp
artpol84: I was relying on the following (pretty old) discussion:
https://stackoverflow.com/questions/4590626/c-static-variables-in-multithreaded-program

Seems like it was later fixed and MT safety is ensured.
In this case, why are we messing with "call once" at all? Can't we just initialize everything in the constructor?

artpol84: I suggest putting in the comment the explanation on why it's ok and the reference to the standard

artpol84: Or just do everything in call_once - then it's explicit and there will be no concerns

src/utils/common/util.h
artpol84: I remember stumbling on this as well, but don't remember the details. So I thought you would know off the top of your head.

src/core/nixl_plugin_manager.cpp
w1ldptr: nixlPluginManager ctor already has bunch of init code. 
Rationale for putting another part of the init code into registerBuiltinPlugins() isn't obvious to me so I am hesitant to change it without need.

src/core/nixl_plugin_manager.cpp
```diff
@@ -187,14 +191,10 @@ nixlPluginManager::nixlPluginManager() {
...
+    static std::once_flag registered;
     static nixlPluginManager instance;
 
-    // Only register built-in plugins once
-    static bool registered = false;
-    if (!registered) {
-        instance.registerBuiltinPlugins();
-        registered = true;
-    }
+    std::call_once(registered, [&]() { instance.registerBuiltinPlugins(); });
```
artpol84: @aranadive can we move this into the constructor?

src/api/cpp/nixl_types.h
w1ldptr: Renamed to nixl_thread_sync_* according to our discussion.

src/core/nixl_plugin_manager.cpp
w1ldptr: Moved registerBuiltinPlugins to ctor.

test/gtest/mt.cpp
w1ldptr: Right, removed static.

test/gtest/mt.cpp
w1ldptr: Extracted magic numbers to class fields.

src/core/plugin_manager.h
w1ldptr: TBH I'm not entirely convinced that it is worth it to use Abseil implementations just for the sake of it, e.g. in cases like this one when we just using the same functionality as already provided by std or when it isn't justified by differences in performance between their respective implementations.

However, since it appears to be general project direction to use Abseil as much as possible I've added the change as a dedicated commit in the series so that the implementation would remain valid with or without it and leaving it to the consideration of maintainers.

artpol84 (approved):

src/core/plugin_manager.h
aranadive: The deadlock detection isnt enabled by default in non-debug builds so we will need additional flags there. I agree with Vlad here that if there are things provided by std we use those over abseil unless the abseil provided ones have a high quantified benefit. That doesnt seem to be the case here - the std::mutex seems good enough.

artpol84:
\ok to test 888bb2d27fdb4e53f1641daaf008dbbf06377a5e

artpol84:
\ok to test 888bb2d27fdb4e53f1641daaf008dbbf06377a5e

src/api/cpp/nixl_params.h
tstamler: not a typo, see #167

src/core/plugin_manager.h
artpol84: Agree with @aranadive

artpol84 (approved):
@tstamler looks good!

aranadive (approved):

PR #157, opened 2025-04-16
BINDINGS/RUST: added rust bindings

roiedanino:
Added rust binding package, taken from dynamo: 
https://github.com/ai-dynamo/dynamo/tree/phendricks/genericize-kvblock/lib/bindings/nixl-sys


Cargo.lock
tstamler, line 1: how did you generate this lock file? It looks like it has a lot of windows packages, and we are not building/supporting NIXL for windows right now

Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
+members = ["src/bindings/rust"]
+
+[workspace.package]
+description = "NIXL - NVIDIA Interconnect Library"
```
tstamler: NVIDIA Inference Xfer Library...

meson.build
```diff
@@ -40,6 +41,12 @@ else
     include_directories : include_directories(cuda_inc_path))
 endif
 
+# Add pthread dependency for thread support
+threads_dep = dependency('threads')
```
tstamler: these should not be necessary except for very old versions of GCC and STL, they should be automatically included in STL

meson.build
```diff
@@ -40,6 +41,12 @@ else
...
+
+# Add dl library for dynamic loading
```
tstamler: same as above

src/bindings/python/__init__.py
tstamler, line 1: we already have an __init__.py in the api/python directory, this isn't needed

src/bindings/rust/Cargo.toml
tstamler, line 1: why do we need 2 Cargo.tomls?

src/bindings/rust/build.rs
```diff
@@ -0,0 +1,43 @@
...
+        .extra_warnings(true)
+        .flag("-std=c++17")
+        .flag("-fPIC")
+        .flag("-Wno-unused-parameter")
+        .flag("-Wno-unused-variable")
+        .include("../../api/cpp")  // Add path to nixl.h
+        .include("../../core")     // Add path to other headers
+        .include("../../infra")    // Add path to other headers
+        .file("wrapper.cpp")
+        .compile("wrapper");
```
tstamler, line 40: is this how we should be building the wrapper? is meson not able to do this?

src/bindings/rust/meson.build
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Find cargo
+cargo = find_program('cargo', required: true)
+
+# Build the Rust library
+rust_lib = custom_target('nixl-sys',
+  output: ['libnixl_sys.a'],
+  input: ['Cargo.toml', 'build.rs', 'wrapper.h', 'wrapper.cpp'] + files('src/lib.rs'),
```
tstamler: I guess same comment as above, mostly just curious if this is the standard way to include C++ wrappers in a Rust distribution in meson

src/bindings/rust/src/lib.rs
tstamler, line 1: I don't know enough about rust to properly review this file

src/bindings/rust/wrapper.cpp
tstamler, line 1: again, I don't have enough expertise to say if this is the best way to do anything, but it feels like this will be painful to maintain

src/core/meson.build
```diff
@@ -33,7 +33,7 @@ nixl_lib = library('nixl',
                    'nixl_agent.cpp',
                    'nixl_plugin_manager.cpp',
                    include_directories: [ nixl_inc_dirs, utils_inc_dirs ],
-                   dependencies: nixl_lib_deps,
+                   dependencies: [nixl_lib_deps, dl_dep],
```
tstamler: not going to comment on all of these and then the other pthread dependencies, but as I said I've only seen this be necessary with old Linux distributions (specifically manylinux environment) which should be an exceptional case and not the standard. Modern C++ will have these dependencies included automatically

tstamler (changes requested):
two final comments:

the wrapper.cpp/h files are C bindings, which are not rust specific. These can be useful for generating other bindings, or from users who just want to use C, and should have a dedicated C bindings directory.

I thought Ryan had some tests, not seeing any here. Either look in dynamo or follow up with him, we should not be merging new bindings/APIs without any kind of testing or documentation.

roiedanino:
@tstamler Yes, this is still a mess, forgot to mark it as a draft, sorry

src/bindings/rust/build.rs
ryanolson: we can compile the c wrapper anywhere. 

I'd almost rather have it built separately and found alongside libnixl.so as libnixl-capi.so or something similar.

src/core/meson.build
yosefe: i've seen we need libdl and pthread on redhat8

ryanolson:
Required for https://github.com/ai-dynamo/dynamo/pull/758

src/bindings/rust_nixl/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[package]
+name = "rust_nixl"
+version = "0.1.0"
+edition = "2021"
```
tstamler: let's make version match NIXL version (0.1.1) and will update for 0.2.0. I'm also guessing we can update edition to be current year

roiedanino: Regarding the edition it means the cargo edition not this crate edition

ryanolson: the crate should be called "nixl-sys" this is the common pattern used in rust when there should be a c api present and installed and the crate using some mechanism to find the .h and .so, then triggered the bindgen.

ryanolson: currently, the crate I provided also compiles the c api, but those should be separated

tstamler (approved):

src/bindings/rust_nixl/Cargo.toml
tstamler: We are moving the separation of the C API out to a future release, because its not up to date with the current NIXL API and will take significant work to update. Until we can have a complete, verified C API we don't want to advertise it as a set of standalone bindings. 

Otherwise, let's update the name and then merge.

src/bindings/rust_nixl/build.rs
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    let nixl_root =  env::var("NIXL_PREFIX")
```
gleon99: Minor: excessive space

gleon99: +Add some `_dir` or `_path` prefixes to these vars

src/bindings/rust_nixl/build.rs
```diff
@@ -0,0 +1,69 @@
...
+        .flag("-std=c++17")
+        .flag("-fPIC")
+        .include(&nixl_include)
+        // Change ABI flag if necessary to match your precompiled libraries:
+            //    .flag("-D_GLIBCXX_USE_CXX11_ABI=0")
+        .flag("-Wno-unused-parameter")
+        .flag("-Wno-unused-variable")
+        .compile("wrapper");
+
+    // Link against NIXL libraries in correct order
```
gleon99: Duplicates lines 29-31?

src/bindings/rust_nixl/build.rs
gleon99: Why removed `pkg_config`? Maybe restore it, as the 1st prio method?

src/bindings/rust_nixl/meson.build
```diff
@@ -0,0 +1,49 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+rustc = find_program('rustc', required: false)
+cargo = find_program('cargo', required: false)
+
+
+nixl_include_dir = include_directories('../../api/cpp', '../../infra', '../../core')
```
gleon99: Can we avoid relative paths here and use `meson.project_source_root()` instead?

src/bindings/rust_nixl/src/lib.rs
```diff
@@ -0,0 +1,2209 @@
...
+            .next()
+            .unwrap()
+            .unwrap();
+        assert_eq!(notif, b"notification");
+
+        // Verify memory patterns
+        assert!(storage1.as_slice().iter().all(|&x| x == 0xbb));
+        assert!(storage2.as_slice().iter().all(|&x| x == 0xbb));
+
+        drop(xfer_req);
```
gleon99: Are these really needed?

roiedanino: @ryanolson ?

src/bindings/rust_nixl/meson.build
roiedanino:
```
../src/bindings/rust/meson.build:21:19: ERROR: Tried to form an absolute path to a dir in the source tree.
You should not do that but use relative paths instead, for
directories that are part of your project.
```

src/bindings/rust_nixl/src/lib.rs
ryanolson: for right now keep them.  I think that removing the drop will error due to the borrow checker.

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+    FailedToCreateXferRequest,
+    #[error("Failed to create registration descriptor list")]
+    RegDescListCreationFailed,
+    #[error("Failed to add registration descriptor")]
+    RegDescAddFailed,
+}
+
+/// Memory types supported by NIXL
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MemType {
```
mkhazraee, line 97: Why unknown? there are 5 specific ones.

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+    Dram,
+    Gpu,
+}
+
+impl From<nixl_capi_mem_type_t> for MemType {
+    fn from(mem_type: nixl_capi_mem_type_t) -> Self {
+        match mem_type {
+            1 => MemType::Dram,
+            2 => MemType::Gpu,
+            _ => MemType::Unknown,
```
mkhazraee, line 114: Same here. Not sure even we want the unknow as 6th.

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+    }
+}
+
+impl fmt::Display for MemType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // SAFETY: We know the memory type is valid and the string will be available
+        let mut str_ptr = ptr::null();
+        unsafe {
+            let mem_type = match self {
+                MemType::Unknown => 0,
```
mkhazraee: again not sure why unknown, and not all 5 types.

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+                    _phantom: std::marker::PhantomData,
+                })
+            }
+            -1 => Err(NixlError::InvalidParam),
+            _ => Err(NixlError::BackendError),
+        }
+    }
+}
+
+impl MemList {
```
mkhazraee, line 1098: follow up on mem_list issue.

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+        }
+    }
+
+    /// Add a descriptor from a type implementing NixlDescriptor
+    ///
+    /// # Safety
+    /// The caller must ensure that:
+    /// - The descriptor remains valid for the lifetime of the list
+    /// - The memory region pointed to by the descriptor remains valid
+    pub fn add_storage_desc<D: NixlDescriptor + 'a>(
```
mkhazraee, line 1270: storage?

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+    fn size(&self) -> usize;
+}
+
+/// A trait for types that can be added to NIXL descriptor lists
+pub trait NixlDescriptor: MemoryRegion {
+    /// Get the memory type for this descriptor
+    fn mem_type(&self) -> MemType;
+
+    /// Get the device ID for this memory region
+    fn device_id(&self) -> u32;
```
mkhazraee: updated to be u64

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+        Some(self.data.as_ptr())
+    }
+}
+
+impl NixlDescriptor for SystemStorage {
+    fn mem_type(&self) -> MemType {
+        MemType::Dram
+    }
+
+    fn device_id(&self) -> u32 {
```
mkhazraee: u64

src/bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,2197 @@
...
+        Ok(RegistrationHandle {
+            agent: self.inner.clone(),
+            ptr: unsafe { descriptor.as_ptr() }.ok_or(NixlError::InvalidParam)? as usize,
+            size: descriptor.size(),
+            dev_id: descriptor.device_id(),
+            mem_type: descriptor.mem_type(),
+        })
+    }
+
+    pub fn register_memory_with_args(
```
mkhazraee, line 528: Why we have two methods for register, one with args and one without, but not for transfer handles?

src/bindings/rust/src/lib.rs
mkhazraee, line 1: Really hard to read file, should be split to a few files. One for descriptors, one for tests at least. Also it would be good to follow the order we have in nixl.h or _api.py. I'm not sure we're exposing all the APIs, as hard to follow.

src/bindings/rust/wrapper.h
```diff
@@ -0,0 +1,195 @@
...
+  NIXL_CAPI_SUCCESS = 0,
+  NIXL_CAPI_ERROR_INVALID_PARAM = -1,
+  NIXL_CAPI_ERROR_BACKEND = -2,
+  NIXL_CAPI_ERROR_INVALID_STATE = -3,
+  NIXL_CAPI_IN_PROG = 1,
+} nixl_capi_status_t;
+
+// Memory types enum (matching nixl's memory types)
+typedef enum {
+  NIXL_CAPI_MEM_UNKNOWN = 0,
```
mkhazraee: Not sure why we have UNKNOWN and not the 5 that we should

src/bindings/rust/wrapper.h
```diff
@@ -0,0 +1,195 @@
...
+nixl_capi_status_t nixl_capi_post_xfer_req(
+    nixl_capi_agent_t agent, nixl_capi_xfer_req_t req_hndl, nixl_capi_opt_args_t opt_args);
+
+nixl_capi_status_t nixl_capi_get_xfer_status(nixl_capi_agent_t agent, nixl_capi_xfer_req_t req_hndl);
+
+nixl_capi_status_t nixl_capi_release_xfer_req(nixl_capi_agent_t agent, nixl_capi_xfer_req_t req);
+
+nixl_capi_status_t nixl_capi_destroy_xfer_req(nixl_capi_xfer_req_t req);
+
+// Descriptor list functions
```
mkhazraee, line 167: Let's be consistent with nixl_bindings.cpp in src/bindings/python. Several important ones like gtetType or trim or verifySorted are missing, while I don't think we want to expose resize and has_overlaps here.

src/bindings/rust/src/lib.rs
roiedanino: `device_id` seems to be `uint32_t` in the entire repo

src/bindings/rust/wrapper.h
roiedanino: Hi Moein, I've added missing memory types, changed device id from u32 to u64, and moved tests to `tests.rs`.
Regarding consistency with Python bindings, I'm not sure we will be able to do it in time for the release, therefore, we should probably do another PR fixing that in the next one. What do you think?

gleon99: @mkhazraee to proceed with merging this PR, we prefer to postpone improvements.

tstamler (approved):

src/bindings/rust/src/lib.rs
```diff
@@ -95,16 +95,22 @@ pub enum NixlError {
 /// Memory types supported by NIXL
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum MemType {
-    Unknown,
     Dram,
-    Gpu,
+    Vram,
+    Bulk,
```
mkhazraee: This is block not bulk.

PR #159, opened 2025-04-16
Agent API: New method to get partial MD and allow partial MD loads

yanivz-nv:
* This allows one agent to notify remote agents about new memory regions
* This allows guardrails between agents so that each remote agent can transfer only to a subset of regions.
* Introducing new internal nixlSectionDesc which is used to cache the remote metadata blob to make sure an existing remote descriptor did not change its metadata during a new loadRemoteMD.

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -365,6 +365,26 @@ class nixlAgent {
         getLocalMD (nixl_blob_t &str) const;
 
         /**
+         * @brief  Get partial metadata blob for this agent, to be given to other agents.
+         *         If `descs` is empty, only backends' connection info is included in the metadata,
+         *         regardless of the value of `extra_params->includeConnInfo` and `descs` memory type.
+         *         If `descs` is non-empty, the metadata of the descriptors in the list are included,
+         *         and if `extra_params->includeConnInfo` is true, the connection info of the
+         *         backends supporting the memory type is also included.
+         *         If `extra_params->backends` is non-empty, only the descriptors supported by the
+         *         backends in the list and the backends' connection info are included in the metadata.
+         *
+         * @param  descs         [in]  Descriptor list to include in the metadata
+         * @param  str           [out] The serialized metadata blob
+         * @param  extra_params  [in]  Optional extra parameters used in getting partial metadata
+         * @return nixl_status_t       Error code if call was not successful
+         */
+        nixl_status_t
+        getLocalPartialMD(nixl_reg_dlist_t  &descs,
+                          nixl_blob_t &str,
+                          const nixl_opt_args_t* extra_params = nullptr) const;
+
+        /**
          * @brief  Load other agent's metadata and unpack it internally. Now the local
          *         agent can initiate transfers towards the remote agent.
          *
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -134,6 +134,12 @@ class nixlAgentOptionalArgs {
          * @var makeXferReq boolean to skip merging consecutive descriptors, used in makeXferReq.
          */
         bool skipDescMerge = false;
+
+        /**
+         * @var includeConnInfo boolean to include connection information in the metadata,
+         *                      used in getLocalPartialMD.
+         */
+        bool includeConnInfo = false;
 };
 /**
  * @brief A typedef for a nixlAgentOptionalArgs
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -53,12 +53,13 @@ class nixlAgentData {
 
         // Bookkeping for local connection metadata and user handles per backend
         std::unordered_map<nixl_backend_t, nixlBackendH*> backendHandles;
-        std::unordered_map<nixl_backend_t, std::string>   connMD;
+        std::unordered_map<nixl_backend_t, nixl_blob_t>   connMD;
 
         // Local section, and Remote sections and their available common backends
         nixlLocalSection*                                        memorySection;
 
-        std::unordered_map<std::string, std::set<nixl_backend_t>,
+        std::unordered_map<std::string,
+                           std::unordered_map<nixl_backend_t, nixl_blob_t>,
                            std::hash<std::string>, strEqual>     remoteBackends;
         std::unordered_map<std::string, nixlRemoteSection*,
                            std::hash<std::string>, strEqual>     remoteSections;
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -272,8 +272,8 @@ nixlAgent::registerMem(const nixl_reg_dlist_t &descs,
     for (size_t i=0; i<backend_list->size(); ++i) {
         nixlBackendEngine* backend = (*backend_list)[i];
         // meta_descs use to be passed to loadLocalData
-        nixl_meta_dlist_t meta_descs(descs.getType(), false);
-        ret = data->memorySection->addDescList(descs, backend, meta_descs);
+        nixl_sec_dlist_t sec_descs(descs.getType(), false);
+        ret = data->memorySection->addDescList(descs, backend, sec_descs);
         if (ret == NIXL_SUCCESS) {
             if (backend->supportsLocal()) {
                 if (data->remoteSections.count(data->name) == 0)
@@ -281,7 +281,7 @@ nixlAgent::registerMem(const nixl_reg_dlist_t &descs,
                           new nixlRemoteSection(data->name);
 
                 ret = data->remoteSections[data->name]->loadLocalData(
-                                                        meta_descs, backend);
+                                                        sec_descs, backend);
                 if (ret == NIXL_SUCCESS)
                     count++;
                 else
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -987,6 +987,81 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
 }
 
 nixl_status_t
+nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
+                             nixl_blob_t &str,
+                             const nixl_opt_args_t* extra_params) const {
+    backend_list_t tmp_list;
+    backend_list_t *backend_list;
+    nixl_status_t ret;
+
+    if (!extra_params || extra_params->backends.size() == 0) {
+        if (descs.descCount() != 0) {
+            // Non-empty dlist, return backends that support the memory type
+            backend_list = &data->memToBackend[descs.getType()];
+            if (backend_list->empty())
+                return NIXL_ERR_NOT_FOUND;
+        } else {
+            // Empty dlist, return all backends
+            backend_list = &tmp_list;
+            for (const auto & elm : data->backendEngines)
+                backend_list->push_back(elm.second);
+        }
+    } else {
+        backend_list = &tmp_list;
+        for (const auto & elm : extra_params->backends)
+            backend_list->push_back(elm->engine);
+    }
+
+    // First find all relevant engines and their conn info.
+    // Best effort, ignore if no conn info.
+    backend_set_t selected_engines;
+    std::vector<typename decltype(data->connMD)::iterator> found_iters;
+    for (const auto &backend : *backend_list) {
+        auto it = data->connMD.find(backend->getType());
+        if (it == data->connMD.end())
+            continue;
+        found_iters.push_back(it);
+        selected_engines.insert(backend);
+    }
+
+    nixlSerDes sd;
+    ret = sd.addStr("Agent", data->name);
+    if(ret)
+        return ret;
+
+    // Only add connection info if requested via extra_params or empty dlist
+    size_t conn_cnt = ((extra_params && extra_params->includeConnInfo) || descs.descCount() == 0) ?
+                      found_iters.size() : 0;
+    ret = sd.addBuf("Conns", &conn_cnt, sizeof(conn_cnt));
+    if(ret)
+        return ret;
+
+    for (size_t i = 0; i < conn_cnt; i++) {
+        ret = sd.addStr("t", found_iters[i]->first);
+        if(ret)
+            return ret;
+        ret = sd.addStr("c", found_iters[i]->second);
+        if(ret)
+            return ret;
+    }
+
+    // No engines found, but there are descs, this is an error
+    if (selected_engines.size() == 0 && descs.descCount() > 0)
+        return NIXL_ERR_BACKEND;
+
+    ret = sd.addStr("", "MemSection");
+    if(ret)
+        return ret;
+
+    ret = data->memorySection->serializePartial(&sd, selected_engines, descs);
+    if(ret)
+        return ret;
+
+    str = sd.exportStr();
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
 nixlAgent::sendLocalMD (const std::string remote_ip, const int port) const {
 
     if(remote_ip.size() == 0){
@@ -1063,10 +1138,6 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
     if(ret)
         return ret;
 
-    // TODO: add support for marginal updates, then conn_cnt might be 0
-    if (conn_cnt<1)
-        return NIXL_ERR_INVALID_PARAM;
-
     for (size_t i=0; i<conn_cnt; ++i) {
         nixl_backend = sd.getStr("t");
         if (nixl_backend.size() == 0)
@@ -1101,7 +1172,7 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
     }
 
     // No common backend, no point in loading the rest, unexpected
-    if (count == 0)
+    if (count == 0 && conn_cnt > 0)
         return NIXL_ERR_BACKEND;
 
     if (sd.getStr("") != "MemSection")
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1001,7 +1001,7 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
     }
 
     // First find all relevant engines and their conn info.
-    // Best effort, ignore if no conn info.
+    // Best effort, ignore if no conn info (meaning backend doesn't support remote).
     backend_set_t selected_engines;
     std::vector<typename decltype(data->connMD)::iterator> found_iters;
     for (const auto &backend : *backend_list) {
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -285,7 +285,7 @@ nixlAgent::registerMem(const nixl_reg_dlist_t &descs,
                 if (ret == NIXL_SUCCESS)
                     count++;
                 else
-                    data->memorySection->remDescList(meta_descs, backend);
+                    data->memorySection->remDescList(descs, backend);
             } else {
                 count++;
             }
@@ -308,7 +308,6 @@ nixlAgent::deregisterMem(const nixl_reg_dlist_t &descs,
 
     backend_set_t     backend_set;
     nixl_status_t     ret, bad_ret=NIXL_SUCCESS;
-    nixl_xfer_dlist_t trimmed = descs.trim();
 
     if (!extra_params || extra_params->backends.size() == 0) {
         backend_set_t* avail_backends;
@@ -325,15 +324,7 @@ nixlAgent::deregisterMem(const nixl_reg_dlist_t &descs,
 
     // Doing best effort, and returning err if any
     for (auto & backend : backend_set) {
-        nixl_meta_dlist_t resp(descs.getType(),
-                               descs.isSorted());
-
-        // Now just populating the metadata part, inside remDescList,
-        // getIndex is used to make sure exact match
-        ret = data->memorySection->populate(trimmed, backend, resp);
-        if (ret != NIXL_SUCCESS)
-            bad_ret = ret;
-        ret = data->memorySection->remDescList(resp, backend);
+        ret = data->memorySection->remDescList(descs, backend);
         if (ret != NIXL_SUCCESS)
             bad_ret = ret;
     }
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -337,24 +337,24 @@ nixlAgent::makeConnection(const std::string &remote_agent,
                           const nixl_opt_args_t* extra_params) {
     nixlBackendEngine* eng;
     nixl_status_t ret;
-    std::set<nixl_backend_t>* backend_set;
+    std::set<nixl_backend_t> backend_set;
     int count = 0;
 
     if (data->remoteBackends.count(remote_agent) == 0)
         return NIXL_ERR_NOT_FOUND;
 
     if (!extra_params || extra_params->backends.size() == 0) {
-        backend_set = &data->remoteBackends[remote_agent];
-        if (backend_set->empty())
+        if (data->remoteBackends[remote_agent].empty())
             return NIXL_ERR_NOT_FOUND;
+        for (auto & [r_bknd, conn_info] : data->remoteBackends[remote_agent])
+            backend_set.insert(r_bknd);
     } else {
-        backend_set = new std::set<nixl_backend_t>();
         for (auto & elm : extra_params->backends)
-            backend_set->insert(elm->engine->getType());
+            backend_set.insert(elm->engine->getType());
     }
 
     // For now trying to make all the connections, can become best effort,
-    for (auto & backend: *backend_set) {
+    for (auto & backend: backend_set) {
         if (data->backendEngines.count(backend)!=0) {
             eng = data->backendEngines[backend];
             ret = eng->connect(remote_agent);
@@ -363,9 +363,6 @@ nixlAgent::makeConnection(const std::string &remote_agent,
             count++;
         }
     }
-
-    if (extra_params && extra_params->backends.size() > 0)
-        delete backend_set;
 
     if (ret)
         return ret;
@@ -1109,7 +1106,7 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
     int count = 0;
     nixlSerDes sd;
     size_t conn_cnt;
-    std::string conn_info;
+    nixl_blob_t conn_info;
     nixl_backend_t nixl_backend;
     nixlBackendEngine* eng;
     nixl_status_t ret;
@@ -1140,12 +1137,14 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
         // Current agent might not support a remote backend
         if (data->backendEngines.count(nixl_backend)!=0) {
 
-            // No need to reload same conn info, (TODO to cache the old val?)
-            if (data->remoteBackends.count(remote_agent)!=0)
-                if (data->remoteBackends[remote_agent].count(nixl_backend)!=0) {
-                    count++;
-                    continue;
-                }
+            // No need to reload same conn info, error if it changed
+            if (data->remoteBackends.count(remote_agent) != 0 &&
+                data->remoteBackends[remote_agent].count(nixl_backend) != 0) {
+                if (data->remoteBackends[remote_agent][nixl_backend] != conn_info)
+                    return NIXL_ERR_NOT_ALLOWED;
+                count++;
+                continue;
+            }
 
             eng = data->backendEngines[nixl_backend];
             if (eng->supportsRemote()) {
@@ -1153,7 +1152,7 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
                 if (ret)
                     return ret; // Error in load
                 count++;
-                data->remoteBackends[remote_agent].insert(nixl_backend);
+                data->remoteBackends[remote_agent].emplace(nixl_backend, conn_info);
             } else {
                 // If there was an issue and we return error while some connections
                 // are loaded, they will be deleted in the backend destructor.
@@ -1200,8 +1199,8 @@ nixlAgent::invalidateRemoteMD(const std::string &remote_agent) {
     }
 
     if (data->remoteBackends.count(remote_agent)!=0) {
-        for (auto & elm: data->remoteBackends[remote_agent])
-            data->backendEngines[elm]->disconnect(remote_agent);
+        for (auto & it: data->remoteBackends[remote_agent])
+            data->backendEngines[it.first]->disconnect(remote_agent);
         data->remoteBackends.erase(remote_agent);
         ret = NIXL_SUCCESS;
     }
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -66,7 +66,7 @@ class nixlLocalSection : public nixlMemSection {
                                    nixl_meta_dlist_t &remote_self);
 
         // Each nixlBasicDesc should be same as original registration region
-        nixl_status_t remDescList (const nixl_meta_dlist_t &mem_elms,
+        nixl_status_t remDescList (const nixl_reg_dlist_t &mem_elms,
                                    nixlBackendEngine* backend);
 
         nixl_status_t serialize(nixlSerDes* serializer) const;
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -44,6 +44,8 @@ public:
     using nixlMetaDesc::nixlMetaDesc;
 
     nixl_blob_t serialize() const {
+        // Serialize only the meta blob. metadataP is private so we don't include it.
+        // The other side will deserialize it as nixlBlobDesc.
         return nixlBasicDesc::serialize() + metaBlob;
     }
 
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -79,9 +79,6 @@ class nixlMemSection {
 
 
 class nixlLocalSection : public nixlMemSection {
-    private:
-        nixl_status_t serializeSections(nixlSerDes* serializer,
-                                        const section_map_t &sections) const;
     public:
         nixl_status_t addDescList (const nixl_reg_dlist_t &mem_elms,
                                    nixlBackendEngine* backend,
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -67,6 +67,10 @@ class nixlLocalSection : public nixlMemSection {
 
         nixl_status_t serialize(nixlSerDes* serializer) const;
 
+        nixl_status_t serializePartial(nixlSerDes* serializer,
+                                       const backend_set_t &backends,
+                                       const nixl_reg_dlist_t &mem_elms) const;
+
         ~nixlLocalSection();
 };
 
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -30,12 +30,13 @@
 typedef std::pair<nixl_mem_t, nixlBackendEngine*>              section_key_t;
 typedef std::set<nixlBackendEngine*>                           backend_set_t;
 typedef std::unordered_map<nixl_backend_t, nixlBackendEngine*> backend_map_t;
+typedef std::map<section_key_t,   nixl_meta_dlist_t*>          section_map_t;
 
 
 class nixlMemSection {
     protected:
         std::array<backend_set_t, FILE_SEG+1>         memToBackend;
-        std::map<section_key_t,   nixl_meta_dlist_t*> sectionMap;
+        section_map_t                                 sectionMap;
 
     public:
         nixlMemSection () {};
@@ -56,6 +57,9 @@ class nixlLocalSection : public nixlMemSection {
         nixl_reg_dlist_t getStringDesc (
                                const nixlBackendEngine* backend,
                                const nixl_meta_dlist_t &d_list) const;
+
+        nixl_status_t serializeSections(nixlSerDes* serializer,
+                                        const section_map_t &sections) const;
     public:
         nixl_status_t addDescList (const nixl_reg_dlist_t &mem_elms,
                                    nixlBackendEngine* backend,
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -30,8 +30,34 @@
 typedef std::pair<nixl_mem_t, nixlBackendEngine*>              section_key_t;
 typedef std::set<nixlBackendEngine*>                           backend_set_t;
 typedef std::unordered_map<nixl_backend_t, nixlBackendEngine*> backend_map_t;
-typedef std::map<section_key_t,   nixl_meta_dlist_t*>          section_map_t;
 
+/**
+ * @brief Section descriptor for nixl
+ *
+ * This class is used to store a section descriptor for nixl.
+ * It is derived from nixlMetaDesc and contains the meta blob for the section.
+ */
+class nixlSectionDesc : public nixlMetaDesc {
+public:
+    nixl_blob_t metaBlob;
+
+    using nixlMetaDesc::nixlMetaDesc;
+
+    nixl_blob_t serialize() const {
+        return nixlBasicDesc::serialize() + metaBlob;
+    }
+
+    inline friend bool operator==(const nixlSectionDesc &lhs, const nixlSectionDesc &rhs) {
+        return (static_cast<nixlMetaDesc>(lhs) == static_cast<nixlMetaDesc>(rhs));
+    }
+
+    inline void print(const std::string &suffix) const {
+        nixlMetaDesc::print(", meta blob: " + metaBlob + suffix);
+    }
+};
+
+typedef nixlDescList<nixlSectionDesc>               nixl_sec_dlist_t;
+typedef std::map<section_key_t, nixl_sec_dlist_t*>  section_map_t;
 
 class nixlMemSection {
     protected:
@@ -54,16 +80,12 @@ class nixlMemSection {
 
 class nixlLocalSection : public nixlMemSection {
     private:
-        nixl_reg_dlist_t getStringDesc (
-                               const nixlBackendEngine* backend,
-                               const nixl_meta_dlist_t &d_list) const;
-
         nixl_status_t serializeSections(nixlSerDes* serializer,
                                         const section_map_t &sections) const;
     public:
         nixl_status_t addDescList (const nixl_reg_dlist_t &mem_elms,
                                    nixlBackendEngine* backend,
-                                   nixl_meta_dlist_t &remote_self);
+                                   nixl_sec_dlist_t &remote_self);
 
         // Each nixlBasicDesc should be same as original registration region
         nixl_status_t remDescList (const nixl_reg_dlist_t &mem_elms,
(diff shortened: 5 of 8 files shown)


src/core/nixl_agent.cpp
```diff
@@ -332,24 +331,24 @@ nixlAgent::makeConnection(const std::string &remote_agent,
                           const nixl_opt_args_t* extra_params) {
     nixlBackendEngine* eng;
     nixl_status_t ret;
-    std::set<nixl_backend_t>* backend_set;
+    std::set<nixl_backend_t> backend_set;
```
mkhazraee, line 340: The point was not to do the extra copy unless necessary. Unless there is a crucial reason to. And even then a separate PR not here.

src/core/nixl_agent.cpp
```diff
@@ -16,6 +16,7 @@
  */
 
 #include <iostream>
+#include <cassert>
```
mkhazraee: Why cassert? For this level library asserts don't make much sense, we removed them from another file that had them. We don't interrupt the flow, at most we return errors.

src/api/cpp/nixl_types.h
```diff
@@ -19,6 +19,7 @@
 #include <vector>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
```
mkhazraee: why?

src/core/agent_data.h
```diff
@@ -40,11 +41,12 @@ class nixlAgentData {
         std::unordered_map<nixl_backend_t, std::string>   connMD;
 
         // Local section, and Remote sections and their available common backends
-        nixlLocalSection*                                        memorySection;
+        std::unique_ptr<nixlLocalSection>                        memorySection;
```
mkhazraee: Separate PR, and to be discussed, not relevant to this PR.

src/core/nixl_agent.cpp
```diff
@@ -977,6 +973,55 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::getLocalPartialMD(const std::vector<nixl_backend_t> &backends,
+                             const std::vector<nixl_reg_dlist_t *> &dlist_vec,
+                             nixl_blob_t &str) const {
+    size_t conn_cnt = backends.size();
```
mkhazraee: Let's check which ones do support remote, in case user passes bad parameters.

mkhazraee: Since they might return an empty backend list to say any backend that has these elements, best effort might make more sense, that error out.

src/infra/mem_section.h
```diff
@@ -27,15 +27,18 @@
...
+typedef std::set<nixlBackendEngine*>                              backend_set_t;
+typedef std::unordered_map<nixl_backend_t, nixlBackendEngine*>    backend_map_t;
+typedef std::unordered_map<nixlBackendEngine*, nixl_meta_dlist_t> backend_dlist_map_t;
 
 class nixlMemSection {
     protected:
+        // TODO: merge memToBackend and sectionMap?
         std::array<backend_set_t, FILE_SEG+1>         memToBackend;
-        std::map<section_key_t,   nixl_meta_dlist_t*> sectionMap;
+        std::array<backend_dlist_map_t, FILE_SEG+1>   sectionMap;
```
mkhazraee: Number of backends is small, number of memory types per each is small, let's not have a 2 level map for this. Complicates things down the line. There were lots of back and forth between backends first or memories first, we landed on this, don't change.

test/nixl/nixl_test.cpp
mkhazraee, line 1: Just limit the changes to the MD update PR. If you want to update this test, a separate PR.

src/api/cpp/nixl_descriptors.h
```diff
@@ -232,12 +232,24 @@ class nixlDescList {
          * @param d_list other nixlDescList object of the same type
          */
         nixlDescList(const nixlDescList<T> &d_list) = default;
+        /**
+         * @brief Move constructor for nixlDescList
+         *
+         * @param d_list other nixlDescList object to move from
+         */
+        nixlDescList(nixlDescList<T> &&d_list) = default;
```
mkhazraee: Based on the change to getLocalPartialMD I'm not positive we need this.

src/api/cpp/nixl_descriptors.h
```diff
@@ -232,12 +232,24 @@ class nixlDescList {
...
          *
          * @param d_list nixlDescList object
          */
         nixlDescList& operator=(const nixlDescList<T> &d_list) = default;
+        /**
+         * @brief Move assignment operator for nixlDescList
+         *
+         * @param d_list other nixlDescList object to move from
+         */
+        nixlDescList& operator=(nixlDescList<T> &&d_list) = default;
```
mkhazraee: Based on the change to getLocalPartialMD I'm not positive we need this.

src/api/cpp/nixl.h
```diff
@@ -323,6 +323,20 @@ class nixlAgent {
...
+         *
+         * @param  backends [in]    List of backends to include in the metadata
+         * @param  descs    [in]    List of descriptor lists to include in the metadata
+         * @param  op       [in]    Metadata operation type (load/unload)
+         * @param  str      [out]   The serialized metadata blob
+         * @return nixl_status_t    Error code if call was not successful
+         */
+        nixl_status_t
+        getLocalPartialMD(const std::vector<nixl_backend_t> &backends,
+                          const std::vector<nixl_reg_dlist_t *> &dlist_vec,
```
mkhazraee: I see what you wanted to do here now, but seeing the overheads the flexibility is adding doesn't seem like worth it, with the double pointers added all over. User can specify a descriptor list, and get the metadata for it for specific backends, or all. Simplicity is more important here. Either they want all the info, or they want info for a specific desc list. They can merge the desc lists if they want. 

One thing is we talked about having a parameter that do we want the conn info to be included or not. User can get a conn info only of all the backends if that argument is set to one, but the reg_dlist_t is empty. So that way they can get conn info only, or without conn info, and then for specific memories.

Let's get the extra_params for the backends to be consistent with the rest of methods.

src/core/nixl_agent.cpp
yanivz-nv: I understand this, but now `remoteBackends[remote_agent]` is a map instead of a set because we cache the old conn_info, so I can't simply point to it.
Either we do this, or we need to maintain a new map of maps `remoteBacknedsConnInfo` which is pretty redundant.

src/core/nixl_agent.cpp
yanivz-nv: As far as I'm aware, one can use asserts to validate we don't reach impossible states internally and internal code is not broken, as opposed to illegal user input, and on checks that don't make sense to check on release.
Anyway, removed it and changed to return error.

src/api/cpp/nixl_types.h
yanivz-nv: mb, I used it at first but not eventually and didn't remove...

src/core/agent_data.h
yanivz-nv: Done

src/core/nixl_agent.cpp
yanivz-nv: Done

src/infra/mem_section.h
yanivz-nv: I dropped this commit.
Although I'm not sure it's more beneficial this way, since we use a map (BST) instead of an array + unordered map (array + hash), but now irrelevant for the feature after API update.

test/nixl/nixl_test.cpp
yanivz-nv: ACK

src/api/cpp/nixl_descriptors.h
yanivz-nv: Good practice when defining copy constructors to also define move constructors (even if default). It's not a must for now since the copy is default so I removed it.

src/api/cpp/nixl_descriptors.h
yanivz-nv: Same ^

src/core/agent_data.h
```diff
@@ -42,7 +42,8 @@ class nixlAgentData {
         // Local section, and Remote sections and their available common backends
         nixlLocalSection*                                        memorySection;
 
-        std::unordered_map<std::string, std::set<nixl_backend_t>,
+        std::unordered_map<std::string,
```
mkhazraee, line 61: I now understand why you did this. One thing to note, we avoid std::string when it has another typedef. So the connInfo should be nixl_blob_t practically at the end of this line.

I had the idea of making a struct to bundle the two hashmaps with the same key, but since it were two and also the second map (remoteSections) was tiny, it was useful to have quick checks for existence. Just putting the idea out there if in future we had to add more fields.

src/core/nixl_agent.cpp
mkhazraee: Ugh, this was why I didn't merge the two. But anyways, that's cleaner, and a copy is not that expensive. Let's go with this, but just with the updated struct. I think this is a clean solution.

src/api/cpp/nixl.h
```diff
@@ -323,6 +323,21 @@ class nixlAgent {
...
+         *
+         * @param  descs         [in]  Descriptor list to include in the metadata
+         * @param  get_conn_info [in]  Whether to include connection information in the metadata
+         * @param  str           [out] The serialized metadata blob
+         * @param  extra_params  [in]  Optional extra parameters used in getting partial metadata
+         * @return nixl_status_t       Error code if call was not successful
+         */
+        nixl_status_t
+        getLocalPartialMD(nixl_reg_dlist_t  &descs,
+                          bool get_conn_info,
```
mkhazraee: let's call it include_conn_info, good chance we move it to extra_params, sounds like an optional parameter, but for now it's fine to be here until we finalize the API. Maybe we will end up puting both descs and this in extra_params.

src/api/cpp/backend/backend_aux.h
```diff
@@ -90,6 +90,7 @@ class nixlMetaDesc : public nixlBasicDesc {
   public:
         // To be able to point to any object
         nixlBackendMD* metadataP;
+        nixl_blob_t remoteMetaInfo;
```
mkhazraee: I like this addition, but it would add an empty string to each single descriptor on each side of transfer as we're using nixlMetaDesc for transfer. So let' do 2 things about it:

- let's inherit from nixlMetaDesc, something like nixlSectionDesc. And add this new field there. So for agent bookkeepings, we use this, but when we create a metadesc for a transfer, it won't include this field. I suggest to put the definition of nixlRecordDesc in mem_section, as that's where these book keepings are being done.

- in RegisterMem, I check if it supportsLocal, I call loadLocalMD. Add a check that if supportsRemote, call getPublic data and put that value here. That way you already have the values for all the registered memories cached upon their registration, and in getLocalMD we don't have to call getPublic anymore (this was a todo iirc, we can get rid of that).

src/core/nixl_agent.cpp
mkhazraee: Oh I separated the maps before to enable this, but not a big deal. Generally makeConnection is optional parameter and not in critical path, so fine to have the extra copy.

src/core/agent_data.h
mkhazraee: If other places connInfo is still string and not blob_t, please update in this PR, thanks.

src/api/cpp/nixl.h
mkhazraee: I also ran it by Tim, let's keep the method separate and not merge it with getLocalMD. Just by moving the flag to extra params, it'll become:
nixl_status_t getLocalPartialMD(nixl_reg_dlist_t  &descs, nixl_blob_t &str, const nixl_opt_args_t* extra_params = nullptr) const;

It's similar to registerMem/deregisterMem which is good to be consistent, just has an output. Later we can add the remove flag inside extra_params without cluttering the main API.

src/api/cpp/backend/backend_aux.h
mkhazraee: Actually while at it, let's do a minor fix I had in mind. The reason there is that copyMeta as a separate method instead of just copying the metadata in populate method is we were using c++11 at first, I couldn't do constexpr, so it would always assume the case for if input is nixlBasicDesc, and error out. But if we change the very initial if in populate to constexpr instead of std::is_same<nixlBasicDesc, T>::value, we can just do the copy in populate. After doing this you can remove the copyMeta from nixlBasicDesc/nixlBlobDesc/nixlMetaDesc.

src/infra/nixl_memory_section.cpp
```diff
@@ -195,6 +195,41 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
     return NIXL_SUCCESS;
 }
 
+nixl_status_t nixlLocalSection::serializePartial(nixlSerDes* serializer,
```
mkhazraee, line 295: Maybe we can do some code refactoring, the original serialize is calling this method per backend almost. Since there are both conn part and desc part it's not that obvious what the best refactoring, but seems like doable.

src/infra/nixl_memory_section.cpp
```diff
@@ -195,6 +195,41 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
...
+                                                 const backend_set_t &backends,
+                                                 const nixl_reg_dlist_t &mem_elms) const {
+    // TODO: consider concatenating 2 serializers instead of using mem_elms_to_serialize
+    std::vector<std::pair<nixlBackendEngine*, nixl_reg_dlist_t>> mem_elms_to_serialize;
+    nixl_xfer_dlist_t trimmed = mem_elms.trim();
```
etoledano: Not needed as a stack variable seems like

src/infra/nixl_memory_section.cpp
```diff
@@ -195,6 +195,41 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
...
+    nixl_mem_t nixl_mem = mem_elms.getType();
+    nixl_status_t ret;
+    size_t seg_count;
```
etoledano: Its [encouraged](https://www.reddit.com/r/C_Programming/comments/nquux5/comment/h0gtm4f/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) to declare variables as close as possible to the usage.

src/infra/nixl_memory_section.cpp
```diff
@@ -195,6 +195,41 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
...
+
+    for (auto &backend : backends) {
```
etoledano: `for (const auto& backend: backends) {`

would work?

src/core/nixl_agent.cpp
```diff
@@ -981,6 +981,73 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
...
 
+nixl_status_t
+nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
+                             bool get_conn_info,
+                             nixl_blob_t &str,
+                             const nixl_opt_args_t* extra_params) const {
+    backend_set_t selected_engines;
+    backend_list_t tmp_list;
+    backend_list_t *backend_list;
+    nixl_backend_t nixl_backend;
```
etoledano: Unused

src/core/nixl_agent.cpp
```diff
@@ -981,6 +981,73 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
...
+    } else {
+        backend_list = &tmp_list;
+        for (auto & elm : extra_params->backends)
+            backend_list->push_back(elm->engine);
+    }
+
+    // First find all relevant engines and their conn info.
+    // Best effort, ignore if no conn info.
+    std::vector<typename decltype(data->connMD)::iterator> found_iters;
+    for (auto &backend : *backend_list) {
```
etoledano: ` for (const auto &backend : *backend_list) {`

src/infra/nixl_memory_section.cpp
yanivz-nv: It is a bit expensive operation, we should do it once and not every iteration of the loop

src/api/cpp/nixl.h
yanivz-nv: Changed name and moved to extra_params.

src/core/nixl_agent.cpp
yanivz-nv: Done.

src/core/nixl_agent.cpp
yanivz-nv: Done

src/infra/nixl_memory_section.cpp
yanivz-nv: Added a commit with a refactoring possibility. Additional improvement to it would be that sectionMap is a map of uniqe_ptr / instances instead of pointers, then we avoid new+delete in serializeParital.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done.

src/infra/nixl_memory_section.cpp
yanivz-nv: `backends` is already const so AFAIU it shouldn't matter, but done.

src/core/agent_data.h
yanivz-nv: connMD was string, updated to blob.

src/api/cpp/backend/backend_aux.h
yanivz-nv: Added new nixlSectionDesc, but not sure I understand the fix in the 2nd comment. Also not sure what was the point of copyMeta instead of just implementing a copy constructor and what is different now with constexpr.

src/infra/mem_section.h
```diff
@@ -31,11 +31,43 @@ typedef std::pair<nixl_mem_t, nixlBackendEngine*>              section_key_t;
...
+ * This class is used to store a section descriptor for nixl.
+ * It is derived from nixlMetaDesc and contains the meta blob for the section.
+ */
+class nixlSectionDesc : public nixlMetaDesc {
+public:
+    nixl_blob_t metaBlob;
+
+    using nixlMetaDesc::nixlMetaDesc;
+
+    nixl_blob_t serialize() const {
```
mkhazraee, line 46: Sounds buggy, specially with string headers, we should use proper serializer class maybe.

src/infra/mem_section.h
```diff
@@ -31,11 +31,43 @@ typedef std::pair<nixl_mem_t, nixlBackendEngine*>              section_key_t;
...
+        return nixlBasicDesc::serialize() + metaBlob;
+    }
+
+    inline friend bool operator==(const nixlSectionDesc &lhs, const nixlSectionDesc &rhs) {
+        return (static_cast<nixlMetaDesc>(lhs) == static_cast<nixlMetaDesc>(rhs));
+    }
+
+    inline void copyMeta(const nixlSectionDesc &meta) {
```
mkhazraee: We said to remove. Unless there was some other use case

src/infra/mem_section.h
```diff
@@ -31,11 +31,43 @@ typedef std::pair<nixl_mem_t, nixlBackendEngine*>              section_key_t;
...
+        nixlMetaDesc::copyMeta(meta);
+        metaBlob = meta.metaBlob;
+    }
+
+    inline void print(const std::string &suffix) const {
```
mkhazraee, line 56: If we wanna print, potentially we can add print for the new field too.

yanivz-nv: Done

src/infra/mem_section.h
yanivz-nv: Done

src/api/cpp/backend/backend_aux.h
yanivz-nv: Rebased on top of your desc refactor as discussed.

test/nixl/agent_example.cpp
```diff
@@ -172,6 +172,156 @@ void test_side_perf(nixlAgent* A1, nixlAgent* A2, nixlBackendH* backend, nixlBac
...
+            dst_mem_lists[update_idx].addDesc(dst_desc);
+
+            // Fill source buffers with test pattern
+            memset(src_bufs[update_idx][buf_idx], 0xbb, BUF_SIZE);
+        }
+    }
+
+    std::string remote_name;
+
+    // Main test loop - update metadata multiple times
```
mkhazraee: Let's test the connInfo flag too, where you get connInfo only with empty desc_list. Not sure in the same test, with a parameter, or a separate test. 

For now let's return the conninfo of all backends when the desclist is empty, not limiting to backends that support that memory type. Let's add this to the description of the method in nixl.h too, what an empty desclist means.

We also need a test for guardrail mode, right now you're getting the same result as getLocalMD, just checking the loads to work fine. But what if some part is missing, and there is a transfer req for that, it should fail and that can be asserted.

src/infra/nixl_memory_section.cpp
```diff
@@ -88,58 +156,73 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
-                return ret2;
         }
 
         *lp = mem_elms[i]; // Copy the basic desc part
         if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) ||
              (nixl_mem == FILE_SEG)) && (lp->len==0))
             lp->len = SIZE_MAX; // File has no range limit
 
-        target->addDesc(local_meta);
+
```
mkhazraee: nitpick, extra enter.

src/infra/nixl_memory_section.cpp
```diff
@@ -88,58 +156,73 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
-            for (int j=0; j<i; ++j) {
-                index = target->getIndex(mem_elms[j]);
-                backend->deregisterMem
-                    ((*(const nixl_meta_dlist_t*)target)[index].metadataP);
-                target->remDesc(index);
+        if (backend->supportsRemote()) {
+            ret = backend->getPublicData(local_sec.metadataP, self_sec.metaBlob);
+            if (ret != NIXL_SUCCESS) {
+                if (backend->supportsLocal())
+                    backend->unloadMD(local_sec.metadataP);
```
mkhazraee: Good call that we need unloadMD, but there are two problems:
1- I think it should be self_sec.metadataP, as that was the output of loadLocalMD.
2- Backend might return the same input pointer as output pointer for loadLocalMD, actually it does most of the case. We should add a pointer check, that if the pointers of remote and local were not the same, do the unloadMD. 

It might even get trickier if they output a pointer to the middle of the metadata that input pointer is pointing to. But that's on them, as they can distinguish and then skip the unload as it's part of the larger local metadata object.

src/infra/nixl_memory_section.cpp
```diff
@@ -88,58 +156,73 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
     }
-    return NIXL_SUCCESS;
+
+    // Abort in case of error
+    if (ret != NIXL_SUCCESS) {
+        for (int j = 0; j < i; ++j) {
+            int index = target->getIndex(mem_elms[j]);
+
+            if (backend->supportsLocal())
+                backend->unloadMD((*target)[index].metadataP);
```
mkhazraee: same point as above, and we need to get the remote_self ones.

src/infra/nixl_memory_section.cpp
```diff
@@ -148,18 +231,17 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
...
+    nixl_sec_dlist_t *target = it->second;
 
     for (auto & elm : mem_elms) {
         int index = target->getIndex(elm);
         // Errorful situation, not sure helpful to deregister the rest,
         // registering back what was deregistered is not meaningful.
         // Can be secured by going through all the list then deregister
         if (index<0)
-            return NIXL_ERR_UNKNOWN;
+            return NIXL_ERR_NOT_FOUND;
```
mkhazraee, line 248: I wonder if we should change the comment. Previously this case should've never happened, only in register and deregister of main agent code it was called. Now I see you removed the populate in deregister, might need more thinking.

src/infra/nixl_memory_section.cpp
```diff
@@ -148,18 +231,17 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
...
 
-        backend->deregisterMem
-            ((*(const nixl_meta_dlist_t*)target)[index].metadataP);
```
mkhazraee, line 257: Good catch, we changed some consts in the method parameters, this was left out to be changed.

src/infra/nixl_memory_section.cpp
```diff
@@ -172,39 +254,78 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
     return NIXL_SUCCESS;
 }
 
-nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
+nixl_status_t nixlLocalSection::serializeSections(nixlSerDes* serializer,
```
mkhazraee: I think this method can be local to this file, not in the nixlLocalSection, just used as a helper function for the other two methods.

src/infra/nixl_memory_section.cpp
mkhazraee: yeah, nice. For unique_ptr we can do it in this PR or a separate one, don't mind either way. But if just for a few members, as we should do a more proper one across all the code.

src/infra/nixl_memory_section.cpp
```diff
@@ -172,39 +254,78 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
...
+        auto it = sectionMap.find(sec_key);
+        if (it == sectionMap.end())
+            continue;
+
+        // TODO: consider section_map_t to be a map of unique_ptr or instance of nixl_meta_dlist_t.
+        //       This will avoid the need to delete the nixl_sec_dlist_t instances.
+        const nixl_sec_dlist_t *base = it->second;
+        nixl_sec_dlist_t *resp = new nixl_sec_dlist_t(nixl_mem, mem_elms.isSorted());
+        for (const auto &desc : trimmed) {
+            int index = base->getIndex(desc);
```
mkhazraee, line 318: Wonder if we just typecast the desc to nixlBasicDesc would be fine/more-efficient, than doing the trim() at the beginning.

src/infra/nixl_memory_section.cpp
```diff
@@ -148,18 +231,17 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
...
         // Can be secured by going through all the list then deregister
         if (index<0)
-            return NIXL_ERR_UNKNOWN;
+            return NIXL_ERR_NOT_FOUND;
 
-        backend->deregisterMem
-            ((*(const nixl_meta_dlist_t*)target)[index].metadataP);
+        backend->deregisterMem((*target)[index].metadataP);
         target->remDesc(index);
     }
```
mkhazraee, line 257: Regarding the next if statement, there was a conversation that maybe during backend creation, we create all the possible keys for that backend, so populating the sectionMap with empty desc_lists. For that we need to add a check when we look for a key, to also see the found desc_list is empty. If we go that path, then this if becomes extra. But don't think we need it in this PR, for future.

src/core/nixl_agent.cpp
```diff
@@ -311,15 +310,7 @@ nixlAgent::deregisterMem(const nixl_reg_dlist_t &descs,
...
-        nixl_meta_dlist_t resp(descs.getType(),
-                               descs.isSorted());
-
-        // Now just populating the metadata part, inside remDescList,
-        // getIndex is used to make sure exact match
-        ret = data->memorySection->populate(trimmed, backend, resp);
-        if (ret != NIXL_SUCCESS)
-            bad_ret = ret;
-        ret = data->memorySection->remDescList(resp, backend);
+        ret = data->memorySection->remDescList(descs, backend);
```
mkhazraee, line 327: There is a problem, we want to check that all the descriptors in the passed descs are available, and if so then delete them all. Right now if there are valid elements and some invalid ones, it'll deregister the initial parts and that puts the system in a bad situation. In the use case of registerMem we already know all of them are proper.

src/infra/nixl_memory_section.cpp
```diff
@@ -88,58 +156,73 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
+                backend->unloadMD((*target)[index].metadataP);
+            backend->deregisterMem((*target)[index].metadataP);
+            target->remDesc(index);
+        }
+        remote_self.clear();
+    }
+    return ret;
 }
 
 // Per each nixlBasicDesc, the full region that got registered should be deregistered
```
mkhazraee: Let's clarify the comment I've put, was confusing for myself too. The comment is referring to why I was using populate, to expand to the full region, but we don't want that. 

If we want to keep the method as is, we can replace it with a comment that this method assumes all the input descriptors can be found. That was the case in registerMem and with populate in deregisterMem. 

Another approach is to first check all elements to be available, then remove them. Since the removal would change the indexes, not sure what's the best way to do it, so maybe just an extra loop at the beginning is fine. If we want it to be more optimized, maybe we can have a flag that says don't check, so the first loop is skipped in case of register. But since that's only triggered in error case, not a big deal.

src/core/nixl_agent.cpp
yanivz-nv: Added initial checking loop inside `remDescList`.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done

src/infra/nixl_memory_section.cpp
yanivz-nv: Done with self_sec.metadataP and added ptr check.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done.

src/infra/nixl_memory_section.cpp
yanivz-nv: Added an extra loop, no skip flag since it will optimize only error flow.

src/infra/nixl_memory_section.cpp
yanivz-nv: Changed the comment(s)

src/infra/nixl_memory_section.cpp
yanivz-nv: ACK

src/infra/nixl_memory_section.cpp
yanivz-nv: IMO if this is fine for now I'll do it in a separate PR since this one is already big.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done. First implemented it before removing getStringDesc so I wasn't able to do it.

src/infra/nixl_memory_section.cpp
yanivz-nv: You are right and we already do it in other places. Done.

test/nixl/agent_example.cpp
yanivz-nv: Done.

src/infra/nixl_descriptors.cpp
```diff
@@ -369,7 +370,7 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
     // descriptor. std::string_view(typeid(T).name()) is compiler dependent
     if (std::is_same<nixlBasicDesc, T>::value)
         ret = serializer->addStr("nixlDList", "nixlBDList");
-    else if (std::is_same<nixlBlobDesc, T>::value)
+    else if (std::is_same<nixlBlobDesc, T>::value || std::is_same<nixlSectionDesc, T>::value)
```
mkhazraee, line 374: Hmm, I think we shouldn't serialize this, same as metadesc, and the if we have in deserialize.

src/infra/nixl_memory_section.cpp
```diff
@@ -183,58 +159,75 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
-        if ((ret1!=NIXL_SUCCESS) || (ret2!=NIXL_SUCCESS)) {
-            for (int j=0; j<i; ++j) {
-                index = target->getIndex(mem_elms[j]);
-                backend->deregisterMem
-                    ((*(const nixl_meta_dlist_t*)target)[index].metadataP);
-                target->remDesc(index);
+        if (backend->supportsRemote()) {
+            ret = backend->getPublicData(local_sec.metadataP, self_sec.metaBlob);
+            if (ret != NIXL_SUCCESS) {
+                if (backend->supportsLocal() && self_sec.metadataP != local_sec.metadataP)
```
mkhazraee, line 193: Add a 1 liner comment why we check the pointers.

src/infra/nixl_memory_section.cpp
```diff
@@ -183,58 +159,75 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
-    return NIXL_SUCCESS;
+
+    // Abort in case of error
+    if (ret != NIXL_SUCCESS) {
+        for (int j = 0; j < i; ++j) {
+            int index = target->getIndex(mem_elms[j]);
+
+            if (backend->supportsLocal()) {
+                int self_index = remote_self.getIndex(mem_elms[j]);
+                if (self_index >= 0 && remote_self[self_index].metadataP != (*target)[index].metadataP)
```
mkhazraee, line 221: self_index being negative shouldn't happen, maybe a comment on that.

src/infra/nixl_memory_section.cpp
```diff
@@ -243,18 +236,22 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_meta_dlist_t &mem_elms,
...
+        if (index < 0)
+            return NIXL_ERR_NOT_FOUND;
+    }
 
-        backend->deregisterMem
-            ((*(const nixl_meta_dlist_t*)target)[index].metadataP);
+    for (auto & elm : mem_elms) {
+        int index = target->getIndex(elm);
+        // Already checked, this should never happen
+        if (index < 0)
```
mkhazraee: I say remove the if as the comment is there. Maybe mention can be added for debug mode.

src/infra/nixl_memory_section.cpp
```diff
@@ -183,58 +159,75 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
+            }
+            backend->deregisterMem((*target)[index].metadataP);
+            target->remDesc(index);
+        }
+        remote_self.clear();
+    }
+    return ret;
 }
 
 // Per each nixlBasicDesc, the full region that got registered should be deregistered
```
mkhazraee: Let's remove the comment. Now things are full match, no need for any comments anymore..

src/infra/nixl_memory_section.cpp
```diff
@@ -322,26 +363,32 @@ nixl_status_t nixlRemoteSection::addDescList (
...
-    nixlMetaDesc out;
+    nixlSectionDesc out;
     nixlBasicDesc *p = &out;
     nixl_status_t ret;
     for (int i=0; i<mem_elms.descCount(); ++i) {
-        // TODO: remote might change the metadata, have to keep stringDesc to compare
-        //       if we support partial updates. Also Can add overlap checks (erroneous)
-        if (target->getIndex((const nixlBasicDesc) mem_elms[i]) < 0) {
+        // TODO: Can add overlap checks (erroneous)
+        int idx = target->getIndex(static_cast<nixlBasicDesc>(mem_elms[i]));
```
mkhazraee: I later made the method to use reference as I forgot initially, as you fixed above, don't think we need the typecast anymore.

test/nixl/agent_example.cpp
```diff
@@ -172,6 +172,201 @@ void test_side_perf(nixlAgent* A1, nixlAgent* A2, nixlBackendH* backend, nixlBac
...
+        assert(status == NIXL_SUCCESS);
+
+        status = A2->registerMem(dst_mem_lists[update], &extra_params2);
+        assert(status == NIXL_SUCCESS);
+    }
+
+    // Test metadata update with only backends and empty descriptor list
+    std::cout << "Metadata update - backends only\n";
+
+    A1->invalidateRemoteMD(agent2); // precaution, don't care about return value
```
mkhazraee: maybe you can actually assert that it's an error or so.

test/nixl/agent_example.cpp
mkhazraee, line 1: I'm debating to separate the test or not as it got really long. @tstamler what do you think? Also should we change agent_example.cpp and desc_example.cpp to agent_test.cpp and desc_test.cpp? I thought we did it some point in time.

tstamler: Let's leave the others alone for now as they might be imminently replaced by gtests. A lot of this test specifically won't be covered by gtests, but some of it might. I think let's evaluate after we have some Agent gtests, see what is and isn't covered and decide how to dissect it.

test/nixl/agent_example.cpp
```diff
@@ -172,6 +172,201 @@ void test_side_perf(nixlAgent* A1, nixlAgent* A2, nixlBackendH* backend, nixlBac
...
+        assert(status == NIXL_SUCCESS);
+        assert(remote_name == agent2);
+
+        // Make sure registered descriptors are updated
+        nixlDlistH *dst_side;
+        status = A1->prepXferDlist(agent2, dst_mem_lists[update].trim(), dst_side, &extra_params1);
+        assert(status == NIXL_SUCCESS);
+        assert(dst_side != nullptr);
+
+        // Make sure unregistered descriptors are not updated
```
mkhazraee: Is this "not loaded" instead of unregistered?

src/infra/nixl_descriptors.cpp
yanivz-nv: It's needed from `serializeSections`, I agree it might be a bit weird that it is serialized the same as BlobDesc and then the remote deserializes it as a BlobDesc.
Maybed I can add another `else if SectionDesc` and then build a temp BlobDesc to serialize. WDYT?
image

src/infra/nixl_memory_section.cpp
yanivz-nv: Done

src/infra/nixl_memory_section.cpp
yanivz-nv: Done

src/infra/nixl_memory_section.cpp
yanivz-nv: Agreed and done.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done.

test/nixl/agent_example.cpp
yanivz-nv: I didn't want to assert because we receive A1 and A2 from outside, so it might succeed.

test/nixl/agent_example.cpp
yanivz-nv: Right, it's a bad comment from before some updates I did to the test. Fixed the comments in this loop.

yanivz-nv:
/ok to test [3910791](https://github.com/ai-dynamo/nixl/pull/159/commits/39107918007b77a8dd0240b86019d8a1dccd20e4)

mkhazraee:
/ok to test [3910791](https://github.com/ai-dynamo/nixl/pull/159/commits/39107918007b77a8dd0240b86019d8a1dccd20e4)

src/infra/nixl_descriptors.cpp
mkhazraee: Oh now I see, then add a comment we want to serialize desclist of secdesc same as blobdesc, just not spend cycles to convert, and in deserialize it would show up as blobdesc.

src/infra/nixl_memory_section.cpp
mkhazraee: Minor update, making it more clear (maybe becomes 2 lines): 
"A backend might use the same object for both initiator/target side of a transfer, so no need for unloadMD in that case."

test/nixl/agent_example.cpp
mkhazraee: Maybe update the comment then, in the previous line, and we don't care either way. The comment is confusing a tad.

test/nixl/agent_example.cpp
```diff
@@ -172,6 +172,201 @@ void test_side_perf(nixlAgent* A1, nixlAgent* A2, nixlBackendH* backend, nixlBac
...
+        assert(status == NIXL_SUCCESS);
+        assert(remote_name == agent2);
+
+        // Make sure loaded descriptors are updated
+        nixlDlistH *dst_side;
+        status = A1->prepXferDlist(agent2, dst_mem_lists[update].trim(), dst_side, &extra_params1);
+        assert(status == NIXL_SUCCESS);
+        assert(dst_side != nullptr);
+
+        // Make sure non-loaded descriptors are not updated
```
mkhazraee: not-loaded

src/core/nixl_agent.cpp
```diff
@@ -995,6 +983,81 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
```
mkhazraee, line 987: There is code duplication between this and getLocalMD, can we do a namespace method to refactor?

src/core/nixl_agent.cpp
```diff
@@ -995,6 +983,81 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
...
+                backend_list->push_back(elm.second);
+        }
+    } else {
+        backend_list = &tmp_list;
+        for (const auto & elm : extra_params->backends)
+            backend_list->push_back(elm->engine);
+    }
+
+    // First find all relevant engines and their conn info.
+    // Best effort, ignore if no conn info.
```
mkhazraee: No conn info means it doesn't support Remote, right? Let's mention that in the comment. Also maybe you can check for that here, or for now just mention it in the comment.

src/infra/nixl_descriptors.cpp
yanivz-nv: Added a comment here and also in nixlSectionDesc::serialize.

src/infra/nixl_memory_section.cpp
yanivz-nv: Done

test/nixl/agent_example.cpp
yanivz-nv: Done.

test/nixl/agent_example.cpp
yanivz-nv: Done

src/core/nixl_agent.cpp
yanivz-nv: Updated comment

PR #161, opened 2025-04-17
UCX: Fix memory leak in nixlUcxEngine

yanivz-nv:
Caught it by chance with address sanitizer (meson flag: -Db_sanitize=address,undefined)

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -348,6 +348,7 @@ nixlUcxEngine::~nixlUcxEngine () {
     vramFiniCtx();
     delete uw;
     delete uc;
+    free(workerAddr);
 }
 
 /****************************************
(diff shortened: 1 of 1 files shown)


tstamler:
Need @artpol84 to double check, I think there might have been specific reasons we didn't want to free these at these points, but I don't remember. I will launch CI to make sure to run tests as well

tstamler:
/ok to test https://github.com/ai-dynamo/nixl/commit/dec94a5323f253c8c3ceac8e859763278aec42cd

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -587,6 +588,7 @@ nixl_status_t nixlUcxEngine::registerMem (const nixlBlobDesc &mem,
 
     out = (nixlBackendMD*) priv; //typecast?
 
+    free((void*)rkey_addr);
```
brminich, line 591: btw, is there a reason to copy packed rkey in `nixlUcxWorker::packRkey` instead of using the one returned by UCX?

tstamler: I suppose it would be ok, we would need to add a separate util function to later call ucp_rkey_buffer_release. It would happen much later so not sure the implications of waiting a long time between receiving the rkey and calling release.

PR #137, opened 2025-04-09
Abseil-CPP Integration

etoledano:
This is a clone of this PR, from a branch: https://github.com/ai-dynamo/nixl/pull/130

PR #163, opened 2025-04-17
Debug infra: Add NIXL logging and assertion macros 

ovidiusm:
Debug infra: Add NIXL logging macros

# Goal

The goal of the debug infrastructure is to plant extensive checks that we can remove in release builds since they may hurt performance; may help NIXL users in debug builds to develop their applications by spotting misuse of the API, unexpected or should-not-have-happened cases.

In general, follow the mature logging support design from UCX.

# Implementation

Leverage Abseil support, which provides logging functionality.

Define:

* Logging macros with different log levels (NIXL_ERROR, NIXL_LOG etc.) that can be selectively compiled out in release builds (by setting ABSL_MIN_LOG_LEVEL during build) and can also be configured at runtime
* Checks with assertions, that are either removed from release builds (NIXL_ASSERT) or more important checks that are compiled in all builds (NIXL_ASSERT_ALWAYS), similar to ucs_assert/ucs_assert_always
* Set default log level at compile time (strips logs/asserts)
* Configurable log level at runtime based on env variable NIXL_LOG_LEVEL

# Out of scope

NIXL code to be updated separately to replace existing stdout/err prints with the logs, and the asserts with the new checks.

# Deps

https://github.com/ai-dynamo/nixl/pull/137 needs to be merged first


src/utils/common/nixl_log.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <string>
+
+namespace {
+
+// Helper class to ensure logging is initialized automatically.
+class LoggingInitializer {
+public:
+    LoggingInitializer() {
+        // Initialize Abseil logging system.
+        absl::InitializeLog();
```
tstamler, line 32: I'm not familiar with absl logging, just wanted to verify that its not any overhead. Typically how I would implement this would be a macro where the log operations are basically no-ops unless the log level is right, but maybe absl does that.

src/utils/common/nixl_log.h
```diff
@@ -0,0 +1,84 @@
...
+ *-----------------------------------------------------------------------------*
+ * Ordered by severity (highest to lowest)
+ * Usage: NIXL_INFO << "Message part 1 " << variable << " message part 2";
+ */
+
+/*
+ * Logs a message and terminates the program unconditionally.
+ * Maps to Abseil LOG(FATAL). Use for unrecoverable errors.
+ */
+#define NIXL_FATAL LOG(FATAL)
```
etoledano, line 35: Why do we need to wrap `LOG` macros? doesn't seem to shorter, moreover it is a bit confusing since `LOG` is dropped and `NIXL_<Severity>`  can imply for actions that aren't logging.

src/utils/common/meson.build
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+nixl_common_inc = include_directories('.')
+
+nixl_common_deps = [
+    absl_dep, # Dependency from root meson.build
```
etoledano, line 19: Can be changed to just `absl_log_dep`

src/utils/common/nixl_log.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        // Set log level based on environment variable NIXL_LOG_LEVEL
+        // Levels: TRACE, DEBUG, INFO, WARNING, ERROR, FATAL
+        const char* env_log_level = std::getenv("NIXL_LOG_LEVEL");
+        std::string level_str = "INFO"; // Default level
+        if (env_log_level != nullptr) {
+            level_str = absl::AsciiStrToUpper(env_log_level);
+        }
```
etoledano, line 40: Seems this class initializes just the once, in this case maybe consider do it in compile time:

```
# meson_options.txt

option('log_level', type: 'combo', choices : , ['trace', 'debug', 'info', 'warning', 'error', 'fatal', '*'] value: 'info', description: 'Log Level')
``` 

```
# meson.build

add_project_arguments('-DLOG_LEVEL_' + get_option('log_level').to_upper() , language: 'cpp')
```

```
// src/utils/common/nixl_log.cpp

// Helper class to ensure logging is initialized automatically.                 
class LoggingInitializer {                                                      
public:                                                                         
    LoggingInitializer() {                                            
        // Initialize Abseil logging system.                                    
        absl::InitializeLog();                                                  
                                                                                
        // Levels: TRACE, DEBUG, INFO, WARNING, ERROR, FATAL
        if constexpr (std::strcmp(LOG_LEVEL, "TRACE") == 0) {
            absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);
            absl::SetVLogLevel("*", 2); 
        } else if constexpr (std::strcmp(LOG_LEVEL, "INFO") == 0) {
			...
        }
    }                                                                           
};    
```

Can even do it w/o a class using `__attribute__((constructor))`:

```
namespace                                                                       
{                                                                               
                                                                                
void LoggingInitializer() __attribute__((constructor));                         
void LoggingInitializer()                                                       
{                                                                               
    // Initialize Abseil logging system.                                        
    absl::InitializeLog();                                                      
                                                                                
    // Set log level based on environment variable NIXL_LOG_LEVEL               
    // Levels: TRACE, DEBUG, INFO, WARNING, ERROR, FATAL                        
    if constexpr (std::strcmp(LOG_LEVEL, "TRACE") == 0) {                       
        absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);                  
        absl::SetVLogLevel("*", 2);                                             
    } else if constexpr (std::strcmp(LOG_LEVEL, "INFO") == 0) {                 
    }                                                                           
}                                                                               
}  // anonymous namespace           
```

src/utils/common/nixl_log.cpp
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+             // Unknown level, default to WARNING
+             absl::SetMinLogLevel(absl::LogSeverityAtLeast::kWarning);
+             absl::SetVLogLevel("*", 0);
+        }
+    }
+};
+
+// Executes the initializer when the program starts before main() is called.
+static LoggingInitializer global_initializer_instance;
```
etoledano, line 69: Consider using `__attribute__((constructor))`

src/utils/common/nixl_log.h
ovidiusm: You're right that the macros aren't necessarily shorter than the direct Abseil calls.

The main reasons for introducing these `NIXL_` wrappers are:

1.  Conditional Compilation/Stripping: This is a key benefit. The wrappers allow us to control which log levels are compiled into different build types more easily. As noted in the comments, levels like `NIXL_WARN`, `NIXL_INFO`, `NIXL_DEBUG`, and `NIXL_TRACE` are intended to be stripped from release builds. While Abseil provides mechanisms for this (like `DLOG`, `DVLOG`, and build flags), these wrappers offer a centralized place to manage the project's specific policy for log stripping, particularly for levels like `INFO` and `WARNING` which Abseil's `LOG` macro doesn't strip by default in release builds.

2.  Parity with UCX: This approach aligns our logging levels and default stripping behavior (stripping INFO/WARN and below in release builds) with the logging levels from UCX (`ucs_log_level_t`). Also, UCX has more granular logging levels than Abseil, which these macros make possible.

3.  Abstraction: A layer over the Abseil logging library is useful if we need to add any customizations in the future, otherwise we have to updating every logging call site throughout the codebase.

4.  Consistency: Using the `NIXL_` prefix helps enforce a consistent logging API style across the project.

While `NIXL_FATAL` itself might seem like a simple alias for `LOG(FATAL)`, maintaining the wrapper keeps it consistent with the other levels where the abstraction, stripping, and parity benefits are more significant. Hopefully, the `NIXL_` prefix combined with the severity level (FATAL, ERROR, etc.) makes the logging intent clear enough.

src/utils/common/nixl_log.cpp
ovidiusm: If the log level option has a default value, then debug and release builds will use the same log level unless it is overridden explicitly. Is this your intent?

If not, I suggest adjusting this such that if the log_level option is set to some default, Meson will set the LOG_LEVEL_... definition to LOG_LEVEL_DEBUG for debug builds and LOG_LEVEL_ERROR for release, debugoptimized, or plain builds. If the user explicitly selects a log level like info or trace, that specific level will be used regardless of the build type.

src/utils/common/nixl_log.cpp
ovidiusm: Abseil does indeed provide that: https://github.com/abseil/abseil.github.io/blob/master/docs/cpp/guides/logging.md

src/utils/common/nixl_log.cpp
ovidiusm: Done

ovidiusm:
Reopened as https://github.com/ai-dynamo/nixl/pull/180 on the main branch

src/utils/common/nixl_log.cpp
etoledano: My general intention is to [prefer compile time checks](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#p5-prefer-compile-time-checking-to-run-time-checking), and conduct the log level check at compile time.

src/utils/common/nixl_log.h
etoledano: I think that Abseil implements `(void)0` when the log level doesn't apply, but I can accept with the consistency to mark with `NIXL_` prefix. What I am still not sure is readability, since `LOG(*)` does imply logging which can be interpeted to `operator<<`, `NIXL_WARN` not so much.

Does it makes sense to use the following syntex?
```

// nixl_log.h
namespace nixl {
namespace logging {
template <typename... Args>
void Warn(const char* format, Args... args) {
    LOG(WARN) << absl::StrFormat(format, args...);
}
}; // namespace logging
}; // namespace nixl

// file.cpp

namspace nixl {

int f() {
    logging::Warn("Received argument %d but expected %d", x, y);
}
}; // namespace nixl

```

etoledano: Just to make my previous note more clear, I assume we plan on having the nixl context as a namespace (@yosefe) rather than a prefix, which is more common in C++, and in this case we will be  forced to make this logger an object with `<<` operator, which probably can be more simple to do using the suggested approach (with templates), rather than something like this:

```

namespace nixl {

class LogWarn {
    LogWarn& operator<<(const std::string& message) {
        LOG(WARN)<< message;
        return *this;
    }

};

static const LogWarn log_warn;

}; // namespace nixl

```

src/utils/common/nixl_log.cpp
ovidiusm: In that case I will remove the config via env var, and the log level will be fixed during compilation.

src/utils/common/nixl_log.h
ovidiusm: Thanks, this is helpful. Would you prefer to use exclusively printf-style logging utility functions (like `void Warn(const char* format, Args... args)`) instead of streams? Or both?

etoledano: I don't have a particular preference, I think making it a stream like would be a bit more clumsy to implement, maybe I'm wrong.
My main intention is to have the code fit the guidelines, in this case with scoping through a namespace by a logical structure that makes sense, and avoiding doing that with a `define` since since [MACROS are the source of all evil](https://arne-mertz.de/2019/03/macro-evil/)

By the way, I wasn't sure how/if abseil handles it, but in case it doesn't, then to have the log not ruined by several threads using IO all together (`std::cout`, file, ...), we should probably have some logic in those `Warn` function, which the `MACRO` couldn't have provided.

ovidiusm: We will not be able to avoid macros here since we lose line number information. And C++ style streams are easier to use than printf-like syntax for printing strings and other objects...

PR #165, opened 2025-04-17
python API: small change to provide clarity to check_remote_xfer

tstamler:


src/api/python/_api.py
```diff
@@ -516,13 +516,18 @@ def update_notifs(self, backends: list[str] = []) -> dict[str, list[bytes]]:
...
     @return True if the notification is found, False otherwise.
     """
 
     def check_remote_xfer_done(
-        self, remote_agent_name: str, lookup_msg: bytes, backends: list[str] = []
+        self,
+        remote_agent_name: str,
+        lookup_tag: bytes,
+        backends: list[str] = [],
+        full_search=False,
```
mkhazraee: full_search is a tad confusing. prefixed-tag or something like that might be more clear. Also let's add it to the params of the mehtod.

src/api/python/_api.py
```diff
@@ -533,8 +538,10 @@ def check_remote_xfer_done(
 
         if remote_agent_name in self.notifs:
             for msg in self.notifs[remote_agent_name]:
-                if lookup_msg in msg:
-                    message = lookup_msg
+                if (full_search and lookup_tag in msg) or (
```
mkhazraee: Not sure how python handles this, maybe repeating the code but having the if statement outside is better, not sure. If stayed this way, let's put the paranthesis more consistent, it's a tad confusing now (not full_search and msg.startswith(lookup_tag)) in on line.

tstamler: I don't really want to just repeat code, python should be handling this correctly in my basic testing. And I understand what you want with the parenthesis, but black won't let me...

src/api/python/_api.py
tstamler: I went with tag_is_prefix, I think that sounds a little better, otherwise updated the comment

mkhazraee (approved):

PR #167, opened 2025-04-18
API: add 3 new NB APIs for metadata transfer with socket implementation

tstamler:
-Three new NB API calls for sending/receiving metadata directly from other agents
-Listener thread implementation to handle connections and metadata transfer
-Socket P2P metadata transfer implementation
-update nixl_text.cpp to test these functions


src/api/cpp/nixl.h
```diff
@@ -313,6 +313,38 @@ class nixlAgent {
...
+        /*** Metadata handling through direct channels (p2p socket and ETCD) ***/
+        /**
+         * @brief  Send your own agent metadata to a remote location.
+         *
+         * @param  remote_ip     If specified, this is the remote IP address to send your metadata to
+         *                       If unspecified, this will send your data to the metadata server.
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        sendLocalMD (const std::string remote_ip = "") const;
```
mkhazraee: Some food for thoughts, most prob none of them are applicable:
- Maybe we should make it remote ip and port, and define it in a struct. But if it makes sense to use a fixed port between the agents, then nevermind.
- Maybe we can put the agent name here, so it's more consistent with other APIs. But then the question is where to get the mapping between the two. If it's only in agent creation, that loses the dynamicity aspect.
- Piggybacking on the previous one, and more like fetchRemoteMD, maybe we can put both agent name and remote_ip. if agentName is self, send to MD. But this might be making things confusing for no reason.

src/api/cpp/nixl.h
```diff
@@ -313,6 +313,38 @@ class nixlAgent {
...
+        /**
+         * @brief  Fetch other agent's metadata and unpack it internally.
+         *
+         * @param  remote_name      Name of remote agent to fetch from ETCD or socket.
+         * @param  remote_ip        If unspecified, will fetch from ETCD, otherwise will try socket
+         * @return nixl_status_t    Error code if call was not successful
+         */
+        nixl_status_t
+        fetchRemoteMD (const std::string remote_name,
+                       const std::string remote_ip = "");
```
mkhazraee: I suggest to typedef std::string for remote_ip to be more clear. We can use some libs, if for socket we anyways have to, or just do our own typedef.

src/core/agent_data.h
```diff
@@ -47,7 +56,18 @@ class nixlAgentData {
         std::unordered_map<std::string, nixlRemoteSection*,
                            std::hash<std::string>, strEqual>     remoteSections;
 
-        nixlAgentData(const std::string &name, const nixlAgentConfig &cfg);
+        nixlMDStreamListener                                 listener;
+        std::unordered_map<std::string, nixlMDStreamClient>  remoteSockets;
+        std::thread                                          commThread;
```
mkhazraee: I think in UCX we have some of this, not sure if we want to put them in an internal class for refactoring, if they become very similar to when Artem brings the progress thread in.

src/core/nixl_agent.cpp
```diff
@@ -57,12 +57,119 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
...
+
+            if(new_fd != -1){
+                //new to convert fd to IP address and add to client map
+                //but don't setupClient()
+            }
+        }
+
+
+        //second, do agent commands
+		this->getCommWork(work_queue);
```
mkhazraee: tabs

mkhazraee: Just saying in case Vim is playing tricks again, better sooner to catch that than later.

src/core/nixl_agent.cpp
```diff
@@ -57,12 +57,119 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
...
+				case SOCK_SEND:
+                    //not connected
+                    if(client == this->remoteSockets.end()) {
+                        nixlMdStreamClient new_client(request.second, DEFAULT_COMM_PORT);
+                        this->remoteSockets[request.second] = new_client;
+                        new_client.setupClient();
+                        client = new_client;
+                    }
+                    nixl_blob_t my_MD;
+                    assert(myAgent->getLocalMD(my_MD) == NIXL_SUCCESS);
```
mkhazraee: I assume you're gonna remove the assert, just during dev? we would error not assert anything, not to break the upper layers.

src/core/nixl_agent.cpp
```diff
@@ -57,12 +57,119 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
     }
 }
 
+void nixlAgentData::commWorker(){
```
mkhazraee: I'm not a fan of having methods in nixlAgentData, it's just where to hold the data. And then you need the pointer back to the agent which is not the best.

On the other hand I see adding a private method would sound like why we had the nixlAgentData in the first place. Maybe a middle ground would be only 1-2 new private methods in the agent, without any parameters, listener_callback and progress_callback, so it's obvious it's what for, and still 1 liner not confusing, and we don't need to provide that much documentation, as it's not part of the API. Maybe you can talk with Artem how he planned to implement his progress thread for the agent.

mkhazraee: This is assuming with at most 1 input you can distinguish between the 3 methods, or even maybe have some flag in the agentData, so when the function is called, you look at that. Didn't go through the details of implementation to know what I'm saying makes any sense.

src/api/cpp/nixl.h
tstamler: Yes, my biggest challenge is how to reconcile agent name and IP address. Also, with the port number, I think the big question is going to be: do we want to allow multiple agents per node, in different processes, or do we want to disallow that? If we want to allow that, I think we need to allow port to be specified while listening (probably in agent constructor), as well as maintaining explicit mappings between agent name and IP address.

src/core/nixl_agent.cpp
tstamler: noooooooo (you're right)

src/core/agent_data.h
tstamler: yes, we wanted to have a NIXL thread pool instead of separate UCX thread/listener thread but I kind of don't think it'll happen for 0.2.0. When its time for the NIXL thread pool, we can relocate.

src/core/nixl_agent.cpp
tstamler: Definitely agree I didn't feel good about putting them in nixlAgentData, but at the time I definitely felt worse about putting them in a public API header. If as you say its ok to have them defined as private methods without documentation/much implementation its probably all good then. I can also probably get rid of the back pointer to the Agent in the nixlAgentData then as well.

src/api/cpp/nixl.h
mkhazraee: Very good call. Def, right now dynamo is using 1 agent per gpu.

src/core/nixl_agent.cpp
mkhazraee: I think that would be a better one than this, but how did Artem implement the progress thread? My assumption was there is a function, always running, let's say with some sleep times while we don't have callbacks. And each time it can call the relevant progress calls. In other words, a thread that is spawned in agent constructor, and killed in destructor.

tstamler: This is basically how its already being updated, I've moved the actual progress function to the agent.

src/core/nixl_agent.cpp
tstamler: gone

src/api/cpp/nixl.h
tstamler: I've added support for multiple ports

src/api/cpp/nixl.h
```diff
@@ -313,6 +314,44 @@ class nixlAgent {
...
+        /**
+         * @brief  Send your own agent metadata to a remote location.
+         *
+         * @param  remote_ip     If specified, this is the remote IP address to send your metadata to
+         *                       If unspecified, this will send your data to the metadata server.
+         * @param  port          Specific port to connect to, otherwise use DEFAULT_COMM_PORT
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        sendLocalMD (const std::string remote_ip = "", const int port = 0) const;
```
mkhazraee, line 329: Debating if we should add a new nixl_type for this. in some sense if the optional are passed, both should be passed, specially you didn't put to a default value. And I'm not sure if we end up adding any extra_params to it, doubt. Just optional params being a pain in C++. But not a big deal, ip and port are clear what they mean instead.

src/core/nixl_agent.cpp
```diff
@@ -57,9 +60,234 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
     }
 }
 
+int connectToIP(std::string ip_addr, int port) {
```
mkhazraee: Not sure, maybe better to return status instead of throwing errors? To make things not crash.

src/core/nixl_agent.cpp
```diff
@@ -57,9 +60,234 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
...
+
+            switch(req_command) {
+                case SOCK_SEND:
+                {
+                    //not connected
+                    if(client == data->remoteSockets.end()) {
+                        int new_client = connectToIP(req_ip, req_port);
+                        data->remoteSockets[req_ip] = new_client;
+                        client_fd = new_client;
+                    } else
```
mkhazraee: add { } after else, looks a tad weird, and might cause confusion

src/core/nixl_agent.cpp
```diff
@@ -57,9 +60,234 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
...
+                    data->remoteSockets.erase(ip_address);
+                } else {
+                    throw std::runtime_error("Received socket message with bad header" + header + "\n");
+                }
+            }
+        }
+
+        //relatively large delay for now
+        nixlTime::us_t start = nixlTime::getUs();
+        while( (start + 100000) > nixlTime::getUs()) {
```
mkhazraee: Let's make it a parameter like for ucx prog thread.

src/core/nixl_agent.cpp
mkhazraee: Also this API and all the way to and including listener, I'm wondering if we should put in a separate file. Infra style, something we use inside the agent.

src/api/cpp/nixl.h
tstamler: I separated them because both are technically optional:

IP can be unspecified to send to ETCD
Port can be unspecified to use default port

src/core/nixl_agent.cpp
tstamler: yes, these internal functions should probably just return errors instead of throwing exceptions. separate file might make sense

src/core/nixl_agent.cpp
tstamler: done

src/core/nixl_agent.cpp
tstamler: done

src/core/nixl_agent.cpp
tstamler: done

src/api/cpp/nixl.h
mkhazraee: hmm, there is a tad of dependence, if no IP is given, port won't be checked. But I see your point. 

One thing was we said let's not have several optional in a function header, but these are clear what they mean I guess.

src/api/cpp/nixl_params.h
```diff
@@ -42,16 +46,28 @@ class nixlAgentConfig {
...
+         * @brief  Agent configuration constructor for enabling various features.
          * @param use_prog_thread  flag to determine use of progress thread
+         * @param use_prog_thread  flag to determine use of listener thread
+         * @param port             specify port for listener thread to listen on
          * @param pthr_delay_us    Optional delay for pthread in us
          */
-        nixlAgentConfig(const bool use_prog_thread, const uint64_t pthr_delay_us=0) {
-            this->useProgThread = use_prog_thread;
-            this->pthrDelay     = pthr_delay_us;
+        nixlAgentConfig(const bool use_prog_thread, const bool use_listen_thread=false, const int port=0, const uint64_t pthr_delay_us=0, const uint64_t lthr_delay_us = 100000) {
```
mkhazraee: Should we do before constructor style as their number is increasing and all are specific member init?

src/core/agent_data.h
```diff
@@ -19,9 +19,23 @@
 
 #include "common/str_tools.h"
 #include "mem_section.h"
+#include "stream/metadata_stream.h"
+
+#define DEFAULT_COMM_PORT 8888
```
mkhazraee: Let's follow the constexpr guideline.

src/core/agent_data.h
```diff
@@ -47,7 +61,18 @@ class nixlAgentData {
...
+        std::unordered_map<std::string, int>  remoteSockets;
+        std::thread                           commThread;
+        std::vector<nixl_comm_req_t>          commQueue;
+        std::mutex                            commLock;
+        bool                                  commThreadStop;
+
+        void enqueueCommWork(nixl_comm_req_t request);
+        void getCommWork(std::vector<nixl_comm_req_t> &req_list);
+
+        nixlAgentData(nixlAgent* my_agent, const std::string &name, const nixlAgentConfig &cfg);
```
mkhazraee: We don't need my_agent anymore, right? I don't see you using it in the nixl_agent.cpp either

src/core/meson.build
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-nixl_lib_deps = [nixl_infra, serdes_interface, dl_dep]
```
mkhazraee, line 16: Just wondering what was the dl_dep that we had, don't recall.

src/core/nixl_agent.cpp
```diff
@@ -91,10 +92,24 @@ nixlAgent::nixlAgent(const std::string &name,
...
+        data->listener = new nixlMDStreamListener(my_port);
+        data->listener->setupListener();
+        data->commThreadStop = false;
+        data->commThread = std::thread(&nixlAgent::commWorker, this);
+    }
 }
 
 nixlAgent::~nixlAgent() {
+    if(data->config.useListenThread) {
+        data->commThreadStop = true;
```
mkhazraee, line 108: Do we really need this? Seems like it's always the same value as config.useListenThread

src/utils/stream/metadata_stream.h
```diff
@@ -52,15 +52,15 @@ class nixlMDStreamListener: public nixlMetadataStream {
...
-        void            setupListener();
-        void            acceptClient();
         void            acceptClientsAsync();
         void            recvFromClients(int clientSocket);
 
     public:
         nixlMDStreamListener(int port);
         ~nixlMDStreamListener();
 
+        int         acceptClient();
```
mkhazraee, line 62: Do you know why Vish put that in private? Like there is any reasoning not to expose it? It could be initially we had most things here and they were moved out, so no reason.

src/core/nixl_listener.cpp
```diff
@@ -0,0 +1,253 @@
...
+                            std::cerr << "Listener thread could not connect to IP " << req_ip;
+                            break;
+                        }
+                        data->remoteSockets[req_ip] = new_client;
+                        client_fd = new_client;
+                    } else {
+                        client_fd = client->second;
+                    }
+
+                    sendCommMessage(client_fd, std::string("NIXLCOMM:SENT" + my_MD));
```
mkhazraee: SENT is a little confusing. Maybe something like LOAD or so is better, not sure. 
In the next phase we should make it more consistent with ETCD messages, but for now no need to be compatible across versions of NIXL, so all good to change later.

src/core/nixl_listener.cpp
```diff
@@ -0,0 +1,253 @@
...
+                        if(new_client == -1) {
+                            std::cerr << "Listener thread could not connect to IP " << req_ip;
+                            break;
+                        }
+                        data->remoteSockets[req_ip] = new_client;
+                        client_fd = new_client;
+                    } else
+                        client_fd = client->second;
+
+                    sendCommMessage(client_fd, std::string("NIXLCOMM:SEND"));
```
mkhazraee, line 171: SEND might not be the best either, maybe FETCH, or MDREQ or something like that. Not a big deal though, if we do LOAD up there, SEND is also a command.

src/core/nixl_listener.cpp
```diff
@@ -0,0 +1,253 @@
...
+        for (const auto& [ip_address, socketClient] : data->remoteSockets ) {
+            std::string commands;
+            std::vector<std::string> command_list;
+            nixl_status_t ret;
+
+            ssize_t recv_bytes = recvCommMessage(socketClient, commands);
+
+            if(recv_bytes == 0 || recv_bytes == -1) continue;
+
+            command_list = str_split_substr(commands, "NIXLCOMM:");
```
mkhazraee, line 204: Idea is same person might have sent a couple commands, but they're pushed together due to how we send in socket? I thought they would end up in separate messages and we have to iterate over them.

src/core/nixl_listener.cpp
```diff
@@ -0,0 +1,253 @@
...
+                } else {
+                    throw std::runtime_error("Received socket message with bad header" + header + ", critically failing\n");
+                }
+            }
+        }
+
+        //relatively large delay for now
+        nixlTime::us_t start = nixlTime::getUs();
+        while( (start + data->config.lthrDelay) > nixlTime::getUs()) {
+            std::this_thread::yield();
```
mkhazraee, line 238: I don't recall if it was not set for ucx progress, we had a different path, meaning no yield, or not,

src/api/cpp/nixl.h
```diff
@@ -33,6 +33,7 @@ class nixlAgent {
     private:
         /** @var  data  The members in agent class wrapped into single nixlAgentData member. */
         nixlAgentData* data;
+        void commWorker();
```
mkhazraee: let's make it nixl_status_t, just not to be void and consistent with other methods. Anyways we don't care about it's output.

src/core/meson.build
tstamler: this was a bad rebase on my part, they should both be there. dl_dep is for old versions of linux so I didn't catch it

src/core/agent_data.h
tstamler: yes, my bad

src/core/nixl_agent.cpp
tstamler: this is the destructor, it allows the user thread to signal to the listener thread to stop execution

src/core/nixl_listener.cpp
tstamler: It'll just be sent as part of a big socket buffer. Its unlikely there will be more than one or two messages at a time from a single remote client, but this feels safest

src/core/nixl_listener.cpp
tstamler: this is copied right from UCX progress thread

src/utils/stream/metadata_stream.h
tstamler: it was only being called implicitly by the startListenerForClient, not sure why

src/api/cpp/nixl.h
tstamler: moving to standalone function anyway

mkhazraee (approved):

PR #168, opened 2025-04-18
DOCA plugin: enable GPU device backend postXfer

e-ago:


src/api/cpp/backend/backend_engine.h
```diff
@@ -85,6 +85,9 @@ class nixlBackendEngine {
         // Determines if a backend supports progress thread.
         virtual bool supportsProgTh () const = 0;
 
+        // Determines if a backend supports posting tranfers requests from GPU.
+        virtual bool supportsTreqGpu () const = 0;
```
mkhazraee: Maybe we can set it as an initparam that it's supported or not, but I think gpu init is considerable feature to have such a function for it. 

If we agree to go down this path, we have to update our other backends too. A little bit not backward compatible, still at a stage we might be able to do. Or we can make it not pure virtual, so by default backends assume it's not supported. We might do this for progress/notification too, set by default to no, keeping backward compatibility.

I would do a minor rename, supportsGpuInitiated be more clear.

src/api/cpp/backend/backend_engine.h
```diff
@@ -114,6 +117,16 @@ class nixlBackendEngine {
                                         const nixl_opt_b_args_t* opt_args=nullptr
                                        ) = 0;
 
+        // Preparing a request to be posted from the GPU
```
mkhazraee: Since it's exact same API with one additional argument, that means it's an optional parameter, should go to opt_args.

src/api/cpp/backend/backend_plugin_doca_common.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef __BACKEND_PLUGIN_DOCA_COMMON_H
+#define __BACKEND_PLUGIN_DOCA_COMMON_H
+
+#include <nixl_types.h>
+#include <doca_rdma.h>
+
+#define DOCA_XFER_REQ_SIZE 512
+
+struct docaXferReqGpu {
```
mkhazraee: This should go to utils, same as ucx. It's specific to backends that do gpu initiated. It's not nixl backend level.

meson.build
```diff
@@ -117,6 +117,8 @@ if get_option('install_headers')
   install_headers('src/utils/common/nixl_time.h', install_dir: prefix_inc + '/utils/common')
   install_headers('src/api/cpp/backend/backend_engine.h', install_dir: prefix_inc + '/backend')
   install_headers('src/api/cpp/backend/backend_aux.h', install_dir: prefix_inc + '/backend')
+  install_headers('src/api/cpp/backend/backend_plugin_doca_common.h', install_dir: prefix_inc + '/backend')
```
mkhazraee: Following my later comment, it's not nixl level, it's plugin level, so should go to utils not here.

src/api/cpp/backend/backend_plugin_doca_device.cuh
```diff
@@ -0,0 +1,66 @@
...
+#ifndef __BACKEND_PLUGIN_DOCA_DEVICE_H
+#define __BACKEND_PLUGIN_DOCA_DEVICE_H
+
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <doca_gpunetio_dev_buf.cuh>
+#include <doca_gpunetio_dev_rdma.cuh>
+#include "backend_plugin_doca_common.h"
+
+__device__ inline void prepXferReqGpu_DocaBlock(uintptr_t backendHandleGpu)
```
mkhazraee: Part of utils, and specific to that backend, not in api/cpp/backend/*

src/api/cpp/nixl.h
```diff
@@ -221,6 +221,41 @@ class nixlAgent {
...
+         * @param  operation      Operation for transfer (e.g., NIXL_WRITE)
+         * @param  local_descs    Local descriptor list
+         * @param  remote_descs   Remote (or loopback) descriptor list
+         * @param  remote_agent   Remote (or self) agent name for accessing the remote (local) data
+         * @param  req_hndl [out] Transfer request GPU memory handle output
+         * @param  extra_params   Optional extra parameters used in creating a transfer request
+         * @return nixl_status_t  Error code if call was not successful
+         */
+         nixl_status_t
+         createXferReq (const nixl_xfer_op_t &operation,
```
mkhazraee: Same here, one parameter is being added, that means optional and should go to extra_params. We have nixlBackendReqH that any backend can inherit, so gpu-initiated backend should just inherit that, and from user point of view they won't know what's happening underneath.

mkhazraee: So not a new API, just a new param in extra_params.

src/api/cpp/nixl_types.h
```diff
@@ -29,6 +29,10 @@ class nixlBackendH;
 class nixlXferReqH;
 class nixlAgentData;
 
+typedef struct nixlXferReqHGpu {
```
mkhazraee: As mentioned earlier, this is not a nixl type, just the gpu initiated backend should inherit from nixlBackendReqH  for its usage.

src/api/cpp/nixl.h
mkhazraee: We can add a second new param in extra_param, gpuInitited, if set to true, agent would use the backends that can deliver this, and also have the flag set (supportsTreqGpu which I suggested to change to supportsGpuInitiated)

src/api/cpp/nixl_types.h
tstamler: The tricky part about this ReqH is it needs to be exposed to the user to give to the CUDA kernel

src/api/cpp/backend/backend_plugin_doca_device.cuh
e-ago: it can be wherever, but when installing nixl, this header must be public so the application can include it

src/api/cpp/backend/backend_engine.h
e-ago: changing the name. Function already added in all the other backend

src/api/cpp/nixl_types.h
e-ago: @mkhazraee  The application needs visibility into the `nixlXferReqHGpu` object (can't just do a forward declaration) as it has to launch a CUDA kernel passing the GPU communications handlers. I can create a class like this in `transfer_request.h` : 
```
class nixlXferReqHGpu : public nixlXferReqH {
    private:
        uintptr_t backendHandleGpu;

    public:
        uintptr_t getGpuInitiatedHandle() const {
            return backendHandleGpu;
        }

        inline nixlXferReqHGpu() { }

        inline ~nixlXferReqHGpu() { }

    friend class nixlAgent;
};
```

but then, from the application I get this error:
```
../test/nixl/nixl_doca_device_test.cu(304): error: pointer or reference to incomplete type "nixlXferReqHGpu" is not allowed
          launch_initiator_send_kernel_v2(stream, buf[0].addr, buf[0].len, gtreq->getGpuInitiatedHandle());
```

This is because `transfer_request.h` is not public header visible at application level.

src/api/cpp/backend/backend_engine.h
```diff
@@ -114,6 +117,16 @@ class nixlBackendEngine {
...
+        virtual nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
+                                        const nixl_meta_dlist_t &local,
+                                        const nixl_meta_dlist_t &remote,
+                                        const std::string &remote_agent,
+                                        nixlBackendReqH* &handle,
+                                        uintptr_t &handle_gpu,
```
tstamler: I don't think we need this separate prepXfer function anymore if we also have the getGpuXferH

src/plugins/cuda_gds/gds_backend.h
```diff
@@ -161,6 +165,16 @@ class nixlGdsEngine : public nixlBackendEngine {
...
 
+        nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
+                                const nixl_meta_dlist_t &local,
+                                const nixl_meta_dlist_t &remote,
+                                const std::string &remote_agent,
+                                nixlBackendReqH* &handle,
+                                uintptr_t &handle_gpu,
+                                const nixl_opt_b_args_t* opt_args) {
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
```
tstamler: as before, I don't think we need this anymore

src/plugins/doca/doca_backend.cpp
```diff
@@ -870,26 +869,125 @@ nixl_status_t nixlDocaEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+}
+
+nixl_status_t nixlDocaEngine::prepXfer (const nixl_xfer_op_t &operation,
+	const nixl_meta_dlist_t &local,
+	const nixl_meta_dlist_t &remote,
+	const std::string &remote_agent,
+	nixlBackendReqH* &handle,
+	uintptr_t &handle_gpu,
+	const nixl_opt_b_args_t* opt_args)
+{
```
tstamler: as before, I don't think this version of prepXfer is needed anymore

src/plugins/doca/doca_backend.h
```diff
@@ -237,6 +231,14 @@ class nixlDocaEngine : public nixlBackendEngine {
                                 nixlBackendReqH* &handle,
                                 const nixl_opt_b_args_t* opt_args=nullptr);
 
+        nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
+                                    const nixl_meta_dlist_t &local,
+                                    const nixl_meta_dlist_t &remote,
+                                    const std::string &remote_agent,
+                                    nixlBackendReqH* &handle,
+                                    uintptr_t &handle_gpu,
+                                    const nixl_opt_b_args_t* opt_args);
```
tstamler: one more

src/plugins/ucx/ucx_backend.h
```diff
@@ -238,6 +239,16 @@ class nixlUcxEngine : public nixlBackendEngine {
...
 
+        nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
+                                    const nixl_meta_dlist_t &local,
+                                    const nixl_meta_dlist_t &remote,
+                                    const std::string &remote_agent,
+                                    nixlBackendReqH* &handle,
+                                    uintptr_t &handle_gpu,
+                                    const nixl_opt_b_args_t* opt_args) {
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
```
tstamler: one more

src/plugins/ucx_mo/ucx_mo_backend.h
```diff
@@ -206,6 +207,16 @@ class nixlUcxMoEngine : public nixlBackendEngine {
...
 
+    nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
+                                const nixl_meta_dlist_t &local,
+                                const nixl_meta_dlist_t &remote,
+                                const std::string &remote_agent,
+                                nixlBackendReqH* &handle,
+                                uintptr_t &handle_gpu,
+                                const nixl_opt_b_args_t* opt_args) {
+        return NIXL_ERR_NOT_SUPPORTED;
+    }
```
tstamler: one more

tstamler (approved):
Next things to think about are the best ways to do checkXferReq and invalidateXferReq

e-ago:
is it ok if I create another PR for it?

src/api/cpp/backend/backend_engine.h
```diff
@@ -85,6 +85,9 @@ class nixlBackendEngine {
         // Determines if a backend supports progress thread.
         virtual bool supportsProgTh () const = 0;
 
+        // Determines if a backend supports posting transfers requests from GPU.
+        virtual bool supportsGpuInitiated () const = 0;
```
mkhazraee: Let's make it not pure virtual, as it would break the build for all other backends. 
virtual bool supportsGpuInitiated () const { return false; }
(We plan to do the same for supportsProgTh and supportNotifs too in a separate PR.)

src/api/cpp/backend/backend_engine.h
```diff
@@ -129,6 +132,8 @@ class nixlBackendEngine {
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) = 0;
 
+        //Return GPU handler for Xfer Request Handler, if supported
+        virtual nixl_status_t getGpuXferH(const nixlBackendReqH* handle, nixlXferReqHGpu* gpu_hndl) = 0;
```
mkhazraee: Similar, we don't define methods specific to a flag to be pure virtual. 
virtual nixl_status_t getGpuXferH(const nixlBackendReqH* handle, nixlXferReqHGpu* gpu_hndl) {
    return NIXL_ERR_BACKEND;
}

src/core/nixl_agent.cpp
```diff
@@ -612,10 +612,12 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
                          const nixl_xfer_dlist_t &remote_descs,
                          const std::string &remote_agent,
                          nixlXferReqH* &req_hndl,
-                         const nixl_opt_args_t* extra_params) const {
+                         const nixl_opt_args_t* extra_params) const
+{
```
mkhazraee: Let's not make such styling changes, very soon we would have a proper clang formatter.

src/core/nixl_agent.cpp
```diff
@@ -659,7 +661,8 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     // TODO: merge descriptors back to back in memory (like makeXferReq).
     // TODO [Perf]: Avoid heap allocation on the datapath, maybe use a mem pool
 
-    nixlXferReqH *handle   = new nixlXferReqH;
+    handle = new nixlXferReqH;
```
mkhazraee: Same comment about styling.

src/core/nixl_agent.cpp
```diff
@@ -817,6 +826,14 @@ nixlAgent::queryXferBackend(const nixlXferReqH* req_hndl,
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::getGpuXferH (const nixlXferReqH* req_hndl,
+             nixlXferReqHGpu* gpu_hndl) const {
+    req_hndl->engine->getGpuXferH(req_hndl->backendHandle, gpu_hndl);
```
mkhazraee, line 834: Let's add a check if the backend supports gpu initiated, user might call it on a wrong transfer request, and also that method might return error, so we should return ret.

src/core/nixl_agent.cpp
```diff
@@ -817,6 +826,14 @@ nixlAgent::queryXferBackend(const nixlXferReqH* req_hndl,
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::getGpuXferH (const nixlXferReqH* req_hndl,
+             nixlXferReqHGpu* gpu_hndl) const {
```
mkhazraee: nixlXferReqHGpu* &gpu_hndl --> it's reference pointer since it's output.

src/plugins/cuda_gds/gds_backend.h
mkhazraee, line 1: With the fixed mentioned above, this change is unnecessary.

src/plugins/ucx/ucx_backend.h
mkhazraee, line 1: With the fixed mentioned above, this change is unnecessary.

src/plugins/ucx_mo/ucx_mo_backend.h
mkhazraee, line 1: With the fixed mentioned above, this change is unnecessary.

mkhazraee:
I've put some comments about the API changes to make them cleaner. 

Talking with Gal, one point seemed unclear. createXferReq and makeXferReq don't do any registrations, they're just some internal nixl bookkeeping to find the relevant metadata handler pointers for the backend. Also the transfers are not know at the beginning of the code, they need to be created within the code. So it potentially would be more beneficial if we make them also as a cuda kernel, and then postXfer and checkStatus and Notif would be also cuda kernal. In other words we can keep registration and metadata handling in cpu, but the transfer APIs in cuda. Going just some transfer methods adds some back and forth between cpu and gpu, potentially killing the benefits, at least for doing kernel fusion. Still there are some issues with device mode, for instance the remote agents can be added and removed during the life time of application, and they impact what transfer handles can be created or not. If we do the metadata management in cpu, then that calls for some shared memory or some more back and forth between cpu and gpu to be figured out (this is more control path at least, the creation of transfer handles are very dynamic, not complete data path, but could be right before each transfer).

With this said, I think we need a doc on pros and cons of each approach, between cudastreams and full cuda that can be fused. Not only from the high level benefits of kernel fusion, but specifically for an example use case of nixl, where transfers are created in the middle of the runtime based on user enquiries, or remote agents can come and go. Now what part of code needs to be run on cpu and what part on GPU. I suggest to make this doc and have to get on the same page and have a clear vision of the two implementations, then we can make the decision which one has higher priority and more benefits (ran this with Omri too). We can talk over the doc or 1-to-1 to speed up the process.

e-ago:
I've applied all the remaining comments. @mkhazraee @tstamler  please approve . Will create another PR once we consolidate the design considerations.
This way, in parallel, I can start measuring performance with benchmarks

mkhazraee (approved):

PR #169, opened 2025-04-19
Add Mooncake backend in NIXL

alogfans:
Mooncake transfer engine is a high-performance, zero-copy data transfer library. To achieve better performance in NIXL, we have designed an new backend based on Mooncake transfer engine. 

# Usage
1. Build the install [Mooncake](https://github.com/kvcache-ai/Mooncake) manually:
```
git clone https://github.com/kvcache-ai/Mooncake.git
cd Mooncake
bash dependencies.sh
mkdir build
cd build
cmake .. -DBUILD_SHARED_LIBS=ON
make -j
sudo make install # compulsory
```
2. Build NIXL with option `enable_mooncake_backend` set as true
3. You can use Mooncake Backend if you create backend by `agent.createBackend("Mooncake", init_param, backend)`. See `examples/cpp/nixl_benchmark.cpp` as an example.


alogfans:
It's ready for review. Could the NIXL community take a look?

stmatengss:
@mkhazraee Hi, PTAL. Hope to see your review. THX!

src/plugins/mooncake/meson.build
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cpp = meson.get_compiler('cpp')
+mooncake_lib = cpp.find_library('transfer_engine')
+
+compile_flags = []
+if cuda_dep.found()
+    compile_flags = [ '-DHAVE_CUDA' ]
+endif
```
tstamler: do we need CUDA in the plugin itself? just from a quick look I'm not seeing any meaningful usage

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include <net/if.h>
+#include <netdb.h>
+#include <sys/socket.h>
+
+#ifdef HAVE_CUDA
+
+#include <cuda_runtime.h>
+#include <cufile.h>
+
+#endif
```
tstamler: same as above, I'm not seeing any CUDA being used in the backend itself, so no need to include these

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+nixlMooncakeEngine::nixlMooncakeEngine (const nixlBackendInitParams* init_params)
+: nixlBackendEngine (init_params) {
+    local_agent_name_ = init_params->localAgent;
+    auto ips = findLocalIpAddresses();
+    std::string segment_name = "127.0.0.1";
+    if (!ips.empty()) segment_name = ips[0];
+    if (getenv("NIXL_MOONCAKE_IP_ADDR"))
+        segment_name = std::string(getenv("NIXL_MOONCAKE_IP_ADDR"));
```
tstamler, line 64: we now support P2P metadata transfer (as well as through ETCD) in the NIXL northbound API itself (see sendLocalMD, fetchRemoteMD), it would be better to leverage these instead of establishing a socket connection in the backend itself Generally we don't want to require environment variables except for advanced features, but I see here its optional to default to the findLocalIPAddresses() function

tstamler: Having said all this, if its difficult to decouple the P2P socket setup from the mooncake TransferEngine object, it might be ok to do it this way. 

Looking for others' opinions

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+    auto priv = (nixlMooncakeBackendMD *) meta;
+    int err = unregisterLocalMemory(engine_, priv->addr);
+    delete priv;
+    return err == 0 ? NIXL_SUCCESS : NIXL_ERR_BACKEND;
+}
+
+nixl_status_t nixlMooncakeEngine::getPublicData (const nixlBackendMD* meta,
+                                                 std::string &str) const {
+    return NIXL_SUCCESS;
+}
```
tstamler, line 171: typically, this would be how you would expose metadata about a descriptor that you need to publicize for other agents to be able to transfer to it. 

For example, with UCX and other RDMA-based transports, one node would need to provide an rkey for a memory region in order for remote nodes to access that memory. Maybe the mooncake TransferEngine does this exchange separately behind the scenes

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+nixl_status_t
+nixlMooncakeEngine::loadLocalMD (nixlBackendMD* input,
+                                 nixlBackendMD* &output)
+{
+    output = nullptr;
+    return NIXL_SUCCESS;
+}
```
tstamler, line 181: this function would be how we would support local transfers, meaning performing transfers between buffers registered to the same agent. Since nothing seems to be happening here, I would assume the Mooncake engine would not support such transfers, which is ok, just worth noting.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+nixl_status_t nixlMooncakeEngine::loadRemoteMD (const nixlBlobDesc &input,
+                                                const nixl_mem_t &nixl_mem,
+                                                const std::string &remote_agent,
+                                                nixlBackendMD* &output)
+{
+    output = nullptr;
+    return NIXL_SUCCESS;
+}
```
tstamler, line 192: this is the remote complementary function to getPublicData. Using rkeys as an example again, when one node exports the rkey and transfers it to another node through our metadata transfer APIs, this function is called by the NIXL agent to interpret the received metadata. In the rkey case, it can be stored in the nixlBackendMD that is returned from this function.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+nixl_status_t nixlMooncakeEngine::unloadMD (nixlBackendMD* input) {
+    return NIXL_SUCCESS;
+}
```
tstamler, line 199: this would be how you destroy a metadata object, but as seen we aren't returning metadata objects

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+nixl_status_t nixlMooncakeEngine::connect(const std::string &remote_agent) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (!connected_agents_.count(remote_agent)) {
+        connected_agents_[remote_agent] = AgentInfo{};
+    }
+    return NIXL_SUCCESS;
+}
```
tstamler, line 99: I don't think this is quite proper usage for connect. Connect is meant to do any required connection setup that happens _after_ loadRemoteConnInfo is called. In some cases, there might not need to be any extra setup that happens. My main concern here is overwriting the remote_agent map entry to an empty struct.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (target_id < 0) return NIXL_ERR_BACKEND;
+    transfer_request_t *request = new transfer_request_t[request_count];
+    for (size_t index = 0; index < request_count; ++index) {
+        if (local[index].len != remote[index].len) return NIXL_ERR_INVALID_PARAM;
+        request[index].opcode = (operation == NIXL_READ) ? OPCODE_READ : OPCODE_WRITE;
+        request[index].source = (void *)local[index].addr;
+        request[index].target_offset = remote[index].addr;
+        request[index].length = local[index].len;
+        request[index].target_id = target_id;
+    }
```
tstamler, line 250: prepXfer is where we typically do pre-processing for transfer requests, and then postXfer is where we actually submit the transfer. It might make sense to move some of this work to prepXfer instead (especially openSegment) but only if you think it makes sense.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -0,0 +1,252 @@
...
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (!connected_agents_.count(remote_agent))
+            return NIXL_ERR_INVALID_PARAM;
+        remote_ip_and_port = connected_agents_[remote_agent].ip_and_port;
+    }
+    if (local.descCount() != remote.descCount()) return NIXL_ERR_INVALID_PARAM;
+    size_t request_count = local.descCount();
+    uint64_t batch_id = allocateBatchID(engine_, request_count);
+    if (batch_id == INVALID_BATCH) return NIXL_ERR_BACKEND;
+    auto target_id = openSegment(engine_, remote_ip_and_port.c_str());
```
tstamler: I'm worried that this is doing TCP operations in the fast path, at least on the first open, so it might be worth looking into putting this in prepXfer or even better just connect if possible

src/plugins/mooncake/mooncake_backend.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include "transfer_engine_c.h"
+
+class nixlMooncakeEngine : public nixlBackendEngine {
+    public:
+        nixlMooncakeEngine(const nixlBackendInitParams* init_params);
+        ~nixlMooncakeEngine();
+
+        bool supportsRemote () const { return true; }
+        bool supportsLocal () const { return true; }
```
tstamler, line 44: as mentioned before, if this backend doesn't support transfers within the same agent, this should be marked false. Maybe you allow this transparently, I'm just not sure.

tstamler (commented):
This is a really good first implementation for the backend! Just a few additional notes:

For new plugins, we really want to see unit tests we can add for CI if we are going to merge the plugin into the main tree, not just performance benchmarks. 

I saw this requires git cloning and installing Mooncake manually. To more easily integrate this, I would encourage you to look into making this a CMake subproject inside of meson, maybe with a wrap file. Otherwise, we can look into adding instructions to a README specific to this backend. 

I'm hoping to get others to look as well. Thanks!

src/plugins/mooncake/mooncake_backend.cpp
alogfans: 1. Users don't need to specify environment variable in most case, except they found any connection issues by default settings. 
2. Metadata management and data transfer are currently tightly coupled in Mooncake Transfer Engine. We appreciate NIXL's design and will refactor our code, and give options to allow users manage metadata exchange by themselves. But perhaps not very soon.

src/plugins/mooncake/meson.build
alogfans: We have remove it in the updated commit.

src/plugins/mooncake/mooncake_backend.h
alogfans: Mooncake supports both local and remote transfer.

alogfans:
@tstamler We have updated our code to respond your good comments. 
- We port the unit test code from UCX backend to the Mooncake backend, and we have checked that it works well.
- We move the metadata exchange step (i.e. open segment in the context of Mooncake Transfer Engine) to the loadRemoteConnInfo function, so that postXfer contains only transferring steps. We also add comments about the functions that we purposely set them as no-action.

Happy International Workers' Day :)

tstamler:
@alogfans 

Hi again, thanks for updating! We have a handful of follow up requests, but then I think we will be able to merge an initial Mooncake plugin.

- We still want to improve the build environment. By default, we think the Mooncake plugin should be disabled to accomdate users who don't have Mooncake installed.  Alternatively, setting up meson so that it will automatically detect if Mooncake is available to build would be better. Otherwise, a simple bash script to clone and install mooncake would be helpful.
- Update the README and add inline TODO comments in the unimplemented metadata functions as planned for future, and maybe some high level overview of what happens under the hood for metadata during a transfer. Our concern is in having a plugin in the main tree that other users might look at and try to emulate. while the APIs are not fully implemented. It might cause confusion down the road. Also stating this is initial implementation should alleviate this problem.
- I'm concerned there might be a memory leak here: https://github.com/alogfans/nixl/blob/e5346bdee7295d718f11c5d627161443a87e4c05/src/plugins/mooncake/mooncake_backend.cpp#L222 with this transfer_request_t being allocated with new and not deleted.
- There might be another memory leak in the case when xferReqH are reposted. NIXL supports reposting the same request handles multiple times without releasing it in between. In your postXfer implementation, I think this will work, but it will overwrite the existing nixlMooncakeBackendReqH, creating a memory leak. You can see how the UCX backend tries to reuse reqH here: https://github.com/ai-dynamo/nixl/blob/3298d77174dd2580c1c114272f34febd908d579a/src/plugins/ucx/ucx_backend.cpp#L841
- We will work to verify the unit test you provided, and then add it to our CI once the initial PR is done. 

If you can address these, I think we'll be in a position to merge an initial version into our repo. Thanks again for your work!

alogfans:
@tstamler  We have addressed all issues in the latest commits.

src/plugins/mooncake/meson.build
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cpp = meson.get_compiler('cpp')
+mooncake_lib = cpp.find_library('transfer_engine')
+
+compile_flags = []
+
+if 'Mooncake' in static_plugins
```
aranadive, line 21: Since you are building a static plugin as well, please modify src/core/meson.build to link it to the plugin manager.

alogfans: Fixed.

PR #172, opened 2025-04-22
backend:ucx: Implement handler reuse

artpol84:
Enable handler reuse

### DIFF
--- a/test/unit/plugins/ucx/ucx_backend_test.cpp
+++ b/test/unit/plugins/ucx/ucx_backend_test.cpp
@@ -40,6 +40,72 @@ static void checkCudaError(cudaError_t result, const char *message) {
 #endif
 
 
+class testHndlIterator {
+private:
+    bool reuse;
+    bool set;
+    bool prepare;
+    bool release;
+    nixlBackendReqH* handle;
+public:
+    testHndlIterator(bool _reuse) {
+        reuse = _reuse;
+        if (reuse) {
+            prepare = true;
+            release = false;
+        } else {
+            prepare = true;
+            release = true;
+        }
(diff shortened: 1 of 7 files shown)


test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -624,18 +708,25 @@ int main()
...
         if (n_vram_dev > 1) {
-            test_inter_agent_transfer(thread_on[i],
+            test_inter_agent_transfer(thread_on[i], false,
+                                      ucx[i][0], VRAM_SEG, dev_ids[0],
+                                      ucx[i][1], VRAM_SEG, dev_ids[1]);
+            test_inter_agent_transfer(thread_on[i], true,
                                       ucx[i][0], VRAM_SEG, dev_ids[0],
                                       ucx[i][1], VRAM_SEG, dev_ids[1]);
-            test_inter_agent_transfer(thread_on[i],
+              test_inter_agent_transfer(thread_on[i], true,
```
tstamler: extra spaces

tstamler (approved):

yosefe:
@artpol84 seems this PR complicates the code, does it provide any performance improvement?

artpol84:
@yosefe it doesn't improve the performance. It addresses the missing functionality.
The backend is expected to allow reusing the handle, and UCX backend doesn't allow it causing the requirement to have a workaround that this PR is removing.
I was requested to fix it

artpol84:
Also, it's not so much of a complication but rather restructuring to enable missing functionality.
In some ways, the code became simpler in my opinion.

PR #180, opened 2025-04-22
Debug infra: Add NIXL logging and assertion macros

ovidiusm:
Note: this is a clone of PR https://github.com/ai-dynamo/nixl/pull/163 which was based on a different branch

Debug infra: Add NIXL logging macros

# Goal

The goal of the debug infrastructure is to plant extensive checks that we can remove in release builds since they may hurt performance; may help NIXL users in debug builds to develop their applications by spotting misuse of the API, unexpected or should-not-have-happened cases.

In general, follow the mature logging support design from UCX.

# Implementation

Leverage Abseil support, which provides logging functionality.

Define:

* Logging macros with different log levels (NIXL_ERROR, NIXL_LOG etc.) that can be selectively compiled out in release builds (by setting ABSL_MIN_LOG_LEVEL during build) and can also be configured at runtime
* Checks with assertions, that are either removed from release builds (NIXL_ASSERT) or more important checks that are compiled in all builds (NIXL_ASSERT_ALWAYS), similar to ucs_assert/ucs_assert_always
* Set default log level at compile time (strips logs/asserts)
* Configurable log level at runtime based on env variable NIXL_LOG_LEVEL

# Out of scope

NIXL code to be updated separately to replace existing stdout/err prints with the logs, and the asserts with the new checks.

# Deps

https://github.com/ai-dynamo/nixl/pull/137 needs to be merged first


meson.build
```diff
@@ -71,6 +71,43 @@ if get_option('disable_gds_backend')
...
+# Logging
+
+# Determine the actual log level based on user choice or build type
+log_level_opt = get_option('log_level')
+buildtype = get_option('buildtype')
+
+actual_log_level = log_level_opt
+if actual_log_level == 'auto'
+  if buildtype == 'debug'
+    actual_log_level = 'debug'
```
brminich: should it be `trace`?

meson.build
```diff
@@ -71,6 +71,43 @@ if get_option('disable_gds_backend')
...
+  else # release, debugoptimized, or plain
+    actual_log_level = 'error'
```
brminich: i'd also include `warning`

src/utils/common/nixl_log.cpp
```diff
@@ -0,0 +1,90 @@
...
+void InitializeNixlLogging() __attribute__((constructor));
+
+void InitializeNixlLogging()
+{
+    // Initialize Abseil logging system.
+    absl::InitializeLog();
+
+    // First, check for runtime override via environment variable
+    const char* env_log_level = std::getenv("NIXL_LOG_LEVEL");
+    if (env_log_level != nullptr) {
```
brminich, line 63: maybe make it more compact using some map containing log level name as a key and pair of `absl::LogSeverityAtLeast:` ans int as a value?

ovidiusm: The request I got in https://github.com/ai-dynamo/nixl/pull/163 was to avoid any level computation at runtime, instead have the level fixed during compilation. Hence the ifdefs. But I think they are quite clear.

etoledano: If there is a requirement for environment variable, then there shouldn't probably be another way to accomplish the same results.
I assumed we prefer doing it static (compilation) to have more control over logging, e,g. by zero overhead, locking and so on.

meson.build
```diff
@@ -71,6 +71,43 @@ if get_option('disable_gds_backend')
     add_project_arguments('-DDISABLE_GDS_BACKEND', language: 'cpp')
 endif
 
+# Logging
+
+# Determine the actual log level based on user choice or build type
+log_level_opt = get_option('log_level')
```
tstamler, line 77: is there any way to make the log level specified at runtime similar to UCX? a lot, if not most users will not be building from source and would still like to benefit from log printing

tstamler: yeah, I see there was discussion about it, but I think NIXL_LOG_LEVEL was removed...

etoledano: If runtime, shouldn't be changeable since MT.

ovidiusm: What we have in UCX is that logs below a minimum level are stripped from the binary for performance reasons in release builds, but users can still choose a level above the minimum at library initialization through an environment variable.

This is useful particularly in debug builds, since the most verbose level is very noisy. It can be annoying for the user to have to recompile and redeploy the binary to change the log level.

MT should not be a blocker for implementing this if we want it.

tstamler: @ovidiusm This sounds good to me and what we want, is it easy to enable this?

brminich: i strongly disagree with making log-level compile-time option. This is practically useless. We have to define an environment variable managing log level in run time. @ovidiusm cam you pls revert the last changes?
cc @yosefe

ovidiusm: Yes, I added back the code that had been removed before. Also unified the logic using a map. Thanks for the feedback.

src/utils/common/nixl_log.cpp
ovidiusm: We just need a fallback for when the env variable is not set. And this fallback needs to be different for debug and release builds (since we want to strip verbose logging in release for performance reasons).

This is exactly what the latest version does.

meson.build
```diff
@@ -71,6 +71,43 @@ if get_option('disable_gds_backend')
...
+buildtype = get_option('buildtype')
+
+actual_log_level = log_level_opt
+if actual_log_level == 'auto'
+  if buildtype == 'debug'
+    actual_log_level = 'trace'
+  else # release, debugoptimized, or plain
+    actual_log_level = 'warning'
```
brminich: sorry for confusion let's also have info in the release build, so at least some important information is available

src/utils/common/nixl_log.h
```diff
@@ -0,0 +1,84 @@
...
+ * Maps to Abseil LOG(FATAL). Use for unrecoverable errors.
+ */
+#define NIXL_FATAL LOG(FATAL)
+
+/* Logs messages unconditionally (maps to Abseil ERROR level) */
+#define NIXL_ERROR LOG(ERROR)
+
+/*
+ * Logs messages unconditionally (maps to Abseil WARNING level)
+ * Stripped from release buids.
```
brminich: it is not

src/utils/common/nixl_log.h
```diff
@@ -0,0 +1,84 @@
...
+ */
+#define NIXL_WARN LOG(WARNING)
+
+/*
+ * Logs messages unconditionally (maps to Abseil INFO level)
+ * Stripped from release buids.
```
brminich: pls make it available in release builds. But the default log level should be warn anyway

src/utils/common/nixl_log.cpp
```diff
@@ -0,0 +1,101 @@
...
+    // Check environment variable, it has priority over compile-time default.
+    const char* env_log_level = std::getenv("NIXL_LOG_LEVEL");
+    std::string env_level_str_upper;
+    if (env_log_level != nullptr) {
+        env_level_str_upper = absl::AsciiStrToUpper(env_log_level);
+        if (kLogLevelMap.contains(env_level_str_upper)) {
+            level_to_use = env_level_str_upper;
+        } else {
+            // Fall back to kDefaultLogLevel if env var is invalid
+            invalid_env_var = true;
```
brminich, line 69: what if we fallback here, but  just error level is defined at compile time

tstamler (approved):

PR #183, opened 2025-04-23
lthread: make socket lookup key both IP and port 

tstamler:
Make socket lookup key both IP and port to enable multiple agents from one remote node

### DIFF
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -36,6 +36,8 @@ typedef enum { SOCK_SEND, SOCK_FETCH, SOCK_INVAL } nixl_comm_t;
 // 4) Metadata to send (for sendLocalMD calls)
 typedef std::tuple<nixl_comm_t, std::string, int, nixl_blob_t> nixl_comm_req_t;
 
+typedef std::pair<std::string, int> nixl_socket_peer_t;
+
 class nixlAgentData {
     private:
         std::string     name;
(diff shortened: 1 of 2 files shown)


src/core/nixl_listener.cpp
```diff
@@ -132,8 +133,12 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
             int req_port = std::get<2>(request);
             std::string my_MD = std::get<3>(request);
 
+            nixl_socket_peer_t req_sock;
```
mkhazraee: I think you can init directly, or use std::make_pair.

src/core/nixl_listener.cpp
```diff
@@ -132,8 +133,12 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
...
+            req_sock.first = req_ip;
+            req_sock.second = req_port;
+
             //use remote IP for socket lookup
```
mkhazraee: Nit pick, add space after //

src/core/agent_data.h
```diff
@@ -36,6 +36,8 @@ typedef enum { SOCK_SEND, SOCK_FETCH, SOCK_INVAL } nixl_comm_t;
 // 4) Metadata to send (for sendLocalMD calls)
 typedef std::tuple<nixl_comm_t, std::string, int, nixl_blob_t> nixl_comm_req_t;
 
+typedef std::pair<std::string, int> nixl_socket_peer_t;
```
mkhazraee, line 39: I do like the name and debating if API should use this. Pros and cons:
- For ETCD both are not passed (if no ip, we don't care about port). --> nicer interface and naming with this
- For socket, they can skip the port in current implementation, or have to put 0 here. --> nicer in current implementation

tstamler: I opted to keep it the way it is because for ETCD they still don't have to and won't specify anything either way. The call will just be sendLocalMD(), but I'm open to moving it. As you said, users will just need to specify port 0 if they want to use default port, which isn't my favorite thing.

mkhazraee (approved):

PR #186, opened 2025-04-23
Fix typo in blocking_send_recv_example

VertexC:


tstamler:
Thanks for the suggestion, but this is already going to be fixed in #187 !

PR #187, opened 2025-04-24
python API: add metadata exchange calls

tstamler:
Extend the new NB APIs for metadata exchange to the python API, with updated example

### DIFF
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -576,6 +576,36 @@ class nixl_agent:
         return self.agent.getLocalMD()
 
     """
+    @brief Get the metadata of the local agent.
+
+    @return Metadata of the local agent, in bytes.
+    """
+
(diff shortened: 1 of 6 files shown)


src/api/python/_api.py
```diff
@@ -575,6 +588,38 @@ def send_notif(
...
+        self.agent.fetchRemoteMD(remote_agent, ip_addr, port)
+
+    """
+    @brief Invalidate your own metadata in the central metadata server, or from a specific peer.
+
+    @param ip_addr If specified, will only send invalidation to one peer by IP address.
+    @param port    If specified, will try to send to specific port.
+    """
+
+    def remove_local_metadata(self, ip_addr: str = "", port: int = 0):
```
mkhazraee: remove is confusing, invalidate is more clear for this scenario.

mkhazraee (approved):

PR #188, opened 2025-04-24
API: sendPartial and python API updates

tstamler:
- Add sendPartialLocalMD to API in both C++ and Python
- Add getPartialLocalMD API to Python
- New test for sending/transferring with partial metadata

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -337,6 +337,8 @@ class nixlAgent {
          *         backends supporting the memory type is also included.
          *         If `extra_params->backends` is non-empty, only the descriptors supported by the
          *         backends in the list and the backends' connection info are included in the metadata.
+         *         If 'extra_params->ip_addr' is set, the metadata will only be sent to a single peer.
+         *         If 'extra_params->port' can be set in addition to IP address.
          *
          * @param  descs         [in]  Descriptor list to include in the metadata
          * @param  str           [out] The serialized metadata blob
@@ -345,9 +347,7 @@ class nixlAgent {
          */
         nixl_status_t
         sendLocalPartialMD(nixl_reg_dlist_t  &descs,
-                           const nixl_opt_args_t* extra_params = nullptr,
-                           const std::string remote_ip = "",
-                           const int port = 0) const;
+                           const nixl_opt_args_t* extra_params = nullptr) const;
 
         /**
          * @brief  Fetch other agent's metadata and unpack it internally.
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -320,13 +320,14 @@ class nixlAgent {
         /**
          * @brief  Send your own agent metadata to a remote location.
          *
-         * @param  remote_ip     If specified, this is the remote IP address to send your metadata to
+         * @param  extra_params  Only to optionally specify IP address and/or port.
+         *                       If specified, this will enable peer to peer sending of your metadata.
          *                       If unspecified, this will send your data to the metadata server.
-         * @param  port          Specific port to connect to, otherwise use DEFAULT_COMM_PORT
+         *
          * @return nixl_status_t Error code if call was not successful
          */
         nixl_status_t
-        sendLocalMD (const std::string remote_ip = "", const int port = 0) const;
+        sendLocalMD (const nixl_opt_args_t* extra_params = nullptr) const;
 
         /**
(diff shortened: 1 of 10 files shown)


src/api/cpp/nixl.h
```diff
@@ -328,6 +328,27 @@ class nixlAgent {
...
+         *         backends in the list and the backends' connection info are included in the metadata.
+         *
+         * @param  descs         [in]  Descriptor list to include in the metadata
+         * @param  str           [out] The serialized metadata blob
+         * @param  extra_params  [in]  Optional extra parameters used in getting partial metadata
+         * @return nixl_status_t       Error code if call was not successful
+         */
+        nixl_status_t
+        sendLocalPartialMD(nixl_reg_dlist_t  &descs,
+                           const nixl_opt_args_t* extra_params = nullptr,
```
mkhazraee: extra_params should be last I would say. That might be a reason to merge ip/port to extra_params, hmm.

src/api/cpp/nixl.h
```diff
@@ -320,39 +320,61 @@ class nixlAgent {
         /**
          * @brief  Send your own agent metadata to a remote location.
          *
-         * @param  remote_ip     If specified, this is the remote IP address to send your metadata to
+         * @param  extra_params  Only to optionally specify IP address and/or port.
+         *                       If specified, this will enable peer to peer sending of your metadata.
          *                       If unspecified, this will send your data to the metadata server.
-         * @param  port          Specific port to connect to, otherwise use DEFAULT_COMM_PORT
+         *
```
mkhazraee, line 327: Let's mention the default port too. In peer to peer sending, if port is not specified, DEFAULT_COMM_PORT will be used.

src/api/cpp/nixl.h
```diff
@@ -320,39 +320,61 @@ class nixlAgent {
...
+         * @brief  Send partial metadata blob for this agent to peer or central metadata server
+         *         If `descs` is empty, only backends' connection info is included in the metadata,
+         *         regardless of the value of `extra_params->includeConnInfo` and `descs` memory type.
+         *         If `descs` is non-empty, the metadata of the descriptors in the list are included,
+         *         and if `extra_params->includeConnInfo` is true, the connection info of the
+         *         backends supporting the memory type is also included.
+         *         If `extra_params->backends` is non-empty, only the descriptors supported by the
+         *         backends in the list and the backends' connection info are included in the metadata.
+         *         If 'extra_params->ip_addr' is set, the metadata will only be sent to a single peer.
+         *         If 'extra_params->port' can be set in addition to IP address.
```
mkhazraee: weird sentence. Maybe similar to the previous comment.

src/api/cpp/nixl_types.h
```diff
@@ -140,6 +140,17 @@ class nixlAgentOptionalArgs {
...
         bool includeConnInfo = false;
+
+        /**
+         * @var ipAddr Used to specify the IP address of a remote peer for metadata transfer.
+         *                      used in sendLocalPartialMD.
+         */
+        std::string ipAddr;
+        /**
+         * @var port Used to specify the port of a remote peer, ipAddr must also be set
+         *                      used in sendLocalPartialMD.
```
mkhazraee: Let's update other methods too, and mention the default port here. Or even maybe initialize it to that? instead of checking in the file? How about this?

src/api/cpp/nixl.h
mkhazraee: Actually thinking more about it, let's just mention this in the nixl_types.h, the DEFAULT_COMM_PORT.

mkhazraee (approved):

src/core/nixl_agent.cpp
```diff
@@ -1050,52 +1050,65 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
...
+nixlAgent::sendLocalMD (const nixl_opt_args_t* extra_params) const {
 
-    if(remote_ip.size() == 0){
+    if(extra_params->ipAddr.size() == 0){
         std::cerr << "ETCD not supported yet, please specify IP\n";
         return NIXL_ERR_NOT_SUPPORTED;
     }
     int send_port = default_comm_port;
-    if(port != 0) send_port = port;
+    if(extra_params->port != 0) send_port = extra_params->port;
```
mkhazraee: I think send_port can be always extra_params->port.

mkhazraee (approved):

PR #198, opened 2025-04-25
Add NIXL logging to agent and plugin manager

ovidiusm:
# Goal

Replace std:cerr calls with log messages.

Will make several small PRs migrating gradually the codebase to log messages, instead of one large PR, due to the frequent conflicts.

# Implementation

* Added log messages to 2 files
* Fixed Abseil log level initialization ([log level has to be set](https://github.com/abseil/abseil.github.io/issues/467) before calling absl::InitializeLog; and absl::SetStderrThreshold is required)
* Made the utils/common library a dynamic library instead of static, otherwise we end up with multiple instances of it loaded that do multiple initializations


src/core/nixl_plugin_manager.cpp
```diff
@@ -292,9 +294,9 @@ void nixlPluginManager::discoverPluginsFromDir(const std::string& dirpath) {
 
             // Try to load the plugin
             auto plugin = loadPlugin(plugin_name);
-            // if (plugin) { // To be used in logging infra
-            //    std::cout << "Loaded plugin " << plugin_name << std::endl;
-            // }
+            if (plugin) {
+                NIXL_DEBUG << "Discovered and loaded plugin: " << plugin_name;
```
brminich: maybe this is something we can print in release? so use INFO here?

src/utils/common/nixl_log.cpp
```diff
@@ -93,6 +90,8 @@ void InitializeNixlLogging()
     const LogLevelSettings& settings = (it != kLogLevelMap.end()) ? it->second : kLogLevelMap.at(kDefaultLogLevel);
     absl::SetMinLogLevel(settings.min_severity);
     absl::SetVLogLevel("*", settings.vlog_level);
+    absl::SetStderrThreshold(settings.min_severity);
```
brminich, line 93: if want to use stderr, maybe use it for warnings (anf higher)?

ovidiusm: The idea is that abseil supports multiple log sinks (in case the program sends logs over the network to a collector) and one can set different logging levels for stderr and other sinks. But for now we are not using multiple sinks, so let's set it the same until we add sinks if we need to.

src/core/nixl_plugin_manager.cpp
aranadive: Will be very verbose in release. Debug is fine.

aranadive (approved):

PR #199, opened 2025-04-25
Add ETCD support for NIXL MD

aranadive:


src/core/agent_data.h
```diff
@@ -74,6 +78,16 @@ class nixlAgentData {
         void enqueueCommWork(nixl_comm_req_t request);
         void getCommWork(std::vector<nixl_comm_req_t> &req_list);
 
+        // ETCD-related methods and data
+        std::string makeEtcdKey(const std::string& metadata_type) const;
+        nixl_status_t storeMetadataInEtcd(const std::string& metadata_type, const nixl_blob_t& metadata) const;
+        nixl_status_t fetchMetadataFromEtcd(const std::string& agent_name, const std::string& metadata_type, nixl_blob_t& metadata) const;
+        nixl_status_t removeMetadataFromEtcd(const std::string& metadata_type) const;
```
tstamler: these don't actually seem to use anything from nixlAgentData, I think they should be a standalone class probably. Or at least just independent functions in nixl_listener.cpp. But maybe I'm missing something

src/api/cpp/nixl_params.h
```diff
@@ -33,8 +33,12 @@ class nixlAgentConfig {
         bool     useProgThread;
         /** @var Enable listener thread */
         bool     useListenThread;
+	    /** @var Enable etcd connection */
+	    bool	 useEtcd;
         /** @var Port for listener thread to use */
         int      listenPort;
+	    /** @var IP:PORT to connect for etcd server */
+	    std::string etcd_ep;
```
tstamler: I think we can simplify this. I don't think specifying the ETCD endpoint in the code makes sense, as people will usually want to run it on a deployment where they can automate things and not need to recompile their applications. 

My suggestion is we can just make it reliant on an environment variable for users that want to use ETCD, since the barrier for entry for ETCD users will already be higher. If the env is set, you can try to use ETCD with the metadata functions, if its not we return an error. What do you think?

src/core/nixl_agent.cpp
```diff
@@ -22,6 +22,9 @@
 #include "transfer_request.h"
 #include "agent_data.h"
 #include "plugin_manager.h"
+#include <etcd/Client.hpp>
+
+std::string default_etcd_namespace = "/nixl/agents";
```
tstamler: probably should make this an env too, but this is probably a good default

examples/cpp/nixl_etcd_example.cpp
mkhazraee, line 1: Not a big deal, maybe good to do a single transfer too. Later we can make it more complete to fail on partial load that was not available, but not that important now.

meson_options.txt
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
```
mkhazraee, line 14: Should we have a enable_etcd flag, if not set, not to use it and depend on it?

src/api/cpp/nixl_params.h
mkhazraee: We can print a warning, if ETCD env vars are set but we can't connect to it in constructor. Otherwise, add a new error type for etcd, if user uses etcd methods. Also if listener thread is not set, all the p2p or etcd methods should return error. @tstamler I forgot if we already do this check or it automatically happens.

One minor concern is if we want to connect to ETCD during constructor, it might take time, so better to be in listener thread. In that scenario we might need to print the warning from the listener thread that can get tricky.

src/api/cpp/nixl_params.h
```diff
@@ -61,16 +65,21 @@ class nixlAgentConfig {
          * @param use_listen_thread  flag to determine use of listener thread
          * @param port               specify port for listener thread to listen on
          * @param pthr_delay_us      Optional delay for pthread in us
-         * @param pthr_delay_us      Optional delay for listener thread in us
+         * @param lthr_delay_us      Optional delay for listener thread in us
+	     * @param etcd_ep            Optional endpoint in IP:PORT to connect to etcd
```
mkhazraee: please replace with space instead of tab

src/api/cpp/nixl_params.h
```diff
@@ -61,16 +65,21 @@ class nixlAgentConfig {
...
          */
         nixlAgentConfig (const bool use_prog_thread,
                          const bool use_listen_thread=false,
+			             const bool use_etcd=false,
```
mkhazraee: tabs

src/core/agent_data.h
mkhazraee: Sounds more like the metadata_stream, so better to put them in a separate file there. Specially if we want to support configurable use of etcd during compile.

src/core/nixl_agent.cpp
mkhazraee: I wonder if this should go to nixl_types.h, or to the new etcd file we make, so later we figure the best practice.

src/core/nixl_agent.cpp
```diff
@@ -101,6 +104,17 @@ nixlAgent::nixlAgent(const std::string &name,
         data->commThreadStop = false;
         data->commThread = std::thread(&nixlAgentData::commWorker, data, this);
     }
+    try {
+        if (cfg.useEtcd && !cfg.etcd_ep.empty()) {
```
mkhazraee: Do we connect to etcd through this code?

src/core/nixl_agent.cpp
```diff
@@ -1051,64 +1065,78 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
...
-    data->enqueueCommWork(std::make_tuple(SOCK_SEND, extra_params->ipAddr, extra_params->port, myMD));
+    // If IP is provided, use socket-based communication
+    if (extra_params && !extra_params->ipAddr.empty()) {
+        data->enqueueCommWork(std::make_tuple(SOCK_SEND, extra_params->ipAddr, extra_params->port, myMD));
+        return NIXL_SUCCESS;
+    }
 
-    return NIXL_SUCCESS;
+    // If no IP is provided, use etcd
+    return data->storeMetadataInEtcd("metadata", myMD);
```
mkhazraee: Is it a separate process? If not I think better to go to listener thread, sounds like a blocking connection.

src/core/nixl_agent.cpp
```diff
@@ -1051,64 +1065,78 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
...
+    // If IP is provided, use socket-based communication
+    if (extra_params && !extra_params->ipAddr.empty()) {
+        data->enqueueCommWork(std::make_tuple(SOCK_FETCH, extra_params->ipAddr, extra_params->port, ""));
+        return NIXL_SUCCESS;
     }
 
-    data->enqueueCommWork(std::make_tuple(SOCK_FETCH, extra_params->ipAddr, extra_params->port, ""));
+    // If no IP is provided, use etcd
+    nixl_blob_t remote_metadata;
+    nixl_status_t ret = data->fetchMetadataFromEtcd(remote_name, "metadata", remote_metadata);
```
mkhazraee: def blocking, should go to listener thread.

src/core/nixl_agent.cpp
```diff
@@ -1051,64 +1065,78 @@ nixlAgent::getLocalPartialMD(nixl_reg_dlist_t &descs,
...
-        return NIXL_ERR_NOT_SUPPORTED;
+    // If IP is provided, use socket-based communication
+    if (extra_params && !extra_params->ipAddr.empty()) {
+        data->enqueueCommWork(std::make_tuple(SOCK_INVAL, extra_params->ipAddr, extra_params->port, ""));
+        return NIXL_SUCCESS;
     }
 
-    data->enqueueCommWork(std::make_tuple(SOCK_INVAL, extra_params->ipAddr, extra_params->port, ""));
+    // If no IP is provided, use etcd
+    nixl_status_t ret1 = data->removeMetadataFromEtcd("metadata");
```
mkhazraee: Again I think blocking.

src/core/agent_data.h
aranadive: These use all the etcd related info in AgentData. But perhaps if we rely on the listener thread to enqueue etcd operations then these can be local to listener thread.

src/core/nixl_agent.cpp
aranadive: Just used an env var now.

src/api/cpp/nixl_params.h
aranadive: Removed this.

src/core/nixl_agent.cpp
aranadive: Moved this to listener.

src/core/nixl_agent.cpp
aranadive: Moved.

src/core/nixl_agent.cpp
aranadive: In listener now.

src/core/nixl_agent.cpp
aranadive: Its async from etcd api but lets move to listener thread.

meson_options.txt
aranadive: Not required now I think.

examples/cpp/nixl_etcd_example.cpp
aranadive: Done.

src/api/cpp/nixl_params.h
aranadive: We now only connect when the thread starts but verify/print in the constructor that etcd support is available.

tstamler (approved):

PR #202, opened 2025-04-25
refactor: updates for kv block manager

ryanolson:


src/bindings/rust/src/lib.rs
```diff
@@ -127,7 +128,7 @@ impl fmt::Display for MemType {
                 MemType::Block => 2,
                 MemType::Object => 3,
                 MemType::File => 4,
-                MemType::Unknown => 5
+                MemType::Unknown => 5,
```
roiedanino, line 131: Why? Convention?

ryanolson: that change was generate from fixing clippy warnings and/or `cargo fmt`

```
cargo clippy --fix --allow-dirty
cargo fmt
```

I don't define the rule, I just follow them.  In dynamo, CI will block us on clippy warnings.

tstamler (approved):

PR #206, opened 2025-04-27
Debug Infra: Fix NIXL log level

brminich:
Currently, `INFO` is used by default for release builds and `TRACE` for debug builds. Both `INFO` and `TRACE` should be available as the maximum log levels for release and debug builds respectively, but the default level should be set to `WARN`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -86,9 +86,6 @@ if actual_log_level == 'auto'
   endif
 endif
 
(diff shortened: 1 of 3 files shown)


yosefe (approved):

etoledano:
Do we have somewhere to document logging usage besides the meson? if makes sense...

PR #208, opened 2025-04-28
refactor: Prefer "using" over "typedef" guideline implementation

etoledano:
Apply `Prefer "using" over "typedef"` guidelines in src/ This commit doesn't change functionality, it is strictly visual.


etoledano:
Moving to fork

PR #209, opened 2025-04-28
nixlAgent: (refactor) Apply smart pointer resource management

etoledano:
Refactor nixlAgent code to comply with smart pointers resource management according to guidelines.

### DIFF
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -56,21 +56,21 @@ class nixlAgentData {
         std::unordered_map<nixl_backend_t, nixl_blob_t>   connMD;
 
         // Local section, and Remote sections and their available common backends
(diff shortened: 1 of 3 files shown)


yosefe:
can we make unique_ptr for nixlAgentData in nixlAgent so users cannot copy it (but only move)?

etoledano:
> can we make unique_ptr for nixlAgentData in nixlAgent so users cannot copy it (but only move)?

I would sure love to do that in another (slightly bigger) commit, explaining:

Currently `nixlAgent` is set to be the exclusive owner of `nixlAgentData` (or more precise - whoever is friend's with this class), which is done by making `nixlAgentData`'s destructor private and have `nixlAgent` as a friend, a private destructor prevents from using smart pointers without over complicating things.

PR #210, opened 2025-04-28
Agent API: support unloading parital remote MD

yanivz-nv:
- New API enum nixl_md_op_t as argument for getLocaPartialMD and sendLocalPartialMD to indicate if the MD is meant for remote load or unload.
- In loadRemoteMD load or unload based on the serialized op.
- Update nixlMemSection::sectionMap to hold unique_ptrs instead of raw ptrs for easier memory management.

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -328,6 +328,7 @@ class nixlAgent {
 
         /**
          * @brief  Get partial metadata blob for this agent, to be given to other agents.
+         *         `op` determines whether the metadata will be used for loading or unloading on the remote.
          *         If `descs` is empty, only backends' connection info is included in the metadata,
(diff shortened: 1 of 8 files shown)


src/api/cpp/nixl.h
```diff
@@ -349,6 +350,7 @@ class nixlAgent {
          */
         nixl_status_t
         sendLocalPartialMD(nixl_reg_dlist_t  &descs,
+                           nixl_md_op_t op,
```
mkhazraee: Let's put it as part of extra params, and a boolean flag to remove. By default it's set to False, if user wants to use it before deregister, we would.

That's not the intended use anyways, just enabling it for savy users, so let's not put it in the main API. Mention in the comments about the new field this is relevant for these 2 methods, and should be called before deregister.

yanivz-nv: Done.

PR #211, opened 2025-04-28
AGENT: Use std::unique_ptr for agent data

yosefe:
## Why
Use move semantics for nixlAgentData

### DIFF
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -96,6 +96,7 @@ class nixlAgentData {
         void enqueueCommWork(nixl_comm_req_t request);
         void getCommWork(std::vector<nixl_comm_req_t> &req_list);
 
+    public:
         nixlAgentData(const std::string &name, const nixlAgentConfig &cfg);
         ~nixlAgentData();
 
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -72,6 +72,8 @@ nixlAgentData::nixlAgentData(const std::string &name,
         NIXL_DEBUG << "NIXL ETCD is disabled";
     }
 #endif // HAVE_ETCD
+    if (name.empty())
+        throw std::invalid_argument("Agent needs a name");
 
     memorySection = new nixlLocalSection();
 }
@@ -98,13 +100,9 @@ nixlAgentData::~nixlAgentData() {
 }
 
 /*** nixlAgent implementation ***/
-nixlAgent::nixlAgent(const std::string &name,
-                     const nixlAgentConfig &cfg) {
-    if (name.size() == 0)
-        throw std::invalid_argument("Agent needs a name");
-
-    data = new nixlAgentData(name, cfg);
-
+nixlAgent::nixlAgent(const std::string &name, const nixlAgentConfig &cfg) :
+    data(std::make_unique<nixlAgentData>(name, cfg))
+{
     if(cfg.useListenThread) {
         int my_port = cfg.listenPort;
         if(my_port == 0) my_port = default_comm_port;
@@ -114,12 +112,13 @@ nixlAgent::nixlAgent(const std::string &name,
 
     if (data->useEtcd || cfg.useListenThread) {
         data->commThreadStop = false;
-        data->commThread = std::thread(&nixlAgentData::commWorker, data, this);
+        data->commThread =
+            std::thread(&nixlAgentData::commWorker, data.get(), this);
     }
 }
 
 nixlAgent::~nixlAgent() {
-    if (data->useEtcd || data->config.useListenThread) {
+    if (data && (data->useEtcd || data->config.useListenThread)) {
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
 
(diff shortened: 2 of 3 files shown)


src/core/agent_data.h
```diff
@@ -77,6 +77,8 @@ class nixlAgentData {
         void getCommWork(std::vector<nixl_comm_req_t> &req_list);
 
         nixlAgentData(const std::string &name, const nixlAgentConfig &cfg);
+
+    public:
         ~nixlAgentData();
```
etoledano, line 101: Moving the C'tor to public would allow you to use `make_unique` instead of `reset(new ...)`

src/core/nixl_agent.cpp
```diff
@@ -93,27 +93,31 @@ nixlAgent::nixlAgent(const std::string &name,
...
+    if(data && data->config.useListenThread) {
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
         if(data->listener) delete data->listener;
     }
-    delete data;
 }
 
+nixlAgent::nixlAgent(nixlAgent &&other) noexcept = default;
+nixlAgent& nixlAgent::operator=(nixlAgent &&other) noexcept = default;
```
etoledano, line 133: This can be in the header

src/core/nixl_agent.cpp
```diff
@@ -93,27 +93,31 @@ nixlAgent::nixlAgent(const std::string &name,
...
         data->commThreadStop = false;
-        data->commThread = std::thread(&nixlAgentData::commWorker, data, this);
+        data->commThread =
+            std::thread(&nixlAgentData::commWorker, data.get(), this);
     }
 }
 
 nixlAgent::~nixlAgent() {
-    if(data->config.useListenThread) {
+    if(data && data->config.useListenThread) {
```
etoledano: `data` can't be `nullptr` at this point

src/core/agent_data.h
yosefe: right, but it will allow user to create nixlAgentData and we want to prevent this

src/core/nixl_agent.cpp
yosefe: it can't, because it depends on knowing the details of nixlAgentData class (to generate unique_ptr operations) which is just forward declared in the header.

src/core/nixl_agent.cpp
yosefe: data can be empty unique_ptr because of move semantics of unique_ptr (when the content is moved to another nixlAgent). Actually the right thing is probably do everything in nixlAgentData ctor/dtor but it's a bigger change

src/core/agent_data.h
etoledano: Right, this relates to the refactor I was talking about in the previous commit.
[Pimpl](https://en.cppreference.com/w/cpp/language/pimpl) is a common way to do that, such that by force `nixlAgentData` will not be visible. In this case it would require moving some stuff around to hide the actual implementation.

src/core/nixl_agent.cpp
etoledano: It worked when I tried it, maybe I am missing something, resolving anyway.

yosefe: it didn't work when i tried to add nixlAgent to an std::vector (it will be in another PR)

src/core/agent_data.h
yosefe: do you mean that since nixlAgentData is not defined in API header files it's ok to make its constructor public? 
IMO a private ctor also prevents any class except the friend nixlAgent from creating it

etoledano: Having the header not exposed + using friendship + private c'tor is one way to control who can own it and who is exposed, which as you can tell creates unnecessary complications and in it for itself is complicated.

I am saying that `nixlAgentData` is simply an internal data structure for `nixlAgent`, that should be declared inside `nixl_agent.cpp` and then there is no need for tricks.

```
// nixl.h

class nixlAgent {
private:
  class nixlAgentData; // Forward declaration
  std::unique_ptr<nixlAgentData> data;
  ...
};

// nixl_agent.cpp

class nixlAgent::nixlAgentData {
  ...
};

nixlAgent::nixlAgent(): data(std::make_unique<nixlAgentData>(...) {
  ...
}

```

yosefe: Agree that nesting nixlAgentData as private class is better. So i'll make c'tor public in this PR and we can move 
nixlAgentData inside nixlAgent in next PR to make changes incremental, WDYT?

etoledano: SG!

yosefe: done

tstamler (approved):

src/core/nixl_agent.cpp
```diff
@@ -114,23 +112,26 @@ nixlAgent::nixlAgent(const std::string &name,
...
 }
 
 nixlAgent::~nixlAgent() {
-    if (data->useEtcd || data->config.useListenThread) {
+    if (data && (data->useEtcd || data->config.useListenThread)) {
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
 
         if(data->config.useListenThread) {
             if(data->listener) delete data->listener;
```
w1ldptr, line 126: Having to manually deallocate parts of 'data' on cleanup kinda defeats the purpose of automatic memory management with RAII. Looking at the code it is not immediately obvious to me why data->listener even needs to be a pointer in the first place, so consider refactoring it to be by-value member of nixlAgentData.

yosefe: Agree that nixlAgentData also needs refactoring. However IMO we need to protect nixlAgent usage first, since refactoring nixlAgentData can be a deeper change.

w1ldptr: I suggested this because making listener either by-value member, or, if not possible, another unique_ptr seemed like a trivial change to me that really makes nixlAgentData compatible with RAII idiom, but I'm not insisting on it.

PR #212, opened 2025-04-28
CI: Add AWS EFA testing infrastructure 

Alexey-Rivkin:
## What?
Add automated testing infrastructure for NIXL on AWS EFA environment via a new `contrib/aws-efa/ `module and GitHub Actions workflow.

## Why?
Ensure NIXL compatibility with AWS EFA environment by:

- Automate validation for C++ and Python implementations
- Verify AWS-specific networking features
- Catche EFA-specific regressions
- Support both CI and manual testing

## How?
Implement:

- Bash testing script for job management
- AWS Batch configuration files
- New GitHub Actions workflow for PRs
- Testing on NVIDIA PyTorch containers

### DIFF
--- a/.github/workflows/aws_efa_validation.yml
+++ b/.github/workflows/aws_efa_validation.yml
@@ -1,52 +1,66 @@
 name: NVIDIA NIXL AWS EFA Validation
 
 on:
-  push:
-    branches:
-      - main
-      - "pull-request/[0-9]+"
-  workflow_dispatch:
+  pull_request:
 
 jobs:
   run_aws_tests:
-    name: Run AWS Tests - ${{ matrix.test_type }}
+    name: Run AWS Tests - ${{ matrix.test_name }}
+    environment: SWX_AWS
     runs-on: ubuntu-latest
+    env:
+      AWS_DEFAULT_REGION: eu-central-1
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     strategy:
       fail-fast: false
       matrix:
-        test_type: [cpp, python]
-    container:
-      image: amazon/aws-cli
-      options: --entrypoint=""
+        include:
+          - test_name: "C++ Tests"
+            test_scripts:
+              - .gitlab/test_cpp.sh $NIXL_INSTALL_DIR
+          - test_name: "Python Tests"
+            test_scripts:
+              - .gitlab/test_python.sh $NIXL_INSTALL_DIR
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
 
-      - name: Install additional tools
+      - name: Setup
         run: |
-          yum install -y jq gettext tar gzip git
+          set -exE
+          sudo apt-get -qq update
+          sudo apt-get -qq install -y curl gettext git gzip jq tar unzip
+
+          # Install kubectl
           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
-          install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+          sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
 
-      - name: Configure AWS credentials
-        run: |
-          aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws configure set aws_secret_access_key ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws configure set region eu-central-1
+          # Install AWS CLI
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip -qq awscliv2.zip
+          sudo ./aws/install --update
+
+          # Verify AWS credentials
+          aws sts get-caller-identity >/dev/null
 
       - name: Run AWS tests
-        id: aws_tests
         working-directory: ./contrib/aws-efa
         timeout-minutes: 180
         run: |
-          ./aws_test.sh "${{ matrix.test_type }}" 2>&1 | tee test_output.log
-          JOB_ID=$(grep 'JOB_ID=' test_output.log | head -n1 | cut -d= -f2)
-          echo "job_id=$JOB_ID" >> $GITHUB_OUTPUT
+          set -o pipefail
+          test_cmd='${{ join(matrix.test_scripts, ' && ') }}'
+          ./aws_test.sh "$test_cmd" 2>&1 | tee test_output.log
 
       - name: Cleanup AWS resources
         if: always()
         working-directory: ./contrib/aws-efa
         run: |
+          JOB_ID=$(grep 'JOB_ID=' test_output.log | head -n1 | cut -d= -f2)
+          echo "JOB_ID=$JOB_ID"
           aws batch terminate-job \
-            --job-id ${{ steps.aws_tests.outputs.job_id }} \
+            --job-id "$JOB_ID" \
             --reason 'Terminated by GitHub Actions' || true
--- a/.github/workflows/aws_efa_validation.yml
+++ b/.github/workflows/aws_efa_validation.yml
@@ -1,7 +1,10 @@
-name: NVIDIA NIXL AWS EFA Validation
+name: AWS EFA NIXL Validation
 
 on:
-  pull_request:
+  push:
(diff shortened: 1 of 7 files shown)


contrib/aws-efa/README.md
```diff
@@ -0,0 +1,76 @@
+# NIXL AWS Testing
```
yosefe, line 1: I would put aws-efa folder under github workflows

contrib/aws-efa/aws_test.sh
```diff
@@ -0,0 +1,107 @@
...
+        ;;
+esac
+
+# Construct command to run in AWS
+setup_cmd="set -x && \
+    git clone ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY} && \
+    cd nixl && \
+    ${GIT_CHECKOUT_CMD}"
+build_cmd=".gitlab/build.sh \${NIXL_INSTALL_DIR} \${UCX_INSTALL_DIR}"
+test_script="test_${TEST_TYPE}"
```
yosefe: maybe we can pass the test script path as argument? so we can run as many scripts we want

contrib/aws-efa/aws_test.sh
```diff
@@ -0,0 +1,107 @@
...
+export AWS_CMD="${setup_cmd} && ${build_cmd} && .gitlab/${test_script}.sh \${NIXL_INSTALL_DIR}"
```
dpressle: This should be a script by itself that gets parameters and the batch cmd will just run this script

contrib/aws-efa/README.md
dpressle: I thought we agreed to have our logic in contrib?
Its also not best practice to put build scripts inside .github folder
You should not put business logic in a build system place what if tomorrow we will move to a new build system we will have to move all these scripts and update the logic, it must be seperated.

tstamler:
Hey, this is pretty cool! Just a couple comments:

-We have nixlbench in the main repo now, it would be nice to see it integrated in some configuration (doesn't need to be right now)
-Why are you deleting the pre-commit yaml? Current CI is obviously not going to be happy about that, but maybe you have a plan

Alexey-Rivkin:
> Hey, this is pretty cool! Just a couple comments:
> 
> -We have nixlbench in the main repo now, it would be nice to see it integrated in some configuration (doesn't need to be right now) -Why are you deleting the pre-commit yaml? Current CI is obviously not going to be happy about that, but maybe you have a plan

Hi, thanks for the feedback!
This is still a work in progress - please don't judge (yet). 
`.pre-commit-config.yaml` conflicted with my local one, but I've since realized the mistake and fetched the original. 
As for nixlbench, let's have it in a separate PR since this one is already pretty big.

contrib/aws-efa/README.md
yosefe: ok

contrib/aws-efa/aws_test.sh
Alexey-Rivkin: Done. Now we can add multiple scripts per test suite.
https://github.com/ai-dynamo/nixl/pull/212/files#diff-0d0db0974df99ada87d9b3aa237c7bc6aebd8ebeb696841cc362e73cf78f4783R24

.github/workflows/aws_efa_validation.yml
```diff
@@ -0,0 +1,66 @@
...
+              - .gitlab/test_python.sh $NIXL_INSTALL_DIR
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Verify AWS credentials
+        run: aws sts get-caller-identity >/dev/null
```
dpressle: How can you run this before you install aws cli in the Install tools step (even if it works its confusing and redundant) ?
Can we make it simpler and combine these 2 steps into one and call it preparation or something

dpressle (changes requested):

.github/workflows/aws_efa_validation.yml
Alexey-Rivkin: Good point! Reordered and combined into a single step.

dpressle (approved):

yosefe (approved):

.github/workflows/aws_efa_validation.yml
```diff
@@ -0,0 +1,66 @@
+name: NVIDIA NIXL AWS EFA Validation
+
+on:
+  pull_request:
```
dpressle: Does this works for both branches and forks?

.github/workflows/aws_efa_validation.yml
```diff
@@ -0,0 +1,66 @@
+name: NVIDIA NIXL AWS EFA Validation
```
dpressle: I think NVIDIA in the name means this workflow runs on Nvidia lab i think we should call it:
AWS EFA NIXL Validation

dpressle (changes requested):

.github/workflows/aws_efa_validation.yml
Alexey-Rivkin: 70be865429fcb69512ebd1f745087963c943359b

.github/workflows/aws_efa_validation.yml
Alexey-Rivkin: 70be865429fcb69512ebd1f745087963c943359b

dpressle (approved):

.github/workflows/aws_efa_validation.yml
```diff
@@ -0,0 +1,69 @@
+name: AWS EFA NIXL Validation
+
+on:
+  push:
+    branches:
+    - main
+    - "pull-request/[0-9]+"
```
pvijayakrish, line 7: This change looks good and should let the PRs from external contributors to run the changes on AWS batch once an internal reviewer gives the "ok to test" and the copy-pr-bot copies them over.

.github/workflows/aws_efa_validation.yml
```diff
@@ -0,0 +1,69 @@
...
+          # Verify AWS credentials
+          aws sts get-caller-identity >/dev/null
+
+      - name: Run AWS tests
+        working-directory: ./contrib/aws-efa
+        timeout-minutes: 180
+        run: |
+          set -o pipefail
+          test_cmd='${{ join(matrix.test_scripts, ' && ') }}'
+          ./aws_test.sh "$test_cmd" 2>&1 | tee test_output.log
```
pvijayakrish, line 59: Could you please share how the results of this job are reported? Additionally, could you provide the link to the AWS Batch dashboard where these tests were executed?

pvijayakrish: How to get read/view permission to the dashboard?

Alexey-Rivkin: Pass/fail status propagation:
When the pod exits, we poll AWS for exit status. If the final state is FAILED, the script exits non-zero, so Github Actions marks the flow as failed.
On success we exit 0, which lets the workflow turn green.
https://github.com/ai-dynamo/nixl/blob/arivkin/aws_tests/contrib/aws-efa/aws_test.sh#L120-L125

Alexey-Rivkin: I have no access to the AWS dashboard.

.github/workflows/aws_efa_validation.yml
Alexey-Rivkin: Standing on the shoulders of giants - copied the trigger part from the existing flow.

Alexey-Rivkin:
> PR looks good. Could you please share how the changes are tested and how the results from AWS batch will be reported back?

### Results propagation from AWS
When the pod exits, we poll AWS for exit status. If the final state is FAILED, the script exits non-zero, so Github Actions marks the flow as failed.
On success we exit 0, which lets the workflow turn green.
https://github.com/ai-dynamo/nixl/blob/arivkin/aws_tests/contrib/aws-efa/aws_test.sh#L120-L125

### Testing done
The changes are validated through the AWS EFA workflow, which provides end-to-end testing by executing NIXL tests on AWS infrastructure and reporting the results back to GitHub. Negative tests were also conducted to ensure robustness.

.github/workflows/aws_efa_validation.yml
pvijayakrish: Was the tests run on EFA? Please help understand where this was tested

Alexey-Rivkin: All tests run on AWS, which exclusively implements EFA instead of traditional InfiniBand

pvijayakrish: Approving to unblock the team. Without the dashboard access, please note that the debugging test failures can get challenging. Because, we are only polling for the exit status.

pvijayakrish (approved):

PR #213, opened 2025-04-28
UTILS/BUILD: Logging infra cleanups

yosefe:
## Why
Simplify build scripts. absl::log provides DVLOG macro that is stripped in compile time with NDEBUG. So just need to define NDEBUG for release builds.
We want to keep most of the logs in release build, and remove only the ones in data path - NIXL_TRACE.

### DIFF
--- a/src/utils/common/nixl_log.h
+++ b/src/utils/common/nixl_log.h
@@ -49,9 +49,8 @@
 
 /*
  * Logs messages unconditionally (maps to Abseil verbosity level 1)
- * Stripped from release buids.
  */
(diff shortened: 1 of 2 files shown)


src/utils/common/nixl_log.h
```diff
@@ -49,9 +49,8 @@
 
 /*
  * Logs messages unconditionally (maps to Abseil verbosity level 1)
- * Stripped from release buids.
  */
-#define NIXL_DEBUG DVLOG(1)
+#define NIXL_DEBUG VLOG(1)
```
yosefe, line 53: Don't strip DEBUG

PR #216, opened 2025-04-28
fix: set version to 0.2.0

saturley-hall:


pyproject.toml
```diff
@@ -19,7 +19,7 @@ build-backend = "mesonpy"
```
nv-anants, line 19: one more reference - https://github.com/ai-dynamo/nixl/blob/main/meson.build#L16

nv-anants (approved):

PR #218, opened 2025-04-28
API: add option to use ETCD for metadata exchange to release branch

tstamler:


saturley-hall (approved):
Approved as from a cherry-pick of this approved on main

PR #223, opened 2025-04-30
refactor: Prefer "using" over "typedef" guideline implementation

etoledano:
Apply `Prefer "using" over "typedef"` guidelines in src/ This commit doesn't change functionality, it is strictly visual.

Copy of https://github.com/ai-dynamo/nixl/pull/208

### DIFF
--- a/src/utils/common/nixl_time.h
+++ b/src/utils/common/nixl_time.h
@@ -23,10 +23,10 @@ namespace nixlTime {
 
     using namespace std::chrono;
 
-    typedef uint64_t ns_t;
-    typedef uint64_t us_t;
-    typedef uint64_t ms_t;
-    typedef uint64_t sec_t;
+    using ns_t = uint64_t;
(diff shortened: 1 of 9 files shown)


src/utils/common/nixl_time.h
```diff
@@ -23,10 +23,10 @@ namespace nixlTime {
 
     using namespace std::chrono;
 
-    typedef uint64_t ns_t;
```
yosefe, line 26: i would consider removing this wrapper and use absl::Now()

yosefe (approved):

src/utils/common/nixl_time.h
etoledano: Sounds good, I've added it to a list for the refactors to do (since this PR's subject is very specific), OK?

etoledano: @yosefe Resolving to promote merging

PR #226, opened 2025-04-30
API: New API proposals

yosefe:


src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+                std::vector<struct obj::Agent> accessList;
+            };
+
+            struct DelMemSpace {
+                struct obj::MemSpace memSpace;
+            };
+
+            struct Msg {
+                Opcode                 opcode;
+                union {
```
w1ldptr, line 88: Unions in C++ with non-POD typed members have terrible ergonomics. Just try to make this simple example work: https://www.programiz.com/online-compiler/5TjIfXfqAW48k

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+              const bool                                podClientAutoPoll,
+              const RcvSignalCb                         rcvSignalCb,
+              const bool                                autoPoll)
+              : podClient(podClientServerID, podClientRcvMsgCb, podClientAutoPoll) {}
+
+        ~Agent();
+
+        void postMemXfer(class MemXfer &memXfer);
+
+        void getCompletedMemXfers(std::vector<class MemXfer> &memXfers);
```
w1ldptr, line 220: Passing references to containers is error prone since it is unclear whether the caller is responsible to clear the container before calling the function again or the callee should do that, for example. Return the container instance instead for cleaner, more maintainable interface.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+
+        void postSndSignal(const class Signal &signal);
+
+        void getCompletedSndSignals(std::vector<class Signal> &signals);
```
w1ldptr, line 224: ditto

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+                         const std::vector<int> &dst_indices,
+                         const std::vector<int> &src_indices);
+
+    ///////////////////////////////////////////////////////////////////
+    // Completion checking
+
+    // Poll Option1: check if a specific transfer is done
+    // TODO: allow returning failures
+    // TODO move it to XferRequest class?
+    bool isXferDone(const XferRequest &xfer_request);
```
w1ldptr, line 102: Rename to isXferCompleted() to consistency with the following method.

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+    ///////////////////////////////////////////////////////////////////
+    // Signals / Notifications
+
+    // Post a signal to a remote agent
+    // Signal is ordered w.r.t. to any previous transfers to the same remote
+    // agent.
+    void postMessage(std::string_view agent_name, const std::string &message);
+
+    // Option1: poll for all signals
+    void pollMessages(std::multimap<std::string_view, std::string> &signals);
```
w1ldptr, line 116: Who owns the string that string_view points to in this case? The agent instance? How it the lifetime of the target is accounted for considering that caller can retain the resulting multimap indefinitely. 

PS also, return the container instead of modifying by reference.

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+            std::string_view agent_name);
+};
+
+class Agent {
+public:
+    class XferList;
+    class XferRequest;
+    class OptionalParams;
+
+    Agent(std::string_view name, const Config &cfg);
```
w1ldptr, line 48: Wouldn't an agent instance need to save its name somewhere? If yes, why this is not a const string &?

w1ldptr (commented):
On a general note I wonder whether we would also need some kind of notification-based async data path API instead or in addition to polling based. This would allow better async bindings for other languages, especially Rust IMO.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+        } // namespace obj
+
+        namespace msg {
+            enum class Opcode {
+                ADD_AGENT,
+                DEL_MEMSPACE
+            };
+
+            struct AddAgent {
+                struct obj::Agent agent;
```
rakhmets, line 69: `struct` keyword can be removed here and below.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+            struct MemXferBuf {
+                class mem::Space space;
+                size_t           ofst;
+                size_t           len;
+            };
+
+            PostState            postState;
+            PostCompletionStatus postCompletionStatus;
+
+            // dst and src must have the same length
```
rakhmets, line 180: If these two collections must have the same size, then it should be a single collection containing both src and dst.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+            POSTED,
+            COMPLETED
+        };
+
+        enum class PostCompletionStatus {
+            OK,
+            ERROR
+        };
+
+        class MemXfer {
```
rakhmets, line 169: I think this should be an abstract class, containing only pure virtual functions.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+                class mem::Space space;
+                size_t           ofst;
+                size_t           len;
+            };
+
+            PostState            postState;
+            PostCompletionStatus postCompletionStatus;
+
+            // dst and src must have the same length
+            MemXfer(const std::vector<struct MemXferBuf> &dst,
```
rakhmets, line 181: `std::vector` cannot be used here, if cross-compiler ABI support is a requirement for the library.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+            };
+
+            struct MemSpace {
+                struct Agent agent;
+                uint32_t     key;
+            };
+        } // namespace obj
+
+        namespace msg {
+            enum class Opcode {
```
ColinNV, line 63: This could be `enum class Opcode : bool` to save space. Similar for other enums (with `bool` or `std::uint8_t` as appropriate).

src/api/cpp/nixl_v2_1.h
ColinNV: This could use `std::variant<>` to make it both safer and easier to use (with IMO negligible overhead).

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+            };
+        } // namespace msg
+
+        class Client {
+            public:
+                using RcvMsgCb = std::function<void(const struct Msg &msg)>;
+
+                Client(const struct AgentID &serverID,
+                       const RcvMsgCb       &rcvMsgCb,
+                       const bool           autoPollRcvMsg);
```
ColinNV, line 103: The `bool` argument could be replaced by an enum for better readability at the call sites.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+
+        class Space {
+        public:
+            struct Local {
+                Type     type;
+                int      dev_id;
+                uint64_t len;
+            };
+
+            Space(const struct Local &local);
```
ColinNV, line 130: Is it intentional that this constructor is not marked `explicit`? Same for the one below.

src/api/cpp/nixl_v2_1.h
```diff
@@ -0,0 +1,232 @@
...
+            Space(const struct pod::obj::MemSpace &podMemSpace);
+
+            // generate a key for granting access to a local mem space for remote agents
+            uint32_t genPodMemSpaceKeyForLocalMemSpace();
```
ColinNV, line 134: This, and several other functions, should probably be declared with `[[nodiscard]]`?

src/api/cpp/nixl_v2_1.h
ColinNV: If the function does not clear the out-parameter it allows the caller to accumulate the results from multiple calls in a single container. For convenience one could add a second function with the same name and no parameter that wraps this function and returns a vector.

src/api/cpp/nixl_v2_1.h
ColinNV: And offering only a function that takes a `std::vector` is not ideal when the caller has only a single object (that needs to be put in a vector just for the call) or, say, a `std::array`. Usually in this case I'd make a private function that takes `const T*` and a `std::size_t` and then have inline forwarders for the various types I want to support.

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+
+class Config {
+};
+
+class MemDesc {
+public:
+    enum class MemType { DRAM, VRAM, BLK, OBJ, FILE, PODMEM };
+
+    // Local memory
+    MemDesc(uint64_t address, size_t length, MemType type, uint64_t dev_id);
```
ColinNV, line 35: The API could use dedicated `enum class` types for `dev_id` (and possibly `address`) to catch mixing up the arguments at compile time. Might also apply in other places.

src/api/cpp/nixl_v2_2.h
ColinNV: Assuming your question is "why is `name` not a `const std::string&`", the short answer is because initializing a `std::string m_name` member of `class Agent` from a `std::string_view` is more efficient in some cases, think passing a string literal; the `std::string_view` prevents a temporary `std::string` from being constructed for the argument (or us hoping that the compiler is smart enough, and allowed to, get rid of it.)

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+    // Posting list of operations
+
+    // Post all operations in the list
+    // There is no ordering guarantee between different transfer requests or
+    // between operations in the same transfer request.
+    XferRequest postXfer(const XferList &xfer_list);
+
+    // Post operations in the list with specific dst and src indices
+    XferRequest postXfer(const XferList &xfer_list,
+                         const std::vector<int> &dst_indices,
```
ColinNV, line 93: Why are the indices `int` and not `std::size_t` (or realistically `std::uint32_t`)?

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+    // Poll Option2: get all completed transfers
+    void getCompletedXfers(std::vector<XferRequest> &xfer_requests);
+
+    ///////////////////////////////////////////////////////////////////
+    // Signals / Notifications
+
+    // Post a signal to a remote agent
+    // Signal is ordered w.r.t. to any previous transfers to the same remote
+    // agent.
+    void postMessage(std::string_view agent_name, const std::string &message);
```
ColinNV, line 113: Why is `message` not also a `std::string_view`?

src/api/cpp/nixl_v2_2.h
ColinNV: And does the result always need to be sorted, or would it make sense to offer a second, slightly more efficient, function that returns a `std::vector< std::pair< std::string, std::string > >` (with no guarantee on the ordering)?

src/api/cpp/nixl_v2_1.h
yosefe: We can clarify that the function does NOT clear the container. Returning a container forces a memory allocation.

src/api/cpp/nixl_v2_2.h
yosefe: Yes, the agent owns the string view (since the first element is the name of the remote agent, and wanted to avoid extra string copy)

src/api/cpp/nixl_v2_2.h
yosefe: unsigned?

src/api/cpp/nixl_v2_2.h
yosefe: right, it can be string_view

src/api/cpp/nixl_v2_1.h
w1ldptr: Returning a vector _does not_ force a memory allocation. The resulting vector will be move-constructed by just copying the pointer to data from the source vector instead of deep-copying the contents.

ColinNV: The vector itself will need to allocate memory even when the return value is moved into place; passing a mutable reference allows the (re)use of a pre-allocated (think `reserve()`) vector.

w1ldptr: The API itself implies quite "heavy" implementation with all the std containers it uses and the amount of features it present to the user. IMO premature micro-optimizations should take lower priority to the simplicity and robustness of the API in a high level C++ application that dynamically allocates memory on a whim (or, more likely, a Python or Rust app communicating with the lib over another layer of bindings on top of it) as opposed to, say, DPDK or kernel API.

Just my 02 cents 🤷‍♂️

src/api/cpp/nixl_v2_2.h
w1ldptr: So if, for example, when agent->invalidateRemoteMD() (using a method from existing API to simplify) is called concurrently user could suddenly end up with dangling pointer to a deallocated string? Just to save a string copy... when already constructing a std::multimap instance?

src/api/cpp/nixl_v2_2.h
w1ldptr: I mean, one could argue that in cases where extra copying is a concern it would be better to just make the argument std::string so that it would be copied once during the call to the constructor and then used to move-construct m_name instead of always having intermediate string_view object. But this gets rather theoretical considering we are talking about a class that would realistically get instantiated once during the application startup so I'm not going to argue about it further.

src/api/cpp/nixl_v2_1.h
w1ldptr: Agree, std::variant would be much more ergonomic to  use.

src/api/cpp/nixl_v2_2.h
ColinNV: ...or `unsigned`, sure.

src/api/cpp/nixl_v2_2.h
ColinNV: Sure, the `Agent( std::string name ) : m_name( std::move( name ) )` pattern works too.

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+    // Prepare transfer list
+
+    // Prepare local and remote buffers for transfer:
+    // - Register local buffers
+    // - Potentially calculate the communication pattern and breakdown to
+    //   collective operations
+    // - Potentially send a message to the remote agent(s) asking to resolve the
+    //   remote addresses to memory access keys, or use one-sided RDMA_READ
+    //   to resolve remote keys.
+    // - List sizes and corresponding buffer sizes are the same.
```
rakhmets, line 72: This should be reflected in the API by replacing the two collections with a single collection of pairs (or structures with two fields).

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+
+    // Post all operations in the list
+    // There is no ordering guarantee between different transfer requests or
+    // between operations in the same transfer request.
+    XferRequest postXfer(const XferList &xfer_list);
+
+    // Post operations in the list with specific dst and src indices
+    XferRequest postXfer(const XferList &xfer_list,
+                         const std::vector<int> &dst_indices,
+                         const std::vector<int> &src_indices);
```
rakhmets, line 94: If these two vectors must have the same size, then the argument should be a vector of pairs/structures with two fields.

src/api/cpp/nixl_v2_2.h
```diff
@@ -0,0 +1,125 @@
...
+
+    // Local memory
+    MemDesc(uint64_t address, size_t length, MemType type, uint64_t dev_id);
+
+    // Remote memory
+    MemDesc(uint64_t address, size_t length, MemType type, uint64_t dev_id,
+            std::string_view agent_name);
+};
+
+class Agent {
```
rakhmets, line 42: I recommend using an abstract class as an interface. E.g.,
```C++
class IAgent {
public:
    using Ptr = std::unique_ptr<IAgent>;

    virtual ~IAgent() = default;
    virtual XferRequest postXfer(const XferList &xferList) = 0;
    virtual XferRequest postXfer(const XferList &xfer_list, const XferIndices &xferIndices) = 0;
    // etc.
};
```

And separate creation of the object from the interface.
```C++
IAgent::Ptr createAgent(const Config &config);
```

ColinNV: > ```c++
> class IAgent {
> public:
>     using Ptr = std::unique_ptr<IAgent>;
> };
> ```

I'd suggest removing the `Ptr` type alias; reading `IAgent::Ptr` in other places doesn't convey as much as reading `std::unique_ptr< IAgent >`.

And for its role `IAgent` should be made "noncopyable" by adding:
```
   IAgent( IAgent&& ) = delete;
   IAgent( const IAgent& ) = delete;
   void operator=( IAgent&& ) = delete;
   void operator=( const IAgent& ) = delete;
```

rakhmets: agree, when I wrote the comment, I used `std::unique_ptr<IAgent>` several times, so decided to add an alias. But now it is redundant.

PR #228, opened 2025-04-30
tstamler: fix cuda deps

tstamler:


src/core/meson.build
```diff
@@ -32,7 +32,7 @@ endif
 
 disable_gds_backend = get_option('disable_gds_backend')
 if not disable_gds_backend and cuda_dep.found()
```
aranadive: Do you still need the cuda_dep check here? Or maybe rename it to gds_deps to make it clearer?

vvenkates27: gds gets installed with cudatools.

aranadive (approved):

PR #231, opened 2025-05-01
NIXLBench: Use directory listing and generation for barrier

tvegas1:
Proposal to use etcd `client->ls()` for barrier implementation and try to reduce sleep. Also currently we perform put()/get() which are not atomic as per my understanding.

### DIFF
--- a/benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
+++ b/benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
@@ -376,22 +376,21 @@ int xferBenchEtcdRT::reduceSumDouble(double *local_value, double *global_value, 
 }
 
 int xferBenchEtcdRT::barrier(const std::string& barrier_id) {
+    int count = 0;
+    std::string barrier_suffix = "barrier/" + barrier_id + "/" + std::to_string(barrier_gen);
+    std::string barrier_key = namespace_prefix + barrier_suffix + "/";
+    barrier_gen++; // In case same barrier is reused too quickly
+
     try {
-        int count = -1;
-        std::string barrier_suffix = "barrier/" + barrier_id + "/" + std::to_string(barrier_gen);
-        std::string barrier_key = namespace_prefix + barrier_suffix + "/";
-
-        barrier_gen++; // In case same barrier is reused too quickly
-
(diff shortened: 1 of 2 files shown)


benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
```diff
@@ -377,98 +377,45 @@ int xferBenchEtcdRT::reduceSumDouble(double *local_value, double *global_value,
...
-        int retries = 0;
-        int expected_count = global_size;
-
-        while (!barrier_complete && retries < 30) { // 5 minutes timeout (300 seconds)
-            resp = client->get(count_key).get();
-            if (resp.error_code() == 0) {
-                current_count = std::stoi(resp.value().as_string());
+        int count = -1;
+        std::string barrier_suffix = "barrier/" + barrier_id + "/" + std::to_string(barrier_gen);
+        std::string barrier_key = namespace_prefix + barrier_suffix + "/";
```
aranadive: Remove the / here at the end since you add one below as well.

aranadive (approved):

benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
```diff
@@ -377,98 +377,45 @@ int xferBenchEtcdRT::reduceSumDouble(double *local_value, double *global_value,
...
-            resp = client->get(ready_key).get();
-            if (resp.error_code() == 0 && resp.value().as_string() == "true") {
-                ready = true;
-            } else {
-                std::this_thread::sleep_for(std::chrono::seconds(1));
-                retries++;
+        for (auto retries = 0; retries < 60; ++retries) {
+            auto resp = client->ls(barrier_key).get();
+            if (resp.error_code()) {
+                goto fail;
```
brminich: can you use exception instead?

tvegas1: fixed

benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
tvegas1: removed below

brminich (approved):

sb17v (approved):

PR #235, opened 2025-05-02
POSIX: Add more logs

tvegas1:
Some nodes might not support io_uring. Add more related details to CI and code.

### DIFF
--- a/src/plugins/posix/posix_backend.h
+++ b/src/plugins/posix/posix_backend.h
@@ -22,7 +22,6 @@
 #include <string>
 #include <vector>
 #include <liburing.h>
-#include <stdexcept>
 #include "backend/backend_engine.h"
 
 class uringQueue {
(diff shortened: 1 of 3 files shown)


tvegas1:
@tstamler, wondering if some CI nodes might be missing io_uring support as some related failures were seen?

src/plugins/posix/posix_backend.h
```diff
@@ -43,8 +44,9 @@ class uringQueue {
         nixl_status_t checkCompleted();
         struct io_uring_sqe *getSqe();
 
-        enum class UringError {
-            INIT,
+        class UringError : public std::runtime_error {
+            public:
+                using std::runtime_error::runtime_error;
```
etoledano: Can we make it simpler using `enum class` like `OperationError`?

tvegas1: i'm not sure if can carry errno/msg within enum class, could you please provide code suggestion?

tstamler:
> @tstamler, wondering if some CI nodes might be missing io_uring support as some related failures were seen?

Yes, I think there is a problem with some Gitlab workers not exposing io_uring system calls correctly. I need to follow up with the Ops team that setup the Gitlab CI but this will also hopefully be fixed when moving to purely Github workers for CI.

src/plugins/posix/posix_backend.h
tvegas1: removed and adapted to output details of the failure

PR #236, opened 2025-05-02
UCX backend error handling

evgeny-leksikov:
- Added naive remote failure handling to UCX backend.
 - Add gtest `UcxTestFixture`
 - Related code refactoring

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -43,18 +43,19 @@ std::string nixlEnumStrings::xferOpStr (const nixl_xfer_op_t &op) {
 
 std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
     switch (status) {
-        case NIXL_IN_PROG:           return "NIXL_IN_PROG";
-        case NIXL_SUCCESS:           return "NIXL_SUCCESS";
-        case NIXL_ERR_NOT_POSTED:    return "NIXL_ERR_NOT_POSTED";
-        case NIXL_ERR_INVALID_PARAM: return "NIXL_ERR_INVALID_PARAM";
-        case NIXL_ERR_BACKEND:       return "NIXL_ERR_BACKEND";
-        case NIXL_ERR_NOT_FOUND:     return "NIXL_ERR_NOT_FOUND";
-        case NIXL_ERR_MISMATCH:      return "NIXL_ERR_MISMATCH";
-        case NIXL_ERR_NOT_ALLOWED:   return "NIXL_ERR_NOT_ALLOWED";
-        case NIXL_ERR_REPOST_ACTIVE: return "NIXL_ERR_REPOST_ACTIVE";
-        case NIXL_ERR_UNKNOWN:       return "NIXL_ERR_UNKNOWN";
-        case NIXL_ERR_NOT_SUPPORTED: return "NIXL_ERR_NOT_SUPPORTED";
-        default:                     return "BAD_STATUS";
+        case NIXL_IN_PROG:               return "NIXL_IN_PROG";
+        case NIXL_SUCCESS:               return "NIXL_SUCCESS";
+        case NIXL_ERR_NOT_POSTED:        return "NIXL_ERR_NOT_POSTED";
+        case NIXL_ERR_INVALID_PARAM:     return "NIXL_ERR_INVALID_PARAM";
+        case NIXL_ERR_BACKEND:           return "NIXL_ERR_BACKEND";
+        case NIXL_ERR_NOT_FOUND:         return "NIXL_ERR_NOT_FOUND";
+        case NIXL_ERR_MISMATCH:          return "NIXL_ERR_MISMATCH";
+        case NIXL_ERR_NOT_ALLOWED:       return "NIXL_ERR_NOT_ALLOWED";
+        case NIXL_ERR_REPOST_ACTIVE:     return "NIXL_ERR_REPOST_ACTIVE";
+        case NIXL_ERR_UNKNOWN:           return "NIXL_ERR_UNKNOWN";
+        case NIXL_ERR_NOT_SUPPORTED:     return "NIXL_ERR_NOT_SUPPORTED";
+        case NIXL_ERR_REMOTE_DISCONNECT: return "NIXL_ERR_REMOTE_DISCONNECT";
+        default:                         return "BAD_STATUS";
     }
 }
 
@@ -824,7 +825,7 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) {
 
     NIXL_LOCK_GUARD(data->lock);
     // If the status is done, no need to recheck.
-    if (req_hndl->status != NIXL_SUCCESS) {
+    if (req_hndl->status == NIXL_IN_PROG) {
         // Check if the remote was invalidated before completion
         if (data->remoteSections.count(req_hndl->remoteAgent) == 0) {
             delete req_hndl;
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -277,7 +277,8 @@ void nixlPluginManager::discoverPluginsFromDir(const std::string& dirpath) {
     std::error_code ec;
     std::filesystem::directory_iterator dir_iter(dir_path, ec);
     if (ec) {
-        std::cerr << "Error accessing directory: " << ec.message() << std::endl;
+        std::cerr << "Error accessing directory(" << dir_path << "):"
+                  << ec.message() << std::endl;
         return;
     }
 
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -494,7 +494,7 @@ nixl_status_t nixlUcxEngine::endConn(const std::string &remote_agent) {
 
     nixlUcxConnection &conn = remoteConnMap[remote_agent];
 
-    if(uw->disconnect_nb(conn.ep) < 0) {
+    if(uw->disconnect_nb(conn.getEp()) < 0) {
         return NIXL_ERR_BACKEND;
     }
 
@@ -592,7 +592,7 @@ nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
     //agent names should never be long enough to need RNDV
     flags |= UCP_AM_SEND_FLAG_EAGER;
 
-    ret = uw->sendAm(conn.ep, CONN_CHECK,
+    ret = uw->sendAm(conn.getEp(), CONN_CHECK,
                      &hdr, sizeof(struct nixl_ucx_am_hdr),
                      (void*) localAgent.data(), localAgent.size(),
                      flags, req);
@@ -625,18 +625,20 @@ nixl_status_t nixlUcxEngine::disconnect(const std::string &remote_agent) {
 
         nixlUcxConnection &conn = remoteConnMap[remote_agent];
 
-        hdr.op = DISCONNECT;
-        //agent names should never be long enough to need RNDV
-        flags |= UCP_AM_SEND_FLAG_EAGER;
+        if (conn.getEp().checkTxState() == NIXL_SUCCESS) {
+            hdr.op = DISCONNECT;
+            //agent names should never be long enough to need RNDV
+            flags |= UCP_AM_SEND_FLAG_EAGER;
 
-        ret = uw->sendAm(conn.ep, DISCONNECT,
-                        &hdr, sizeof(struct nixl_ucx_am_hdr),
-                        (void*) localAgent.data(), localAgent.size(),
-                        flags, req);
+            ret = uw->sendAm(conn.getEp(), DISCONNECT,
+                             &hdr, sizeof(struct nixl_ucx_am_hdr),
+                             (void*) localAgent.data(), localAgent.size(),
+                             flags, req);
 
-        //don't care
-        if(ret == NIXL_IN_PROG){
-            uw->reqRelease(req);
+            //don't care
+            if (ret == NIXL_IN_PROG) {
+                uw->reqRelease(req);
+            }
         }
     }
 
@@ -659,14 +661,13 @@ nixl_status_t nixlUcxEngine::loadRemoteConnInfo (const std::string &remote_agent
     }
 
     nixlSerDes::_stringToBytes((void*) addr, remote_conn_info, size);
-    ret = uw->connect(addr, size, conn.ep);
+    ret = uw->connect(addr, size, conn.getEp());
     if (ret) {
         return NIXL_ERR_BACKEND;
     }
 
     conn.remoteAgent = remote_agent;
-    conn.connected = false;
-
+    // TODO: should we use move semantics here?
     remoteConnMap[remote_agent] = conn;
 
     delete[] addr;
@@ -753,7 +754,7 @@ nixlUcxEngine::internalMDHelper (const nixl_blob_t &blob,
     char *addr = new char[size];
     nixlSerDes::_stringToBytes(addr, blob, size);
 
-    int ret = uw->rkeyImport(conn.ep, addr, size, md->rkey);
+    int ret = uw->rkeyImport(conn.getEp(), addr, size, md->rkey);
     if (ret) {
         // TODO: error out. Should we indicate which desc failed or unroll everything prior
         return NIXL_ERR_BACKEND;
@@ -865,10 +866,10 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
 
         switch (operation) {
         case NIXL_READ:
-            ret = uw->read(rmd->conn.ep, (uint64_t) raddr, rmd->rkey, laddr, lmd->mem, lsize, req);
+            ret = uw->read(rmd->conn.getEp(), (uint64_t) raddr, rmd->rkey, laddr, lmd->mem, lsize, req);
             break;
         case NIXL_WRITE:
-            ret = uw->write(rmd->conn.ep, laddr, lmd->mem, (uint64_t) raddr, rmd->rkey, lsize, req);
+            ret = uw->write(rmd->conn.getEp(), laddr, lmd->mem, (uint64_t) raddr, rmd->rkey, lsize, req);
             break;
         default:
             return NIXL_ERR_INVALID_PARAM;
@@ -880,7 +881,7 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
     }
 
     rmd = (nixlUcxPublicMetadata*) remote[0].metadataP;
-    ret = uw->flushEp(rmd->conn.ep, req);
+    ret = uw->flushEp(rmd->conn.getEp(), req);
     if (_retHelper(ret, intHandle, req)) {
         return ret;
     }
@@ -951,7 +952,7 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
     // TODO: replace with mpool for performance
     ser_msg = new std::string(ser_des.exportStr());
 
-    ret = uw->sendAm(conn.ep, NOTIF_STR,
+    ret = uw->sendAm(conn.getEp(), NOTIF_STR,
                      &hdr, sizeof(struct nixl_ucx_am_hdr),
                      (void*) ser_msg->data(), ser_msg->size(),
                      flags, req);
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -41,10 +41,13 @@ struct nixl_ucx_am_hdr {
 class nixlUcxConnection : public nixlBackendConnMD {
     private:
         std::string remoteAgent;
-        nixlUcxEp ep;
-        volatile bool connected;
+        // NOTE: nixlUcxEp object must be alive as long as ucp_ep_h is alive
+        //       since the object is passed to ucp_ep_create_nbx as err_cb
+        //       argument.
+        std::shared_ptr<nixlUcxEp> ep = std::make_shared<nixlUcxEp>();
 
     public:
+        nixlUcxEp& getEp() { return *ep; }
         // Extra information required for UCX connections
 
     friend class nixlUcxEngine;
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -14,15 +14,149 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <nixl_types.h>
 #include <vector>
 #include <string>
 #include <cstring>
-#include <cassert>
 
 #include "ucx_utils.h"
+#include "common/nixl_log.h"
 
 using namespace std;
 
+static nixl_status_t ucx_status_to_nixl(ucs_status_t status)
+{
+    if (status == UCS_OK) [[likely]] {
+        return NIXL_SUCCESS;
+    }
+
+    switch(status) {
+    case UCS_INPROGRESS:
+        return NIXL_IN_PROG;
+    case UCS_ERR_CONNECTION_RESET:
+        return NIXL_ERR_REMOTE_DISCONNECT;
+    case UCS_ERR_INVALID_PARAM:
+        return NIXL_ERR_INVALID_PARAM;
+    default:
+        return NIXL_ERR_BACKEND;
+    }
+}
+
+void nixlUcxEp::err_cb(void *arg, ucp_ep_h ucp_ep, ucs_status_t status)
+{
+    nixlUcxEp *ep = reinterpret_cast<nixlUcxEp*>(arg);
+    ucs_status_ptr_t request;
+
+    NIXL_DEBUG << "ep " << ep->getHandle() << ": state " << ep->state
+               << ", UCX error handling callback was invoked with status "
+               << status << " (" << ucs_status_string(status) << ")";
+
+    NIXL_ASSERT(ep->getHandle() == ucp_ep);
+
+    switch(ep->state) {
+    case NIXL_UCX_EP_STATE_NULL:
+    case NIXL_UCX_EP_STATE_FAILED:
+        // The error was already handled, nothing to do
+    case NIXL_UCX_EP_STATE_DISCONNECTED:
+        // The EP has been disconnected, nothing to do
+        return;
+    case NIXL_UCX_EP_STATE_CONNECTED:
+        ep->setState(NIXL_UCX_EP_STATE_FAILED);
+        request = ucp_ep_close_nb(ucp_ep, UCP_EP_CLOSE_MODE_FORCE);
+        if (UCS_PTR_IS_PTR(request)) {
+            ucp_request_free(request);
+        }
+        return;
+    default:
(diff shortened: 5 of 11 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -181,11 +240,35 @@ int nixlUcxWorker::epAddr(uint64_t &addr, size_t &size)
...
-
-static void err_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
+static void err_cb(void *arg, ucp_ep_h ucp_ep, ucs_status_t status)
 {
-    printf("error handling callback was invoked with status %d (%s)\n",
-           status, ucs_status_string(status));
+    nixlUcxEp *ep = (nixlUcxEp*)arg;
+    ucs_status_ptr_t request;
+
+    printf("ep %p: state %d, error handling callback was invoked with status %d (%s)\n",
```
brminich: better to replace with absl based NIXL logging

src/utils/ucx/ucx_utils.cpp
```diff
@@ -181,11 +240,35 @@ int nixlUcxWorker::epAddr(uint64_t &addr, size_t &size)
...
+        // The EP has been disconnected, nothing to do
+        return;
+    case NIXL_UCX_EP_STATE_CONNECTED:
+        request = ucp_ep_close_nb(ucp_ep, UCP_EP_CLOSE_MODE_FORCE);
+        if (UCS_PTR_IS_PTR(request)) {
+            ucp_request_free(request);
+        }
+        break;
+    default:
+        assert(0 && "Invalid endpoint state");
```
brminich: better to use `NIXL_ASSERT_ALWAYS`

src/utils/ucx/ucx_utils.cpp
```diff
@@ -181,11 +240,35 @@ int nixlUcxWorker::epAddr(uint64_t &addr, size_t &size)
...
 {
-    printf("error handling callback was invoked with status %d (%s)\n",
-           status, ucs_status_string(status));
+    nixlUcxEp *ep = (nixlUcxEp*)arg;
+    ucs_status_ptr_t request;
+
+    printf("ep %p: state %d, error handling callback was invoked with status %d (%s)\n",
+           ep->getHandle(), ep->getState(), status, ucs_status_string(status));
+
+    assert(ep->getHandle() == ucp_ep);
```
brminich: `NIXL_ASSERT`?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -24,6 +24,65 @@
...
+
+    switch(getState()) {
+    case NIXL_UCX_EP_STATE_NULL:
+        // The EP has not been connected, nothing to do
+        assert(eph == nullptr);
+        return 0;
+    case NIXL_UCX_EP_STATE_DISCONNECTED:
+        // The EP has been disconnected, nothing to do
+        assert(eph == nullptr);
+        return 0;
```
```suggestion
    case NIXL_UCX_EP_STATE_NULL:
    case NIXL_UCX_EP_STATE_DISCONNECTED:
        // The EP has not been connected, or already disconnected -  nothing to do
        NIXL_ASSERT(eph == nullptr);
        return 0;
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -24,6 +24,65 @@
...
+        }
+
+        if (worker == nullptr) {
+            ucp_request_free(request);
+            eph = nullptr;
+            return 0;
+        }
+        break;
+    default:
+        assert(0 && "Invalid endpoint state");
```
brminich: pls use NIXL_ASSERT

src/utils/ucx/ucx_utils.h
```diff
@@ -30,12 +30,29 @@ typedef enum {
...
 
 public:
-    friend class nixlUcxWorker;
+    ucp_ep_h getHandle() const { return eph; }
+    ucp_ep_h* getHandlePtr() { return &eph; }
+    nixl_ucx_ep_state_t getState() { return state; }
+    void setState(nixl_ucx_ep_state_t ep_state) { state = ep_state; }
+    int close(ucp_worker_h worker) { return closeImpl(worker, ucp_ep_close_flags_t(0)); }
+    int closeForce() { return closeImpl(nullptr, UCP_EP_CLOSE_FLAG_FORCE); }
+    int closeNb() { return closeImpl(nullptr, ucp_ep_close_flags_t(0)); }
```
brminich: maybe return `nixl_status_t`?

examples/cpp/nixl_example.cpp
```diff
@@ -64,8 +64,58 @@ void printParams(const nixl_b_params_t& params, const nixl_mem_list_t& mems) {
...
+        return NIXL_SUCCESS;
+    }
+    
+    if (status == NIXL_ERR_BACKEND) {
+        std::cout << "Backend error, exiting\n";
+        return NIXL_ERR_BACKEND;
+    }
+
+    std::cout << "Failed with unexpected error: "
+              << nixlEnumStrings::statusStr(status) << std::endl;
```
brminich: maybe
```suggestion
    if (status == NIXL_ERR_BACKEND) {
        std::cout << "Backend error, exiting\n";
    } else if (status != NIXL_SUCCESS) {
         std::cout << "Failed with unexpected error: "
                  << nixlEnumStrings::statusStr(status) << std::endl;
    }
    
    return status;
```

brminich (commented):
LGTM, have some minor comments

examples/cpp/nixl_example.cpp
```diff
@@ -78,7 +126,8 @@ int main()
 
     // populate required/desired inits
     nixlAgent A1(agent1, cfg);
-    nixlAgent A2(agent2, cfg);
+    // allocate A2 on the heap to kill it to emulate a failed connection
+    nixlAgent *A2 = new nixlAgent(agent2, cfg);
```
iyastreb: consider using unique_ptr instead of raw pointer, and `reset` when needed to delete

src/utils/ucx/ucx_utils.h
```diff
@@ -17,26 +17,48 @@
...
-#include "nixl.h"
-
 enum nixl_ucx_mt_t {
     NIXL_UCX_MT_SINGLE,
     NIXL_UCX_MT_CTX,
     NIXL_UCX_MT_WORKER,
     NIXL_UCX_MT_MAX
 };
 
+enum nixl_ucx_ep_state_t{
```
iyastreb:
```suggestion
enum nixl_ucx_ep_state_t {
```

src/utils/ucx/ucx_utils.h
```diff
@@ -17,26 +17,48 @@
...
+    nixl_ucx_ep_state_t state{nixl_ucx_ep_state_t::NIXL_UCX_EP_STATE_NULL};
+
+    nixl_status_t closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags);
 
 public:
-    friend class nixlUcxWorker;
+    ucp_ep_h getHandle() const { return eph; }
+    ucp_ep_h* getHandlePtr() { return &eph; }
+    nixl_ucx_ep_state_t getState() { return state; }
+    void setState(nixl_ucx_ep_state_t ep_state) { state = ep_state; }
```
iyastreb: do we want to log state change with debug/trace level?

examples/cpp/nixl_example.cpp
```diff
@@ -78,7 +126,8 @@ int main()
 
     // populate required/desired inits
     nixlAgent A1(agent1, cfg);
-    nixlAgent A2(agent2, cfg);
+    // allocate A2 on the heap to kill it to emulate a failed connection
```
tstamler: In general, this is meant to be an example application we point potential new users to, and UCX error handling is probably not something we want to potentially confuse new users with. Having a standalone unit test for this feature would be best, ideally this would be a gtest of some kind. 

The other changes to separate out some functions are nice though, but maybe worth a separate PR.

meson.build
```diff
@@ -27,15 +27,12 @@ cpp = meson.get_compiler('cpp')
...
 
-# Forced to ignore this error due to:
-# https://github.com/abseil/abseil-cpp/issues/1779
-# This must be global since subprojects cannot be assigned with arguments.
-# Adding this configuration only for release build to consider it while developing.
-if get_option('buildtype') == 'release'
-  abseil_flags = cpp.get_supported_arguments('-Wno-error=maybe-uninitialized', '-Wno-maybe-uninitialized')
-  add_global_arguments(abseil_flags, language: 'cpp')
-endif
-abseil_proj = subproject('abseil-cpp')
```
tstamler: why are we removing this?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -625,18 +625,20 @@ nixl_status_t nixlUcxEngine::disconnect(const std::string &remote_agent) {
...
+            flags |= UCP_AM_SEND_FLAG_EAGER;
 
-        ret = uw->sendAm(conn.ep, DISCONNECT,
-                        &hdr, sizeof(struct nixl_ucx_am_hdr),
-                        (void*) localAgent.data(), localAgent.size(),
-                        flags, req);
+            ret = uw->sendAm(conn.getEp(), DISCONNECT,
+                             &hdr, sizeof(struct nixl_ucx_am_hdr),
+                             (void*) localAgent.data(), localAgent.size(),
+                             flags, req);
```
tstamler, line 636: looks like some tabs, we don't want

src/utils/common/meson.build
```diff
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Get required Abseil dependencies from the subproject
+# Initialize Abseil subproject with static library option
+abseil_proj = subproject('abseil-cpp', default_options: ['default_library=static'])
```
tstamler: I see where this is going now, any reason we want abseil to be static?

tstamler (commented):
I think these are good changes at a high level!

src/utils/common/meson.build
evgeny-leksikov: I'm not a meson expert and just started to work with nicx.. I needed to made ucx backed working with `NIXL_DEBUG/NIXL_ASSERT/etc` :) according to https://github.com/ai-dynamo/nixl/pull/236#discussion_r2072645973. I would appreciate for help with pointing to right approach.

tstamler: no problem! I think I can explain. I believe you'll just want to use the nixl_common_deps dependency, defined in our utils: 

https://github.com/ai-dynamo/nixl/blob/21ef52711c636b7b69eb2b89595694747251cb3f/src/utils/common/meson.build#L36

you can see how its used in our core library:

https://github.com/ai-dynamo/nixl/blob/21ef52711c636b7b69eb2b89595694747251cb3f/src/core/meson.build#L18

evgeny-leksikov: thanks! I'll try to enable this way

evgeny-leksikov: @tstamler it seems working, pls take a look

examples/cpp/nixl_example.cpp
evgeny-leksikov: reverted, all the testing moved to gtests

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: why? AFAIU according to other places, I fixed alignment in the code which I modified. Or am I missing something?

src/plugins/ucx/ucx_backend.h
```diff
@@ -41,10 +41,10 @@ struct nixl_ucx_am_hdr {
 class nixlUcxConnection : public nixlBackendConnMD {
     private:
         std::string remoteAgent;
-        nixlUcxEp ep;
-        volatile bool connected;
+        std::shared_ptr<nixlUcxEp> ep = std::make_shared<nixlUcxEp>();
```
iyastreb, line 47: I might miss the previous discussion, but what's the rational of using shared pointer here?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -14,15 +14,151 @@
...
+        setState(NIXL_UCX_EP_STATE_CONNECTED);
+    }
+
+    return status;
+}
+
+nixl_status_t
+nixlUcxEp::closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags)
+{
+    ucs_status_ptr_t request  = nullptr;
```
iyastreb: align or remove space

src/utils/ucx/ucx_utils.cpp
```diff
@@ -14,15 +14,151 @@
...
+        // The EP has not been connected, or already disconnected.
+        // Nothing to do.
+        NIXL_ASSERT(eph == nullptr);
+        return NIXL_SUCCESS;
+    case NIXL_UCX_EP_STATE_FAILED:
+        // The EP was closed in error callback, just return error.
+        eph = nullptr;
+        return NIXL_ERR_REMOTE_DISCONNECT;
+    case NIXL_UCX_EP_STATE_CONNECTED:
+        request = ucp_ep_close_nbx(eph, &req_param);
```
iyastreb, line 124: Handler for connected state is split into 2 parts: first one in switch, and the second one is after switch block
I would propose to keep it in the same place, maybe after switch:
```
    switch(state) {
    case NIXL_UCX_EP_STATE_CONNECTED:
        // Handled below
        break;
...
    }

    ucs_status_ptr_t request = ucp_ep_close_nbx(eph, &req_param);
    ...
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -383,18 +451,14 @@ nixl_status_t nixlUcxWorker::sendAm(nixlUcxEp &ep, unsigned msg_id,
...
+    request = ucp_am_send_nbx(ep.getHandle(), msg_id, hdr, hdr_len, buffer, len, &param);
 
-    if (request == NULL ) {
-        return NIXL_SUCCESS;
-    } else if (UCS_PTR_IS_ERR(request)) {
-        /* TODO: MSW_NET_ERROR(priv->net, "unable to complete UCX request (%s)\n",
-                         ucs_status_string(UCS_PTR_STATUS(request))); */
-        return NIXL_ERR_BACKEND;
+    if (UCS_PTR_IS_PTR(request)) {
+        req = (void*)request;
```
iyastreb, line 455: Is it ok to use C style casts?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -426,71 +490,59 @@ nixl_status_t nixlUcxWorker::read(nixlUcxEp &ep,
...
-    if (request == NULL ) {
-        return NIXL_SUCCESS;
-    } else if (UCS_PTR_IS_ERR(request)) {
-        /* TODO: MSW_NET_ERROR(priv->net, "unable to complete UCX request (%s)\n",
-                         ucs_status_string(UCS_PTR_STATUS(request))); */
-        return NIXL_ERR_BACKEND;
+    ucs_status_ptr_t request = ucp_get_nbx(ep.getHandle(), laddr, size, raddr,
+                                           rk.rkeyh, &param);
+    if (UCS_PTR_IS_PTR(request)) {
+        req = (void*)request;
```
iyastreb, line 504: Q for casts: do we put a space between cast type and var? Like in other places in this PR

src/utils/ucx/ucx_utils.h
```diff
@@ -32,11 +32,48 @@ enum nixl_ucx_mt_t {
 };
 
 class nixlUcxEp {
+    enum nixl_ucx_ep_state_t {
```
iyastreb, line 35: do we want to use C++11 `enum class`?

src/utils/ucx/ucx_utils.h
```diff
@@ -32,11 +32,48 @@ enum nixl_ucx_mt_t {
...
+    ucp_ep_h            eph{nullptr};
+    nixl_ucx_ep_state_t state{NIXL_UCX_EP_STATE_NULL};
+
+    static void err_cb(void *arg, ucp_ep_h ucp_ep, ucs_status_t status);
+    void setState(nixl_ucx_ep_state_t new_state);
+    nixl_status_t closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags);
 
 public:
-    friend class nixlUcxWorker;
+    // TODO: Add read/write methods to avoid using raw handle outside.
```
iyastreb, line 50: Typically public section part comes first in C++ class, followed by private part, for readability purposes
But I'm not sure if this is a NIXL guideline

src/utils/ucx/ucx_utils.h
```diff
@@ -32,11 +32,48 @@ enum nixl_ucx_mt_t {
...
+    ucp_ep_h getHandle() const { return eph; }
```
iyastreb, line 51: IMO ideally we should not expose raw `ucp_ep_h` to the outside.
We should probably neither accept `ucp_worker_h` as input argument, but a corresponding NIXL wrapper instead
Up to you

test/gtest/ucx_tests.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include <nixl_types.h>
+#include "nixl.h"
+
+namespace gtest {
+namespace ucx {
+
+namespace nixl {
+    static std::unique_ptr<nixlAgent> createAgent(const std::string& name)
+    {
+        return std::make_unique<nixlAgent>(name, nixlAgentConfig(true));
```
iyastreb, line 29: can we use C++17 features? Then we can consider std::string_view
If not, then we can pass by value + std::move
```
static std::unique_ptr<nixlAgent> createAgent(std::string name)
{
    return std::make_unique<nixlAgent>(std::move(name), nixlAgentConfig(true));

File: test/gtest/ucx_tests.cpp [hunk 24]
```diff
@@ -0,0 +1,217 @@
...
+    {
+        std::vector<nixl_backend_t> plugins;
+        nixl_status_t status = agent.getAvailPlugins(plugins);
+        EXPECT_EQ(status, NIXL_SUCCESS);
+        auto it = std::find(plugins.begin(), plugins.end(), "UCX");
+        EXPECT_NE(it, plugins.end()) << "UCX plugin not found";
+
+        nixl_b_params_t params;
+        nixl_mem_list_t mems;
+        status = agent.getPluginParams("UCX", mems, params);
```
iyastreb, line 42: can use {} instead of params?

test/gtest/ucx_tests.cpp
```diff
@@ -0,0 +1,217 @@
...
+        EXPECT_EQ(NIXL_SUCCESS, status);
+        EXPECT_NE(nullptr, backend_handle);
+        return backend_handle;
+    }
+
+    static nixl_opt_args_t createExtraParams(nixlBackendH* backend)
+    {
+        nixl_opt_args_t extra_params;
+        extra_params.backends = {backend};
+        return extra_params;
```
iyastreb, line 56: can we just `return {backend}:`?

test/gtest/ucx_tests.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+        nixlBasicDesc rReq_dst;
+        nixl_xfer_dlist_t rReq_descs(DRAM_SEG);
+        nixl::fillRegList(rReq_descs, rReq_dst, rData);
+
+        sExtraParams.notifMsg = "notification";
+        sExtraParams.hasNotif = true;
+        nixlXferReqH* req_handle;
+        EXPECT_EQ(NIXL_SUCCESS, sAgent->createXferReq(NIXL_WRITE, sReq_descs,
+                                                    rReq_descs, "receiver",
```
iyastreb: alignment

test/gtest/ucx_tests.cpp
```diff
@@ -0,0 +1,217 @@
...
+                  << std::endl;
+    }
+
+    static void test_xfer(bool receiver_failure = false)
+    {
+        const size_t len                  = 256;
+        std::unique_ptr<nixlAgent> sAgent = nixl::createAgent("sender");
+        std::unique_ptr<nixlAgent> rAgent = nixl::createAgent("receiver");
+
+        nixlBackendH* sBackend = nixl::createUcxBackend(*sAgent);
```
iyastreb, line 133: Is this object destroyed automatically by agent dtor?
Sorry, don't have nixl codebase yet

src/core/nixl_agent.cpp
```diff
@@ -824,7 +825,7 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) {
 
     NIXL_LOCK_GUARD(data->lock);
     // If the status is done, no need to recheck.
-    if (req_hndl->status != NIXL_SUCCESS) {
+    if (req_hndl->status == NIXL_IN_PROG) {
```
artpol84, line 828: This seems to be missing other error cases.
So "else" that would check for "!= NIXL_SUCCESS" is needed to detect real errors.

src/core/nixl_plugin_manager.cpp
```diff
@@ -277,7 +277,8 @@ void nixlPluginManager::discoverPluginsFromDir(const std::string& dirpath) {
     std::error_code ec;
     std::filesystem::directory_iterator dir_iter(dir_path, ec);
     if (ec) {
-        std::cerr << "Error accessing directory: " << ec.message() << std::endl;
+        std::cerr << "Error accessing directory(" << dir_path << "):"
```
artpol84, line 280: I thought there was NIXL-derived streams for I/O. See here:
https://github.com/ai-dynamo/nixl/pull/257/files#diff-70d20dbd8f24f38c9d54eb28a879ee16bc492a43d107103617789e3ed29e0826R458

src/plugins/ucx/ucx_backend.cpp
artpol84: @tstamler the "if" statement above was introduced

src/plugins/ucx/ucx_backend.h
evgeny-leksikov: We did not have discussion on that yet, the reason is we cannot copy `nixlUcxEp` object since it's an argument for `nixlUcxEp::err_cb`

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: IMO these parts are separated logically, 1st part just initiates closing, 2nd one is blocking part.

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: I see in many places, I did not find exact nixl code style (btw @tstamler do we have some nixl code style guide?)
mostly I kept existing code as-is and used c++ type cast in new code

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: I think we'll get an A in https://github.com/ai-dynamo/nixl/pull/236/files#r2077514210

src/utils/ucx/ucx_utils.h
evgeny-leksikov: I reused existing code style in the file

src/utils/ucx/ucx_utils.h
evgeny-leksikov: see comment with TODO line above

test/gtest/ucx_tests.cpp
evgeny-leksikov: > can we use C++17 features? Then we can consider std::string_view

why?

> If not, then we can pass by value + std::move

why? name is a const reference

test/gtest/ucx_tests.cpp
evgeny-leksikov: why?

test/gtest/ucx_tests.cpp
evgeny-leksikov: I don't think so, could you explain?

test/gtest/ucx_tests.cpp
evgeny-leksikov: it should be, dtor is protected

test/gtest/ucx_tests.cpp
```diff
@@ -0,0 +1,217 @@
...
+        EXPECT_EQ(NIXL_SUCCESS, sAgent->loadRemoteMD(rMeta, sMeta_remote));
+        std::string rMeta_remote;
+        EXPECT_EQ(NIXL_SUCCESS, rAgent->loadRemoteMD(sMeta, rMeta_remote));
+
+        nixlBasicDesc sReq_src;
+        nixl_xfer_dlist_t sReq_descs(DRAM_SEG);
+        nixl::fillRegList(sReq_descs, sReq_src, sData);
+
+        if (receiver_failure) {
+            rAgent->deregisterMem(rDlist, &rExtraParams);
```
artpol84, line 167: IIUC, at `loadRemoteMD()`, `ucp_ep_create` is invoked. This doesn't guarantee that a connection is established.
Does it make sense to test 3 paths here:
* error-free
* rAgent destroyed right after `loadRemoteMD()` (as now)
* a round of successful communication is followed by rAgent destroy (this exercise another path). Even though this is not visible to NIXL, IMO it is worth testing.

artpol84 (approved):

artpol84:
\ok to test 273abe21c9d68048600df5b5fcab4babd7eb4bb2

test/gtest/ucx_tests.cpp
evgeny-leksikov: per offline discussion agreed to extend testing in next PR

src/core/nixl_plugin_manager.cpp
evgeny-leksikov: by default gtest was trying to load plugins from default linux path, see https://github.com/ai-dynamo/nixl/pull/236/files#diff-992f3a6e2aa64bb5770f78d106d9beaa3f618681f0497e08e82da6363e62e481R115

src/core/nixl_agent.cpp
evgeny-leksikov: `req_hndl->status` already has correct error status, whether success or failure

src/utils/ucx/ucx_utils.cpp
```diff
@@ -14,15 +14,151 @@
...
+        request = ucp_ep_close_nb(ucp_ep, UCP_EP_CLOSE_MODE_FORCE);
+        if (UCS_PTR_IS_PTR(request)) {
+            ucp_request_free(request);
+        }
+        break;
+    default:
+        NIXL_FATAL << "Invalid endpoint state: " << ep->state;
+    }
+
+    NIXL_DEBUG << "ep " << ep->getHandle() << ": state " << ep->state;
```
brminich: seems it is not needed given the trace at line 50 above

brminich (approved):
👍 besides minor comment

src/core/nixl_agent.cpp
evgeny-leksikov: btw, in old logic the real error was missed due to:
```
        req_hndl->status = req_hndl->engine->checkXfer(
                                     req_hndl->backendHandle);
```
set `req_hndl->status` to `NIXL_SUCCESS` if no new notifications, I'm not sure if it's bug in the engine or expected behavour but I saw explicit comment for this, so decided to fix it here.

PR #237, opened 2025-05-02
CI: reduce size of posix test

tstamler:


test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -36,7 +36,7 @@
 namespace {
     const size_t page_size = sysconf(_SC_PAGESIZE);
 
-    constexpr int default_num_transfers = 4 * 1024;
+    constexpr int default_num_transfers = 512;
```
aranadive: Shouldnt this should be passed as an cli arg? That way we only run a few in CI and not block a run for too long.

tstamler: you right, it already supports this

PR #245, opened 2025-05-03
BINDINGS/RUST: Example, fix building, tests

ikryukov:
- Added simple example using nixl in Rust in single process
- Fixed build: added libs and default path tweaks
- Fixed tests:

```
test_get_backend_params: Modified to prefer the UCX plugin over GDS to avoid GPU dependency
test_basic_agent_lifecycle: Added proper error handling and graceful degradation for UCX networking issues
```

- Made tests code Rust idiomatic: error handling, removing unwrap, using `match` 
- Updated `cc` version


examples/rust/Cargo.lock
```diff
@@ -0,0 +1,1060 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
roiedanino, line 2: Is it standard in Rust to track `Cargo.lock` in git?

examples/rust/Cargo.lock
```diff
@@ -0,0 +1,1060 @@
...
+checksum = "32a4622180e7a0ec044bb555404c800bc9fd9ec262ec147edd5989ccd0c02cd3"
+
+[[package]]
+name = "windows_aarch64_msvc"
+version = "0.52.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09ec2a7bb152e2252b53fa7803150007879548bc709c039df7627cabbd05d469"
+
+[[package]]
+name = "windows_i686_gnu"
```
roiedanino, line 1039: @tstamler was bothered by that in my recent PR, maybe its fine

examples/rust/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[package]
+name = "nixl-examples"
+version = "0.1.0"
```
roiedanino: should be 0.1.1

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,241 @@
...
+    // Create opt args with backends
+    let mut opt_args1 = OptArgs::new()?;
+    opt_args1.add_backend(&backend1)?;
+
+    let mut opt_args2 = OptArgs::new()?;
+    opt_args2.add_backend(&backend2)?;
+
+    // Register memory
+    storage1.register(&agent1, &opt_args1)?;
+    storage2.register(&agent2, &opt_args2)?;
```
roiedanino: Do we also need to deregister memory?

src/bindings/rust/build.rs
```diff
@@ -17,29 +17,47 @@ use std::env;
 use std::path::PathBuf;
 
 fn main() {
-    let nixl_root_path = env::var("NIXL_PREFIX")
-                            .unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
+    let nixl_root_path =
+        env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
     let nixl_include_path = format!("{}/include", nixl_root_path);
-    let nixl_lib_path = format!("{}/lib64", nixl_root_path);
+    let nixl_lib_path = format!("{}/lib/x86_64-linux-gnu", nixl_root_path);
```
roiedanino: Is this change always correct?

src/bindings/rust/build.rs
```diff
@@ -17,29 +17,47 @@ use std::env;
...
         .compiler("g++") // Ensure we're using the C++ compiler
         .file("wrapper.cpp")
         .flag("-std=c++17")
         .flag("-fPIC")
         .include(&nixl_include_path)
         // Change ABI flag if necessary to match your precompiled libraries:
-            //    .flag("-D_GLIBCXX_USE_CXX11_ABI=0")
+        //    .flag("-D_GLIBCXX_USE_CXX11_ABI=0")
         .flag("-Wno-unused-parameter")
-        .flag("-Wno-unused-variable")
```
roiedanino, line 47: Any reason to delete that?

src/bindings/rust/build.rs
ikryukov: according to docker file - yes
```
ENV NIXL_PREFIX=/usr/local/nixl
ENV NIXL_PLUGIN_DIR=/usr/local/nixl/lib/x86_64-linux-gnu/plugins
RUN echo "/usr/local/nixl/lib/x86_64-linux-gnu" > /etc/ld.so.conf.d/nixl.conf && \
    echo "/usr/local/nixl/lib/x86_64-linux-gnu/plugins" >> /etc/ld.so.conf.d/nixl.conf && \
    ldconfig
```

src/bindings/rust/build.rs
ikryukov: it is not deleted, just added conditional define and do compile call few lines later

examples/rust/Cargo.lock
ikryukov: yes, for example here: https://github.com/rust-lang/rust/blob/master/Cargo.lock

examples/rust/src/single_process_example.rs
ikryukov: here we register memory that stores `RegistrationHandle` and it implements `Drop` (sort of destructor) that deregister memory for us.

ikryukov:
I have revised the comments, please take a look. @roiedanino

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    // Create UCX backends
+    println!("Creating UCX backends...");
+    let (_, params1) = agent1.get_plugin_params("UCX")?;
+    let (_, params2) = agent2.get_plugin_params("UCX")?;
+
+    let backend1 = agent1.create_backend("UCX", &params1)?;
+    let backend2 = agent2.create_backend("UCX", &params2)?;
+    println!("Created UCX backends");
```
roiedanino: The logs are a bit verbose. Maybe we can log just the important steps and errors?

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,241 @@
...
+
+            if !notifs.is_empty()? {
+                println!("Agent2 received notification");
+                received_notification = true;
+            }
+        }
+
+        if !completed || !received_notification {
+            thread::sleep(Duration::from_millis(100));
+        }
```
roiedanino, line 210: Why?

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    if !completed || !received_notification {
+        println!("Warning: Transfer or notification did not complete within timeout period.");
```
roiedanino: error?

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,241 @@
...
+    let data2_after = storage2.as_slice();
+    println!("\nAfter transfer:");
+    println!("Agent1 memory bytes[0..4]: {:?}", &data1[0..4]);
+    println!("Agent2 memory bytes[0..4]: {:?}", &data2_after[0..4]);
+
+    if data2_after[0] == 0xaa {
+        println!("SUCCESSFUL - Memory was updated with Agent1's pattern (0xaa)!");
+    } else {
+        println!("FAILED - Memory was not updated as expected.");
+        println!("  Expected: 0xaa, Got: 0x{:02x}", data2_after[0]);
```
roiedanino: Error exit?
We are planning adding rust tests to CI and that example also can be part of it, so we would want to fail the CI when this example fails

src/bindings/rust/build.rs
```diff
@@ -22,11 +22,17 @@ fn main() {
...
+
     // Tell cargo to look for shared libraries in the specified directories
     println!("cargo:rustc-link-search={}", nixl_lib_path);
+    // Add rpath to embed library path in the binary
+    println!("cargo:rustc-link-arg=-Wl,-rpath,{}", nixl_lib_path);
 
     // Build the C++ wrapper
-    cc::Build::new()
+    let mut cc_builder = cc::Build::new();
+    cc_builder
```
roiedanino, line 83: Maybe:

```suggestion
    let mut cc_builder = cc::Build::new()
        .cpp(true)
        .compiler("g++")
        ...
```

src/bindings/rust/src/tests.rs
```diff
@@ -22,6 +22,27 @@
...
+
+    // Helper function to create an agent with error handling
+    fn create_test_agent(name: &str) -> Result<Agent, NixlError> {
+        Agent::new(name)
+    }
+
+    // Helper function to find a plugin by name
+    fn find_plugin(plugins: &StringList, name: &str) -> Result<String, NixlError> {
+        let mut found_plugin = None;
+        for i in 0..plugins.len()? {
```
roiedanino: Maybe we can:
```rust
    match plugins.iter().find(|&plugin| plugin == name) {
         Some(plugin_name) => plugin_name.to_string(),
        None => Err("Plugin not found"),

    }

```

src/bindings/rust/src/tests.rs
```diff
@@ -326,117 +355,304 @@ mod tests {
...
+                let mut notifs = NotificationMap::new()?;
+                let start = std::time::Instant::now();
+
+                while start.elapsed() < timeout {
+                    match agent2.get_notifications(&mut notifs, None) {
+                        Ok(_) if !notifs.is_empty()? => {
+                            println!("Got notifications");
+                            break;
+                        }
+                        Ok(_) => std::thread::sleep(Duration::from_millis(100)),
```
roiedanino, line 443: why?

src/bindings/rust/src/tests.rs
```diff
@@ -326,117 +355,304 @@ mod tests {
...
+            // Should have notifications from NotifSender
+            if let Some(Ok(agent_name)) = agents.next() {
+                assert_eq!(agent_name, "NotifSender");
+
+                // Verify notification content
+                let notifications = notifs.get_notifications(agent_name)?;
+                let notif_count = notifs.get_notifications_size(agent_name)?;
+
+                // May have 1 or 2 notifications depending on whether both were processed
+                assert!(notif_count > 0, "Should have at least one notification");
```
roiedanino, line 560: should also check notifications.len() == notif_count

src/bindings/rust/src/tests.rs
```diff
@@ -101,29 +122,37 @@ mod tests {
...
         }
 
-        // Verify we can access the memory types
+        // Print memory types
         for mem_type in backend_mems.iter() {
-            println!("Backend memory type: {:?}", mem_type.unwrap());
+            if let Ok(mem_type) = mem_type {
+                println!("Backend memory type: {:?}", mem_type);
+            }
         }
```
roiedanino: Maybe its too verbose?

roiedanino (dismissed):

examples/rust/src/single_process_example.rs
ikryukov: Will add tracing and adjusting all print statements at different levels, such as debugging, information, warnings, errors, and so on.

src/bindings/rust/src/tests.rs
ikryukov: removed

src/bindings/rust/src/tests.rs
ikryukov: It is not displayed on cargo tests anyway.

examples/rust/src/single_process_example.rs
ikryukov: removed

src/bindings/rust/build.rs
ikryukov: I cannot do this due to borrow checker.

src/bindings/rust/src/tests.rs
ikryukov: It is not trivial, since the iterator returned would be consumed.

examples/rust/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+[package]
+name = "nixl-examples"
+version = "0.1.1"
+edition = "2021"
+
+[dependencies]
+nixl-sys = { path = "../../src/bindings/rust" }
+clap = { version = "4.5.38", features = ["derive"] }
```
ryanolson, line 23: should this be a dev-dependency?

examples/rust/src/single_process_example.rs
```diff
@@ -0,0 +1,235 @@
...
+use std::thread;
+use std::time::Duration;
+
+use tracing::{debug, error, info};
+
+// Use the same agent names as in working C++ example
+const AGENT1_NAME: &str = "Agent001";
+const AGENT2_NAME: &str = "Agent002";
+
+fn main() -> Result<(), Box<dyn Error>> {
```
ryanolson, line 30: this test look a lot like: https://github.com/ai-dynamo/nixl/blob/main/src/bindings/rust/src/tests.rs#L329

we should move all these to `<cargo-root>/tests` rather than be in the `src` directory

src/bindings/rust/build.rs
```diff
@@ -22,11 +22,17 @@ fn main() {
     let nixl_include_path = format!("{}/include", nixl_root_path);
     let nixl_lib_path = format!("{}/lib/x86_64-linux-gnu", nixl_root_path);
 
+    // Check if etcd is enabled via environment variable
+    let etcd_enabled = env::var("HAVE_ETCD").map(|v| v != "0").unwrap_or(false);
+
     // Tell cargo to look for shared libraries in the specified directories
     println!("cargo:rustc-link-search={}", nixl_lib_path);
+    // Add rpath to embed library path in the binary
+    println!("cargo:rustc-link-arg=-Wl,-rpath,{}", nixl_lib_path);
```
ryanolson: I don't think we want to do this.  Often times we do, but in this case we I don't believe we want to do this.

We will eventually compile and get the path via png-config, or use a stubs.cc.  if we embed the rpath with the stubs, we will not pick up the actual libs in the install path.

src/bindings/rust/build.rs
```diff
@@ -36,10 +42,22 @@ fn main() {
...
+    // Define HAVE_ETCD if etcd is enabled
+    if etcd_enabled {
+        cc_builder.define("HAVE_ETCD", "1");
+    }
+
+    cc_builder.compile("wrapper");
 
     // Link against NIXL libraries in correct order
+    // Only link against etcd-cpp-api if it's enabled
+    if etcd_enabled {
```
ryanolson, line 104: we should never use c++ etcd.  we can use the rust etcd-client crate if we wish to use etcd.

src/bindings/rust/src/lib.rs
```diff
@@ -628,6 +630,230 @@ impl Agent {
         self.inner.write().unwrap().invalidate_all_remotes()
     }
 
+    /// Send this agent's metadata to etcd
+    ///
+    /// This enables other agents to discover this agent's metadata via etcd.
+    ///
+    /// # Arguments
+    /// * `opt_args` - Optional arguments for sending metadata
+    pub fn send_local_md(&self, opt_args: Option<&OptArgs>) -> Result<(), NixlError> {
```
ryanolson, line 637: I would feature flag this around using the rust etcd-client.  

in dynamo, we already have that available to use as part of the discovery service.

src/bindings/rust/src/lib.rs
```diff
@@ -628,6 +630,230 @@ impl Agent {
...
+    /// Send a notification to a remote agent
+    ///
+    /// # Arguments
+    /// * `remote_agent` - Name of the remote agent to send notification to
+    /// * `message` - The notification message to send
+    /// * `backend` - Optional backend to use for sending the notification
+    ///
+    /// # Returns
+    /// `Ok(())` if the notification was sent successfully
+    pub fn send_notification(
```
ryanolson, line 807: thanks for adding this.  I was going to but didn't get the chance.

src/bindings/rust/wrapper.h
```diff
@@ -82,6 +82,18 @@ nixl_capi_status_t nixl_capi_load_remote_md(nixl_capi_agent_t agent, const void*
 // Invalidate remote agent metadata
 nixl_capi_status_t nixl_capi_invalidate_remote_md(nixl_capi_agent_t agent, const char* remote_agent);
 
+// Invalidate local metadata in etcd
+nixl_capi_status_t nixl_capi_invalidate_local_md(nixl_capi_agent_t agent, nixl_capi_opt_args_t opt_args);
+
+// Check if remote metadata is available
+nixl_capi_status_t nixl_capi_check_remote_md(nixl_capi_agent_t agent, const char* remote_name, nixl_capi_xfer_dlist_t descs);
+
+// Send local metadata to etcd
```
ryanolson, line 91: same comment above for native etcd.

ryanolson (dismissed):
Overall great PR, thanks for the contributions. 

Changes:
- drop rpath 
- drop c++ etcd bits, optionally add native `etcd-client`; however note, that dynamo and do that as well. overall, I think that etcd should not be part of the bindings, but used as a dev-dependency and the native client used in the examples/tests.

I do think we should add a tokio feature flag to the bindings and add in some of the multi-threaded runtime tokio bits to the bindings.

src/bindings/rust/build.rs
ikryukov: C++ etcd is used by Nixl (actually it is optional), but not Rust. Here we want to know whether c++ nixl was built with etcd support or not.

examples/rust/Cargo.toml
ikryukov: No, since this is necessary for example app.

examples/rust/src/single_process_example.rs
ikryukov: The main goal was to demonstrate how to use nixl with Rust by analogy to C++/Python examples.

src/bindings/rust/src/lib.rs
ikryukov: Dynamo is responsible for managing metadata by itself, but in other use cases it might be useful?

src/bindings/rust/build.rs
ikryukov: fixed

aranadive:
@ikryukov are all comments resolved? @ryanolson & @roiedanino to confirm.

ikryukov:
> @ikryukov are all comments resolved? @ryanolson & @roiedanino to confirm.
From my side, yes.

src/bindings/rust/build.rs
roiedanino: As `/lib64` is the correct path for Red Hat-based distributions (CI is running on ubuntu image), perhaps it would be better to use `os_info` and set path accordingly?

aranadive (approved):

PR #246, opened 2025-05-04
TEST: Add data transfer test in gtest

yosefe:
Add a gtest to run data transfer over UCX backend
The test could be extended to other backends and for GPU memory

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ucx/TestTransfer
[ RUN      ] ucx/TestTransfer.RandomSizes/0
[     INFO ] 4096x8x10=327680 bytes in 0.00483286 seconds (0.0631459 GB/s)
[     INFO ] 32768x64x100=209715200 bytes in 0.0124641 seconds (15.67 GB/s)
[     INFO ] 1000000x100x1000=100000000000 bytes in 4.02394 seconds (23.1446 GB/s)
[       OK ] ucx/TestTransfer.RandomSizes/0 (5676 ms)
[----------] 1 test from ucx/TestTransfer (5676 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (5676 ms total)
[  PASSED  ] 1 test.
```

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -56,13 +56,6 @@ class nixlAgent {
         nixlAgent &operator=(nixlAgent&&) noexcept;
 
         /**
-         * @brief  Get the name of the agent
-         *
-         * @return std::string The name of the agent
-         */
-        std::string getName () const;
-
-        /**
          * @brief  Discover the available supported plugins found in the plugin paths
          *
          * @param  plugins [out] Vector of available backend plugins
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -51,6 +51,13 @@ class nixlAgent {
         ~nixlAgent ();
 
         /**
+         * @brief  Get the name of the agent
+         *
+         * @return std::string The name of the agent
+         */
+        std::string getName () const;
+
+        /**
          * @brief  Discover the available supported plugins found in the plugin paths
          *
          * @param  plugins [out] Vector of available backend plugins
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -24,6 +24,7 @@
 #include "nixl_types.h"
 #include "nixl_params.h"
 #include "nixl_descriptors.h"
+#include <memory>
 
 /**
  * @class nixlAgent
@@ -32,7 +33,7 @@
 class nixlAgent {
     private:
         /** @var  data  The members in agent class wrapped into single nixlAgentData member. */
-        nixlAgentData* data;
+        std::unique_ptr<nixlAgentData> data;
 
     public:
         /*** Initialization and Registering Methods ***/
@@ -49,6 +50,10 @@ class nixlAgent {
          * @brief Destructor for nixlAgent object
          */
         ~nixlAgent ();
+
+        /* Declare move operations (needed because we declared a destructor) */
+        nixlAgent(nixlAgent&&) noexcept;
+        nixlAgent &operator=(nixlAgent&&) noexcept;
 
         /**
          * @brief  Get the name of the agent
--- a/test/gtest/common.h
+++ b/test/gtest/common.h
@@ -1,0 +1,42 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iostream>
+#include <iomanip>
+
+namespace gtest {
+
+class Logger {
+public:
+    Logger(const std::string &title = "INFO")
+    {
+        std::cout << "[ " << std::setw(8) << title << " ] ";
+    }
+
+    ~Logger()
+    {
+        std::cout << std::endl;
+    }
+
+    template<typename T> Logger &operator<<(const T &value)
+    {
+        std::cout << value;
+        return *this;
+    }
+};
+
+} // namespace gtest
--- a/test/gtest/test_transfer.cpp
+++ b/test/gtest/test_transfer.cpp
@@ -159,20 +159,27 @@ protected:
         }
     }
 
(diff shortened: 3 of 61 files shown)


yosefe:
When running in CI, this is the test output:
```
[ RUN      ] ucx/TestTransfer.RandomSizes/0
[1746346069.603371] [runner-s5xjxous1-project-170402-concurrent-0:207  :0]          parser.c:2326 UCX  WARN  unused environment variable: UCX_INSTALL_DIR
[1746346069.603371] [runner-s5xjxous1-project-170402-concurrent-0:207  :0]          parser.c:2326 UCX  WARN  (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
[     INFO ] 4096x8x10=327680 bytes in 0.00155445 seconds (0.196324 GB/s)
[     INFO ] 32768x64x100=209715200 bytes in 0.126199 seconds (1.54765 GB/s)
[     INFO ] 1000000x100x1000=100000000000 bytes in 34.5672 seconds (2.69424 GB/s)
[       OK ] ucx/TestTransfer.RandomSizes/0 (34783 ms)
[----------] 1 test from ucx/TestTransfer (34783 ms total)
```

test/gtest/common.h
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iostream>
+#include <iomanip>
+
+namespace gtest {
+
+class Logger {
```
etoledano, line 23: Have you considered using [RecordProperty](https://google.github.io/googletest/reference/testing.html#Test::RecordProperty) to track values to the XML, and for errors/warning (if needed)  `abseil::Log`, for consistency?

yosefe: I've tried it (following your suggestion) and IMO it's less convenient because the properties values are buried in the XML report and need more tools to extract and print them nicely.
Regarding absl: we'd probably want to use INFO log level, but then gtest would also need to raise the default log level to INFO, WDYT?

etoledano: Sorry I didn't get this right away, I thought those prints are for errors so I assumed it makes sense to allow log level to be adjusted globally, but if its for info then it'll probably flood the log with code's prints. I guess we can use different log sinks and make gtest log to a specific one, not sure if it over complicates or not, no strong opinion on that.

test/gtest/test_transfer.cpp
```diff
@@ -0,0 +1,263 @@
...
+                NIXL_WRITE,
+                makeDescList<nixlBasicDesc>(src_buffers, src_mem_type),
+                makeDescList<nixlBasicDesc>(dst_buffers, dst_mem_type),
+                to->getName(), xfer_req);
+        ASSERT_EQ(status, NIXL_SUCCESS);
+        EXPECT_NE(xfer_req, nullptr);
+
+        auto start_time = absl::Now();
+        for (size_t i = 0; i < repeat; i++) {
+            status = from->postXferReq(xfer_req);
```
tvegas1: ideally add `extra_params` here and to `doTransfer()` to add test instance when plugin supports notifications that skips line 210

tvegas1: idea: add multiple `->postXferReq()` beign started, then `->getXferStatus()` and measure duration. (ucx plugin: adding a `ucp_ep_fence()/strong flush` would likely block those independent `->postXferReq()` and slow down?)

tvegas1: (add threading ability)

test/gtest/test_transfer.cpp
```diff
@@ -0,0 +1,263 @@
...
+            ASSERT_GE(status, NIXL_SUCCESS);
+
+            status = from->genNotif(to->getName(), NOTIF_MSG);
```
tvegas1: i did not understand what genNotif() followed by waitForOneNotif() do? if it's not related to the transfer we might remove it and or test it separately? we don't want to incidentally progress() line 210, which might impact the testing of `->getXferStatus()` line 215?

tvegas1: also do we have ordering guarantee between genNotif() and postXferReq? ucx plugin EFA/Multirail etc i understand we don't have it?

do we need to instead repeatedly call `->getXferStatus()` while making sure we also progress the other end?

src/api/cpp/nixl.h
```diff
@@ -55,6 +55,13 @@ class nixlAgent {
         nixlAgent(nixlAgent&&) noexcept;
         nixlAgent &operator=(nixlAgent&&) noexcept;
 
+        /**
+         * @brief  Get the name of the agent
+         *
+         * @return std::string The name of the agent
+         */
+        std::string getName () const;
```
tstamler: can we do this without an API change? we typically rely on users to know their own name. Exposing a get function for the name could be something worth adding, but right now it doesn't fit the rest of the API header, so at the very least I think we should move it to a separate PR.

test/gtest/test_transfer.cpp
yosefe: IMO a more generic way to to send notification explicitly

yosefe: since each transfer request already contains a list of descriptors, we have several outstanding messages. we can extend it in the future and also add multi threading.

test/gtest/test_transfer.cpp
yosefe: 1. it's on different agents: "from" is posting transfer req + notification, and then we wait for it to be received on the target agent "to". I'll add explicit check for that.
2. Yes, the notification should be received after xfer req is received. I guess next step would be to add data validation. getNotifs should progress both sides. but i can change the progress to check xfer on one side and do getNotifs on other side.

src/api/cpp/nixl.h
yosefe: will remove it for now (though IMO it's good to have it in general, at least for C++ API)

test/gtest/test_transfer.cpp
tvegas1: it contains a list of descriptors but posts rma in sequence, then flush (could become fence?). if we parallel poxtXferReq, some of them will already do flush (could become fence?) while others are still posting rma. so flush/fence of a postXferReq could have impact on unrelated postXferReq.

agree this is idea for later PR

tvegas1: but the inline has ordering guarantee

test/gtest/test_transfer.cpp
tvegas1: in practice could be hard to observe as I do not see how the order is guaranteed (arrival order of getXferStatus and getNotif

tstamler (approved):
Looks good to me to merge after rebase

test/gtest/test_transfer.cpp
yosefe: you are right, unfortunately there is no ordering guarantee in the API today.

yosefe: fixed

test/gtest/test_transfer.cpp
yosefe: done

tvegas1 (approved):

PR #248, opened 2025-05-04
Cost estimation

ovidiusm:
Defines and implements the cost estimation code for transfers

Depends on https://github.com/openucx/ucx/pull/10672


src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
     return NIXL_IN_PROG;
 }
 
+/////////////////////////////// COST ESTIMATION API /////////////////////////////////////////////
+// TODO: these definitions need to appear in UCX not NIXL. Provided for code review purposes.
+
+/**
+ * @enum ucp_op_type_t
+ * @brief UCP operation types.
+ *
```
brminich: can u pls open the corresponding PR in UCX GH?

src/utils/ucx/meson.build
```diff
@@ -13,13 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ucx_utils_dep = [ ucx_dep ]
+ucx_utils_dep = [ ucx_dep, nixl_common_dep ]
```
brminich: is it unrelated fixes?

src/utils/ucx/ucx_utils.h
```diff
@@ -124,6 +124,24 @@ class nixlUcxWorker {
...
+     * @param local_addr     Pointer to the local buffer (used for memory type detection).
+     * @param length         Size of the message to estimate cost for.
+     * @param nixl_op        NIXL operation type (NIXL_READ or NIXL_WRITE).
+     * @param duration_sec   [out] Estimated duration in seconds.
+     * @return nixl_status_t NIXL_SUCCESS on success,
+     *                       NIXL_ERR_NOT_SUPPORTED if estimation is unavailable,
+     *                       NIXL_ERR_BACKEND on other UCX errors.
+     */
+    nixl_status_t estimateCost(nixlUcxEp ep,
+                               void* local_addr,
```
brminich: we'd need both buffers - origin and target (also needs to be reflected in UCX API)

src/api/cpp/nixl_types.h
```diff
@@ -116,6 +116,18 @@ using nixl_b_params_t = std::unordered_map<std::string, std::string>;
  */
 using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
 
+/**
+ * @class nixlCostEstimate
+ * @brief Cost estimation for a transfer request.
+ */
+class nixlCostEstimate {
```
brminich: we'd need to make it backward compatible. Pls consider using struct with version, pimpl or some other mechanism

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+        (data->remoteSections.count(req_hndl->remoteAgent) == 0)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (!req_hndl->engine) {
+        // Should not happen if req_hndl is valid
+        return NIXL_ERR_UNKNOWN;
+    }
+
+    // Call the backend engine's implementation
```
brminich: minor: i'd avoid such comments, as it is obvious from the code

src/plugins/ucx/meson.build
```diff
@@ -23,15 +23,15 @@ endif
 if 'UCX' in static_plugins
     ucx_backend_lib = static_library('UCX',
                'ucx_backend.cpp', 'ucx_backend.h', 'ucx_plugin.cpp',
-               dependencies: [nixl_infra, ucx_utils_dep, serdes_interface, cuda_dep, ucx_dep, thread_dep],
-               include_directories: nixl_inc_dirs,
+               dependencies: [nixl_infra, nixl_common_dep, ucx_utils_dep, serdes_interface, cuda_dep, ucx_dep, thread_dep],
```
brminich: why is it needed?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+        estimate->duration = 0.0;
+        return NIXL_SUCCESS;
+    }
+
+    // Iterate through descriptors to estimate cost for each pair.
+    for (i = 0; i < lcnt; i++) {
+        // Use addr/len from nixlMetaDesc - assumes these are correct for the segment
+        void *laddr = (void*) local[i].addr;
+        size_t lsize = local[i].len;
+        // void *raddr = (void*) remote[i].addr; // Not needed for cost params
```
brminich: we'd need that as well for proper estimation

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+/**
+ * @brief UCP cost estimation parameters field mask.
+ *
+ * The enumeration allows specifying which fields in @ref ucp_ep_cost_params_t are
+ * present. It is used to enable backward compatibility support.
+ */
+enum ucp_ep_cost_params_field {
+    UCP_EP_COST_PARAM_FIELD_BUFFER       = UCS_BIT(0), /**< buffer */
+    UCP_EP_COST_PARAM_FIELD_MESSAGE_SIZE = UCS_BIT(1), /**< message_size */
+    UCP_EP_COST_PARAM_FIELD_OP_EP        = UCS_BIT(2), /**< ep */
```
brminich: ep should be a separate paramater in the API, as it is function performed on an ep

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+/**
+ * @struct ucp_ep_cost_params_t
+ * @brief Parameters for the cost estimation query.
+ */
+typedef struct ucp_ep_cost_params {
+    /** Mask of fields set. */
+    uint64_t field_mask;
+
+    /** Pointer to the local memory region. */
+    void *buffer;
```
brminich: we also need pointer to the remote buffer

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+ *
+ * This function queries the cost estimation for a transfer request based on the provided parameters.
+ *
+ * @param [in] ep             Endpoint handle for the transfer request.
+ * @param [in] cost_params    Parameters for the cost estimation query.
+ * @param [out] cost_estimate Pointer to the cost estimation structure to be populated.
+ *
+ * @return UCS_OK on success or an error code on failure.
+ */
+ucs_status_t ucp_ep_query_cost(const ucp_ep_cost_params_t *cost_params, ucp_ep_cost_t *cost_estimate)
```
```suggestion
ucs_status_t ucp_ep_query_cost(const ucp_ep_cost_params_t *cost_params, ucp_ep_cost_t *cost_attr)
```

brminich: maybe `ucp_ep_estimate_perf`, like in UCT? query_cost also sounds good, let's check others opinion

src/api/cpp/nixl_types.h
iyastreb: One way of implementing backward compatibility that I've implemented before at CERN is a sort of named data type container:
https://cmw.web.cern.ch/rdaJavaAPI/cern/cmw/Data.html, https://cmw.web.cern.ch/rdaCppAPI/cpp_common/html/classrdaData.html

Main goal: allow to exchange values whose types are not fixed at compile time, backward compatibility.
A Data object may contain zero or more DataEntry objects; each entry is uniquely identified by a string tag. The entries can be accessed sequentially or randomly by tags.
Data object may be serialized very efficiently into a byte array (similar to TLV (type-length-value))

Another way is to use extensions - preserve some opaque pointer in the structure that can be implemented later

src/api/cpp/backend/backend_engine.h
```diff
@@ -114,6 +114,26 @@ class nixlBackendEngine {
...
+         *
+         * @param operation      NIXL operation type (NIXL_READ or NIXL_WRITE).
+         * @param local          Local metadata descriptor list.
+         * @param remote         Remote metadata descriptor list.
+         * @param remote_agent   Remote agent name.
+         * @param estimate       Pointer to nixlCostEstimate structure to store the result.
+         * @return nixl_status_t NIXL_SUCCESS if cost estimation is successful,
+         *                       error code on other errors.
+         */
+        virtual nixl_status_t estimateXferCost(const nixl_xfer_op_t &operation,
```
iyastreb, line 119: Do we use exceptions or error codes?

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::estimateXferCost(nixlXferReqH* req_hndl,
+                            nixlCostEstimate* estimate) const
+{
+    NIXL_LOCK_GUARD(data->lock);
```
iyastreb: Can we acquire a lock after NULL pointer check?

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+    }
+
+    // Check if the remote agent connection info is still valid
+    // (assuming cost estimation requires connection info like transfers)
+    if (!req_hndl->remoteAgent.empty() &&
+        (data->remoteSections.count(req_hndl->remoteAgent) == 0)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (!req_hndl->engine) {
```
iyastreb, line 782: maybe do this check before taking the lock too?

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+    NIXL_LOCK_GUARD(data->lock);
+
+    if (!req_hndl || !estimate) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    // Check if the remote agent connection info is still valid
+    // (assuming cost estimation requires connection info like transfers)
+    if (!req_hndl->remoteAgent.empty() &&
+        (data->remoteSections.count(req_hndl->remoteAgent) == 0)) {
```
iyastreb, line 777: I don't see the declaration of `remoteSections`, but it it's a std::map - can we use `contains` methods to check the key presence?

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (!req_hndl->engine) {
+        // Should not happen if req_hndl is valid
+        return NIXL_ERR_UNKNOWN;
+    }
+
+    // Call the backend engine's implementation
+    return req_hndl->engine->estimateXferCost(req_hndl->backendOp,
```
iyastreb, line 787: Is keeping a lock needed while executing this function?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+    }
+
+    if (lcnt == 0) {
+        // Nothing to estimate
+        estimate->duration = 0.0;
+        return NIXL_SUCCESS;
+    }
+
+    // Iterate through descriptors to estimate cost for each pair.
+    for (i = 0; i < lcnt; i++) {
```
iyastreb: maybe declare i here, it's used only inside the loop

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+    nixlUcxPublicMetadata *rmd;
+    double total_duration_sec = 0.0;
+
+    if (lcnt != rcnt) {
+        NIXL_ERROR << "Local (" << lcnt << ") and remote (" << rcnt
+                   << ") descriptor lists differ in size for cost estimation";
+        return NIXL_ERR_MISMATCH;
+    }
+
+    if (lcnt == 0) {
```
iyastreb: this block is redundant

src/api/cpp/backend/backend_engine.h
ColinNV: +1 for exceptions.

src/core/nixl_agent.cpp
ColinNV: Shouldn't the arguments be references, rendering the `NULL` check superfluous?

src/core/nixl_agent.cpp
ColinNV: That's a C++20 function, are we there yet?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+
+    // Iterate through descriptors to estimate cost for each pair.
+    for (i = 0; i < lcnt; i++) {
+        // Use addr/len from nixlMetaDesc - assumes these are correct for the segment
+        void *laddr = (void*) local[i].addr;
+        size_t lsize = local[i].len;
+        // void *raddr = (void*) remote[i].addr; // Not needed for cost params
+        size_t rsize = remote[i].len;
+
+        lmd = (nixlUcxPrivateMetadata*) local[i].metadataP;
```
ColinNV: This should use a C++-style cast as they can be grepped for and don't quite as easily do too much.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
 }
 
+// Estimate transfer cost using ucp_ep_query_cost
+nixl_status_t nixlUcxEngine::estimateXferCost (const nixl_xfer_op_t &operation,
+                                               const nixl_meta_dlist_t &local,
+                                               const nixl_meta_dlist_t &remote,
+                                               const std::string &agent_name,
+                                               nixlCostEstimate* estimate)
+{
+    size_t lcnt = local.descCount();
```
ColinNV: Looks like this, and `rcnt`, can be `const`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+    size_t rcnt = remote.descCount();
+    size_t i;
+    nixl_status_t ret;
+    nixlUcxPrivateMetadata *lmd;
```
ColinNV: The declarations should be moved down to where they are initialized.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+// TODO: these definitions need to appear in UCX not NIXL. Provided for code review purposes.
+
+/**
+ * @enum ucp_op_type_t
+ * @brief UCP operation types.
+ *
+ * This enum defines the types of operations that can be performed using UCP.
+ * It covers all the routines available in the UCP API for transferring data.
+ */
+typedef enum {
```
ColinNV: Should be `enum class`, possibly with a specific size? And the `typedef` is not needed in C++.

src/api/cpp/nixl_types.h
ColinNV: That looks like it would handle every conceivable forward and backward compatibility, albeit by being quite heavyweight... Are we sure we need this much?

iyastreb: Yes, it's quite heavyweight, as it was designed to represent any possible data exchange between client and server (including nested structures). I'm not sure we want it just for the purpose of backward compatibility
However we might consider some lightweight version of it

src/core/nixl_agent.cpp
```diff
@@ -755,6 +755,36 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+    if (!req_hndl->remoteAgent.empty() &&
+        (data->remoteSections.count(req_hndl->remoteAgent) == 0)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (!req_hndl->engine) {
+        // Should not happen if req_hndl is valid
+        return NIXL_ERR_UNKNOWN;
+    }
+
```
w1ldptr, line 786: Body of this function is mostly sanity checks. Some of them are duplicated from postXferReq(), other are superfluous since we create nixlXferReqH and its contents are opaque to the user (so why would req_hndl->engine be NULL, for example, unless it is a bug in our makeXferReq()?).

I suggest to either remove majority of these or extract them out into some 'checkXferSanity' function that could be shared with other nixlAgent methods and allows us to keep the code readable. Even better, nixlXferReqH could be refactored into proper C++ class where all these things would be ctor parameters so it wouldn't be possible to instantiate the object with bogus data since ctor would throw an exception when passed invalid engine pointer or empty remote name.

src/api/cpp/nixl_types.h
```diff
@@ -116,6 +116,18 @@ using nixl_b_params_t = std::unordered_map<std::string, std::string>;
...
+    public:
+        /**
+         * @var duration Estimated transfer duration in seconds.
+         */
+        double duration;
```
w1ldptr: std::chrono::duration?

src/api/cpp/nixl_types.h
w1ldptr: What is the point of these tricks in this particular case? Pimpl approach is sometimes used to preserve ABI when changing layout of _complex opaque_ classes in bugfix releases. This class is trivial and transparent since the implementation exposes the contents directly as public members.

src/core/nixl_agent.cpp
w1ldptr: With the only model currently supported by NIXL (strict) nixlAgent must always take the lock when calling engine callbacks.

src/utils/ucx/ucx_utils.cpp
w1ldptr: IMO for API that returns result in time units 'cost' or 'latency' would be more appropriate than 'perf'.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+    nixlUcxPublicMetadata *rmd;
+    double total_duration_sec = 0.0;
+
+    if (lcnt != rcnt) {
+        NIXL_ERROR << "Local (" << lcnt << ") and remote (" << rcnt
+                   << ") descriptor lists differ in size for cost estimation";
+        return NIXL_ERR_MISMATCH;
+    }
```
w1ldptr, line 998: Move sanity checks to agent so that all engines implementing cost estimation API won't have to duplicate them.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+        // void *raddr = (void*) remote[i].addr; // Not needed for cost params
+        size_t rsize = remote[i].len;
+
+        lmd = (nixlUcxPrivateMetadata*) local[i].metadataP;
+        rmd = (nixlUcxPublicMetadata*) remote[i].metadataP;
+
+        if (!lmd || !rmd) {
+            NIXL_ERROR << "Null metadata found in descriptor lists at index " << i << " during cost estimation";
+            return NIXL_ERR_INVALID_PARAM;
+        }
```
w1ldptr: Isn't this enforced by nixlMemSection::populate()?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,75 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+
+        if (lsize != rsize) {
+             NIXL_ERROR << "Local size (" << lsize << ") != Remote size (" << rsize
+                        << ") at index " << i << " during cost estimation";
+            return NIXL_ERR_INVALID_PARAM; // Should have been caught by agent layer, but check anyway
```
w1ldptr: Why check anyway if it is makeXferReq() responsibility?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,154 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+    UCP_EP_COST_PARAM_FIELD_MESSAGE_SIZE = UCS_BIT(1), /**< message_size */
+    UCP_EP_COST_PARAM_FIELD_OP_EP        = UCS_BIT(2), /**< ep */
+    UCP_EP_COST_PARAM_FIELD_OP_TYPE      = UCS_BIT(3), /**< op_type */
+};
+
+/**
+ * @struct ucp_ep_cost_params_t
+ * @brief Parameters for the cost estimation query.
+ */
+typedef struct ucp_ep_cost_params {
```
w1ldptr: Typedef for structure name is not needed in C++.

src/utils/ucx/ucx_utils.cpp
ovidiusm: https://github.com/openucx/ucx/pull/10672

src/utils/ucx/meson.build
ovidiusm: For linking the logging code

src/api/cpp/nixl_types.h
ovidiusm: Pimpl is overkill here and would impact performance

src/plugins/ucx/meson.build
ovidiusm: We need to link the logging code since nixl_common is a separate shared object

src/api/cpp/backend/backend_engine.h
ovidiusm: Error codes. This API will not be used directly but through python wrappers.

src/api/cpp/nixl_types.h
ovidiusm: The options are:
* version code (simple but a bit heavy documentation, usage not easy without reading the docs  -- e.g. which field is supported in version 3?)
* field mask (similar to version code but more clear and readable code)
* padding (error prone)
* pimpl (perf impact)

=> chose field mask

src/api/cpp/nixl_types.h
ovidiusm: We are middleware between python and UCX

UCX returns double

Python uses double

I would not convert to chrono::duration because we need to do conversion and then conversion back to double, there is no point

src/core/nixl_agent.cpp
ovidiusm: Yes indeed. Better to use refs.

src/core/nixl_agent.cpp
ovidiusm: Thanks for the good suggestions. I simplified the code greatly.
Regarding refactoring.  Right now there is no other comm backend so it's premature to refactor as we don't know yet the requirements => out of scope of this change. So in this PR I will keep the checks in the most specific location where they are used.

src/plugins/ucx/ucx_backend.cpp
ovidiusm: I disagree, the sanity checks are specific to what this implementation does, we do not know what other comm libraries might do

src/plugins/ucx/ucx_backend.cpp
ovidiusm: Better to check for null pointers than to crash

src/utils/ucx/ucx_utils.h
ovidiusm: Addressed as we discussed by passing the mem/rkey handles instead

src/plugins/ucx/ucx_backend.cpp
ovidiusm: working with handles now

src/utils/ucx/ucx_utils.cpp
ovidiusm: Agreed

src/utils/ucx/ucx_utils.cpp
ovidiusm: using handles now

src/utils/ucx/ucx_utils.cpp
ovidiusm: switched to existing UCX API

src/core/nixl_agent.cpp
ovidiusm: switched to references

src/plugins/ucx/ucx_backend.cpp
ovidiusm: even better, use the getters

src/utils/ucx/ucx_utils.cpp
ovidiusm: moved to UCX

src/utils/ucx/ucx_utils.cpp
ovidiusm: moved to UCX

src/plugins/ucx/ucx_backend.cpp
ovidiusm: because if we do not check and they happen to be different we will crash. Better to check and log

src/core/nixl_agent.cpp
ovidiusm: Maybe it would be nice to have but let's keep the older version for compatibility

src/api/cpp/nixl_types.h
w1ldptr: > We are middleware between python and UCX
> 

I'm not sure we are Python middleware yet since we have users in both Rust and C++...

> UCX returns double
> 
> Python uses double

https://docs.python.org/3/library/datetime.html#examples-of-usage-timedelta

> 
> I would not convert to chrono::duration because we need to do conversion and then conversion back to double, there is no point

Use better types appropriate for specific programming language that would make it more convenient for the user instead reducing everything to lowest common denominator of basic C types.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Better to fix makeXferReq() to ensure that it doesn't happen and that all engines benefit from it instead of putting random defensive checks all around our code base.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: If the pointer contains some garbage unexpected value, then you only check for one of 2pow64 potential wrong values while ignoring other 2pow64-1.

Instead of trying to catch NULLs in random places all around the code base ensure that they are not propagated to engines from agent in the first place.

src/api/cpp/nixl_types.h
ovidiusm: std::chrono::duration does not offer ABI compatibility so we cannot use it here

src/api/cpp/backend/backend_engine.h
rakhmets, line 119: Does it make sense to move towards making this class abstract (with only pure virtual functions)? I think this would be better from API perspective. I see that there are already many non-purely virtual functions and class fields here. However, adding another non-purely virtual method takes us a step further from this idea.

src/api/cpp/nixl_types.h
w1ldptr: I'm not sure I'm following. Nixl Agent API already exposes both std templates (like vector and string) and our own template classes like nixlDescList. What exactly makes this new API you are introducing so special? When was the decision regarding "cannot use" made?

From what I remember the issue was discussed on C++ guidelines meeting and it was decided to use rich C++ types in APIs _assuming they are from std or our own_ (i.e. no Abseil). @yosefe ?

src/api/cpp/nixl_types.h
```diff
@@ -116,6 +117,60 @@ using nixl_b_params_t = std::unordered_map<std::string, std::string>;
...
+ */
+enum nixl_cost_estimate_fields_t {
+    NIXL_COST_ESTIMATE_DURATION = 1 << 0,
+};
+
+/**
+ * @class nixlCostEstimate
+ * @brief Cost estimation for a transfer request.
+ */
+class nixlCostEstimate {
```
rakhmets: I don't think this file is a good place for this class. I think we should only keep here aliases, enums, constant expressions, structures.

P.S. nixlAgentOptionalArgs should be a structure. (not related to changes in this PR)

src/api/cpp/backend/backend_engine.h
```diff
@@ -114,6 +114,26 @@ class nixlBackendEngine {
...
+                                               const nixl_meta_dlist_t &local,
+                                               const nixl_meta_dlist_t &remote,
```
rakhmets, line 121: Do we want to provide the ability to pass lists with different sizes? I have the same doubt about postXfer and prepXfer. If not so, then it should be a single list of pairs of local and remote metadates.

src/api/cpp/nixl_types.h
rakhmets: Please use chrono to make unit of time duration explicit. If you need to mention something in comments like it was done for getter and setter (... in seconds ...), then it is a signal to use std::chrono.

Compilers do not read comments. Comments are less precise than code. Comments are not updated as consistently as code. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code

Please see example from this section: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed

std::chrono::seconds should be passed to setter (good), or the setter should be template method (better).

brminich: do not see a strong reason to use chrono if it is going to be used by rust/python code thru the corresponding bindings

src/plugins/ucx/ucx_backend.cpp
brminich: i'd suggest to add NIXL_ASSERT instead of check

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -827,6 +828,63 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+            return NIXL_ERR_INVALID_PARAM;
+        }
+
+        double duration_sec = 0.0;
+        nixl_status_t ret = uw->estimateCost(rmd->conn.ep,    // Remote Endpoint
+                                             lmd->mem,        // Local buffer address
+                                             rmd->rkey,       // Remote buffer address
+                                             lsize,           // Message size
+                                             operation,       // NIXL operation type
+                                             duration_sec);   // Output duration
```
brminich: i'd remove these comments, they do not seem to add value
also rkey is not a buffer address and mem is not a local  address

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,54 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+    }
+    ucp_ep_evaluate_perf_param_t params = {
+        .field_mask = UCP_EP_PERF_PARAM_FIELD_MESSAGE_SIZE |
+                      UCP_EP_PERF_PARAM_FIELD_RKEY         |
+                      UCP_EP_PERF_PARAM_FIELD_MEM_H        |
+                      UCP_EP_PERF_PARAM_FIELD_OP_TYPE,
+        .message_size = size,
+        .rkey         = rk.rkeyh,
+        .mem_h        = mem.memh,
+        .op_type      = ucx_op_type,
```
brminich: minor: can align all fields by `=` as it is already partially done

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,54 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+
+    ucs_status_t status = ucp_ep_evaluate_perf(ep.eph, &params, &cost_result);
+    if (status != UCS_OK) {
+        NIXL_ERROR << "ucp_ep_evaluate_perf failed: " << ucs_status_string(status);
+        return NIXL_ERR_BACKEND;
+    }
+    if (!(cost_result.field_mask & UCP_EP_PERF_ATTR_FIELD_ESTIMATED_TIME)) {
+        NIXL_ERROR << "ucp_ep_evaluate_perf failed: no duration returned";
+        return NIXL_ERR_BACKEND;
+    }
```
brminich: pls remove, it is supposed to be true if status is success

src/utils/ucx/ucx_utils.cpp
```diff
@@ -471,6 +472,54 @@ nixl_status_t nixlUcxWorker::write(nixlUcxEp &ep,
...
+            break;
+        default:
+            NIXL_ERROR << "Unsupported NIXL operation type: " << nixl_op;
+            return NIXL_ERR_INVALID_PARAM;
+    }
+    ucp_ep_evaluate_perf_param_t params = {
+        .field_mask = UCP_EP_PERF_PARAM_FIELD_MESSAGE_SIZE |
+                      UCP_EP_PERF_PARAM_FIELD_RKEY         |
+                      UCP_EP_PERF_PARAM_FIELD_MEM_H        |
+                      UCP_EP_PERF_PARAM_FIELD_OP_TYPE,
```
brminich: these fields will only appear in UCX 1.20, need to handle the case when older UCX is used

src/api/cpp/nixl_types.h
w1ldptr: AFAIK there are C++ users of NIXL.

rakhmets: > do not see a strong reason to use chrono if it is going to be used by rust/python code thru the corresponding bindings

I posted links to the two topics in my previous comment. This is a C++ code. And I recommend to follow C++ guidelines. And the guidelines are pretty simple in this case. You can also find bad examples, and better solutions there. Copied the reasons from the topics below.

- *Make interfaces precisely and strongly typed*
Reason: Types are the simplest and best documentation, improve legibility due to their well-defined meaning, and are checked at compile time. Also, precisely typed code is often optimized better.
- *Don’t say in comments what can be clearly stated in code*
Reason: Compilers do not read comments. Comments are less precise than code. Comments are not updated as consistently as code.

src/api/cpp/backend/backend_engine.h
ovidiusm: Let's discuss and do any major refactoring separately

src/api/cpp/backend/backend_engine.h
ovidiusm: Let's raise this point in the next meeting and discuss separately

src/bindings/python/nixl_bindings.cpp
```diff
@@ -426,6 +426,12 @@ PYBIND11_MODULE(_bindings, m) {
                    py::arg("remote_descs"), py::arg("remote_agent"),
                    py::arg("notif_msg") = std::string(""),
                    py::arg("backend") = std::vector<uintptr_t>({}))
+        .def("estimateXferCost", [](nixlAgent &agent, uintptr_t reqh) -> std::chrono::duration<double> {
+                std::chrono::duration<double> duration;
+                nixl_status_t status = agent.estimateXferCost(reinterpret_cast<nixlXferReqH&>(reqh), duration);
```
brminich: minor: note that seems `ret` is used for `nixl_status_t` in all other funcs

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -855,6 +855,53 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
     return NIXL_SUCCESS;
 }
 
+// Estimate transfer cost using ucp_ep_query_cost
```
brminich: not sure we need this comment, but
```suggestion
// Estimate transfer cost using ucp_ep_evaluate_perf
```

brminich:
@yosefe @mkhazraee can you pls review?

src/api/cpp/nixl.h
```diff
@@ -231,6 +233,17 @@ class nixlAgent {
...
 
+        /**
+         * @brief Estimate the cost (e.g., duration) of executing a transfer request.
+         *
+         * @param req_hndl       Transfer request handle
+         * @param estimate [out] Estimated duration of the transfer
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        estimateXferCost(const nixlXferReqH &req_hndl,
```
mkhazraee: I think just the duration is not sufficient, we need some level of certainty. For example this number is analytical, coming from some better level of estimation, coming from telemetry done by the agent, coming from telemetry done by the backend, or some even more advanced method to do the estimation.

One other thing is if we need to provide a standard deviation from this result. Specially if we consider the telemetry cases which are more representative than analytical models, not sure if there is a better method than std, @brminich any thoughts?

src/api/cpp/nixl.h
```diff
@@ -231,6 +233,17 @@ class nixlAgent {
...
+                         std::chrono::duration<double> &duration) const;
```
mkhazraee: I suggest to output a ns value. u64 of ns is way more than enough, no need to have std::chrono in output. or if ns is not practical, us. That's what we have in nixl_time too.

src/api/cpp/nixl.h
mkhazraee: We can start with the analytical model and give 0 as std output, and some value, maybe 0, for the confidence. Just to have the capability in the API. Not sure for the confidence we want to do a float, or define some levels as enum, more fan of the latter, defining the float can be challenging and get into the nuances that comparing values becomes meaningless.

src/api/cpp/nixl.h
yosefe: IMO std::chrono::duration is a standard C++ way for this, and should be used (or replace) nixlTime, there is no need to reimplement in nixl system services provided by C++ std lib..

rakhmets: As I mentioned in another comment the interfaces should be precisely and strongly typed. And to avoid problems and discussions about units of measurement, we can do the following:
```C++
template<class rep, class period>   
nixl_status_t estimateXferCost(const nixlXferReqH &req_hndl,
                               std::chrono::duration<rep, period> &duration) const
{
    // std::chrono::duration_cast<std::chrono::duration<rep, period>>(value);
}
```
and the interface will accept any time duration unit.
```C++
std::chrono::microseconds us;
auto status = nixlAgent::getEstimateXferCost(h, us);

std::chrono::nanoseconds ns;
status = nixlAgent::getEstimateXferCost(h, ns);
```

rakhmets: Another option is to mention units explicitly, e.g.:
```C++
nixl_status_t
estimateXferCost(const nixlXferReqH &req_hndl,
                 std::chrono::microseconds &duration) const;
```

ovidiusm: Thanks, let's go with `std::chrono::microseconds`

src/core/nixl_agent.cpp
```diff
@@ -767,7 +767,8 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
 
 nixl_status_t
 nixlAgent::estimateXferCost(const nixlXferReqH &req_hndl,
-                            std::chrono::microseconds &duration) const
+                            std::chrono::microseconds &duration,
+                            const nixl_opt_args_t* extra_params) const
```
ColinNV, line 770: Could we use `std::optional< nixl_opt_args_t >` for the optional parameters?

brminich: currently it follows the same approach as other APIs, maybe better keep it and then change all APIs at once if we decide to do so

src/api/cpp/nixl.h
ovidiusm: It may make sense to expose the error in the API when we have a better understanding of how it will be implemented, because it is currently not clear which option we will choose

ovidiusm: Added an error parameter as discussed offline

src/api/cpp/nixl_types.h
mkhazraee: I come from a different angle to this, at the end of the day user cares about this is a micro second value, and we can simply say duration_us and it would be the same level of being clear than saying std::chrono::microseconds (if you really want to be verbose, say xfer_time_microseconds even). And introducing a new std::chrono might be confusing for the users, specially Agent users. If you're a plugin developer, then we can be more clear and specify these.

We had a similar conversation for pthrDelay back in the days. What we did was to use u64 for the NB, so it's simple for the user (and also what Mikhail said), and then for SB and plugins we did nixlTime::us_t pthrDelay; That ways if there is any compiler checks that can be done as intended (actually nixlTime::us_t is just u64, but placeholder if we wanted it to be different).

w1ldptr: Actually, pthrDelay is a great example why we _shouldn't do this_ in the API. I wanted to change it during the progress thread refactoring to MS (since it is what POSIX poll() API accepts) and the code just continued to compile after the change while being completely incorrect because they all just aliases for uint64_t. While with std::chrono::duration I would get something like:

`error: no match for 'operator=' (operand types are 'std::chrono::milliseconds' {aka 'std::chrono::duration<long int, std::ratio<1, 1000> >'} and 'std::chrono::microseconds' {aka 'std::chrono::duration<long int, std::ratio<1, 1000000> >'})`

On top of that [user gets](https://en.cppreference.com/w/cpp/chrono/duration) nice QoL features. What exactly is "confusing" about any of that?

w1ldptr: On top of that I don't quite get why we have this discussion yet again when the issue of "random C-like vs modern C++ API" has been addressed and decision on this has been made.

mkhazraee: BTW regarding std::vector<int> and std::string they're the only use cases we have std. For string it's just the string in C, so it was fine. For vector we had the conversation that this is equal to a list in python. 

We were even thinking of even making it a int list and a len, but it was more confusing and 1 extra argument, or introducing a new type for nixl with this without adding extra param. All these while std::vector is well know and pybind supports it, didn't make sense. So for each of them we had a conversation. 

Now for std::chrono::microsecond, it's just a number, doing that change doesn't change number of parameters or clarity, while it's less common. And regarding compiler checks, having the conversation between NB and SB provides that.

w1ldptr: > BTW regarding std::vector and std::string they're the only use cases we have std. For string it's just the string in C, so it was fine. For vector we had the conversation that this is equal to a list in python.

First of all, it is not since I also see unordered_maps passed around. Second of all, I don't see how this consideration change the current direction towards rich C++ API.

> We were even thinking of even making it a int list and a len, but it was more confusing and 1 extra argument, or introducing a new type for nixl with this without adding extra param. All these while std::vector is well know and pybind supports it, didn't make sense. So for each of them we had a conversation.

I think one of the motivating factors for high-level decision towards C++ API was specifically _not_ having such conversations in every single PR.

> Now for std::chrono::microsecond, it's just a number, doing that change doesn't change number of parameters or clarity, while it's less common.

It sure changes clarity and readability of code when I see literals like '2ms' (i.e. not "just a number") passed somewhere as opposed to random double like '2.0' (because it is indeed "just a number" and I would have to open the function declaration to read that the random double constant is passed as parameters named 'something_ms'). Passing reference to instance of 'std::chrono::milliseconds duration' is also clearer than a reference to 'double duration'.

> And regarding compiler checks, having the conversation between NB and SB provides that.

Can you clarify how exactly are you going to compiler-check someone passing nixlTime::us_t where nixlTime::ms_t is expected?

ovidiusm: As discussed offline, I have added a _us suffix to make it obvious that what the unit is, and switched to signed int since that is the unit used consistently in other parts of the code.

Also added an enum field that describes how the estimate is computed. The caller may use this info to decide how to use the estimate.

Overall we have 3 output params, and we considered moving them to a struct, but the code is slightly more readable as it is and it also forces the client to fix the API in case of any further changes.

The `extra_params` field may be desired in the future, but there is no usecase now, so I will remove it.

mkhazraee: That error is exactly why it's less user friendly. Dealing with integers versus dealing with chrono and extra checks it has. For case of pthrDelay, user just wants to indicate a number, not deal with time. Here user wants to get a number to decide to use NIXL for transfer or not. Simple, you get a us value, not have to deal with this is ms or us or ns or any of that. Those printing capabilities don't bring any benefit, while those limitations adds more hassle and confusion for the user.

Regarding pthrDelay, if for the backend you want to do chrono instead of nixlTime, that's fine as I mentioned. In our talks we concluded for backend implementation that someone savvy or ourselves are developing, we follow C++ guidelines. For NB, we keep discussions as simplicity is of utmost importance here.

w1ldptr: I'm not entirely convinced that dismissing my and Raul's suggestions as "confusing" without any meaningful explanation is constructive and contribute meaningfully. As is just unilaterally marking all suggestions you disagree with as "resolved" BTW...

w1ldptr: > That error is exactly why it's less user friendly.

WDYM? Compiler checks are definitely a good and user-friendly feature of any API.

> Dealing with integers versus dealing with chrono and extra checks it has.

You are not _dealing_ with checks, they _help_ you write correct and self-explanatory code.

> For case of pthrDelay, user just wants to indicate a number, not deal with time.

I have to disagree. As a user I _definitely_ want to indicate a time unit when setting an option like 'pthrDelay', not some random double like '0.123' whatever that may be.

> Those printing capabilities don't bring any benefit, while those limitations adds more hassle and confusion for the user.

I just don't see why having strong compiler checks and QoL features add more hassle and confusion, sorry.

ColinNV: When used well the C++ type system can go a long way to "if it compiles it'll be correct", which is a layer of bug prevention that IMO should be used as much as possible.

That means choosing the most specialized type available (and ging beyond by creating new, dedicated "simple" types to make conversions explicit instead of sticking everything in, say, a `std::string`.)

In this case we should definitely use `std::chrono` and let the language help us prevent an entire class of bugs.

ColinNV: > do not see a strong reason to use chrono if it is going to be used by rust/python code thru the corresponding bindings

Thinking about this again, one never knows where it might be used from in the future, therefore I'd strongly argue to not compromise on the C++ side of things.

src/api/cpp/nixl_types.h
```diff
@@ -134,6 +135,15 @@ extern const std::string default_metadata_label;
  */
 extern const std::string default_partial_metadata_label;
 
+
+/**
+ * @enum nixl_cost_estimate_t
+ * @brief An enumeration of cost estimation types for transfer cost estimation.
+ */
+enum nixl_cost_estimate_t {
+    NIXL_COST_ESTIMATE_UCX_EP = 0, // UCX endpoint-based cost estimate
```
mkhazraee: We shouldn't make it be backend specific. Something like NIXL_COST_ANALYTICAL_BACKEND might be a better start.

.pre-commit-config.yaml
```diff
@@ -49,7 +49,7 @@ repos:
     rev: v2.2.4
     hooks:
       - id: codespell
-        args: ["--toml", "pyproject.toml"]
+        args: ["--toml", "pyproject.toml", "--ignore-words-list=crate"]
```
brminich: is it needed after #327?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -426,6 +426,14 @@ PYBIND11_MODULE(_bindings, m) {
                    py::arg("remote_descs"), py::arg("remote_agent"),
                    py::arg("notif_msg") = std::string(""),
                    py::arg("backend") = std::vector<uintptr_t>({}))
+        .def("estimateXferCost", [](nixlAgent &agent, uintptr_t reqh) -> std::tuple<double, double, int> {
```
mkhazraee: I suggest to be consistent across all 3 APIs, usec as int, not double only in python.

src/api/cpp/backend/backend_engine.h
```diff
@@ -114,6 +114,20 @@ class nixlBackendEngine {
                                         const nixl_opt_b_args_t* opt_args=nullptr
                                        ) const = 0;
 
+        // Estimate the cost (duration) of a transfer operation.
+        virtual nixl_status_t estimateXferCost(const nixl_xfer_op_t &operation,
+                                               const nixl_meta_dlist_t &local,
+                                               const nixl_meta_dlist_t &remote,
+                                               const std::string &remote_agent,
+                                               nixlBackendReqH* const &handle,
```
mkhazraee, line 123: I understand why we did this to be aligned with postXfer, maybe in future we do an API change that post and get cost only have a single input of (nixlBackendReqH* const &handle), that aligns better with NB. Initially we didn't have the prep call, and didn't want to touch the implementation. But maybe something good to do.

src/api/cpp/nixl_types.h
rakhmets: >That error is exactly why it's less user friendly. Dealing with integers versus dealing with chrono and extra checks it has. For case of pthrDelay, user just wants to indicate a number, not deal with time. Here user wants to get a number to decide to use NIXL for transfer or not. Simple, you get a us value, not have to deal with this is ms or us or ns or any of that. Those printing capabilities don't bring any benefit, while those limitations adds more hassle and confusion for the user.

>Regarding pthrDelay, if for the backend you want to do chrono instead of nixlTime, that's fine as I mentioned. In our talks we concluded for backend implementation that someone savvy or ourselves are developing, we follow C++ guidelines. For NB, we keep discussions as simplicity is of utmost importance here.

I cannot agree with this statement.

This is exactly the same problem that is described here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed

I have already copied quotes from this topic in other comments.
This interface corresponds to the example that is marked as 'bad'.
```C++
nixl_status_t estimateXferCost(int64_t &duration_us) const;
```
, because using the interface leaves ambiguities:
```C++
int64_t xyz;
auto status = nixlAgent->estimateXferCost(xyz);
```

And I have already given an example of an interface that can correspond to the 'Example, good' section.
```C++
template<class rep, class period>   
nixl_status_t estimateXferCost(std::chrono::duration<rep, period> &duration) const {
    //...
}
```
, with the following usage in clients code:
```C++
std::chrono::microseconds xyz;
auto status = nixlAgent->estimateXferCost(xyz);
```
, or:
```C++
std::chrono::nanoseconds xyz;
auto status = nixlAgent->estimateXferCost(xyz);
```
, etc.

And invalid usage will be caught at compile time.

src/core/nixl_agent.cpp
```diff
@@ -762,6 +762,39 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
 }
 
+nixl_status_t
+nixlAgent::estimateXferCost(const nixlXferReqH *req_hndl,
+                            std::chrono::microseconds &duration,
+                            std::chrono::microseconds &err_margin,
+                            nixl_cost_estimate_t &source,
+                            const nixl_opt_args_t* extra_params) const
+{
+    NIXL_LOCK_GUARD(data->lock);
```
brminich: should it be `NIXL_SHARED_LOCK_GUARD(data->lock);` now?
(note it was merged recently)

src/api/cpp/nixl.h
```diff
@@ -231,6 +232,23 @@ class nixlAgent {
...
+         * @param err_margin   [out] Estimated error margin of the transfer
+         * @param source       [out] Source of the cost estimate
+         * @param extra_params Optional extra parameters
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        estimateXferCost(const nixlXferReqH* req_hndl,
+                         std::chrono::microseconds &duration,
+                         std::chrono::microseconds &err_margin,
+                         nixl_cost_estimate_t &source,
```
mkhazraee: I see it's source of the cost model, but source being is an overloaded term. @brminich any other ideas? I can't come up with anything from top of my mind, just throwing out there in case you had other candidates.

mkhazraee (approved):

src/api/cpp/nixl.h
brminich: name is ok imo, i just want to check what will be the behavior when some backend supports different methods for cost estimation? do we want to allow user to request a particular method? If not, how is this source value going to be used?

yosefe: IMO it should be removed, it's unclear, we could add source as input parameter later on in "extra_params"

yosefe (commented):
LGTM except removing "source" and "endpoint" from api

src/api/cpp/nixl.h
mkhazraee: It's an output not input. It's just indicating what was the output mode. If a backend supports more, it would return the most accurate one. For example, maybe a backend needs more telemetry before it can provide a more accurate estimate, so until that point it gives the lower accurate output.

brminich: would not err_margin be enough? How is this output parameter supposed to be used?

mkhazraee: Overall this is part of the confidence level, is it an analytical model, or based on telemetry or somewhere in the middle.

Another aspect is when we add telemetry to agent, if a backend doesn't provide any type of estimation, we can use the telemetry data, and we indicate "where it's coming from and what's its mode" with this enum. Also let's say we end up supporting multiple type of err_margins, not sure if it happens just keeping the option open, we can use this flag to indicate that too.

mkhazraee: I hear you, but let's say an analytical model can also provide a std variation, but that's not the same as telemetry based. And there was an idea maybe we provide p99 or something, that needs an indication too.

brminich: how can such info be used? E.g. if backends return different values obtained using different methods? How will it be compared?

mkhazraee: The problem is we can't expect all backends to be able to give same level of precision, considering we have storage backends too. Also doing a confidence level in form of a double, from 0 to 1, exactly causes the issue of how to compare across different backends. So I was thinking of having a limited set, and backends say which of this limited set they're using as their confidence level. Confidence level is not quite the same as err_margin.

So what I had in mind was to have 2-3 main levels, for example analytical, heuristic, telemetry, and then if it's coming from Agent or backend. (And then the idea of maybe we have different mode for the margin, we can reuse, but this is an if case, for now let's say all is std variation).

brminich: ok

mkhazraee: Talking more with Mikhail, let's change source to mode or method or something like that. Source is confusing. Maybe we can do 
nixl_cost_estimate_t &source --> nixl_cost_t &estimate_mode or nixl_cost_t &estimate_method
As it's estimation mode/method at the end of the day, and nixl has a cost API, and it's cost_t being this sounds reasonable IMO.

src/api/cpp/nixl_types.h
```diff
@@ -135,6 +135,15 @@ extern const std::string default_metadata_label;
  */
 extern const std::string default_partial_metadata_label;
 
+
+/**
+ * @enum nixl_cost_estimate_t
+ * @brief An enumeration of cost estimation types for transfer cost estimation.
+ */
+enum nixl_cost_estimate_t {
+    NIXL_COST_ESTIMATE_HEURISTIC_EP = 0, // Endpoint-based heuristic cost estimate
```
mkhazraee: Maybe NIXL_COST_ANALYTICAL_BKND not to bring EP here, and also ESTIMATE sounds redundant, and you mentioned it's more ANALYTICAL than a HEURISTIC that gets updated.

brminich (approved):

PR #252, opened 2025-05-05
UCX: Only send request notification when requests have completed

tvegas1:
Flush does not imply fencing, so transfer target can end up being notified even before the transfer has completed. This is particularly visible on platforms with weaker ordering (RMA/AM) and with more lanes, as AM might not even be one of the RMA lanes. Perf seen in nixlbench does not seem affected (DRAM and VRAM) by the fix.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -250,17 +250,9 @@ private:
     std::optional<std::pair<std::string,std::string>> notif;
 
 public:
-    void notif_schedule(const std::string& agent,
-                        const std::string& message) {
-        notif = std::make_pair(agent, message);
+    auto& notification() {
+        return notif;
     }
-
-    std::pair<const std::string&,const std::string&> notif_args() const {
-        return {notif.value().first, notif.value().second};
-    }
-
-    bool notif_pending() const { return notif.has_value(); }
-    void notif_consume() { notif.reset(); }
 
     nixlUcxBackendH(nixlUcxWorker* _uw){
         uw = _uw;
@@ -907,9 +899,15 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
     if (opt_args && opt_args->hasNotif) {
         if (ret == NIXL_SUCCESS) {
             ret = notifSendPriv(remote_agent, opt_args->notifMsg, req);
-            _retHelper(ret, intHandle, req);
+            if (_retHelper(ret, intHandle, req)) {
+                return ret;
+            }
+
+            if (ret == NIXL_IN_PROG) {
+                ret = intHandle->status();
+            }
         } else if (ret == NIXL_IN_PROG) {
-            intHandle->notif_schedule(remote_agent, opt_args->notifMsg);
+            intHandle->notification().emplace(remote_agent, opt_args->notifMsg);
         }
     }
 
@@ -919,14 +917,20 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
 nixl_status_t nixlUcxEngine::checkXfer (nixlBackendReqH* handle)
 {
     nixlUcxBackendH *intHandle = (nixlUcxBackendH *)handle;
-    nixl_status_t status = intHandle->status();
 
-    if (status == NIXL_SUCCESS && intHandle->notif_pending()) {
+    nixl_status_t status = intHandle->status();
+    auto& notif = intHandle->notification();
+    if (status == NIXL_SUCCESS && notif.has_value()) {
         nixlUcxReq req;
-        const auto& [remote_agent, msg] = intHandle->notif_args();
-        status = notifSendPriv(remote_agent, msg, req);
-        intHandle->notif_consume();
-        _retHelper(status, intHandle, req);
+        status = notifSendPriv(notif.value().first, notif.value().second, req);
+        notif.reset();
+        if (_retHelper(status, intHandle, req)) {
+            return status;
+        }
+
+        if (status == NIXL_IN_PROG) {
+            status = intHandle->status();
+        }
     }
 
     return status;
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -247,7 +247,13 @@ private:
     nixlUcxWorker* uw;
 
     // Notification to be sent after completion of all requests
-    std::optional<std::pair<std::string,std::string>> notif;
+    struct Notif {
+	    std::string agent;
+	    nixl_blob_t payload;
+	    Notif(const std::string& remote_agent, const nixl_blob_t& msg)
+		    : agent(remote_agent), payload(msg) {}
+    };
+    std::optional<Notif> notif;
 
 public:
     auto& notification() {
@@ -903,9 +909,7 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
                 return ret;
             }
 
-            if (ret == NIXL_IN_PROG) {
-                ret = intHandle->status();
-            }
+            ret = intHandle->status();
         } else if (ret == NIXL_IN_PROG) {
             intHandle->notification().emplace(remote_agent, opt_args->notifMsg);
         }
@@ -922,15 +926,13 @@ nixl_status_t nixlUcxEngine::checkXfer (nixlBackendReqH* handle)
     auto& notif = intHandle->notification();
     if (status == NIXL_SUCCESS && notif.has_value()) {
         nixlUcxReq req;
-        status = notifSendPriv(notif.value().first, notif.value().second, req);
+        status = notifSendPriv(notif->agent, notif->payload, req);
         notif.reset();
         if (_retHelper(status, intHandle, req)) {
             return status;
         }
 
-        if (status == NIXL_IN_PROG) {
-            status = intHandle->status();
-        }
+        status = intHandle->status();
     }
 
     return status;
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -17,6 +17,8 @@
 #include "ucx_backend.h"
 #include "serdes/serdes.h"
 
+#include <optional>
+
 #ifdef HAVE_CUDA
 
 #include <cuda_runtime.h>
@@ -244,7 +246,21 @@ private:
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    std::optional<std::pair<std::string,std::string>> notif;
+
 public:
+    void notif_schedule(const std::string& agent,
+                        const std::string& message) {
+        notif = std::make_pair(agent, message);
+    }
+
+    std::pair<const std::string&,const std::string&> notif_args() const {
+        return {notif.value().first, notif.value().second};
+    }
+
+    bool notif_pending() const { return notif.has_value(); }
+    void notif_consume() { notif.reset(); }
 
     nixlUcxBackendH(nixlUcxWorker* _uw){
         uw = _uw;
@@ -861,8 +877,6 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
             return NIXL_ERR_INVALID_PARAM;
         }
 
-        // TODO: remote_agent and msg should be cached in nixlUCxReq or another way
-
         switch (operation) {
         case NIXL_READ:
             ret = uw->read(rmd->conn.ep, (uint64_t) raddr, rmd->rkey, laddr, lmd->mem, lsize, req);
@@ -879,27 +893,43 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
         }
     }
 
+    /*
+     * Flush keeps intHandle non-empty until the operation is actually
+     * completed, which can happen after local requests completion.
+     */
     rmd = (nixlUcxPublicMetadata*) remote[0].metadataP;
     ret = uw->flushEp(rmd->conn.ep, req);
     if (_retHelper(ret, intHandle, req)) {
         return ret;
     }
 
-    if(opt_args && opt_args->hasNotif) {
-        ret = notifSendPriv(remote_agent, opt_args->notifMsg, req);
-        if (_retHelper(ret, intHandle, req)) {
-            return ret;
+    ret = intHandle->status();
+    if (opt_args && opt_args->hasNotif) {
+        if (ret == NIXL_SUCCESS) {
+            ret = notifSendPriv(remote_agent, opt_args->notifMsg, req);
+            _retHelper(ret, intHandle, req);
+        } else if (ret == NIXL_IN_PROG) {
+            intHandle->notif_schedule(remote_agent, opt_args->notifMsg);
         }
     }
 
-    return intHandle->status();
+    return ret;
 }
 
 nixl_status_t nixlUcxEngine::checkXfer (nixlBackendReqH* handle)
 {
     nixlUcxBackendH *intHandle = (nixlUcxBackendH *)handle;
+    nixl_status_t status = intHandle->status();
 
-    return intHandle->status();
+    if (status == NIXL_SUCCESS && intHandle->notif_pending()) {
+        nixlUcxReq req;
+        const auto& [remote_agent, msg] = intHandle->notif_args();
+        status = notifSendPriv(remote_agent, msg, req);
+        intHandle->notif_consume();
+        _retHelper(status, intHandle, req);
+    }
+
+    return status;
 }
 
 nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle)


artpol84 (approved):
Thank you, @tvegas1.
Looks good to me

artpol84:
@brminich FWIW this doesn't fix the issue I observe with UCX 1.19 and intra-agent (loopback) transfer.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,50 @@ class nixlUcxBackendH : public nixlBackendReqH {
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    struct Notif {
```
yosefe, line 250: why need another subclass for Notif? can we make it part of the request handle directly?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -325,10 +359,39 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+{
+    nixl_status_t ret;
+    nixlUcxReq req;
+
+    if (!has_notification) {
+        return nullptr;
+    }
+
+    ret = engine->notifSendPriv(remote_agent, msg, req);
+    if (ret == NIXL_SUCCESS) {
```
yosefe: can we use _retHelper?

yosefe (commented):
Can we add CI test for this?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -325,10 +359,39 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+        req = nullptr;
+    } else if (ret != NIXL_IN_PROG) {
+        std::cerr << "Error: push notification send returned: " << ret;
```
brminich: pls use NIXL logging

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

tvegas1:
> Can we add CI test for this?

probably needs #246

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    void notif_clear() {
+        has_notification = false;
+        engine = NULL;
+        remote_agent.clear();
+        msg.clear();
+    }
+
 public:
+    void notif_schedule(nixlUcxEngine* ucx_engine,
+                        std::string agent, std::string message) {
```
yosefe: const std::string &

src/plugins/ucx/ucx_backend.h
```diff
@@ -224,6 +224,8 @@ class nixlUcxEngine : public nixlBackendEngine {
         //public function for UCX worker to mark connections as connected
         nixl_status_t checkConn(const std::string &remote_agent);
         nixl_status_t endConn(const std::string &remote_agent);
+
+        friend class nixlUcxBackendH;
```
yosefe: why needed?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    bool has_notification;
+    nixlUcxEngine* engine;
+    std::string remote_agent;
+    std::string msg;
+
+    void notif_clear() {
+        has_notification = false;
+        engine = NULL;
+        remote_agent.clear();
+        msg.clear();
```
yosefe: why needed?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
     nixlUcxBackendH(nixlUcxWorker* _uw){
         uw = _uw;
+        notif_clear();
     }
 
     void append(nixlUcxIntReq *req) {
         head.link(req);
     }
 
+    bool is_empty() {
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+        return head.next() == NULL;
```
rakhmets: nullptr

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    bool has_notification;
+    nixlUcxEngine* engine;
```
rakhmets: std::shared_ptr

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+
+    void notif_clear() {
+        has_notification = false;
+        engine = NULL;
+        remote_agent.clear();
+        msg.clear();
+    }
+
 public:
+    void notif_schedule(nixlUcxEngine* ucx_engine,
```
rakhmets: std::shared_ptr

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
 
+    // Notification to be sent after completion of all requests
+    bool has_notification;
+    nixlUcxEngine* engine;
+    std::string remote_agent;
+    std::string msg;
+
+    void notif_clear() {
+        has_notification = false;
+        engine = NULL;
```
rakhmets: nullptr

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    bool has_notification;
```
```C++
bool has_notification() const {
    return engine != nullptr;
}
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,16 +244,41 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    nixlUcxEngine* engine;
+    std::string remote_agent;
+    std::string msg;
+
+    void notif_clear() {
```
rakhmets: Can be replaced by `engine = nullptr;`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -279,55 +304,81 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-                        out_ret = NIXL_IN_PROG;
-                        break;
-                    default:
-                        /* Any other ret value is ERR and will be returned */
-                        return ret;
-                }
+nixl_status_t nixlUcxBackendH::status()
+{
+    nixlUcxIntReq *req = head.next();
+    nixl_status_t out_ret = NIXL_SUCCESS;
```
```suggestion
    auto out_ret = NIXL_SUCCESS;
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -279,55 +304,81 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+
+    if (NULL == req) {
```
rakhmets: nullptr

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -279,55 +304,81 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+        /* No pending transmissions */
+        return NIXL_SUCCESS;
+    }
+
+    /* Go over all request updating their status */
+    while(req) {
+        nixl_status_t ret;
+        if (!req->is_complete()) {
+            ret = uw->test((nixlUcxReq)req);
```
rakhmets: Please avoid C-style cast.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -279,55 +304,81 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-            }
-            req = next_req;
+    /* Remove completed requests keeping the first one as
+       request representative */
+    req = head.unlink();
+    while(req) {
+        nixlUcxIntReq *next_req = req->unlink();
+        if (req->is_complete()) {
+            _internalRequestReset(req);
+            uw->reqRelease((nixlUcxReq)req);
```
rakhmets: Please avoid C-style cast.

src/plugins/ucx/ucx_backend.h
tvegas1: to do `engine->notifSendPriv()` inside `handle->status()`

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to take as a separate rework since there are other cases

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to do separate rework for that, as there are many instances.

src/plugins/ucx/ucx_backend.cpp
tvegas1: this is existing, suggest to take it separately as for shared_ptr / nullptr and there are other instances.

src/plugins/ucx/ucx_backend.cpp
tvegas1: save memory

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to take it separately

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to do separate rework for that, as there are many instances.

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to do separate rework for that, it is existing code and there are many instances.

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to do separate rework for that, as there are many instances.

src/plugins/ucx/ucx_backend.cpp
tvegas1: suggest to do separate rework for that, as there are many instances.

src/plugins/ucx/ucx_backend.cpp
rakhmets: You can just replace `NULL` by `nullptr` in new added code. It won't require any additional changes.

src/plugins/ucx/ucx_backend.h
rakhmets: These two classes don't look like they should be coupled.
It seems that `nixlUcxEngine` should send notification depending on returning value of `nixlUcxBackendH::status()`.

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.h
tvegas1: fixed, but please note #257 seems to be adding `nixlUcxEngine &eng;` to `nixlUcxBackendH` anyways.

rakhmets: Thanks! My comment was mostly about making a friend class. Because it reduces encapsulation. I don't mind adding a class as a filed if it's really necessary. However, I don't think it's necessary to do this even in that PR.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -879,27 +893,43 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
...
+    if (status == NIXL_SUCCESS && intHandle->notif_pending()) {
+        nixlUcxReq req;
+        const auto& [remote_agent, msg] = intHandle->notif_args();
+        status = notifSendPriv(remote_agent, msg, req);
+        intHandle->notif_consume();
+        _retHelper(status, intHandle, req);
+    }
 
-    return intHandle->status();
+    return status;
```
brminich, line 938: shoule not we return `intHandle->status()` here, if we added notif request to the handle?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,7 +246,21 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    std::optional<std::pair<std::string,std::string>> notif;
+
 public:
+    void notif_schedule(const std::string& agent,
+                        const std::string& message) {
+        notif = std::make_pair(agent, message);
+    }
+
+    std::pair<const std::string&,const std::string&> notif_args() const {
+        return {notif.value().first, notif.value().second};
```
brminich: it will throw an exception if notif is not set, maybe return special value in this case instead?

rakhmets: value_or can be used here

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,7 +246,21 @@ class nixlUcxBackendH : public nixlBackendReqH {
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    std::optional<std::pair<std::string,std::string>> notif;
+
 public:
+    void notif_schedule(const std::string& agent,
+                        const std::string& message) {
+        notif = std::make_pair(agent, message);
```
rakhmets: please use emplace

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -879,27 +893,43 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
...
-        ret = notifSendPriv(remote_agent, opt_args->notifMsg, req);
-        if (_retHelper(ret, intHandle, req)) {
-            return ret;
+    ret = intHandle->status();
+    if (opt_args && opt_args->hasNotif) {
+        if (ret == NIXL_SUCCESS) {
+            ret = notifSendPriv(remote_agent, opt_args->notifMsg, req);
+            _retHelper(ret, intHandle, req);
+        } else if (ret == NIXL_IN_PROG) {
+            intHandle->notif_schedule(remote_agent, opt_args->notifMsg);
```
rakhmets: I don't think it is correct to do it in this way.
`nixlUcxEngine::postXfer` has two independent parameters: `const std::string &remote_agent` and `nixlBackendReqH* &handle`. There is no limitation to call the method with two different agents, with the same handle. The first agent may not receive the notification in this case.

brminich: i do not think it is supported. Also it is existing logic

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -244,7 +246,13 @@ class nixlUcxBackendH : public nixlBackendReqH {
     nixlUcxIntReq head;
     nixlUcxWorker* uw;
 
+    // Notification to be sent after completion of all requests
+    std::optional<std::pair<std::string,std::string>> notif;
```
w1ldptr: Any particular reason to make this std::pair instead of a structure with descriptive field names so that the reader could easily understand that it is the remote agent name and a notification message? Especially when both fields have the same type so it is very easy to make mistakes that compiler wouldn't catch by using 'first' instead of 'second' or vice versa.

tvegas1: added structure

src/plugins/ucx/ucx_backend.cpp
tvegas1: should be fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: readded to carry the notifications in the optional

src/plugins/ucx/ucx_backend.cpp
tvegas1: should be fixed

brminich (approved):

src/plugins/ucx/ucx_backend.cpp
artpol84: This is an internal API, so I'd expect that NIXL core should handle that.
De-facto you are limited to use only one target on a handler at the time.

rakhmets: > i do not think it is supported. Also it is existing logic

The issue with the new logic. However, if it's not supported, then this is not an issue.

rakhmets: >This is an internal API, so I'd expect that NIXL core should handle that.
De-facto you are limited to use only one target on a handler at the time.

I see. Then it seems to be ok to store `remote_agent` in `intHandle`.

PR #254, opened 2025-05-05
pytest: add more fixtures and unit tests for API features

tstamler:
Explore some extra pytest features and edge cases inside the API, with two API fixes:

The op lookup in initializing a transfer would return throw a KeyError exception, so there's no point in following it with an if statement.

Putting an empty nixlList of the wrong type into get_xfer_descs or get_reg_descs would create a seg fault, because we are dereferencing the first item in an empty list. By checking for both of these first, we can avoid this.

### DIFF
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -371,26 +371,22 @@ class nixl_agent:
         skip_desc_merge: bool = False,
     ) -> nixl_xfer_handle:
         op = self.nixl_ops[operation]
-        if op:
-            handle_list = []
-            for backend_string in backends:
-                handle_list.append(self.backends[backend_string])
+        handle_list = []
+        for backend_string in backends:
+            handle_list.append(self.backends[backend_string])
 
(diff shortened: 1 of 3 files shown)


tstamler:
For whoever reviews this, consider for the unit tests if what the current API is returning (KeyError, empty list, nixlError, etc.) is what we actually would want it to return.

aranadive (approved):

src/api/python/_api.py
```diff
@@ -371,26 +371,22 @@ def make_prepped_xfer(
...
-                op,
-                local_xfer_side,
-                local_indices,
-                remote_xfer_side,
-                remote_indices,
-                notif_msg,
-                handle_list,
-                skip_desc_merge,
-            )
+        handle_list = []
```
mkhazraee, line 374: the if op was in case users passed a bad input, to error out early. We handle it as an exception somewhere?

mkhazraee (approved):

PR #256, opened 2025-05-06
UCX: Fix potential memory leak with genNotif()

tvegas1:
In UCX plugin, requests are always kept constructed. When marked completed, they are always destructed/constructed before release. For `genNotif()`, they are returned to UCX with a valid `amBuffer`. When UCX provides it again to the user, `amBuffer` is not used anymore but still allocated, so we need to free it before overwriting it.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -198,21 +198,13 @@ class nixlUcxIntReq : public nixlLinkElem<nixlUcxIntReq> {
     private:
         int _completed;
     public:
-        std::string *amBuffer;
+        std::unique_ptr<std::string> amBuffer;
 
         nixlUcxIntReq() : nixlLinkElem() {
             _completed = 0;
-            amBuffer = NULL;
         }
 
-        ~nixlUcxIntReq() {
-            _completed = 0;
-            if (amBuffer) {
-                delete amBuffer;
-            }
-        }
-
-        bool is_complete() { return _completed; }
+        bool is_complete() const { return _completed; }
         void completed() { _completed = 1; }
 };
 
@@ -927,7 +919,6 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
                                            const std::string &msg, nixlUcxReq &req)
 {
     nixlSerDes ser_des;
-    std::string *ser_msg;
     nixlUcxConnection conn;
     // TODO - temp fix, need to have an mpool
     static struct nixl_ucx_am_hdr hdr;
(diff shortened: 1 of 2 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -958,6 +958,10 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
 
     if (ret == NIXL_IN_PROG) {
         nixlUcxIntReq* nReq = (nixlUcxIntReq*)req;
+        if (nReq->amBuffer) {
```
brminich: where does the leak happen? Isn't it destroyed in `_internalRequestReset`?

tvegas1: leak can happen if that new request what previously released by `genNotif()`

yosefe: could #246 + running with ASAN catch it?

tvegas1: yes two genNotif() calls should do

etoledano: Can `ser_msg` be changed to a smart pointer to fix this issue? `delete` doesn't assign `nullptr` AFAIK so its not trusted to rely on.

tvegas1: moved to smart pointer (delete was deleting newly created, leaving amBuff properly untouched)

brminich: seems the leak also happens in `_internalRequestInit`

brminich: request management seems to be incorrect. I'd suggest to revise it instead of fixing one of the leaks

tvegas1: there is no leak on _internalRequestInit. request management can surely be discussed

brminich (approved):

tstamler (approved):
sorry about the serdes cleanup

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -949,18 +940,16 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
     ser_des.addStr("name", localAgent);
     ser_des.addStr("msg", msg);
     // TODO: replace with mpool for performance
-    ser_msg = new std::string(ser_des.exportStr());
 
+    auto buffer = std::make_unique<std::string>(std::move(ser_des.exportStr()));
     ret = uw->sendAm(conn.ep, NOTIF_STR,
                      &hdr, sizeof(struct nixl_ucx_am_hdr),
-                     (void*) ser_msg->data(), ser_msg->size(),
+                     (void*)buffer->data(), buffer->size(),
```
etoledano, line 984: nit: ‘static_cast’

src/utils/serdes/serdes.cpp
```diff
@@ -50,27 +50,27 @@ std::string nixlSerDes::getStr(const std::string &tag){
...
-    
+
     //move past string plus | delimiter
     des_offset += len + 1;
 
     return ret;
 }
 
-/* Ser/Des for Byte buffers */
+// Byte buffers serialization
```
etoledano, line 73: Many changes due to missing clang, understood.

PR #244, opened 2025-05-03
Topic/ucx multi gpu/final

artpol84:

PR #257, opened 2025-05-06
Extend UCX engine with multi worker support

w1ldptr:
This is a preparatory series for concurrency model development but also stands on its own as an improvement for UCX progress thread.

First, it refactors UCX engine and utils code:

- Move methods that are not actually related to particular worker to classes they belong to.
- Improve memory safety by using smart pointers across the whole code. Use strictest smart pointer that fits the use-case (i.e. unique_ptr where possible).
- Don't arbitrary create temporary copies of objects where it is not strictly necessary. Access the instances via an iterator to the container that owns them.
- Properly initialize/cleanup nixlUcpEp instances via class ctor/dtor instead of member methods that can be called zero or multiple times per object instead of exactly once.

Then, it allows user to instantiate UCX engine with multiple workers by setting engine custom parameter "num_workers"  value. This reduces lock contention on individual workers in multi-threaded environment.

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -27,7 +27,7 @@ using namespace std;
 
 static nixl_status_t ucx_status_to_nixl(ucs_status_t status)
 {
-    if (status == UCS_OK) [[likely]] {
+    if (status == UCS_OK) {
         return NIXL_SUCCESS;
     }
 
@@ -428,9 +428,7 @@ std::unique_ptr<char []> nixlUcxWorker::epAddr(size_t &size)
     ucp_worker_attr_t wattr;
     ucs_status_t status;
 
-    wattr.field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS |
-                       UCP_WORKER_ATTR_FIELD_ADDRESS_FLAGS;
-    wattr.address_flags = UCP_WORKER_ADDRESS_FLAG_NET_ONLY;
+    wattr.field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS;
     status = ucp_worker_query(worker, &wattr);
     if (UCS_OK != status) {
         // TODO: printf
--- a/.github/workflows/aws_efa_validation.yml
+++ b/.github/workflows/aws_efa_validation.yml
@@ -1,0 +1,69 @@
+name: AWS EFA NIXL Validation
+
+on:
+  push:
+    branches:
+    - main
+    - "pull-request/[0-9]+"
+
+jobs:
+  run_aws_tests:
+    name: Run AWS Tests - ${{ matrix.test_name }}
+    environment: SWX_AWS
+    runs-on: ubuntu-latest
+    env:
+      AWS_DEFAULT_REGION: eu-central-1
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test_name: "C++ Tests"
+            test_scripts:
+              - .gitlab/test_cpp.sh $NIXL_INSTALL_DIR
+          - test_name: "Python Tests"
+            test_scripts:
+              - .gitlab/test_python.sh $NIXL_INSTALL_DIR
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Setup
+        run: |
+          set -exE
+          sudo apt-get -qq update
+          sudo apt-get -qq install -y curl gettext git gzip jq tar unzip
+
+          # Install kubectl
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+          # Install AWS CLI
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip -qq awscliv2.zip
+          sudo ./aws/install --update
+
+          # Verify AWS credentials
+          aws sts get-caller-identity >/dev/null
+
+      - name: Run AWS tests
+        working-directory: ./contrib/aws-efa
+        timeout-minutes: 180
+        run: |
+          set -o pipefail
+          test_cmd='${{ join(matrix.test_scripts, ' && ') }}'
+          ./aws_test.sh "$test_cmd" 2>&1 | tee test_output.log
+
+      - name: Cleanup AWS resources
+        if: always()
+        working-directory: ./contrib/aws-efa
+        run: |
+          JOB_ID=$(grep 'JOB_ID=' test_output.log | head -n1 | cut -d= -f2)
+          echo "JOB_ID=$JOB_ID"
+          aws batch terminate-job \
+            --job-id "$JOB_ID" \
+            --reason 'Terminated by GitHub Actions' || true
--- a/.github/workflows/copyright-check.ps1
+++ b/.github/workflows/copyright-check.ps1
@@ -145,7 +145,7 @@ $global:copyright_results = @{
 
 # === end common.ps1 extensions ===
 
-$ignored_files = @('.clang-format', '.gitattributes', '.gitignore', '.gitkeep', '.patch', 'Cargo.lock', 'LICENSE', 'uv.lock', 'rust-toolchain.toml', 'codespell.txt')
+$ignored_files = @('.clang-format', '.gitattributes', '.gitignore', '.gitkeep', '.patch', 'Cargo.lock', 'LICENSE', 'uv.lock', 'rust-toolchain.toml', 'codespell.txt', 'aws_job_def.json')
 write-debug "<copyright-check> ignored_files = ['$($ignored_files -join "','")']."
 $ignored_paths = @('.github', '.mypy_cache', '.pytest_cache')
 write-debug "<copyright-check> ignored_paths = ['$($ignored_paths -join "','")']."
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,8 @@
 # Abseil-CPP
 subprojects/abseil-cpp-*
 subprojects/packagecache/
+
+.vscode
+
+# AWS EFA test files
+contrib/aws-efa/aws_vars.json
--- a/README.md
+++ b/README.md
@@ -36,15 +36,9 @@ pip install nixl
 
 `$ sudo apt install build-essential cmake pkg-config`
 
-If you want to build POSIX plugin:
-`$ sudo apt install liburing-dev`
-
 ### Fedora:
 
 `$ sudo dnf install gcc-c++ cmake pkg-config`
-
-If you want to build POSIX plugin:
-`$ sudo dnf install liburing-devel`
 
 ### Python
 
@@ -99,6 +93,14 @@ $ mkdir build && cd build
 $ cmake ..
 $ make -j$(nproc) && make install
 ```
+
+### Additional plugins
+
+Some plugins may have additional build requirements, see them here:
+
+- [Mooncake](src/plugins/mooncake/README.md)
+- [POSIX](src/plugins/posix/README.md)
+- [GDS](src/plugins/cuda_gds/README.md)
 
 ## Getting started
 ### Build & install
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -35,7 +35,13 @@ RUN apt-get update -y && \
     protobuf-compiler-grpc \
     libcpprest-dev \
     etcd-server \
-    etcd-client
+    etcd-client \
+    autotools-dev \
+    automake \
+    libtool \
+    libz-dev \
+    flex \
+    build-essential
 
 WORKDIR /workspace
 RUN git clone https://github.com/etcd-cpp-apiv3/etcd-cpp-apiv3.git &&\
@@ -58,6 +64,7 @@ RUN wget --tries=3 --waitretry=5 "https://static.rust-lang.org/rustup/archive/1.
     chmod -R a+w $RUSTUP_HOME $CARGO_HOME
 
 WORKDIR /workspace/nixl
+COPY . /workspace/nixl
 
 # LD_LIBRARY_PATH is needed for auditwheel to find libcuda.so.1
 # Set incorrectly to `compat/lib` in cuda-dl-base image
@@ -69,6 +76,30 @@ RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
     uv pip install --upgrade meson pybind11 patchelf
 
 COPY . /workspace/nixl
+
+RUN rm -rf /usr/lib/ucx
+RUN rm -rf /opt/hpcx/ucx
+
+RUN cd /usr/local/src && \
+     git clone https://github.com/openucx/ucx.git && \
+     cd ucx && 			     \
+     git checkout v1.19.x &&	     \
+     ./autogen.sh && ./configure     \
+         --enable-shared             \
+         --disable-static            \
+         --disable-doxygen-doc       \
+         --enable-optimizations      \
+         --enable-cma                \
+         --enable-devel-headers      \
+         --with-cuda=/usr/local/cuda \
+         --with-verbs                \
+         --with-dm                   \
+         --with-gdrcopy=/usr/local   \
+         --with-efa                  \
+         --enable-mt &&              \
+     make -j &&                      \
+     make -j install-strip &&        \
+     ldconfig
 
 RUN rm -rf build && \
     mkdir build && \
--- a/contrib/aws-efa/README.md
+++ b/contrib/aws-efa/README.md
@@ -1,0 +1,73 @@
+# NIXL AWS Testing
+
+This directory contains scripts and configuration files for running NIXL tests on AWS infrastructure using AWS Batch and EKS.
+
+## Overview
+
+The AWS test infrastructure allows NIXL to be automatically tested on AWS Elastic Fabric Adapter (EFA) environments through GitHub Actions. It leverages AWS Batch to manage compute resources and job execution.
+
+## Prerequisites
+
+- AWS account with access to EKS and AWS Batch
+- Pre-configured AWS job queue: `ucx-nxil-jq`
+- Pre-configured AWS EKS cluster: `ucx-ci`
+- Properly registered job definition: `NIXL-Ubuntu-JD`
+
+## Files
+
+- **aws_test.sh**: Main script that submits and monitors AWS Batch jobs
+- **aws_vars.template**: Template file for AWS Batch job configuration
+- **aws_job_def.json**: Job definition (Registered once, for reference only)
+
+## GitHub Actions Integration
+
+The script is designed to run in GitHub Actions when branches are updated.
+It relies on the `GITHUB_REF` environment variable to determine which branch or commit to test. The value can be a branch name (e.g., `main`, `feature-xyz`) or a commit SHA.
+
+### Required GitHub Secrets
+
+The following secrets are configured in the GitHub environment:
+
+- `AWS_ACCESS_KEY_ID`: AWS access key with permissions for AWS Batch and EKS
+- `AWS_SECRET_ACCESS_KEY`: Corresponding secret access key
+
+## Manual Execution
+
+To run the tests manually:
+
+1. Set your AWS account using `aws configure` command or env vars.
+2. Substitute GH variables:
+
+```bash
+# Branch or commit SHA to test
+export GITHUB_REF="main"
+
+# Other required variables
+export GITHUB_SERVER_URL="https://github.com"
+export GITHUB_REPOSITORY="ai-dynamo/nixl"
+
+# Run the script with your test command(s)
+./aws_test.sh ".gitlab/test_cpp.sh /opt/nixl"
+
+# Multiple test commands can be chained with '&&'
+./aws_test.sh ".gitlab/test_cpp.sh /opt/nixl && .test_script2.sh param123"
+```
+
+## Test Execution Flow
+
+The AWS test script:
+
+1. Generates AWS job configuration from template
+2. Submits a job to AWS
+3. Monitors AWS job execution
+4. Streams logs from the AWS pod
+5. Reports success or failure
+
+## Container Image
+
+The script uses the container image: `nvcr.io/nvidia/pytorch:25.02-py3`
+You can override this by setting the `CONTAINER_IMAGE` environment variable:
+
+```bash
+export CONTAINER_IMAGE="your-custom-image:tag"
+```
--- a/contrib/aws-efa/aws_job_def.json
+++ b/contrib/aws-efa/aws_job_def.json
@@ -1,0 +1,54 @@
+{
+    "jobDefinitionName": "NIXL-Ubuntu-JD",
+    "type": "container",
+    "parameters": {},
+    "timeout": {
+        "attemptDurationSeconds": 10800
+    },
+    "tags": {
+        "Project": "NIXL",
+        "Environment": "Testing"
+    },
+    "eksProperties": {
+        "podProperties": {
+            "hostNetwork": true,
+            "imagePullSecrets": [],
+            "containers": [
+                {
+                    "image": "nvcr.io/nvidia/pytorch:25.02-py3",
+                    "command": [
+                        "/bin/bash",
+                        "-c",
+                        "sleep 1h"
+                    ],
+                    "args": [],
+                    "env": [],
+                    "resources": {
+                        "limits": {
+                            "memory": "167936Mi",
+                            "cpu": "70"
+                        }
+                    },
+                    "volumeMounts": [
+                        {
+                            "name": "efa-volume",
+                            "mountPath": "/dev/infiniband/uverbs0"
+                        }
+                    ],
+                    "securityContext": {
+                        "privileged": true
+                    }
+                }
+            ],
+            "initContainers": [],
+            "volumes": [
+                {
+                    "name": "efa-volume",
+                    "hostPath": {
+                        "path": "/dev/infiniband/uverbs0"
+                    }
+                }
+            ]
+        }
+    }
+}
--- a/contrib/aws-efa/aws_test.sh
+++ b/contrib/aws-efa/aws_test.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -exE -o pipefail
+
+usage() {
+    echo ""
+    echo "Description: Run NIXL tests on AWS infrastructure"
+    echo ""
+    echo "Usage: $0 <test script> <test script args>"
+    echo ""
+    echo "Example: $0 .gitlab/test_cpp.sh \$NIXL_INSTALL_DIR && .test_script123.sh param123"
+    echo ""
+    echo "Required environment variables:"
+    echo "  GITHUB_REF        - Git reference to checkout (e.g., main, branch-xyz, commit SHA)"
+    echo "  GITHUB_SERVER_URL - GitHub server URL (e.g., \"https://github.com\")"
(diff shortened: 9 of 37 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -336,52 +341,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
 
     vramApplyCtx();
 
+    {
+        std::unique_lock<std::mutex> lock(pthrActiveLock);
+        pthrActive = true;
+    }
+    pthrActiveCV.notify_one();
+
+    std::vector<pollfd> poll_fds;
```
yosefe: consider using pre-created epoll fd

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -421,15 +452,21 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
-                           _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER);
-    uw = new nixlUcxWorker(uc);
-    uw->epAddr(n_addr, workerSize);
-    workerAddr = (void*) n_addr;
+    uc = std::make_shared<nixlUcxContext>(devs, sizeof(nixlUcxIntReq),
+                                          _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER,
+                                          init_params->enableProgTh);
+    for (unsigned int i = 0; i < init_params->numWorkers; i++) {
+        uint64_t n_addr;
+        auto uw = std::make_unique<nixlUcxWorker>(uc);
```
yosefe: maybe something like
`const auto & uw = uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -629,14 +666,16 @@ nixl_status_t nixlUcxEngine::disconnect(const std::string &remote_agent) {
...
-        if(ret == NIXL_IN_PROG){
-            uw->reqRelease(req);
+        nixl_status_t ret = NIXL_SUCCESS;
+        for (size_t i = 0; i < uws.size(); i++) {
+            nixlUcxReq req;
+            ret = getWorker(i)->sendAm(conn.getEp(i), DISCONNECT,
+                                 &hdr, sizeof(struct nixl_ucx_am_hdr),
+                                 (void*) localAgent.data(), localAgent.size(),
+                                 flags, req);
+            //don't care
```
yosefe: maybe unite common code with nixlUcxEngine::connect?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -592,29 +625,33 @@ nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
...
-                     &hdr, sizeof(struct nixl_ucx_am_hdr),
-                     (void*) localAgent.data(), localAgent.size(),
-                     flags, req);
-
-    if(ret < 0) {
-        return ret;
+    bool error = false;
+    nixl_status_t ret = NIXL_SUCCESS;
+    std::vector<nixlUcxReq> reqs;
+    for (size_t i = 0; i < uws.size(); i++) {
```
yosefe, line 610: why sometimes we iterate over uws and in other times from 0 to uws.size() and use getWorker(i) ?

src/utils/ucx/ucx_utils.h
```diff
@@ -126,6 +118,18 @@ class nixlUcxWorker {
 
     void reqRelease(nixlUcxReq req);
     void reqCancel(nixlUcxReq req);
+
+    /* Memory management */
+    static int memReg(std::shared_ptr<nixlUcxContext> &ctx, void *addr, size_t size, nixlUcxMem &mem);
+    static size_t packRkey(std::shared_ptr<nixlUcxContext> &ctx, nixlUcxMem &mem, uint64_t &addr, size_t &size);
+    static void memDereg(std::shared_ptr<nixlUcxContext> &ctx, nixlUcxMem &mem);
+
```
yosefe: shall we move these functions to nixlUcxContext?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: What do you mean? They are already pre-created, i.e. they are created once before the main loop starts.

If you mean during class init, then I would prefer the init code to stay here because:

1. I don't like putting random data to be visible to whole instance so any other method could access it as opposed to keeping things local.
2. I only want to create the descriptors when progress thread is enabled and the most obvious way to do it is in the progress thread itself instead of adding more conditionals into class ctor.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Because in some places I need the index to access other data. In this case I'm using it for "conn.getEp(i)". When the index is not needed I prefer to keep the loop simple by iterating over uws directly.

src/utils/ucx/ucx_utils.h
w1ldptr: Basically that would require making nixlUcxContext to be friend class of nixlUcxMem and nixlUcxRkey due to the current design (all their data is private and is only accessible to their friend class nixlUcxWorker). I personally don't have a strong opinion on this but Artem suggested he would prefer to make them static and leave them in nixlUcxWorker as opposed to extending friend classes.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: It is not clear what would be the common code. Even the main loop body is slightly different due to disconnect just discarding the request immediately.

Can you suggest any nicer approach than bunch of ugly conditionals like "if (disconnect) do_this else do_that" everywhere? (which are IMO worse than just keeping the function implementations separate)

src/plugins/ucx/ucx_backend.cpp
yosefe: i mean to use epoll_wait() instead of poll()

src/plugins/ucx/ucx_backend.cpp
yosefe: i guess it will be a function the gets the active message id (CONN_CHECK/DISCONNECT) and sends it on all workers and returns a vector of requests. here we can discard it.

yosefe:
Would we need to drop some locks in nixlAgent for real concurrency?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Hmm, epoll_wait() would likely perform similarly or worse on small number of descriptors like we have here (it is not like this loop is going to handle 10k TCP sockets or something) and on top of that it is not UNIX spec and is not portable (so you won't be able to run NIXL without modifications on BSD if need be).

w1ldptr:
> Would we need to drop some locks in nixlAgent for real concurrency?

The current plan is to use read-write lock which would allow running data path APIs concurrently.

src/utils/ucx/ucx_utils.h
yosefe: i would make their fields public (or use constructor+getters)

artpol84: Actually, I don't like the static functions approach much, and maybe friend is better. 

My reasoning:
* As mem, ep, etc. are handles, it's not right to expose them to the outside world.
* On the other hand, UCX components are tightly coupled - memory is registered with UCP context and later used to post with UCP endpoint which is derived from UCP worker.

So, in my opinion, the friend approach would be the most natural and easiest to implement.
Getters will intro unnecessary code without benefits.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: And what do you do in case of an error on Nth sendAm() call? Return some ugly pair<errcode, req_vector> so caller could manually call uw->test() or uw->reqRelease() on all the previous successful requests?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -58,11 +59,14 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
     ucp_params.features = UCP_FEATURE_RMA | UCP_FEATURE_AMO32 | UCP_FEATURE_AMO64 | UCP_FEATURE_AM;
+    if (wakeup)
+        ucp_params.features |= UCP_FEATURE_WAKEUP;
+
     switch(mt_type) {
     case NIXL_UCX_MT_SINGLE:
-    case NIXL_UCX_MT_WORKER:
         ucp_params.mt_workers_shared = 0;
         break;
+    case NIXL_UCX_MT_WORKER:
```
artpol84, line 336: @yosefe @brminich just wanted to double-check that this is required when the worker-level MT is enabled.
It makes perfect sense to me and @w1ldptr, but want to double-check

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -865,10 +924,10 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
 
         switch (operation) {
         case NIXL_READ:
-            ret = uw->read(rmd->conn.ep, (uint64_t) raddr, rmd->rkey, laddr, lmd->mem, lsize, req);
+            ret = uw->read(rmd->conn.getEp(workerId), (uint64_t) raddr, rmd->getRkey(workerId), laddr, lmd->mem, lsize, req);
             break;
         case NIXL_WRITE:
-            ret = uw->write(rmd->conn.ep, laddr, lmd->mem, (uint64_t) raddr, rmd->rkey, lsize, req);
+            ret = uw->write(rmd->conn.getEp(workerId), laddr, lmd->mem, (uint64_t) raddr, rmd->getRkey(workerId), lsize, req);
```
artpol84: As UCX engine is a friend of both EP and RMD (which is OK IMO as they are tightly coupled), do we need "getters" here?
If in the future we add some additional checks in debug mode (on index, for example), it might make sense.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -242,13 +242,12 @@ static void _internalRequestReset(nixlUcxIntReq *req) {
 class nixlUcxBackendH : public nixlBackendReqH {
 private:
     nixlUcxIntReq head;
-    nixlUcxWorker* uw;
+    nixlUcxEngine &eng;
```
artpol84, line 240: Note #252 as it touches the same area

artpol84: https://github.com/ai-dynamo/nixl/pull/252/files#diff-70d20dbd8f24f38c9d54eb28a879ee16bc492a43d107103617789e3ed29e0826R247

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/utils/ucx/ucx_utils.h
w1ldptr: Moved rkey import functions to nixlUcxEp and memory registration functions to nixlUcxContext.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -336,52 +341,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+                    if (status != UCS_ERR_BUSY)
+                        NIXL_ERROR << "Couldn't arm worker, error " << status << std::endl;
+                    all_armed = false;
+                    break;
+                }
+            }
+            if (!all_armed || pthrStop)
+                break;
+
+            if (poll(poll_fds.data(), poll_fds.size(), pthrDelay) > 0) {
```
artpol84: Another tiny optimization: to speed up the cleanup, we can add a pipe[] from the main thread to the progress thread so it can be woken up instantly if agent is shutting down.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -391,7 +419,7 @@ void nixlUcxEngine::progressThreadStop()
         return;
     }
 
-    pthrStop = 1;
+    pthrStop = true;
```
artpol84, line 411: here we'd write in to the pipe to wake the progress thread.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -421,19 +452,27 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
-                           _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER);
-    uw = new nixlUcxWorker(uc);
-    uw->epAddr(n_addr, workerSize);
-    workerAddr = (void*) n_addr;
-
-    uw->regAmCallback(CONN_CHECK, connectionCheckAmCb, this);
-    uw->regAmCallback(DISCONNECT, connectionTermAmCb, this);
-    uw->regAmCallback(NOTIF_STR, notifAmCb, this);
+    uc = std::make_shared<nixlUcxContext>(devs, sizeof(nixlUcxIntReq),
+                                          _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER,
```
artpol84: Note that it's being fixed in https://github.com/ai-dynamo/nixl/pull/244. If it's accepted, we would need to fix this during rebase (we don't always need MT_WORKER, only if we are in the MT mode)

https://github.com/ai-dynamo/nixl/pull/244/files#diff-70d20dbd8f24f38c9d54eb28a879ee16bc492a43d107103617789e3ed29e0826R322

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -421,19 +452,27 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+                                          init_params->enableProgTh);
+    for (unsigned int i = 0; i < init_params->numWorkers; i++) {
+        uint64_t n_addr;
+        const auto & uw = uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));
+        uw->epAddr(n_addr, workerSize);
```
artpol84: This is a memory leak as epAddr allocates memory for the address.

We want to do it once at the end of the loop. And my suggestion is to do it for the worker[0].
Then in the progress it will be easy to figure out of `progressNotif` needs to be called (all notifs are coming on the worker which address was shared with others)

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -494,14 +531,14 @@ nixl_status_t nixlUcxEngine::endConn(const std::string &remote_agent) {
 
     nixlUcxConnection &conn = remoteConnMap[remote_agent];
 
-    if(uw->disconnect_nb(conn.ep) < 0) {
-        return NIXL_ERR_BACKEND;
-    }
-
+    bool error = false;
+    for (size_t i = 0; i < conn.eps.size(); i++)
+        if(getWorker(i)->disconnect_nb(conn.getEp(i)) < 0)
```
artpol84: I wonder if its a good idea to have disconnect as an EP routine.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -592,29 +627,33 @@ nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
...
-                     flags, req);
-
-    if(ret < 0) {
-        return ret;
+    bool error = false;
+    nixl_status_t ret = NIXL_SUCCESS;
+    std::vector<nixlUcxReq> reqs;
+    for (size_t i = 0; i < uws.size(); i++) {
+        reqs.emplace_back();
+        ret = getWorker(i)->sendAm(conn.getEp(i), CONN_CHECK,
```
artpol84: Same for communication (put/get/am).
Will make it slightly more readable.

doesn't need to be this PR though.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -735,32 +785,40 @@ nixl_status_t
...
+    std::vector<char> addr(size);
+    nixlSerDes::_stringToBytes(addr.data(), blob, size);
 
-    int ret = uw->rkeyImport(conn.ep, addr, size, md->rkey);
-    if (ret) {
-        // TODO: error out. Should we indicate which desc failed or unroll everything prior
+    bool error = false;
+    for (size_t worker_id = 0; worker_id < uws.size(); worker_id++) {
+        nixlUcxRkey rkey;
+        error = md->conn.getEp(worker_id).rkeyImport(addr.data(), size, rkey);
```
artpol84: Yeah ... we already opened a can of worms on EP adding rkeyImport (and I think it's good)

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -735,32 +785,40 @@ nixl_status_t
...
+        if (error)
+            // TODO: error out. Should we indicate which desc failed or unroll everything prior
+            break;
+        md->rkeys.push_back(rkey);
+    }
+    if (error) {
+        for (size_t worker_id = 0; worker_id < md->rkeys.size(); worker_id++)
```
artpol84: my 5 cents is that for such local iterators that are mentioned multiple times in the expr. below, it's better to use something short, like `wid` or `wId`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -336,52 +341,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
     while (!pthrStop) {
-        int i;
-        for(i = 0; i < noSyncIters; i++) {
-            uw->progress();
+        bool made_progress = false;
+        for (auto &uw: uws)
+            while (uw->progress())
+                made_progress = true;
+
+        if (made_progress) {
```
artpol84: As mentioned in connection establishment, we can only have notifications on one worker (I suggest worker[0]).
So we can save few cycles calling `notifProgress()` only if progress was on worker[0].

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -336,52 +341,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    }
+
     while (!pthrStop) {
-        int i;
-        for(i = 0; i < noSyncIters; i++) {
-            uw->progress();
+        bool made_progress = false;
+        for (auto &uw: uws)
+            while (uw->progress())
+                made_progress = true;
```
artpol84: TBH I don't like this separation of progress and arming. It feels you can efficiently combine the two like below. It also allows you to avoid touching/arming workers that have not work to do that might save few cycles.

```C

// Append the pipe descriptor allowing to interrupt and stop at any time.
poll_fds.push_back({pipe_fds[1], POLLIN, 0});

vector<nixlUcxWorker> active = uws;

while(true) {
    for (auto &uw: active) {
        // TODO: for fairness, maybe add some limit
        while (true) {
            bool hadProgr = false;
            while (uw->progress()) {
                hadProgr = true;
            }
            ucs_status_t status = ucp_worker_arm(uw->getWorker());
            if (status != UCS_OK) {
               ...
               continue;  // try one more time
            }
            break; // Armed
        }
        if (uw == uws[0]) {
            notifProgress();
        }
    }

    // No active UCX workers in the set
    active.clear();
    
    // the poll may be triggered by UCX fd's OR by the pipe FD owned by the user thread 
    while (poll(poll_fds.data(), poll_fds.size(), pthrDelay) == 0);

    if (pthrStop) {
        break;
    }

    // Only go over UCX workers 
    // (ignore pip_fd as it only triggers once when the thread exits)
    for (size_t wId = 0; wId < uws.size(); wId++) {
        if (poll_fds[wId].revents & POLLIN) {
            active.push_back(uws[wId]);
            poll_fds[wId].revents = 0;
        }
    }
}
```

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I'm not sure it would be a straight-up optimization since the code will now have to always check for single special descriptor index to determine whether it is that pipe or UCX wait fd, or implement some tricks making the code harder to understand (like in your proposed implementation), or at least have poll() always check one more descriptor, so it is more like a tradeoff between marginally faster shutdown and simpler progress thread code, but okay...

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Why wouldn't we want getters? They afford us a flexibility of changing implementation details without also modifying 10 different places when ep is accessed in the code and without any trade-offs for using them.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: WDYM?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Good catch, thanks!

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Makes sense, done.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Looks good, pushed slightly simplified version of this that also manages active workers a bit differently due to unique_ptrs not being copy-able.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
artpol84: let's discuss on a call

artpol84 (approved):

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.h
```diff
@@ -118,7 +127,7 @@ class nixlUcxEngine : public nixlBackendEngine {
         notif_list_t notifPthrPriv, notifPthr;
 
         // Map of agent name to saved nixlUcxConnection info
-        std::unordered_map<std::string, nixlUcxConnection,
+        std::unordered_map<std::string, std::shared_ptr<nixlUcxConnection>,
```
yosefe: maybe define type for connection ptr?

src/plugins/ucx/ucx_backend.h
```diff
@@ -227,6 +238,9 @@ class nixlUcxEngine : public nixlBackendEngine {
         //public function for UCX worker to mark connections as connected
         nixl_status_t checkConn(const std::string &remote_agent);
         nixl_status_t endConn(const std::string &remote_agent);
+
+        const std::unique_ptr<nixlUcxWorker> &getWorker(size_t worker_id) const;
```
yosefe: make it inline? and return nixlUcxWorker&

src/api/cpp/backend/backend_aux.h
```diff
@@ -51,6 +51,8 @@ class nixlBackendInitParams {
 
         bool              enableProgTh;
         nixlTime::us_t    pthrDelay;
+
+        unsigned int      numWorkers;
```
brminich: is it needed for other backends? maybe it can be custom param?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -157,6 +142,157 @@ nixlUcxEp::closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags)
...
+        setState(NIXL_UCX_EP_STATE_CONNECTED);
+    else
+        throw std::runtime_error("failed to create ep");
+}
+
+ nixlUcxEp::~nixlUcxEp()
+ {
+     nixl_status_t status = disconnect_nb();
+     if (status)
+         NIXL_ERROR << "Failed to disconnect ep with status " << status << std::endl;
```
```suggestion
         NIXL_ERROR << "Failed to disconnect ep with status " << status;
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -157,6 +142,157 @@ nixlUcxEp::closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags)
 
     return ucx_status_to_nixl(status);
 }
+nixlUcxEp::nixlUcxEp(ucp_worker_h worker, void* addr)
```
brminich, line 146:
```suggestion

nixlUcxEp::nixlUcxEp(ucp_worker_h worker, void* addr)
```

src/utils/ucx/ucx_utils.cpp
brminich: `mt_workers_shared = 1` is needed when there are multiple workers used by different threads. And it is not needed if there is one worker  used by multiple threads

src/api/cpp/backend/backend_aux.h
w1ldptr: enableProgTh and pthrDelay are also UCX-specific, so I added numWorkers here for consistency. I don't have a strong opinion either way.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -342,52 +348,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-        //     cnt++;
-        // }
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        active.clear();
+        while (poll(poll_fds.data(), poll_fds.size(), -1) == 0);
```
yosefe: shall we handle EINTR and EAGAIN? or failures from poll() in general?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -342,52 +348,75 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+
+        for (size_t wid = 0; wid < uws.size() && wid < poll_fds.size(); wid++) {
+            if (poll_fds[wid].revents & POLLIN) {
+                active.push_back(wid);
```
yosefe: do we really have to maintain active() array? could just rearm what is needed in this loop
(could reset poll_fds[wid].revents to POLLIN before entering the loop to unite code)

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -416,19 +446,39 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     std::vector<std::string> devs; /* Empty vector */
     nixl_b_params_t* custom_params = init_params->customParams;
 
+    if (!init_params->numWorkers) {
+        this->initErr = true;
+        return;
+    }
```
yosefe: space line after

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -416,19 +446,39 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
             return;
         }
     }
 
+    if (init_params->enableProgTh) {
+        pthrOn = true;
+        int err = pipe(pthrStopFds);
+        if (err) {
+            NIXL_ERROR << "Failed to create progress thread pipe with error " << err;
+            this->initErr = true;
```
yosefe: why need this-> prefix?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -592,35 +626,37 @@ nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
...
+            error = true;
+            break;
+        }
     }
 
     //wait for AM to send
-    while(ret == NIXL_IN_PROG){
-        ret = uw->test(req);
-    }
+    for (size_t i = 0; i < reqs.size(); i++)
```
yosefe, line 624: shall we initialize ret before this loop?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Nevermind, will refactor it anyway to re-arm the same array.

src/plugins/ucx/ucx_backend.h
w1ldptr: There is no point in using unique_ptr when you just give everyone pointer or reference to internal object. What exactly is the issue with returning the reference to the smart pointer?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
w1ldptr: removed

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
w1ldptr: This PR became too big and disjointed. Removed the progress thread refactoring patch to be submitted separately.

src/plugins/ucx/ucx_backend.h
w1ldptr: done

src/plugins/ucx/ucx_backend.h
yosefe: right, was going to update this comment. so just make it inline

w1ldptr: done

src/api/cpp/backend/backend_aux.h
brminich: imo progress thread sounds more general than worker, i'd hide it from the common params
@yosefe, wdyt?

w1ldptr: Another approach could be to not expose this config to the user at all and instead use 1 worker when mt model is 'strict' or some predetermined value >1 when the model allows concurrent execution. WDYT?

w1ldptr: Moved it to custom param.

brminich (approved):

yosefe (approved):

PR #258, opened 2025-05-07
NIXLBench: Bug fixes to properly create desclist and calculate stat + read support for nvshmem worker

sb17v:
* Enable READ support for NVSHMEM
* Introduce synchronization to properly calculate time
  * Introduce sync after (i)exchangeIOV and (ii)Warmup is complete
* Bug fix during printStats and createTransferDescLists
  * Pass isTarget in printStats to restrict target processes from reducing result across procs (Temp fix)
  * Fix createTransferDescLists to provide address within range with RR pattern
* Add cpprest lib during nvcc compilation for NVSHMEM

### DIFF
--- a/benchmark/nixlbench/src/main.cpp
+++ b/benchmark/nixlbench/src/main.cpp
@@ -80,10 +80,11 @@ static std::vector<std::vector<xferBenchIOV>> createTransferDescLists(xferBenchW
 
         for (const auto &iov : iov_list) {
             for (size_t i = 0; i < count; i++) {
-                size_t offset = ((i * stride) % iov.len);
+                size_t dev_offset = ((i * stride) % iov.len);
 
                 for (size_t j = 0; j < batch_size; j++) {
-                    xfer_list.push_back(xferBenchIOV((iov.addr + offset) + (j * block_size),
+                    size_t block_offset = ((j * block_size) % iov.len);
+                    xfer_list.push_back(xferBenchIOV((iov.addr + dev_offset) + block_offset,
                                                       block_size,
                                                       iov.devId));
                 }
@@ -119,7 +120,7 @@ static int processBatchSizes(xferBenchWorker &worker,
             if (IS_PAIRWISE_AND_SG()) {
                 // TODO: This is here just to call throughput reduction
                 // Separate reduction and print
-                xferBenchUtils::printStats(block_size, batch_size, 0);
+                xferBenchUtils::printStats(true, block_size, batch_size, 0);
             }
         } else if (worker.isInitiator()) {
             std::vector<std::vector<xferBenchIOV>> remote_trans_lists(worker.exchangeIOV(local_trans_lists));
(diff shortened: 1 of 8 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -671,6 +673,9 @@ std::variant<double, int> xferBenchNixlWorker::transfer(size_t block_size,
         return std::variant<double, int>(ret);
     }
 
+    // Synchronize to ensure all processes have completed the warmup (iter and polling)
+    synchronize();
```
tvegas1, line 677: worth noting that `synchronize()` return time still can have 1sec difference between ranks as we cannot poll etcd more frequently

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -575,6 +575,8 @@ xferBenchNixlWorker::exchangeIOV(const std::vector<std::vector<xferBenchIOV>> &l
             }
         }
     }
+    // Ensure all processes have completed the exchange with a barrier/sync
+    synchronize();
```
tvegas1, line 579: why do we need synchronize() here since rt->send/recv/Char are acked?

benchmark/nixlbench/src/main.cpp
```diff
@@ -80,10 +80,11 @@ static std::vector<std::vector<xferBenchIOV>> createTransferDescLists(xferBenchW
 
         for (const auto &iov : iov_list) {
             for (size_t i = 0; i < count; i++) {
-                size_t offset = ((i * stride) % iov.len);
+                size_t dev_offset = ((i * stride) % iov.len);
 
                 for (size_t j = 0; j < batch_size; j++) {
-                    xfer_list.push_back(xferBenchIOV((iov.addr + offset) + (j * block_size),
+                    size_t block_offset = ((j * block_size) % iov.len);
```
w1ldptr, line 86: Wonder whether these loops should be somehow scaled to have resulting xfer_list.size() to be divided by batch_size while every one of xferBenchIOV in xfer_list to include batch_size number of ops and it all should total to the same amount of read/written memory.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
sb17v: Yes, you are right. ETCD send-recv implementation is a blocking op. However, the barrier here ensures all OOB exchanges are completed before the actual transfer/poll starts, even if a separate runtime is used with non-blocking send-recv. This also ensures that the transfer and poll start at the same time. Hence, I am not in favor of removing it.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
sb17v: Thanks for the info. It's okay, as this sync call is outside the timer and should not impact the performance measurement.

benchmark/nixlbench/src/main.cpp
w1ldptr: Nvm, I see that batch size is taken into account when calculating the stats instead of scaling the amount of transfer requests.

w1ldptr (approved):

aranadive (approved):

Issue #260, opened 2025-05-07
The --use-cuda parameter in blocking_send_recv_example.py raises a RuntimeError


ZhangShuaiyi:
**target**
```
python3 blocking_send_recv_example.py --ip 127.0.0.1 --mode target --use_cuda true
```
**initiator**
```
python3 blocking_send_recv_example.py --ip 127.0.0.1 --use_cuda true
```
**initiator** results in a RuntimeError
```
MD listener is listening on port 8888...
Backend UCX was instantiated
Initialized NIXL agent: initiator
initiator Tensors: [tensor([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.], device='cuda:0'), tensor([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.], device='cuda:0')]
Initiator sending to 127.0.0.1
Ready for transfer
Traceback (most recent call last):
  File "/workspace/nixl/examples/python/blocking_send_recv_example.py", line 135, in <module>
    if not torch.allclose(tensor, torch.ones(10)):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

PR #261, opened 2025-05-07
python: use torch.set_default_device in blocking_send_recv_example.py

ZhangShuaiyi:
Fix: #260
Use `torch.set_default_device("cuda:0")` to select the device and simplify the code for the `--use-cuda` parameter , which was introduced in #221.


tstamler (approved):
thanks for noticing, if it passes in CI I think its a good change

PR #262, opened 2025-05-07
posix: Updated implementation to support both AIO and Uring

vvenkates27:
Update posix implementation to support both AIO and Uring implementations.

### DIFF
--- a/src/plugins/posix/meson.build
+++ b/src/plugins/posix/meson.build
@@ -13,35 +13,81 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Check for liburing dependency
-liburing_dep = dependency('liburing', required: true)
+# Check for libaio dependency (required)
+compiler = meson.get_compiler('cpp')
+libaio_dep = compiler.find_library('aio', required: false)
+
+# If libaio is not found, skip building the POSIX plugin entirely
+if not libaio_dep.found()
+    message('libaio not found, POSIX plugin will not be built')
+    subdir_done()
+endif
+
+# Configure compile flags and set AIO as default
+compile_defs = ['-DHAVE_LIBAIO']
+message('libaio found, building POSIX plugin')
+
+# Try to find liburing (optional) - first try pkg-config
+liburing_dep = dependency('liburing', required: false)
+
+# If pkg-config fails, try finding it as a system library
+if not liburing_dep.found()
+    liburing_dep = compiler.find_library('uring', required: false)
+endif
+
+have_liburing = liburing_dep.found()
+
+if have_liburing
+    compile_defs += ['-DHAVE_LIBURING']
+    message('liburing found, adding io_uring support')
+else
+    # Create a not found dependency to avoid build errors
+    liburing_dep = dependency('', required: false)
+    message('liburing not found, building with AIO support only')
+endif
 
 # Get Abseil dependencies
 absl_log_dep = dependency('absl_log', required: true)
 
-if 'POSIX' in static_plugins
-    posix_backend_lib = static_library('POSIX', 'posix_backend.cpp', 'posix_backend.h', 'posix_plugin.cpp',
-                                       dependencies: [nixl_infra, liburing_dep, absl_log_dep],
-                                       link_args: ['-luring'],
-                                       include_directories: [nixl_inc_dirs, utils_inc_dirs],
-                                       install: false,
-                                       cpp_args: compile_flags,
-                                       name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
-else
-    posix_backend_lib = shared_library('POSIX', 'posix_backend.cpp', 'posix_backend.h', 'posix_plugin.cpp',
-                                       dependencies: [nixl_infra, liburing_dep, absl_log_dep],
-                                       include_directories: [nixl_inc_dirs, utils_inc_dirs],
-                                       link_args: ['-luring'],
-                                       install: true,
-                                       cpp_args: ['-fPIC'],
-                                       name_prefix: 'libplugin_',  # Custom prefix for plugin libraries
-                                       install_dir: plugin_install_dir)
-    if get_option('buildtype') == 'debug'
+# Add pthread dependency
+thread_dep = dependency('threads')
+
+# Define base source files - conditionally include uring_queue.cpp
+posix_sources = [
+    'posix_backend.cpp',
+    'posix_backend.h',
+    'posix_plugin.cpp',
+    'aio_queue.cpp'  # Always include AIO source since it's required
+]
+
+# Add io_uring source if available
+if have_liburing
+    posix_sources += ['uring_queue.cpp']
+endif
+
+# Prepare dependencies list - only include liburing if found
+plugin_deps = [nixl_infra, nixl_common_dep, absl_log_dep, thread_dep, libaio_dep]
+plugin_link_args = ['-laio']
+
+if have_liburing
+    plugin_deps += [liburing_dep]
+    plugin_link_args += ['-luring']
+endif
+
+# Build the plugin using project-wide include directories
+posix_backend_lib = shared_library('POSIX',
+              posix_sources,
+              dependencies: plugin_deps,
+              link_args: plugin_link_args,
+              cpp_args: compile_defs,
+              name_prefix: 'libplugin_',
+              include_directories: [nixl_inc_dirs, utils_inc_dirs],
+              install: true,
+              install_dir: plugin_install_dir)
+
+if get_option('buildtype') == 'debug'
         run_command('sh', '-c',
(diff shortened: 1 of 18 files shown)


src/plugins/posix/meson.build
```diff
@@ -1,47 +1,93 @@
...
+# Configure compile flags and set AIO as default
+compile_defs = ['-DHAVE_LIBAIO']
+message('libaio found, building POSIX plugin')
+
+# Try to find liburing (optional) - first try pkg-config
+liburing_dep = dependency('liburing', required: false)
+
+# If pkg-config fails, try finding it as a system library
+if not liburing_dep.found()
+    liburing_dep = compiler.find_library('uring', required: false)
```
aranadive: Shouldnt the required be true here if you want to find it?

vvenkates27: Keeping it optional, if its not ignore building the code.

src/plugins/posix/posix_backend.cpp
```diff
@@ -1,286 +1,354 @@
...
+    public:
+        static bool isAioAvailable() {
+            return QueueFactory::isAioAvailable();
+        }
+
+        static bool isUringAvailable() {
+            return QueueFactory::isUringAvailable();
+        }
+
+        static nixl_status_t shouldUseAio(const nixl_b_params_t* custom_params, bool& use_aio) {
```
dfarge: Could consider returning a tuple<nixl_status_t, bool> instead

src/plugins/posix/posix_backend.cpp
```diff
@@ -1,286 +1,354 @@
...
+        }
+    };
+}
+
+// -----------------------------------------------------------------------------
+// IO Queue Factory Implementation
+// -----------------------------------------------------------------------------
+
+std::unique_ptr<nixlPosixQueue> IOQueueFactory::createQueue(bool use_aio, int num_entries, bool is_read,
+                                                           const void* params) {
```
dfarge:
```suggestion
std::unique_ptr<nixlPosixQueue> IOQueueFactory::createQueue(bool use_aio, int num_entries, bool is_read,
                                                            const void* params) {
```

src/plugins/posix/posix_backend.cpp
```diff
@@ -1,286 +1,354 @@
...
+        #else
+        (void)params;
+        return nullptr;
+        #endif
+    }
+}
+
+bool IOQueueFactory::isBackendAvailable(bool use_aio) {
+    return use_aio ? QueueFactory::isAioAvailable() : QueueFactory::isUringAvailable();
+}
```
dfarge: This function is unused

src/plugins/posix/posix_backend.cpp
```diff
@@ -1,286 +1,354 @@
...
+        status = initQueues();
+        if (status != NIXL_SUCCESS) {
+            throw std::runtime_error("Failed to initialize queues: " +
+                                   std::to_string(status));
+        }
+    } catch (const std::exception& e) {
+        NIXL_ERROR << "Failed to initialize queues: " << e.what();
+        status = NIXL_ERR_BACKEND;
+    }
+}
```
dfarge: If the exception from the try block is thrown, shouldn't it print "Failed to initialize queues: Failed to initialize queues: NIXL_ERR_BACKEND" for example?

src/plugins/posix/queue_factory_impl.h
```diff
@@ -0,0 +1,97 @@
...
+
+    static bool isUringAvailable() {
+        #ifdef HAVE_LIBURING
+        try {
+            io_uring_params params = {};
+            auto test_queue = createUringQueue(1, true, &params);
+            return test_queue != nullptr;
+        } catch (...) {
+            return false;
+        }
```
dfarge: Why is this try catch block needed?

src/plugins/posix/uring_queue.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+nixl_status_t UringQueue::init(int entries, const io_uring_params& params, bool read_op) {
+    num_entries = entries;
+    num_completed = 0;
+    num_submitted = 0;
+    is_read = read_op;
+    memset(&uring, 0, sizeof(uring));
+
+    // Start with basic parameters, no special flags
```
dfarge: This comment may be unnecessary, this is already specified in nixlPosixBackendReqH::initQueues

src/plugins/posix/uring_queue.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+UringQueue::~UringQueue() {
+    io_uring_queue_exit(&uring);
+}
+
+nixl_status_t UringQueue::submit() {
+    int ret = io_uring_submit(&uring);
+    if (ret < 0) {
```
dfarge: if 0 < ret < num_entries, then some submissions failed, and the user won't know
```suggestion
    if (ret != num_etnries) {
```

src/plugins/posix/posix_backend.cpp
```diff
@@ -1,286 +1,354 @@
...
+        if (use_aio_) {
+            queue = QueueFactory::createAioQueue(queue_depth_, operation == NIXL_READ);
+            if (!queue) {
+                throw std::runtime_error("Failed to create AIO queue");
+            }
+            NIXL_INFO << "Using AIO backend";
+        } else {
+            #ifdef HAVE_LIBURING
+            // Initialize io_uring parameters with basic configuration
+            struct io_uring_params params = {};
```
dfarge, line 131: Not specifying cq_entries defaults to twice the num of sq depth and is wasteful

src/plugins/posix/uring_queue.cpp
```diff
@@ -0,0 +1,128 @@
...
+        NIXL_ERROR << "io_uring submit failed: " << strerror(-ret);
+        return NIXL_ERR_BACKEND;
+    }
+    num_submitted += ret;
```
dfarge: num_submitted is updated but never read or used

src/plugins/posix/uring_queue.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (!sqe) {
+        NIXL_ERROR << "Failed to get io_uring submission queue entry";
+        return NIXL_ERR_BACKEND;
+    }
+
+    if (is_read) {
+        io_uring_prep_read(sqe, fd, buf, len, offset);
+    } else {
+        io_uring_prep_write(sqe, fd, buf, len, offset);
+    }
```
dfarge: For better performance, in the previous version of the code, this if was avoided by assigning the correct op beforehand

src/plugins/posix/uring_queue.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    // Don't use IOSQE_FIXED_FILE since we're not using registered files
+    sqe->flags = 0;
```
dfarge: Does this do anything?

src/plugins/posix/posix_backend.cpp
vvenkates27: Removed.

src/plugins/posix/posix_backend.cpp
vvenkates27: Yes this error is simplistic, we need a way to properly propogate strEnum here to accomplish that.

src/plugins/posix/posix_backend.cpp
vvenkates27: Yes updated in latest patch.

src/plugins/posix/posix_backend.cpp
vvenkates27: Fixed in latest patch

src/plugins/posix/uring_queue.cpp
vvenkates27: Done Fixed in latest patch.

src/plugins/posix/uring_queue.cpp
vvenkates27: Nothing removed.

src/plugins/posix/posix_backend.cpp
vvenkates27: Fixed.

src/plugins/posix/uring_queue.cpp
vvenkates27: Done.. removed num_submitted.

src/plugins/posix/queue_factory_impl.h
vvenkates27: Removed.

src/plugins/posix/uring_queue.cpp
vvenkates27: removed.

aranadive (approved):

PR #265, opened 2025-05-07
Agent: small fixes to commWorker

yanivz-nv:
* NIXL_LOG doesn't require std::endl
* Replace std::cerr with NIXL_ERROR
* Avoid unnecesary copies with references and move semantics
* commThreadStop should be volatile
* Close remote sockets in Agent destructor for cleaning in case of exceptions
* Wrap static functions in nixl_listener.cpp inside unnamed namespace

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -123,6 +123,12 @@ nixlAgent::~nixlAgent() {
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
 
+        // Close remaining connections from comm thread
+        for (auto &[remote, fd] : data->remoteSockets) {
+            shutdown(fd, SHUT_RDWR);
+            close(fd);
+        }
+
         if(data->config.useListenThread) {
             if(data->listener) delete data->listener;
         }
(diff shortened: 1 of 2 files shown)


src/core/nixl_listener.cpp
```diff
@@ -121,14 +123,14 @@ static nixl_status_t storeMetadataInEtcd(const std::string& agent_name,
         etcd::Response response = client->put(metadata_key, metadata).get();
 
         if (response.is_ok()) {
-            NIXL_DEBUG << "Successfully stored " << metadata_type << " in etcd with key: " << metadata_key << std::endl;
+            NIXL_DEBUG << "Successfully stored " << metadata_type << " in etcd with key: " << metadata_key << " (rev " << response.value().modified_index() << ")";
```
etoledano, line 126: `absl::StrFormat`, all around

src/core/agent_data.h
```diff
@@ -89,7 +89,7 @@ class nixlAgentData {
         std::thread                        commThread;
         std::vector<nixl_comm_req_t>       commQueue;
         std::mutex                         commLock;
-        bool                               commThreadStop;
+        volatile bool                      commThreadStop;
```
etoledano: Why is it volatile?

etoledano (commented):
Thanks Yaniv!

src/core/nixl_agent.cpp
```diff
@@ -122,6 +122,11 @@ nixlAgent::~nixlAgent() {
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
 
+        for (auto &[remote, fd] : data->remoteSockets) {
+            shutdown(fd, SHUT_RDWR);
+            close(fd);
+        }
+
```
tstamler, line 131: all sockets are already closed at the end of nixlAgentData::commWorker (nixl_listener.cpp:535), do we need to duplicate that here?

yanivz-nv: In the cases commWorker throws exceptions the sockets won't be properly closed with adding it here.

tstamler: good call, do you think it makes sense to remove from commWorker altogether then? I guess there's no harm in having it both but might be redundant

src/core/agent_data.h
yanivz-nv: Reverted as discussed.

src/core/nixl_agent.cpp
yanivz-nv: I also considered it but kept both. I think it makes more sense to leave it only in the destructor (done).

tstamler (approved):

PR #271, opened 2025-05-08
feat: rust bindings improvements

ryanolson:
This PR:

- adds stubs
  - this should enable us to compile against a stub variant of the .so 
- adds .take_notifs to NotificationMap
  - extracts all the notifications from the c++ map into a rust map and clears the c++ map
  - avoid having to create a new heap allocated c++ object each team we poll for new notifications
- improving tests
  - basic lifecycle test was always asserting the write was not complete immediately when it could actually complete
- refactor giant lib.rs file to more discrete files


src/bindings/rust/tests/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+    loop {
+        let status = agent1.get_xfer_status(&xfer_req).unwrap();
+
+        if !status {
+            println!("Xfer req completed");
+            break;
+        } else {
+            println!("Xfer req not completed");
+        }
+        std::thread::sleep(std::time::Duration::from_millis(100));
```
roiedanino, line 405: why?

aranadive:
@roiedanino can you review?

src/bindings/rust/src/xfer.rs
roiedanino, line 1: File is empty, do you want to move all xfer-requests APIs? otherwise I think we can delete it

src/bindings/rust/stubs.cpp
```diff
@@ -0,0 +1,426 @@
...
+  printf("nixl error: detected use of the NIXL C API's stub; please update the LD_LIBRARY_PATH to include the nixl library\n");
+  std::abort();
+  return NIXL_CAPI_ERROR_EXCEPTION;
+}
+
+nixl_capi_status_t
+nixl_capi_create_agent(const char* name, nixl_capi_agent_t* agent)
+{
+  return nixl_capi_stub_abort();
+}
```
roiedanino, line 53: I'm not sure how that works, shouldn't we do some meson / build.rs changes so the stub API will be used when nixl is not available?

aranadive:
@ryanolson can you follow up on this?

src/bindings/rust/src/xfer.rs
ryanolson: I meant to migrate more pieces. Let's just keep it empty as the next PR will migrate bits from lib.rs

src/bindings/rust/tests/tests.rs
ryanolson: This is simulating a hard polling loop. 

A callback to trip a Notify would be nice here.

src/bindings/rust/stubs.cpp
ryanolson: We don't build the stubs as part of the build system. 

We only use the stubs when we do not detect a locally installed, preferably from the lack of a pkg-config file. 

This allows us to build and link, but not run. 

CUDA has the same pattern. We can link against the stubs of libcuda.so, then drop that binary anywhere libcuda.so / the driver is installed. 

Allows us to compile on CPU only systems without the drivers installed.

ryanolson:
> @ryanolson can you follow up on this?

Done

ryanolson:
FWIW - I've got #401 incorporated. #401 should be a v0.3.0 merge only, with no need to merge it back to main as this PR should suffice.

src/bindings/rust/tests/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    let mut notifs = NotificationMap::new().unwrap();
+    let notify_map;
+    println!("Waiting for notifications");
+    std::thread::sleep(std::time::Duration::from_millis(100));
+
+    // // see if the values in storage2 are 0xbb
+    // assert!(storage2.as_slice().iter().all(|&x| x == 0xbb));
```
roiedanino: uncomment / delete those lines

src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -0,0 +1,165 @@
...
+            // Get the memory type from the list by checking first descriptor
+            let mut len = 0;
+            match nixl_capi_reg_dlist_len(self.inner.as_ptr(), &mut len) {
+                0 => Ok(()),
+                -1 => Err(NixlError::InvalidParam),
+                _ => Err(NixlError::BackendError),
+            }?;
+            if len > 0 {
+                // TODO: Add API to get descriptor memory type
+                MemType::Unknown
```
roiedanino, line 133: Now we have GetType

src/bindings/rust/tests/tests.rs
ryanolson: done.

src/bindings/rust/src/descriptors/reg.rs
ryanolson: presumably this is on the reg_dlist and probably on the xfer_dlist, correct?

if so, let's resolve this update to the api with a separate PR, this is long enough.

mkhazraee: Overall it would be great if we can remove the Unknown, better to match our C++ and Rust APIs on this. Fine to do it in a separate PR, would be nice to be for 0.3.1, but let's merge this if no problems.

PR #272, opened 2025-05-08
Agent API: new param metadataLabel for allowing user to label md in ETCD mode

yanivz-nv:
* This label is appended to the agent's key to allow custom partial MD keys in the ETCD server.
* Allows guardrailing by storing MD of different memory sections using different keys.

### DIFF
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -123,6 +123,18 @@ using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
 constexpr int default_comm_port = 8888;
 
 /**
+ * @brief A constant to define the default metadata label for ETCD server key.
+ *        Appended to the agent's key prefix to form the full key for metadata.
+ */
+extern const std::string default_metadata_label;
+
+/**
+ * @brief A constant to define the default partial metadata label for ETCD server key.
+ *        Appended to the agent's key prefix to form the full key for partial metadata.
+ */
+extern const std::string default_partial_metadata_label;
+
+/**
  * @class nixlAgentOptionalArgs
  * @brief A class for optional argument that can be provided to relevant agent methods.
  */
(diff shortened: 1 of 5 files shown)


src/api/cpp/nixl_types.h
```diff
@@ -122,6 +122,18 @@ using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
  */
 constexpr int default_comm_port = 8888;
 
+/**
+ * @brief A constant to define the default metadata label for ETCD server key.
+ *        Appended to the agent's key prefix to form the full key for metadata.
+ */
+extern const std::string default_metadata_label;
```
mkhazraee, line 129: Not sure we need a extern, or we should also extern the default_comm_port too. I think here is fine though.

src/api/cpp/nixl_types.h
```diff
@@ -167,6 +179,16 @@ class nixlAgentOptionalArgs {
          *                      used in sendLocalMD, fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
          */
         int port = default_comm_port;
+
+        /**
+         * @var metadataLabel Used to specify the label of the metadata to be sent/fetched/invalidated
+         *                    when working with ETCD metadata server. The label will be appended to the
+         *                    agent's key prefix, and the full key will be used to store/fetch/invalidate
+         *                    the metadata key-value pair.
+         *                    Used in fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
```
mkhazraee: Let's not add the partial removal.

src/core/nixl_listener.cpp
```diff
@@ -211,6 +213,40 @@ static nixl_status_t fetchMetadataFromEtcd(const std::string& agent_name,
     }
 }
 
+static nixl_status_t invalidateMetadataInEtcd(const std::string& agent_name,
```
mkhazraee: Let's put this in another PR.

src/api/cpp/nixl_types.h
```diff
@@ -122,6 +122,18 @@ using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
...
+ * @brief A constant to define the default metadata label for ETCD server key.
+ *        Appended to the agent's key prefix to form the full key for metadata.
+ */
+extern const std::string default_metadata_label;
+
+/**
+ * @brief A constant to define the default partial metadata label for ETCD server key.
+ *        Appended to the agent's key prefix to form the full key for partial metadata.
+ */
+extern const std::string default_partial_metadata_label;
```
tstamler, line 135: I just learned that constexpr for std::string is supported, but only for C++20 and newer, so this is good for now. Just a heads up

tstamler (commented):
I only was able to take a quick look, looks like Moein has the important points covered

src/api/cpp/nixl_types.h
yanivz-nv: Can't be constexpr and AFAIU keeping the definition in .cpp will avoid multiple instances of the same variable.

src/api/cpp/nixl_types.h
yanivz-nv: done

src/core/nixl_listener.cpp
yanivz-nv: Not relevant anymore

src/api/cpp/nixl_types.h
mkhazraee: oh now I follow, later we might move the comm port to listener then too, to be consistent, good for now.

tstamler (approved):

PR #275, opened 2025-05-08
Update CODEOWNERS

yosefe:
## Why
Update owners of UCX plugins
cc @tstamler

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,3 +8,7 @@ CODEOWNERS @ai-dynamo/Devops @ai-dynamo/nixl-maintainers
 /CONTRIBUTING.md @ai-dynamo/Devops
 /LICENSE @ai-dynamo/Devops
 /SECURITY.md @ai-dynamo/Devops
+
+# UCX Plugins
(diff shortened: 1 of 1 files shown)


tstamler (commented):
@yosefe Generally I'm ok with putting extra protection around these backends, but I would maybe just suggest adding more people to the list. I'm just worried that we might reach a point where it is too difficult for only two people to review all UCX PRs.

yosefe:
> @yosefe Generally I'm ok with putting extra protection around these backends, but I would maybe just suggest adding more people to the list. I'm just worried that we might reach a point where it is too difficult for only two people to review all UCX PRs.

Added @gleon99

aranadive:
@yosefe There are also UCX tests in test/unit/plugins/ucx. Do you plan to own those as well?

yosefe:
> @yosefe There are also UCX tests in test/unit/plugins/ucx. Do you plan to own those as well?

No, IMO it's not critical to require specific reviewers for tests

aranadive (approved):

PR #277, opened 2025-05-08
CI: try to build Mooncake plugin

tstamler:


brminich:
@tstamler can we close given #1013

PR #278, opened 2025-05-08
Dockerfile for building EFA wheel

tstamler:
I've verified this builds locally successfully, need @aranadive to confirm it looks good with him though

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -81,7 +81,8 @@ RUN rm -rf /opt/hpcx/ucx
 
 RUN cd /usr/local/src && \
      git clone https://github.com/openucx/ucx.git && \
-     cd ucx && \
+     cd ucx && 			     \
+     git checkout v1.19.x &&	     \
      ./autogen.sh && ./configure     \
          --enable-shared             \
          --disable-static            \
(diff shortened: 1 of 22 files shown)


contrib/Dockerfile
```diff
@@ -35,7 +35,15 @@ RUN apt-get update -y && \
     protobuf-compiler-grpc \
     libcpprest-dev \
     etcd-server \
-    etcd-client
+    etcd-client \
+    autotools-dev \
+    automake \
+    libtool \
+    libz-dev \
+    libiberty-dev \
```
aranadive: What is libiberty-dev package? Also we should have libibverbs-dev from the hpcx/ofed install. It shouldnt be required.

tstamler: I don't remember why I added these but it seems to build fine without them so I have now removed

contrib/Dockerfile
```diff
@@ -70,6 +76,28 @@ RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
...
+     cd ucx && \
+     ./autogen.sh && ./configure     \
+         --enable-shared             \
+         --disable-static            \
+         --disable-doxygen-doc       \
+         --enable-optimizations      \
+         --enable-cma                \
+         --enable-devel-headers      \
+         --with-cuda=/usr/local/cuda \
+         --with-verbs                \
```
aranadive, line 95: There should be --with-efa

aranadive: Also please checkout v1.19.x branch. Thats what we tested with. Main has moved to 1.20.

tstamler: got it, didn't realize. Should be updated now

aranadive (approved):

PR #279, opened 2025-05-08
Add file Uitls for NIXL testing

vvenkates27:


examples/python/nixl_gds_example.py
```diff
@@ -29,6 +30,16 @@
         print("Please specify file path in argv")
         exit(0)
 
+    test_file = sys.argv[1]
+
+    # Check if file exists and create/open it
+    if not file_exists(test_file):
+        print(f"Creating new file: {test_file}")
```
yosefe, line 37: pls use nixl absl-based logger or remove this

src/bindings/python/nixl_utils.cpp
```diff
@@ -39,9 +41,61 @@ void verify_transfer(uintptr_t addr1, uintptr_t addr2, int size) {
     for(int i = 0; i<size; i++) assert(((uint8_t*) addr1)[i] == ((uint8_t*) addr2)[i]);
 }
 
+// File utils wrapper functions
+int open_file(const std::string& path, nixlFileMode mode) {
```
yosefe, line 44: why do we need to expose file operations in python? python already has built in support for this

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,103 @@
...
+        switch (errno) {
+            case EACCES:
+                return NIXL_ERR_NOT_ALLOWED;
+            case EEXIST:
+                return NIXL_ERR_INVALID_PARAM;
+            case ENOENT:
+                return NIXL_ERR_NOT_FOUND;
+            default:
+                return NIXL_ERR_BACKEND;
+        }
```
yosefe: why do we need this wrapper? seems it just translates NIXL constants to/from OS constants

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,103 @@
...
+            case EBADF:
+                return NIXL_ERR_INVALID_PARAM;
+            default:
+                return NIXL_ERR_BACKEND;
+        }
+    }
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t nixlFileUtils::unlinkFile(const std::string& path) {
```
yosefe, line 197: can use std::filesystem instead?

test/unit/utils/file/file_utils_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+    status = nixlFileUtils::unlinkFile(test_file);
+    assert(status == NIXL_SUCCESS);
+
+    // Verify file no longer exists
+    status = nixlFileUtils::fileExists(test_file);
+    assert(status == NIXL_ERR_NOT_FOUND);
+
+    std::cout << "All file utils tests passed!" << std::endl;
+    return 0;
+}
```
yosefe: pls add new tests to gtest. AFAIU, currently this newly added test would not run in CI

examples/python/nixl_gds_example.py
vvenkates27: Will change them.

src/bindings/python/nixl_utils.cpp
vvenkates27: Agreed, for some platforms there were some translation errors reported by customers. This was to allow those customers to use these APIs too as a pass through. To avoid any descriptor translation errors.

src/utils/file/file_utils.cpp
vvenkates27: This wrapper has multiple purposes
1. To Enable opening/closing files from Plugins
2. To support passthorugh such that fds are opened without compatibility issues between RUST/Python/C++
3. To facilitate ease of testing

src/utils/file/file_utils.cpp
vvenkates27: Sure we can do that.

examples/python/nixl_gds_example.py
vvenkates27: Actually absl logger is only for C++ and not python, so cannot use them here.

test/unit/utils/file/file_utils_test.cpp
vvenkates27: Fixed to use gtest.

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+            default:
+                return NIXL_ERR_BACKEND;
+        }
+    }
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t nixlFileUtils::fileExists(const std::string& path) {
+    struct stat st;
+    if (stat(path.c_str(), &st) == -1) {
```
aranadive: Why not use std::filesystem - https://en.cppreference.com/w/cpp/filesystem

examples/python/nixl_gds_example.py
vvenkates27: @yosefe is it possbile to remove this change request since it is not possible to use absl for python currently?

src/bindings/python/nixl_utils.cpp
```diff
@@ -39,9 +41,61 @@ void verify_transfer(uintptr_t addr1, uintptr_t addr2, int size) {
...
+
+    py::enum_<nixlFileOperation>(m, "FileOperation")
+        .value("OPEN", nixlFileOperation::OPEN)
+        .value("CLOSE", nixlFileOperation::CLOSE)
+        .value("UNLINK", nixlFileOperation::UNLINK)
+        .value("STAT", nixlFileOperation::STAT)
+        .export_values();
+
+    m.def("open_file", &open_file, "Open a file with specified mode",
+          py::arg("path"), py::arg("mode"));
```
tstamler: py::arg only needs to be specified for optional arguments

src/bindings/python/nixl_utils.cpp
```diff
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 #include <pybind11/pybind11.h>
+#include <cassert>
```
tstamler: I don't see any asserts in this file (and there shouldn't be any) so should remove this include

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+    // Check if this is a direct I/O prefix (3-letter prefix ending with 'D')
+    bool use_direct = prefixed_path.length() >= 4 && prefixed_path[2] == 'D' && prefixed_path[3] == ':';
+
+    NIXL_DEBUG << absl::StrFormat("Opening with mode: %d, path: %s, direct I/O: %s",
+                                 static_cast<int>(mode), path, use_direct ? "Yes" : "No");
+
+    return openFile(path, mode, fd, use_direct);
+}
+
+std::tuple<bool, nixlFileMode, std::string> nixlFileUtils::parsePrefixedPath(const std::string& prefixed_path) {
```
tstamler, line 128: might be worth a typedef for this tuple, a lot going on here

src/bindings/python/meson.build
```diff
@@ -29,5 +29,6 @@ py.extension_module('_bindings',
 py.extension_module('_utils',
            'nixl_utils.cpp',
            subdir: ('nixl'),
-           dependencies: [pybind_dep],
+           dependencies: [pybind_dep, file_utils_dep],
+           include_directories: [nixl_inc_dirs, utils_inc_dirs],
```
tstamler, line 33: is the nixl_inc_dirs necessary or can we just have utils_inc_dirs?

src/bindings/python/nixl_utils.cpp
yosefe: do you mean that fd returned by python is not mapping to devid used by nixl? can you give a specific example?
maybe the translation should be vice-versa? if really needed
this PR adds a lot of "boiler plate" code and we need a really good reason to duplicate/wrap so many existing system services

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+            return ((mode == nixlFileMode::READ_WRITE_DIRECT || use_direct) ? O_DIRECT : 0) |
+                   O_RDWR | O_CREAT;
+
+        default:
+            return -1;
+    }
+}
+
+nixl_status_t nixlFileUtils::openFile(const std::string& path, nixlFileMode mode, int& fd, bool use_direct) {
+    int flags = getModeFlags(mode, use_direct);
```
yosefe, line 55: same here. what is the reason to define NIXL flags, error codes, etc for filesystem access?
maybe just need specific things that don't exist such as prefixedPath?
it adds more code we need to maintain, review, and potentially fix issues

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+    // Extract path (everything after the colon)
+    std::string path = prefixed_path.substr(colon_pos + 1);
+    if (path.empty()) {
+        NIXL_ERROR << absl::StrFormat("Invalid path: no path after prefix '%s:'", prefix);
+        return {false, nixlFileMode::READ_ONLY, ""};
+    }
+
+    // Map prefix to file mode
+    nixlFileMode mode;
+    if (prefix == "RD") {
```
yosefe, line 161: where are these prefixes defined and documented? is this part of user-facing API?
can we use unordered_map here?

src/utils/file/file_utils.cpp
yosefe: this comment is not resolved

yosefe (changes requested):

src/utils/file/file_utils.cpp
vvenkates27: Updated now.

src/utils/file/file_utils.cpp
vvenkates27: We need to handle filesystem open/close for the following reasons:
1. To Enable opening/closing files from Plugins (during register/der-register)
2. To support passthorugh such that fds are opened without compatibility issues between RUST/Python/C++
3. To facilitate ease of testing

src/utils/file/file_utils.cpp
vvenkates27: These will be part of optional metadata that user can specify
This was supposed to be part of a separate patch, came in here due to rebase issue. Nevertheless, it hasn't been documented yet, but will be introduced for file operations during register/deregister

src/bindings/python/nixl_utils.cpp
vvenkates27: removed

src/bindings/python/nixl_utils.cpp
vvenkates27: removed.

src/utils/file/file_utils.cpp
vvenkates27: Done.

src/bindings/python/meson.build
vvenkates27: it includes nixl_types.h so we will need nixl_inc_dirs.

src/utils/file/file_utils.cpp
vvenkates27: you mean to say you don't want NIXL error codes? Since its wrapped this was done for consistency and ease of use in NIXL plugins.

PR #280, opened 2025-05-08
GTEST/UCX: extend testing

evgeny-leksikov:
+ add NIXL_READ
 + add BASIC_XFER and LOAD_REMOTE_THEN_FAIL cases
 + related refactoring
 + extended ucx_status_to_nixl

@artpol84 FYI

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -33,11 +33,13 @@ static nixl_status_t ucx_status_to_nixl(ucs_status_t status)
     switch(status) {
     case UCS_INPROGRESS:
         return NIXL_IN_PROG;
+    case UCS_ERR_NOT_CONNECTED:
     case UCS_ERR_CONNECTION_RESET:
         return NIXL_ERR_REMOTE_DISCONNECT;
     case UCS_ERR_INVALID_PARAM:
         return NIXL_ERR_INVALID_PARAM;
     default:
+        NIXL_DEBUG << "Not converted UCX status: " << ucs_status_string(status);
         return NIXL_ERR_BACKEND;
     }
 }
--- a/test/gtest/ucx_tests.cpp
+++ b/test/gtest/ucx_tests.cpp
@@ -24,11 +24,6 @@ namespace gtest {
 namespace ucx {
 
 namespace nixl {
-    static std::unique_ptr<nixlAgent> createAgent(const std::string& name)
-    {
-        return std::make_unique<nixlAgent>(name, nixlAgentConfig(true));
-    }
-
     static nixlBackendH* createUcxBackend(nixlAgent& agent)
     {
         std::vector<nixl_backend_t> plugins;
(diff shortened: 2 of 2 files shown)


test/gtest/ucx_tests.cpp
```diff
@@ -49,168 +44,286 @@ namespace nixl {
...
+              << std::endl;
+}
+
+UcxTestFixture::~UcxTestFixture() {
+    setenv("NIXL_PLUGIN_DIR", m_plugin_dir_backup.c_str(), 1);
+    std::cout << "restore NIXL_PLUGIN_DIR: " << getenv("NIXL_PLUGIN_DIR")
+              << std::endl;
+}
+
+template<UcxTestFixture::TestType test_type, enum nixl_xfer_op_t op>
```
artpol84: Out of curiosity, what is the advantage of using the template vs. just passing these as arguments here?

artpol84:
Thanks @evgeny-leksikov!
This covers everything now!

src/utils/ucx/ucx_utils.cpp
```diff
@@ -33,11 +33,13 @@ static nixl_status_t ucx_status_to_nixl(ucs_status_t status)
     switch(status) {
     case UCS_INPROGRESS:
         return NIXL_IN_PROG;
+    case UCS_ERR_NOT_CONNECTED:
     case UCS_ERR_CONNECTION_RESET:
         return NIXL_ERR_REMOTE_DISCONNECT;
     case UCS_ERR_INVALID_PARAM:
         return NIXL_ERR_INVALID_PARAM;
     default:
+        NIXL_DEBUG << "Not converted UCX status: " << ucs_status_string(status);
```
yosefe: NIXL_WARN << "Unexpected UCX error "

yosefe (commented):
As a general comment, i think we should try to avoid backend specific tests as much as possible and write tests over nixlAgent API.
Can we extend test_transfer.cpp , for example?

test/gtest/ucx_tests.cpp
```diff
@@ -49,168 +44,286 @@ namespace nixl {
...
+} // namespace nixl
 
-    static nixl_status_t
-    wait_for_completion(nixlAgent& agent, nixlXferReqH* req_handle)
-    {
-        nixl_status_t status;
+class UcxTestFixture : public testing::Test {
+    class Agent {
+        struct MemDesc {
+            MemDesc() : m_dlist(DRAM_SEG), m_desc(), m_data(m_data_size) {}
```
gleon99: `= std::vector<std::byte>  m_data = std::vector<std::byte>(m_data_size); ` in line 75 instead?

test/gtest/ucx_tests.cpp
```diff
@@ -49,168 +44,286 @@ namespace nixl {
...
+        void destroy();
+        void fillRegList(nixl_xfer_dlist_t& dlist, nixlBasicDesc& desc) const;
+        std::string getLocalMD() const;
+        void loadRemoteMD(const std::string& remote_name);
+        nixl_status_t createXferReq(const nixl_xfer_op_t& op,
+                                    nixl_xfer_dlist_t& sReq_descs,
+                                    nixl_xfer_dlist_t& rReq_descs,
+                                    nixlXferReqH*& req_handle) const;
+        nixl_status_t postXferReq(nixlXferReqH* req_handle) const;
+        nixl_status_t wait_for_completion(nixlXferReqH* req_handle);
```
gleon99: Why mixed methods naming notation?
`wait_for_completion` vs. `postXferReq`

evgeny-leksikov:
> As a general comment, i think we should try to avoid backend specific tests as much as possible and write tests over nixlAgent API. Can we extend test_transfer.cpp , for example?

These tests were added in scope of adding the err handling feature to UCX backend, so I was focused on testing the UCX backend. We could extend testing as another activity with aligning the feature support in other backends. Also, I would propose to add specific variants to test each backend separately instead of default one.

test/gtest/ucx_tests.cpp
evgeny-leksikov: from functionality perspective - no difference, the main advantage - compile time optimizations, in this particular case helper functions which just convert one constant to other one will be completely compiled out.

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: fixed

yosefe:
> These tests were added in scope of adding the err handling feature to UCX backend, so I was focused on testing the UCX backend. We could extend testing as another activity with aligning the feature support in other backends. Also, I would propose to add specific variants to test each backend separately instead of default one.

Is the feature of error handling exposed somehow in nixlAgent API or could be tested on nixlAgent level?

evgeny-leksikov:
> Is the feature of error handling exposed somehow in nixlAgent API or could be tested on nixlAgent level?

All the testing is done at Agent level, the only UCX specific here is explicit UCX plugin loading.

test/gtest/ucx_tests.cpp
evgeny-leksikov: fixed

test/gtest/ucx_tests.cpp
evgeny-leksikov: fixed

yosefe:
> > Is the feature of error handling exposed somehow in nixlAgent API or could be tested on nixlAgent level?
> 
> All the testing is done at Agent level, the only UCX specific here is explicit UCX plugin loading.

So let's rename the test class and function names accordingly?

evgeny-leksikov:
> > > Is the feature of error handling exposed somehow in nixlAgent API or could be tested on nixlAgent level?
> > 
> > 
> > All the testing is done at Agent level, the only UCX specific here is explicit UCX plugin loading.
> 
> So let's rename the test class and function names accordingly?

Could you please be more specific? In current state the tests are working with UCX backend only, from my point of view the tests names correspond to what they are doing.

yosefe:
> > > > Is the feature of error handling exposed somehow in nixlAgent API or could be tested on nixlAgent level?
> > > 
> > > 
> > > All the testing is done at Agent level, the only UCX specific here is explicit UCX plugin loading.
> > 
> > 
> > So let's rename the test class and function names accordingly?
> 
> Could you please be more specific? In current state the tests are working with UCX backend only, from my point of view the tests names correspond to what they are doing.

see test_transfer.cpp for example

evgeny-leksikov:
> > Could you please be more specific? In current state the tests are working with UCX backend only, from my point of view the tests names correspond to what they are doing.
> 
> see test_transfer.cpp for example

take a look now

yosefe:
> > > Could you please be more specific? In current state the tests are working with UCX backend only, from my point of view the tests names correspond to what they are doing.
> > 
> > 
> > see test_transfer.cpp for example
> 
> take a look now

The test is still called "TestUcxBackend" and it's in the file "ucx_tests.cpp"
if there is nothing really UCX specific in this test i'd expect it to be called TestErrorHandling, for example. It could be instantiated only on UCX backend (for now)
Since the purpose of this PR is to extend testing for error handling (AFAIU) let's make it more generic

evgeny-leksikov:
> The test is still called "TestUcxBackend" and it's in the file "ucx_tests.cpp" if there is nothing really UCX specific in this test i'd expect it to be called TestErrorHandling, for example. It could be instantiated only on UCX backend (for now) Since the purpose of this PR is to extend testing for error handling (AFAIU) let's make it more generic

done

test/gtest/error_handling.cpp
```diff
@@ -0,0 +1,329 @@
...
+    m_mem.fillData();
+}
+
+bool TestErrorHandling::Agent::dataCmp(const TestErrorHandling::Agent& other) const {
+    return m_mem.m_data == other.m_mem.m_data;
+}
+
+TestErrorHandling::TestErrorHandling() {
+    // Set up test environment
+    m_plugin_dir_backup = getenv("NIXL_PLUGIN_DIR");
```
aranadive: If NIXL_PLUGIN_DIR is not set in env this will cause seg fault. So you first need to check if its available then set it to the string var. Similar on restore path you need to change how the original path is restored.

aranadive (approved):

test/gtest/error_handling.cpp
```diff
@@ -0,0 +1,338 @@
...
+                m_params = { .backends = {backend} };
+                nixl::fillRegList(m_dlist, m_desc, m_data);
+            }
+
+            void fillData() {
+                std::fill(m_data.begin(), m_data.end(), std::byte(std::rand()));
+            }
+
+            static constexpr size_t m_data_size = 256;
+            std::vector<std::byte>  m_data = std::vector<std::byte>(m_data_size);
```
brminich, line 77: maybe
`std::vector<std::byte> m_data{m_data_size};`

test/gtest/error_handling.cpp
```diff
@@ -0,0 +1,338 @@
...
+void TestErrorHandling::testXfer() {
+    m_Initiator.init("initiator");
+    m_Target.init("target");
+
+    exchangeMetaData();
+
+    for (size_t i = 0; i < numIter<test_type>(); ++i) {
+        nixlXferReqH* req_handle = postXfer(op, isFailure<test_type>(i));
+        nixl_status_t status     = m_Initiator.waitForCompletion(req_handle);
+
```
brminich, line 248: minor: could use auto type

brminich (commented):
👍 
just couple minor comments

test/gtest/error_handling.cpp
evgeny-leksikov: I don't think it's a good idea - it works unless type of vector element is not "convertable" to `size_t`

test/gtest/error_handling.cpp
evgeny-leksikov: I don't like `auto` for simple types, in many cases it disimproves readability

brminich (approved):

src/plugins/ucx/ucx_plugin.cpp
```diff
@@ -41,7 +41,8 @@ namespace
 
    [[nodiscard]] nixl_b_params_t get_backend_options() {
        return {
-	 { "ucx_devices", "" }
+        { "ucx_devices", "" },
+        { "ucx_error_handling_mode", "none" } // or "peer"
```
brminich, line 45: let's also add for `UcxMoPlugin`?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -117,41 +119,24 @@ nixlUcxEp::closeImpl(ucp_worker_h worker, ucp_ep_close_flags_t flags)
...
 
-        if (worker == nullptr) {
-            ucp_request_free(request);
-            eph = nullptr;
-            return NIXL_SUCCESS;
-        }
-        break;
+        ucp_request_free(request);
+        eph = nullptr;
+        return NIXL_SUCCESS;
```
brminich, line 124: do we need to wait for completion in certain cases? E.g. when worker is destroyed?

src/plugins/ucx/ucx_plugin.cpp
evgeny-leksikov: Let's add in next PR, then I can try to add it in gtest

PR #287, opened 2025-05-09
Refactor: Some C++ best practices.

ColinNV:


aranadive:
Thanks for the PR. @ColinNV could you add a bit more context in your comment about which C++ best practices are you using here?

ColinNV:
@aranadive Sure, let's see what we've got:

- Use a smart pointer and RAII instead of manual memory management (simplifies code and helps with exception safety.)
- Use the parameter initializer list instead of assigning member variables in the body of the constructor (idiomatic C++ and in sometimes more efficient.)
- Declare more functions as `const`, `noexcept`, `explicit` and/or `override` to increase readability and thereby maintainability (arguably `const` and `explicit` should be the default...), and in the case of `explicit` and `override` catch some mistakes at compile-time.
- Return "expensive" objects by const-reference from member "getter functions" as is usually done; leaves it up to the caller to make an expensive copy only when needed.

I believe that covers most of the changes except for a couple of code reformattings, mostly not putting a space in front of the opening bracket in a function declaration for consistency, and not putting brackets around a returned expression since `return` is not a function call and the brackets are superfluous.

Hope this answers the question :-)

src/api/cpp/backend/backend_engine.h
```diff
@@ -53,39 +54,35 @@ class nixlBackendEngine {
...
     public:
-        nixlBackendEngine (const nixlBackendInitParams* init_params)
-            : localAgent(init_params->localAgent) {
-
-            this->backendType  = init_params->type;
-            this->initErr      = false;
-            this->customParams = new nixl_b_params_t(*(init_params->customParams));
+        explicit nixlBackendEngine (const nixlBackendInitParams* init_params)
+            : backendType( init_params->type ),
+              customParams( new nixl_b_params_t( *(init_params->customParams) ) ),  // TODO: Use std::make_unique (C++14).
```
yosefe: use make_unique 
extra () around init_params

ColinNV: Will change to `make_unique` (though I'm still not clear why customParams is a pointer instead of directly the `nixl_b_params_t`).

w1ldptr: There are many places in NIXL where objects are manually allocated on heap and passed around as pointers instead of using them by-value or by-reference. We have been removing such cases and I suggest to also make customParams here a by-value class member instead of a (smart) pointer.

src/api/cpp/backend/backend_engine.h
```diff
@@ -26,87 +27,89 @@
...
         }
 
-        nixl_status_t getInitParam(const std::string &key, std::string &value) {
-            if (customParams->count(key)==0) {
+        [[nodiscard]] nixl_status_t getInitParam(const std::string &key, std::string &value) {
+            if (customParams.count(key)==0) {
                 return NIXL_ERR_INVALID_PARAM;
             } else {
-                value = (*customParams)[key];
+	        value = customParams.at(key);
```
w1ldptr: nit: looking up the iterator only once, checking whether it is equal to customParams.end() and, if not, assigning iterator->second to value would be more elegant (and would also allow you to add const modifier to this function I assume).

ColinNV: I wanted to concentrate mostly on the interface declarations, but you're right, while I'm already at it...

yosefe:
@ColinNV pls not conflicts must be resolved to run CI

ColinNV:
Now I'm getting an error about missing AWS credentials, not quite sure how to proceed:

https://github.com/ColinNV/nixl/actions/runs/14993139038/job/42121012201

src/api/cpp/backend/backend_engine.h
```diff
@@ -26,87 +27,87 @@
...
     protected:
         // Members that can be accessed by the child (localAgent cannot be modified)
-        bool              initErr;
+        bool              initErr = false;
         const std::string localAgent;
 
-        nixl_status_t setInitParam(const std::string &key, const std::string &value) {
-            if (customParams->count(key)==0) {
-                (*customParams)[key] = value;
+        [[nodiscard]] nixl_status_t setInitParam(const std::string &key, const std::string &value) {
```
brminich, line 38: since this is an API file can it break compilation of some app/client code?

src/api/cpp/backend/backend_engine.h
```diff
@@ -26,87 +27,87 @@
...
-        nixlBackendEngine (const nixlBackendInitParams* init_params)
-            : localAgent(init_params->localAgent) {
-
-            this->backendType  = init_params->type;
-            this->initErr      = false;
-            this->customParams = new nixl_b_params_t(*(init_params->customParams));
+        explicit nixlBackendEngine (const nixlBackendInitParams* init_params)
+            : backendType( init_params->type ),
+              customParams( *init_params->customParams ),
+              localAgent(init_params->localAgent) {
```
brminich: seems like different style is used in the project, without extra spaces in the brackets

src/api/cpp/backend/backend_engine.h
ColinNV: Yes, if they ignore the return value (which they shouldn't do (blindly)).

src/api/cpp/backend/backend_engine.h
ColinNV: Sorry, falling back to my personal style (which I find more readable).

src/api/cpp/backend/backend_engine.h
```diff
@@ -26,66 +27,66 @@
...
+        void operator=(const nixlBackendEngine&) = delete;
+
+        virtual ~nixlBackendEngine() = default;
 
-        bool getInitErr() { return initErr; }
-        nixl_backend_t getType () const { return backendType; }
-        nixl_b_params_t getCustomParams () const { return *customParams; }
+        bool getInitErr() const noexcept { return initErr; }
+        const nixl_backend_t& getType() const noexcept { return backendType; }
+        const nixl_b_params_t& getCustomParams() const noexcept { return customParams; }
```
brminich, line 71: why is it  better to expose private members by ref?

ColinNV: If you return by value then the object (in this case: container!) is always copied. When returning by const-reference the caller of the getter gets to decide whether it really needs a copy.

brminich: i was wondering about some kind of lifetime issues given that it is not a data path. However I do not have a strong opinion here, so ok.

brminich (approved):

PR #291, opened 2025-05-11
TEXT/XFER: Add fast gtest mode and data check

yosefe:
## Why
Allow executing gtest in fast mode (with --fast command line argument) along with adding data validation

## How
- Validate data in test_transfer using PRNG pattern
- Use absl::Flags for parsing gtest arguments


test/gtest/test_transfer.cpp
```diff
@@ -243,25 +294,50 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
-    constexpr std::array<std::tuple<size_t, size_t, size_t>, 3> test_cases = {
-        {{4096, 8, 3},
-         {32768, 64, 3},
-         {1000000, 100, 3}}
+    const std::vector<std::tuple<size_t, size_t, size_t>> test_cases      = {
+        {{4096, 8, 10000}, {32768, 64, 1000}, {1000000, 32, 100}}
+    };
+    const std::vector<std::tuple<size_t, size_t, size_t>> fast_test_cases = {
+        {{65536, 8, 1}}
     };
```
w1ldptr, line 337: This approach is a bit crude and non-configurable. How about adding CLI parameter like "testRunMultiplier" that will have small default value (so that the tests would run quickly by default) but could be changed by the caller to any amount of iterations they want?

test/gtest/test_transfer.cpp
```diff
@@ -51,12 +51,46 @@ class MemBuffer : std::shared_ptr<void> {
         return size;
```
w1ldptr, line 51: On line 37: when capturing mem_type (stack-allocated ctor parameter) by reference in lambda what would the reference point to by the time the class is being destroyed? Don't you need to capture by-value instead?

test/gtest/test_transfer.cpp
```diff
@@ -51,12 +51,46 @@ class MemBuffer : std::shared_ptr<void> {
...
     }
 
+    void fill(uint64_t seed)
+    {
+        for (auto &value : *this) {
```
w1ldptr, line 56: This whole approach of inheriting from a shared_ptr, then extending it with bunch of methods to allow iteration, etc. make the code a bit hard to read and understand. 
Instead, why not make MemBuffer a simple template with single specialization for nixl_mem_t::DRAM_SEG that would just be a wrapper over std::vector<uint64_t> (returning vector's iterators for begin(), end(), etc.) and all methods like fill, check, alignedSize, etc. - local static functions that are not members of the class?

test/gtest/test_transfer.cpp
yosefe: i'd expect different tests to have different ways to understand "fast mode". for example test for small messages could still run more iterations. since this is functionality tests and not a benchmark IMO controlling num of iterations of each test is overkill

test/gtest/test_transfer.cpp
yosefe: yes we can capture by value

test/gtest/test_transfer.cpp
w1ldptr: Hence "multiplier" in the parameter I proposed so you could set different base defaults (10x for small messages, 5x for intermediate, 1x for large) and multiple these by "testRunMultiplier". Anyway, I'm not really insisting on it, just thought that a factor parameter for number of iteration is more elegant approach than bunch of conditionals "if (fastMode())..." that our test code will soon be riddled with.

test/gtest/test_transfer.cpp
yosefe: yes, the problem here is that the fill/check methods will not be directly useful for GPU memory anyway. will try to refactor.

test/gtest/test_transfer.cpp
yosefe: should the multiplier affect also message sizes or just num of iterations ?

w1ldptr: I guess just the number of iterations. Hopefully 1 iteration of large message size shouldn't take too much time even on laptop?

aranadive:
@yosefe any update on this?

PR #293, opened 2025-05-12
UCX: Add debug logs for initialization failure

yosefe:
## Why
To debug failures like reported in https://nvidia.slack.com/archives/C08MF9LBA4E/p1746951585024519


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -418,6 +420,7 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
 
     if (init_params->enableProgTh) {
         if (!nixlUcxContext::mtLevelIsSupproted(NIXL_UCX_MT_WORKER)) {
+            NIXL_DEBUG << "UCX library does not support multi-threading";
```
brminich: should not it be a warning or at least info?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -432,6 +435,7 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     workerAddr = uw->epAddr(workerSize);
 
     if (workerAddr == nullptr) {
+        NIXL_DEBUG << "Failed to get UCX worker address";
```
brminich: should not it be a warning/info?

yosefe: fixed

src/plugins/ucx/ucx_backend.cpp
yosefe: fixed

brminich (approved):

PR #294, opened 2025-05-12
UCX: handle ucp_worker_create failure

evgeny-leksikov:
+ handle ucp_worker_create failure


src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,15 +406,15 @@ nixlUcxWorker::nixlUcxWorker(std::shared_ptr<nixlUcxContext> &_ctx): ctx(_ctx)
...
 
     status = ucp_worker_create(ctx->ctx, &worker_params, &worker);
-    if (status != UCS_OK)
-    {
-       // TODO: MSW_NET_ERROR(priv->net, "failed to create ucp_worker (%s)\n", ucs_status_string(status));
-        return;
+    if (status != UCS_OK) {
+        auto err_str = std::string("Failed to create UCX worker: ") +
+                       ucs_status_string(status);
+        NIXL_ERROR << err_str;
```
yosefe, line 404: IMO this PR should just add error logs and we need to introduce exception error handling as broader scope change

evgeny-leksikov: https://github.com/ai-dynamo/nixl/blob/498fe2ac0b36f7223c8d9c66508de814928602d3/src/utils/ucx/ucx_utils.cpp#L163

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,15 +406,15 @@ nixlUcxWorker::nixlUcxWorker(std::shared_ptr<nixlUcxContext> &_ctx): ctx(_ctx)
...
+        throw std::runtime_error(err_str);
```
brminich, line 405: maybe ` throw std::exception();` and no need for `err_str`, as you do not use it anyway

evgeny-leksikov: I used the same exception type as existing one for `ucp_ep_create`

evgeny-leksikov: Also `std::runtime_error` does not have default ctor

brminich: seems odd to generate an error string which you are not going to use. Can't you use default `std::exception`?

evgeny-leksikov: I can, but IMO better to have it aligned

evgeny-leksikov:
Merge issue, converted to draft to revert

src/utils/ucx/ucx_utils.cpp
yosefe: ok

gleon99 (approved):

brminich (approved):

PR #295, opened 2025-05-12
Mark subset of data path methods with 'const' modifier

w1ldptr:
In C++ marking method as 'const' doesn't automatically guarantee thread-safety but allows compiler to catch obvious shared state modifications which is a good first step towards making contributors consider thread safety when implementing new backends or modifying existing ones. This series marks subset of agent data path methods (reateXferReq, postXferReq, getXferStatus, releaseXferReq, genNotif) and the backend methods that they call (prepXfer, postXfer, checkXfer, releaseReqH, genNotif) as const, including laying the groundwork to make the change compile/work:

- Refactor GDS backend to extract mutable state into dedicated class that protects its data with mutex.
- Mooncake already implements its own locking. Make the lock mutable to allow taking it from const methods. Also, cosmetic change to use iterator to access connected_agents_ elements instead of operator[] to allow using it from const methods.
- Perform simple cosmetic changes in ucx and ucx_mo to use const references and const iterators from the data path methods to make them compatible with 'const' modifier.

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -254,7 +254,7 @@ class nixlAgent {
          * @return nixl_status_t NIXL_IN_PROG or error code if call was not successful
          */
         nixl_status_t
-        getXferStatus (nixlXferReqH* req_hndl);
+        getXferStatus (nixlXferReqH* req_hndl) const;
 
         /**
          * @brief  Query the backend associated with `req_hndl`. E.g., if for genNotif
@@ -276,7 +276,7 @@ class nixlAgent {
          * @return nixl_status_t Error code if call was not successful
          */
         nixl_status_t
-        releaseXferReq (nixlXferReqH* req_hndl);
+        releaseXferReq (nixlXferReqH* req_hndl) const;
 
         /**
          * @brief  Release the prepared descriptor list handle `dlist_hndl`
@@ -319,7 +319,7 @@ class nixlAgent {
         nixl_status_t
         genNotif (const std::string &remote_agent,
                   const nixl_blob_t &msg,
-                  const nixl_opt_args_t* extra_params = nullptr);
+                  const nixl_opt_args_t* extra_params = nullptr) const;
 
         /*** Metadata handling through side channel ***/
         /**
--- a/src/plugins/cuda_gds/gds_backend.cpp
+++ b/src/plugins/cuda_gds/gds_backend.cpp
@@ -172,7 +172,7 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        const nixl_meta_dlist_t &remote,
                                        const std::string &remote_agent,
                                        nixlBackendReqH* &handle,
-                                       const nixl_opt_b_args_t* opt_args)
+                                       const nixl_opt_b_args_t* opt_args) const
 {
     nixlGdsBackendReqH* gds_handle = new nixlGdsBackendReqH();
     size_t buf_cnt = local.descCount();
@@ -273,7 +273,7 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
     return NIXL_SUCCESS;
 }
 
-nixlGdsIOBatch* nixlGdsEngine::getBatchFromPool(unsigned int size) {
+nixlGdsIOBatch* nixlGdsEngine::getBatchFromPool(unsigned int size) const {
     const std::lock_guard<std::mutex> lock(batch_pool_lock);
     // Use a pre-allocated batch if available
     if (!batch_pool.empty()) {
@@ -286,7 +286,7 @@ nixlGdsIOBatch* nixlGdsEngine::getBatchFromPool(unsigned int size) {
     return nullptr;
 }
 
-void nixlGdsEngine::returnBatchToPool(nixlGdsIOBatch* batch) {
+void nixlGdsEngine::returnBatchToPool(nixlGdsIOBatch* batch) const {
     const std::lock_guard<std::mutex> lock(batch_pool_lock);
     // Only keep up to batch_pool_size batches
         batch_pool.push_back(batch);
@@ -297,7 +297,7 @@ nixl_status_t nixlGdsEngine::postXfer(const nixl_xfer_op_t &operation,
                                       const nixl_meta_dlist_t &remote,
                                       const std::string &remote_agent,
                                       nixlBackendReqH* &handle,
-                                      const nixl_opt_b_args_t* opt_args)
+                                      const nixl_opt_b_args_t* opt_args) const
 {
     nixlGdsBackendReqH* gds_handle = (nixlGdsBackendReqH*)handle;
 
@@ -334,7 +334,7 @@ nixl_status_t nixlGdsEngine::postXfer(const nixl_xfer_op_t &operation,
 
 nixl_status_t nixlGdsEngine::createAndSubmitBatch(const std::vector<GdsTransferRequestH>& requests,
                                                   size_t start_idx, size_t batch_size,
-                                                  std::vector<nixlGdsIOBatch*>& batch_list)
+                                                  std::vector<nixlGdsIOBatch*>& batch_list) const
 {
     nixlGdsIOBatch* batch = getBatchFromPool(batch_size);
     if (!batch) {
@@ -368,7 +368,7 @@ nixl_status_t nixlGdsEngine::createAndSubmitBatch(const std::vector<GdsTransferR
     return NIXL_SUCCESS;
 }
 
-nixl_status_t nixlGdsEngine::checkXfer(nixlBackendReqH* handle)
+nixl_status_t nixlGdsEngine::checkXfer(nixlBackendReqH* handle) const
 {
     nixlGdsBackendReqH *gds_handle = (nixlGdsBackendReqH *)handle;
 
@@ -396,7 +396,7 @@ nixl_status_t nixlGdsEngine::checkXfer(nixlBackendReqH* handle)
     return status;
 }
 
-nixl_status_t nixlGdsEngine::releaseReqH(nixlBackendReqH* handle)
+nixl_status_t nixlGdsEngine::releaseReqH(nixlBackendReqH* handle) const
 {
 
     nixlGdsBackendReqH *gds_handle = (nixlGdsBackendReqH *) handle;
--- a/src/plugins/cuda_gds/gds_backend.cpp
+++ b/src/plugins/cuda_gds/gds_backend.cpp
@@ -274,6 +274,7 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
 }
 
 nixlGdsIOBatch* nixlGdsEngine::getBatchFromPool(unsigned int size) {
+    const std::lock_guard<std::mutex> lock(batch_pool_lock);
     // Use a pre-allocated batch if available
     if (!batch_pool.empty()) {
         nixlGdsIOBatch* batch = batch_pool.back();
@@ -286,6 +287,7 @@ nixlGdsIOBatch* nixlGdsEngine::getBatchFromPool(unsigned int size) {
 }
 
 void nixlGdsEngine::returnBatchToPool(nixlGdsIOBatch* batch) {
+    const std::lock_guard<std::mutex> lock(batch_pool_lock);
     // Only keep up to batch_pool_size batches
         batch_pool.push_back(batch);
 }
--- a/src/plugins/cuda_gds/gds_backend.h
+++ b/src/plugins/cuda_gds/gds_backend.h
@@ -25,6 +25,7 @@
 #include <fcntl.h>
 #include <list>
 #include <vector>
+#include <mutex>
 #include "gds_utils.h"
 #include "backend/backend_engine.h"
 
@@ -87,6 +88,8 @@ class nixlGdsEngine : public nixlBackendEngine {
     private:
         gdsUtil *gds_utils;
         std::unordered_map<int, gdsFileHandle> gds_file_map;
+
+        std::mutex batch_pool_lock;
         std::list<nixlGdsIOBatch*> batch_pool;
         unsigned int batch_pool_size;  // Renamed from pool_size
         unsigned int batch_limit;      // Added for configurable batch limit
--- a/src/plugins/cuda_gds/gds_backend.h
+++ b/src/plugins/cuda_gds/gds_backend.h
@@ -89,17 +89,17 @@ class nixlGdsEngine : public nixlBackendEngine {
         gdsUtil *gds_utils;
         std::unordered_map<int, gdsFileHandle> gds_file_map;
 
-        std::mutex batch_pool_lock;
-        std::list<nixlGdsIOBatch*> batch_pool;
+        mutable std::mutex batch_pool_lock;
+        mutable std::list<nixlGdsIOBatch*> batch_pool;
         unsigned int batch_pool_size;  // Renamed from pool_size
         unsigned int batch_limit;      // Added for configurable batch limit
         unsigned int max_request_size; // Added for configurable request size
 
-        nixlGdsIOBatch* getBatchFromPool(unsigned int size);
-        void returnBatchToPool(nixlGdsIOBatch* batch);
+        nixlGdsIOBatch* getBatchFromPool(unsigned int size) const;
+        void returnBatchToPool(nixlGdsIOBatch* batch) const;
         nixl_status_t createAndSubmitBatch(const std::vector<GdsTransferRequestH>& requests,
                                            size_t start_idx, size_t batch_size,
-                                           std::vector<nixlGdsIOBatch*>& batch_list);
+                                           std::vector<nixlGdsIOBatch*>& batch_list) const;
         nixl_status_t createBatches(const nixl_xfer_op_t &operation,
                                    const nixl_meta_dlist_t &local,
                                    const nixl_meta_dlist_t &remote,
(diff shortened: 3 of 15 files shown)


src/api/cpp/nixl.h
```diff
@@ -254,7 +254,7 @@ class nixlAgent {
          * @return nixl_status_t NIXL_IN_PROG or error code if call was not successful
          */
         nixl_status_t
-        getXferStatus (nixlXferReqH* req_hndl);
+        getXferStatus (nixlXferReqH* req_hndl) const;
```
yosefe, line 257: does it mean this function is not allowed to poll the network, or polling the network (including update QP/CQ state for example) is not considered changing a shared state ?

w1ldptr: 'Const' modifier on a method in C++ only guarantees that 'this' object representation is not modified by the method. For example, accessing elements from unordered map member with operator[] is not allowed in const method because operator[] creates new element with the provided key before returning it in case the element doesn't exist (hence changing 'connected_agents_[remote_agent].segment_id' to 'agent->second.segment_id' in another place in this patch set).

In this particular function accessing the engine and modifying its state is allowed because it is a field of nixlXferReqH req_hndl, not of nixlAgent instance 'this':
```
        req_hndl->status = req_hndl->engine->checkXfer(
                                     req_hndl->backendHandle);
```

Then we mark checkXfer itself with 'const' so, for example, it can't change remoteConnMap in ucx_backend but it can still do things like the following because 1. I already refactored remoteConnMap accesses to use iterator find instead of operator[] in previous series 2. getEp() returns a pointer to an object so changing that object doesn't change nixlUcxEngine instance:
```
    ret = search->second->getEp(worker_id)->sendAm(NOTIF_STR,
                                                   &hdr, sizeof(struct nixl_ucx_am_hdr),
                                                   (void*)buffer->data(), buffer->size(),
                                                   flags, req);
```

Overall const methods in C++ can be a little confusing at first and don't provide any strong guarantees with regards to "deep" constness, but they are still good for preventing obvious unsafe things like described above.

yosefe: Thanks. I see most methods in nixlAgent already marked as const according to this reasoning. should genNotif also be const?

w1ldptr: Not sure about genNotif since I didn't audit it for the purpose of rw locking.

yosefe: > Not sure about genNotif since I didn't audit it for the purpose of rw locking.

maybe we should add it to make the PR more complete? Seems all other API methods are marked properly as const when possible.

w1ldptr: done

src/plugins/ucx_mo/ucx_mo_backend.cpp
```diff
@@ -507,11 +507,11 @@ nixlUcxMoEngine::prepXfer (const nixl_xfer_op_t &operation,
     }
 
     // Check that remote agent is known
-    remote_comm_it_t it = remoteConnMap.find(remote_agent);
+    remote_comm_const_it_t it = remoteConnMap.find(remote_agent);
```
yosefe: auto ?

src/plugins/ucx_mo/ucx_mo_backend.h
```diff
@@ -100,6 +100,7 @@ class nixlUcxMoEngine : public nixlBackendEngine {
     // Map of agent name to saved nixlUcxConnection info
     using remote_conn_map_t = std::map<std::string, nixlUcxMoConnection>;
     using remote_comm_it_t = remote_conn_map_t::iterator;
+    using remote_comm_const_it_t = remote_conn_map_t::const_iterator;
```
yosefe: maybe remove it

test/gtest/mocks/mock_dram_engine.cpp
```diff
@@ -51,8 +51,8 @@ nixl_status_t MockDramBackendEngine::prepXfer(const nixl_xfer_op_t &operation,
                                              const nixl_meta_dlist_t &remote,
                                              const std::string &remote_agent,
                                              nixlBackendReqH *&handle,
-                                             const nixl_opt_b_args_t *opt_args) {
-  sharedState++;
+                                             const nixl_opt_b_args_t *opt_args) const {
+  assert(sharedState > 0);
```
yosefe, line 55: maybe return failure / EXPECT_EQ instead of assert?

ColinNV: Isn't this an internal class invariant (for which assert is often appropriate?)

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
+}
+
+nixlGdsIOBatch *nixlGdsBatchPool::getBatch(unsigned int size) {
+    const std::lock_guard<std::mutex> lock(batchPoolLock);
+
+    // Use a pre-allocated batch if available
+    if (!batchPool.empty()) {
+        nixlGdsIOBatch* batch = batchPool.back();
+        batchPool.pop_back();
+        batch->reset();
```
ColinNV: The reset function can be called outside of the mutex lock.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
+};
+
+nixlGdsBatchPool::nixlGdsBatchPool(unsigned int poolSize_, unsigned int batchLimit_): poolSize(poolSize_), batchLimit(batchLimit_)  {
+        for (unsigned int i = 0; i < poolSize; i++)
+            batchPool.push_back(new nixlGdsIOBatch(batchLimit));
+}
+
+nixlGdsBatchPool::~nixlGdsBatchPool() {
+    for (auto* batch : batchPool)
+        delete batch;
```
ColinNV: Please use `std::unique_ptr` in `batchPool` and get rid of the destructor.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
 /** Create a batch pool of size 32 */
 #define DEFAULT_BATCH_POOL_SIZE 32
 
+class nixlGdsBatchPool {
+public:
+    nixlGdsBatchPool(unsigned int poolSize_, unsigned int batchLimit_);
+    ~nixlGdsBatchPool();
+
+    nixlGdsIOBatch *getBatch(unsigned int size);
```
ColinNV: This could return a smart pointer that automatically returns the object to the pool, though that would add some overhead.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
+    const std::lock_guard<std::mutex> lock(batchPoolLock);
+
+    // Only keep up to batch_pool_size batches
+    batchPool.push_back(batch);
+}
+
 nixlGdsEngine::nixlGdsEngine(const nixlBackendInitParams* init_params)
     : nixlBackendEngine(init_params)
 {
     gds_utils = new gdsUtil();
```
ColinNV: The initialization should be in the member initializer list.

src/plugins/cuda_gds/gds_backend.h
```diff
@@ -83,20 +84,18 @@ class nixlGdsBackendReqH : public nixlBackendReqH {
         }
 };
 
+class nixlGdsBatchPool;
+
 class nixlGdsEngine : public nixlBackendEngine {
     private:
+        std::unique_ptr<nixlGdsBatchPool> batch_pool;
         gdsUtil *gds_utils;
```
ColinNV, line 89: This looks like it should be a `std::unique_ptr`, too.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -227,15 +227,16 @@ nixl_status_t nixlMooncakeEngine::postXfer (const nixl_xfer_op_t &operation,
...
                                             nixlBackendReqH* &handle,
-                                            const nixl_opt_b_args_t* opt_args)
+                                            const nixl_opt_b_args_t* opt_args) const
 {
     auto priv = (nixlMooncakeBackendReqH *) handle;
     int segment_id;
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        if (!connected_agents_.count(remote_agent))
+        auto agent = connected_agents_.find(remote_agent);
```
ColinNV: Variable `agent` can be `const`?

src/api/cpp/nixl.h
ColinNV: > 'Const' modifier on a method in C++ only guarantees that 'this' object representation is not modified by the method.
> [...]
> Overall const methods in C++ can be a little confusing at first and don't provide any strong guarantees with regards to "deep" constness, but they are still good for preventing obvious unsafe things like described above.

`Const` greatly increases code readability, which alone makes it worthwhile to use whenever possible, even when it's usefulness for thread safety is - unfortunately - less than one might initially hope.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
+    void putBatch(nixlGdsIOBatch *batch);
+
+    unsigned int getBatchLimit() const {
+        return batchLimit;
+    }
+private:
+    const unsigned int poolSize;
+    const unsigned int batchLimit;
+    std::mutex batchPoolLock;
+    std::list<nixlGdsIOBatch*> batchPool;
```
ColinNV: This should be `std::vector` as only `push_back` and `pop_back` are used.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
+};
+
+nixlGdsBatchPool::nixlGdsBatchPool(unsigned int poolSize_, unsigned int batchLimit_): poolSize(poolSize_), batchLimit(batchLimit_)  {
+        for (unsigned int i = 0; i < poolSize; i++)
```
ColinNV: If `batchPool` is changed to a `std::vector` this can call `reserve` before filling.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
...
 
+class nixlGdsBatchPool {
+public:
+    nixlGdsBatchPool(unsigned int poolSize_, unsigned int batchLimit_);
+    ~nixlGdsBatchPool();
+
+    nixlGdsIOBatch *getBatch(unsigned int size);
+    void putBatch(nixlGdsIOBatch *batch);
+
+    unsigned int getBatchLimit() const {
```
ColinNV: Should also be `[[nodiscard]]` and `noexcept`. The `int` is redundant (and why are we using `unsigned` instead of `std::size_t`)?

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -27,14 +28,63 @@
 /** Create a batch pool of size 32 */
 #define DEFAULT_BATCH_POOL_SIZE 32
 
+class nixlGdsBatchPool {
+public:
+    nixlGdsBatchPool(unsigned int poolSize_, unsigned int batchLimit_);
+    ~nixlGdsBatchPool();
+
```
ColinNV: This class could be made noncopyable by `delete`-ing the usual copy constructors and assignment operators.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -80,11 +130,7 @@ nixlGdsEngine::nixlGdsEngine(const nixlBackendInitParams* init_params)
         return;
     }
 
-    // Initialize the batch pool
-    for (unsigned int i = 0; i < batch_pool_size; i++) {
-        batch_pool.push_back(new nixlGdsIOBatch(batch_limit));
-    }
-
+    batch_pool = std::make_unique<nixlGdsBatchPool>(batch_pool_size, batch_limit);
```
ColinNV: The code to determine the two parameters should be factored out into separate functions (or similar), then `batch_pool` can be initialized in the member initializer list, at which point we can (probably?) drop the `unique_ptr` and keep it by-value.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: As opposed to UCX engine in previous series I'm explicitly _not_ attempting any kind of large scale refactoring in GDS. They tests are too sparse, have to be manually executed on cluster environment and coverage is completely missing from CI which makes me not comfortable doing it. Instead, I'm extracting _exactly_ what they have into a standalone class and adding a lock on top of it.

ColinNV: Understood.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: There are multiple error paths in that code, so I can't move it out to a helper until we remove that ugly initErr thing and use exceptions instead.

ColinNV: We should coordinate on that because I'd like to get that change done ASAP.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Having simple lock guard for the whole function body is straightforward highly-maintainable easy to read code. I'm not comfortable messing with it for imaginary micro-optimization that would never show up in any actual profiling data.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Again, I'm not attempting any large-scale GDS engine refactoring here, only extracting shared data path state into a basic helper class with a lock.

test/gtest/mocks/mock_dram_engine.cpp
w1ldptr: This is just a mock class for my unit tests. All const members assert on shared state so that tsan would see it as a 'read' and all non-const members increment sharedState so that tsan would detect it as a 'write'.

src/plugins/cuda_gds/gds_backend.cpp
ColinNV: Inverting the logic of the `if` would make it not too bad, but I agree that since `reset()` is cheap it's not necessary.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: > Should also be `[[nodiscard]]` and `noexcept`.

Ack.

> The `int` is redundant 

Some code guidelines prescribe always writing int after 'unsigned', etc. Here authors chose to add redundant 'int' if you check the definition of previous 'batch_limit' class variable that this methods substitutes. I don't personally have strong preference either way.

> (and why are we using `unsigned` instead of `std::size_t`)?

Just copied verbatim what GDS authors used. Looking at other GDS APIs it seems that this is unsigned int because nixlGdsIOBatch ctor accepts unsigned int.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: I would prefer this series to be merged first since the following RW lock series relies on it.

w1ldptr: (BTW I strongly agree that moving to exceptions in ctors should be done sooner rather than later)

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: 'batch' objects are exposed through APIs of this class. I don't see any straightforward way to just wrap them in unique_ptr without refactoring whole GDS engine for that.

ColinNV: You can use plain pointers to the outside but still use unique pointers internally.

w1ldptr: How? When we remove the batch in getBatch() function we also remove it from batchPool list. Now, if batchPool list holds unique ptrs instead of regular pointers, then at this point the underlying object gets deallocated unless unique ptr ownership is moved somewhere else instead. What am I missing?

ColinNV: The unique pointer has a release function:

https://en.cppreference.com/w/cpp/memory/unique_ptr/release

w1ldptr: And then you initialize the new one when user returns the object back? What is the point of it in such case?

ColinNV: Yes, at least you are "clean" inside the pool class, can get rid of the destructor, and don't leak memory if an allocation (other than the first) fails in the loop in the constructor.

yosefe (approved):

w1ldptr:
Similar to extracting UCX backend progress function refactoring to a standalone PR from UCX multiworker series I've also simplified GDS change to the bare minimum in order to concentrate on the actual purpose of this change.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.cpp
w1ldptr: Extracted nixlGdsBatchPool class out of this PR.

src/plugins/cuda_gds/gds_backend.h
w1ldptr: Let's keep large scale GDS refactoring separate.

w1ldptr:
> There are two stray `#include <mutex>` that could presumably be removed.

Thanks for noticing. Removed the redundant include.

artpol84 (approved):
UCX part looks great to me!

GDS changes are a bit concerning as these "const" changes bring a bit dangerous "comfort" when you start marking members as `mutable`.

On the other hand, it has a good side as now only limited set of members can be changed and it is easier to control. Probably this outweighs the aforementioned concern.

ColinNV:
> GDS changes are a bit concerning as these "const" changes bring a bit dangerous "comfort" when you start marking members as `mutable`.

As long as it's only mutexes that are marked mutable I wouldn't be too concerned; everything else should be looked at very carefully, agree.

PR #297, opened 2025-05-12
Refactor UCX progress thread to remove busy-polling

w1ldptr:
Instead of busy-polling use functionality provided by UCX that exposes file descriptors that can be monitored for events with standard UNIX APIs. Implement following changes:

 - When progress thread is enabled also enable UCP feature that exposes file descriptors for wake up.

 - In UCX progress thread arm every worker for wake up via file descriptor and sleep while waiting for events with poll().

 - Use eventfd file descriptor to wake up the thread for cancellation.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -18,8 +18,11 @@
 #include "ucx_backend.h"
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
+#include "common/nixl_log.h"
 
 #include <optional>
+#include <string.h>
+#include <unistd.h>
 #include "absl/strings/numbers.h"
 
 #ifdef HAVE_CUDA
@@ -28,8 +31,6 @@
 #include <cufile.h>
 
 #endif
-
-static constexpr int const noSyncIters = 32;
 
 /****************************************
  * CUDA related code
@@ -359,34 +360,50 @@ void nixlUcxEngine::progressFunc()
     }
     pthrActiveCV.notify_one();
 
+    // Set POLLIN event on all worker fds so that the main loop would process them all on first iteration
+    for (size_t wid = 0; wid < pollFds.size() - 1; wid++)
+        pollFds[wid].revents = POLLIN;
+
+    bool pthrStop = false;
     while (!pthrStop) {
-        int i;
-        for(i = 0; i < noSyncIters; i++) {
-            for (auto &uw: uws)
-                uw->progress();
+        for (size_t wid = 0; wid < pollFds.size() - 1; wid++) {
+            if (!(pollFds[wid].revents & POLLIN))
+                continue;
+            pollFds[wid].revents = 0;
+
+            bool made_progress = false;
+            ucs_status_t status = UCS_INPROGRESS;
+            const auto &uw = uws[wid];
+            do {
+                while (uw->progress())
+                    made_progress = true;
+
+                status = ucp_worker_arm(uw->getWorker());
+            } while (status == UCS_ERR_BUSY);
+            NIXL_ASSERT(status == UCS_OK);
+
+            if (made_progress && !wid)
+                notifProgress();
         }
-        notifProgress();
-        // TODO: once NIXL thread infrastructure is available - move it there!!!
 
-        // {
-        //     static uint64_t cnt = 0;
-        //     if ( !(cnt % 1000000)) {
-        //         std::cout << "Progress round" << std::endl;
-        //     }
-        //     cnt++;
-        // }
+        while (poll(pollFds.data(), pollFds.size(), -1) <= 0)
+            NIXL_ERROR << "Call to poll() was interrupted, retrying. Error: " << strerror(errno);
 
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+        if (pollFds.back().revents & POLLIN) {
+            pollFds.back().revents = 0;
+
+            char signal;
+            int ret = read(pollFds.back().fd, &signal, sizeof(signal));
+            if (ret < 0)
+                NIXL_ERROR << "read() on control pipe failed. Error: " << strerror(errno);
+
+            pthrStop = true;
         }
     }
 }
 
 void nixlUcxEngine::progressThreadStart()
 {
-    pthrStop = false;
     {
         std::unique_lock<std::mutex> lock(pthrActiveLock);
         pthrActive = false;
@@ -410,7 +427,11 @@ void nixlUcxEngine::progressThreadStop()
         return;
     }
 
-    pthrStop = true;
+    const char signal = 'X';
+    int ret = write(pthrControlPipe[1], &signal, sizeof(signal));
+    if (ret < 0)
+        NIXL_ERROR << "write to progress thread control pipe failed, error: "
+                   << strerror(errno);
     pthr.join();
 }
 
@@ -431,11 +452,19 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     nixl_b_params_t* custom_params = init_params->customParams;
 
     if (init_params->enableProgTh) {
+        pthrOn = true;
         if (!nixlUcxContext::mtLevelIsSupproted(NIXL_UCX_MT_WORKER)) {
             NIXL_ERROR << "UCX library does not support multi-threading";
             this->initErr = true;
             return;
         }
+        if (pipe(pthrControlPipe) < 0) {
+            NIXL_ERROR << "Couldn't create progress thread control pipe, error: " << strerror(errno);
+            this->initErr = true;
+            return;
+        }
+    } else {
+        pthrOn = false;
     }
 
     if (custom_params->count("device_list")!=0)
@@ -446,7 +475,8 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         numWorkers = 1;
 
     uc = std::make_shared<nixlUcxContext>(devs, sizeof(nixlUcxIntReq),
-                                          _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER);
+                                          _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER,
+                                          pthrOn);
     for (unsigned int i = 0; i < numWorkers; i++)
         uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));
 
@@ -459,16 +489,24 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         return;
     }
 
+    if (pthrOn) {
+        for (auto &uw: uws) {
+            int fd;
+            ucs_status_t ret = ucp_worker_get_efd(uw->getWorker(), &fd);
+            if (ret != UCS_OK) {
+                NIXL_ERROR << "Couldn't obtain fd for a worker, status: " << ucs_status_string(ret);
+                initErr = true;
+                return;
+            }
+
+            pollFds.push_back({fd, POLLIN, 0});
+        }
+        pollFds.push_back({pthrControlPipe[0], POLLIN, 0});
+    }
+
     uw->regAmCallback(CONN_CHECK, connectionCheckAmCb, this);
     uw->regAmCallback(DISCONNECT, connectionTermAmCb, this);
     uw->regAmCallback(NOTIF_STR, notifAmCb, this);
-
-    if (init_params->enableProgTh) {
-        pthrOn = true;
-        pthrDelay = init_params->pthrDelay;
-    } else {
-        pthrOn = false;
-    }
 
     // Temp fixup
     if (getenv("NIXL_DISABLE_CUDA_ADDR_WA")) {
@@ -499,6 +537,10 @@ nixlUcxEngine::~nixlUcxEngine () {
     }
 
     progressThreadStop();
+    if (pthrOn) {
+        close(pthrControlPipe[0]);
+        close(pthrControlPipe[1]);
+    }
     vramFiniCtx();
 }
 
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -318,7 +318,8 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
                                size_t req_size,
                                nixlUcxContext::req_cb_t init_cb,
                                nixlUcxContext::req_cb_t fini_cb,
-                               nixl_ucx_mt_t __mt_type)
+                               nixl_ucx_mt_t __mt_type,
+                               bool prog_thread)
 {
     ucp_params_t ucp_params;
     ucp_config_t *ucp_config;
@@ -328,6 +329,9 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
 
     ucp_params.field_mask = UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_MT_WORKERS_SHARED;
     ucp_params.features = UCP_FEATURE_RMA | UCP_FEATURE_AMO32 | UCP_FEATURE_AMO64 | UCP_FEATURE_AM;
+    if (prog_thread)
+        ucp_params.features |= UCP_FEATURE_WAKEUP;
+
     switch(mt_type) {
     case NIXL_UCX_MT_SINGLE:
         ucp_params.mt_workers_shared = 0;
--- a/test/unit/plugins/mooncake/mooncake_backend_test.cpp
+++ b/test/unit/plugins/mooncake/mooncake_backend_test.cpp
@@ -658,7 +658,10 @@ int main()
 #ifdef HAVE_CUDA
     int dev_ids[2] = { 0 , 0 };
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
+        n_vram_dev = 0;
+    }
 
     std::cout << "Detected " << n_vram_dev << " CUDA devices" << std::endl;
     if (n_vram_dev > 1) {
--- a/test/unit/plugins/ucx/ucx_backend_test.cpp
+++ b/test/unit/plugins/ucx/ucx_backend_test.cpp
@@ -688,7 +688,10 @@ int main()
 #ifdef HAVE_CUDA
     int dev_ids[2] = { 0 , 0 };
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
+        n_vram_dev = 0;
+    }
 
     std::cout << "Detected " << n_vram_dev << " CUDA devices" << std::endl;
     if (n_vram_dev > 1) {
--- a/test/unit/plugins/ucx_mo/ucx_mo_backend_test.cpp
+++ b/test/unit/plugins/ucx_mo/ucx_mo_backend_test.cpp
@@ -562,7 +562,10 @@ int main()
 int ndevices = NUM_WORKERS;
 #ifdef HAVE_CUDA
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
+        n_vram_dev = 0;
+    }
     std::cout << "Detected " << n_vram_dev << " CUDA devices" << std::endl;
 #endif
 
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -25,6 +25,7 @@
 #include <memory>
 #include <condition_variable>
 #include <atomic>
+#include <poll.h>
 
 #include "nixl.h"
 #include "backend/backend_engine.h"
@@ -114,10 +115,10 @@ class nixlUcxEngine : public nixlBackendEngine {
         std::mutex pthrActiveLock;
         std::condition_variable pthrActiveCV;
         bool pthrActive;
-        std::atomic_bool pthrStop;
         bool pthrOn;
         std::thread pthr;
-        nixlTime::us_t pthrDelay;
+        int pthrControlPipe[2];
+        std::vector<pollfd> pollFds;
 
         /* CUDA data*/
         std::unique_ptr<nixlUcxCudaCtx> cudaCtx;
--- a/src/utils/ucx/ucx_utils.h
+++ b/src/utils/ucx/ucx_utils.h
@@ -131,7 +131,7 @@ public:
     using req_cb_t = void(void *request);
     nixlUcxContext(std::vector<std::string> devices,
                    size_t req_size, req_cb_t init_cb, req_cb_t fini_cb,
-                   nixl_ucx_mt_t mt_type);
+                   nixl_ucx_mt_t mt_type, bool prog_thread);
     ~nixlUcxContext();
 
     static bool mtLevelIsSupproted(nixl_ucx_mt_t mt_type);
@@ -169,6 +169,9 @@ public:
 
     void reqRelease(nixlUcxReq req);
     void reqCancel(nixlUcxReq req);
+
+    /* Worker access */
+    ucp_worker_h getWorker() const { return worker; }
 };
 
 #endif
--- a/test/unit/utils/ucx/ucx_am_test.cpp
+++ b/test/unit/utils/ucx/ucx_am_test.cpp
@@ -88,8 +88,8 @@ int main()
     devs.push_back("mlx5_0");
 
     std::shared_ptr<nixlUcxContext> c[2] = {
-        std::make_shared<nixlUcxContext>(devs, 0, nullptr, nullptr, NIXL_UCX_MT_SINGLE),
-        std::make_shared<nixlUcxContext>(devs, 0, nullptr, nullptr, NIXL_UCX_MT_SINGLE)
+        std::make_shared<nixlUcxContext>(devs, 0, nullptr, nullptr, NIXL_UCX_MT_SINGLE, false),
+        std::make_shared<nixlUcxContext>(devs, 0, nullptr, nullptr, NIXL_UCX_MT_SINGLE, false)
     };
 
     nixlUcxWorker w[2] = {
--- a/test/unit/utils/ucx/ucx_worker_test.cpp
+++ b/test/unit/utils/ucx/ucx_worker_test.cpp
@@ -86,8 +86,8 @@ int main()
     // in CI it would be goot to test both SHM and IB
     //devs.push_back("mlx5_0");
     std::shared_ptr<nixlUcxContext> c[2] = {
-        std::make_shared<nixlUcxContext>(devs, sizeof(requestData), nixlUcxRequestInit, nullptr, NIXL_UCX_MT_SINGLE),
-        std::make_shared<nixlUcxContext>(devs, sizeof(requestData), nixlUcxRequestInit, nullptr, NIXL_UCX_MT_SINGLE)
+        std::make_shared<nixlUcxContext>(devs, sizeof(requestData), nixlUcxRequestInit, nullptr, NIXL_UCX_MT_SINGLE, false),
+        std::make_shared<nixlUcxContext>(devs, sizeof(requestData), nixlUcxRequestInit, nullptr, NIXL_UCX_MT_SINGLE, false)
     };
 
     nixlUcxWorker w[2] = {


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
     }
     pthrActiveCV.notify_one();
 
+    std::vector<pollfd> poll_fds;
+    for (auto &uw: uws) {
+        int fd;
+        if (ucp_worker_get_efd(uw->getWorker(), &fd) == UCS_OK)
+            poll_fds.push_back({fd, POLLIN, 0});
+        else
+            NIXL_ERROR << "Couldn't obtain fd for a worker";
```
yosefe: would the progress thread switch to busy polling? or need to make it fatal error?
also we could keep poll_fds on the engine instance and init them before starting the thread

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
...
+    std::vector<bool> active(poll_fds.size() - 1, true);
     while (!pthrStop) {
-        int i;
-        for(i = 0; i < noSyncIters; i++) {
-            for (auto &uw: uws)
-                uw->progress();
+        for (size_t wid = 0; wid < active.size(); wid++) {
+            if (!active[wid])
+                continue;
+            active[wid] = false;
```
yosefe: seems that if ucp_worker_get_efd failed the indexes in uws and poll_fds may not match

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
...
+
+            bool made_progress = false;
+            ucs_status_t status = UCS_INPROGRESS;
+            const auto &uw = uws[wid];
+            while (status != UCS_OK) {
```
yosefe: minor: IMO do..while is better fit here

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
...
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        while (poll(poll_fds.data(), poll_fds.size(), -1) <= 0) {
+            // Any other error would indicate bug in UCX
+            NIXL_ASSERT(!errno || errno == ENOMEM || errno == EINTR || errno == EAGAIN);
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
yosefe: why need sleep?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
...
-        // }
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        while (poll(poll_fds.data(), poll_fds.size(), -1) <= 0) {
+            // Any other error would indicate bug in UCX
+            NIXL_ASSERT(!errno || errno == ENOMEM || errno == EINTR || errno == EAGAIN);
```
yosefe: can we print NIXL_ERROR with errno details?
and if EAGAIN/EINTR i'd also print NIXL_DEBUG or NIXL_TRACE

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +358,48 @@ void nixlUcxEngine::progressFunc()
...
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        }
+
+        for (size_t wid = 0; wid < active.size(); wid++) {
+            if (poll_fds[wid].revents & POLLIN) {
+                active[wid] = true;
```
yosefe: do we really need "active vector"? what if we initialized revents to POLLIN before the while (!pthrStop) loop? and set revents=0 in the loop that calls ucp_worker_arm?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -409,6 +431,7 @@ void nixlUcxEngine::progressThreadStop()
     }
 
     pthrStop = true;
+    while (write(pthrStopFds[1], "x", 1) != 1);
```
yosefe: what if pipe is full? maybe no need to block?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: To allow kernel clean up some memory. If we got to the place where poll() returns ENOMEM or EAGAIN, then some low priority processes are likely being killed during that time.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: agree

src/plugins/ucx/ucx_backend.cpp
w1ldptr: > we could keep poll_fds on the engine instance and init them before starting the thread

ack

src/plugins/ucx/ucx_backend.cpp
w1ldptr: agree, I'll refactor it

w1ldptr: refactored it to use eventfd which is a better fit for this

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
w1ldptr: done

src/plugins/ucx/ucx_backend.cpp
w1ldptr: moved poll_fds init into ctor

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +359,37 @@ void nixlUcxEngine::progressFunc()
     }
     pthrActiveCV.notify_one();
 
+    // Set POLLIN event on all fd so that the main loop would process them all on first iteration
+    for (auto &pfd: pollFds)
+            pfd.revents = POLLIN;
```
iyastreb: alignment

src/utils/ucx/ucx_utils.cpp
```diff
@@ -318,7 +318,8 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
                                size_t req_size,
                                nixlUcxContext::req_cb_t init_cb,
                                nixlUcxContext::req_cb_t fini_cb,
-                               nixl_ucx_mt_t __mt_type)
+                               nixl_ucx_mt_t __mt_type,
```
iyastreb, line 321: I think it should be named mt_type, like in declaration?

src/plugins/ucx/ucx_backend.cpp
iyastreb: maybe std::this_thread::yield() then?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -357,27 +359,37 @@ void nixlUcxEngine::progressFunc()
...
-        //     }
-        //     cnt++;
-        // }
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        while (poll(pollFds.data(), pollFds.size(), -1) <= 0) {
```
iyastreb: I admit I know little about NIXL design
Wouldn't be sufficient here to use a blocking queue? (mutex+cond var+queue)
What is the rational for using poll+fds?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I've been removing "yield" calls wherever I see them since people keep inserting them to the code for no reason.

Let's dive into what yield actually does:

> The exact behavior of this function depends on the implementation, in particular on the mechanics of the OS scheduler in use and the state of the system. For example, a first-in-first-out realtime scheduler (SCHED_FIFO in Linux) would suspend the current thread and put it on the back of the queue of the same-priority threads that are ready to run, and if there are no other threads at the same priority, yield has no effect.
The POSIX equivalent of this function is sched_yield.

Now, let's check what [sched_yield does on Linux](https://man7.org/linux/man-pages/man2/sched_yield.2.html):

> DESCRIPTION
>        sched_yield() causes the calling thread to relinquish the CPU.
>        The thread is moved to the end of the queue for its static
>        priority and a new thread gets to run.

> CAVEATS
>        sched_yield() is intended for use with real-time scheduling
>        policies (i.e., SCHED_FIFO or SCHED_RR).  Use of sched_yield()
>        with nondeterministic scheduling policies such as SCHED_OTHER is
>        unspecified and very likely means your application design is
>        broken.
>        If the calling thread is the only thread in the highest priority
>        list at that time, it will continue to run after a call to
>        sched_yield().
>        Avoid calling sched_yield() unnecessarily or inappropriately
>        (e.g., when resources needed by other schedulable threads are
>        still held by the caller), since doing so will result in
>        unnecessary context switches, which will degrade system
>        performance.

So sched_yield is designed to facilitate cooperative multitasking in real time environment. We are not running this with real time scheduler so the call just bombards the kernel with unnecessary calls to scheduler that likely immediately return control to the calling thread. Moreover, with preemptive multitasking kernel can suspend a task whenever it wants, there is no need to call sched_yield for it since it is needed only to release the CPU to other tasks of same priority.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I'm not familiar with UCX design but Yossi suggested to poll on eventfds exposed by UCX as the best approach for implementing non-busy-polling progress thread. Why do you suggest mutex+condvar+queue would be better for this case?

src/utils/ucx/ucx_utils.cpp
w1ldptr: Notice that I'm not introducing the naming, just adding additional parameter. This one seems to be named like that because there is already nixlUcxContext::mt_type class field with that name. Anyway, I'm removing mt_type parameter from this constructor in the upcoming PR that implement RW locking, so don't think it is worth further consideration at the moment.

src/plugins/ucx/ucx_backend.cpp
yosefe: AFAIK the kernel will call schedule() at the end of the syscall (poll) anyway. so explicit sleep seems redundant.

src/plugins/ucx/ucx_backend.cpp
iyastreb: It's not that I propose, rather wondering if this task can be solved with that simple approach.
Blocking queue impl is much more simple and straightforward comparing to polling FDs, so I just want to check what are justifications/requirements for FDs.
@yosefe ?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I'm adding sleep not to release the control to scheduler (Linux doesn't require our cooperation to suspend this thread any time it wants to), I'm adding it to gracefully allow OOM killer to free up some memory during that time instead of basically busy-waiting by calling poll() repeatedly.

I don't feel strongly about this either way, so can just remove it if you think it is redundant.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Could you sketch a high-level implementation? How do you sleeping-wait for events on multiple UCX workers with mutex+condvar+queue?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,27 +361,36 @@ void nixlUcxEngine::progressFunc()
...
+            // Any other error would indicate bug in UCX
+            NIXL_ASSERT(!errno || errno == ENOMEM || errno == EINTR || errno == EAGAIN)
+                << "Call to poll() failed with unexpected error: " << strerror(errno);
+            if (errno == EINTR || errno == EAGAIN)
+                NIXL_TRACE << "Call to poll() was interrupted, retrying. Error: " << strerror(errno);
```
brminich: imo, this trace is not really needed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -411,6 +425,11 @@ void nixlUcxEngine::progressThreadStop()
     }
 
     pthrStop = true;
+    uint64_t counter_inc = 1;
+    int ret = write(pthrControlFd, &counter_inc, sizeof(counter_inc));
+    if (ret < 0)
+        NIXL_ERROR << "Write to eventfd should never fail unless the program failed to reset the counter. Error: "
```
brminich: maybe just
```suggestion
        NIXL_ERROR << "write to eventfd failed, error: "
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,27 +361,36 @@ void nixlUcxEngine::progressFunc()
...
+            bool made_progress = false;
+            ucs_status_t status = UCS_INPROGRESS;
+            const auto &uw = uws[wid];
+            do {
+                while (uw->progress())
+                    made_progress = true;
+
+                status = ucp_worker_arm(uw->getWorker());
+            } while (status != UCS_OK);
+
```
brminich:
```suggestion
            } while (status == UCS_ERR_BUSY);
            
            NIXL_ASSERT(status == UCS_OK);

```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,27 +361,36 @@ void nixlUcxEngine::progressFunc()
...
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        while (poll(pollFds.data(), pollFds.size(), -1) <= 0) {
+            // Any other error would indicate bug in UCX
+            NIXL_ASSERT(!errno || errno == ENOMEM || errno == EINTR || errno == EAGAIN)
+                << "Call to poll() failed with unexpected error: " << strerror(errno);
```
brminich: maybe 
```suggestion
                << "poll() failed with unexpected error: " << strerror(errno);
```

src/plugins/ucx/ucx_backend.cpp
brminich: using poll is the way UCX supports sleep-based waiting (wakeup mode/feature)

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -459,17 +488,25 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+            ucs_status_t ret = ucp_worker_get_efd(uw->getWorker(), &fd);
+            if (ret != UCS_OK) {
+                NIXL_ERROR << "Couldn't obtain fd for a worker, status: " << ucs_status_string(ret);
+                initErr = true;
+                return;
+            }
+
+            pollFds.push_back({fd, POLLIN, 0});
+        }
+        pollFds.push_back({pthrControlFd, POLLIN, 0});
```
brminich: can you use `ucp_worker_signal` instead of `pthrControlFd`?

w1ldptr: I thought about it and I don't think I can. If ucp_worker_signal() is called during the loop that progresses workers and before that specific worker has been armed, then the signal would be ignored. After that the progress thread would sleep on poll() indefinitely while the thread that called progressThreadStop() deadlocks waiting on join().

So basically we would be introducing very rare and hard to debug deadlock, unless I'm missing something.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Yossi specifically asked for it. We are going in circles by now...

src/plugins/ucx/ucx_backend.cpp
brminich: makes sense
probably you could check `pthrStop` right before going to poll, but this also does not look nice, so up to you

w1ldptr: That would work I think, but the code will become quite brittle when any small rearrangement can introduce hard to debug deadlock that is quite challenging to cover with tests for a very little gain IMO...

brminich: ok

brminich (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -396,6 +408,9 @@ void nixlUcxEngine::progressThreadStart()
         // not enabled
         return;
     }
+    uint64_t counter;
+    // This resets eventfd internal counter to 0
+    std::ignore = read(pthrControlFd, &counter, sizeof(counter));
```
artpol84: To be on the safe side, maybe better to keep reading until it's done

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -411,6 +426,11 @@ void nixlUcxEngine::progressThreadStop()
     }
 
     pthrStop = true;
+    uint64_t counter_inc = 1;
+    int ret = write(pthrControlFd, &counter_inc, sizeof(counter_inc));
```
artpol84: I have a concern that on relaxed memory ordering systems, like ARM, the below may not be sufficient:
 ```C
pthrStop = true
int ret = write(pthrControlFd, &counter_inc, sizeof(counter_inc));
```

 ```C
poll()
check(pthrStop)
```

I'll let others correct me, but this is my understanding that was nicely summarized by chatGPT (see below).
To be on the safe side, we can either use `membarrier(MEMBARRIER_CMD_PRIVATE_EXPEDITED)` (https://man7.org/linux/man-pages/man2/membarrier.2.html)

OR (which is much simpler IMO) check that `pthrControlFd` has triggered the poll and read out its value in the progress function and make the decision based on the read value rather than on the shared variable.

```
ChatGPT excerpt:

On ARM systems, the behavior of system calls like write() and poll()
with respect to memory barriers is not guaranteed to enforce full
memory ordering—you must not rely on them as memory barriers,
especially for synchronization between threads or cores.

What these syscalls do:
write(fd, buf, len) sends data to a file descriptor.

poll(fds, nfds, timeout) waits for events on file descriptors.

These system calls do not perform any explicit synchronization
on user memory. Their main effects are on kernel-visible state
(e.g., buffers, descriptors), not on shared memory between threads/cores.

System calls do enter the kernel, which typically includes a context
switch or privilege level change. This can cause:
* Pipeline flushing
* Instruction serialization
* Cache/TLB effects
But these are side effects, not guarantees. On ARM, without an explicit
barrier, reads and writes can still be reordered around a syscall from the
perspective of another thread or CPU.
```

w1ldptr: I'm not an expert in C++ atomics, but the [default memory ordering memory_order_seq_cst](https://en.cppreference.com/w/cpp/atomic/memory_order) establishes ordering and provides memory barrier:

> A load operation with this memory order performs an acquire operation, a store performs a release operation, and read-modify-write performs both an acquire operation and a release operation, plus a single total order exists in which all threads observe all modifications in the same order
> 
> All memory writes (including non-atomic and relaxed atomic) that happened-before the atomic store from the point of view of thread A, become visible side-effects in thread B. That is, once the atomic load is completed, thread B is guaranteed to see everything thread A wrote to memory.
> 
> On strongly-ordered systems — x86, SPARC TSO, IBM mainframe, etc. — release-acquire ordering is automatic for the majority of operations. No additional CPU instructions are issued for this synchronization mode; only certain compiler optimizations are affected (e.g., the compiler is prohibited from moving non-atomic stores past the atomic store-release or performing non-atomic loads earlier than the atomic load-acquire). On weakly-ordered systems (ARM, Itanium, PowerPC), special CPU load or memory fence instructions are used.
> 
> Total sequential ordering requires a full memory fence CPU instruction on all multi-core systems. This may become a performance bottleneck since it forces the affected memory accesses to propagate to every core.

Doesn't look like we need anything else here.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Reads on eventfd have [well defined semantics](https://man7.org/linux/man-pages/man2/eventfd.2.html):

> [read(2)](https://man7.org/linux/man-pages/man2/read.2.html)
>
>               Each successful [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) returns an 8-byte integer.  A
>               [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) fails with the error EINVAL if the size of the
>               supplied buffer is less than 8 bytes.
> 
>               The value returned by [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) is in host byte order—that
>               is, the native byte order for integers on the host machine.
> 
>               The semantics of [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) depend on whether the eventfd
>               counter currently has a nonzero value and whether the
>               EFD_SEMAPHORE flag was specified when creating the eventfd
>               file descriptor:
> 
>               •  If EFD_SEMAPHORE was not specified and the eventfd
>                  counter has a nonzero value, then a [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) returns 8
>                  bytes containing that value, and the counter's value is
>                  reset to zero.
> 
>               •  If EFD_SEMAPHORE was specified and the eventfd counter
>                  has a nonzero value, then a [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) returns 8 bytes
>                  containing the value 1, and the counter's value is
>                  decremented by 1.
> 
>               •  If the eventfd counter is zero at the time of the call
>                  to [read(2)](https://man7.org/linux/man-pages/man2/read.2.html), then the call either blocks until the
>                  counter becomes nonzero (at which time, the [read(2)](https://man7.org/linux/man-pages/man2/read.2.html)
>                  proceeds as described above) or fails with the error
>                  EAGAIN if the file descriptor has been made nonblocking.

Let's properly understand and rely on semantics of primitives that we use: 8byte read on eventfd descriptor resets the underlying counter to 0 or immediately returns EAGAIN if it is 0 already, no need for anything else.

src/plugins/ucx/ucx_backend.cpp
iyastreb: I meant just a typical blocking queue to implement producer-consumer pattern:
```
template <typename T>
class BlockingQueue {
    // called by producer
    void push(T item) {
        std::unique_lock<std::mutex> lock(m_mutex);
        m_queue.push(std::move(item));
        m_cond_var.notify_one();
    }

    // called in a loop by consumer(s)
    T pop() {
        std::unique_lock<std::mutex> lock(m_mutex);
        m_cond_var.wait(lock, [this] { return !m_queue.empty(); });
        T item = std::move(m_queue.front());
        m_queue.pop();
        return item;
    }

    std::queue<T> m_queue;
    std::mutex m_mutex;
    std::condition_variable m_cond_var;
};
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,27 +361,37 @@ void nixlUcxEngine::progressFunc()
...
-        // }
-
-        /* Wait for predefined number of */
-        us_t start = getUs();
-        while( (start + pthrDelay) > getUs()) {
-            std::this_thread::yield();
+
+        while (poll(pollFds.data(), pollFds.size(), -1) <= 0) {
+            // Any other error would indicate bug in UCX
+            NIXL_ASSERT(!errno || errno == ENOMEM || errno == EINTR || errno == EAGAIN)
```
yosefe: i'd prefer NIXL_ERROR log message and retry the polling, rather than crashing the application with assert. Morevover, NIXL_ASSERT could also be disabled in release. and then we don't have err indication

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -431,11 +451,20 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     nixl_b_params_t* custom_params = init_params->customParams;
 
     if (init_params->enableProgTh) {
+        pthrOn = true;
         if (!nixlUcxContext::mtLevelIsSupproted(NIXL_UCX_MT_WORKER)) {
             NIXL_ERROR << "UCX library does not support multi-threading";
             this->initErr = true;
             return;
         }
+        pthrControlFd = eventfd(0, EFD_NONBLOCK);
```
yosefe: is event fd posix compliant? since posix compliancy argument justified using poll rather than epoll

test/unit/plugins/mooncake/mooncake_backend_test.cpp
```diff
@@ -658,7 +658,10 @@ int main()
 #ifdef HAVE_CUDA
     int dev_ids[2] = { 0 , 0 };
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
```
yosefe, line 662: NIXL_DEBUG

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -688,7 +688,10 @@ int main()
 #ifdef HAVE_CUDA
     int dev_ids[2] = { 0 , 0 };
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
```
yosefe, line 692: NIXL_DEBUG

test/unit/plugins/ucx_mo/ucx_mo_backend_test.cpp
```diff
@@ -562,7 +562,10 @@ int main()
 int ndevices = NUM_WORKERS;
 #ifdef HAVE_CUDA
     int n_vram_dev;
-    cudaGetDeviceCount(&n_vram_dev);
+    if (cudaGetDeviceCount(&n_vram_dev) != cudaSuccess) {
+        std::cout << "Call to cudaGetDeviceCount failed, assuming 0 devices";
```
yosefe, line 566: NIXL_DEBUG

src/plugins/ucx/ucx_backend.cpp
w1ldptr: We are asserting because any other error means invariant violation in our library. We don't open new FDs dynamically, inherit them after forking from parent process or anything like that. All the descriptors are created on startup with verified success status, so getting EFAULT or EINVAL should never happen. Retrying will just lock the thread in infinite loop (it is not like the poll file descriptors array is going to fix itself on the next iteration) and all the user thread will get stuck waiting for progress that never happens.

test/unit/plugins/ucx/ucx_backend_test.cpp
w1ldptr: Both mooncake_backend_test and mooncake_backend_test use std streams throughout their implementations. Don't think adding single NIXL_DEBUG is a good practice.

test/unit/plugins/ucx_mo/ucx_mo_backend_test.cpp
w1ldptr: ditto

test/unit/plugins/mooncake/mooncake_backend_test.cpp
w1ldptr: ditto

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Good catch, I will switch it back to pipes.

BTW the argument for using poll() was that it is simpler to use while at the same time as fast or faster than epoll() on small amount of descriptors.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Nevermind, refactored control channel back to pipes.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Since I needed to refactor the control channel back to pipes anyway, I've also removed pthrStop atomic and implemented cancellation based purely on control channel as you suggested.

artpol84 (approved):

test/unit/plugins/ucx/ucx_backend_test.cpp
yosefe: ok, didnt notice it is test code

src/plugins/ucx/ucx_backend.cpp
yosefe: Well I guess you can use this argument to say that any error from library calls (also, cuda) should crash the application, not sure it is the right approach. maybe the progress thread should just exit or do busy polling?
also, what about the error not being checked in release mode?

w1ldptr: > Well I guess you can use this argument to say that any error from library calls (also, cuda) should crash the application, not sure it is the right approach. 

I already explained in my previous comment why this won't be an "any error" but an invariant violation, e.g. a bug in our design.

> also, what about the error not being checked in release mode?

Well, there is another assert version that doesn't get compiled-out in release mode which we could use.

Anyway, I'm not married to this assert so removing it now...

PR #301, opened 2025-05-13
fix: remove nixlGdsIOBatch::destroyBatch to prevent double cuFileBatchIODestroy call

### LINKED ISSUE #300: GDS plugin triggers "ERROR Batch Ctx version mismatch in cuFileBatchGetContextFromID"
Calling cuFileBatchIODestroy() twice on the same CUfileBatchHandle_t causes

```bash
python examples/python/nixl_gds_example.py /path/to/dir
```

cufile.log:
```
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  0:324 Batch Ctx version mismatch in cuFileBatchGetContextFromID
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  0:369 Did not find any valid context in cuFileBatchGetContextFromID
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  cufio_batch:2735 Batch state is not in IDLE state, try destroy later
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  0:324 Batch Ctx version mismatch in cuFileBatchGetContextFromID
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  0:369 Did not find any valid context in cuFileBatchGetContextFromID
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  cufio_batch:2735 Batch state is not in IDLE state, try destroy later
```

The issue appears when nixlGdsIOBatch::destroyBatch() is invoked in nixlGdsEngine’s destructor and later the nixlGdsIOBatch destructor calls cuFileBatchIODestroy() again.

https://github.com/ai-dynamo/nixl/blob/498fe2ac0b36f7223c8d9c66508de814928602d3/src/plugins/cuda_gds/gds_backend.cpp#L408-L416

https://github.com/ai-dynamo/nixl/blob/498fe2ac0b36f7223c8d9c66508de814928602d3/src/plugins/cuda_gds/gds_utils.cpp#L144-L147

https://github.com/ai-dynamo/nixl/blob/498fe2ac0b36f7223c8d9c66508de814928602d3/src/plugins/cuda_gds/gds_utils.cpp#L108-L114

The code for a minimal PoC is shown below.
```cpp
#include <cuda_runtime.h>
#include <cufile.h>

int main(int argc, char *argv[]) {
    CUfileBatchHandle_t batch_id;
    unsigned int batch_size = 10;
    cudaSetDevice(0);
    cuFileDriverOpen();
    cuFileBatchIOSetUp(&batch_id, batch_size);
    
    // double call
    cuFileBatchIODestroy(batch_id);
    cuFileBatchIODestroy(batch_id); // Batch Ctx version mismatch
    return 0;
}
```

suzusuzu:
Remove the nixlGdsIOBatch::destroyBatch method to prevent a double call to cuFileBatchIODestroy, which was causing the following runtime error(cufile.log):

```
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  0:324 Batch Ctx version mismatch in cuFileBatchGetContextFromID
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  0:369 Did not find any valid context in cuFileBatchGetContextFromID
13-05-2025 19:29:14:429 [pid=124535 tid=124535] ERROR  cufio_batch:2735 Batch state is not in IDLE state, try destroy later
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  0:324 Batch Ctx version mismatch in cuFileBatchGetContextFromID
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  0:369 Did not find any valid context in cuFileBatchGetContextFromID
13-05-2025 19:29:14:445 [pid=124535 tid=124535] ERROR  cufio_batch:2735 Batch state is not in IDLE state, try destroy later
```

There was a note suggesting we “Explicitly destroy the batch before deleting” but the nixlGdsIOBatch destructor already handles that cleanup. Removing the redundant destroyBatch call ensures we only invoke cuFileBatchIODestroy once.

This change closes #300.


src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -408,13 +408,6 @@ nixl_status_t nixlGdsEngine::releaseReqH(nixlBackendReqH* handle) const
 }
 
 nixlGdsEngine::~nixlGdsEngine() {
```
vvenkates27: @suzusuzu  Thanks, You are right about not requiring destoryBatch() as the destructor handles it but you will still need to loop through the batch_pool to delete the batches. The batch_pool is a std:list which will not automatically delete the pointers. If you can add that portion back this patch would be correct.

vvenkates27 (changes requested):
@suzusuzu  We can remove destroyBatch member of the gds_utils to explicitely remove batches, but we still need to delete batches from the pool in the GdsEngine destructor.
Please can you fix that?

src/plugins/cuda_gds/gds_backend.cpp
suzusuzu: @vvenkates27 Thank you for reviewing. I apologize for introducing a memory leak. I’ve restored the relevant part. Could you please take another look at this patch?

vvenkates27 (approved):

Issue #298, opened 2025-05-12
nixl-sys crate fails to find nixl libraries on arm64 builds


nv-anants:
Error trace while building nixl-sys on arm
```
#58 104.4   = note: some arguments are omitted. use `--verbose` to show all linker arguments
#58 104.4   = note: /opt/rh/gcc-toolset-14/root/usr/libexec/gcc/aarch64-redhat-linux/14/ld: cannot find -lnixl: No such file or directory
#58 104.4           /opt/rh/gcc-toolset-14/root/usr/libexec/gcc/aarch64-redhat-linux/14/ld: cannot find -lnixl_build: No such file or directory
#58 104.4           /opt/rh/gcc-toolset-14/root/usr/libexec/gcc/aarch64-redhat-linux/14/ld: cannot find -lserdes: No such file or directory
#58 104.4           collect2: error: ld returned 1 exit status
```

Updating ld_library_path doesn't resolve this issue.

Currently build.rs in rust bindings is set to look only in x86 path - https://github.com/ai-dynamo/nixl/blob/main/src/bindings/rust/build.rs#L23

PR #302, opened 2025-05-13
feat: add ARM (aarch64) build support for nixl-sys

nv-anants:
support for building NIXL Rust bindings on arm64 architectures. 

CARGO_CFG_TARGET_ARCH is set automatically during cargo builds
https://doc.rust-lang.org/cargo/reference/environment-variables.html

closes: #298 
closes: OPS-62

### DIFF
--- a/src/bindings/rust/build.rs
+++ b/src/bindings/rust/build.rs
@@ -20,7 +20,15 @@ fn main() {
     let nixl_root_path =
         env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
     let nixl_include_path = format!("{}/include", nixl_root_path);
-    let nixl_lib_path = format!("{}/lib/x86_64-linux-gnu", nixl_root_path);
+
+    // Determine architecture based on target
+    let arch = match env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_else(|_| "x86_64".to_string()).as_str() {
+        "x86_64" => "x86_64",
+        "aarch64" => "aarch64",
+        other => panic!("Unsupported architecture: {}", other),
+    };
+
+    let nixl_lib_path = format!("{}/lib/{}-linux-gnu", nixl_root_path, arch);
 
     // Tell cargo to look for shared libraries in the specified directories
     println!("cargo:rustc-link-search={}", nixl_lib_path);


aranadive (approved):
Seems okay to me

src/bindings/rust/build.rs
```diff
@@ -20,7 +20,15 @@ fn main() {
     let nixl_root_path =
         env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
     let nixl_include_path = format!("{}/include", nixl_root_path);
-    let nixl_lib_path = format!("{}/lib/x86_64-linux-gnu", nixl_root_path);
+
+    // Determine architecture based on target
+    let arch = match env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_else(|_| "x86_64".to_string()).as_str() {
+        "x86_64" => "x86_64",
+        "aarch64" => "aarch64",
```
ryanolson, line 27: do we also want arm64?

src/bindings/rust/build.rs
```diff
@@ -20,7 +20,15 @@ fn main() {
...
+        other => panic!("Unsupported architecture: {}", other),
+    };
+
+    let nixl_lib_path = format!("{}/lib/{}-linux-gnu", nixl_root_path, arch);
```
ryanolson, line 31: we might need to update nixl to install to this path.  it will for x86_64, but not sure if that is hardcoded or if that needs to be updated for other archs.

src/bindings/rust/build.rs
nv-anants: yes. aarch64 == 64bit arm process i think

src/bindings/rust/build.rs
aranadive: Only the prefix is - nixl_root_path. Since the arm/x86 is arch dep, it should be determined by installer.

nv-anants: i tried building ai-dynamo-runtime wheel on arm64 and the path works for me
```
#58 108.2 🖨  Copied external shared libraries to package dynamo.libs directory:
#58 108.2     /opt/nvidia/nvda_nixl/lib/aarch64-linux-gnu/libstream.so
#58 108.2     /opt/nvidia/nvda_nixl/lib/aarch64-linux-gnu/libserdes.so
#58 108.2     /opt/nvidia/nvda_nixl/lib/aarch64-linux-gnu/libnixl_build.so
#58 108.2     /opt/nvidia/nvda_nixl/lib/aarch64-linux-gnu/libnixl.so
#58 108.2     /opt/nvidia/nvda_nixl/lib/aarch64-linux-gnu/libnixl_common.so
```

nv-anants: pasting offline conversation
>Ryan Olson
perfect!
mark those as resolved. thanks for validating

PR #303, opened 2025-05-13
Agent: Improve MD fetch and invalidate flow in ETCD mode using watchers

yanivz-nv:
* Refactor commWorker to move all ETCD functionality into new class nixlEtcdClient.
* When watching for specific key we try to fetch, avoid trying to loadRemoteMD on deletion events.
* Implement timeout when setting a watcher after trying to fetch a non-existing key.
* Setup watchers on prefix of remote agent after fetching its metadata.
* invalidateLocalMD will remove all keys with agent's prefix, then all remote watchers will be notified and invalidateRemoteMD.
* Relevant tests depend on PR [315](https://github.com/ai-dynamo/nixl/pull/315)

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -306,14 +306,14 @@ public:
                 return;
             }
             const auto &event = response.events()[0];
-            if (event.event_type() == etcd::Event::EventType::PUT) {
-                NIXL_ERROR << "Unexpected PUT: " << event.kv().key()
-                           << " (rev " << event.kv().modified_index() << ") = " << event.kv().as_string();
-
-            } else if (event.event_type() == etcd::Event::EventType::DELETE_) {
-                NIXL_DEBUG << "DELETE: " << event.kv().key() << " (rev " << event.kv().modified_index() << ")";
+            if (event.event_type() == etcd::Event::EventType::DELETE_) {
+                NIXL_DEBUG << "Watcher DELETE: " << event.kv().key()
+                           << " (rev " << event.kv().modified_index() << ")";
                 std::lock_guard<std::mutex> lock(invalidated_agents_mutex);
                 invalidated_agents.push_back(agent_name);
+            } else {
+                NIXL_ERROR << "Watcher for " << event.kv().key() << " received unexpected event from etcd: "
+                           << event.event_type();
             }
         };
 
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -214,7 +214,8 @@ public:
 
             if (response.is_ok()) {
                 metadata = response.value().as_string();
-                NIXL_DEBUG << "Successfully fetched key: " << metadata_key << " (rev " << response.value().modified_index() << ")";
+                NIXL_DEBUG << "Successfully fetched key: " << metadata_key
+                           << " (rev " << response.value().modified_index() << ")";
                 return NIXL_SUCCESS;
             } else {
                 NIXL_ERROR << "Failed to fetch key: " << metadata_key << " from etcd: " << response.error_message();
@@ -294,7 +295,8 @@ public:
         // DELETE events are enqueued to be deleted in commWorker (can't be done inside the Watcher callback)
         auto process_response = [this, agent_name](etcd::Response response) -> void {
             if (!response.is_ok()) {
-                NIXL_ERROR << "Watcher failed to watch agent " << agent_name << " from etcd: " << response.error_message();
+                NIXL_ERROR << "Watcher failed to watch agent " << agent_name
+                           << " from etcd: " << response.error_message();
                 return;
             }
             NIXL_DEBUG << "Watcher received " << response.events().size() << " events from etcd";
@@ -484,8 +486,9 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
                     if (ret != NIXL_SUCCESS) {
                         NIXL_ERROR << "Failed to load remote metadata: " << ret;
                         break;
-                    }else if (remote_agent_from_md != remote_agent) {
-                        NIXL_ERROR << "Metadata mismatch for agent: " << remote_agent << " from md: " << remote_agent_from_md;
+                    } else if (remote_agent_from_md != remote_agent) {
+                        NIXL_ERROR << "Metadata mismatch for agent: " << remote_agent
+                                   << " from md: " << remote_agent_from_md;
                         break;
                     }
                     NIXL_DEBUG << "Successfully loaded metadata for agent: " << remote_agent;
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -277,9 +277,6 @@ public:
         nixl_status_t ret = fetchMetadataFromEtcd(remote_agent, metadata_label, remote_metadata);
         if (ret == NIXL_SUCCESS) {
             return NIXL_SUCCESS;
-        } else if (ret == NIXL_ERR_INVALID_PARAM) {
-            NIXL_ERROR << "Metadata was invalidated for agent: " << remote_agent;
-            return NIXL_ERR_INVALID_PARAM;
         }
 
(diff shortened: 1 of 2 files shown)


yanivz-nv:
For tests need to first merge PR #315.

src/core/nixl_listener.cpp
```diff
@@ -99,162 +100,255 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
+    std::string makeKey(const std::string& agent_name,
+                        const std::string& metadata_type) {
+        std::stringstream ss;
+        ss << namespace_prefix << "/" << agent_name << "/" << metadata_type;
+        return ss.str();
+    }
+
+public:
+    nixlEtcdClient(const std::string& my_agent_name) {
+        std::string etcd_endpoints = std::getenv("NIXL_ETCD_ENDPOINTS");
```
aranadive: This can segfault if NIXL_ETCD_ENDPOINTS isnt set. Suggestion is to check if its not null then assign.

src/core/nixl_listener.cpp
```diff
@@ -99,162 +100,255 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
+        // DELETE events are enqueued to be deleted in commWorker (can't be done inside the Watcher callback)
+        auto process_response = [this, agent_name](etcd::Response response) -> void {
+            if (!response.is_ok()) {
+                NIXL_ERROR << "Watcher failed to watch agent " << agent_name << " from etcd: " << response.error_message();
+                return;
+            }
+            NIXL_DEBUG << "Watcher received " << response.events().size() << " events from etcd";
+            if (response.events().size() != 1) {
+                NIXL_ERROR << "Watcher agent " << agent_name << " received unexpected number of events from etcd: "
+                        << response.events().size();
```
etoledano: `absl::StrFormat` will make it more readable, for `printf` style format.
All around...

src/core/nixl_listener.cpp
```diff
@@ -99,162 +100,255 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
-}
 
-// Create etcd client with specified endpoints or from environment variable
-static std::unique_ptr<etcd::Client> createEtcdClient(std::string etcd_endpoints) {
-    try {
-        // Sanity check
-        if (etcd_endpoints.size() == 0) {
-            throw std::runtime_error("No etcd endpoints provided");
+    // Setup a watcher for an agent's metadata invalidation if it doesn't already exist
+    void setupAgentWatcher(const std::string &agent_name) {
```
etoledano, line 313: Don't you prefer to return a status? (all around)

src/core/nixl_listener.cpp
```diff
@@ -99,162 +100,255 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
+        // Check if etcd client is available
+        if (!etcd) {
+            NIXL_ERROR << "ETCD client not available";
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
+
+        // Create key for agent's metadata
+        std::string metadata_key = makeKey(agent_name, metadata_type);
+        try {
+            // Fetch metadata from etcd
```
etoledano: Most of the comments I see in this PR doesn't seem to align with the guidelines, e.g.
[NL.1: Don’t say in comments what can be clearly stated in code
](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code)

src/core/nixl_listener.cpp
yanivz-nv: IMHO I don't think it makes it more readable since we still do `NIXL_ERROR <<`, and if we want printf style format why not use printf style logging library?

```C++
NIXL_ERROR << "Watcher agent " << agent_name << " received unexpected number of events from etcd: "
           << response.events().size();

NIXL_ERROR << absl::StrFormat("Watcher agent %s received unexpected number of events from etcd: %ul",
                              agent_name, response.events().size());
```

src/core/nixl_listener.cpp
yanivz-nv: We currently don't do anything in commWorker thread loop if these methods fail except printing errors, so I didn't think it was necessary. WDYT?

src/core/nixl_listener.cpp
yanivz-nv: Fixed

src/core/nixl_listener.cpp
yanivz-nv: Fixed.

yanivz-nv:
Added additional bug fix: applying timeout when watching for non-existing key during fetch.

src/core/nixl_listener.cpp
```diff
@@ -99,162 +101,257 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
-            NIXL_INFO << "Agent " << agent_name << " is marked as invalid, removing all keys";
-            removeMetadataFromEtcd(agent_name, namespace_prefix, client);
+    // Fetch metadata from etcd or wait for it to be available
+    nixl_status_t fetchOrWaitForMetadataFromEtcd(const std::string& remote_agent,
+                                                 const std::string& metadata_label,
+                                                 nixl_blob_t& remote_metadata) {
+        nixl_status_t ret = fetchMetadataFromEtcd(remote_agent, metadata_label, remote_metadata);
+        if (ret == NIXL_SUCCESS) {
+            return NIXL_SUCCESS;
+        } else if (ret == NIXL_ERR_INVALID_PARAM) {
```
aranadive: We dont return INVALID_PARAM from fetchMetadataFromEtcd. Should you modify fetchMetadataFromEtcd to return INVALID_PARAM for an invalidated key?

yanivz-nv: It's actually an impossible statement that I didn't remove after dealing changing the invalidation flow. Now the client can either fetch an existing key or watch a non-existing, which can timeout if the key was already invalidated. Removed it now.

src/core/nixl_listener.cpp
```diff
@@ -99,162 +101,254 @@ ssize_t recvCommMessage(int fd, std::string &msg){
...
+                return;
+            }
+            NIXL_DEBUG << "Watcher received " << response.events().size() << " events from etcd";
+            if (response.events().size() != 1) {
+                NIXL_ERROR << "Watcher agent " << agent_name << " received unexpected number of events from etcd: "
+                           << response.events().size();
+                return;
+            }
+            const auto &event = response.events()[0];
+            if (event.event_type() == etcd::Event::EventType::PUT) {
```
aranadive: Why do we need to track PUT for remote agent? Wouldnt this lead to wrong wakeups for this watcher?

src/core/nixl_listener.cpp
```diff
@@ -378,57 +472,25 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
...
+                        break;
                     }
+
+                    std::string remote_agent_from_md;
+                    ret = myAgent->loadRemoteMD(remote_metadata, remote_agent_from_md);
+                    if (ret != NIXL_SUCCESS) {
+                        NIXL_ERROR << "Failed to load remote metadata: " << ret;
+                        break;
+                    }else if (remote_agent_from_md != remote_agent) {
+                        NIXL_ERROR << "Metadata mismatch for agent: " << remote_agent << " from md: " << remote_agent_from_md;
```
aranadive: You can wrap code after remote_agent to next line

src/core/nixl_listener.cpp
yanivz-nv: It actually shouldn't happen because each agent puts its prefix key at creation. I kept it so we can see a print of unexpected behaviour. Also the etcd::Watcher class watches forever for any number of events until canceled/destroyed.

src/core/nixl_listener.cpp
yanivz-nv: Fixed, along with other places.

aranadive (approved):

PR #306, opened 2025-05-13
Benchmark: make it possible to build container with UCX built from source

ovidiusm:
Adds an option to the benchmark container build.sh script to make it possible to build from source UCX.

Rationale: it is common to need to test UCX and NIXL changes in tandem with both libraries built from source, for testing and CI. 

Example:

```bash
./benchmark/nixlbench/contrib/build.sh --ucx /path/to/ucx
```

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -45,8 +45,8 @@ RUN echo "INFO: Using UCX from base image (UCX=${UCX})."
 
 # --- Stage 2b: Represents building UCX from source ---
 FROM os_setup_stage AS ucx_custom_image
-RUN mkdir -p /tmp/ucx_build
-COPY --from=ucx . /tmp/ucx_build
+RUN mkdir -p /workspace/ucx
+COPY --from=ucx . /workspace/ucx
 
 RUN echo "INFO: Starting custom UCX build..." && \
     apt-get update -y && \
@@ -55,21 +55,16 @@ RUN echo "INFO: Starting custom UCX build..." && \
         libnuma-dev librdmacm-dev ibverbs-providers && \
     echo "INFO: Removing pre-existing UCX installations..." && \
     rm -rf /usr/local/ucx /opt/hpcx/ucx && \
-    cd /tmp/ucx_build && \
+    cd /workspace/ucx && \
     ./autogen.sh && \
     echo "INFO: Building UCX..." && \
-    ./configure --prefix=/usr/local/ucx \
-                --with-cuda=/usr/local/cuda \
-                --enable-mt \
-                --without-go \
-                --disable-logging \
-                --disable-debug \
-                --disable-assertions \
-                --disable-params-check && \
+    ./contrib/configure-release --prefix=/usr/local/ucx \
+                                --with-cuda=/usr/local/cuda \
+                                --enable-mt \
+                                --without-go && \
     make -j$(nproc) && \
     make install && \
     cd / && \
-    rm -rf /tmp/ucx_build && \
     echo "INFO: Finished building and installing UCX to /usr/local/ucx."
 
 # --- Stage 3: UCX Image Selection ---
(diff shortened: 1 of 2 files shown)


benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -34,6 +39,49 @@ RUN apt-get update -y && \
...
+                --without-go \
+                --disable-logging \
+                --disable-debug \
+                --disable-assertions \
+                --disable-params-check && \
+    make -j$(nproc) && \
+    make install && \
+    cd / && \
+    rm -rf /tmp/ucx_build && \
+    echo "INFO: Finished building and installing UCX to /usr/local/ucx."
```
brminich, line 68: shall we update LD_LIBRARY_PATH then?

benchmark/nixlbench/contrib/build.sh
```diff
@@ -63,7 +65,7 @@ get_options() {
         --nixl)
             if [ "$2" ]; then
                 NIXL_BUILD_CONTEXT_ARGS="--build-context nixl=$2"
-		NIXL_SRC=$2
+		        NIXL_SRC=$2
```
brminich: no need to change

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -34,6 +39,49 @@ RUN apt-get update -y && \
...
+    ./autogen.sh && \
+    echo "INFO: Building UCX..." && \
+    ./configure --prefix=/usr/local/ucx \
+                --with-cuda=/usr/local/cuda \
+                --enable-mt \
+                --without-go \
+                --disable-logging \
+                --disable-debug \
+                --disable-assertions \
+                --disable-params-check && \
```
brminich: @tvegas1 would it make sense to introduce an option to build either for ib or for efa?

tvegas1: it will build and work for both if rdma-core has efa support/header.

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -34,6 +39,49 @@ RUN apt-get update -y && \
...
     etcd-client
 
+# --- Stage 2a: Represents using UCX from the base image ---
+FROM os_setup_stage AS ucx_upstream_image
+RUN echo "INFO: Using UCX from base image (UCX=${UCX})."
+
+# --- Stage 2b: Represents building UCX from source ---
+FROM os_setup_stage AS ucx_custom_image
+RUN mkdir -p /tmp/ucx_build
+COPY --from=ucx . /tmp/ucx_build
```
tvegas1: any use of putting it in /workspace/ucx, and keep code as done for others?

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -34,6 +39,49 @@ RUN apt-get update -y && \
...
+    apt-get update -y && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        autoconf automake libtool pkg-config make g++ \
+        libnuma-dev librdmacm-dev ibverbs-providers && \
+    echo "INFO: Removing pre-existing UCX installations..." && \
+    rm -rf /usr/local/ucx /opt/hpcx/ucx && \
+    cd /tmp/ucx_build && \
+    ./autogen.sh && \
+    echo "INFO: Building UCX..." && \
+    ./configure --prefix=/usr/local/ucx \
```
tvegas1: maybe use configure-release?

benchmark/nixlbench/contrib/Dockerfile
ovidiusm: The base image we use already has UCX installed in that location, so it is already set up

benchmark/nixlbench/contrib/Dockerfile
ovidiusm: good idea

benchmark/nixlbench/contrib/build.sh
brminich: unresolved

ovidiusm: Let's fix the indentation? It was intentional

brminich: can you pls explain way need to change indentation here? There is no any other change

brminich (approved):

tvegas1 (approved):

PR #310, opened 2025-05-14
nixlbench: Add support for POSIX plugin

aranadive:
Use POSIX plugin with nixlbench. It follows same flow as GDS. 
Also, add another num-files options to create multiple files. Using just 0 file for now.

### DIFF
--- a/src/plugins/posix/aio_queue.cpp
+++ b/src/plugins/posix/aio_queue.cpp
@@ -105,13 +105,6 @@ nixl_status_t aioQueue::checkCompleted() {
             num_completed++;
             aiocb.aio_fildes = 0;  // Mark as completed
             aiocb.aio_nbytes = 0;
-
-            // Log progress periodically
-            if (10 <= num_entries) {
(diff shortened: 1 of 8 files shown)


src/plugins/posix/aio_queue.cpp
```diff
@@ -107,8 +107,10 @@ nixl_status_t aioQueue::checkCompleted() {
             aiocb.aio_nbytes = 0;
 
             // Log progress periodically
-            if (num_completed % (num_entries / 10) == 0) {
-                NIXL_INFO << "Queue progress: " << (num_completed * 100.0 / num_entries) << "% complete";
```
sb17v, line 111: This one causes Arithmetic Exception in case batch_size is less than 10

sb17v (approved):

src/plugins/posix/aio_queue.cpp
aranadive: Removed

PR #313, opened 2025-05-15
python api add node_num to support multi node

ayrnb:
If a decode node includes multiple instances , the devid of different instances is the same, transferring the kvcache will result in an error. So, I use rank instead of local_rank to represent its devid.


tstamler:
Hi @ayrnb , can you provide some more details of your usage? It's not clear to me what the purpose of these changes would be. You mentioned using multiple instances (I guess agents) on a single node. devID should be the GPU device id, but for example with vLLM we recommend not mixing GPU device ids. Basically just using one NIXL agent per GPU. 

However, even just this one agent per GPU suggestion is just for the short term. Ideally, we would want to have one NIXL agent for node that can handle every GPU. UCX-MO backend should support this now, and UCX should natively support this in 1.19. Let me know what you think.

ayrnb:
@tstamler  The disaggregation on  vllm+nixl. For example, we have TP=8 for Prefill and 2*node(TP=16 )for Decode , because  ` self.dst_xfer_side_handles[engine_id][i] = self.nixl_wrapper.prep_xfer_dlist(self._remote_agents[engine_id][self.rank * tp_multiplier + i], descs)` . the prep_xfer_dlist desc's devid is global rank, but the kv register desc's devid is local rank, they do not match.

mkhazraee:
Hi @ayrnb, I see your point, however merging GPUs from different nodes causes different behavior among different descriptors. For instance, that's one of the reasons we force a single memory type per descriptor list, that way all the elements in a transfer have some similarity, and that helps with providing SLAs and cost estimates.

We let the user of NIXL take care of how they want to manage across nodes, instead of from one node starting a transfer on another node, it's rather straightforward to do, while it can become misleading and hard to debug for scenarios that one  node is initiating transfers from the other and becomes a complicated scenario.

On the note about vLLM, we're collaborating and actually TP16 is almost ready. For instance this PR is a relevant optimization: https://github.com/vllm-project/vllm/pull/19080

src/api/cpp/nixl_descriptors.h
```diff
@@ -24,6 +24,9 @@
 #include <algorithm>
 #include "nixl_types.h"
 
+static constexpr uint64_t NIXL_DEFAULT_GLOBAL_DEVID = 0;
```
roiedanino, line 27:
```suggestion
static constexpr uint64_t NIXL_DEFAULT_GLOBAL_DEV_ID = 0;
```

src/bindings/python/nixl_bindings.cpp
```diff
@@ -228,7 +228,13 @@ PYBIND11_MODULE(_bindings, m) {
...
                 nixl_reg_dlist_t new_list(mem, sorted, descs.size());
                 for(long unsigned int i = 0; i<descs.size(); i++)
-                    new_list[i] = nixlBlobDesc(descs[i][0].cast<uintptr_t>(), descs[i][1].cast<size_t>(), descs[i][2].cast<uint64_t>(), descs[i][3].cast<std::string>());
+                    new_list[i] = nixlBlobDesc(
+                        descs[i][0].cast<uintptr_t>(),
+                        descs[i][1].cast<size_t>(),
+                        descs[i][2].cast<uint64_t>(),
+                        descs[i][3].cast<std::string>(),
+                        descs[i][4].cast<uint64_t>(),
+                        descs[i][5].cast<uint64_t>());
```
roiedanino, line 237: Can you please add a short description for each field here? (same for the rest of the methods here)

PR #314, opened 2025-05-15
Implement new synchronization read-write locking mode

w1ldptr:
First, preparatory commits implementing new test and extending POSIX file backend with locking.

Then, the new sync mode implementation that changes existing NIXL agent mutex from std to Abseil implementation which supports both regular exclusive locking and read-write modes. Add new 'RW' sync mode and take the lock in the read mode in a subset of NIXL data-path APIs (reateXferReq, postXferReq, getXferStatus, releaseXferReq, genNotif).

On top of that perform small UCX Context class refactoring to initialize UCX context and workers with the most appropriate setting according to the configuration requested by the user (sync mode, progress thread on/off, number of workers).

### DIFF
--- a/src/core/sync.h
+++ b/src/core/sync.h
@@ -19,6 +19,7 @@
 #include "common/util.h"
 #include "nixl_params.h"
 #include "absl/synchronization/mutex.h"
+#include <shared_mutex>
 
 class nixlLock {
     public:
@@ -26,14 +27,30 @@ class nixlLock {
         {}
 
         void lock() {
-            if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
+            if (syncMode != nixl_thread_sync_t::NIXL_THREAD_SYNC_NONE) {
                 m.Lock();
             }
         }
 
+        void lock_shared() {
+            if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
+                m.Lock();
+            } else if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_RW) {
+                m.ReaderLock();
+            }
+        }
+
         void unlock() {
+            if (syncMode != nixl_thread_sync_t::NIXL_THREAD_SYNC_NONE) {
+                m.Unlock();
+            }
+        }
+
+        void unlock_shared() {
             if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
                 m.Unlock();
+            } else if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_RW) {
+                m.ReaderUnlock();
             }
         }
 
@@ -43,5 +60,6 @@ class nixlLock {
 };
 
 #define NIXL_LOCK_GUARD(lock) const std::lock_guard<nixlLock> UNIQUE_NAME(lock_guard) (lock)
+#define NIXL_SHARED_LOCK_GUARD(lock) const std::shared_lock<nixlLock> UNIQUE_NAME(lock_guard) (lock)
 
 #endif /* SYNC_H */
--- a/src/core/sync.h
+++ b/src/core/sync.h
@@ -18,7 +18,7 @@
 #define SYNC_H
 #include "common/util.h"
 #include "nixl_params.h"
-#include <mutex>
+#include "absl/synchronization/mutex.h"
 
 class nixlLock {
     public:
@@ -27,19 +27,19 @@ class nixlLock {
 
         void lock() {
             if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
-                m.lock();
+                m.Lock();
             }
         }
 
         void unlock() {
             if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
-                m.unlock();
+                m.Unlock();
             }
         }
 
     private:
         nixl_thread_sync_t syncMode;
-        std::mutex m;
+        absl::Mutex m;
 };
 
 #define NIXL_LOCK_GUARD(lock) const std::lock_guard<nixlLock> UNIQUE_NAME(lock_guard) (lock)
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -475,8 +475,8 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         numWorkers = 1;
 
     uc = std::make_shared<nixlUcxContext>(devs, sizeof(nixlUcxIntReq),
-                                          _internalRequestInit, _internalRequestFini, NIXL_UCX_MT_WORKER,
-                                          pthrOn);
+                                          _internalRequestInit, _internalRequestFini,
+                                          pthrOn, numWorkers, init_params->syncMode);
     for (unsigned int i = 0; i < numWorkers; i++)
         uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));
 
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -18,8 +18,11 @@
 #include "ucx_backend.h"
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
+#include "common/nixl_log.h"
 
 #include <optional>
+#include <string.h>
+#include <unistd.h>
 #include "absl/strings/numbers.h"
 
 #ifdef HAVE_CUDA
@@ -28,8 +31,6 @@
 #include <cufile.h>
 
 #endif
-
-static constexpr int const noSyncIters = 32;
 
 /****************************************
  * CUDA related code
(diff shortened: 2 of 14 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -318,32 +318,24 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
     ucp_config_t *ucp_config;
     ucs_status_t status = UCS_OK;
 
-    mt_type = __mt_type;
+    mt_type = (sync_mode >= nixl_thread_sync_t::NIXL_THREAD_SYNC_RW || prog_thread) ?
+        NIXL_UCX_MT_WORKER : NIXL_UCX_MT_SINGLE;
 
-    ucp_params.field_mask = UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_MT_WORKERS_SHARED;
+    ucp_params.field_mask = UCP_PARAM_FIELD_FEATURES | UCP_PARAM_FIELD_MT_WORKERS_SHARED |
+        UCP_PARAM_FIELD_ESTIMATED_NUM_EPS;
```
brminich: pls, remove. est num eps is how many eps we expect per worker

src/plugins/posix/aio_queue.h
```diff
@@ -28,6 +29,7 @@ class nixlPosixBackendReqH;
 
 class aioQueue : public nixlPosixQueue {
     private:
+        std::mutex lock;                   // Protects all internal state from concurrent modification
```
brminich: is nixlLock supposed to be used for plugins also?

w1ldptr: No. We need this mutex to protect queue internals when engine is executed concurrently on multiple threads because NIXL lock is taken in read mode by upper layer.

src/utils/ucx/ucx_utils.cpp
w1ldptr: ack

src/plugins/posix/aio_queue.h
w1ldptr: After consulting with posix plugin developers it appears that the locking in those queues is redundant since the queues are per-transfer request and aren't shared. Removed that commit from the series.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,34 +360,50 @@ void nixlUcxEngine::progressFunc()
...
+    for (size_t wid = 0; wid < pollFds.size() - 1; wid++)
+        pollFds[wid].revents = POLLIN;
+
+    bool pthrStop = false;
     while (!pthrStop) {
-        int i;
-        for(i = 0; i < noSyncIters; i++) {
-            for (auto &uw: uws)
-                uw->progress();
+        for (size_t wid = 0; wid < pollFds.size() - 1; wid++) {
```
brminich, line 369: is it from #297?

w1ldptr: It seems that every time I force-push the branch that this PR #297 is based on I also need to rebase and push this one, otherwise Github treats outdated base branch commit as belonging to this PR. Fixed it now, thanks.

test/nixl/nixl_test.cpp
```diff
@@ -204,7 +204,7 @@ static void initiatorThread(nixlAgent &agent, nixl_opt_args_t *extra_params,
 }
 
 static void runTarget(const std::string &ip, int port) {
-    nixlAgentConfig cfg(true, true, port, nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT);
```
artpol84, line 207: Should we test both?

test/gtest/multi_threading.cpp
```diff
@@ -156,8 +158,35 @@ TEST_F(MultiThreadingTestFixture, ConcurrentTransfers) {
...
+    auto transfer_sequence = [&]() {
+        nixl_status_t status = local_agent.genNotif(remote_agent_name, "hello!");
+        EXPECT_EQ(status, NIXL_SUCCESS);
+    };
+
+    std::thread t1(transfer_sequence);
+    std::thread t2(transfer_sequence);
+
+    t1.join();
+    t2.join();
```
artpol84, line 185: Two things:
1. I think that to be meaningful, this test must check the consistency of the received notifications
2. Maybe we want to be aggressive with threads and spawn more than 2
3. Should we try to make the race more possible by (a) doing multiple iterations; (b) synchronizing the beginning of notification generation

src/utils/ucx/ucx_utils.cpp
```diff
@@ -318,32 +318,22 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+                               bool prog_thread,
+                               unsigned long num_workers,
+                               nixl_thread_sync_t sync_mode)
 {
     ucp_params_t ucp_params;
     ucp_config_t *ucp_config;
     ucs_status_t status = UCS_OK;
 
-    mt_type = __mt_type;
+    mt_type = (sync_mode >= nixl_thread_sync_t::NIXL_THREAD_SYNC_RW || prog_thread) ?
```
artpol84, line 333: This doesn't seem a right comparison given the threading modes declaration:
```C
enum class nixl_thread_sync_t {
    NIXL_THREAD_SYNC_NONE,
    NIXL_THREAD_SYNC_STRICT,
    NIXL_THREAD_SYNC_RW,
    NIXL_THREAD_SYNC_DEFAULT = NIXL_THREAD_SYNC_NONE,
};
```

Should it be 
```
sync_mode > nixl_thread_sync_t::NIXL_THREAD_SYNC_NONE
```
?

artpol84: I'd rather vote for some kind of more reliable statement, like maybe having a `NIXL_THREAD_NEED_LOCKS()` macro defined next to `nixl_thread_sync_t` and just used here to avoid similar mistakes in future (if this is indeed a mistake, of course)

artpol84 (changes requested):

test/gtest/multi_threading.cpp
w1ldptr: I think this was discussed already when introducing the thread sanitizer tests initially - they don't require neither more threads nor more iterations. If there is unsafe access to memory, then tsan will catch it on first iterations with two threads.

test/nixl/nixl_test.cpp
w1ldptr: I don't see why - both models take exactly the same lock by exactly the same macro. And if tsan doesn't produce any errors with more permissive mode of that lock than it won't also with the more strict.

src/utils/ucx/ucx_utils.cpp
w1ldptr: What is the mistake here? In the following code the result of this conditional is mapped to one of the following:
```
    UCS_THREAD_MODE_SERIALIZED, /**< Multiple threads can access, but only one at a time */
    UCS_THREAD_MODE_MULTI,      /**< Multiple threads can access concurrently */
```

Unless I'm missing something, simultaneous access from multiple threads to a worker instance is only possible either when the locking model is permissive (e.g. multiple concurrent user data path threads) or when the progress thread is enabled (which will run concurrently with user data path threads irrespective of the synchronization model chosen).

test/gtest/multi_threading.cpp
w1ldptr: Regarding consistency - this test uses mock backend which doesn't actually pass any notifications between agents so there is nothing to verify. GenNotif() is used by nixl_test from multi-threaded scenario though (which I extended with multi-threading to be the smoke test).

src/utils/ucx/ucx_utils.cpp
brminich: do I understand correctly that with `NIXL_THREAD_SYNC_STRICT` thread safety will be ensured on NIXL level using NIXL_LOCK_GUARD? And we do not need to have mt support in UCX then?
in general I agree with @artpol84 that current initialization of `mt_type` is confusing, can you pls add some comment or initialize it somehow more explicitly?

w1ldptr: > do I understand correctly that with NIXL_THREAD_SYNC_STRICT thread safety will be ensured on NIXL level using NIXL_LOCK_GUARD? And we do not need to have mt support in UCX then?

Yes.

> in general I agree with @artpol84 that current initialization of mt_type is confusing, can you pls add some comment or initialize it somehow more explicitly?

I will add a comment.

w1ldptr: done

brminich (approved):

src/core/sync.h
```diff
@@ -18,30 +18,48 @@
...
+            }
+        }
+
+        void lock_shared() {
             if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_STRICT) {
-                m.lock();
+                m.Lock();
+            } else if (syncMode == nixl_thread_sync_t::NIXL_THREAD_SYNC_RW) {
+                m.ReaderLock();
             }
```
etoledano, line 40: Two notes about this class:


1. As I recall you preferred to not make the config statically determined, and since the options seems to expand, WDYT about making it use callbacks? its probably more readable vs. list of conditions.

```
class nixlLock{
public:
    nixlLock(const nixl_thread_sync_t sync_mode)
    {
        if (...) {
            lockFunc = [this](){ mutex_.lock(); };
            unlockFunc = [this](){ mutex_.unlock(); };
        } else {
            lockFunc =  {};
            unlockFunc =  {};
        }
    }

    void lock() {
        lockFunc();
    }

    void unlock() {
        unlockFunc();
    }

private:
    std::mutex mutex_;
    std::function<void()> lockFunc;
    std::function<void()> unlockFunc;
};
```

2. This class (`nixlLock`) should not be exposed at all since you wrap it nicely and its not straight forward lock seems like. There are several common ways to do that, I think the easiest is to make it in `internal` namespace and keep it in the header, but feel free to pick any other solution (e.g. using a `*.cpp` file).

src/utils/ucx/ucx_utils.cpp
```diff
@@ -318,32 +318,26 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
     ucp_params_t ucp_params;
     ucp_config_t *ucp_config;
     ucs_status_t status = UCS_OK;
 
-    mt_type = __mt_type;
+    // With strict synchronization model nixlAgent serializes access to backends, with more
+    // permissive models backends need to account for concurrent access and ensure their internal
+    // state is properly protected. Progress thread creates internal concurrency in UCX backend
+    // irrespective of nixlAgent synchronization model.
+    mt_type = (sync_mode >= nixl_thread_sync_t::NIXL_THREAD_SYNC_RW || prog_thread) ?
```
etoledano, line 333: Perhaps I'm overlooking something, but I can't identify a metric under which the sequence none, strict, rw, default is monotonic.

w1ldptr: ack

src/core/sync.h
w1ldptr: Implemented suggestion 1. Regarding 2 - std::lock_guard and std::shared_lock both call nixlLock which needs to be exposed in order for such code to compile.

PR #315, opened 2025-05-15
Agent: Bug fix: apply timeout to socket connection and add gtests for md exchange

yanivz-nv:
* Before, if the remote does not exist or is never ready, the commWorker
   blocks forever.
* This commit allows to try connection for 1 second and fail after
   timeout.
* Also avoid throwing exception on failures during handling of remote
   messages.
* New gtest fixture for metadata exchange tests using get/load and
   send/fetch with sockets. ETCD mode in a separate PR.

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -522,7 +522,10 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
                     std::string remote_agent;
                     ret = myAgent->loadRemoteMD(remote_md, remote_agent);
                     if(ret != NIXL_SUCCESS) {
-                        throw std::runtime_error("loadRemoteMD in listener thread failed, critically failing\n");
+                        NIXL_ERROR << "loadRemoteMD in listener thread failed for md from peer "
+                                   << socket_iter->first.first << ":" << socket_iter->first.second
+                                   << " with error " << ret;
+                        continue;
                     }
                     // not sure what to do with remote_agent
                 } else if(header == "SEND") {
@@ -535,7 +538,8 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
                     myAgent->invalidateRemoteMD(remote_agent);
                     break;
                 } else {
-                    throw std::runtime_error("Received socket message with bad header" + header + ", critically failing\n");
+                    NIXL_ERROR << "Received socket message with bad header" + header + " from peer "
+                               << socket_iter->first.first << ":" << socket_iter->first.second;
                 }
             }
 
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -39,32 +39,55 @@ int connectToIP(std::string ip_addr, int port) {
     listenerAddr.sin_port   = htons(port);
     listenerAddr.sin_family = AF_INET;
 
+    if (inet_pton(AF_INET, ip_addr.c_str(), &listenerAddr.sin_addr) <= 0) {
+        NIXL_ERROR << "inet_pton failed for ip_addr: " << ip_addr;
+        return -1;
+    }
+
+    // Create a non-blocking socket
     int ret_fd = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);
     if (ret_fd == -1) {
+        NIXL_ERROR << "socket creation failed for ip_addr: " << ip_addr << " and port: " << port;
         return -1;
     }
 
-    if (inet_pton(AF_INET, ip_addr.c_str(),
-                  &listenerAddr.sin_addr) <= 0) {
+    // Connect will return immediately with EINPROGRESS
+    int ret = connect(ret_fd, (struct sockaddr*)&listenerAddr, sizeof(listenerAddr));
+    if (ret < 0 && errno != EINPROGRESS) {
         close(ret_fd);
         return -1;
     }
 
-    //make connect block for now to avoid ambiguity in send right after
-    int orig_flags = fcntl(ret_fd, F_GETFL, 0);
-    int new_flags = orig_flags ^ O_NONBLOCK;
+    // Use select to wait for connection with timeout
+    fd_set write_fds;
+    FD_ZERO(&write_fds);
+    FD_SET(ret_fd, &write_fds);
 
-    fcntl(ret_fd, F_SETFL, new_flags);
+    struct timeval tv;
+    tv.tv_sec = 1;
+    tv.tv_usec = 0;
 
-    if (connect(ret_fd, (struct sockaddr*)&listenerAddr,
-                    sizeof(listenerAddr)) < 0) {
-        perror("async connect");
+    ret = select(ret_fd + 1, NULL, &write_fds, NULL, &tv);
+    if (ret <= 0) {
+        if (ret < 0) {
+            NIXL_ERROR << "select failed for ip_addr: " << ip_addr << " and port: " << port
+                       << " with error: " << strerror(errno);
+        } else {
+            NIXL_ERROR << "select timed out for ip_addr: " << ip_addr << " and port: " << port;
+        }
         close(ret_fd);
         return -1;
     }
 
-    //make nonblocking again
-    fcntl(ret_fd, F_SETFL, orig_flags);
+    // Check if connection was successful
+    int error = 0;
+    socklen_t len = sizeof(error);
+    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0 || error != 0) {
+        NIXL_ERROR << "getsockopt failed for ip_addr: " << ip_addr << " and port: " << port
+                   << " with error: " << strerror(error);
+        close(ret_fd);
+        return -1;
+    }
 
     return ret_fd;
 }
--- a/test/gtest/metadata_exchange.cpp
+++ b/test/gtest/metadata_exchange.cpp
@@ -1,0 +1,299 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <gtest/gtest.h>
+#include "nixl.h"
+#include "common.h"
+
+namespace gtest {
+namespace metadata_exchange {
+
+class MemBuffer {
+public:
+    MemBuffer(size_t size) :
+        vec_(size)
+    {
+    }
+
+    operator uintptr_t() const
+    {
+        return reinterpret_cast<uintptr_t>(vec_.data());
+    }
+
+    nixlBasicDesc getBasicDesc() const
+    {
+        return nixlBasicDesc(static_cast<uintptr_t>(*this), vec_.size(), dev_id_);
+    }
+
+    nixlBlobDesc getBlobDesc() const
+    {
+        return nixlBlobDesc(getBasicDesc(), "");
+    }
+
+    size_t getSize() const
+    {
+        return vec_.size();
+    }
+
+private:
+    std::vector<std::byte> vec_;
+    constexpr static uint64_t dev_id_ = 0;
+};
+
+class MetadataExchangeTestFixture : public testing::Test {
+
+    struct AgentContext {
+        std::unique_ptr<nixlAgent> agent;
+        const std::string name;
+        const int port;
+        nixlBackendH *backend_handle = nullptr;
+        std::vector<MemBuffer> buffers;
+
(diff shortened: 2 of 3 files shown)


test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+    status = dst.agent->fetchRemoteMD(src.name, &fetch_args);
+    ASSERT_EQ(status, NIXL_SUCCESS);
+
+    std::this_thread::sleep_for(sleep_time);
+
+    status = dst.agent->checkRemoteMD(src.name, {DRAM_SEG});
+    ASSERT_EQ(status, NIXL_SUCCESS);
+
+    // TODO: support invalidate local after remote fetch? (currently raises exception)
```
tstamler: I'm surprised this doesn't work, it should be a separate PR but good to look into

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,548 @@
...
+    // Case 1: Use unregistered descriptors
+    MemBuffer unregistered_buffer(1024, DRAM_SEG);
+    nixl_reg_dlist_t unregistered_descs(DRAM_SEG);
+    unregistered_descs.addDesc(unregistered_buffer.getBlobDesc());
+
+    status = src.agent->sendLocalPartialMD(unregistered_descs, &send_args);
+    ASSERT_NE(status, NIXL_SUCCESS);
+
+    // Case 2: Attempt to load connection info on agent without backend
+    // TODO: This currently raises exception in commWorker thread
```
tstamler: not sure what the expectation is here, if there are no backends, there are no metadata to exchange

src/core/nixl_listener.cpp
```diff
@@ -39,32 +39,56 @@ int connectToIP(std::string ip_addr, int port) {
...
+    fd_set write_fds;
+    FD_ZERO(&write_fds);
+    FD_SET(ret_fd, &write_fds);
+
+    // TODO: can we do less than 5 seconds?
+    struct timeval tv;
+    tv.tv_sec = 5;
+    tv.tv_usec = 0;
+
+    ret = select(ret_fd + 1, NULL, &write_fds, NULL, &tv);
```
tstamler, line 70: this is good, this is a feature I wanted to add. I feel like one second might even be fine, it should fail quicker than that if there is a network problem

test/gtest/metadata_exchange.cpp
yanivz-nv: Right, what I wonder is if whether we want to raise an exception on a remote agent just because the local agent sent an invalid/corrupted MD or didn't wait for the remote to create backends. Basically the local agent can easily kill the remote agent.

test/gtest/metadata_exchange.cpp
yanivz-nv: Yes because connections established from the remote are on a different port than the remote listenPort, so we need a way to know it or maybe use a mapping of name->(ip,port). Updated the comment with more info.

src/core/nixl_listener.cpp
yanivz-nv: Right, I just wondered how much do we tolerate the agents not being in sync, that is the remote commWorker might be busy with other work which is currently blocking. Changed to 1s.

test/gtest/metadata_exchange.cpp
tstamler: oh, I see now. Yes, I agree this should not kill the remote agent. If its a quick fix you can add it, or make it a future TODO.

yanivz-nv: I can easily change it to just log error instead of throw. A future TODO may be to implement an ACK mechanism to notify the sender on success/failure.

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+private:
+    static void *allocate(size_t size, nixl_mem_t mem_type)
+    {
+        switch (mem_type) {
+        case DRAM_SEG:
+            return malloc(size);
+        default:
+            return nullptr; // TODO
+        }
+    }
```
etoledano: Why do we allocate and not use smart pointers?

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+        AgentContext(std::unique_ptr<nixlAgent> agent, std::string name, int port) :
+            agent(std::move(agent)), name(std::move(name)), port(port)
+        {
+        }
+
+        void createAgentBackend()
+        {
+            nixl_status_t status = agent->createBackend("UCX", {}, backend_handle);
+            ASSERT_EQ(status, NIXL_SUCCESS);
+            ASSERT_NE(backend_handle, nullptr);
```
etoledano: 1. Why assert and not except?
2. `status` isn't needed seems like

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+        const int port;
+        nixlBackendH *backend_handle = nullptr;
+        std::vector<MemBuffer> buffers;
+
+        AgentContext(std::unique_ptr<nixlAgent> agent, std::string name, int port) :
+            agent(std::move(agent)), name(std::move(name)), port(port)
+        {
+        }
+
+        void createAgentBackend()
```
etoledano, line 91: PTAL at Menachem's agent test PR and see if can extract common functionality

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+            ASSERT_EQ(status, NIXL_SUCCESS);
+        }
+    };
+
+protected:
+
+    void SetUp() override
+    {
+        // Get random port between 10000 and 65535
+        int port_base = 10000 + (std::rand() % (65535 - 10000 + 1));
```
etoledano: Can reuse & refactor Menachem's `GetRandomByte`
But anyway shouldn't have magic numbers.

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+    // Invalid metadata
+    status = dst.agent->loadRemoteMD("invalid", remote_name);
+    ASSERT_NE(status, NIXL_SUCCESS);
+
+    // Remote does not exist so cannot invalidate
+    status = dst.agent->invalidateRemoteMD(src.name);
+    ASSERT_NE(status, NIXL_SUCCESS);
+}
+
+TEST_F(MetadataExchangeTestFixture, GetLocalPartialAndLoadRemote)
```
etoledano, line 209: PTAL at mencham's PR for reference for shorter tests

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+    status = dst.agent->invalidateRemoteMD(src.name);
+    ASSERT_EQ(status, NIXL_SUCCESS);
+
+    // Step 3: Get and load again but with extra params
+
+    nixl_opt_args_t extra_params;
+    extra_params.backends.push_back(src.backend_handle);
+    extra_params.includeConnInfo = true;
+
+    status = src.agent->getLocalPartialMD(valid_descs, md, &extra_params);
```
etoledano: All around, no need for status seems like.

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+    status = dst.agent->checkRemoteMD(src.name, {DRAM_SEG});
+    ASSERT_EQ(status, NIXL_ERR_NOT_FOUND);
+}
+
+TEST_F(MetadataExchangeTestFixture, SocketFetchRemoteAndInvalidateLocal)
+{
+    for (size_t i = 0; i < agents.size(); i++)
+        agents[i].createAgentBackend();
+
+    size_t count = (std::rand() % 10) + 2;
```
etoledano: Magic numbers.

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,570 @@
...
+    // send_args.ipAddr = dst.ip;
+    // send_args.port = dst.port;
+
+    // status = src.agent->invalidateLocalMD(&send_args);
+    // ASSERT_EQ(status, NIXL_SUCCESS);
+
+    // std::this_thread::sleep_for(sleep_time);
+
+    // status = dst.agent->checkRemoteMD(src.name, {DRAM_SEG});
+    // ASSERT_EQ(status, NIXL_ERR_NOT_FOUND);
```
etoledano: Please do not have TODO's in code, bugs/tasks tracking should be done else where.

test/gtest/metadata_exchange.cpp
yanivz-nv: This is an implementation of a shared_ptr, this class inherits from shared_ptr. It is similar to the class in `test/gtest/test_transfer.cpp`

test/gtest/metadata_exchange.cpp
yanivz-nv: 1. Because the test shouldn't continue if the expression is false.
2. Why not? Checking the return value.

test/gtest/metadata_exchange.cpp
yanivz-nv: WDYM? Checking the return value.

test/gtest/metadata_exchange.cpp
etoledano: 1. I think this is pretty much correct for every line in the test...
2. Its always an lvalue, why not have the function call in the gtest MACRO then?

test/gtest/metadata_exchange.cpp
etoledano: `EXPECT_EQ(src.agent->getLocalPartialMD(valid_descs, md, &extra_params), NIXL_SUCCESS)`

test/gtest/metadata_exchange.cpp
etoledano: I don't understand why, inheriting from std classes is almost never recommended, why not hold a pointer like all the other classes? how does this inheritance benefit? not sure how another implementation is related, maybe its not needed there too.

test/gtest/metadata_exchange.cpp
yanivz-nv: ASSERT_EQ causes the test to abort immediately if the assertion fails, while EXPECT_EQ allows the test to continue running other operations.
In this test, I'm performing a sequence of dependent operations where later steps rely on earlier ones succeeding. For example, I need to successfully create backends and register buffers before exchanging metadata. If that fails, it can cause subsequent errors.
I think the current ASSERT_EQ usage is appropriate because:
1. Most steps depend on previous ones succeeding
2. Continuing after a fundamental step fails would likely cause crashes or misleading errors, or in general an undefined behavior.

I could use EXPECT_EQ for the final assertions about invalidation since no subsequent operations depend on them, but the ASSERT_EQ approach clearly communicates that each step must succeed for the test to be valid.

test/gtest/metadata_exchange.cpp
yanivz-nv: IMO it reduces readability but OK...

test/gtest/metadata_exchange.cpp
yanivz-nv: Refactored it with a different approach, more specific for my test cases.

test/gtest/metadata_exchange.cpp
yanivz-nv: Removed comment and opened issue https://github.com/ai-dynamo/nixl/issues/339

yanivz-nv (commented):
Comments regarding Menachem's PR will be addressed once his PR is ready for merge.

test/gtest/metadata_exchange.cpp
etoledano: I understand the urge to separate between must-happen and should-happen operation, I'm not sure it would hold as a methodology so I suggested to simplify. I agree with you regarding the first point, regarding the second one, `status` can be dismissed giving a more fluent test.

test/gtest/metadata_exchange.cpp
etoledano: Not sure how having to write two lines each time instead of one is more readable nor convenient, I do understand it can become handy when debugging but that's basically writing production for debug, but anyway, I'm positive it'll help when trying to make the function fit the screen or keep it small.

test/gtest/metadata_exchange.cpp
yanivz-nv: Fixed

test/gtest/metadata_exchange.cpp
yanivz-nv: These tests are long by design because they test more advanced internal states other than simply getMD+loadMD

test/gtest/metadata_exchange.cpp
yanivz-nv: Fixed magic number. I'm not sure how Menachem's GetRandomByte answers my needs since it's not a byte. Changed to something similar but relevant for my case.

test/gtest/metadata_exchange.cpp
yanivz-nv: Done

test/gtest/metadata_exchange.cpp
yanivz-nv: As discussed, will not be done for this release.

aranadive (approved):

test/gtest/metadata_exchange.cpp
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    size_t getSize() const
+    {
+        return vec_.size();
+    }
+
+private:
+    std::vector<std::byte> vec_;
+    constexpr static uint64_t dev_id_ = 0;
```
roiedanino, line 70: Why underscores suffixes? @yanivz-nv  Can fix that?

PR #316, opened 2025-05-15
fix: memory leaks

rafiw:
fix some memory leaks and creating default data->remoteSections[remote_name]


rafiw:
I think this is better
```
if (data->backendEngines.count(backend) != 0)
    if (data->remoteSections[remote_name]->populate(
        descs, data->backendEngines[backend], dummy) == NIXL_SUCCESS)
    return NIXL_SUCCESS;
```

PR #318, opened 2025-05-15
fix: memory leak

rafiw:


src/core/nixl_agent.cpp
```diff
@@ -884,6 +885,10 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) {
 nixl_status_t
 nixlAgent::releasedDlistH (nixlDlistH* dlist_hndl) const {
     NIXL_LOCK_GUARD(data->lock);
+    for (auto it = dlist_hndl->descs.begin(); it != dlist_hndl->descs.end();) {
+        delete it->second;
+        it = dlist_hndl->descs.erase(it);
```
yosefe: can we do it in dlist_hndl destructor?

rafiw: I looked again and found that it was actually cleaned in it destructor, so removed

src/core/nixl_agent.cpp
```diff
@@ -1317,10 +1318,13 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
             return NIXL_SUCCESS;
         } else {
             nixl_meta_dlist_t dummy(descs.getType(), descs.isSorted());
-            for (const auto& [backend, conn_info] : data->remoteBackends[remote_name])
-                if (data->remoteSections[remote_name]->populate(
-                          descs, data->backendEngines[backend], dummy) == NIXL_SUCCESS)
-                    return NIXL_SUCCESS;
+            if (auto it = data->remoteBackends.find(remote_name); it != data->remoteBackends.end()) {
+                for (const auto& [backend, conn_info] : it->second)
```
tstamler, line 1322: what is the purpose of breaking out this if statement?

rafiw: sanity check that data->backendEngines[backend] exists to prevent undefined behavior from the populate function

yosefe: maybe use data->remoteSections.find(remote_name) ? and combine with the if in line 1321?

rafiw: `data->remoteSections[remote_name]` exists for sure this is the if in line 1316
The question is when iterating the remote backends in `data->remoteBackends` are we 100% sure tha this backend is in the engine list `data->backendEngines[backend]`, from the code in line 1141 I'm not sure it's true.

mkhazraee: Yes, line 1174 in loadRemoteMD guarantees that, only if this statement is true we add to data->remoteBackends:
if (data->backendEngines.count(nixl_backend)!=0)

mkhazraee: I suggest to add a comment instead, to clarify we thought this true. And that being said, I found a bug, in loadRemoteMD when there is an error, we delete the remoteSection but not the remoteBackend.

aranadive:
#361 fixes this

Issue #222, opened 2025-04-29
Memory registration fails with multiple GPUs in single process using UCX_MO backend


sb17v:
When attempting to register memory across multiple GPUs in a single process using the UCX_MO backend, the registration fails if all descriptors are added to a list and registered at once. The failure appears to stem from the current CUDA context being incorrectly assumed during registration.

```
nixl_reg_dlist_t desc_list(seg_type);

for(int i = 0; i < num_gpus; i++) {
  nixlBlobDesc desc;
  
  cudaSetDevice(i);
  cudaMalloc(&addr, buffer_size);
  
  desc.addr = addr;
  desc.len = buffer_size;
  desc.devId = i;
  
  desc_list.addDesc(desc)
}

agent->registerMem(desc_list, &opt_args)
```

However, calling registerMem() inside the loop for each device resolves the issue.

The problem occurs because registerMem() internally uses the current CUDA context to register memory. After the loop exits, the CUDA context is set to the last device, and all descriptors are registered in that context. This leads to errors when trying to register memory not associated with the current (last) device.

```
[1744077545.770044] [eos0014:3234413:0]    cuda_copy_md.c:687  UCX  ERROR cuMemGetAddressRange(0x7ffc60000000) error: named symbol not found
```

aranadive:
@sb17v thanks for the fix.

PR #324, opened 2025-05-17
UCX: fix multi GPU scenario (#222)

sb17v:
* Set CUDA device to avoid failure during registration
* Fix NIXLBench stat calculation for single process multi-GPU scenario

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -772,6 +772,8 @@ nixl_status_t nixlUcxEngine::registerMem (const nixlBlobDesc &mem,
         }
         if (need_restart) {
             progressThreadRestart();
+            // set the ctx for main thread
+            vramApplyCtx();
         }
     }
 
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -488,6 +488,11 @@ void xferBenchUtils::printStats(bool is_target, size_t block_size, size_t batch_
 
     total_data_transferred = ((block_size * batch_size) * num_iter); // In Bytes
     avg_latency = (total_duration / (num_iter * batch_size)); // In microsec
+    if (IS_PAIRWISE_AND_MG()) {
+        total_data_transferred *= xferBenchConfig::num_initiator_dev; // In Bytes
+        avg_latency /= xferBenchConfig::num_initiator_dev; // In microsec
+    }
+
     throughput = (((double) total_data_transferred / (1024 * 1024)) /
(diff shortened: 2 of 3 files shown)


aranadive (approved):

yosefe (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -720,6 +720,9 @@ nixl_status_t nixlUcxEngine::registerMem (const nixlBlobDesc &mem,
         if (need_restart) {
             progressThreadRestart();
         }
+#if HAVE_CUDA
+        cudaSetDevice(mem.devId);
```
tvegas1: it should probably be integrated with vramUpdateCtx() that is already handling cuda context setting?

brminich (commented):
is it possible to add a gtest for ucx BE fix?

src/plugins/ucx/ucx_backend.cpp
brminich: can you use (update if needed) `nixlUcxCudaCtx`?  Note currentlu cuda driver API is used in UCX backend, not cuda runtime.

brminich (changes requested):

src/plugins/ucx/ucx_backend.cpp
mkhazraee: IIRC we did find the cudacontext during first registration, and in later ones checked that it's the same one for UCX (as opposed to MO-UCX). One point was the device id might be different that cuda context, let's say in single GPU container mode the devID is always 0. And IIRC we did save it in some member of ucx backend class, so I think we should use that here.

On top of that, I think we got a feedback that during load of remote, the local context should be set to. The issue with the listener thread. Maybe we can add that one line in this PR two, just similarly setting the cudacontext #if HAVE_CUDA during loadRemoteMD.

brminich: regarding listener thread @tvegas1  is  about to submit a patch + test for it

sb17v: Thanks, I have added vramApplyCtx() to set the ctx with the correct devId for the main thread.

sb17v: @brminich, I have used the vramApplyCtx (), which uses the cudaSetCtx () to set the current context. We are not using CUDA RT API now.
The problem was that we were not setting the correct context (w/ correct devId) for the main thread. I have changed it, and tested that it is working.

brminich (approved):

brminich (approved):

tvegas1 (approved):

PR #327, opened 2025-05-18
pre-commit: Ignore crate in codespell

aranadive:


yosefe:
why don't we restore .codespellrc?

aranadive:
> why don't we restore .codespellrc?

Mainly to confine it to pre-commit. If we decide to add a lot of words maybe we can consider bringing .codespellrc back.

yosefe:
.codespellrc can be useful also for IDEs (and extensible for more words)

aranadive:
> .codespellrc can be useful also for IDEs (and extensible for more words)

I wasnt sure if having the .codespellrc also caused requirement of codespelll package so mostly removed it out of precaution. Maybe once if we have words we can bring it back.

aranadive:
@yosefe can you review?

mkhazraee (approved):

PR #329, opened 2025-05-19
Posix: Updated Uring Queue implementation

dfarge:


src/plugins/posix/aio_queue.cpp
```diff
@@ -116,7 +116,7 @@ nixl_status_t aioQueue::checkCompleted() {
     return (num_completed == num_entries) ? NIXL_SUCCESS : NIXL_IN_PROG;
 }
 
-nixl_status_t aioQueue::prepareIO(int fd, void* buf, size_t len, off_t offset) {
+nixl_status_t aioQueue::prepIO(int fd, void* buf, size_t len, off_t offset) {
```
etoledano, line 119: Whats wrong with `prepareIO`?

src/api/cpp/nixl_descriptors.h
```diff
@@ -269,6 +269,14 @@ class nixlDescList {
             { return descs.begin(); }
         inline typename std::vector<T>::iterator end()
             { return descs.end(); }
+        /**
+         * @brief Returns a view of a subrange of descriptors without copying data
```
etoledano: It doesn't seems you are returning a view, or am I missing something? `assign` strictly copies values

Plus, where is `slice` used?

src/plugins/posix/posix_backend.cpp
```diff
@@ -47,9 +46,9 @@ class BackendManager {
                     if (value == "true" || value == "1") {
                         if (!BackendManager::isAioAvailable()) {
                             NIXL_ERROR << "AIO backend requested but not available";
-                            return {NIXL_ERR_NOT_SUPPORTED, false};
+                            return nixlPosixQueue::queue_t::UNSUPPORTED;
```
etoledano: You can declare `using` in function scope, this is way too long to repeat.
All around...

src/plugins/posix/posix_backend.cpp
```diff
@@ -96,75 +95,70 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
...
+nixl_status_t nixlPosixBackendReqH::initQueues() {
     try {
-        if (use_aio) {
-            queue = QueueFactory::createAioQueue(queue_depth_, operation == NIXL_READ);
-            if (!queue) {
-                throw std::runtime_error("Failed to create AIO queue");
-            }
-            NIXL_INFO << "Using AIO backend";
-        } else {
 #ifdef HAVE_LIBURING
```
etoledano: Why do we have `ifdef` all around? can we not have? this is very very hard to read and C++ offers extensive handling for static checks.

src/plugins/posix/posix_backend.cpp
```diff
@@ -202,31 +187,22 @@ nixlPosixBackendReqH::~nixlPosixBackendReqH() {
...
 
 nixlPosixEngine::nixlPosixEngine(const nixlBackendInitParams* init_params)
-    : nixlBackendEngine(init_params) {
-    use_aio = true;
-    auto [init_status, should_use_aio] = BackendManager::shouldUseAio(init_params->customParams);
-    if (init_status != NIXL_SUCCESS) {
+    : nixlBackendEngine(init_params)
+    , queue_type_(BackendManager::getQueueType(init_params->customParams)) {
+    if (queue_type_ == nixlPosixQueue::queue_t::UNSUPPORTED) {
         initErr = true;
```
etoledano, line 240: Since you are using exception (as should), why do we need `initErr`?

src/plugins/posix/posix_backend.cpp
```diff
@@ -202,31 +187,22 @@ nixlPosixBackendReqH::~nixlPosixBackendReqH() {
...
         NIXL_ERROR << "Failed to initialize POSIX backend - requested backend not available";
         return;
     }
-    use_aio = should_use_aio;
-    NIXL_INFO << "POSIX backend initialized using " << (use_aio ? "AIO" : "io_uring") << " backend";
+    NIXL_INFO << absl::StrFormat("POSIX backend initialized using %s backend",
+                                 queue_type_ == nixlPosixQueue::queue_t::AIO ? "AIO" : "io_uring");
```
etoledano: If you need extra functionality beside enumation, you should extend `nixlPosixQueue::queue_t` to support this, e.g. stringify

src/plugins/posix/posix_backend.cpp
```diff
@@ -96,75 +95,70 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
...
-    }
-
-    const int desc_count = static_cast<int>(local.descCount());
-    for (int i = 0; i < desc_count; ++i) {
-        status = queue->prepareIO(
-            remote[i].devId,
-            reinterpret_cast<void*>(local[i].addr),
-            remote[i].len,
-            remote[i].addr
+    for (auto [local_it, remote_it] = std::make_pair(local.begin(), remote.begin());
```
etoledano, line 205: Is it necessary to copy each item while iterating in a fast path function? Why not use references?
All around

src/plugins/posix/posix_backend.cpp
```diff
@@ -96,75 +95,70 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
...
-
-            queue = QueueFactory::createUringQueue(queue_depth_, operation == NIXL_READ, &params);
-            if (!queue) {
-                throw std::runtime_error("Failed to create io_uring queue");
-            }
-            NIXL_INFO << "Using io_uring backend";
+        // Initialize io_uring parameters with basic configuration
+        // Start with basic parameters, no special flags
+        // We can add optimizations like SQPOLL later
+        struct io_uring_params params = {};
```
etoledano: [ES.21: Don’t introduce a variable (or constant) before you need to use it
](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es21-dont-introduce-a-variable-or-constant-before-you-need-to-use-it)

src/plugins/posix/posix_backend.cpp
```diff
@@ -96,75 +95,70 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
...
+#endif
+        switch (queue_type_) {
+            case nixlPosixQueue::queue_t::AIO:
+                queue = QueueFactory::createAioQueue(queue_depth_, operation == NIXL_READ);
+                if (!queue) {
+                    throw std::runtime_error("Failed to create AIO queue");
+                }
+                NIXL_INFO << "Using AIO backend";
```
etoledano: Do we really want that many prints in our log? Don't we have debug log?

src/plugins/posix/posix_backend.cpp
```diff
@@ -301,23 +295,34 @@ nixl_status_t nixlPosixEngine::postXfer(const nixl_xfer_op_t &operation,
...
+    return posix_handle->checkXfer();
 }
 
 nixl_status_t nixlPosixEngine::releaseReqH(nixlBackendReqH* handle) const {
-    delete static_cast<nixlPosixBackendReqH*>(handle);
+    auto posix_handle = dynamic_cast<nixlPosixBackendReqH*>(handle);
+    if (!posix_handle) {
+        NIXL_ERROR << "Invalid handle type";
+        return NIXL_ERR_INVALID_PARAM;
+    }
```
etoledano: This seems pretty repetitive no?
How about create a utility to fetch `nixlPosixBackendReqH` and throw if something isn't right?

src/infra/nixl_descriptors.cpp
```diff
@@ -417,6 +417,16 @@ void nixlDescList<T>::print() const {
...
 }
 
+template <class T>
+nixlDescList<T> nixlDescList<T>::slice(size_t start_idx, size_t end_idx) const {
+    if (start_idx >= descs.size() || end_idx > descs.size() || start_idx >= end_idx) {
+        return nixlDescList<T>(type, sorted, 0);  // Return empty list with same type and sorted property
+    }
+    nixlDescList<T> result(type, sorted, end_idx - start_idx);
+    result.descs.assign(descs.begin() + start_idx, descs.begin() + end_idx);
+    return result;
```
etoledano: 1. Instead of dealing with inclusive/exclusive, and count on the reader to read and understand the API description, consider replacing to start & size.
2. For performance, you create temporary `nixlDescList` which initiate N values with default c'tor (I assume), then you copy K values using `assign`, then you return a copy of the created object copying it one more time. I hope this function isn't planned to be used anywhere there is a performance consideration. BTW, there are ways to improve, e.g. by using smart pointers for the allocation to prevent the copy-by-value on return, and using a dedicated c'tor to copy the list, overall do just one copy.

src/plugins/posix/aio_queue.h
```diff
@@ -45,7 +44,7 @@ class aioQueue : public nixlPosixQueue {
         ~aioQueue();
         nixl_status_t submit() override;
         nixl_status_t checkCompleted() override;
-        nixl_status_t prepareIO(int fd, void* buf, size_t len, off_t offset) override;
+        nixl_status_t prepIO(int fd, void* buf, size_t len, off_t offset) override;
```
etoledano, line 47: No descriptions here? All around...

src/plugins/posix/posix_backend.cpp
```diff
@@ -38,7 +37,7 @@ class BackendManager {
             return QueueFactory::isUringAvailable();
         }
 
-        static std::tuple<nixl_status_t, bool> shouldUseAio(const nixl_b_params_t* custom_params) {
+        static nixlPosixQueue::queue_t getQueueType(const nixl_b_params_t* custom_params) {
```
etoledano: Why is this a static function and not independent one? I don't see how you u leverage belonging to `BackendManager`, also in the other functions.
For context logical relations C++ provides namespaces of course.

src/plugins/posix/posix_backend.h
```diff
@@ -62,84 +61,106 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
...
+                                                     local_count, remote_count)) {}
+    };
+
+    class QueueInitError : public std::runtime_error {
+        public:
+            QueueInitError(nixl_status_t status)
+                : std::runtime_error(absl::StrFormat("Failed to initialize queues: %s",
+                                                     nixlEnumStrings::statusStr(status))) {}
+    };
 };
```
etoledano: How does class helps you instead of `std::runtime_error` directly?

src/plugins/posix/posix_backend.cpp
```diff
@@ -202,31 +187,22 @@ nixlPosixBackendReqH::~nixlPosixBackendReqH() {
...
-    return NIXL_SUCCESS;
-}
-
-
-bool nixlPosixEngine::validatePrepXferParams(const nixl_xfer_op_t &operation,
-                                             const nixl_meta_dlist_t &local,
-                                             const nixl_meta_dlist_t &remote,
-                                             const std::string &remote_agent,
-                                             const std::string &local_agent) const {
+bool nixlPosixEngine::isValidPrepXferParams(const nixl_xfer_op_t &operation,
```
etoledano: Does this function uses internals? if not, why not in unnamed namespace in the source file?
Function that can be in unnamed namespace, should be there.
All around.

src/plugins/posix/posix_backend.h
```diff
@@ -62,84 +61,106 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
...
+
+    nixl_status_t checkXfer(nixlBackendReqH* handle) const override;
+    nixl_status_t releaseReqH(nixlBackendReqH* handle) const override;
 
-    nixl_status_t checkXfer(nixlBackendReqH* handle) const;
-    nixl_status_t releaseReqH(nixlBackendReqH* handle) const;
+    nixl_status_t loadLocalMD(nixlBackendMD* input, nixlBackendMD* &output) override {
+        output = input;
+        return NIXL_SUCCESS;
+    }
```
etoledano, line 130: This isn't something the user can do on its own? Why is this function needed?

src/plugins/posix/queue_factory_impl.h
```diff
@@ -52,13 +52,6 @@ class QueueFactory {
             }
             return std::make_unique<UringQueue>(num_entries, *params, is_read);
         }
-#else
-        static std::unique_ptr<nixlPosixQueue> createUringQueue(int num_entries, bool is_read, const void* params) {
```
etoledano, line 56: Where did it move?
BTW, you don't need to use `(void)` you can simply not name the argument.

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,35 @@
...
+#include "absl/strings/str_format.h"
 #include "common/nixl_log.h"
 
 namespace {
+    // io_urings internally rounds up size to exp of 2. We define the size to be exp of 2 so there aren't "hidden" unused uring sqes
     static constexpr unsigned int max_posix_ring_size_log = 10;
-    static constexpr unsigned int max_posix_ring_size = 1 << max_posix_ring_size_log;
+    // At most 2 unique uring_params: all but the last uring use max_posix_ring_size; the last handles the remainder.
+    // E.g., for max_posix_ring_size = 4 and 15 entries: uring sizes = [4, 4, 4, 3].
+    constexpr int max_num_uring_params = 2;static constexpr unsigned int max_posix_ring_size = 1 << max_posix_ring_size_log;
```
etoledano: This seems off line.

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,35 @@
...
+        throw std::runtime_error(absl::StrFormat("Failed to initialize io_uring instance: %s", strerror(errno)));
     }
 
     // Log the features supported by this io_uring instance
-    NIXL_INFO << "io_uring features:"
-              << " SQPOLL=" << ((mutable_params.features & IORING_FEAT_SQPOLL_NONFIXED) ? "yes" : "no")
-              << " IOPOLL=" << ((mutable_params.features & IORING_FEAT_FAST_POLL) ? "yes" : "no");
+    NIXL_INFO << absl::StrFormat("io_uring features: SQPOLL=%s IOPOLL=%s",
+                                 ((mutable_params.features & IORING_FEAT_SQPOLL_NONFIXED) ? "yes" : "no"),
+                                 ((mutable_params.features & IORING_FEAT_FAST_POLL) ? "yes" : "no"));
```
etoledano: This is hard to read, how about stringify the features mask?

src/plugins/posix/uring_queue.cpp
```diff
@@ -74,10 +68,10 @@ UringQueue::~UringQueue() {
 nixl_status_t UringQueue::submit() {
     int ret = io_uring_submit(&uring);
     if (ret != num_entries) {
-        NIXL_ERROR << "io_uring submit failed: " << strerror(-ret);
+        NIXL_ERROR << absl::StrFormat("io_uring submit failed: %s", strerror(-ret));
```
etoledano: Why do you need to send negative value of the error to `strerror`? isn't the purpose of the function to receive errno?

src/plugins/posix/uring_queue.cpp
```diff
@@ -106,14 +100,14 @@ nixl_status_t UringQueue::checkCompleted() {
 
     // Log progress periodically
     if (num_completed % (num_entries / 10) == 0) {
```
etoledano: What is this calculation? magic numbers? description? whats the period and of what?

src/plugins/posix/uring_queue.cpp
```diff
@@ -106,14 +100,14 @@ nixl_status_t UringQueue::checkCompleted() {
...
-        NIXL_DEBUG << "Queue progress: "
-                   <<  (num_completed * 100.0 / num_entries) << "% complete";
+        NIXL_DEBUG << absl::StrFormat("Queue progress: %.1f%% complete",
+                                      (num_completed * 100.0 / num_entries));
```
etoledano: `num_completed * 100.0 / num_entries`
Should be a utility, in unnamed namespace, with description.

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,35 @@
...
-              << " SQPOLL=" << ((mutable_params.features & IORING_FEAT_SQPOLL_NONFIXED) ? "yes" : "no")
-              << " IOPOLL=" << ((mutable_params.features & IORING_FEAT_FAST_POLL) ? "yes" : "no");
+    NIXL_INFO << absl::StrFormat("io_uring features: SQPOLL=%s IOPOLL=%s",
+                                 ((mutable_params.features & IORING_FEAT_SQPOLL_NONFIXED) ? "yes" : "no"),
+                                 ((mutable_params.features & IORING_FEAT_FAST_POLL) ? "yes" : "no"));
 
     return NIXL_SUCCESS;
 }
 
 UringQueue::UringQueue(int num_entries, const io_uring_params& params, bool is_read)
```
etoledano: `is_read` instead of `read/write`? not sure I understand why its more readable, or why even not get the function itself or an operation, but boolean flags should make sense, otherwise we'll get code of this sort:
```

class Logger {
public:
    Logger(bool verbose) {
        if (verbose) {
            std::cout << "Verbose logging enabled\n";
        } else {
            std::cout << "Standard logging\n";
        }
    }
};

int main() {
    Logger log1(true);   // What does 'true' mean here?
    Logger log2(false);  // What does 'false' mean?
}
```

src/plugins/posix/posix_backend.h
```diff
@@ -62,84 +61,106 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
...
+                                                     nixlEnumStrings::statusStr(status))) {}
+    };
 };
 
 class nixlPosixEngine : public nixlBackendEngine {
 private:
-    bool use_aio;                             // Use AIO instead of io_uring
-    const nixl_mem_list_t supported_mems = {  // supported memory types
+    const nixlPosixQueue::queue_t queue_type_;
+    const nixl_mem_list_t supported_mems = {
```
etoledano: If u implement the function using it in the header you can make it in an unnamed namespace, or in the function itself, both better than drag this memory along the class when its constant.

src/plugins/posix/posix_backend.cpp
dfarge: initErr is inherited from the nixlBackendEngine parent class, this field is checked by the function that creates engines and errs if its set to true.

etoledano: But it doesn't mean you have to use it as far as I understand.

src/plugins/posix/aio_queue.cpp
dfarge: I aligned it to our other "prepare" functions for example the prepXfer functions.

src/api/cpp/nixl_descriptors.h
dfarge: You are correct. This was for a different version that by the end of it wasn't add to this PR. Removed.

src/plugins/posix/posix_backend.cpp
dfarge: This is because currently the default queue type is aio. liburing has some security issues with docker so its optional. We need to be able to compile with or without liburing.

src/plugins/posix/posix_backend.cpp
dfarge: This is from the function that creates the backends, nixlAgent::createBackend:
```
nixlBackendEngine*    backend = nullptr;
...
if (plugin_handle) {
        // Plugin found, use it to create the backend
        backend = plugin_handle->createEngine(&init_params);
    } else {
        NIXL_ERROR << "Unsupported backend: " << type;
        return NIXL_ERR_NOT_FOUND;
    }

    if (backend) {
        if (backend->getInitErr()) {
            delete backend;
            return NIXL_ERR_BACKEND;
        }
...
```
This function is called by the user directly and this part isn't in a try block, so if we throw in the ctor no one will catch it. The current mechanism is the convention that the engine sets this value to true on error in ctor. (I prefer using exceptions here as well).

src/plugins/posix/posix_backend.cpp
dfarge: I might be wrong but to my understanding it doesn't copy the items, only the iterators.
Can you explain why it would copy the items?

src/plugins/posix/posix_backend.cpp
dfarge: It doesn't compile if it's declared in the switch statement.
image

src/plugins/posix/posix_backend.cpp
dfarge: These aren't my additions, @vvenkates27 ?

src/plugins/posix/posix_backend.cpp
dfarge: I considered it but didn't want to throw out of ctors.

src/infra/nixl_descriptors.cpp
dfarge: Removed this change

src/plugins/posix/aio_queue.h
dfarge: This file isn't a part of the PR really, I changed this function name because uring queue and aio queue both inherit this function.

src/plugins/posix/posix_backend.cpp
dfarge: @vvenkates27, was there a specific logic behind it?
Otherwise, I would move to an anonymous namespace as Eylon suggests

src/plugins/posix/posix_backend.h
dfarge: I explained that in a comment just above there:
`    // Different error classes for different error cases so we can return different error codes to user
`

We catch these at nixlPosixEngine::prepXfer:
```   
    } catch (const nixlPosixBackendReqH::BackendTypeError& e) {
        NIXL_ERROR << absl::StrFormat("Backend type error: %s", e.what());
    } catch (const nixlPosixBackendReqH::InvalidDescriptorCountError& e) {
        NIXL_ERROR << absl::StrFormat("Invalid descriptor count: %s", e.what());
    } catch (const nixlPosixBackendReqH::QueueInitError& e) {
        NIXL_ERROR << absl::StrFormat("Queue initialization error: %s", e.what());
        return NIXL_ERR_BACKEND;
    } catch (const std::exception& e) {
        NIXL_ERROR << absl::StrFormat("Unexpected error: %s", e.what());
        return NIXL_ERR_BACKEND;
    }

File: src/plugins/posix/posix_backend.cpp (hunk 15 above)
Comment (dfarge) [reply src/plugins/posix/posix_backend.cpp]: Moved.

File: src/plugins/posix/posix_backend.h (hunk 16 above)
Comment (dfarge) [reply src/plugins/posix/posix_backend.h:130]: I followed suit from the gds backend.
To my understanding because this is essentially a loop-back the target metadata is the source metadata.

File: src/plugins/posix/queue_factory_impl.h (hunk 17 above)
Comment (dfarge) [reply src/plugins/posix/queue_factory_impl.h:56]: It was removed.
This function is only used inside ifdef HAVE_LIBURING blocks.
If liburing isn't installed then HAVE_LIBURING isn't defined, and these functions don't show in the code that is compiled.

Comment (dfarge) [reply src/plugins/posix/queue_factory_impl.h:56]: Regarding the void I can change that, but this PR isn't about aio.

File: src/plugins/posix/uring_queue.cpp (hunk 18 above)
Comment (dfarge) [reply src/plugins/posix/uring_queue.cpp]: Correct, everything in this namespace is a leftover from a version not part of this PR. Removed.

File: src/plugins/posix/uring_queue.cpp (hunk 19 above)
Comment (dfarge) [reply src/plugins/posix/uring_queue.cpp]: Added stringifyUringFeatures in an anonymous namespace and using it there.

File: src/plugins/posix/uring_queue.cpp (hunk 20 above)
Comment (dfarge) [reply src/plugins/posix/uring_queue.cpp]: There are two types of errors that can occur here:
1. Partial submission - shouldn't happen so it's considered an error. (ret value is positive and smaller than num_entries)
2. Uring error - returns a negative value, so we change it to a positive by -ret.

Updated to distinguish between the two and log the error correctly.

File: src/plugins/posix/uring_queue.cpp (hunk 22 above)
Comment (dfarge) [reply src/plugins/posix/uring_queue.cpp]: Changed.

File: src/plugins/posix/posix_backend.h (hunk 24 above)
Comment (dfarge) [reply src/plugins/posix/posix_backend.h]: Moved it to the function.

File: src/plugins/posix/posix_backend.cpp (hunk 4 above)
Comment (etoledano) [reply src/plugins/posix/posix_backend.cpp]: Can this be gathered to a single ‘#if’ at one place and use c++ templates / other compilation time property to implement?

File: src/plugins/posix/posix_backend.cpp (hunk 5 above)
Comment (etoledano) [reply src/plugins/posix/posix_backend.cpp:240]: Sounds like a good opportunity to refactor, or add a task to do so

File: src/plugins/posix/posix_backend.cpp (hunk 10 above)
Comment (etoledano) [reply src/plugins/posix/posix_backend.cpp]: You can also catch, still saves common logic

File: src/plugins/posix/aio_queue.h (hunk 12 above)
Comment (etoledano) [reply src/plugins/posix/aio_queue.h:47]: aio isn’t part of posix? I thought the PR is to rewrite posix

File: src/api/cpp/backend/backend_aux.h [hunk 25]
```diff
@@ -57,7 +57,7 @@ class nixlBackendInitParams {
 class nixlBackendReqH {
 public:
     nixlBackendReqH() { }
-    ~nixlBackendReqH() { }
+    virtual ~nixlBackendReqH() { } // Made the destructor virtual to allow for proper cleanup
```
mkhazraee: I think this is the only relevant part for me. When we call releaseReqH(backendHandle) practically that is the destructor call, so we didn't make it Virtual. What different here that needs to be called through the base pointer? It's not a big deal, but good to understand what's making it different here. If we decide to keep it virtual, we can remove the comment at the end of the line as that's obvious, I understand it was just for the reviews.

src/meson.build
```diff
@@ -19,6 +19,6 @@ subdir('utils')
 
 #subdir('arch')
 subdir('infra')
-subdir('plugins')
 subdir('core')
+subdir('plugins')
```
aranadive: Why was this moved?

src/plugins/posix/posix_backend.cpp
etoledano: I will list it as a future refactor task, no further steps are needed.

src/plugins/posix/posix_backend.cpp
etoledano: Correct, my mistake.

src/plugins/posix/posix_backend.cpp
etoledano: I assume you were missing a scope for the `case`

src/plugins/posix/posix_backend.h
etoledano: I understand your point.

My main concerns with this design are:

- Each exception type carries two strings.
- The exception type is already chosen at a lower level, but it's translated again at a higher level.

Both of these can be simplified:

- The outer-level log can rely directly on `e.what()`.
- Your exception class can implement a `toStatus()` method that returns a status code of your choice.

```
#include <iostream>

enum nixl_status_t {
    kGood,
    kDescriptor,
    kParam,
};

class InvalidDescriptorCountError : public std::runtime_error {
    private:
        static constexpr nixl_status_t code_ = kDescriptor;
    public:
        InvalidDescriptorCountError()
            : std::runtime_error("Invalid descriptor count") {}
        
        nixl_status_t toStatus() const { return code_; }
};

void riskyFunction() {
    throw InvalidDescriptorCountError();
}

nixl_status_t safeOperation() {
    try {
        riskyFunction();
    } catch (const InvalidDescriptorCountError& e) {
        std::cout << e.what() << std::endl;
        return e.toStatus();
    }
    
    return nixl_status_t::kGood;
}


int main() {
    auto status = safeOperation();
    if (status != nixl_status_t::kGood) {
        std::cout << "Oops!" << std::endl;
    }
}
```

I tend to think this isn't just a matter of style, but rather about simplifying the code as much as possible. That said, I’d be happy to share my thoughts and hear your opinion.

I'm considering a few aspects:

- Code overhead: Creating a separate class for each exception type feels suboptimal if a shorter alternative exists.
- Scalability: Adding a new exception type should ideally be a one-liner, not require a new class.
- Use case: If exceptions are only caught at a higher level and translated into logs, then embedding a string message in the exception itself might not be necessary.

Let me share the optimistic approach.

Points 1 & 2: Use a one-liner (existing) `enum` to represent exception types.
Point 3: Log in the lower level

```
#include <iostream>

enum nixl_status_t {
    kFileNotFound,
    kInvalidInput,
    kNetworkError
};

void riskyFunction() {
    std::cerr << "File not found" << std::endl;
    throw nixl_status_t::kInvalidInput;
}

int main() {
    try {
        riskyFunction();
    } catch (nixl_status_t e) {
        return e;
    }
}
```

The main issue with this approach is that enum types are not derived from std::exception, which affects several aspects of exception handling.

That said, I don't think it's a critical problem in this case, especially since we're dealing with a small internal codebase where this pattern can be acceptable and pragmatic.

However, I believe the following alternative strikes a good balance: it preserves idiomatic C++ practices while addressing the concerns mentioned:

- It allows for type-safe exception handling.
- It supports meaningful error messages.
- It remains lightweight and easy to extend.

```
enum nixl_status_t {
    kGood,
    kFileNotFound,
    kInvalidInput,
    kNetworkError
};

class MyException : public std::exception {
private:
    std::string msg_;
    nixl_status_t code_;
public:
    explicit MyException(std::string msg, nixl_status_t code) : msg_(msg), code_(code) {}

    const char* what() const noexcept override {
        return msg_.c_str();
    }

    nixl_status_t code() const noexcept { return code_; }
};

void riskyFunction() {
    throw MyException("Bad parameter", nixl_status_t::kInvalidInput);
}

void riskyFunctionWithParams(int x) {
    std::string s = "Bad integer " + std::to_string(x);
    throw MyException(s, nixl_status_t::kInvalidInput);
}

nixl_status_t safeWrapper() {
    try {
        riskyFunctionWithParams(5);
    } catch (const MyException &e) {
        std::cout << e.what() << std::endl;
        return e.code();
    }
    
    return kGood;
}

int main() {
    std::cout << "safeWrapper status : " << safeWrapper() << std::endl;
}
```

As I understand it, this is a fairly general-purpose exception that could be useful across the codebase.
If you're okay with this solution, feel free to keep it local for now - I can create a task to promote it to a shared API and distribute it throughout the code later.

WDYT?

src/plugins/posix/queue_factory_impl.h
etoledano: Weird, but it looks to be used in this commit - did u check with and without the `HAVE_LIBURING` flag? That's one of the main reason why `ifdef` is bad...

```
                queue = QueueFactory::createUringQueue(queue_depth_, operation == NIXL_READ, &params);
```

src/plugins/posix/aio_queue.h
dfarge: I only changed the uring part, could possibly do an aio update in the future but I don't think I could make it to 3.0 which has code freeze in a couple days.

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,54 @@
...
-#include "common/status.h"
+#include "absl/strings/str_format.h"
+#include "absl/strings/str_join.h"
 #include "common/nixl_log.h"
 
 namespace {
-    static constexpr unsigned int max_posix_ring_size_log = 10;
-    static constexpr unsigned int max_posix_ring_size = 1 << max_posix_ring_size_log;
-}
+    constexpr unsigned int log_percent_step = 10;
```
etoledano: Should be in the function.

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,54 @@
...
 
-nixl_status_t UringQueue::init(int entries, const io_uring_params& params, bool read_op) {
-    memset(&uring, 0, sizeof(uring));
+    // Log completion percentage at regular intervals (every log_percent_step percent)
+    void logOnPercentStep(unsigned int completed, unsigned int total) {
```
etoledano, line 30: 1. `total = 0' isn't handled
2. No static assertion for `log_percent_step` in range of [1, 100]

Have you considered using `LOG_EVERY_N` (has several flavors) of abseil?

src/plugins/posix/uring_queue.cpp
```diff
@@ -21,38 +21,54 @@
...
+
+        if (features & IORING_FEAT_SQPOLL_NONFIXED) {
+            enabled_features.push_back("SQPOLL");
+        }
+        if (features & IORING_FEAT_FAST_POLL) {
+            enabled_features.push_back("IOPOLL");
+        }
+        if (enabled_features.empty()) {
+            return "none";
+        }
```
etoledano: Is this function called during fastpath? if so, you can consider the following improvement, iterating only over the set bits, if not please ignore and resolve this comment.

(can probably be done shorter with std/builtin functions, just to introduce the concept)

```

std::string stringifyUringFeatures(unsigned int features) {
    static const std::unordered_map<unsigned int, std::string> feature_map = {
        {IORING_FEAT_SQPOLL_NONFIXED, "SQPOLL"},
        {IORING_FEAT_FAST_POLL, "IOPOLL"}
        // Add more features here as needed
    };

    std::vector<std::string> enabled;

    unsigned int remaining = features;
    while (remaining) {
        unsigned int bit = remaining & -remaining;  // Isolate lowest set bit
        remaining &= (remaining - 1);               // Clear that bit

        auto it = feature_map.find(bit);
        if (it != feature_map.end()) {
            enabled.push_back(it->second);
        }
    }

    return enabled.empty() ? "none" : absl::StrJoin(enabled, ", ");
}

```

src/api/cpp/backend/backend_aux.h
dfarge: There's another point to making it virtual.
Before this change the class wasn't truly polymorphic because it didn't have any pure virtual functions (even though there's a comment saying it's a pure virtual class).
This change also allows us to use dynamic cast on instances of this class to derive the child classes.

dfarge: Also, Vishwanath recommended I ask you to take a look on another change in this PR which changes the order of compilation between the core and plugins.
I wanted to use the function that translates nixl_status_t to string on error logging, and it was in the core, so I changed the compilation order.
Are there any implications I'm unaware of for that?

src/meson.build
dfarge: I wanted to use the function that translates nixl_status_t to string on error logging, and it was in the core, so I changed the compilation order.

src/plugins/posix/aio_queue.h
etoledano: OK, let's take it offline to assign a rewrite of the same sort.

src/plugins/posix/posix_backend.cpp
```diff
@@ -96,75 +128,70 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
...
-    } catch (const std::exception& e) {
-        NIXL_ERROR << "Failed to initialize queues: " << e.what();
-        status = NIXL_ERR_BACKEND;
+    nixl_status_t status = initQueues();
+    if (status != NIXL_SUCCESS) {
+        throw QueueInitError(status);
     }
 }
 
-nixl_status_t nixlPosixBackendReqH::initQueues(bool use_aio) {
```
etoledano, line 120: Similar comment as with `is_read`, flags are impossible to understand when looking at the user code

src/plugins/posix/posix_backend.cpp
dfarge: That worked, thank you.
Changed.

src/plugins/posix/posix_backend.cpp
dfarge: Done.

src/plugins/posix/posix_backend.cpp
dfarge: With your suggestion in another comment its now a single #ifdef, is that ok or should I look into templates or other things?
I'm unfamiliar with how to use other options in this context

src/plugins/posix/posix_backend.cpp
dfarge: I see that we can't introduce methods to an enum class, so I added a static queueTypeToString method to nixlPosixQueue for that.

src/plugins/posix/posix_backend.h
dfarge: I like the 3rd option the best as well. I'll make it local for now as suggested.
Do note I split the general exception into two, one that inherits std::runtime_error, and one that inherits::invalid_argument because the distinction made sense to me.
If you think it's better to only have the  std::runtime_error one let me know.

src/plugins/posix/queue_factory_impl.h
dfarge: It's inside an #ifdef HAVE_LIBURING block too.

src/plugins/posix/uring_queue.cpp
dfarge: 1. Added
2. Added

Regarding the Log_Every_N, I saw the 3 options:
* LOG_EVERY_N
* LOG_FIRST_N
* LOG_IF
I think the only appropriate option is LOG_IF, but I don't see how it offers a benefit over the current implementation

src/plugins/posix/posix_backend.h
etoledano: I think you can just inherit from std::exception because you create your own distinct type.

src/plugins/posix/posix_backend.cpp
etoledano: I do see it but not its usage.

Anyway, I don't understand why do it static when you can use the unnamed namespace, this class is internal, this should simply be a utility function in that namespace.

And you can preserve `to_string` name (pretty common) since C++ allows function overloading, so you won't have to worry about confusion, this of course applies as long as this `enum` (as you correctly implemented) is type safe (`enum class`).

```
inline const char* to_string(queue_t type) {
    switch (type) {
        case queue_t::AIO: return "AIO";
        case queue_t::URING: return "URING";
        case queue_t::UNSUPPORTED: return "UNSUPPORTED";
    }
    return "UNKNOWN";
}
```

src/plugins/posix/posix_backend.cpp
etoledano: I'm sorry, but I don't see the change you're referring to.

The main point here is to create a method that can be adopted and preserved - either by using one of the suggested approaches or your own ([like this one](https://github.com/ai-dynamo/nixl/pull/338#discussion_r2097555210)). This helps future developers avoid reinventing the wheel. It's totally fine to start with a local implementation, but the key takeaway is that we should avoid placing `#ifdef` directives anywhere below the `#include` section at the top of the file.

Keeping conditional compilation centralized prevents us from having to maintain inconsistent indentation, guess which code is active under which `#define`, or risk introducing subtle bugs.

I really appreciate your honesty about being unfamiliar with this approach. I'd be more than happy to walk through it with you in detail - it's a great opportunity to learn and practice together.

src/plugins/posix/posix_backend.h
dfarge: Joined to one class that inherits from std::exception.

src/plugins/posix/uring_queue.cpp
dfarge: We discussed this function in a different comment, please thumbs up if I can mark it as resolved.

src/plugins/posix/uring_queue.cpp
dfarge: Took your suggestion but made it a for loop instead.
Didn't find an std/builtin function to make this more concise.

src/plugins/posix/posix_backend.cpp
dfarge: Added such function.

src/plugins/posix/posix_backend.cpp
dfarge: Cool! I didn't think to overload to_string
Moved to an anonymous namespace as well.

src/plugins/posix/posix_backend.cpp
dfarge: Removed both is_read and use_aio.

src/plugins/posix/uring_queue.cpp
dfarge: Changed to passing the operation.

src/plugins/posix/posix_backend.cpp
dfarge: Let's take this offline at the office tomorrow, so I can take you up on the offer to show me the other approaches.

src/plugins/posix/aio_queue.cpp
```diff
@@ -23,8 +23,8 @@
 #include <time.h>
 #include <stdexcept>
 
-aioQueue::aioQueue(int num_entries, bool is_read)
-    : aiocbs(num_entries), num_entries(num_entries), num_completed(0), num_submitted(0), is_read(is_read) {
+aioQueue::aioQueue(int num_entries, nixl_xfer_op_t operation)
```
vvenkates27, line 26: What is the benefit of changing nixl_xfer_op_t?

src/plugins/posix/posix_backend.cpp
vvenkates27: We can remove these, I added them to debug and only triggered when INFO is set

src/plugins/posix/posix_backend.cpp
vvenkates27: They were kept initially because there was no instant specific state. However, I agree with @etoledano  that C++ namespaces would be a more idiomatic way to organize such utility functions. So not opposed to changing them.

src/plugins/posix/posix_backend.cpp
```diff
@@ -271,15 +276,24 @@ nixl_status_t nixlPosixEngine::prepXfer(const nixl_xfer_op_t &operation,
...
     }
 
     try {
         // Create a params map with our backend selection
         nixl_b_params_t params;
-        params["use_uring"] = use_aio ? "false" : "true";
-        params["use_aio"] = use_aio ? "true" : "false";
+        switch (queue_type_) {
+            case nixlPosixQueue::queue_t::AIO:
+                params["use_aio"] = "true";
```
vvenkates27, line 277: It appears you are setting params based on some queue type?
The queue type must be changed based on user backend params.

src/plugins/posix/posix_backend.h
vvenkates27: This is correct.

src/api/cpp/backend/backend_aux.h
mkhazraee: Hmm, actually I think there was a time that we made it virtual and then in some last minute change we just let it be as it was not used, don't recall what happened. But it makes sense to be like nixlMemSection that is actually pure virtual, so no object can be made from it. For that class not only we put =0 at the end, but also define an empty destructor as it had std members in it. Here that's not necessary, but maybe making it pure even makes sense. Unless I'm forgetting an issue with that caused us to do it this way, doubt.

mkhazraee: I saw Adit already started the conversation about the core and plugin, he knows more about the build infra, so follow up with him on that.

src/plugins/posix/posix_backend.cpp
dfarge: Removed in all places except in the nixlPosixEngine ctor.

src/plugins/posix/posix_backend.cpp
dfarge: Moved to an anonymous namespace.

src/api/cpp/backend/backend_aux.h
dfarge: I left it virtual and not pure virtual because that would require some changes in UCX and that is outside the scope of this PR.
Perhaps we should consider making it pure virtual in a different commit.

src/plugins/posix/aio_queue.cpp
dfarge: Readability and consistency with the user api.
Easier to read 
`aioQueue(1024, NIXL_WRITE)`
than
`aioQueue(1024, false)`

src/plugins/posix/posix_backend.cpp
dfarge: That queue_type is derived directly from the queue type picked for the nixlPosixEngine, so it is based solely on the user's backend params.

src/meson.build
dfarge: Do you see any potential issue with this change?
Worst case it's ok not to use that function

aranadive: yes, this will break static plugins as the static plugin libraries are linked to the plugin manager in core. You can move that function to src/utils/common/nixl_log.cpp?

dfarge: I reverted the change, I also removed the log that used the function. It wasn't necessary to begin with.

dfarge:
/ok to test https://github.com/ai-dynamo/nixl/commit/1aaebec9574da577ee346bc2ab8574909b52c659

src/plugins/posix/posix_backend.cpp
dfarge: We decided to postpone this for now.

vvenkates27 (approved):
Please add a proper commit message before landing.

aranadive (approved):

PR #336, opened 2025-05-19
UCX: Set a CUDA context when loading remote MD from listener thread

tvegas1:
When using listener thread to load remote metadata, UCX plugin fails to import some keys because the CUDA context is missing. Make sure to use one and retain it. Assumption is that we only need it when using VRAM memory, UCX anyways detects the correct device when actually using memory/keys. If no context exist/context is inactive, the retain call will keep it until backend is destroyed.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -55,8 +55,6 @@ public:
 class nixlUcxCudaDevicePrimaryCtx {
 #ifndef HAVE_CUDA
 public:
-
-    nixlUcxCudaDevicePrimaryCtx(int ordinal) {}
     bool push() { return false; }
     void pop() {};
 #else
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -55,21 +55,21 @@ public:
 class nixlUcxCudaDevicePrimaryCtx {
 #ifndef HAVE_CUDA
 public:
+
     nixlUcxCudaDevicePrimaryCtx(int ordinal) {}
     bool push() { return false; }
     void pop() {};
 #else
-    int m_ordinal;
-    CUdevice m_device;
-    CUcontext m_context;
+    static constexpr int defaultCudaDeviceOrdinal = 0;
+    int m_ordinal{defaultCudaDeviceOrdinal};
+    CUdevice m_device{CU_DEVICE_INVALID};
+    CUcontext m_context{nullptr};
 public:
-    nixlUcxCudaDevicePrimaryCtx(int ordinal)
-        : m_ordinal(ordinal), m_context(nullptr) {}
 
     bool push() {
         CUcontext context;
 
-        CUresult res = cuCtxGetCurrent(&context);
+        const auto res = cuCtxGetCurrent(&context);
         if (res != CUDA_SUCCESS || context != nullptr) {
             return false;
         }
@@ -593,7 +593,7 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         cuda_addr_wa = true;
     }
 
-    m_cudaPrimaryCtx = std::make_shared<nixlUcxCudaDevicePrimaryCtx>(0);
+    m_cudaPrimaryCtx = std::make_shared<nixlUcxCudaDevicePrimaryCtx>();
     vramInitCtx();
     progressThreadStart();
 }
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -52,78 +52,70 @@ public:
     int cudaSetCtx();
 };
 
-class nixlUcxCudaCtxRetained {
-#ifdef HAVE_CUDA
+class nixlUcxCudaDevicePrimaryCtx {
+#ifndef HAVE_CUDA
+public:
+    nixlUcxCudaDevicePrimaryCtx(int ordinal) {}
+    bool push() { return false; }
+    void pop() {};
+#else
+    int m_ordinal;
     CUdevice m_device;
     CUcontext m_context;
 public:
-    nixlUcxCudaCtxRetained(CUdevice device) : m_device(device) {
-        CUresult res = cuDevicePrimaryCtxRetain(&m_context, m_device);
-        if (res != CUDA_SUCCESS) {
-            m_device = -1;
-            m_context = nullptr;
+    nixlUcxCudaDevicePrimaryCtx(int ordinal)
+        : m_ordinal(ordinal), m_context(nullptr) {}
+
+    bool push() {
+        CUcontext context;
+
+        CUresult res = cuCtxGetCurrent(&context);
+        if (res != CUDA_SUCCESS || context != nullptr) {
+            return false;
         }
+
+        if (m_context == nullptr) {
+            CUresult res = cuDeviceGet(&m_device, m_ordinal);
+            if (res != CUDA_SUCCESS) {
+                return false;
+            }
+
+            res = cuDevicePrimaryCtxRetain(&m_context, m_device);
+            if (res != CUDA_SUCCESS) {
+                m_context = nullptr;
+                return false;
+            }
+        }
+
+        return cuCtxPushCurrent(m_context) == CUDA_SUCCESS;
     }
-    ~nixlUcxCudaCtxRetained() {
-        if (m_device > -1) {
+
+    void pop() {
+        cuCtxPopCurrent(nullptr);
+    }
+
+    ~nixlUcxCudaDevicePrimaryCtx() {
+        if (m_context != nullptr) {
             cuDevicePrimaryCtxRelease(m_device);
         }
-    }
-    CUcontext get() const {
-        return m_context;
     }
 #endif
 };
 
 class nixlUcxCudaCtxGuard {
-#ifndef HAVE_CUDA
+    nixlUcxCudaDevicePrimaryCtxPtr m_primary;
 public:
     nixlUcxCudaCtxGuard(nixl_mem_t nixl_mem,
-                        std::unique_ptr<nixlUcxCudaCtxRetained>& retain) {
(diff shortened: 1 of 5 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,80 @@ class nixlUcxCudaCtx {
...
+#ifdef HAVE_CUDA
+    CUdevice m_device;
+    CUcontext m_context;
+public:
+    nixlUcxCudaCtxRetained(CUdevice device) : m_device(device) {
+        CUresult res = cuDevicePrimaryCtxRetain(&m_context, m_device);
+        if (res != CUDA_SUCCESS) {
+            m_device = -1;
+            m_context = nullptr;
+        }
```
```suggestion
        const auto res = cuDevicePrimaryCtxRetain(&m_context, m_device);
        if (res != CUDA_SUCCESS) {
            m_device = CU_DEVICE_INVALID;
            m_context = nullptr;
        }
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,80 @@ class nixlUcxCudaCtx {
...
+    }
+    ~nixlUcxCudaCtxRetained() {
+        if (m_device > -1) {
```
```suggestion
        if (m_device > CU_DEVICE_INVALID) {
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,80 @@ class nixlUcxCudaCtx {
...
+            cuDevicePrimaryCtxRelease(m_device);
+        }
+    }
+    CUcontext get() const {
+        return m_context;
+    }
+#endif
+};
+
+class nixlUcxCudaCtxGuard {
```
rakhmets: Maybe it would be better in this way. Because CUcontext is encapsulated in nixlUcxCudaDevicePrimaryCtx.
```C++
class nixlUcxCudaDevicePrimaryCtx {
    CUdevice m_device   = CU_DEVICE_INVALID;
    CUcontext m_context = nullptr;

public:
    using Ptr = std::shared_ptr<nixlUcxCudaDevicePrimaryCtx>;
    nixlUcxCudaDevicePrimaryCtx(int ordinal);
    virtual ~nixlUcxCudaDevicePrimaryCtx();
    void push();
    void pop();
}

nixlUcxCudaDevicePrimaryCtx::nixlUcxCudaDevicePrimaryCtx(int ordinal)
{
    const auto res = cuDeviceGet(&m_device, ordinal);
    // handle res != CUDA_SUCCESS
}

nixlUcxCudaDevicePrimaryCtx::~nixlUcxCudaDevicePrimaryCtx()
{
    if (m_context != nullptr) {
        const auto res = cuDevicePrimaryCtxRelease(m_device);
        // handle res != CUDA_SUCCESS
    }
}

void nixlUcxCudaDevicePrimaryCtx::push()
{
    if (m_context == nullptr) {
        const auto res = cuDevicePrimaryCtxRetain(&m_context, m_device);
        // handle res != CUDA_SUCCESS
    }

    const auto res = cuCtxPushCurrent(m_context);
    // handle res != CUDA_SUCCESS
}

void nixlUcxCudaDevicePrimaryCtx::pop()
{
    const auto res = cuCtxPopCurrent(nullptr);
    // handle res != CUDA_SUCCESS
}

class nixlUcxCudaCtxGuard {
    nixlUcxCudaDevicePrimaryCtx::Ptr m_cudaDevicePrimaryCtx = nullptr;

public:
    nixlUcxCudaCtxGuard(nixl_mem_t nixl_mem,
                        nixlUcxCudaDevicePrimaryCtx::Ptr cudaDevicePrimaryCtx);
    virtual ~nixlUcxCudaCtxGuard();
}

nixlUcxCudaCtxGuard::nixlUcxCudaCtxGuard(
        nixl_mem_t nixl_mem,
        nixlUcxCudaDevicePrimaryCtx::Ptr cudaDevicePrimaryCtx)
{
    if (nixl_mem != VRAM_SEG) {
        return;
    }

    const auto res = cuCtxGetCurrent(&context);
    // handle res != CUDA_SUCCESS

    if (context != nullptr) {
        return; // There is already a context set
    }

    m_cudaDevicePrimaryCtx = cudaDevicePrimaryCtx;
    if (m_cudaDevicePrimaryCtx) {
        m_cudaDevicePrimaryCtx->push();
    }
}

nixlUcxCudaCtxGuard::~nixlUcxCudaCtxGuard()
{
    if (m_cudaDevicePrimaryCtx) {
        m_cudaDevicePrimaryCtx->pop();
    }
}

// make_shared<nixlUcxCudaDevicePrimaryCtx>(0) in nixlUcxEngine ctor
```

src/plugins/ucx/ucx_backend.cpp
tvegas1: reworked

src/plugins/ucx/ucx_backend.cpp
tvegas1: reworked

src/plugins/ucx/ucx_backend.cpp
tvegas1: moved cuda calls to first class

test/gtest/test_transfer.cpp
```diff
@@ -54,9 +56,13 @@ class MemBuffer : std::shared_ptr<void> {
 private:
     static void *allocate(size_t size, nixl_mem_t mem_type)
     {
+        void *ptr;
+
         switch (mem_type) {
         case DRAM_SEG:
             return malloc(size);
+        case VRAM_SEG:
+            return cudaSuccess == cudaMalloc(&ptr, size)? ptr : nullptr;
```
brminich, line 67: shoud it be guarded with HAVE_CUDA?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,72 @@ class nixlUcxCudaCtx {
     int cudaSetCtx();
 };
 
+class nixlUcxCudaDevicePrimaryCtx {
+#ifndef HAVE_CUDA
```
brminich, line 56: id' try to unite it with `nixlUcxCudaCtx` somehow, but let's do it later in next PR

rakhmets: I don't think they should be united. These are two different classes, similar to the combination of a mutex and a lock.

tvegas1:
CI failure, CUDA device not available:
```
ERROR: The NVIDIA Driver is present, but CUDA failed to initialize.  GPU functionality will not be available.
   [[ Unable to initialize CUDA driver (error ???) ]]
```

test/gtest/test_transfer.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
tvegas1: agree

brminich (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,72 @@ class nixlUcxCudaCtx {
...
+public:
+    nixlUcxCudaDevicePrimaryCtx(int ordinal) {}
+    bool push() { return false; }
+    void pop() {};
+#else
+    int m_ordinal;
+    CUdevice m_device;
+    CUcontext m_context;
```
rakhmets: I would recommend to use default member initialization. In this case, we will get predictable behavior in the case of using default ctor. And you can delete `m_context(nullptr)` from ctor then.
```suggestion
    int m_ordinal{0};
    CUdevice m_device{CU_DEVICE_INVALID};
    CUcontext m_context{nullptr};
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,72 @@ class nixlUcxCudaCtx {
...
+public:
+    nixlUcxCudaDevicePrimaryCtx(int ordinal)
+        : m_ordinal(ordinal), m_context(nullptr) {}
+
+    bool push() {
+        CUcontext context;
+
+        CUresult res = cuCtxGetCurrent(&context);
```
rakhmets: CUresult can be replaced by `const auto`/`auto`.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -52,6 +52,72 @@ class nixlUcxCudaCtx {
...
+    ~nixlUcxCudaDevicePrimaryCtx() {
+        if (m_context != nullptr) {
+            cuDevicePrimaryCtxRelease(m_device);
+        }
+    }
+#endif
+};
+
+class nixlUcxCudaCtxGuard {
+    nixlUcxCudaDevicePrimaryCtxPtr m_primary;
```
rakhmets, line 104:
```suggestion
    nixlUcxCudaDevicePrimaryCtxPtr m_primary{nullptr};
```

src/plugins/ucx/ucx_backend.cpp
rakhmets: And I think it is better to have `constexpr int defaultCudaDeviceOrdinal{0}`. And use it as default value for the class filed `m_ordinal{defaultCudaDeviceOrdinal}`.

And you can use this value for creation of the shared pointer, or use default ctor without arguments.

src/plugins/ucx/ucx_backend.cpp
tvegas1: but shared_ptr is already setting to nullptr?

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed

src/plugins/ucx/ucx_backend.cpp
rakhmets: Right, this is redundant.

src/plugins/ucx/ucx_backend.cpp
tvegas1: fixed (also probably a mater of taste)

brminich (approved):

PR #340, opened 2025-05-20
fix: add python and rust attribution files

saturley-hall:
Enumerate rust and python dependencies for NIXL project


nv-anants (approved):

dmitry-tokarev-nv (approved):
:shipit:

PR #344, opened 2025-05-21
clang: Add clang format file

aranadive:
Lets start with this version of the file and see if the changes to nixl agent and a ucx header are okay.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,71 @@
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: Empty
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterReturnType: All
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: false
+  AfterNamespace: false
+  AfterObjCDeclaration: false
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeCatch: true
+  BeforeElse: false
+  BeforeLambdaBody: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: true
+ColumnLimit: 100
+CompactNamespaces: false
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 8
+DerivePointerAlignment: false
+IndentCaseLabels: false
+IndentPPDirectives: None
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: Inner
+
+PackConstructorInitializers: CurrentLine
+PointerAlignment: Right
+ReferenceAlignment: Right
+ReflowComments: true
+SortIncludes: false
+SortUsingDeclarations: false
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: false
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCtorInitializerColon: true
+SpaceBeforeInheritanceColon: true
+SpaceBeforeParens: ControlStatements
+SpaceBeforeRangeBasedForLoopColon: true
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles: false
+SpacesInContainerLiterals: false
+SpacesInSquareBrackets: false
+TabWidth: 4
+UseTab: Never
+...


aranadive:
If you use split diff view it easier to see the differences.

.clang-format
```diff
@@ -0,0 +1,71 @@
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
```
w1ldptr: Aligning function arguments by bracket looks nicer IMO, but no strong opinion.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  BeforeCatch: true
+  BeforeElse: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Linux
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: true
+ColumnLimit: 100
```
w1ldptr, line 39: Lower column limit looks okay in C project, but with long C++ type names it tends to get ugly in places. Look at the declaration of remoteConnMap for example.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
```
w1ldptr: Break after return type seems much nicer to me, but this is purely subjective.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+AlignEscapedNewlines: Left
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
```
w1ldptr: This is deprecated and might conflict with AlwaysBreakAfterReturnType.

.clang-format
```diff
@@ -0,0 +1,71 @@
+---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: true
```
w1ldptr: This likely doesn't work since 'true' is not one of [the accepted values](https://releases.llvm.org/18.1.6/tools/clang/docs/ClangFormatStyleOptions.html#allowshortblocksonasingleline). However, I would propose just setting it to 'Empty' (i.e. only put empty {} blocks on the same line), otherwise it would make code very hard to read in some places.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: true
+  AfterNamespace: true
```
w1ldptr: This makes nested namespace definitions quite verbose:
```
namespace A
{ 
  namespace B 
  { 
    namespace C 
    { 
      // ... 
    } 
  } 
}
```

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  AfterObjCDeclaration: false
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeCatch: true
+  BeforeElse: false
```
w1ldptr, line 32: Also add "BeforeLambdaBody: false" ?

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Linux
+BreakBeforeInheritanceComma: false
```
w1ldptr: I don't see this one in any recent-ish clanf-format manuals, so I assume it has been deprecated and removed.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: true
+ColumnLimit: 100
+CompactNamespaces: false
```
w1ldptr, line 40: 'true' looks much nicer with any kind of nested namespaces IMO.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
```
w1ldptr: This has been deprecated and I assume will also be removed soon. Use PackConstructorInitializers instead.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 8
+DerivePointerAlignment: false
+IndentCaseLabels: true
```
w1ldptr: Why enable this? With long C++ typenames and 100 column lines it would make many switch blocks look terrible.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+IndentPPDirectives: None
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: All
```
w1ldptr: As soon as we start using namespaces this will just unnecessarily indent all code. Even worse with nested namespaces.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+
+PointerAlignment: Right
+ReflowComments: true
+SortIncludes: false
+SortUsingDeclarations: false
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: true
```
w1ldptr: Not having a space when declaring instances of templates like "std::map<std::string, void *>" but having them in the definitions like "template <int> void foo()" looks a bit weird.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCtorInitializerColon: true
+SpaceBeforeInheritanceColon: true
+SpaceBeforeParens: ControlStatements
+SpaceBeforeRangeBasedForLoopColon: true
```
w1ldptr, line 63: Hmm, never encountered this kind of formatting "for (auto &elm : remoteSections)" in  C++ code.

.clang-format
aranadive: Done

.clang-format
aranadive: If we keep indentation minimal, maybe its okay. We can increase later if required.

.clang-format
aranadive: So toplevel here. Also changed to BreakAfterReturnType.

.clang-format
aranadive: Removed.

.clang-format
aranadive: Yup, Empty it is.

.clang-format
aranadive: Removed.. and used PackConstructorInitializers

.clang-format
aranadive: Maybe inner is better for now. If its still too much, ask for None.

.clang-format
aranadive: I tihnk its a bit cleaner to know the loop.

.clang-format
aranadive: We dont have too many nested NS as yet. We might in the future, so set to false now.

.clang-format
aranadive: Removed.

mkhazraee (approved):

.clang-format
w1ldptr: So the first problem I see with any kind of indentation for namespaces is that we don't use namespaces at the moment, then we commit to clang-format reformatting everything for some release, and then someone adds a namespace afterwards and whole code has to be reformatted yet again.

Second issue is that if we end with namespace style like "nvda::dyn::nixl" or something, then even with Inner we will have 8 char base indentation which would look even worse due to 100char line size limit.

Also, the whole idea of namespace indentation is when you really break things into multiple namespaces even in a single file. Looking at the current structure of NIXL API I don't see where any kind of such fine-grained namespacing would fit and I assume we will just end up with everything wrapped in a single namespace. Stylistically I don't have strong opinion on this, just thinking out loud how it fits our current and future API so feel free to ignore.

aranadive: Yeah so for now i added Inner instead of all. We save atleast some indentation and column space. We dont a plan a for namespace style as yet or major changes for APIs. I think inner achieves a balance for now and we can switch to none when there is a plan to use multiple nested namespaces.

w1ldptr (approved):

PR #345, opened 2025-05-21
one to one mapping from thread to ucx_worker for better performance

Pyngon:
Previously, each thread choose ucx_worker by using a hash of thread id. Most of the time, this would lead to multiple thread using the same ucx_worker and cause contention.

Additionally, added an argument "threading_mode" for user to choose the threading mode that used by UCX.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -123,6 +123,9 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
                 }
             }
         }
+        if (0 == xferBenchConfig::backend.compare(XFERBENCH_BACKEND_UCX)) {
+            backend_params["num_workers"] = std::to_string(xferBenchConfig::num_threads);
+        }
 
         if (gethostname(hostname, 256)) {
             std::cerr << "Failed to get hostname" << std::endl;
@@ -165,7 +168,10 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
         exit(EXIT_FAILURE);
     }
 
-    agent->createBackend(backend_name, backend_params, backend_engine);
+    if (agent->createBackend(backend_name, backend_params, backend_engine) != NIXL_SUCCESS) {
+        std::cerr << "Backend creation fail: " << xferBenchConfig::backend << std::endl;
+        exit(EXIT_FAILURE);
+    }
 }
 
 xferBenchNixlWorker::~xferBenchNixlWorker() {
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -124,7 +124,7 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
             }
         }
         if (0 == xferBenchConfig::backend.compare(XFERBENCH_BACKEND_UCX)) {
-            backend_params["num_workers"] = std::to_string(xferBenchConfig::num_threads);
+            backend_params["num_dedicated_workers"] = std::to_string(xferBenchConfig::num_threads);
         }
 
         if (gethostname(hostname, 256)) {
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -459,8 +459,9 @@ void nixlUcxEngine::progressFunc()
             } while (status == NIXL_IN_PROG);
             NIXL_ASSERT(status == NIXL_SUCCESS) << ", status: " << status;
 
-            if (made_progress && !wid)
+            if (made_progress && wid == recvNotifWorkerId) {
                 notifProgress();
+            }
         }
         timeout = false;
 
@@ -556,8 +557,21 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
         devs = str_split((*custom_params)["device_list"], ", ");
 
     const auto num_workers_iter = custom_params->find("num_workers");
-    if (num_workers_iter == custom_params->end() || !absl::SimpleAtoi(num_workers_iter->second, &numWorkers))
-        numWorkers = 1;
+    if (num_workers_iter == custom_params->end() ||
+        !absl::SimpleAtoi(num_workers_iter->second, &numWorkers)) {
+        // Default to have 1 shared worker and 0 dedicated worker
+        numWorkers = 0;
+    }
+
+    const auto num_shared_iter = custom_params->find("num_shared_workers");
+    if (num_shared_iter == custom_params->end() ||
+        !absl::SimpleAtoi(num_shared_iter->second, &numSharedWorkers)) {
+        numSharedWorkers = 1;
+    }
+
+    if (numWorkers + numSharedWorkers == 0) {
+        throw std::invalid_argument("Total number of workers must be greater than 0");
+    }
 
     ucp_err_handling_mode_t err_handling_mode;
     const auto err_handling_mode_it =
@@ -576,11 +590,29 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
                                           numWorkers,
                                           init_params->syncMode);
 
-    for (size_t i = 0; i < numWorkers; i++) {
-        uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode));
+    for (size_t i = 0; i < numSharedWorkers; i++) {
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode, true));
     }
 
-    workerAddr = uws.front()->epAddr();
+    for (size_t i = 0; i < numWorkers; i++) {
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode, false));
+        freeWorkers.push(i + numSharedWorkers);
+    }
+
+    const auto &uw = uws.front();
+    workerAddr = uw->epAddr();
+    recvNotifWorkerId = 0; // First worker from the list
+
+    if (workerAddr.empty()) {
+        NIXL_ERROR << "Failed to get UCX worker address";
+        initErr = true;
+        return;
+    }
+
+    if (initThreadMapping() != NIXL_SUCCESS) {
+        initErr = true;
+        return;
+    }
 
     if (pthrOn) {
         for (auto &uw: uws) {
@@ -628,6 +660,7 @@ nixlUcxEngine::~nixlUcxEngine() {
     }
 
     vramFiniCtx();
+    destroyThreadMapping();
 }
 
 /****************************************
@@ -932,8 +965,14 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        nixlBackendReqH* &handle,
                                        const nixl_opt_b_args_t* opt_args) const
 {
+    size_t worker_id;
+    nixl_status_t status = getWorkerIdWithPreference(false, worker_id);
+    if (status != NIXL_SUCCESS) {
+        return status;
+    }
+
     /* TODO: try to get from a pool first */
-    nixlUcxBackendH *intHandle = new nixlUcxBackendH(*this, getWorkerId());
+    nixlUcxBackendH *intHandle = new nixlUcxBackendH(*this, worker_id);
 
     handle = (nixlBackendReqH*)intHandle;
     return NIXL_SUCCESS;
@@ -1205,14 +1244,19 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
 {
     nixl_status_t ret;
     nixlUcxReq req;
-    size_t wid = getWorkerId();
 
-    ret = notifSendPriv(remote_agent, msg, req, wid);
+    size_t worker_id;
+    ret = getWorkerIdWithPreference(true, worker_id);
+    if (ret != NIXL_SUCCESS) {
+        return ret;
+    }
+
+    ret = notifSendPriv(remote_agent, msg, req, worker_id);
 
     switch(ret) {
     case NIXL_IN_PROG:
         /* do not track the request */
-        getWorker(wid)->reqRelease(req);
+        uws[worker_id]->reqRelease(req);
     case NIXL_SUCCESS:
         break;
     default:
@@ -1220,4 +1264,105 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
         return ret;
     }
     return NIXL_SUCCESS;
+}
+
+/****************************************
+ * Thread to worker mapping
+ *****************************************/
+
+void
+nixlUcxEngine::threadMapDestructor(void *arg) {
+    auto worker_info = static_cast<nixlUcxWorkerInfo *>(arg);
+
+    /* When a thread exits, disassociate it from the worker,
+     * so that the worker could be reused by another thread.
+     */
+    worker_info->engine->pushFreeWorker(worker_info->workerId);
+
+    // TODO: memory leak if engine obj is destroy before thread exit?
+    delete worker_info;
+}
+
+nixl_status_t
+nixlUcxEngine::initThreadMapping() {
+    int err = pthread_key_create(&keyThreadToWorker, threadMapDestructor);
+    if (err) {
+        NIXL_ERROR << "Failed to create keyThreadToWorker, err:" << err;
+        return NIXL_ERR_BACKEND;
+    }
+    return NIXL_SUCCESS;
+}
+
+void
+nixlUcxEngine::destroyThreadMapping() {
+    int err = pthread_key_delete(keyThreadToWorker);
+    if (err) {
+        NIXL_WARN << "Failed to delete keyThreadToWorker, err:" << err;
+    }
+}
+
+nixl_status_t
+nixlUcxEngine::getFreeDedicatedWorkerId(size_t &worker_id) const {
+    const std::lock_guard<std::mutex> lock(workersMutex);
+    if (freeWorkers.empty()) {
+        return NIXL_ERR_BACKEND;
+    }
+
+    worker_id = freeWorkers.front();
+    freeWorkers.pop();
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlUcxEngine::getDedicatedWorkerId(size_t &worker_id) const {
+    auto worker_info = static_cast<nixlUcxWorkerInfo *>(pthread_getspecific(keyThreadToWorker));
+    if (worker_info != nullptr) {
+        worker_id = worker_info->workerId;
+        return NIXL_SUCCESS;
+    }
+
+    nixl_status_t status = getFreeDedicatedWorkerId(worker_id);
+    if (status != NIXL_SUCCESS) {
+        return status;
+    }
+
+    worker_info = new nixlUcxWorkerInfo(this, worker_id);
+    if (worker_info == nullptr) {
+        pushFreeWorker(worker_id);
+        NIXL_ERROR << "Failed to allocate nixlUcxWorkerInfo";
+        return NIXL_ERR_BACKEND;
+    }
+
+    int err = pthread_setspecific(keyThreadToWorker, worker_info);
+    if (err) {
+        delete worker_info;
+        pushFreeWorker(worker_id);
+        NIXL_ERROR << "Failed to set keyThreadToWorker, workerId:" << worker_id << "err:" << err;
+        return NIXL_ERR_BACKEND;
+    }
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlUcxEngine::getSharedWorkerId(size_t &worker_id) const {
+    if (numSharedWorkers == 0) {
+        /* no shared workers */
+        return NIXL_ERR_BACKEND;
+    }
+
+    worker_id = std::hash<std::thread::id>{}(std::this_thread::get_id()) % numSharedWorkers;
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlUcxEngine::getWorkerIdWithPreference(bool prefer_shared, size_t &worker_id) const {
+    nixl_status_t status =
+        prefer_shared ? getSharedWorkerId(worker_id) : getDedicatedWorkerId(worker_id);
+    if (status == NIXL_SUCCESS) {
+        return status;
+    }
+
+    // If the preferred worker type is not available, get the other type
+    return prefer_shared ? getDedicatedWorkerId(worker_id) : getSharedWorkerId(worker_id);
 }
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -532,7 +532,7 @@ void nixlUcxEngine::progressThreadRestart()
 nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
     : nixlBackendEngine(init_params),
       pthrControlPipe{0, 0} {
-    size_t numWorkers;
+    size_t numDedicatedWorkers;
     std::vector<std::string> devs; /* Empty vector */
     nixl_b_params_t* custom_params = init_params->customParams;
 
@@ -561,18 +561,18 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
 
     const auto num_workers_iter = custom_params->find("num_workers");
     if (num_workers_iter == custom_params->end() ||
-        !absl::SimpleAtoi(num_workers_iter->second, &numWorkers)) {
+        !absl::SimpleAtoi(num_workers_iter->second, &numSharedWorkers)) {
         // Default to have 1 shared worker and 0 dedicated worker
-        numWorkers = 0;
-    }
-
-    const auto num_shared_iter = custom_params->find("num_shared_workers");
-    if (num_shared_iter == custom_params->end() ||
-        !absl::SimpleAtoi(num_shared_iter->second, &numSharedWorkers)) {
         numSharedWorkers = 1;
     }
 
-    if (numWorkers + numSharedWorkers == 0) {
+    const auto num_shared_iter = custom_params->find("num_dedicated_workers");
+    if (num_shared_iter == custom_params->end() ||
+        !absl::SimpleAtoi(num_shared_iter->second, &numDedicatedWorkers)) {
+        numDedicatedWorkers = 0;
+    }
+
+    if (numDedicatedWorkers + numSharedWorkers == 0) {
         throw std::invalid_argument("Total number of workers must be greater than 0");
     }
 
@@ -590,14 +590,14 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
                                           _internalRequestInit,
                                           _internalRequestFini,
                                           pthrOn,
-                                          numWorkers,
+                                          numDedicatedWorkers + numSharedWorkers,
                                           init_params->syncMode);
 
     for (size_t i = 0; i < numSharedWorkers; i++) {
         uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode, true));
     }
 
-    for (size_t i = 0; i < numWorkers; i++) {
+    for (size_t i = 0; i < numDedicatedWorkers; i++) {
         uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode, false));
         freeWorkers.push(i + numSharedWorkers);
     }
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -19,7 +19,6 @@
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
 #include "common/nixl_log.h"
-
 #include <optional>
 #include <limits>
 #include <string.h>
@@ -42,6 +41,10 @@ namespace {
         }
     }
 }
+
+// Static member definitions
+std::unordered_set<nixlUcxEngine *> nixlUcxEngine::engineSet;
+std::mutex nixlUcxEngine::engineSetMutex;
 
 /****************************************
  * CUDA related code
@@ -639,6 +642,8 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
     m_cudaPrimaryCtx = std::make_shared<nixlUcxCudaDevicePrimaryCtx>();
     vramInitCtx();
     progressThreadStart();
+
+    nixlUcxEngine::addEngine(this);
 }
 
 nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
@@ -649,7 +654,17 @@ nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
 }
 
 // Through parent destructor the unregister will be called.
-nixlUcxEngine::~nixlUcxEngine() {
+nixlUcxEngine::~nixlUcxEngine () {
+    // per registered memory deregisters it, which removes the corresponding metadata too
+    // parent destructor takes care of the desc list
+    // For remote metadata, they should be removed here
+    if (this->initErr) {
+        // Nothing to do
+        return;
+    }
+
+    nixlUcxEngine::removeEngine(this);
+
     progressThreadStop();
     if (pthrOn) {
         for (const auto pthr_control_pipe : pthrControlPipe) {
@@ -1272,22 +1287,32 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
 
 void
 nixlUcxEngine::threadMapDestructor(void *arg) {
-    auto worker_info = static_cast<nixlUcxWorkerInfo *>(arg);
+    auto engine = static_cast<nixlUcxEngine *>(arg);
+
+    std::lock_guard<std::mutex> lock(nixlUcxEngine::engineSetMutex);
+    if (!nixlUcxEngine::isEngineExist(engine)) {
+        NIXL_DEBUG << "Engine not found in threadMapDestructor, engine:" << engine;
+        return;
+    }
 
     /* When a thread exits, disassociate it from the worker,
      * so that the worker could be reused by another thread.
      */
-    worker_info->engine->pushFreeWorker(worker_info->workerId);
-
-    // TODO: memory leak if engine obj is destroy before thread exit?
-    delete worker_info;
+    engine->pushFreeWorker(reinterpret_cast<size_t>(pthread_getspecific(engine->pKeyWorkerId)));
 }
 
 nixl_status_t
 nixlUcxEngine::initThreadMapping() {
-    int err = pthread_key_create(&keyThreadToWorker, threadMapDestructor);
+    int err = pthread_key_create(&pKeyEngine, threadMapDestructor);
     if (err) {
-        NIXL_ERROR << "Failed to create keyThreadToWorker, err:" << err;
+        NIXL_ERROR << "Failed to create pKeyEngine, err:" << err;
+        return NIXL_ERR_BACKEND;
+    }
+
+    err = pthread_key_create(&pKeyWorkerId, NULL);
+    if (err) {
+        NIXL_ERROR << "Failed to create pKeyWorkerId, err:" << err;
+        pthread_key_delete(pKeyEngine);
         return NIXL_ERR_BACKEND;
     }
     return NIXL_SUCCESS;
@@ -1295,9 +1320,14 @@ nixlUcxEngine::initThreadMapping() {
 
 void
 nixlUcxEngine::destroyThreadMapping() {
-    int err = pthread_key_delete(keyThreadToWorker);
+    int err = pthread_key_delete(pKeyEngine);
     if (err) {
-        NIXL_WARN << "Failed to delete keyThreadToWorker, err:" << err;
+        NIXL_WARN << "Failed to delete pKeyEngine, err:" << err;
+    }
+
+    err = pthread_key_delete(pKeyWorkerId);
+    if (err) {
+        NIXL_WARN << "Failed to delete pKeyWorkerId, err:" << err;
     }
 }
 
@@ -1315,9 +1345,9 @@ nixlUcxEngine::getFreeDedicatedWorkerId(size_t &worker_id) const {
 
 nixl_status_t
 nixlUcxEngine::getDedicatedWorkerId(size_t &worker_id) const {
-    auto worker_info = static_cast<nixlUcxWorkerInfo *>(pthread_getspecific(keyThreadToWorker));
-    if (worker_info != nullptr) {
-        worker_id = worker_info->workerId;
+    auto engine = static_cast<nixlUcxEngine *>(pthread_getspecific(pKeyEngine));
+    if (engine != nullptr) {
+        worker_id = reinterpret_cast<size_t>(pthread_getspecific(pKeyWorkerId));
         return NIXL_SUCCESS;
     }
 
@@ -1326,18 +1356,16 @@ nixlUcxEngine::getDedicatedWorkerId(size_t &worker_id) const {
         return status;
     }
 
-    worker_info = new nixlUcxWorkerInfo(this, worker_id);
-    if (worker_info == nullptr) {
+    int err = pthread_setspecific(pKeyWorkerId, reinterpret_cast<void *>(worker_id));
+    if (err) {
         pushFreeWorker(worker_id);
-        NIXL_ERROR << "Failed to allocate nixlUcxWorkerInfo";
+        NIXL_ERROR << "Failed to set pKeyWorkerId, workerId:" << worker_id << "err:" << err;
         return NIXL_ERR_BACKEND;
     }
-
-    int err = pthread_setspecific(keyThreadToWorker, worker_info);
+    err = pthread_setspecific(pKeyEngine, this);
     if (err) {
-        delete worker_info;
         pushFreeWorker(worker_id);
-        NIXL_ERROR << "Failed to set keyThreadToWorker, workerId:" << worker_id << "err:" << err;
+        NIXL_ERROR << "Failed to set pKeyEngine, workerId:" << worker_id << "err:" << err;
         return NIXL_ERR_BACKEND;
     }
 
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -626,7 +626,6 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
 
     // TODO: in case of UCX error handling is enabled, we can clean up AM based connections error
     //       handling, if user requested disabled error handling, we dont care about it.
-    auto &uw = uws.front();
     uw->regAmCallback(CONN_CHECK, connectionCheckAmCb, this);
     uw->regAmCallback(DISCONNECT, connectionTermAmCb, this);
     uw->regAmCallback(NOTIF_STR, notifAmCb, this);
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -18,6 +18,7 @@
 #define NIXL_SRC_PLUGINS_UCX_UCX_BACKEND_H
 
 #include <vector>
+#include <queue>
 #include <cstring>
 #include <iostream>
 #include <thread>
@@ -110,6 +111,8 @@ class nixlUcxEngine
         std::unique_ptr<nixlUcxContext> uc;
         std::vector<std::unique_ptr<nixlUcxWorker>> uws;
         std::string workerAddr;
+        size_t recvNotifWorkerId;
+        size_t numSharedWorkers;
 
         /* Progress thread data */
         std::mutex pthrActiveLock;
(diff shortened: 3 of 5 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1170,3 +1202,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
 }
+
+/****************************************
+ * Thread to worker mapping
+*****************************************/
+
+nixlUcxThreadToWorker::nixlUcxThreadToWorker(size_t num_workers, nixl_ucx_mt_t mode)
+    : numWorkers(num_workers), mode(mode)
+{
+    pthread_key_create(&keyThreadToWorker, nullptr);
```
yosefe: pls add destructor that would remove entry from the map and continue using single threaded mode if possible. this means the logic of reusing worker must be changed, for example use reference counting.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -500,6 +520,8 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         return;
     }
 
+    threadWorkerMap = std::make_unique<nixlUcxThreadToWorker>(numWorkers, mode);
```
yosefe: can it be direct member var and not a pointer?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1154,7 +1182,11 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
 {
     nixl_status_t ret;
     nixlUcxReq req;
-    size_t wid = getWorkerId();
+    size_t wid;
+    ret = threadWorkerMap->threadToWorkerId(wid);
```
yosefe: space line before

src/plugins/ucx/ucx_backend.cpp
Pyngon: yosefe, thanks for the review. Could you help to clarify what do you mean here? worker mode could not be changed after the initialization.

src/plugins/ucx/ucx_backend.cpp
Pyngon: It can't be a direct member because prepXfer is a const function, when it get the worker id it will need to increment the atomic nextWorkerId field.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -512,9 +512,24 @@ void nixlUcxEngine::progressThreadRestart()
 nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
 : nixlBackendEngine (init_params) {
     unsigned long numWorkers;
+    nixl_ucx_mt_t mode = NIXL_UCX_MT_MAX;
     std::vector<std::string> devs; /* Empty vector */
     nixl_b_params_t* custom_params = init_params->customParams;
 
+    if (custom_params->count("worker_mode") != 0) {
```
iyastreb: I would propose to have a separate parse function, that we can also reuse in other places:
```
nixl_ucx_mt_t parseNixlUcxMt(const std::string& str) {
    if (str == "single") {
        return NIXL_UCX_MT_SINGLE;
...
}

File: benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp [hunk 5]
```diff
@@ -128,6 +128,10 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
                 }
             }
         }
+        if (0 == xferBenchConfig::backend.compare(XFERBENCH_BACKEND_UCX)) {
+            backend_params.emplace("num_workers", std::to_string(xferBenchConfig::num_threads));
+            backend_params.emplace("worker_mode", xferBenchConfig::threading_mode);
```
iyastreb: maybe we better keep identical names in params and config for better code clarity?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -512,9 +512,24 @@ void nixlUcxEngine::progressThreadRestart()
 nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
```
iyastreb: since init_params can't be NULL, we can pass by ref

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -512,9 +512,24 @@ void nixlUcxEngine::progressThreadRestart()
...
+    if (custom_params->count("worker_mode") != 0) {
+        if (custom_params->at("worker_mode") == "single") {
+            mode = NIXL_UCX_MT_SINGLE;
+        } else if (custom_params->at("worker_mode") == "serialized") {
+            mode = NIXL_UCX_MT_CTX;
+        } else if (custom_params->at("worker_mode") == "multi") {
+            mode = NIXL_UCX_MT_WORKER;
+        } else {
+            NIXL_ERROR << "Invalid worker mode: " << custom_params->at("worker_mode");
+            this->initErr = true;
```
iyastreb: It doesn't seem right that params validation is done in the class consuming params
Ideally we should validate params before using them, right after params construction.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -527,6 +542,11 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
             this->initErr = true;
             return;
         }
+        if (mode == NIXL_UCX_MT_SINGLE || mode == NIXL_UCX_MT_CTX) {
+            NIXL_ERROR << "Non-NIXL_UCX_MT_WORKER mode does not support progress thread";
+            this->initErr = true;
```
iyastreb: IMO `initErr` is bad design pattern - it means that we construct semi-valid object, which is not good.
For example, we might open a pipe before, but then mark object as invalid without closing the pipe.
And it's not closed in the dtor as well, because now object is in error state, so we have a resource leak.

As a rule, ctor should either create a valid object or throw an exception, creating semi-valid object is error prone

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -128,6 +128,10 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
                 }
             }
         }
+        if (0 == xferBenchConfig::backend.compare(XFERBENCH_BACKEND_UCX)) {
```
tvegas1, line 126: shall we include UCX_MO too?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+            }
+            cur_id = ++nextWorkerId;
+            if (cur_id > numWorkers) {
+                return NIXL_ERR_NOT_SUPPORTED;
+            }
+        }
+        pthread_setspecific(keyThreadToWorker, (void*)cur_id);
+    }
+
+    // 0 mean the id not exists, so we save the worker id as + 1
```
iyastreb: look like discrepancy between comment and source code

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+}
+
+nixl_status_t nixlUcxThreadToWorker::threadToWorkerId(size_t &worker_id)
+{
+    size_t cur_id = (size_t)pthread_getspecific(keyThreadToWorker);
+    if (cur_id == 0) {
+        if (mode == NIXL_UCX_MT_WORKER) {
+            do {
+                cur_id = ++nextWorkerId;
+            } while (cur_id == 0);
```
iyastreb: I think cycle is redundant

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+            cur_id = ((cur_id - 1) % numWorkers) + 1;
+        } else {
+            if (nextWorkerId >= numWorkers) {
```
iyastreb: maybe this check is not needed, it's duplicated by the next one

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+                return NIXL_ERR_NOT_SUPPORTED;
+            }
+            cur_id = ++nextWorkerId;
+            if (cur_id > numWorkers) {
```
iyastreb: I would suggest to return size_t, and have an assertion here instead of status

src/plugins/ucx/ucx_backend.h
```diff
@@ -135,6 +148,8 @@ class nixlUcxEngine : public nixlBackendEngine {
         std::unordered_map<std::string, ucx_connection_ptr_t,
                            std::hash<std::string>, strEqual> remoteConnMap;
 
+        // Thread to worker mapping
+        std::unique_ptr<nixlUcxThreadToWorker> threadWorkerMap;
```
iyastreb: Can we solve this problem simpler?
If engine is a singleton maybe we can just use function as before:
```
size_t getWorkerId() const {
    static atomic<size_t> idCounter(0); // global should be ok as long as engine is singleton
    static thread_local size_t threadId = idCounter++;
    if (mode == NIXL_UCX_MT_WORKER) {
         return threadId % uws.size();
     } else {
         // assert on thread count < num workers, etc
     }
}

Comment (Pyngon) [reply src/plugins/ucx/ucx_backend.h]: Is engine a singleton? It does not look like it is singleton.

File: src/plugins/ucx/ucx_backend.cpp (hunk 12 above)
Comment (Pyngon) [reply src/plugins/ucx/ucx_backend.cpp]: Try to prevent keep incrementing ++nextWorkerId and eventually overflow.

File: src/plugins/ucx/ucx_backend.cpp (hunk 11 above)
Comment (Pyngon) [reply src/plugins/ucx/ucx_backend.cpp]: In the case when nextWorkerId is INT_MAX, it looks cleaner this way.

File: benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp (hunk 9 above)
Comment (Pyngon) [reply benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp:126]: This patch focus on UCX only, and UCX_MO does not need this parameter now. I will leave it to the future patch if UCX_MO also needs something similar.

File: src/plugins/ucx/ucx_backend.cpp (hunk 11 above)
Comment (iyastreb) [reply src/plugins/ucx/ucx_backend.cpp]: If the intention is to handle INT_MAX, then pls consider adding explicit assertion on that.
IMO cycle is not needed, as it just hides the issue instead of reporting it (even if this issue cannot happen anyway) and makes code less readable

File: src/plugins/ucx/ucx_backend.cpp (hunk 12 above)
Comment (iyastreb) [reply src/plugins/ucx/ucx_backend.cpp]: ok, then single check before increment is sufficient, but not both of them

File: src/plugins/ucx/ucx_backend.h (hunk 14 above)
Comment (iyastreb) [reply src/plugins/ucx/ucx_backend.h]: Right, it seems it's not a singleton, so what I proposed is not an option

Still I think this class can be simplified to do only one thing (according to single responsibility principle): assign unique ID per each thread, without doing anything else like checking mode or workers number:
```
class nixlUcxThreadMap { // just a thread map
    private:
        pthread_key_t keyThreadToWorker;
        std::atomic<size_t> nextWorkerId;

    public:
        size_t getThreadId() {
            size_t cur_id = (size_t)pthread_getspecific(keyThreadToWorker);
            if (cur_id == 0) {
                cur_id = ++nextWorkerId;
            }
            return cur_id;
        }
};
```
And all the checking logic can be implemented in engine, because it has already all the data to decide. IMO no need to move this logic to the map.

Also, there is no need to have a unique_ptr here, plain object is enough.

File: src/plugins/ucx/ucx_backend.cpp [hunk 15]
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
 }
+
+/****************************************
+ * Thread to worker mapping
+*****************************************/
+
+nixlUcxThreadToWorker::nixlUcxThreadToWorker(size_t num_workers, nixl_ucx_mt_t mode)
+    : numWorkers(num_workers), mode(mode)
+{
+    pthread_key_create(&keyThreadToWorker, nullptr);
```
w1ldptr: Please use C++ 'thread_local' for much more straightforward code in this class.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I'm not entrily convinced in usefulness of this parameter to begin with. Previously UCX MT mode was deduced from other parameters (progress thread, requested number of workers, etc.) to set to the least restrictive _but correct_ mode according to them. Now, user can set anything. Afterwards there is some validation like rejecting modes incompatible with progress thread, but it is not exhaustive. For example, the user can request NIXL_UCX_MT_SINGLE with NIXL_THREAD_SYNC_RW sync mode which allows concurrent access from nixlAgent but initializes ucp worker without locking.

Could you describe your motivation for introducing this complication?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1238,3 +1271,46 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+    if (cur_id == 0) {
+        if (mode == NIXL_UCX_MT_WORKER) {
+            do {
+                cur_id = ++nextWorkerId;
+            } while (cur_id == 0);
+            cur_id = ((cur_id - 1) % numWorkers) + 1;
+        } else {
+            if (nextWorkerId >= numWorkers) {
+                return NIXL_ERR_NOT_SUPPORTED;
+            }
```
w1ldptr: So if the user requested, for example, STRICT mode, no progress thread and didn't specify any 'worker mode', then their ucxEngine instance will stop working as soon as it had been used from more than num_workers threads? Why?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: You can make the atomic mutable though.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Right, the current code was thought out before I see the recent patch that use NIXL mode to determine the ucx_worker mode. The original thought is, by default the worker_mode will be UCX_MULTI and we use a round robin to assign thread to worker, and user does not have to worry about anything. And if the user want to squeeze more performance, they can specify the UCX_SINGLE mode explicitly and use the same number of thread and ucx_worker for the task.

With the NIXL sync mode in the picture now, we do have to rethink what we want to do. When NIXL sync mode is STRICT and no prog thread, we can just give it a worker. But in the case when NIXL sync mode is NONE, we have to do something to ensure no 2 threads using the worker or fail here.

src/plugins/ucx/ucx_backend.cpp
Pyngon: From what I know, thread_local has to be static, which will only has 1 copy per class. But we need to have 1 copy per instance.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
tstamler: in general, I think its fine to not put a lot more effort into UCX_MO if UCX is going to natively support multiple GPUs well

src/plugins/ucx/ucx_backend.cpp
w1ldptr: > And if the user want to squeeze more performance, they can specify the UCX_SINGLE mode explicitly and use the same number of thread and ucx_worker for the task.

I don't think that would be a valid configuration from UCX PoV because in such mode only the thread that instantiated the backend can access any of the workers:
`UCS_THREAD_MODE_SINGLE,     /**< Only the master thread can access (i.e. the thread that initialized the context; multiple threads may exist and never access) */`

This kinda proves my point from another comment thread about configuration becomes quite complicated with this option.

> With the NIXL sync mode in the picture now, we do have to rethink what we want to do. When NIXL sync mode is STRICT and no prog thread, we can just give it a worker. But in the case when NIXL sync mode is NONE, we have to do something to ensure no 2 threads using the worker or fail here

I think we can always just give it a worker. It is up to the user to ensure synchronized access in NONE mode.

src/plugins/ucx/ucx_backend.cpp
Pyngon: In the latest commit, added the destructor and allow reusing of worker when the previous thread exit.

src/plugins/ucx/ucx_backend.cpp
Pyngon: In the latest commit, put the variable in the engine itself and make it mutable.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Remove this param in the latest commit, do not allow user to specify the worker mode directly. The worker mode will be derived from the progThread and nixl sync mode param.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: param removed.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Removed this param in the latest commit.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Worker allocation logic has been reworked in the latest commit. Let's restart the review from there.

src/plugins/ucx/ucx_backend.cpp
Pyngon: In the latest commit, we will create some number of shared workers, if no dedicated worker is available, we will assigned the thread a shared worker.

Pyngon:
Changes on the latest commit
1. Divide ucp_worker to 2 types, dedicated workers and shared workers. Dedicated workers will be associated with 1 thread only at any given time, and shared workers could be used by multiple thread. Dedicated workers will be associated with a thread when a thread createXferReq for the first time. In the case where no dedicated workers is available, the thread will use a shared worker. Users could specify number of dedicated worker with "num_workers" param, and specify number of shared workers with "num_shared_workers" param.

2. Allow reusing of dedicated worker when a thread associated with it has exited.

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -64,6 +64,7 @@ DEFINE_int32(num_files, 1, "Number of files used by benchmark");
 DEFINE_int32(num_initiator_dev, 1, "Number of device in initiator process");
 DEFINE_int32(num_target_dev, 1, "Number of device in target process");
 DEFINE_bool(enable_pt, false, "Enable Progress Thread (only used with nixl worker)");
+
```
iyastreb: redundant

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -547,6 +548,15 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     if (num_workers_iter == custom_params->end() || !absl::SimpleAtoi(num_workers_iter->second, &numWorkers))
         numWorkers = 1;
 
+    if (custom_params->count("num_shared_workers") == 0 ||
```
iyastreb: why not to do it as above with iterator, to have a single map lookup

src/utils/ucx/ucx_utils.h
```diff
@@ -143,6 +145,15 @@ class nixlUcxContext {
     int memReg(void *addr, size_t size, nixlUcxMem &mem);
     std::unique_ptr<char []> packRkey(nixlUcxMem &mem, size_t &size);
     void memDereg(nixlUcxMem &mem);
+    size_t getNumAssociatedWorkers() {
```
iyastreb: can be const?

src/utils/ucx/ucx_utils.h
```diff
@@ -175,13 +189,32 @@ class nixlUcxWorker {
...
+    void disassociate() {
+        associated.store(false);
+    }
+    bool isAssociated() const {
+        return associated.load();
+    }
+    size_t getWorkerId() const {
+        return workerId;
+    }
+    std::shared_ptr<nixlUcxContext> getCtx() const {
```
iyastreb: I don't think there are cases when we need to construct a copy of shared_ptr when calling this method. We rather need to access ctx owned by a worker, right?
If so better to return `nixlUcxContext &`, or `const shared_ptr &`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1287,14 +1317,23 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
-    ret = notifSendPriv(remote_agent, msg, req, wid);
+    nixlUcxWorker *worker = getSharedWorker();
+    if (worker == nullptr) {
+        // User may specify 0 shared workers, we will try to get a dedicated
+        // worker in this case
+        worker = getDedicatedWorker();
+        if (worker == nullptr) {
+            return NIXL_ERR_INVALID_PARAM;
+        }
+    }
```
brminich: maybe hide it in some member function. You can create a common function which would either try to get shared or dedicated worker at first depending on some param

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -128,6 +128,10 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
...
+            backend_params["num_workers"] = std::to_string(xferBenchConfig::num_threads);
+            backend_params["num_shared_workers"] = std::to_string(2);
```
brminich: is there a reson to have 2? Otherwise maybe do not specify it and it will be set to 1 by default

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1303,3 +1342,85 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
     return NIXL_SUCCESS;
 }
+
+/****************************************
+ * Thread to worker mapping
+*****************************************/
+
+void nixlUcxEngine::threadMapDestructor(void *arg)
+{
+    nixlUcxWorker *worker = (nixlUcxWorker *)arg;
```
brminich: maybe `nixlUcxWorker *worker = static_cast<nixlUcxWorker *>(arg);`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1303,3 +1342,85 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+                return uws[cur].get();
+            }
+        }
+    }
+    return nullptr;
+}
+
+nixlUcxWorker *nixlUcxEngine::getDedicatedWorker() const
+{
+    nixlUcxWorker *p = (nixlUcxWorker *)pthread_getspecific(keyThreadToWorker);
```
brminich: `nixlUcxWorker *p = static_cast<nixlUcxWorker *>(pthread_getspecific(keyThreadToWorker));`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1303,3 +1342,85 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+        /*
+         * Do not have to search the worker list when all
+         * workers are already associated.
+         */
+        return nullptr;
+    }
+
+    nixlUcxWorker *worker = findAndAssociateDedicatedWorker();
+    if (worker != nullptr) {
+        pthread_setspecific(keyThreadToWorker, worker);
```
brminich: i'd check return status and print an error in case of failure (also for other pthread functions)

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1303,3 +1342,85 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+                 */
+                nextWorkerId = cur + 1;
+                return uws[cur].get();
+            }
+        }
+    }
+    return nullptr;
+}
+
+nixlUcxWorker *nixlUcxEngine::getDedicatedWorker() const
```
brminich: is it really a const?

aranadive:
@Pyngon any update?

Pyngon:
> @Pyngon any update?

On a vacation last week, will look at this today.

src/plugins/ucx/ucx_backend.cpp
Pyngon: This function has to be called from prepXfer(), so has to be const.
And yes, this function may change the value of a atomic variable which marked as mutable.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Made the changes to have single map lookup.

src/utils/ucx/ucx_utils.h
Pyngon: Right, `const shared_ptr &` is better.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: Ya, nixlbench won't need this. Removed in latest patch.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Done

src/plugins/ucx/ucx_backend.cpp
Pyngon: Agree. This looks like an API used by every plugin, we probably better to have a separate patch to focus on changing this if needed.

src/plugins/ucx/ucx_backend.cpp
Pyngon: This also looks like a pattern that used by the NIXL framework now. Will be good to have a separate patch to change this if needed.

Pyngon:
@aranadive @iyastreb @brminich  Done going through the review and made the changes. Please help to have another look. Thanks!

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -573,12 +584,20 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+                                          init_params->syncMode);
 
     for (unsigned int i = 0; i < numWorkers; i++)
-        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc, i));
 
-    const auto &uw = uws.front();
+    for (unsigned int i = 0; i < numSharedWorkers; i++) {
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc, i + numWorkers, true));
+    }
```
brminich: is there a benefit in having 2 workers (1 shared + 1 dedicated) by default?
maybe it is simpler to have just 1 shared worker by default?

Pyngon: Yes, it is fine to have 1 shared worker by default

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1281,125 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+                return uws[cur].get();
+            }
+        }
+    }
+    return nullptr;
+}
+
+nixlUcxWorker *nixlUcxEngine::getDedicatedWorker() const
+{
+    nixlUcxWorker *p = static_cast<nixlUcxWorker *>(pthread_getspecific(keyThreadToWorker));
```
iyastreb:
```suggestion
    nixlUcxWorker *worker = static_cast<nixlUcxWorker *>(pthread_getspecific(keyThreadToWorker));
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1281,125 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+    if (p != nullptr) {
+        return p;
+    }
+
+    if (uc->getNumAssociatedWorkers() == numDedicatedWorker) {
+        /*
+         * Do not have to search the worker list when all
+         * workers are already associated.
+         */
+        return nullptr;
```
iyastreb: maybe we should print at least debug message?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1281,125 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+    // Get the other worker type when we could not get the prefer one.
+    if (!prefer_shared) {
+        // Use shared worker when we used up the dedicated worker.
+        return getSharedWorker();
+    }
+
+    // User may specify 0 shared workers, we will try to get a dedicated
+    // worker in this case
+    return getDedicatedWorker();
+}
```
iyastreb:
```suggestion
nixlUcxWorker *nixlUcxEngine::getWorkerWithPreference(bool prefer_shared) const
{
nixlUcxWorker *nixlUcxEngine::getWorkerWithPreference(bool prefer_shared) const
{
    nixlUcxWorker *worker = prefer_shared ? getSharedWorker() : getDedicatedWorker();
    if (worker != nullptr) {
        return worker;
    }
    // If the preferred worker type is not available, get the other type
    return prefer_shared ? getDedicatedWorker() : getSharedWorker();
}```

File: src/plugins/ucx/ucx_backend.h [hunk 31]
```diff
@@ -133,6 +134,16 @@ class nixlUcxEngine
         std::unordered_map<std::string, ucx_connection_ptr_t,
                            std::hash<std::string>, strEqual> remoteConnMap;
 
+        // Thread to worker mapping
+        pthread_key_t keyThreadToWorker;
+        mutable std::atomic<size_t> nextWorkerId;
+        nixl_status_t initThreadMapping();
```
iyastreb:
```suggestion

        nixl_status_t initThreadMapping();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1281,125 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+         * workers are already associated.
+         */
+        return nullptr;
+    }
+
+    nixlUcxWorker *worker = findAndAssociateDedicatedWorker();
+    if (worker != nullptr) {
+        int err = pthread_setspecific(keyThreadToWorker, worker);
+        if (err) {
+            NIXL_ERROR << "Failed to set keyThreadToWorker, workeriId:"
```
iyastreb:
```suggestion
            NIXL_ERROR << "Failed to set keyThreadToWorker, workerId:"
```

src/utils/ucx/ucx_utils.h
```diff
@@ -148,19 +148,30 @@ class nixlUcxContext {
     ucp_context_h ctx;
     nixl_ucx_mt_t mt_type;
     ucp_err_handling_mode_t err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
+    mutable std::atomic<size_t> num_associated_workers = 0;
```
iyastreb: IMO nixlUcxEngine is the right place for this variable, to keep the logic in one place. The relationship between `nixlUcxContext` and `nixlUcxEngine` is 1:1, so it should not be an issue

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1281,125 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+{
+    int err = pthread_key_delete(keyThreadToWorker);
+    if (err) {
+        NIXL_WARN << "Failed to delete keyThreadToWorker, err:" << err;
+        return NIXL_ERR_INVALID_PARAM;
+    }
+    return NIXL_SUCCESS;
+}
+
+nixlUcxWorker *nixlUcxEngine::findAndAssociateDedicatedWorker() const
```
iyastreb: TBH the overall approach looks a bit complicated.
Can we reach the same functionality by just keeping a list/set of free workers + single mutex?
I'm not sure that for this simple task we need so much atomic counters
Why not to do it with KISS principle:
```
std::unordered_set<nixlUcxWorker *> freeWorkers;
std::mutex workersMutex;

nixlUcxWorker * associate()
{
   // lock mutex
   // remove elem from freeWorkers
   // add it to thread context
}
void disassociate(nixlUcxWorker * worker)
{
    // lock mutex
    // remove worker from thread context
    // add back to the set
}
```
And no changes in other classes

Pyngon: Yes, this is simpler, I will change it to this way.

src/utils/ucx/ucx_utils.h
Pyngon: Putting this in nixlUcxContext is because of threadMapDestructor (destructor function when a thread exits). This is a static function with no access to the nixlUcxEngine object, it only has a nixlUcxWorker object. As nixlUcxWorker currenly only has a pointer to nixlUcxContext, so I put it here.

If we put it into nixlUcxEngine, i think we will need to add a pointer to the engine in nixlUcxWorker, and as nixlUcxWorker is shared among UCX and UCX_MO (which have different engine), we will keep a void pointer in nixlUcxWorker, and cast the void pointer back to the engine in the threadMapDestructor function. What is your though on this pattern?

src/plugins/ucx/ucx_backend.cpp
Pyngon: made the changes, please help to review again. Thanks.

src/utils/ucx/ucx_utils.h
Pyngon: The latest patch keep the engine as a void pointer in nixlUcxEngine, let me know if you have a different thought. Thanks.

src/plugins/ucx/ucx_backend.h
Pyngon: Done

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1286,92 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+nixlUcxWorker *nixlUcxEngine::getFreeDedicatedWorker() const
+{
+    if (freeWorkers.size() == 0) {
+        return nullptr;
+    }
+
+    const std::lock_guard<std::mutex> lock(workersMutex);
+    nixlUcxWorker *worker = freeWorkers.front();
+    freeWorkers.pop();
+    return worker;
```
iyastreb: Better use empty().
Do we really need this micro-optimization? Maybe:

```suggestion
    const std::lock_guard<std::mutex> lock(workersMutex);
    if (freeWorkers.empty()) {
        return nullptr;
    }

    nixlUcxWorker *worker = freeWorkers.front();
    freeWorkers.pop();
    return worker;
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1244,3 +1286,92 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+        return worker;
+    }
+
+    worker = getFreeDedicatedWorker();
+    if (worker != nullptr) {
+        int err = pthread_setspecific(keyThreadToWorker, worker);
+        if (err) {
+            NIXL_ERROR << "Failed to set keyThreadToWorker, workerId:"
+                << worker->getWorkerId() << "err:" << err;
+            return nullptr;
```
iyastreb: should we push back worker to free pool?

iyastreb (commented):
LGTM besides minor comments

src/plugins/ucx/ucx_backend.cpp
Pyngon: Thanks for catching this. Yes, should check empty after lock

src/plugins/ucx/ucx_backend.cpp
Pyngon: yes, absolutely.

iyastreb (dismissed):

Pyngon:
Hi @yosefe @gleon99 @brminich, could you help to see if this patch is ready to merge? And give me an approval if it is. Thanks!

leibin2014:
@yosefe Can you pls help on the +2 code review? Thanks!

src/plugins/ucx/ucx_backend.h
```diff
@@ -251,12 +272,16 @@ class nixlUcxEngine
         nixl_status_t checkConn(const std::string &remote_agent);
         nixl_status_t endConn(const std::string &remote_agent);
 
-        const std::unique_ptr<nixlUcxWorker> &getWorker(size_t worker_id) const {
-            return uws[worker_id];
+        // Add worker back to the free workers queue
+        void
+        pushFreeWorker(nixlUcxWorker *worker) const {
```
rakhmets: Why do we need this in `public` section?

src/utils/ucx/ucx_utils.h
```diff
@@ -171,13 +175,18 @@ class nixlUcxWorker {
...
 
-    [[nodiscard]] static ucp_worker* createUcpWorker(nixlUcxContext&);
-
-  public:
-    explicit nixlUcxWorker(const std::shared_ptr<nixlUcxContext> &_ctx);
+    [[nodiscard]] static ucp_worker *
+    createUcpWorker(nixlUcxContext &, bool);
 
+public:
+    explicit nixlUcxWorker(const std::shared_ptr<nixlUcxContext> &_ctx,
```
rakhmets: `explicit` has become unnecessary.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1242,3 +1285,93 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+
+nixl_status_t
+nixlUcxEngine::destroyThreadMapping() {
+    int err = pthread_key_delete(keyThreadToWorker);
+    if (err) {
+        NIXL_WARN << "Failed to delete keyThreadToWorker, err:" << err;
+        return NIXL_ERR_INVALID_PARAM;
+    }
+    return NIXL_SUCCESS;
+}
```
rakhmets: The return type should be void.
The method is called from a destructor and the return type cannot be handled. Printing a warning message is enough.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -568,16 +582,30 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+
+    numDedicatedWorker = numWorkers;
 
-    const auto &uw = uws.front();
+    // Initialize freeWorkers queue with dedicated workers
+    for (unsigned int i = 0; i < numWorkers; i++) {
+        freeWorkers.push(uws[i].get());
+    }
+
+    const auto &uw = uws.back();
```
rakhmets: Why do we need to change from `uws.front()` to `uws.back()`.

P.S. I'm not sure I understand how it works. We save worker address of only one worker, and use it to connect endpoints...

src/utils/ucx/ucx_utils.h
```diff
@@ -171,13 +175,18 @@ class nixlUcxWorker {
 private:
     /* Local UCX stuff */
     const std::shared_ptr<nixlUcxContext> ctx;
+    size_t workerId;
```
rakhmets: Looks like an external data that doesn't belong to the class.

It seems the problem is that data only related to `nixlUcxEngine` is stored in `nixlUcxWorker`.
The ids should be managed by `nixlUcxEngine`.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1242,3 +1285,93 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
     return NIXL_SUCCESS;
 }
+
+/****************************************
+ * Thread to worker mapping
+ *****************************************/
+
+void
+nixlUcxEngine::threadMapDestructor(void *arg) {
+    nixlUcxWorker *worker = static_cast<nixlUcxWorker *>(arg);
```
rakhmets: There is no need to duplicate a type name twice. Please use `auto`.
```suggestion
    auto worker = static_cast<nixlUcxWorker *>(arg);
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1242,3 +1285,93 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+
+    /* When a thread exits, disassociate it from the worker,
+     * so that the worker could be reused by another thread.
+     */
+    nixlUcxEngine *engine = static_cast<nixlUcxEngine *>(worker->getEngine());
```
rakhmets: The same as above.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -568,16 +582,30 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+                                          err_handling_mode,
+                                          numWorkers + numSharedWorkers,
+                                          init_params->syncMode);
 
     for (unsigned int i = 0; i < numWorkers; i++)
-        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc));
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc, i, false, this));
+
+    for (unsigned int i = 0; i < numSharedWorkers; i++) {
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(uc, i + numWorkers, true, this));
```
rakhmets: You can store worker ids in `freeWorkers` queue to get rid of storing external information in `nixlUcxWorker`.
And `get*Worker` functions will return worker id in this case.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -557,8 +558,21 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
         devs = str_split((*custom_params)["device_list"], ", ");
 
     const auto num_workers_iter = custom_params->find("num_workers");
-    if (num_workers_iter == custom_params->end() || !absl::SimpleAtoi(num_workers_iter->second, &numWorkers))
-        numWorkers = 1;
+    if (num_workers_iter == custom_params->end() ||
+        !absl::SimpleAtoi(num_workers_iter->second, &numWorkers)) {
+        // Default to have 1 shared worker and 0 dedicated worker
```
rakhmets, line 565: According to the default values `ucxBackendEngine` will create 2 workers, if `params["num_workers"]` is set to 1. Which may be unexpected for the client code. It looks like we need to keep `num_workers` key for shared workers, and add `num_dedicated_workers` option instead.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -527,7 +527,8 @@ void nixlUcxEngine::progressThreadRestart()
 
 nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
 : nixlBackendEngine (init_params) {
-    unsigned long numWorkers;
+    unsigned int numWorkers;
```
rakhmets: Can be removed, and `numDedicatedWorker` can be used instead.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1226,14 +1265,18 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
 {
     nixl_status_t ret;
     nixlUcxReq req;
-    size_t wid = getWorkerId();
 
-    ret = notifSendPriv(remote_agent, msg, req, wid);
+    nixlUcxWorker *worker = getWorkerWithPreference(true);
+    if (worker == nullptr) {
+        return NIXL_ERR_INVALID_PARAM;
```
rakhmets: The error is not related to the passed parameters. `NIXL_ERR_BACKEND`?

src/utils/ucx/ucx_utils.h
```diff
@@ -171,13 +175,18 @@ class nixlUcxWorker {
...
+    void *engine;
```
rakhmets: This field does not look like it belongs to this class. It's only here to return the worker to the queue of free dedicated workers. It should be handled on backend level. E.g. Associate a pthread key with a structure containing pointer to engine and worker id.

src/plugins/ucx/ucx_backend.cpp
Pyngon: What is the default value you mentioned here? From what I know, default value is from the "get_ucx_backend_common_options()" function?

src/plugins/ucx/ucx_backend.cpp
Pyngon: From what I understand so far, the worker address is sent to the remote peer to establish connection. So, all workers in the remote peer will only establish connection with the first worker in local.

src/plugins/ucx/ucx_backend.h
Pyngon: Right, can be moved to private

src/plugins/ucx/ucx_backend.cpp
Pyngon: By the way, in the latest patch I put the shared workers in the front for the notification issue, so changed back to front in the latest patch.

src/utils/ucx/ucx_utils.h
Pyngon: Done in the latest patch

src/plugins/ucx/ucx_backend.cpp
Pyngon: Done

src/plugins/ucx/ucx_backend.cpp
Pyngon: Done

src/plugins/ucx/ucx_backend.cpp
Pyngon: I changed to keep the numSharedWorkers in the class instead, so removed the numSharedWorkers local variable and keep this here.

src/utils/ucx/ucx_utils.h
Pyngon: Done.

src/utils/ucx/ucx_utils.h
Pyngon: I've removed the worker_id and engine arguments, so keep the explicit.

src/plugins/ucx/ucx_backend.cpp
rakhmets: I meant the following:
```C++
nixlBackendInitParams init_params;
init_params.custom_params = {{"num_workers", "1"}};
nixlUcxEngine engine{&init_params}; // creates two nixlUcxWorker
```

Pyngon: ok, i think I understand what you mean now. The name "num_workers" seems to implied total workers, but we are creating more worker than that. I made the change to use "num_dedicated_workers" for dedicate workers and use "num_workers" to indicate shared workers. Please help to have a look again. Thanks

Pyngon: @rakhmets above

iyastreb:
https://github.com/ai-dynamo/nixl/pull/573 should solve this issue:
- added threadpool to execute large post batches in parallel
- split_batch_size option to control the threshold
- round-robin distribution (and attach to thread in TLS) of shared workers (for all cases)

PR #346, opened 2025-05-21
CI: Increase timeout for AWS job status check

Alexey-Rivkin:
## What?
Increase timeout and polling interval for AWS EFA tests.

## Why?
Improve test robustness by allowing more time for concurrent AWS jobs to complete.
Reduce API call frequency and log spamming by interval spacing.

## How?
Simple parameter adjustment in the wait_for_status function:
- Timeout increased: 600s → 1800s (10min → 30min)
- Check interval increased: 60s → 180s (1min → 3min)

### DIFF
--- a/contrib/aws-efa/aws_test.sh
+++ b/contrib/aws-efa/aws_test.sh
@@ -83,8 +83,8 @@ JOB_ID=$(aws batch submit-job \
 # Function to wait for a specific job status
 wait_for_status() {
     local target_status="$1"
-    local timeout=1800
-    local interval=180
+    local timeout="$2"
+    local interval="$3"
     local status=""
     SECONDS=0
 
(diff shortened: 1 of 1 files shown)


brminich (approved):

contrib/aws-efa/aws_test.sh
```diff
@@ -83,8 +83,8 @@ JOB_ID=$(aws batch submit-job \
 # Function to wait for a specific job status
 wait_for_status() {
     local target_status="$1"
-    local timeout=600
-    local interval=60
+    local timeout=1800
+    local interval=180
```
dpressle: This is not ideal, we need to distinguished between different stages of this process and have different timeouts for each.
For example, we should probably wait more time to allocate the resources (RUNNING) but should not wait so much time for the test run it self to end (SUCCEDED|FAILED)
I would add an option to pass timeout value to this function and have different timeouts by type of action.

Alexey-Rivkin: [CI: Use stage-specific timeouts in wait_for_status](https://github.com/ai-dynamo/nixl/pull/346/commits/5f25e2b8f26aa96d4b58cf00d99b869132b13edb)

yosefe (approved):

PR #347, opened 2025-05-21
Timeout from poll() in progress thread after pthrDelay

w1ldptr:
Some UCX transports may still require periodic polling in specific corner case conditions. Re-use 'pthrDelay' config that is a leftover from previous busy-polling implementation to set a timeout for poll() instead of sleeping indefinitely.


w1ldptr:
@evgeny-leksikov please review

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -385,11 +384,15 @@ void nixlUcxEngine::progressFunc()
             if (made_progress && !wid)
                 notifProgress();
         }
+        timeout = false;
 
-        while (poll(pollFds.data(), pollFds.size(), -1) <= 0)
+        int ret;
+        while ((ret = poll(pollFds.data(), pollFds.size(), pthrDelay.count())) < 0)
             NIXL_ERROR << "Call to poll() was interrupted, retrying. Error: " << strerror(errno);
```
evgeny-leksikov: what kind of errors happens here? do we really need errors in log for any type of error? maybe we could silently handle EAGAIN, EINTR, etc or convert to NIXL_DEBUG? How about something like this:
```suggestion
        while (true) {
            int ret = poll(pollFds.data(), pollFds.size(), pthrDelay.count());
            if (ret < 0) {
                if (errno != EAGAIN) && (errno != EINTR)) {
                    NIXL_ERROR << "poll() failed with error: " << strerror(errno) << ", retrying.";
                }
            } else {
                timeout = (ret > 0);
                break;
            }
        }
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -463,6 +466,12 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
             this->initErr = true;
             return;
         }
+
+        // This will ensure that the resulting delay is at least 1ms and fits into int in order for
+        // it to be compatible with poll()
+        pthrDelay = std::chrono::ceil<std::chrono::milliseconds>(
```
brminich, line 536: probably worth changing description of this option in `nixlAgentConfig`, because it has different meaning now

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -360,14 +361,12 @@ void nixlUcxEngine::progressFunc()
...
-    for (size_t wid = 0; wid < pollFds.size() - 1; wid++)
-        pollFds[wid].revents = POLLIN;
-
+    // Set timeout event so that the main loop would progress all workers on first iteration
+    bool timeout = true;
     bool pthrStop = false;
     while (!pthrStop) {
         for (size_t wid = 0; wid < pollFds.size() - 1; wid++) {
-            if (!(pollFds[wid].revents & POLLIN))
+            if (!(pollFds[wid].revents & POLLIN) && !timeout)
```
evgeny-leksikov, line 433: it seems like logic around `!timeout` condition can be simplified, like moving the condition outside of `for(...)` loop

src/plugins/ucx/ucx_backend.cpp
w1ldptr: I don't have an opinion either way I remember many people had some strong opinions about these logs and asserts (later mostly removed in the accepted version). @yosefe @brminich ?

Also, timeout should be 'true' when ret==0.

src/plugins/ucx/ucx_backend.cpp
w1ldptr: Not sure I got you. How would moving this conditional outside of for loop look like?

src/plugins/ucx/ucx_backend.cpp
w1ldptr: ack

src/plugins/ucx/ucx_backend.cpp
brminich: regarding logs i agree with @evgeny-leksikov, but afair it was @yosefe comment to log it, but my impression it was debug or trace before..
maybe it is indeed makes sence not to log for EAGAIN and EINTR to avoid

w1ldptr: Hmm, I can see why we would want to avoid spamming this on "error" level, but being able to determine by the log that progress thread is stuck in infinite EAGAIN loop also has some value IMO.

brminich: can we log it then with DEBUG or TRACE rather than ERRROR?

w1ldptr: Sure, np.

src/api/cpp/nixl_params.h
```diff
@@ -41,11 +41,9 @@ class nixlAgentConfig {
     public:
 
         /**
-         * @var Progress thread frequency knob (in us)
-         *      The progress thread is calling sched_yield to avoid blocking a core
-         *      If pthrDelay time is less than sched_yield time - option has no effect
-         *      Otherwise pthread will be calling sched_yield until the specified
-         *      amount of time has past.
+         * @var Progress thread sleep timeout Progress thread event waiting timeout.
```
brminich: Did you mean?
```suggestion
         * @var Progress thread event waiting timeout.
```

brminich (approved):

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: sorry, I missed that the `!timeout` is a 2nd condition in the statement

evgeny-leksikov (approved):

PR #353, opened 2025-05-22
Vvenkates27/reg dereg file ops

vvenkates27:


src/plugins/cuda_gds/gds_backend.h
```diff
@@ -87,13 +87,14 @@ class nixlGdsBackendReqH : public nixlBackendReqH {
...
         gdsUtil *gds_utils;
-        std::unordered_map<int, gdsFileHandle> gds_file_map;
+        std::unordered_map<uint64_t, gdsFileHandle> gds_file_map;
 
         mutable std::mutex batch_pool_lock;
         mutable std::list<nixlGdsIOBatch*> batch_pool;
         unsigned int batch_pool_size;  // Renamed from pool_size
         unsigned int batch_limit;      // Added for configurable batch limit
         unsigned int max_request_size; // Added for configurable request size
+        bool initErr;
```
rakhmets, line 97: Please remove unused class field.

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+        default:
+            return -1;
+    }
+}
+
+nixl_status_t nixlFileUtils::openFile(const std::string& path, nixlFileMode mode, int& fd, bool use_direct) {
+    int flags = getModeFlags(mode, use_direct);
+    if (flags == -1) {
+        NIXL_ERROR << absl::StrFormat("Invalid mode flags for path: %s", path);
```
rakhmets, line 57: Maybe just:
```suggestion
        NIXL_ERROR << "Invalid mode flags for path: " << path;
```

src/utils/file/file_utils.h
```diff
@@ -0,0 +1,138 @@
...
+    OPEN,       // Open a file
+    CLOSE,      // Close a file
+    UNLINK,     // Delete file
+    STAT        // Get file status
+};
+
+/**
+ * @brief File utilities class for basic file operations
+ */
+class nixlFileUtils {
```
rakhmets, line 56: Please prefer RAII. Smth like:
```C++
class NixlFile {
public:
    NixlFile(std::string_view path, nixlFileMode mode, bool use_direct = false);
    NixlFile(std::string_view prefixed_path);
    ~nixlFile();
    ...

pirvate:
    int fd;
};
```
, where `~nixlFile()` closes opened file.

So you can use it like this:
```C++
void foo(std::string_view prefixed_path) {
    NixlFile nixlFile{prefixed_path};
    ...
    if (smth) return; // no leak
    ...
}

File: src/plugins/cuda_gds/gds_backend.cpp [hunk 4]
```diff
@@ -159,7 +202,19 @@ nixl_status_t nixlGdsEngine::deregisterMem (nixlBackendMD* meta)
...
         gds_utils->deregisterFileHandle(md->handle);
-	gds_file_map.erase(md->handle.fd);
+
+        // Find and erase the entry using devId
+        for (auto it = gds_file_map.begin(); it != gds_file_map.end(); ++it) {
+            if (it->second.fd == md->handle.fd) {
+                gds_file_map.erase(it);
+                break;
+            }
+        }
```
rakhmets, line 212: It seems that you can use `md->handle.fd` to find the entity.
```suggestion
        const auto& it = gds_file_map.find(md->handle.fd);
        if (it != gds_file_map.end()) {
            gds_file_map.erase(it);
        }
```

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -87,6 +92,54 @@ nixlGdsEngine::nixlGdsEngine(const nixlBackendInitParams* init_params)
...
+    }
+
+    // Register new file descriptor
+    status = gds_utils->registerFileHandle(mem.devId, mem.len,
+                                       mem.metaInfo, md->handle);
+    if (status == NIXL_SUCCESS) {
+        gds_file_map[mem.devId] = md->handle;
+        out = (nixlBackendMD*)md;
+    }
+    return status;
```
rakhmets, line 140: Please avoid code duplication.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -87,6 +92,54 @@ nixlGdsEngine::nixlGdsEngine(const nixlBackendInitParams* init_params)
...
+            nixlFileUtils::closeFile(fd);
+            return status;
+        } else if (status != NIXL_ERR_INVALID_PARAM) {
+            // Error during file open (not invalid prefix case)
+            return status;
+        }
+    }
+
+    // Handle raw file descriptor (no prefix or invalid prefix)
+    auto it = gds_file_map.find(mem.devId);
```
rakhmets, line 124: Why don't we check key for presence at line 106?

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,247 @@
...
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    // For WRITE_ONLY, check if file exists and add O_CREAT if it doesn't
+    if (mode == nixlFileMode::WRITE_ONLY) {
+        nixl_status_t status = fileExists(path);
```
da-x, line 63: With this check, the code is prone to the [Time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue. One process can get `true` on the check, followed by `rename` or `unlink` by another process, then the first process will fail creating the file.

I think it is better to be consistent with `READ_WRITE` mode, always provide `O_CREAT` for write, and remove this check.

vvenkates27:
No longer considered for file ops. closing this PR now.

PR #354, opened 2025-05-22
NIXL KV Cache Benchmark

aranadive:
This is a first version of the NIXL KV Cache Storage Benchmark.

The overall goals of the benchmark are as follows:
* Capture KV Cache creation and usage scenarios using synthetic data
* Drive optimizations in NIXL Storage Plugins
* Improve Dynamo and Inference Runtime Integrations

Key Features / Components
* Synthetic Configuration generator
* Workload Traffic Generator
* NIXLBench to run workload
* Part of NIXL - buildable as separate benchmark


.pre-commit-config.yaml
```diff
@@ -23,6 +23,7 @@ repos:
     rev: 5.12.0
     hooks:
       - id: isort
+        args: ["--profile", "black", "--filter-files"]
```
vvenkates27, line 26: What are these changes?

aranadive: We require them for python formatting (import and general code). Both isort and black CI tools can conflict so isort has a profile to behave like black. Otherwise I was getting conflicting python formatting updates for the imports list.

ovidiusm:
The WORKDIR from nixlbench should not have been changed for kvbench. We should be able to run the two independently https://github.com/ai-dynamo/nixl/issues/392

PR #356, opened 2025-05-22
LISTENER: Segment socket-based message transfers

tvegas1:
Ability to send or receive a whole chunk of data on a socket is not always guaranteed, leading to truncation. So we now also send a size to provide segmentation. Added test, but it only reproduces the issue when forcing send() one byte at a time.


test/gtest/test_transfer.cpp
```diff
@@ -363,6 +363,16 @@ TEST_P(TestTransfer, remoteMDFromSocket)
...
 }
 
+TEST_P(TestTransfer, ListenerCommSize) {
+    std::vector<MemBuffer> buffers;
+    createRegisteredMem(getAgent(1), 64, 10000, DRAM_SEG, buffers);
+    auto status = fetchRemoteMD(0, 1);
+    ASSERT_EQ(NIXL_SUCCESS, status);
+    do {
+        status = checkRemoteMD(0, 1);
+    } while (status != NIXL_SUCCESS);
```
w1ldptr: Perhaps this loops needs some limit on number of iterations? Otherwise the test will just get stuck in infinite loop on any breakage that causes checkRemoveMD to return an error.

tvegas1: makes sense fixed

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+                if (!force && done == 0) {
+                    return false; // nothing to read yet
+                }
 
-    return recv_bytes;
+                continue;
+            }
+        }
+
+        throw std::runtime_error(
```
tvegas1, line 156: this is clang-format output

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+    size_t offset = 0, done = 0;
+    uint64_t size = msg.size();
+    struct iovec iov[2];
+
+    iov[0].iov_base = &size;
+    iov[0].iov_len = sizeof(size);
+    iov[1].iov_base = &msg[0];
+    iov[1].iov_len = msg.size();
+
+    for (int i = 0; i < 2;) {
```
```suggestion
    for (size_t i =0, offset = 0, sent = 0; i < iov_size;) {
```

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+        if (bytes < 0) {
+            if (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK) {
+                continue;
+            }
+
+            throw std::runtime_error(
+                    absl::StrFormat("sendCommMessage(fd=%d) %zu/%zu bytes failed, errno=%d",
+                                    fd,
+                                    done,
+                                    size + msg.size(),
```
```suggestion
                                    size + sizeof(size),
```

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+sendCommMessage(int fd, std::string msg) {
+    ssize_t bytes;
+    size_t offset = 0, done = 0;
+    uint64_t size = msg.size();
+    struct iovec iov[2];
+
+    iov[0].iov_base = &size;
+    iov[0].iov_len = sizeof(size);
+    iov[1].iov_base = &msg[0];
+    iov[1].iov_len = msg.size();
```
```suggestion
    constexpr size_t iov_size = 2;
    const iovec iov[iov_size] {{&size, sizeof(size)}, {msg.data(), msg.size()}};
```

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
-size_t sendCommMessage(int fd, std::string msg){
-    size_t bytes;
-    bytes = send(fd, msg.c_str(), msg.size(), 0);
-    if(bytes < 0) {
-        NIXL_ERROR << "Cannot send on socket to fd " << fd;
+void
+sendCommMessage(int fd, std::string msg) {
+    ssize_t bytes;
+    size_t offset = 0, done = 0;
+    uint64_t size = msg.size();
```
```suggestion
    size_t size = msg.size();
```
`size_t` to align with receive method.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+    struct iovec iov[2];
```
rakhmets: Not a recommendation, but an idea to use `struct iov { const void *base; size_t size; };`, so you can use `const size_t` for `size` variable, and `string_view` for `msg` argument.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
-
-    do {
-        one_recv_bytes = recv(fd, buffer, sizeof(buffer), 0);
-        if (one_recv_bytes == -1){
-            if(errno == EAGAIN || errno == EWOULDBLOCK) return recv_bytes;
-            NIXL_ERROR << "Cannot recv on socket fd " << fd;
-            return one_recv_bytes;
+bool
+recvCommMessageType(int fd, void *data, size_t size, bool force = false) {
+    for (size_t done = 0; done < size;) {
```
rakhmets: maybe received, or total_received, instead of done.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
 
-size_t sendCommMessage(int fd, std::string msg){
-    size_t bytes;
-    bytes = send(fd, msg.c_str(), msg.size(), 0);
-    if(bytes < 0) {
-        NIXL_ERROR << "Cannot send on socket to fd " << fd;
+void
+sendCommMessage(int fd, std::string msg) {
+    ssize_t bytes;
+    size_t offset = 0, done = 0;
```
rakhmets: maybe sent, or total_sent instead of done.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,90 @@ int connectToIP(std::string ip_addr, int port) {
...
+bool
+recvCommMessage(int fd, std::string &msg) {
+    size_t size;
+    if (!recvCommMessageType(fd, &size, sizeof(size))) {
+        return false;
+    }
+
+    msg.resize(size);
+    recvCommMessageType(fd, &msg[0], size, true);
+    return true;
```
```suggestion
    return recvCommMessageType(fd, msg.data(), size, true);
```

tstamler (commented):
thanks for the cleanup, looks good to me pending others minor review comments

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: will keep as is

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,84 @@ int connectToIP(std::string ip_addr, int port) {
...
 }
 
-size_t sendCommMessage(int fd, std::string msg){
-    size_t bytes;
-    bytes = send(fd, msg.c_str(), msg.size(), 0);
-    if(bytes < 0) {
-        NIXL_ERROR << "Cannot send on socket to fd " << fd;
+void
+sendCommMessage(int fd, std::string msg) {
+    ssize_t bytes;
```
rakhmets: Please remove, since it's a local variable used only inside the loop. And you can use `auto bytes =...` inside the loop.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,84 @@ int connectToIP(std::string ip_addr, int port) {
...
+    size_t size = msg.size();
+    constexpr size_t iov_size = 2;
+    struct iovec iov[iov_size] = { {&size, sizeof(size)}, {&msg[0], msg.size()} };
```
rakhmets: You can use `msg.data()` here and at line 175 in `recvCommMessage`.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,84 @@ int connectToIP(std::string ip_addr, int port) {
     return ret_fd;
 }
 
-size_t sendCommMessage(int fd, std::string msg){
-    size_t bytes;
-    bytes = send(fd, msg.c_str(), msg.size(), 0);
-    if(bytes < 0) {
-        NIXL_ERROR << "Cannot send on socket to fd " << fd;
+void
+sendCommMessage(int fd, std::string msg) {
```
rakhmets: It seems we can pass by reference here.

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,84 @@ int connectToIP(std::string ip_addr, int port) {
...
-    do {
-        one_recv_bytes = recv(fd, buffer, sizeof(buffer), 0);
-        if (one_recv_bytes == -1){
-            if(errno == EAGAIN || errno == EWOULDBLOCK) return recv_bytes;
-            NIXL_ERROR << "Cannot recv on socket fd " << fd;
-            return one_recv_bytes;
+bool
+recvCommMessageType(int fd, void *data, size_t size, bool force = false) {
+    for (size_t sent = 0; sent < size;) {
+        ssize_t bytes = recv(fd, static_cast<char *>(data) + sent, size - sent, 0);
```
rakhmets: auto?

src/core/nixl_listener.cpp
```diff
@@ -27,6 +27,7 @@
 #include <etcd/Watcher.hpp>
 #include <future>
 #endif // HAVE_ETCD
+#include <absl/strings/str_format.h>
```
rakhmets, line 30: Q. Is it required to add this? I'd prefer to use just `std::string` with `<<`,`>>` (or in some cases `std::to_string`).

src/core/nixl_listener.cpp
tvegas1: currently we cannot `sendCommMessage(client_fd, std::string("NIXLCOMM:LOAD" + my_MD))`

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: i am under understanding that abs::StrFromat() is recommended way for the codebase

rakhmets (approved):

brminich (approved):

test/gtest/test_transfer.cpp
```diff
@@ -173,6 +174,16 @@ class TestTransfer : public testing::TestWithParam<std::string> {
         agent.registerMem(reg_list);
     }
 
+    static bool wait_until_true(std::function<bool()> func, int retries = 500) {
+      bool result;
```
iyastreb: indentation 4 spaces

src/core/nixl_listener.cpp
iyastreb: you can pass by const reference, no need to make a copy

src/core/nixl_listener.cpp
```diff
@@ -94,33 +95,83 @@ int connectToIP(std::string ip_addr, int port) {
     return ret_fd;
 }
 
-size_t sendCommMessage(int fd, std::string msg){
-    size_t bytes;
-    bytes = send(fd, msg.c_str(), msg.size(), 0);
-    if(bytes < 0) {
-        NIXL_ERROR << "Cannot send on socket to fd " << fd;
+void
+sendCommMessage(int fd, std::string msg) {
```
iyastreb: can pass by const reference, no need to make a copy

test/gtest/test_transfer.cpp
tvegas1: fixed

src/core/nixl_listener.cpp
tvegas1: caller is building in parameter like below and i'd like not to change in this pr

`sendCommMessage(client_fd, std::string("NIXLCOMM:LOAD" + my_MD));`

iyastreb: it should work with const reference without changing caller side

tvegas1: fixed

src/core/nixl_listener.cpp
```diff
@@ -93,33 +94,86 @@ int connectToIP(std::string ip_addr, int port) {
...
-
-    do {
-        one_recv_bytes = recv(fd, buffer, sizeof(buffer), 0);
-        if (one_recv_bytes == -1){
-            if(errno == EAGAIN || errno == EWOULDBLOCK) return recv_bytes;
-            NIXL_ERROR << "Cannot recv on socket fd " << fd;
-            return one_recv_bytes;
+bool
+recvCommMessageType(int fd, void *data, size_t size, bool force = false) {
+    for (size_t sent = 0; sent < size;) {
```
brminich: why is it called sent? seems received would be more appropriate.

tvegas1: fixed

brminich (approved):

PR #360, opened 2025-05-22
LISTENER: Add connection creation for invalidation message

tvegas1:
Also create a connected socket if needed when invalidating MD on a remote agent, as RPCs are sent unidirectionally for all messages.

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -422,48 +422,34 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
             const auto client = remoteSockets.find(req_sock);
             int client_fd;
 
+            // not connected
+            if(req_command < SOCK_MAX) {
+                if(client == remoteSockets.end()) {
+                    int new_client = connectToIP(req_ip, req_port);
+                    if(new_client == -1) {
+                        NIXL_ERROR << "Listener thread could not connect to IP " << req_ip << " and port " << req_port;
+                        continue;
+                    }
+                    remoteSockets[req_sock] = new_client;
+                    client_fd = new_client;
+                } else {
+                    client_fd = client->second;
+                }
+            }
+
             switch(req_command) {
                 case SOCK_SEND:
                 {
-                    // not connected
-                    if(client == remoteSockets.end()) {
-                        int new_client = connectToIP(req_ip, req_port);
-                        if(new_client == -1) {
-                            NIXL_ERROR << "Listener thread could not connect to IP " << req_ip << " and port " << req_port;
-                            break;
-                        }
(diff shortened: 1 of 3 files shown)


src/core/nixl_listener.cpp
```diff
@@ -422,48 +422,34 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
             const auto client = remoteSockets.find(req_sock);
             int client_fd;
 
+            // not connected
+            if(req_command < SOCK_MAX) {
```
ovidiusm: Nit: could we add whitespace consistently in `if (` etc?

tvegas1: applied recommended clang-format

src/core/nixl_listener.cpp
```diff
@@ -422,51 +422,35 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
...
+                } else {
                     client_fd = client->second;
-                    sendCommMessage(client_fd, std::string("NIXLCOMM:INVL") + name);
-                    break;
                 }
+            }
+
+            switch(req_command) {
+            case SOCK_SEND: {
+                sendCommMessage(client_fd, std::string("NIXLCOMM:LOAD" + my_MD));
```
rakhmets: `std::string` is redundant here (and below).
```suggestion
                sendCommMessage(client_fd, "NIXLCOMM:LOAD" + my_MD);
```

rakhmets (approved):

src/core/nixl_listener.cpp
tvegas1: fixed

test/gtest/metadata_exchange.cpp
```diff
@@ -359,10 +359,16 @@ TEST_F(MetadataExchangeTestFixture, SocketFetchRemoteAndInvalidateLocal)
...
     std::this_thread::sleep_for(sleep_time);
-
     ASSERT_EQ(dst.agent->checkRemoteMD(src.name, {DRAM_SEG}), NIXL_SUCCESS);
+
+    nixl_opt_args_t invalidate_args;
+    invalidate_args.ipAddr = dst.ip;
+    invalidate_args.port = dst.port;
+
+    ASSERT_EQ(src.agent->invalidateLocalMD(&invalidate_args), NIXL_SUCCESS);
+    std::this_thread::sleep_for(sleep_time);
```
brminich, line 370: is it possible to check some state instead of doing a sleep? Why is sleep needed?

tvegas1: it is standard in this file committed recently, probably needs to use while/check remote md loop instead. imo it would make sense to replace all occurences in this file at once in separate pr

PR #361, opened 2025-05-22
fix: memory leak

mkhazraee:
Follow up to #318


aranadive (approved):

src/core/nixl_agent.cpp
```diff
@@ -658,6 +658,7 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     NIXL_SHARED_LOCK_GUARD(data->lock);
     if (data->remoteSections.count(remote_agent) == 0)
     {
+        delete backend_set;
```
tvegas1, line 661: maybe use std::unique_ptr at this point and remove all delete?

mkhazraee: I just carried over the memory leak fix, but yes that's something we should do throughout the library.

PR #365, opened 2025-05-23
chore: update versions to 0.3.0

nv-anants:
Update versions for 0.3.0 release

### DIFF
--- a/benchmark/nixlbench/meson.build
+++ b/benchmark/nixlbench/meson.build
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project('nixlbench', 'CPP', version: '0.2.0',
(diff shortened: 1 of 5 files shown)


tstamler:
should we be updating the benchmark/nixlbench/meson.build version number as well?

yosefe:
maybe we need CI workflow to make sure all versions are aligned

nv-anants:
> should we be updating the benchmark/nixlbench/meson.build version number as well?

updated to keep it aligned, but this seems like a utility so can follow independent versioning as well..

nv-anants:
> maybe we need CI workflow to make sure all versions are aligned

@yosefe do you mean like a find and grep command to match all versions or is there some tool  that we can leverage to scan the entire repo?

saturley-hall (approved):
🚀 go go release!

PR #367, opened 2025-05-23
UCX: set rma rails to 2 for ucx backend.

jeynmann:
Set `UCX_MAX_RMA_RAILS` to 2 (default is 1) to allow ucx to use maximum 2 ports for RMA operations.
It could make ucx backend get better performance when using a bond type interface.

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -364,6 +364,7 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         ucp_config_modify(ucp_config, "NET_DEVICES", dev_str.c_str());
     }
 
+    ucp_config_modify(ucp_config, "MAX_RMA_RAILS", "2");
     status = ucp_init(&ucp_params, ucp_config, &ctx);
     if (status != UCS_OK) {
         /* TODO: proper cleanup */


yosefe (approved):

src/utils/ucx/ucx_utils.cpp
```diff
@@ -364,6 +364,7 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         ucp_config_modify(ucp_config, "NET_DEVICES", dev_str.c_str());
     }
 
+    ucp_config_modify(ucp_config, "MAX_RMA_RAILS", "2");
```
brminich: i'd check return status and print warn or diag message if it is not ok

src/utils/ucx/ucx_utils.cpp
```diff
@@ -364,6 +364,11 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         ucp_config_modify(ucp_config, "NET_DEVICES", dev_str.c_str());
     }
 
+    status = ucp_config_modify(ucp_config, "MAX_RMA_RAILS", "2");
```
tvegas1, line 390: is there an issue with setting it to 4 instead, as with EFA, going from 2 to 4 effectively doubles the throughput?

brminich: may not work well for small/middle size messages on IB. I'd not do it without perf study

brminich:
ok to test 5f482a7dc1af1e8f4bc7eb6be07c962731fa18bc

tvegas1 (approved):

PR #369, opened 2025-05-23
feat: acquire the agent from the registration handle

ryanolson:


src/bindings/rust/src/lib.rs
```diff
@@ -41,35 +41,35 @@ mod bindings {
...
-    nixl_capi_xfer_dlist_trim, nixl_capi_reg_dlist_trim, nixl_capi_xfer_dlist_rem_desc,
-    nixl_capi_reg_dlist_rem_desc, nixl_capi_xfer_dlist_print, nixl_capi_reg_dlist_print,
-    nixl_capi_agent_make_connection, nixl_capi_agent_prep_xfer_dlist,
-    nixl_capi_agent_make_xfer_req, nixl_capi_create_xfer_dlist_handle,
-    nixl_capi_destroy_xfer_dlist_handle, nixl_capi_estimate_xfer_cost, nixl_capi_gen_notif
+    nixl_capi_xfer_dlist_add_desc, nixl_capi_xfer_dlist_clear, nixl_capi_xfer_dlist_desc_count,
+    nixl_capi_xfer_dlist_get_type, nixl_capi_xfer_dlist_has_overlaps,
+    nixl_capi_xfer_dlist_is_empty, nixl_capi_xfer_dlist_is_sorted, nixl_capi_xfer_dlist_len,
+    nixl_capi_xfer_dlist_print, nixl_capi_xfer_dlist_rem_desc, nixl_capi_xfer_dlist_resize,
+    nixl_capi_xfer_dlist_trim, nixl_capi_xfer_dlist_verify_sorted,
```
roiedanino, line 72: Maybe it would be better to try adding new functions at the end of the list to avoid time-consuming and error-prone merge conflicts.

gleon99:
@ryanolson 
1. Please elaborate what happens here and why.
2. Does "feat" stand for "feature"?
Thanks!

src/bindings/rust/src/lib.rs
gleon99: +1 let's try to avoid unnecessary diffs

mkhazraee:
@ryanolson I'm not quite following the goal, each registration belongs to a single agent, and through that agent you can reach this method. Is this about xfer_desc_list received from a remote agent and you want to attach an agent name to it? 

Also generally I think the rust bindings should be consistent with our cpp API. Not adding/removing capabilities, those can go to a shim layer.

src/bindings/rust/src/lib.rs
ryanolson: the diff is because you are not enforcing `cargo fmt` which is a miss.  CI should fail if `cargo fmt` applies an update.

ryanolson: @nv-anants - we need to find a way to apply the clippy and fmt actions to the nixl-sys crate.

ryanolson:
> @ryanolson
> 
> 1. Please elaborate what happens here and why.
> 2. Does "feat" stand for "feature"?
>    Thanks!

NIXL fails to build on my machine without this change.  I update everything on the host, install nixl, then the bindings fail.  Perhaps it's my gcc / llvm, but the compilation fail on the includes in wrapper.h.

ryanolson:
oh crud.  this is not the PR i thought it was. closing.

PR #370, opened 2025-05-23
nixlbench: automatically detect NIXL lib path based on host CPU

iyastreb:
Currently nixlbench uses hardcoded `x86_64` path to the NIXL libraries, which makes it hard to build under arm64/aarch64.
This PR automatically detects host CPU family and uses the appropriate lib path

### DIFF
--- a/benchmark/nixlbench/meson.build
+++ b/benchmark/nixlbench/meson.build
@@ -40,7 +40,14 @@ nvshmem_lib_path = get_option('nvshmem_lib_path')
 
 # Find required libraries
 # NIXL
-nixl_lib_path = nixl_path + '/lib/x86_64-linux-gnu'
+host_cpu_family = host_machine.cpu_family()
+host_system = host_machine.system().to_lower()
+
+if host_system != 'linux' or host_cpu_family not in ['x86_64', 'aarch64']
+    error('This build only supports Linux on x86_64 or aarch64 architectures.')
+endif
+
+nixl_lib_path = nixl_path + '/lib/' + host_cpu_family + '-linux-gnu'
 nixl_lib = cpp.find_library('nixl', dirs: [nixl_lib_path])
 nixl_build = cpp.find_library('nixl_build', dirs: [nixl_lib_path])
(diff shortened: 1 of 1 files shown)


benchmark/nixlbench/meson.build
```diff
@@ -40,7 +40,7 @@ nvshmem_lib_path = get_option('nvshmem_lib_path')
 
 # Find required libraries
 # NIXL
-nixl_lib_path = nixl_path + '/lib/x86_64-linux-gnu'
+nixl_lib_path = nixl_path + '/lib/' + host_machine.cpu_family() + '-linux-gnu'
```
brminich: maybe add an error when if `host_machine.system() != linux` or f host_machine.cpu_family()  is not `x86_64` or
`aarch64`

iyastreb: done

ovidiusm:
The Dockerfile is also hardcoding these, could we remove those as well or would you prefer we do it separately?

iyastreb:
> The Dockerfile is also hardcoding these, could we remove those as well or would you prefer we do it separately?

Thanks for pointing it, I didn't even know about that Dockerfile)
I think it can be done separately, personally I'm not using that container

brminich (approved):

benchmark/nixlbench/meson.build
```diff
@@ -40,7 +40,12 @@ nvshmem_lib_path = get_option('nvshmem_lib_path')
 
 # Find required libraries
 # NIXL
-nixl_lib_path = nixl_path + '/lib/x86_64-linux-gnu'
+host_cpu_family = host_machine.cpu_family()
+if host_machine.system() != 'linux' or not (host_cpu_family == 'x86_64' or host_cpu_family == 'aarch64')
```
brminich: btw
maybe
```
host_cpu_family = host_machine.cpu_family()
if host_machine.system() != 'linux' or host_cpu_family not in ['x86_64', 'aarch64']
  error('This build only supports Linux on x86_64 or aarch64 architectures.')
endif
```

tvegas1:
seems you would need to remove test.file

brminich (approved):

PR #371, opened 2025-05-23
Gtest transfer test improvements

w1ldptr:
First, refactor the code without changing functionality:

- Extract MemBuffer into dedicated file for re-usability. Make it a proper template parameterized by memory type.
- Make memory type a template parameter over the whole test call chain.

Second, implement multiple improvements in existing test coverage:

- Allow executing test with different batch sizes (number of descriptors in a single Xfer request).
- Allow specifying transfer operation mode and test with both NIXL_WRITE and NIXL_READ.
- Use unique notification message for every Xfer request instead of reusing the same hardcoded one.
- Allow executing test data path loop on specified number of threads.

Third, add new test that validates genNotif() in multi-threaded environment.

### DIFF
--- a/test/gtest/mem_buffer.h
+++ b/test/gtest/mem_buffer.h
@@ -20,10 +20,78 @@
 
 #include "nixl_types.h"
 #include <cstdint>
+#include <stdexcept>
+#include <vector>
+#include <random>
+#include <memory>
+#include <type_traits>
 
 namespace gtest {
 
 template<nixl_mem_t MemType> class MemBuffer;
+
+template<nixl_mem_t MemType> using MemBufferVec = std::vector<MemBuffer<MemType>>;
+
+template<nixl_mem_t MemType> using MemBufferVec2D = std::vector<MemBufferVec<MemType>>;
+
+template<nixl_mem_t MemType>
+std::vector<uint8_t>
+createRandomData(size_t size,
+                 std::mt19937_64 &gen,
+                 std::uniform_int_distribution<uint64_t> &distrib) {
+    size_t aligned_size = (size + 7) & ~7;
+    std::vector<uint8_t> data(aligned_size);
+
+    for (size_t i = 0; i < aligned_size; i += 8) {
+        uint64_t rand_val = distrib(gen);
+        for (size_t j = 0; j < 8; ++j) {
+            data[i + j] = static_cast<uint8_t>(rand_val >> (j * 8));
+        }
+    }
+
+    data.resize(size);
+    return data;
+}
+
+template<nixl_mem_t LocalMemType, nixl_mem_t RemoteMemType, nixl_xfer_op_t Mode>
+void
+initBuffersWithRandomData(MemBufferVec<LocalMemType> &local_buffers,
+                          MemBufferVec<RemoteMemType> &remote_buffers,
+                          size_t count,
+                          size_t size,
+                          std::mt19937_64 &gen,
+                          std::uniform_int_distribution<uint64_t> &distrib) {
+    for (size_t i = 0; i < count; i++) {
+        auto random_data = createRandomData<LocalMemType>(size, gen, distrib);
+        if constexpr (Mode == NIXL_WRITE) {
+            local_buffers.emplace_back(std::move(random_data));
+            remote_buffers.emplace_back(size);
+        } else if constexpr (Mode == NIXL_READ) {
+            local_buffers.emplace_back(size);
+            remote_buffers.emplace_back(std::move(random_data));
+        } else {
+            static_assert(Mode == NIXL_WRITE || Mode == NIXL_READ, "Invalid transfer mode");
+        }
+    }
+}
+
+template<nixl_mem_t LocalMemType, nixl_mem_t RemoteMemType, nixl_xfer_op_t Mode>
+void
+zeroBuffers(MemBufferVec<LocalMemType> &local_buffers,
+            MemBufferVec<RemoteMemType> &remote_buffers,
+            size_t count) {
+    if constexpr (Mode == NIXL_WRITE) {
+        for (size_t i = 0; i < count; i++) {
+            remote_buffers[i].zero();
+        }
+    } else if constexpr (Mode == NIXL_READ) {
+        for (size_t i = 0; i < count; i++) {
+            local_buffers[i].zero();
+        }
+    } else {
+        static_assert(Mode == NIXL_WRITE || Mode == NIXL_READ, "Invalid transfer mode");
+    }
+}
 
 template<> class MemBuffer<DRAM_SEG> {
 public:
@@ -86,10 +154,9 @@ public:
     MemBuffer &
     operator=(const MemBuffer &) = delete;
 
-    MemBuffer(MemBuffer &&other) noexcept : buffer_(other.buffer_), size_(other.size_) {
-        other.buffer_ = nullptr;
-        other.size_ = 0;
-    }
+    MemBuffer(MemBuffer &&other) noexcept :
+            buffer_(std::exchange(other.buffer_, nullptr)),
+            size_(std::exchange(other.size_, 0)) {}
 
     MemBuffer &
     operator=(MemBuffer &&other) noexcept {
@@ -97,10 +164,8 @@ public:
             if (buffer_) {
                 cudaFree(buffer_);
             }
-            buffer_ = other.buffer_;
-            size_ = other.size_;
-            other.buffer_ = nullptr;
-            other.size_ = 0;
+            buffer_ = std::exchange(other.buffer_, nullptr);
+            size_ = std::exchange(other.size_, 0);
         }
         return *this;
     }
@@ -134,6 +199,5 @@ private:
 } // namespace gtest
 
 #endif // HAVE_CUDA
-
 
 #endif /* MEM_BUFFER_H */
--- a/test/gtest/mem_buffer.h
+++ b/test/gtest/mem_buffer.h
@@ -46,6 +46,11 @@ public:
         return buffer_.size();
     }
 
+    void
+    zero() {
+        std::fill(buffer_.begin(), buffer_.end(), 0);
+    }
+
 private:
     std::vector<uint8_t> buffer_;
 };
@@ -114,6 +119,11 @@ public:
     size_t
     size() const {
         return size_;
+    }
+
+    void
+    zero() {
+        // TODO
     }
 
 private:
--- a/test/gtest/mem_buffer.h
+++ b/test/gtest/mem_buffer.h
@@ -1,0 +1,112 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef MEM_BUFFER_H
+#define MEM_BUFFER_H
+
+#include "nixl_types.h"
+#include <cstdint>
+
+namespace gtest {
+
+template<nixl_mem_t MemType> class MemBuffer;
+
+template<> class MemBuffer<DRAM_SEG> {
+public:
+    MemBuffer(size_t size) : buffer_(size) {}
+
+    uintptr_t
+    data() const {
+        return reinterpret_cast<uintptr_t>(buffer_.data());
+    }
+
+    size_t
+    size() const {
+        return buffer_.size();
+    }
+
+private:
+    std::vector<uint8_t> buffer_;
+};
+
+} // namespace gtest
+
+#ifdef HAVE_CUDA
+
+#include <cuda_runtime.h>
+
+namespace gtest {
+
+template<> class MemBuffer<VRAM_SEG> {
+public:
+    MemBuffer(size_t size) : size_(size) {
+        cudaError_t err = cudaMalloc(&buffer_, size);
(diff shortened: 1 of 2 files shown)


test/gtest/test_transfer.cpp
```diff
@@ -312,56 +482,108 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
-                   size, count, repeat,
-                   DRAM_SEG, src_buffers,
-                   DRAM_SEG, dst_buffers);
+TEST_P(TestTransfer, RandomSizes) {
+    // Tuple fields are: size, count, batch_size, repeat, num_threads
+    constexpr std::array<std::tuple<size_t, size_t, size_t, size_t, size_t>, 4> test_cases = {
+            {{40, 250, 1, 1, 4}, {4096, 32, 2, 2, 4}, {32768, 16, 2, 2, 4}, {1000000, 8, 1, 2, 4}}};
+
+    for (const auto &[size, count, batch_size, repeat, num_threads] : test_cases) {
+        doTransfers<DRAM_SEG, DRAM_SEG>(getAgent(0),
```
tvegas1: any way to avoid all those if() else() with copy pasted params + read/write, or while at it templatize NIXL_READ param?
```
auto transfer = &doTransfers<DRAM_SEG, DRAM_SEG>;
auto nixl_op = NIXL_READ;
transfer(..., nixl_op, ...);
```

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,139 @@
...
+            throw std::runtime_error("Failed to allocate CUDA memory");
+        }
+    }
+
+    MemBuffer(std::vector<uint8_t> &&data) {
+        // TODO
+    }
+
+    ~MemBuffer() {
+        if (buffer_) {
```
tvegas1, line 148: i was under impression that codebase was supposed to use m_ prefix?

test/gtest/test_transfer.cpp
```diff
@@ -223,81 +187,287 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+                } else {
+                    thread_local_buffers[thread_id][i].zero();
+                }
+            }
+        }
+    }
+
+    template<nixl_mem_t LocalMemType, nixl_mem_t RemoteMemType>
+    void
+    doTransfer(nixlAgent &from,
```
tvegas1: not too sure, but are we sure we want to templatize up to here? i see that MemBuffer is now template but maybe doTransfer() does the same work any ways

w1ldptr: This will involve bunch of conditionals down the call chain because you will need to instantiate correct MemBuffer.

test/gtest/mem_buffer.h
w1ldptr: I wasn't aware. What makes you think that?

test/gtest/test_transfer.cpp
w1ldptr: Not sure I'm following. Where are any if/else clauses here?

test/gtest/mem_buffer.h
tvegas1: seems anyways no clear side in the codebase. there was some discussion where you mentioned preferring _ suffix for auto completion but i do not remember where it happened.

test/gtest/test_transfer.cpp
tvegas1: actually right, if/else is below:
```
    if (m_cuda_device) {
        doTransfers<VRAM_SEG, VRAM_SEG>(getAgent(0),
                                        getAgentName(0),
                                        getAgent(1),
                                        getAgentName(1),
                                        size,
                                        count,
                                        1,
                                        1,
                                        1,
                                        NIXL_WRITE,
                                        [this]() { exchangeMDIP(); });
    } else {
        doTransfers<DRAM_SEG, DRAM_SEG>(getAgent(0),
                                        getAgentName(0),
                                        getAgent(1),
                                        getAgentName(1),
                                        size,
                                        count,
                                        1,
                                        1,
                                        1,
                                        NIXL_WRITE,
                                        [this]() { exchangeMDIP(); });
    }
```

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,139 @@
...
+    MemBuffer &
+    operator=(MemBuffer &&other) noexcept {
+        if (this != &other) {
+            if (buffer_) {
+                cudaFree(buffer_);
+            }
+            buffer_ = other.buffer_;
+            size_ = other.size_;
+            other.buffer_ = nullptr;
+            other.size_ = 0;
```
rakhmets: Can be replaced by private method `other.reset()` here and in move ctor.

test/gtest/test_transfer.cpp
```diff
@@ -223,81 +187,289 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+               nixlAgent &to,
+               const std::string &to_name,
+               size_t size,
+               size_t count,
+               size_t batch_size,
+               nixl_xfer_op_t mode,
+               std::function<void()> setup_md,
+               const std::vector<std::string> &expected_notifs,
+               const std::vector<MemBuffer<LocalMemType>> &local_buffers,
+               const std::vector<MemBuffer<RemoteMemType>> &remote_buffers) {
```
rakhmets: Please add aliases:
```C++
using MemBufferVec<MemType> = std::vector<MemBuffer<MemType>>;
using MemBufferVec2D<MemType> = std::vector<MemBufferVec<MemType>>;
```
to reduce the amount of duplicate code.

test/gtest/test_transfer.cpp
```diff
@@ -155,21 +120,20 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
     nixlDescList<Desc>
-    makeDescList(const std::vector<MemBuffer> &buffers, nixl_mem_t mem_type)
-    {
-        nixlDescList<Desc> desc_list(mem_type);
-        for (const auto &buffer : buffers) {
-            desc_list.addDesc(Desc(buffer, buffer.getSize(), DEV_ID));
+    makeDescList(Iter begin, Iter end) {
+        nixlDescList<Desc> desc_list(MemType);
+        for (auto it = begin; it != end; ++it) {
+            desc_list.addDesc(Desc(it->data(), it->size(), DEV_ID));
```
```suggestion
        for (; begin != end; ++begin) {
            desc_list.addDesc(Desc(begin->data(), begin->size(), DEV_ID));
```

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,139 @@
...
+
+    MemBuffer(MemBuffer &&other) noexcept : buffer_(other.buffer_), size_(other.size_) {
+        other.buffer_ = nullptr;
+        other.size_ = 0;
+    }
+
+    MemBuffer &
+    operator=(MemBuffer &&other) noexcept {
+        if (this != &other) {
+            if (buffer_) {
```
iyastreb, line 164: I think we can use std::swap both for buffer and size, maybe explicit free is not needed here

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,139 @@
...
+#ifdef HAVE_CUDA
+
+#include <cuda_runtime.h>
+
+namespace gtest {
+
+template<> class MemBuffer<VRAM_SEG> {
+public:
+    MemBuffer(size_t size) : size_(size) {
+        cudaError_t err = cudaMalloc(&buffer_, size);
```
iyastreb, line 137: maybe it makes to create a wrapper for CUDA calls that checks err and throws an exception. This might be a common usage pattern

test/gtest/test_transfer.cpp
```diff
@@ -223,81 +187,289 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+
+            validateNotifications(notif_map, from_name, base_thread_notifs);
+        }
+
+        invalidateMD();
+    }
+
+    void
+    doNotificationTest(nixlAgent &from,
+                       const std::string &from_name,
```
rakhmets: Maybe add `struct TestAgent { nixlAgent agent; std::string name};`. Then you can use it here, and in other methods (`doTransfers`, `doTransfer`). To reduce code duplication.

test/gtest/mem_buffer.h
w1ldptr: Maybe, but as you can see by most of the functions in VRAM specialization being empty with just "TODO" inside I'm not attempting comprehensive CUDA wrapper here since I'm not familiar with it at all. This PR is complex enough as it is so I suggest leaving CUDA for later.

test/gtest/test_transfer.cpp
w1ldptr: Even better, I just pass agent index and obtain instances with GetAgent() and GetAgentName().

test/gtest/mem_buffer.h
w1ldptr: Re-implemented with std::exchange.

test/gtest/test_transfer.cpp
w1ldptr: I couldn't really make it work to save a pointer to a specialized class function template here without lambdas that look even worse. However, it may look better now with refactoring suggested by @rakhmets .

w1ldptr: However, I was able to templatize WRITE/READ mode.

test/gtest/mem_buffer.h
w1ldptr: Hmm, I think that std::exchange is recommended for move ctors and operator= which I used in the new version since the code is moving, now swapping. I don't have a strong opinion either way though, so let me know what you think about it.

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,203 @@
...
+        std::fill(buffer_.begin(), buffer_.end(), 0);
+    }
+
+private:
+    std::vector<uint8_t> buffer_;
+};
+
+} // namespace gtest
+
+#ifdef HAVE_CUDA
```
etoledano, line 128: I don't think condition is needed, if using undefined content within you can declare the template with ```std::_enable_if``` for SFINAE to be triggered, as @dfarge is working on.

w1ldptr: I don't see how we can completely remove ifdef since we would still need to conditionally include cuda_runtime.h. I might be missing something, but it looks like the end result would be both some ifdef and also std:enable_if in the template, which appears like an inferior solution. Could you show a working version of this with ifdef fully removed?

etoledano: I understand.
Ideally having a single `ifdef` to determine headers to include and other definitions sounds more appealing than scattering around the code, if you agree with this then a working solution could be as below, wdyt?

P.S
The code below isn't ideal, we can use functions directly, just took it from a scratchpad of mine to showcase.

```

// nixl_config.h
struct CudaEnabledTag {};
struct CudaDisabledTag {};

#ifdef HAVE_CUDA
using CudaMode = CudaEnabledTag;
#include "cuda_runtime.h"
#else
using CudaMode = CudaDisabledTag;
#endif

// sync.cpp
template<typename Mode, typename Enable = void>
struct syncImpl;

template<typename Mode>
struct syncImpl<Mode, std::enable_if_t<std::is_same<Mode, CudaEnabledTag>::value>> {
   static void sync() {
       std::cout << "sync::CudaEnabledTag" << std::endl;
   }
};

template<typename Mode>
struct syncImpl<Mode, std::enable_if_t<std::is_same<Mode, CudaDisabledTag>::value>> {
   static void sync() {
       std::cout << "sync::CudaDisabledTag" << std::endl;
   }
};


void sync0() {
   syncImpl<CudaMode>::sync();
}

int main() {
   sync0();
   return 0;
}

```

w1ldptr: So instead of single ifdef we get more complicated ifdef with else clause and more complicated template definition. How is this an improvement?

test/gtest/mem_buffer.h
```diff
@@ -0,0 +1,203 @@
...
+#include <cstdint>
+#include <stdexcept>
+#include <vector>
+#include <random>
+#include <memory>
+#include <type_traits>
+
+namespace gtest {
+
+template<nixl_mem_t MemType> class MemBuffer;
```
yosefe, line 31: since gtest is not data path, i think it would simplify the code to make mem_type a run time parameter and not a template argument.
this way we can iterate on all mem type combinations using a simple loop.
imagine having even more memory types (OBJ, FILE, ...) - having this as template would exponentially increase the code size.

yosefe (changes requested):

test/gtest/mem_buffer.h
etoledano: Resolved offline and agreed to proceed after @dfarge PR for reference

test/gtest/mem_buffer.h
w1ldptr: The goal of templatizing MemBuffer was not to improve runtime performance by removing run-time conditionals, but to improve code readability and maintainability by having a clean separate implementations of MemBuffer for every memory type instead of a single class with "if (DRAM) .. else if (VRAM) .. else if ..." in every one of its methods. Having some boilerplate in TEST_P blocks looks like a lesser evil to me.

etoledano: We could consider using templates to replace the `#ifdef` blocks (for CUDA), allowing the memory type to be dynamic. This approach maintains the current code size while leveraging the advantages of templates over conditional compilation.

@yosefe @w1ldptr 
what do you think?

w1ldptr: My original goal was to just add multi-threading to existing tests. After familiarizing myself with the implementation I decided to also perform some obvious QoL and readability improvements not expecting them to be controversial TBH. Guess I'll just put this one aside until we have some consensus on how memory wrappers should look like (from @dfarge work or otherwise) and implement some minimal multi-threading extensions for now.

yosefe: the problem with templated class is that you have to pull up the memory types template argument to all places using the MemBuffer class, like the transfer tests, instead of using simple encapsulation and letting most of the functions in transfer tests be unaware of the memory types.
can use switch/case instead of if/else if it looks better

w1ldptr: Yes, there is a trade-off between compile-time and run-time dispatch of type of memory being tested. We either have to deal with it by 1. copy/pasting some boilerplate in TEST_P handler (trivial code that we don't be changing or debugging much in the future) or 2. in the test themselves (more complicated code like doTransfer) + make all methods of MemBuffer a chain of conditionals (doesn't matter whether chained if/else clauses or big switch statements). To me the first option appears superior.

PR #372, opened 2025-05-23
Use NIXL logging macros consistently

ovidiusm:
Replace remanining cout/cerr calls in src with proper logging macros

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -16,7 +16,6 @@
  */
 
 #include <fcntl.h>
-#include <iostream>
 #include "nixl.h"
 #include "common/nixl_time.h"
 #include "common/str_tools.h"
--- a/src/core/nixl_plugin_manager.cpp
+++ b/src/core/nixl_plugin_manager.cpp
@@ -19,13 +19,11 @@
 #include "nixl.h"
 #include "common/nixl_log.h"
 #include <dlfcn.h>
-#include <iostream>
 #include <filesystem>
 #include <dirent.h>
 #include <unistd.h>  // For access() and F_OK
 #include <cstdlib>  // For getenv
 #include <fstream>
-#include <iostream>
 #include <string>
 #include <map>
 
@@ -277,8 +275,8 @@ void nixlPluginManager::discoverPluginsFromDir(const std::string& dirpath) {
     std::error_code ec;
     std::filesystem::directory_iterator dir_iter(dir_path, ec);
     if (ec) {
-        std::cerr << "Error accessing directory(" << dir_path << "):"
-                  << ec.message() << std::endl;
+        NIXL_ERROR << "Error accessing directory(" << dir_path << "):"
+                   << ec.message();
         return;
     }
 
@@ -361,7 +359,7 @@ void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginC
 
     //Static Plugins are considered pre-loaded
     nixlBackendPlugin* plugin = info.createFunc();
-    NIXL_DEBUG << "Loading static plugin: " << name << std::endl;
+    NIXL_DEBUG << "Loading static plugin: " << name;
     if (plugin) {
         // Register the loaded plugin
         auto plugin_handle = std::make_shared<const nixlPluginHandle>(nullptr, plugin);
--- a/src/infra/nixl_descriptors.cpp
+++ b/src/infra/nixl_descriptors.cpp
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 #include <algorithm>
-#include <iostream>
 #include <functional>
 #include <stdexcept>
 #include "nixl.h"
@@ -23,6 +22,7 @@
 #include "mem_section.h"
 #include "backend/backend_aux.h"
 #include "serdes/serdes.h"
+#include "common/nixl_log.h"
 
 /*** Class nixlBasicDesc implementation ***/
 
@@ -89,8 +89,8 @@ nixl_blob_t nixlBasicDesc::serialize() const {
 }
 
 void nixlBasicDesc::print(const std::string &suffix) const {
-    std::cout << "LOG: Desc (" << addr << ", " << len
-              << ") from devID " << devId << suffix << "\n";
+    NIXL_INFO << "Desc (" << addr << ", " << len
+              << ") from devID " << devId << suffix;
 }
 
 
@@ -409,10 +409,9 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
 template <class T>
 void nixlDescList<T>::print() const {
-    std::cout << "LOG: DescList of mem type " << type
-              << (sorted ? "sorted" : "unsorted") << "\n";
+    NIXL_INFO << "DescList of mem type " << type
+              << (sorted ? "sorted" : "unsorted");
     for (auto & elm : descs) {
-        std::cout << "    ";
         elm.print("");
     }
 }
--- a/src/plugins/cuda_gds/gds_utils.cpp
+++ b/src/plugins/cuda_gds/gds_utils.cpp
@@ -14,8 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include <iostream>
 #include "gds_utils.h"
+#include "common/nixl_log.h"
 
 nixl_status_t gdsUtil::registerFileHandle(int fd,
                                           size_t size,
@@ -31,7 +31,7 @@ nixl_status_t gdsUtil::registerFileHandle(int fd,
 
     status = cuFileHandleRegister(&handle, &descr);
     if (status.err != CU_FILE_SUCCESS) {
-        std::cerr << "file register error:" << std::endl;
+        NIXL_ERROR << "file register error:";
         return NIXL_ERR_BACKEND;
     }
 
@@ -51,7 +51,7 @@ nixl_status_t gdsUtil::registerBufHandle(void *ptr,
 
     status = cuFileBufRegister(ptr, size, flags);
     if (status.err != CU_FILE_SUCCESS) {
-        std::cerr << "Warning: Buffer registration failed - will use compat mode\n";
+        NIXL_WARN << "Warning: Buffer registration failed - will use compat mode";
     }
     return NIXL_SUCCESS;
 }
@@ -62,7 +62,7 @@ nixl_status_t gdsUtil::openGdsDriver()
 
     err = cuFileDriverOpen();
     if (err.err != CU_FILE_SUCCESS) {
-        std::cerr << "Error initializing GPU Direct Storage driver\n";
+        NIXL_ERROR << "Error initializing GPU Direct Storage driver";
         return NIXL_ERR_BACKEND;
     }
     return NIXL_SUCCESS;
@@ -84,7 +84,7 @@ nixl_status_t gdsUtil::deregisterBufHandle(void *ptr)
 
     status = cuFileBufDeregister(ptr);
     if (status.err != CU_FILE_SUCCESS) {
-        std::cerr << "Error De-Registering Buffer\n";
+        NIXL_ERROR << "Error De-Registering Buffer";
         return NIXL_ERR_BACKEND;
     }
     return NIXL_SUCCESS;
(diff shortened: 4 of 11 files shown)


src/utils/stream/metadata_stream.cpp
```diff
@@ -180,18 +178,18 @@ bool nixlMDStreamClient::setupClient() {
 
     if (inet_pton(AF_INET, listenerAddress.c_str(),
                   &listenerAddr.sin_addr) <= 0) {
-        perror("Invalid address/ Address not supported");
+        NIXL_ERROR << "Invalid address/ Address not supported: " << strerror(errno);
```
tvegas1: maybe replace `strerror(errno)` which might not be thread-safe by something like `std::error_code ec(errno, std::generic_category());`?

src/infra/nixl_descriptors.cpp
```diff
@@ -89,8 +89,8 @@ nixl_blob_t nixlBasicDesc::serialize() const {
 }
 
 void nixlBasicDesc::print(const std::string &suffix) const {
-    std::cout << "LOG: Desc (" << addr << ", " << len
-              << ") from devID " << devId << suffix << "\n";
+    NIXL_INFO << "Desc (" << addr << ", " << len
```
brminich, line 93: maybe using `absl::StrFormat` would be easier to read?

here and in other places where many params are printed

src/plugins/cuda_gds/gds_utils.cpp
```diff
@@ -51,7 +51,7 @@ nixl_status_t gdsUtil::registerBufHandle(void *ptr,
 
     status = cuFileBufRegister(ptr, size, flags);
     if (status.err != CU_FILE_SUCCESS) {
-        std::cerr << "Warning: Buffer registration failed - will use compat mode\n";
+        NIXL_WARN << "Warning: Buffer registration failed - will use compat mode";
```
brminich: maybe
```suggestion
        NIXL_WARN << "Buffer registration failed - will use compat mode";
```
as afaiu `warning` will be printed by absl

src/utils/stream/metadata_stream.cpp
```diff
@@ -145,10 +143,10 @@ void nixlMDStreamListener::recvFromClients(int clientSocket) {
               std::string ack = "Message received";
               send(clientSocket, ack.c_str(), ack.size(), 0);
               std::string recv_message(buffer);
-              std::cout << "Message Received" << recv_message <<"\n";
+              NIXL_INFO << "Message Received" << recv_message;
         }
         close(clientSocket);
-        std::cout << "Client Disconnected\n";
+        NIXL_INFO << "Client Disconnected";
```
brminich: maybe use debug for such messages to not to overload info?

src/infra/nixl_descriptors.cpp
rakhmets: I suggest to keep `<<`, because it provides type safety, and doesn't rely on type specifiers.

rakhmets (dismissed):

src/infra/nixl_descriptors.cpp
brminich: ok

ovidiusm: Also log filtering may happen after strformat which impacts performance

src/utils/stream/metadata_stream.cpp
ovidiusm: Done

src/utils/stream/metadata_stream.cpp
ovidiusm: Added nixl_strerror in nixl_log.h and replaces occurences overall

tvegas1 (approved):

aranadive (approved):

src/core/nixl_listener.cpp
```diff
@@ -72,7 +71,7 @@ int connectToIP(std::string ip_addr, int port) {
     if (ret <= 0) {
         if (ret < 0) {
             NIXL_ERROR << "select failed for ip_addr: " << ip_addr << " and port: " << port
-                       << " with error: " << strerror(errno);
+                       << " with error: " << nixl_strerror(errno);
```
brminich: can we add one more NIXL log macro using absl `PLOG`? then we would not need `nixl_strerror`, because it is already handled by absl
the idea is to use it in all places where we want errno to be printed

```
// PLOG()
//
// `PLOG` behaves like `LOG` except that a description of the current state of
// `errno` is appended to the streamed message.
#define PLOG(severity) ABSL_LOG_INTERNAL_PLOG_IMPL(_##severity)

```

src/core/nixl_plugin_manager.cpp
```diff
@@ -361,7 +359,7 @@ void nixlPluginManager::registerStaticPlugin(const char* name, nixlStaticPluginC
 
     //Static Plugins are considered pre-loaded
     nixlBackendPlugin* plugin = info.createFunc();
-    NIXL_DEBUG << "Loading static plugin: " << name << std::endl;
+    NIXL_DEBUG << "Loading static plugin: " << name;
```
brminich: maybe better INFO?
I'd use
info to print some initialization data (printed just once and important to understand what component is initizlized)
debug to print  MD flow (not data path)
trace (just datapath)

yosefe (commented):
LGTM except mikhail's comment

src/core/nixl_plugin_manager.cpp
aranadive: Im not sure about INFO, this could mean a lot of plugins/text output for each agent and end user may not care.

yosefe: INFO is not enabled by default, right?

brminich: yes
I thought to log all objects creation in INFO, MD exchange in debug and data in trace

aranadive: Sorry I thought INFO was default.. WARN is default. Then INFO is okay with me.

brminich: @aranadive what kind of information would you suggest to print with info level?

aranadive: > yes I thought to log all objects creation in INFO, MD exchange in debug and data in trace

Agreed for INFO. But also add obj destruction and success of API functions in debug?
By data do you mean, function args as well?

ovidiusm: Is there anything else to do here, or can we close the conversation?

src/plugins/hf3fs/meson.build
```diff
@@ -21,14 +21,13 @@ threefs_dep = declare_dependency(
   include_directories : include_directories(threefs_inc_path))
 
 # Get Abseil dependencies
```
barneuman: Please remove the comment too

src/plugins/hf3fs/hf3fs_log.h
```diff
@@ -22,7 +22,7 @@
 
 #define HF3FS_LOG_RETURN(error_code, message) \
     do { \
-        NIXL_ERROR << absl::StrFormat("HF3FS error: %d - %s", (error_code), (message)); \
+        NIXL_PERROR << absl::StrFormat("HF3FS error: %d - %s", (error_code), (message)); \
```
barneuman: errno isn't relevant for all the use cases of HF3FS_LOG_RETURN (for example: validateIO)

src/core/nixl_plugin_manager.cpp
brminich: did not we agree to change to INFO?

src/plugins/hf3fs/hf3fs_log.h
ovidiusm: I see. I will revert the change

barneuman:
LGTM

src/core/nixl_plugin_manager.cpp
```diff
@@ -277,8 +275,8 @@ void nixlPluginManager::discoverPluginsFromDir(const std::string& dirpath) {
     std::error_code ec;
     std::filesystem::directory_iterator dir_iter(dir_path, ec);
     if (ec) {
-        std::cerr << "Error accessing directory(" << dir_path << "):"
-                  << ec.message() << std::endl;
+        NIXL_ERROR << "Error accessing directory(" << dir_path << "):"
```
yosefe: add space after :

src/infra/nixl_descriptors.cpp
```diff
@@ -409,10 +409,9 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
 template <class T>
 void nixlDescList<T>::print() const {
-    std::cout << "LOG: DescList of mem type " << type
-              << (sorted ? "sorted" : "unsorted") << "\n";
+    NIXL_INFO << "DescList of mem type " << type
+              << (sorted ? "sorted" : "unsorted");
```
yosefe: missing space after type

src/utils/common/nixl_log.h
```diff
@@ -78,4 +109,17 @@
...
+/*-----------------------------------------------------------------------------*
+ * Helper Functions
+ *-----------------------------------------------------------------------------*/
+
+/*
+ * Get the error message for the given error number. Thread-safe.
+ * @param err: The error number.
+ * @return: The error message.
+ */
+inline std::string nixl_strerror(int err) {
```
yosefe: static

src/utils/stream/metadata_stream.cpp
```diff
@@ -34,7 +34,7 @@ bool nixlMetadataStream::setupStream() {
 
     socketFd = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);
     if (socketFd == -1) {
-        std::cerr << "failed to create stream socket for listener";
+        NIXL_ERROR << "failed to create stream socket for listener";
```
yosefe: NIXL_PERROR

src/utils/stream/metadata_stream.cpp
```diff
@@ -70,33 +70,32 @@ void nixlMDStreamListener::setupListener() {
     int opt = 1;
     if (setsockopt(socketFd, SOL_SOCKET, SO_REUSEADDR,
                    &opt, sizeof(opt)) < 0) {
-        std::cerr << "setsockopt(REUSEADDR) failed while setting up listener for MD\n";
+        NIXL_ERROR << "setsockopt(REUSEADDR) failed while setting up listener for MD";
```
yosefe: NIXL_PERROR

src/utils/stream/metadata_stream.cpp
```diff
@@ -70,33 +70,32 @@ void nixlMDStreamListener::setupListener() {
...
         closeStream();
         return;
     }
 
     if (bind(socketFd, (struct sockaddr*)&listenerAddr,
              sizeof(listenerAddr)) < 0) {
-        std::cerr << "Socket Bind failed while setting up listener for MD\n";
+        NIXL_ERROR << "Socket Bind failed while setting up listener for MD";
```
yosefe: NIXL_PERROR

src/utils/stream/metadata_stream.cpp
```diff
@@ -70,33 +70,32 @@ void nixlMDStreamListener::setupListener() {
...
         closeStream();
         return;
     }
 
     if (listen(socketFd, 128) < 0) {
-        std::cerr << "Listening failed for stream Socket: "
-                  << socketFd  << "\n";
+        NIXL_ERROR << "Listening failed for stream Socket: "
```
yosefe: NIXL_PERROR

src/infra/nixl_descriptors.cpp
```diff
@@ -409,10 +409,9 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
 template <class T>
 void nixlDescList<T>::print() const {
-    std::cout << "LOG: DescList of mem type " << type
-              << (sorted ? "sorted" : "unsorted") << "\n";
+    NIXL_INFO << "DescList of mem type " << type
```
brminich: looking at the code it should be printed with default log level. I do not think error or warning is appropriate, so maybe just use cout?

ovidiusm: Based on the name of the function (print), it sounds like the cout should indeed be used. Just the fact that there was LOG in the original message made me think this should be replaced with a log message. But I think the intent is to actually print on stdout, so let's do that.

yosefe: maybe keep nixlDescList as is for now (use cout), going forward print() should be deprecated in favor of operator<< IMO

brminich (approved):

PR #387, opened 2025-05-26
POSIX: updated queue_factory_impl, and fixed meson logic

dfarge:


meson.build
```diff
@@ -28,6 +28,13 @@ dl_dep = cpp.find_library('dl', required: true)
 rt_dep = cpp.find_library('rt', required: true)
 thread_dep = dependency('threads')
 
+# Check for libaio (for POSIX plugin and test)
```
vvenkates27, line 31: Why was this removed and why is this added back. Can we make sure that this does not fail with RHEL during pip creation?

src/plugins/posix/queue_factory_impl.cpp
```diff
@@ -0,0 +1,92 @@
...
+#ifdef HAVE_LIBURING
+#include "uring_queue.h"
+#endif
+
+// Anonymous namespace for internal template implementations for functions that use the optional liburing
+namespace {
+    struct uringEnabled {};
+    struct uringDisabled {};
+
+    #ifdef HAVE_LIBURING
```
vvenkates27: So this changes uring ifdef to one place alone and uses function pointers to create create routines?

src/plugins/posix/queue_factory_impl.cpp
```diff
@@ -0,0 +1,92 @@
...
+    using uringMode = uringEnabled;
+    #else
+    using uringMode = uringDisabled;
+    #endif
+
+    template <typename Mode, typename Enable = void>
+    struct createUringQueueImpl;
+
+    template <typename Mode>
```
vvenkates27, line 42: inconsistent styling.

src/plugins/posix/queue_factory_impl.cpp
```diff
@@ -0,0 +1,92 @@
...
+        static std::unique_ptr<nixlPosixQueue> create(int num_entries, nixl_xfer_op_t operation) {
+            // Initialize io_uring parameters with basic configuration
+            // Start with basic parameters, no special flags
+            // We can add optimizations like SQPOLL later
+            struct io_uring_params params = {};
+            return std::make_unique<class UringQueue>(num_entries, params, operation);
+        }
+    };
+
+    template <typename Mode>
```
vvenkates27, line 57: Lots of tab/space styling issues throughout this file -- Please fix.

meson.build
```diff
@@ -28,6 +28,13 @@ dl_dep = cpp.find_library('dl', required: true)
...
+aio_dep = cpp.find_library('aio', required: false)
+paio = false
```
aranadive: Lets rename paio to posix_aio

src/plugins/posix/meson.build
```diff
@@ -13,27 +13,19 @@
...
+    'queue_factory_impl.cpp',
     'aio_queue.cpp'  # Always include AIO source since it's required
 ]
 
-if aio_dep.found()
-    paio = cpp.has_function('aio_cancel', prefix: '#include <aio.h>')
-endif
-
 # If libaio is not found, skip building the POSIX plugin entirely
 if aio_dep.found() and paio
```
aranadive: Should just be if paio

test/unit/plugins/posix/meson.build
```diff
@@ -13,13 +13,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Get Abseil dependencies
-absl_log_dep = dependency('absl_log', required: true)
+# Using globally defined aio_dep and paio variables from the root meson.build
+if (aio_dep.found() and paio) or rt_dep.found()
```
aranadive: If paio

meson.build
dfarge: Renamed as suggested

src/plugins/posix/meson.build
dfarge: Correct, changed as suggested

test/unit/plugins/posix/meson.build
dfarge: Correct, changed as suggested

src/plugins/posix/queue_factory_impl.cpp
aranadive: Dont indent the ifdefs

dfarge: Removed the indentation

src/plugins/posix/queue_factory_impl.cpp
dfarge: Can you please elaborate? I don't understand what styling issues you mean
Please note that we define templates for structs and member functions within those structs.

src/plugins/posix/queue_factory_impl.cpp
dfarge: Not exactly.
We define uringMode to be a synonym of struct  uringEnabled/Disabled with "using".
Then we create a template for structs with uringEnabled/Disabled.
In each template we create different member functions - with liburing on uringEnabled and w/o on uringDisabled.
The exposed API calls member functions of the struct created by the template with uringMode.
The unused template isn't compiled in.

dfarge:
I simplified by using a single struct that holds all the functionality instead of a different struct for each function.

meson.build
dfarge: This wasn't removed and added back, it was moved from the posix dir meson.build file to the project's dir's meson.build file.
The change was made so that the dependencies could be used to make decisions on the posix test compilation as well.
I didn't understand your question, can you explain further?

dfarge: @aranadive - Vish and I talked about this, and he asked for your input here. I use ubuntu, and I don't know how to order a RHEL setup for NIXL. How can I check this works? Why should this be a problem on RHEL?

aranadive: The dependency check seems okay to me.

dfarge: Great, thank you

w1ldptr (approved):

aranadive (approved):

PR #389, opened 2025-05-26
Gtest transfer test multi-threading extensions

w1ldptr:
- Use multiple UCX workers and permissive RW sync mode to verify transfer functionality in multi-threaded environment. Extend doTransfer to execute createXferReq/postXferReq/getXferStatus from multiple threads.
- Implement a test that verifies genNotif in multi-threaded environment. Test
is similar to existing ones based on doTransfer but doesn't use any Xfer
request relying only on genNotif API to pass notifications.

### DIFF
--- a/.ci/jenkins/lib/build-matrix.yaml
+++ b/.ci/jenkins/lib/build-matrix.yaml
@@ -1,0 +1,80 @@
+# Build Matrix Configuration for NixL CI Pipeline
+#
+# This file defines the build matrix configuration for the NixL CI pipeline in Jenkins.
+# It specifies the build environment, resources, and test matrix for continuous integration.
+#
+# Key Components:
+# - Job Configuration: Defines timeout, failure behavior, and Kubernetes resources
+# - Docker Images: Specifies the container images used for different build stages
+#   - PyTorch images (24.10 and 25.02) for building and testing
+#   - Podman image for container builds
+# - Matrix Axes: Defines build variations (currently x86_64 architecture)
+# - Build Steps: Sequential steps for building, testing, and container creation
+#
+# When Modified:
+# - Adding/removing Docker images: Affects available build environments
+# - Modifying matrix axes: Changes build variations (e.g., adding architectures)
+# - Adjusting resource limits: Impacts build performance and resource allocation
+# - Adding/removing steps: Changes the build pipeline sequence
+#
+# Note: Changes to this file are tested as part of the PR CI flow no need to test them manually.
+
+---
+job: nixl-ci-build
+
+# Fail job if one of the steps fails or continue
+failFast: false
+
+timeout_minutes: 240
+
+kubernetes:
+  cloud: il-ipp-blossom-prod
+  namespace: swx-media
+  limits: "{memory: 8Gi, cpu: 8000m}"
+  requests: "{memory: 8Gi, cpu: 8000m}"
+
+runs_on_dockers:
+  - { name: "pytorch-25.02-py3", url: "nvcr.io/nvidia/pytorch:25.02-py3" }
+  - { name: "pytorch-24.10-py3", url: "nvcr.io/nvidia/pytorch:24.10-py3" }
+  - { name: "podman-v5.0.2", url: "quay.io/podman/stable:v5.0.2", category: 'tool', privileged: true }
+
+matrix:
+  axes:
+    arch:
+      - x86_64
+      # - aarch64
+
+env:
+  NIXL_INSTALL_DIR: /opt/nixl
+
+steps:
+  - name: Build
+    parallel: false
+    run: |
+      if [[ "${name}" == "pytorch-24.10-py3" ]]; then
+        # distro's meson version is too old project requires >= 0.64.0
+        pip3 install meson
+      fi
+      .gitlab/build.sh ${NIXL_INSTALL_DIR}
+
+  - name: Test CPP
+    parallel: false
+    run: |
+      .gitlab/test_cpp.sh ${NIXL_INSTALL_DIR}
+
+  - name: Test Python
+    parallel: false
+    run: |
+      .gitlab/test_python.sh ${NIXL_INSTALL_DIR}
+
+  - name: Build Docker Image
+    parallel: false
+    containerSelector: "{ name: 'podman.*' }"
+    run: |
+      # change storage driver to improve build performance
+      rm -f /etc/containers/storage.conf ; podman system reset -f || true
+      # symlink podman to docker - scripts works with docker commands
+      ln -sfT $(type -p podman) /usr/bin/docker
+      # install git for building container image
+      yum install -y git
+      contrib/build-container.sh --no-cache
--- a/.ci/jenkins/pipeline/Jenkinsfile
+++ b/.ci/jenkins/pipeline/Jenkinsfile
@@ -1,0 +1,27 @@
+#!/usr/bin/groovy
+
+/*
+ * This is the main Jenkins pipeline definition file that defines the actual build steps
+ * It is referenced by the JJB configuration (proj-jjb.yaml) and executed by Jenkins
+ * The pipeline uses a matrix-based build approach to handle multiple build configurations
+ * NOTE: This file should not be changed by none devops team members
+ */
+
+// Load required pipeline libraries
+// These libraries provide shared functionality and utilities for the build process
+// Requires pipeline-github-lib plugin to load library from github
+@Library('github.com/Mellanox/ci-demo@stable_media')  // External library for matrix build functionality
+
+// Initialize the build matrix
+// The matrix class handles the parallel execution of different build configurations
+// It manages the build environment, test execution, and result reporting
+def matrix = new com.mellanox.cicd.Matrix()
+
+// Execute the main build process
+// This will:
+// 1. Parse the build matrix configuration from .ci/jenkins/lib/build-matrix.yaml
+// 2. Set up the build environment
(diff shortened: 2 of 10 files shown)


yosefe:
overall approach LGTM

test/gtest/test_transfer.cpp
```diff
@@ -325,23 +408,23 @@ const std::string TestTransfer::NOTIF_MSG = "notification";
...
 {
-    // Tuple fields are: size, count, repeat
-    constexpr std::array<std::tuple<size_t, size_t, size_t>, 4> test_cases = {
-        {{40, 1000, 1},
-         {4096, 8, 3},
-         {32768, 64, 3},
-         {1000000, 100, 3}}
+    // Tuple fields are: size, count, repeat, num_threads
+    constexpr std::array<std::tuple<size_t, size_t, size_t, size_t>, 4> test_cases = {
+        {{40, 1000, 1, 4},
```
tvegas1: maybe keep one/two single threaded

test/gtest/test_transfer.cpp
```diff
@@ -257,9 +269,63 @@ class TestTransfer : public testing::TestWithParam<std::string> {
         registerMem(agent, out, mem_type);
     }
 
+    void verifyNotifs(nixlAgent &agent, const std::string &from_name, size_t expected_count)
+    {
+        nixl_notifs_t notif_map;
+        int backoff_ms = 1;
+
+        for (int i = 0; i < 10; i++) {
```
yosefe: maybe define absolute timeout for example 60 seconds?

test/gtest/test_transfer.cpp
yosefe: shall we also test 1 and 2 threads?

test/gtest/test_transfer.cpp
```diff
@@ -269,36 +335,53 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+        for (auto& thread : threads) {
+            thread.join();
         }
-        auto total_time = absl::ToDoubleSeconds(absl::Now() - start_time);
 
-        auto total_size = size * count * repeat;
+        verifyNotifs(to, from_name, repeat * num_threads);
+
+        auto total_time = absl::ToDoubleSeconds(absl::Now() - start_time);
+        auto total_size = size * count * repeat * num_threads;
```
yosefe: IMO better post separate time per each thread since this also measures thread create and xfer prepare

test/gtest/test_transfer.cpp
w1ldptr: Absolute timeout if defined by Gtest itself (you may have noticed it fails when tests are not completed in 30 seconds by default). It is also much more convenient to change it as just a Gtest CLI parameter instead of recompiling.

test/gtest/test_transfer.cpp
```diff
@@ -269,36 +335,53 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+                        NIXL_WRITE,
+                        makeDescList<nixlBasicDesc>(src_buffers, src_mem_type),
+                        makeDescList<nixlBasicDesc>(dst_buffers, dst_mem_type), to_name,
+                        xfer_req, &extra_params);
+                ASSERT_EQ(status, NIXL_SUCCESS);
+                EXPECT_NE(xfer_req, nullptr);
+
+                for (size_t i = 0; i < repeat; i++) {
+                    status = from.postXferReq(xfer_req);
+                    ASSERT_GE(status, NIXL_SUCCESS);
```
```suggestion
                    ASSERT_TRUE((status == NIXL_SUCCESS) || (status == NIXL_IN_PROG));
```

test/gtest/test_transfer.cpp
w1ldptr: Let's have the infrastructure merged and then discuss the optimal per-test setting. Pinpointing good default fits more into the parallel "quick mode" PR.

test/gtest/test_transfer.cpp
w1ldptr: Why are we measuring execution time at all here? This is not a performance test and is only compiled in debug config which results sub-optimal performance anyway.

test/gtest/test_transfer.cpp
yosefe: so why limit to 10 iterations? and have this backoff?

test/gtest/test_transfer.cpp
```diff
@@ -269,36 +335,53 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
-            EXPECT_EQ(status, NIXL_SUCCESS);
+                    bool xfer_done = false;
+                    do {
+                        status = from.getXferStatus(xfer_req);
+                        EXPECT_TRUE((status == NIXL_SUCCESS) || (status == NIXL_IN_PROG));
+                        xfer_done = (status == NIXL_SUCCESS);
+                    } while (!xfer_done);
+
+                    status = from.getXferStatus(xfer_req);
+                    EXPECT_EQ(status, NIXL_SUCCESS);
```
rakhmets: Why do we need an extra check here?

test/gtest/test_transfer.cpp
yosefe: it's not question of best perf, but testing coverage. before this pr we tested xfer with 1 thread, now we don't

test/gtest/test_transfer.cpp
w1ldptr: Arbitrary value, should be enough to ensure that progress thread received all pending notifications and that we also don't block the execution for long.

test/gtest/test_transfer.cpp
```diff
@@ -257,9 +269,63 @@ class TestTransfer : public testing::TestWithParam<std::string> {
         registerMem(agent, out, mem_type);
     }
 
+    void verifyNotifs(nixlAgent &agent, const std::string &from_name, size_t expected_count)
+    {
+        nixl_notifs_t notif_map;
+        int backoff_ms = 1;
```
```suggestion
        std::chrono::milliseconds backoff{1};
```

test/gtest/test_transfer.cpp
w1ldptr: Ok, so to answer your question "shall we also test 1 and 2 threads?" - I don't see much benefit in it. Tell me if you think it is necessary and I'll add more configs.

test/gtest/test_transfer.cpp
w1ldptr: Refactored to simple loop of 10 iteration with 100ms sleep.

test/gtest/test_transfer.cpp
w1ldptr: Changed test cases to run with different number of threads to keep it fast and improve coverage.

test/gtest/test_transfer.cpp
w1ldptr: done

test/gtest/test_transfer.cpp
w1ldptr: removed

test/gtest/test_transfer.cpp
tvegas1: maybe not that important but if test started to fail, first thing would be to try with one thread.

test/gtest/test_transfer.cpp
```diff
@@ -257,47 +269,114 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+
+                auto start_time = absl::Now();
+
+                for (size_t i = 0; i < repeat; i++) {
+                    status = from.postXferReq(xfer_req);
+                    ASSERT_TRUE((status == NIXL_SUCCESS) || (status == NIXL_IN_PROG));
+
+                    bool xfer_done = false;
+                    do {
+                        status = from.getXferStatus(xfer_req);
```
tvegas1, line 355: we can use wait_until_true() instead of this while (you initially suggested rightfully)

w1ldptr: What is wait_until_true()?

test/gtest/test_transfer.cpp
w1ldptr: First test is the one which can actually get some decent concurrent transfers because it executes many requests. Not sure I'm following why you want that specific one to be single-threaded, but let me perhaps reorder the tests instead?

test/gtest/test_transfer.cpp
tvegas1: my bad, it is not merged yet.. I am referring to https://github.com/ai-dynamo/nixl/pull/356/files where we had similar discussion

test/gtest/test_transfer.cpp
tvegas1: this was addressed along with comment below

test/gtest/test_transfer.cpp
w1ldptr: Not sure if wait_until_true() is perfect fit here. In this case I want to assert as soon as getXferStatus returns an error (i.e. any status not IN_PROG or SUCCESS). In your case you just want to spin on checkRemoteMD() returning NOT_FOUND error until it eventually returns SUCCESS or times out.

tvegas1: is there a way to make sure we don't risk infinite loop?

yosefe: maybe add hard timeout

w1ldptr: Extracted timeout and number of retries to be shared for both transfer and notifications tests. Set their values to 1ms and 1000x which results fastest execution time on my setup.

Issue #392, opened 2025-05-27
Nixlbench cannot execute from container anymore


ovidiusm:
PR https://github.com/ai-dynamo/nixl/pull/354 broke running nixlbench from the container:

```
 docker run --privileged --device=/dev/infiniband --net=host --ipc=host --gpus all -e NVIDIA_VISIBLE_DEVICES=all -e NVIDIA_VISIBLE_DEVICES=all --rm -ti $IMG ./nixlbench --etcd-endpoints http://$SERVER:2379 --backend UCX --initiator_seg_type VRAM


==========
== CUDA ==
==========

NVIDIA Release  (build )
CUDA Version 12.8.1.012
Container image Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

Various files include modifications (c) NVIDIA CORPORATION & AFFILIATES.  All rights reserved.

GOVERNING TERMS: The software and materials are governed by the NVIDIA Software License Agreement
(found at https://www.nvidia.com/en-us/agreements/enterprise-software/nvidia-software-license-agreement/)
and the Product-Specific Terms for NVIDIA AI Products
(found at https://www.nvidia.com/en-us/agreements/enterprise-software/product-specific-terms-for-ai-products/).

NOTE: CUDA Forward Compatibility mode ENABLED.
  Using CUDA 12.8 driver version 570.124.06 with kernel driver version 560.35.03.
  See https://docs.nvidia.com/deploy/cuda-compatibility/ for details.

/opt/nvidia/nvidia_entrypoint.sh: line 55: /workspace/nixl/benchmark/kvbench/nixlbench: No such file or directory
```

This is due to changing the latest WORKDIR. Let's please fix that. The container is used for verification and performance regression testing.

aranadive:
@cheese-head for viz.

aranadive:
@ovidiusm - TBF, I think the entire path to nixlbench exec should be provided in the run command. Or atleast cd into the right directory before executing. You have #393 that should fix this. Thanks!

cheese-head:
@ovidiusm - The nixlbench binary is added to the PATH environment here. https://github.com/ai-dynamo/nixl/blob/b1c22edd8fe10e2e5221c107ee51200fce6f09a8/benchmark/nixlbench/contrib/Dockerfile#L133

cheese-head:
@aranadive - I can also package/add kvbench to the PATH environment variable

PR #393, opened 2025-05-27
Fixes for nixlbench container on x86 and arm

ovidiusm:
Make it possible to run nixlbench container on arm

Revert change to WORKDIR from PR #354

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -19,6 +19,7 @@ ARG BASE_IMAGE_TAG="25.03-cuda12.8-devel-ubuntu24.04"
 # UCX argument is either "upstream" (default installed in base image) or "custom" (build from source)
 ARG UCX="upstream"
 ARG DEFAULT_PYTHON_VERSION="3.12"
+ARG TARGETARCH="x86_64"
 
 # --- Stage 1: Common OS setup ---
 FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} AS os_setup_stage
@@ -76,6 +77,7 @@ FROM ucx_${UCX}_image AS ucx_image
 ARG DEFAULT_PYTHON_VERSION
 ARG WHL_PYTHON_VERSIONS="3.12"
 ARG WHL_PLATFORM="manylinux_2_39_x86_64"
+ARG TARGETARCH="x86_64"
 
 WORKDIR /workspace
 RUN git clone https://github.com/etcd-cpp-apiv3/etcd-cpp-apiv3.git &&\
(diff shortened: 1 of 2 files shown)


ovidiusm:
The only thing that I could not figure out is why `uv run meson setup build -Dgds_path=...` has no effect when executed as part of docker build :/

From an interactive shell inside the container it works fine

aranadive:
closes #392

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -97,16 +99,18 @@ RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
     # pybind11 pip install needed for ubuntu 22.04
     uv pip install --upgrade meson pybind11 patchelf pyYAML
 
-RUN rm -rf build && \
+RUN GDS_PATH=$(find /usr/local/cuda/targets -name '*-linux' | head -n 1) && \
+    sed -i "s|option('gds_path'.*)|option('gds_path', type: 'string', value: '$GDS_PATH', description: 'Path to GDS CuFile install')|g" meson_options.txt && \
```
aranadive, line 103: The other way to fix it is to modify the meson.build file to find the right gds path. Maybe this is an okay workaround for the fix right now. But ideally I think we can fix it in meson.build. Im working on adding that now.

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -97,16 +99,18 @@ RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
...
+    rm -rf build && \
     mkdir build && \
-    uv run meson setup build/ --prefix=/usr/local/nixl && \
+    uv run -- meson setup build --prefix=/usr/local/nixl && \
```
aranadive: Why is the -- required?

ovidiusm: It did not make any difference, can be removed

benchmark/nixlbench/contrib/Dockerfile
ovidiusm: That would be great! Thanks!

benchmark/nixlbench/contrib/Dockerfile
aranadive: Lets remove and merge if poss since this may need to be cherry-picked.

aranadive (approved):

benchmark/nixlbench/contrib/Dockerfile
ovidiusm: Done

aranadive (approved):

iyastreb:
> The only thing that I could not figure out is why `uv run meson setup build -Dgds_path=...` has no effect when executed as part of docker build :/
> 
> From an interactive shell inside the container it works fine

I noticed the same. I think the difference may come from the current working dir. When I run `uv run` from nixl dir, then it takes arguments for meson from `pyproject.toml` file and ignores options that I pass

benchmark/nixlbench/contrib/build.sh
```diff
@@ -35,9 +35,22 @@ fi
...
-WHL_PLATFORM=manylinux_2_39_x86_64
 WHL_PYTHON_VERSIONS="3.12"
 OS="ubuntu24"
+ARCH=$(uname -m)
+case "$ARCH" in
+    x86_64)
+        WHL_PLATFORM="manylinux_2_39_x86_64"
+        ;;
+    aarch64)
+        WHL_PLATFORM="manylinux_2_39_aarch64"
```
yosefe, line 46: maybe 
WHL_PLATFORM="manylinux_2_39_${ARCH}"

benchmark/nixlbench/contrib/Dockerfile
yosefe: if we decide to keep gds detection here - can we simplify/break this one liner to several parts?

iyastreb: Fixed in https://github.com/ai-dynamo/nixl/pull/407 by detecting gds path in meson.build

benchmark/nixlbench/contrib/build.sh
iyastreb: Fixed in https://github.com/ai-dynamo/nixl/pull/407

PR #395, opened 2025-05-27
CONTRIB: Enable UCX build for EFA with common sha1

tvegas1:
Add `--efa` option to build container with UCX with EFA SRD support.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -19,6 +19,7 @@ ARG BASE_IMAGE_TAG="25.03-cuda12.8-devel-ubuntu24.04"
 FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG}
 
 ARG DEFAULT_PYTHON_VERSION="3.12"
+ARG UCX_REF="v1.19.x"
 
 RUN apt-get update -y && \
     DEBIAN_FRONTEND=noninteractive apt-get -y install \
@@ -84,7 +85,7 @@ RUN rm -rf /opt/hpcx/ucx
 RUN cd /usr/local/src && \
      git clone https://github.com/openucx/ucx.git && \
      cd ucx && 			     \
-     git checkout v1.19.x &&	     \
+     git checkout $UCX_REF &&        \
      ./autogen.sh && ./configure     \
          --enable-shared             \
          --disable-static            \
--- a/contrib/build-container.sh
+++ b/contrib/build-container.sh
@@ -32,6 +32,7 @@ BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base
 BASE_IMAGE_TAG=25.03-cuda12.8-devel-ubuntu24.04
 WHL_PLATFORM=manylinux_2_39_x86_64
 WHL_PYTHON_VERSIONS="3.12"
+UCX_REF=v1.19.x
 OS="ubuntu24"
 
 get_options() {
@@ -84,6 +85,10 @@ get_options() {
                 missing_requirement $1
             fi
             ;;
+        --efa)
+            # Master branch (v1.20) containing EFA SRD support
+            UCX_REF=3842a65f02a8e2f3635a66189cd82dc86bf065c2
+            ;;
         --)
             shift
             break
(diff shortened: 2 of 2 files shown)


contrib/Dockerfile
```diff
@@ -19,6 +19,7 @@ ARG BASE_IMAGE_TAG="25.03-cuda12.8-devel-ubuntu24.04"
 FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG}
 
 ARG DEFAULT_PYTHON_VERSION="3.12"
+ARG UCX_REF="v1.19.x"
```
aranadive, line 23: Maybe the EFA REF should be default for everything and allow user to select a UCX_REF in command line if required?

contrib/build-container.sh
```diff
@@ -84,6 +85,10 @@ get_options() {
                 missing_requirement $1
             fi
             ;;
+        --efa)
```
brminich: shall we also remove `--with-efa` from ucx configure options in docker file is this option is not provided?

tvegas1: i'd rather not, since efa might still be bundled and this --efa option is not going to persist across many releases.

contrib/Dockerfile
tvegas1: i am under understanding that we want to use v1.19.x by default (validation), and to allow to build container with efa support with --efa.

contrib/build-container.sh
brminich: can you pls elaborate? Configure will fail with the following error if no EFA in env
`configure: error: EFA device support requested but libefa or EFA headers are not found`

yosefe: maybe instead of calling it '--efa' we can call it '--upstream' or '--main' to mean we use latest stable main branch?

tvegas1: > can you pls elaborate? Configure will fail with the following error if no EFA in env `configure: error: EFA device support requested but libefa or EFA headers are not found`

in practice efa is there in rdma-core-devel when building this container (unless badly mistaken), and I meant that even if we remove --with-efa, efa can still be detected and built (and not used when not applicable).

> maybe instead of calling it '--efa' we can call it '--upstream' or '--main' to mean we use latest stable main branch?

--efa was to make sure we have stable sha1, but --upstream is another choice, you prefer this?

contrib/Dockerfile
```diff
@@ -84,7 +85,7 @@ RUN rm -rf /opt/hpcx/ucx
 RUN cd /usr/local/src && \
      git clone https://github.com/openucx/ucx.git && \
      cd ucx && 			     \
-     git checkout v1.19.x &&	     \
+     git checkout $UCX_REF &&        \
```
brminich, line 93: i meant do we need to remove `--with-efa` here or make it conditional?

tvegas1: i'd rather not, since efa might still be bundled and this --efa option is not going to persist across many releases.

tvegas1: and we do not need to remove it as far as i understand it, I detailed in the comment above.

contrib/build-container.sh
yosefe: yes because there may be additional reasons to take upstream version, but on other hand we don't want to pick up automatically changes in ucx code.

yosefe: > > can you pls elaborate? Configure will fail with the following error if no EFA in env `configure: error: EFA device support requested but libefa or EFA headers are not found`
> 
> in practice efa is there in rdma-core-devel when building this container (unless badly mistaken), and I meant that even if we remove --with-efa, efa can still be detected and built (and not used when not applicable).

Since it's there already and works because it's in rdma-core i don't think we need to remove it in this pr

tvegas1: ok than change is only to rename `--efa` to `--upstream`.

aranadive (approved):

contrib/build-container.sh
```diff
@@ -84,6 +85,10 @@ get_options() {
                 missing_requirement $1
             fi
             ;;
+        --ucx-upstream)
+            # Master branch (v1.20) also containing EFA SRD support
+            UCX_REF=3842a65f02a8e2f3635a66189cd82dc86bf065c2
```
brminich: maybe we will need to change it depending on the ongoing debug

yosefe: yes

yosefe (approved):

aranadive:
@tvegas1 can we merge?

tvegas1:
pending the merge of https://github.com/openucx/ucx/pull/10704

tvegas1:
ready for merge

aranadive (approved):

PR #396, opened 2025-05-28
UCX: Refactor nixlUcxWorker (and some related code).

ColinNV:


src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,79 @@ nixlUcxContext::~nixlUcxContext()
...
+    NIXL_FATAL << "Invalid UCX worker type: " << static_cast< std::underlying_type_t< nixl_ucx_mt_t > >( t );
+    std::terminate();
+  }
+
+  struct nixlUcpWorkerParams
+    : ucp_worker_params_t
+  {
+    explicit nixlUcpWorkerParams( const nixl_ucx_mt_t t )
+    {
+      std::memset( this, 0, sizeof( ucp_worker_params_t ) );  // TODO: Needed?
```
rakhmets: Can be removed since the field mask is set explicitly.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,79 @@ nixlUcxContext::~nixlUcxContext()
...
 
-nixlUcxWorker::~nixlUcxWorker()
+}  // namespace
+
+ucp_worker* nixlUcxWorker::createUcpWorker( const std::shared_ptr< nixlUcxContext > &ctx )
 {
-    ucp_worker_destroy(worker);
+  ucp_worker* result = nullptr;
+  const nixlUcpWorkerParams params( ctx->mt_type );
+  const ucs_status_t status = ucp_worker_create(ctx->ctx, &params, &result);
```
rakhmets: Maybe use auto instead of ucs_status_t here and in other places?

src/utils/ucx/ucx_utils.cpp
ColinNV: Thanks for confirming, will remove.

src/utils/ucx/ucx_utils.cpp
ColinNV: When the type is short and/or spelling it out increases readability I prefer to not use `auto`...

rakhmets: The reason I prefer to use `auto` wherever possible is that it doesn't leave any room for interpretation. Otherwise, different contributors uses `auto` differently. Of course, this is not a strict remark, it is more a topic for reflection for now.

ColinNV: Some years ago I briefly tried "use `auto` wherever possible" and found that it can negatively impact readability; now I use it where I feel that it _increases_ readability and/or simplifies code, e.g. when knee-deep in templates and you don't immediately know the type, or the typical result of `std::map::find`.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -75,9 +78,9 @@ void nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status)
             ucp_request_free(request);
         }
         return;
-    default:
-        NIXL_FATAL << "Invalid endpoint state: " << state;
     }
+    NIXL_FATAL << "Invalid endpoint state: " << state;
```
iyastreb, line 84: IMO we can add std::terminate to NIXL_FATAL macro, since it's always supposed to terminate the program:
```
/*
 * Logs a message and terminates the program unconditionally.
 * Maps to Abseil LOG(FATAL). Use for unrecoverable errors.
 */
#define NIXL_FATAL LOG(FATAL) // add std::terminate

File: src/utils/ucx/ucx_utils.cpp [hunk 4]
```diff
@@ -312,23 +315,21 @@ nixl_status_t nixlUcxEp::flushEp(nixlUcxReq &req)
     return ucx_status_to_nixl(UCS_PTR_STATUS(request));
 }
 
-bool nixlUcxContext::mtLevelIsSupproted(nixl_ucx_mt_t mt_type)
+bool nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t mt_type) noexcept
```
iyastreb, line 320: minor: personally I prefer not to pass primitives by const - it makes code longer without clear benefits

src/utils/ucx/ucx_utils.cpp
```diff
@@ -312,23 +315,21 @@ nixl_status_t nixlUcxEp::flushEp(nixlUcxReq &req)
...
-        assert(mt_type < NIXL_UCX_MT_MAX);
-        abort();
+    case nixl_ucx_mt_t::SINGLE:
+        return attr.max_thread_level >= UCS_THREAD_MODE_SERIALIZED;
+    case nixl_ucx_mt_t::CTX:
+    case nixl_ucx_mt_t::WORKER:
+        return attr.max_thread_level >= UCS_THREAD_MODE_MULTI;
     }
-    return false;
+    assert(!"invalid mt type");
```
iyastreb: NIXL_FATAL?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
-        worker_params.thread_mode = UCS_THREAD_MODE_SERIALIZED;
-        break;
-    case NIXL_UCX_MT_WORKER:
-        worker_params.thread_mode = UCS_THREAD_MODE_MULTI;
-        break;
-    default:
-        NIXL_FATAL << "Invalid UCX worker type: " << ctx->mt_type;
+  [[nodiscard]] ucs_thread_mode_t toUcsThreadModeChecked( const nixl_ucx_mt_t t )
+  {
+    switch( t ) {
```
iyastreb: maybe remove extra spaces to have it consistent with the rest of the code
```suggestion
  [[nodiscard]] ucs_thread_mode_t toUcsThreadModeChecked(nixl_ucx_mt_t t)
  {
    switch (t) {
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
-        NIXL_ERROR << err_str;
-        throw std::runtime_error(err_str);
+    NIXL_FATAL << "Invalid UCX worker type: " << static_cast< std::underlying_type_t< nixl_ucx_mt_t > >( t );
+    std::terminate();
+  }
+
+  struct nixlUcpWorkerParams
+    : ucp_worker_params_t
+  {
+    explicit nixlUcpWorkerParams( const nixl_ucx_mt_t t )
```
iyastreb: I don't see in other places these extra spaces in function declaration
Maybe remove them for a consistent look?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
+    {
```
iyastreb: alignment is 4 spaces

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
+      field_mask = UCP_WORKER_PARAM_FIELD_THREAD_MODE;
```
iyastreb: minor: both member fields can be initialized with ctor init list

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
     }
-}
+  };
+
+  static_assert( sizeof( nixlUcpWorkerParams ) == sizeof( ucp_worker_params_t ) );
 
-nixlUcxWorker::~nixlUcxWorker()
+}  // namespace
+
+ucp_worker* nixlUcxWorker::createUcpWorker( const std::shared_ptr< nixlUcxContext > &ctx )
```
iyastreb: I think here we can pass context by reference

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
+  ucp_worker* result = nullptr;
+  const nixlUcpWorkerParams params( ctx->mt_type );
+  const ucs_status_t status = ucp_worker_create(ctx->ctx, &params, &result);
+  if( status != UCS_OK ) {
+    const auto err_str = std::string( "Failed to create UCX worker: " ) +
+                         ucs_status_string( status );
+    NIXL_ERROR << err_str;
+    throw std::runtime_error( err_str );  // result is nullptr -- no leak
+  }
+  return result;
```
iyastreb: maybe name it worker
why not to return std::unique_ptr?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,78 @@ nixlUcxContext::~nixlUcxContext()
...
 }
 
-std::unique_ptr<char []> nixlUcxWorker::epAddr(size_t &size)
+nixlUcxWorker::nixlUcxWorker( const std::shared_ptr< nixlUcxContext > &_ctx )
+  : ctx(_ctx),
+    worker( createUcpWorker( ctx ), &ucp_worker_destroy )
+{}
+
+std::unique_ptr<char []> nixlUcxWorker::epAddr(std::size_t &size)
```
iyastreb: why not std::string?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -544,29 +552,23 @@ int nixlUcxWorker::regAmCallback(unsigned msg_id, ucp_am_recv_callback_t cb, voi
...
+    const ucs_status_ptr_t status = ucp_am_recv_data_nbx(worker.get(), data_desc, buffer, len, param);
+    if(UCS_PTR_IS_ERR(status)) {
         //TODO: error handling
-        return -1;
+        return nullptr;
     }
-    req = (void*)status;
-
-    return 0;
+    return static_cast< nixlUcxReq >( status );;
```
iyastreb: extra ;

src/utils/ucx/ucx_utils.cpp
```diff
@@ -575,16 +577,16 @@ int nixlUcxWorker::getRndvData(void* data_desc, void* buffer, size_t len, const
 
 int nixlUcxWorker::progress()
 {
-    return ucp_worker_progress(worker);
+  return ucp_worker_progress(worker.get());
```
iyastreb, line 565: alignment

src/utils/ucx/ucx_utils.cpp
brminich: 👍, even though the code will be aligned with clang eventually better to use consistent syntax

src/utils/ucx/ucx_utils.cpp
```diff
@@ -544,29 +552,23 @@ int nixlUcxWorker::regAmCallback(unsigned msg_id, ucp_am_recv_callback_t cb, voi
...
-    {
+    if(status != UCS_OK) {
         //TODO: error handling
         return -1;
     }
     return 0;
 }
 
-int nixlUcxWorker::getRndvData(void* data_desc, void* buffer, size_t len, const ucp_request_param_t *param, nixlUcxReq &req)
+nixlUcxReq nixlUcxWorker::getRndvData(void* data_desc, void* buffer, const std::size_t len, const ucp_request_param_t *param)
```
brminich: it is not used, maybe remove it?

src/utils/ucx/ucx_utils.cpp
ColinNV: Sounds ok, but I think this PR should stick to UCX related parts, I'll do it separately.

src/utils/ucx/ucx_utils.cpp
ColinNV: One could argue that `const` should have be the default, given as that is not posssible the only thing we can do is plaster `const` everywhere possible. At that point, an absence of `const` turns into "attention, this variable will change". In other words, I prefer to declare all local variable `const` when possible, including function arguments.

src/utils/ucx/ucx_utils.cpp
ColinNV: Ok

src/utils/ucx/ucx_utils.cpp
ColinNV: Ok

src/utils/ucx/ucx_utils.cpp
ColinNV: Unfortunately not as the fields are inherited from `ucp_worker_params_t`.

src/utils/ucx/ucx_utils.cpp
ColinNV: Since `ucp_worker_create` needs a mutable reference to the "returned" pointer an intermediate step is required and in this case I simply aligned it with the function boundary...

src/utils/ucx/ucx_utils.cpp
ColinNV: Good question, looks like a `std::string` works, wonder why somebody used `std::unique_ptr`.

src/utils/ucx/ucx_utils.cpp
ColinNV: The test still uses it...

src/utils/ucx/ucx_utils.cpp
```diff
@@ -312,23 +315,21 @@ nixl_status_t nixlUcxEp::flushEp(nixlUcxReq &req)
...
-        assert(mt_type < NIXL_UCX_MT_MAX);
-        abort();
+    case nixl_ucx_mt_t::SINGLE:
+        return attr.max_thread_level >= UCS_THREAD_MODE_SERIALIZED;
+    case nixl_ucx_mt_t::CTX:
+    case nixl_ucx_mt_t::WORKER:
+        return attr.max_thread_level >= UCS_THREAD_MODE_MULTI;
     }
-    return false;
+    NIXL_FATAL << "invalid mt type";
```
brminich: pls also print `mt type`

src/utils/ucx/ucx_utils.cpp
brminich: we can remove the test as well

src/utils/ucx/ucx_utils.h
```diff
@@ -14,27 +14,39 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef __UCX_UTILS_H
-#define __UCX_UTILS_H
+#ifndef NIXL_SRC_UTILX_UCX_UCX_UTILS_H
```
brminich: did you mean something like
```suggestion
#ifndef NIXL_SRC_UTILS_UCX_UTILS_H
```

src/utils/ucx/ucx_utils.h
```diff
@@ -137,44 +149,50 @@ class nixlUcxContext {
...
 
+[[nodiscard]] bool nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t) noexcept;
+
 class nixlUcxWorker {
 private:
     /* Local UCX stuff */
-    std::shared_ptr<nixlUcxContext> ctx;
-    ucp_worker_h worker;
+    const std::shared_ptr<nixlUcxContext> ctx;
+    const std::unique_ptr<ucp_worker, void(*)(ucp_worker*)> worker;
```
brminich, line 174: maybe?
```suggestion
    const std::unique_ptr<ucp_worker_t, void(*)(ucp_worker_t*)> worker;
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -653,7 +653,7 @@ nixl_status_t nixlUcxEngine::endConn(const std::string &remote_agent) {
 }
 
 nixl_status_t nixlUcxEngine::getConnInfo(std::string &str) const {
-    str = nixlSerDes::_bytesToString(workerAddr.get(), workerSize);
+    str = nixlSerDes::_bytesToString(workerAddr.data(), workerAddr.size());
```
brminich: seems it could be just `str = workerAddr` now
btw, maybe keep the original format of meta data to avoid extra string copies along the stack?

src/utils/ucx/ucx_utils.h
```diff
@@ -137,44 +149,50 @@ class nixlUcxContext {
...
     /* Memory management */
     int memReg(void *addr, size_t size, nixlUcxMem &mem);
-    std::unique_ptr<char []> packRkey(nixlUcxMem &mem, size_t &size);
+    [[nodiscard]] std::string packRkey(nixlUcxMem &mem);
     void memDereg(nixlUcxMem &mem);
 
     friend class nixlUcxWorker;
 };
 
+[[nodiscard]] bool nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t) noexcept;
```
brminich, line 168: what is the benefit of moving it outside? Maybe still define it as a context method and cache returned attrs in the context, to avoid doing query every time?

src/utils/ucx/ucx_utils.h
ColinNV: The `X` in `UTILX` was a typo, I meant `NIXL_SRC_UTILS_UCX_UCX_UTILS_H`; it's `ucx_utils.h` in a `ucx` directory.

ColinNV: I saw at least one header already use `#pragma once`.

src/utils/ucx/ucx_utils.h
ColinNV: Ok, if that's preferred.

src/utils/ucx/ucx_utils.h
ColinNV: I moved it outside because there was no reason for it to be part of the class. As for caching, the function is only called in the constructor of `nixlUcxEngine`, do we expect that to happen often enough?

brminich: maybe make it a member of `nixlUcxEngine` then?

ColinNV: But it doesn't need access to that class's internals either. Why not a free function?

src/utils/ucx/ucx_utils.h
ColinNV: Doesn't compile out-of-the-box and doesn't seem worth hunting down?

src/utils/ucx/ucx_utils.h
brminich: ok, just thought that making it a property of either context or engine class would be more clear.
can be a separate func though

src/utils/ucx/ucx_utils.h
brminich: ok

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -719,14 +719,14 @@ nixlUcxEngine::connectionTermAmCb (void *arg, const void *header,
...
 nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
-    struct nixl_ucx_am_hdr hdr;
-    uint32_t flags = 0;
+    nixl_ucx_am_hdr hdr;
+    std::uint32_t flags = 0;
 
     if (remote_agent == localAgent)
         return loadRemoteConnInfo (remote_agent,
-                   nixlSerDes::_bytesToString(workerAddr.get(), workerSize));
+            nixlSerDes::_bytesToString(workerAddr.data(), workerAddr.size()));
```
brminich: can it be just `workerAddr`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -719,14 +719,13 @@ nixlUcxEngine::connectionTermAmCb (void *arg, const void *header,
...
 
 nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
-    struct nixl_ucx_am_hdr hdr;
-    uint32_t flags = 0;
-
-    if (remote_agent == localAgent)
-        return loadRemoteConnInfo (remote_agent,
-                   nixlSerDes::_bytesToString(workerAddr.get(), workerSize));
+    nixl_ucx_am_hdr hdr;
+    std::uint32_t flags = 0;
```
iyastreb, line 723: minor: I doubt there is a benefit of using `std::` prefix for fixed-width integers and size_t. And since it's not used in other places I would vote for removing it here. Reason: less typing, easier to read

src/utils/ucx/ucx_utils.cpp
```diff
@@ -406,67 +407,74 @@ nixlUcxContext::~nixlUcxContext()
...
+std::string nixlUcxWorker::epAddr()
 {
     ucp_worker_attr_t wattr;
-    ucs_status_t status;
 
     wattr.field_mask = UCP_WORKER_ATTR_FIELD_ADDRESS;
-    status = ucp_worker_query(worker, &wattr);
+    const ucs_status_t status = ucp_worker_query(worker.get(), &wattr);
     if (UCS_OK != status) {
         // TODO: printf
```
iyastreb: print an error here?

src/plugins/ucx/ucx_backend.cpp
ColinNV: I try not to rely on C++ "features" that only exist because of how it evolved from C.

brminich (approved):

PR #398, opened 2025-05-28
nixlbench: enable VMM CUDA memory allocation

iyastreb:
This PR enables VMM CUDA memory allocation in NIXL to allow MNNVL communication

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -226,10 +226,10 @@ getVramDescCuda(int devid, size_t buffer_size, uint8_t memset_value)
     return std::optional<xferBenchIOV>(std::in_place, (uintptr_t)addr, buffer_size, devid);
 }
 
-#if HAVE_CUDA_FABRIC
 static std::optional<xferBenchIOV>
 getVramDescCudaVmm(int devid, size_t buffer_size, uint8_t memset_value)
 {
+#if HAVE_CUDA_FABRIC
     CUdeviceptr addr = 0;
     CUmemAllocationProp prop = {};
     CUmemAccessDesc access = {};
@@ -278,8 +278,12 @@ getVramDescCudaVmm(int devid, size_t buffer_size, uint8_t memset_value)
 
     return std::optional<xferBenchIOV>(std::in_place, (uintptr_t)addr, buffer_size,
                                        devid, padded_size, handle);
-}
+
+#else
+    std::cerr << "CUDA_FABRIC is not supported" << std::endl;
+    return std::nullopt;
 #endif /* HAVE_CUDA_FABRIC */
+}
 
 static std::optional<xferBenchIOV> getVramDesc(int devid, size_t buffer_size,
                                                bool isInit)
@@ -289,11 +293,10 @@ static std::optional<xferBenchIOV> getVramDesc(int devid, size_t buffer_size,
                                     XFERBENCH_TARGET_BUFFER_ELEMENT;
 
     if (xferBenchConfig::enable_vmm) {
-#if HAVE_CUDA_FABRIC
         return getVramDescCudaVmm(devid, buffer_size, memset_value);
-#endif
+    } else {
+        return getVramDescCuda(devid, buffer_size, memset_value);
     }
-    return getVramDescCuda(devid, buffer_size, memset_value);
 }
 
 std::optional<xferBenchIOV> xferBenchNixlWorker::initBasicDescVram(size_t buffer_size, int mem_dev_id) {
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -32,17 +32,11 @@
 #include <utils/serdes/serdes.h>
 #include <omp.h>
 
-#define USE_VMM 0
 #define ROUND_UP(value, granularity) ((((value) + (granularity) - 1) / (granularity)) * (granularity))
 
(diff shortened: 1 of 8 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -218,72 +212,81 @@ std::optional<xferBenchIOV> xferBenchNixlWorker::initBasicDescDram(size_t buffer
...
+getVramDescCudaVmm(int devid, size_t buffer_size, uint8_t memset_value)
+{
     CUdeviceptr addr = 0;
-    size_t granularity = 0;
     CUmemAllocationProp prop = {};
     CUmemAccessDesc access = {};
 
     prop.type = CU_MEM_ALLOCATION_TYPE_PINNED;
-    // prop.requestedHandleTypes = CU_MEM_HANDLE_TYPE_FABRIC;
+    prop.requestedHandleTypes = CU_MEM_HANDLE_TYPE_FABRIC;
```
brminich, line 238: shall we add compile time check for cuda version to make sure `CU_MEM_HANDLE_TYPE_FABRIC` is available?

iyastreb: done

iyastreb:
> do you think we can add any type of test to CI? (even just compilation)

I think compilation should be covered by existing tests.
And functional test would be harder to add..

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
brminich: can we check for particular symbol presence rather than CUDA version?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -222,72 +216,88 @@ std::optional<xferBenchIOV> xferBenchNixlWorker::initBasicDescDram(size_t buffer
...
+{
+    CHECK_CUDA_ERROR(cudaSetDevice(devid), "Failed to set device");
+    uint8_t memset_value = isInit ? XFERBENCH_INITIATOR_BUFFER_ELEMENT :
+                                    XFERBENCH_TARGET_BUFFER_ELEMENT;
+
+    if (xferBenchConfig::enable_vmm) {
+#if CUDA_VERSION >= 12020
+        return getVramDescCudaVmm(devid, buffer_size, memset_value);
+#else
+        std::cerr << "VMM is not supported in CUDA version " << CUDA_VERSION << std::endl;
```
rakhmets: Maybe we shouldn't add the benchmark option when we build with CUDA <12.2.
Or we should fail earlier, e.g. during benchmark options validation.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -376,15 +386,18 @@ void xferBenchNixlWorker::cleanupBasicDescDram(xferBenchIOV &iov) {
...
+        CHECK_CUDA_DRIVER_ERROR(cuMemUnmap(iov.addr, iov.len),
+                                "Failed to unmap memory");
+        CHECK_CUDA_DRIVER_ERROR(cuMemRelease(iov.handle),
+                                "Failed to release memory");
+        CHECK_CUDA_DRIVER_ERROR(cuMemAddressFree(iov.addr, iov.padded_size),
+                                "Failed to free reserved address");
+    } else {
+        CHECK_CUDA_ERROR(cudaFree((void *)iov.addr),
+                                  "Failed to deallocate CUDA buffer");
+    }
```
rakhmets, line 399: Please consider introducing an object, that will allocate memory in ctor, and free in dtor.

iyastreb: That was my initial intention, but then it requires quite heavy refactoring, so I decided to not do in in this PR

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
iyastreb: done

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
iyastreb: I moved config validation to `loadFromFlags` function that initializes config.
wrt excluding this option - it would be larger change (including multiple files) and I'm not sure about added value. I think error message is good enough.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -222,72 +216,84 @@ std::optional<xferBenchIOV> xferBenchNixlWorker::initBasicDescDram(size_t buffer
...
+                                               bool isInit)
+{
+    CHECK_CUDA_ERROR(cudaSetDevice(devid), "Failed to set device");
+    uint8_t memset_value = isInit ? XFERBENCH_INITIATOR_BUFFER_ELEMENT :
+                                    XFERBENCH_TARGET_BUFFER_ELEMENT;
+
+    if (xferBenchConfig::enable_vmm) {
+#if HAVE_CUDA_FABRIC
+        return getVramDescCudaVmm(devid, buffer_size, memset_value);
+#endif
```
brminich: maybe now move this check inside `getVramDescCudaVmm` and return some error code from there in case it is not defined?

iyastreb: done

brminich (approved):

aranadive (approved):

PR #403, opened 2025-05-29
Support Mooncake backend in nixlbench

Pyngon:


Pyngon:
@aranadive the ci failed with

```
TEST FAILED: Dynamic Plugins are still loaded.
Loaded plugins after unload:
 - MOCK_BACKEND
```

It does not look like it is related to this patch, do you have idea about this error?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -758,15 +766,16 @@ void xferBenchNixlWorker::poll(size_t block_size) {
     total_iter = skip + num_iter;
 
     /* Ensure warmup is done*/
-    while (skip != int(notifs["initiator"].size())) {
-        agent->getNotifs(notifs);
-    }
+    do {
+        status = agent->getNotifs(notifs);
+    } while(skip != int(notifs["initiator"].size()) && status == NIXL_SUCCESS);
```
yosefe: don't we need to exit is status!=SUCCESS and not start "polling for actual iterations"?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -249,9 +256,15 @@ nixl_status_t nixlMooncakeEngine::postXfer (const nixl_xfer_op_t &operation,
         request[index].length = local[index].len;
         request[index].target_id = segment_id;
     }
+
+    // TODO: submitTransfer could not be called multiple times for
+    // the same batch_id, it will fail when users call createXferReq
+    // once and reuse the handle and call postXferReq multiple times.
```
yosefe: if this is the case do we really need to keep priv->batch_id field?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -268,13 +281,22 @@ nixl_status_t nixlMooncakeEngine::checkXfer (nixlBackendReqH* handle) const
         else if (status.status == STATUS_PENDING || status.status == STATUS_WAITING)
             return NIXL_IN_PROG;
     }
+    if (!has_failed) {
+        // workaround the issue where submitTranfer could not be called
+        // multiple times with the same batch id. Free the batchID
+        // here and allocate a new batchID in postXfer.
+        freeBatchID(engine_, priv->batch_id);
+        priv->batch_id = INVALID_BATCH;
```
yosefe: is it bug in mooncake backend or expected behavior?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: ya, should check the status first. Will fix it in next commit.

src/plugins/mooncake/mooncake_backend.cpp
Pyngon: i think we still need that so that we can check the status in checkXfer and free it.

src/plugins/mooncake/mooncake_backend.cpp
Pyngon: Not an expert here, but the error code return from submitTransfer is 2 which looks like it mean TooManyRequest.
(https://github.com/kvcache-ai/Mooncake/blob/f09c501b2a5d73e4d60cdeb612d7d0d54e1ec228/mooncake-transfer-engine/include/common/base/status.h#L36).

Need to read further to understand whether it is bug or expected behavior.

Pyngon: Looked further into the code, mooncake backend will specify the number of requests when allocating the batch_id, https://github.com/ai-dynamo/nixl/blob/main/src/plugins/mooncake/mooncake_backend.cpp#L215

So it will failed when we called the submitTransfer multiple times and exceed the total number of requests.

I have also tried to specify a large size when allocating the batch_id, but the performance is very bad when there is a lot of requests under one batch id. And an additional issue here is the requests in one batch could not exceed 1024 when using Mooncake backend. 

I will update the comment in the PR with this additional information. Do we need further discussion with the Mooncake backend owner to find a solution to get around this limitation?

yosefe: Yes, or get an approval from Mooncake backend owner that this is the right solution

Pyngon: @alogfans Could you help to review this PR regarding the issues mentioned above and see if the current change is good? Thanks.

This patch mainly release the batch_id early in CheckXfer when the transfer is done and allocate a new one when PostXfer is called. I think it shouldn't introduce new issue or limitation to the existing code.

alogfans: @Pyngon I believe such modification won't change the correctness. If it is benefical for performance, I agree with it.

Pyngon: @alogfans The performance should be the same as the existing code. The change is mainly to allow nixlbench to benchmark Mooncake, as nixlbench will call the prepXfer once and use the same nixlBackendH to call PostXfer and CheckXfer multiple times, existing code won't work with that.

If this looks fine to you, please give me an approval for check in. Thanks.

Pyngon:
@aranadive Could you help to run the CI and merge it if this looks good to you?

aranadive:
@Pyngon Please use clang-format on the diffs. After rebase you can use git clang-format HEAD~3 (for the number of commits you have).

Pyngon:
> @Pyngon Please use clang-format on the diffs. After rebase you can use git clang-format HEAD~3 (for the number of commits you have).

@aranadive Done. Please help to merge it if all is good. Thanks.

PR #407, opened 2025-05-30
nixlbench: Add deps for IB devices

iyastreb:
This PR incorporates a few fixes:
- ARM64 fix and GDS path fix is done in https://github.com/ai-dynamo/nixl/pull/408
- Install required dependencies for IB in nixlbench Dockerfile

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -52,7 +52,8 @@ RUN echo "INFO: Starting custom UCX build..." && \
     apt-get update -y && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
         autoconf automake libtool pkg-config make g++ \
-        libnuma-dev librdmacm-dev ibverbs-providers && \
+        libnuma-dev librdmacm-dev ibverbs-providers libibverbs-dev rdma-core \
+        ibverbs-utils libibumad-dev && \
     echo "INFO: Removing pre-existing UCX installations..." && \
     rm -rf /usr/local/ucx /opt/hpcx/ucx && \
     cd /workspace/ucx && \
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -45,7 +45,8 @@ RUN apt-get update -y && \
     build-essential
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
-    --reinstall libibverbs-dev rdma-core ibverbs-utils libibumad-dev
+    --reinstall libibverbs-dev rdma-core ibverbs-utils libibumad-dev \
+    libnuma-dev librdmacm-dev ibverbs-providers
 
 WORKDIR /workspace
 RUN git clone https://github.com/etcd-cpp-apiv3/etcd-cpp-apiv3.git &&\


benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -76,11 +80,12 @@ FROM ucx_${UCX}_image AS ucx_image
 ARG DEFAULT_PYTHON_VERSION
 ARG WHL_PYTHON_VERSIONS="3.12"
 ARG WHL_PLATFORM="manylinux_2_39_x86_64"
+ARG TARGETARCH="x86_64"
```
brminich: it is already set on line 22

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -39,6 +40,9 @@ RUN apt-get update -y && \
     etcd-server \
     etcd-client
 
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y install --reinstall \
+    libibverbs-dev rdma-core ibverbs-utils libibumad-dev
```
brminich: maybe merge it with line 59

benchmark/nixlbench/contrib/Dockerfile
iyastreb: fixed in place

benchmark/nixlbench/contrib/Dockerfile
iyastreb: UCX can be either taken from the base image ("upstream" use case 2a), or built from source ("custom" use case 2b). I think we need to install these deps in both cases, not just when we build UCX from sources. But let me double check that

iyastreb: it's not needed, will move to line 59

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -132,4 +133,4 @@ RUN rm -rf build && \
     uv run meson setup build -Dnixl_path=/usr/local/nixl/ && \
     cd build && ninja
 ENV PATH=/workspace/nixlbench/build:$PATH
-WORKDIR /workspace/nixl/benchmark/kvbench
\ No newline at end of file
+WORKDIR /workspace/nixlbench/build
```
brminich: does it make sense to change,  given that this folder is added to PATH

benchmark/nixlbench/contrib/build.sh
```diff
@@ -195,5 +195,5 @@ BUILD_ARGS+=" --build-arg WHL_PLATFORM=$WHL_PLATFORM"
 BUILD_ARGS+=" --build-arg ARCH=$ARCH"
 
 show_build_options
-
+set -x
```
brminich: why needed

iyastreb: I took it from https://github.com/ai-dynamo/nixl/pull/393, useful for debugging

benchmark/nixlbench/contrib/Dockerfile
iyastreb: Right, now it works even without this change
I think it's still desirable to have this change for consistency: nixlbench pointing to kvbench seems a bit misleading. And also this fix was already approved by Adit, who introduced it

brminich (approved):

benchmark/nixlbench/contrib/Dockerfile
iyastreb: Ok, if we merge this change, then kvbench documentation needs to be updated also, because currently it relies on kvbench WORKDIR. So maybe I just remove this change

benchmark/nixlbench/contrib/build.sh
iyastreb: removed

brminich (approved):

PR #408, opened 2025-05-30
Build and container fixes for aarch64

w1ldptr:
Set default GDS path to point to CUDA install directory and use point lib and include to symlinks instead of hardcoding architecture-specific values.

Currently build scripts and Docker build files assume that the container is built on x86_64 Linux host (by not specifying a platform which makes docker default to host architecture, hardcoding 'x86_64' in several places, etc.), which makes it impossible to create either x86_64 or aarch64 container on Arm host or aarch64 container on x86 host.

Configure target architecture via 'ARCH' docker variable. Set it to x86 by default in docker files (for any users that use it directly) and to host architecture in build-container.sh and nixlbench build.sh. Allow user to specify ARCH value via '--arch' CLI parameter, set docker build platform value accordingly and pass the value to docker build as a build arg.

Fixes NIX-14

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -73,9 +73,10 @@ FROM ucx_${UCX}_image AS ucx_image
 
 # --- Stage 4: Final Image Assembly ---
 # Re-declare ARGs needed in this final stage
+ARG ARCH="x86_64"
 ARG DEFAULT_PYTHON_VERSION
 ARG WHL_PYTHON_VERSIONS="3.12"
-ARG WHL_PLATFORM="manylinux_2_39_x86_64"
+ARG WHL_PLATFORM="manylinux_2_39_$ARCH"
 
 WORKDIR /workspace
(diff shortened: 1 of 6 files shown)


iyastreb:
Few hours ago I fixed the same issue: https://github.com/ai-dynamo/nixl/pull/407

w1ldptr:
> Few hours ago I fixed the same issue: #407

@iyastreb @brminich  this also fixes nixl container build besides nixlbench, allows container crossbuild (i.e. allows user to build x86 in Arm and vice versa) and properly solves the GDS path issue by leveraging symlinks instead of continuing to rely on arch-specific paths.

iyastreb:
> @iyastreb @brminich this also fixes nixl container build besides nixlbench, allows container crossbuild (i.e. allows user to build x86 in Arm and vice versa) and properly solves the GDS path issue by leveraging symlinks instead of continuing to rely on arch-specific paths.

Sure, I like your approach, let's sync up on Monday to remove the common part.
We can maybe pick up IB related changes from my PR and apply here, or commit as 2 independent PRs

PR #410, opened 2025-06-01
clang_format: Add clang format and pre-commit hooks

aranadive:
This adds the required infrastructure for both pre-commit checks in local repo and github repo for clang formatting. The scripts for checking and applying formatting for only the diffs is taken from - https://github.com/barisione/clang-format-hooks. 
The process to use it is to install pre-commit hooks in local repo:
$> ./contrib/scripts/install-precommit.sh
$> git add ..
$> git commit 
... if files are not clang formatted, pre-commit script will modify files and present prompt to accept changes
$> git add ..
$> git commit

You can also modify files in staged commit
$> ./contrib/scripts/apply-format file

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -24,20 +24,45 @@ ARG DEFAULT_PYTHON_VERSION="3.12"
 FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} AS os_setup_stage
 
 # Re-declare for use in this stage
+ARG ARCH="x86_64"
 ARG DEFAULT_PYTHON_VERSION
 RUN apt-get update -y && \
     DEBIAN_FRONTEND=noninteractive apt-get -y install \
     ninja-build \
     pybind11-dev \
+    libclang-dev \
     cmake \
     libgflags-dev \
     libgrpc-dev \
     libgrpc++-dev \
     libprotobuf-dev \
+    libaio-dev \
+    liburing-dev \
     protobuf-compiler-grpc \
     libcpprest-dev \
     etcd-server \
-    etcd-client
+    etcd-client \
+    autotools-dev \
+    automake \
+    libtool \
+    libz-dev \
+    flex \
+    libgtest-dev \
+    build-essential
+
+# Add Mellanox repository and install packages
+RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64-sbsa"; else echo "${ARCH}"; fi) && \
+    export PKG_CONFIG_PATH="/opt/mellanox/doca/lib/${ARCH_SUFFIX}-linux-gnu/pkgconfig:/opt/mellanox/dpdk/lib/${ARCH_SUFFIX}-linux-gnu/pkgconfig:$PKG_CONFIG_PATH" && \
+    curl -fsSL https://linux.mellanox.com/public/repo/doca/3.0.0/ubuntu24.04/${ARCH_SUFFIX}/GPG-KEY-Mellanox.pub | \
+    gpg --dearmor | tee /usr/share/keyrings/mellanox-archive-keyring.gpg && \
+    echo "deb [signed-by=/usr/share/keyrings/mellanox-archive-keyring.gpg] https://linux.mellanox.com/public/repo/doca/3.0.0/ubuntu24.04/${ARCH_SUFFIX} ./" | \
+    tee /etc/apt/sources.list.d/mellanox.list && \
+    DEBIAN_FRONTEND=noninteractive apt update -y && \
+    apt install -y --no-install-recommends \
+        mlnx-dpdk mlnx-dpdk-dev \
+        doca-sdk-common doca-sdk-dma doca-sdk-dpdk-bridge \
+        doca-sdk-eth doca-sdk-flow doca-sdk-rdma doca-all \
+        doca-sdk-gpunetio libdoca-sdk-gpunetio-dev
 
 # --- Stage 2a: Represents using UCX from the base image ---
 FROM os_setup_stage AS ucx_upstream_image
--- a/contrib/build-container.sh
+++ b/contrib/build-container.sh
@@ -61,6 +61,14 @@ get_options() {
                 missing_requirement $1
             fi
             ;;
+        --wheel-base)
+            if [ "$2" ]; then
+                WHL_BASE=$2
+                shift
+            else
+                missing_requirement $1
+            fi
+            ;;
         --os)
             if [ "$2" ]; then
                 OS=$2
@@ -103,7 +111,6 @@ get_options() {
         --arch)
             if [ "$2" ]; then
                 ARCH=$2
-                WHL_PLATFORM=${WHL_BASE}_${ARCH}
                 shift
             else
                 missing_requirement $1
@@ -128,8 +135,10 @@ get_options() {
 
     if [[ $OS == "ubuntu22" ]]; then
         BASE_IMAGE_TAG=24.10-cuda12.6-devel-ubuntu22.04
-        WHL_PLATFORM=manylinux_2_34_${ARCH}
+        WHL_BASE=${WHL_BASE:-manylinux_2_34}
     fi
+
+    WHL_PLATFORM=${WHL_BASE}_${ARCH}
 
     if [ -z "$TAG" ]; then
         TAG="--tag nixl:${VERSION}"
@@ -152,6 +161,7 @@ show_help() {
     echo "usage: build-container.sh"
     echo "  [--base base image]"
     echo "  [--base-image-tag base image tag]"
+    echo "  [--wheel-base base platform for wheel builds]"
     echo "  [--no-cache disable docker build cache]"
     echo "  [--os [ubuntu24|ubuntu22] to select Ubuntu version]"
     echo "  [--tag tag for image]"


.clang-format
```diff
@@ -0,0 +1,71 @@
...
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: Empty
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: WithoutElse
```
ColinNV, line 13: Suggest changing to `Never` for consistency and readability.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  BeforeElse: false
+  BeforeLambdaBody: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: true
+ColumnLimit: 100
+CompactNamespaces: false
+ConstructorInitializerIndentWidth: 8
```
ColinNV: This and the next one could be reduced to 4 to match IndentWidth.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+ReferenceAlignment: Right
+ReflowComments: true
+SortIncludes: false
+SortUsingDeclarations: false
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: false
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCtorInitializerColon: true
+SpaceBeforeInheritanceColon: true
+SpaceBeforeParens: NonEmptyParentheses
```
ColinNV, line 62: Suggest changing this to `Never`, and then using
```
SpacesInAngles = true
SpacesInParentheses = true
SpacesInSquareBrackets = true
```
to provide for a more uniform readable experience.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  AfterExternBlock: false
+  BeforeCatch: true
+  BeforeElse: false
+  BeforeLambdaBody: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: true
+ColumnLimit: 100
```
ColinNV, line 39: This feels a bit low for today, suggest going to around 150.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeCatch: true
+  BeforeElse: false
+  BeforeLambdaBody: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: AfterColon
```
ColinNV: Suggest setting both this and `BreakInheritanceList` to `BeforeColon`.

ColinNV:
Btw, the overall look I would go for can be seen in files like these:

https://github.com/taocpp/PEGTL/blob/work/include/tao/pegtl/parse_error.hpp
https://github.com/taocpp/PEGTL/blob/work/include/tao/pegtl/internal/mmap_file_posix.hpp

If I were to get one wish it would be the spaces within all kinds of brackets, and no spaces before them.

.clang-format
mkhazraee: The corner case is "if (ret) return ret;" which is quite common in our code, and making it 2 liner is not the best either. The chances of someone misusing this rule for longer ifs is low, so I know it's taking it with a grain of salt.

.clang-format
mkhazraee: I say 150 is too high. Still we look at laptop and half screens, I say 120 is the upper limit. If you feel strongly about this, let's go 120.

.clang-format
mkhazraee: I understand where you are coming from, in my personal codes I usually don't use space when using a call, similar to how I use asserts, but we had this debate before, similar arguments and eventually we went for spacing, to be more similar to if/for etc and sometimes  a space between name and first arg helps. So I think we can keep it as is, and for empty parentheses the latter is not the case, so it's fine.

.clang-format
```diff
@@ -0,0 +1,71 @@
...
+DerivePointerAlignment: false
+IndentCaseLabels: false
+IndentPPDirectives: None
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: Inner
+
+PackConstructorInitializers: CurrentLine
+PointerAlignment: Right
```
mkhazraee, line 52: We had a long discussion on this. Since we have pointer reference output for reqH and so, which is a reference to a pointer, we said let's do "T* &name", and be consistent with this throughout: all references get attached to the right, while all *s attach left to the type. Right now "T *&name" looks a tad strange. Also at the end of the day being pointer is the type. The only corner case is defining multiple pointers in the same line, which is not that common, taking it with a grain of salt.

.clang-format
```diff
@@ -0,0 +1,71 @@
+---
+AccessModifierOffset: -4
```
mkhazraee, line 2: I like -2 more than -4, not to be fully aligned, but not strongly defending it, I see the pros and cons, just don't like the class term and something inside it being aligned.

.clang-format
mkhazraee: (If clang cannot support it, then I'm more fan of keeping the pointer to the left and reference to the right, and ignoring this specific case. But not strongly arguing for it.)

.clang-format
ColinNV: Not sure why that corner case should get special treatment.

.clang-format
ColinNV: So that would produce this?
```
if (expression)
function (argument, argument)
function()
array[index]
templated<parameter> foo
templated<> bar
```

I find the following more readable and consistent, but of course that's subjective :-)
```
if( expression )
function( argument, argument )
function()
array[ index ]
templated< parameter > foo
templated<> bar
```

.clang-format
ColinNV: I'm happy for any increase!

.clang-format
ColinNV: Hopefully a reference to a pointer is so rare that it doesn't matter if it's ugly. Given the choice I'd prefer pointers and references attached to the type, for consistency, since they are all part of the type, and also because sometimes there is no name to attach the reference to:

```
class my_class
{
   my_class( my_class&& ) = delete;
   my_class( const my_class& ) = delete;
};
```

(Ok, you can just put a name there, but I usually try to avoid unnecessary things.)

If the consensus is on pointers-with-types, references-with-names I'll live with it :-)

.clang-format
ColinNV: Are you saying you don't like this (which I personally would prefer)?

```
class foo
{
public:
   foo();

private:
   stuff;
};
```

.clang-format
w1ldptr: Yeah, space before function argument list looks a bit weird. In the previous PR that was reverted the value of this option was "ControlStatements" which resulted more conventional look. @aranadive why did you decide to change it?

Anyway, from my point of view having bad style is better than no style, so no strong objections from me ;)

.clang-format
w1ldptr: Note that indenting arbitrary stuff with fractional indent (2 spaces instead of 4 in this case) will make it quite challenging for developers to configure their preferred editors to match the style. Yes, they can still just autoreformat everything during commit, but it is usually nicer to see the code in its final form during development.

aranadive:
@ColinNV @w1ldptr I have been using https://zed0.co.uk/clang-format-configurator/ to see how the clang formatter looks (this supports not all clang options we use but most). Based on it, I felt that the spaceBeforeParens,  shortIfStatementsOnASingleLine deserved a relook and updated their values. But if there is a strong preference otherwise, Ill revert to the earlier PR values.

.clang-format
aranadive: Yes, I prefer 0 here to indent based on rest of clang. But -4 was agreed earlier so went with that.

.clang-format
aranadive: Moein prefers my_class* &class1 .. That doesnt seem to be doable with current clang formatters. Best I can get is my_class *&class1. Not sure if anyone has better ideas.

.clang-format
mkhazraee: Agreed it's subjective, maybe I'm used to if (statement), so ControlStatements or this NonEmptyParentheses look to me. If Adit and Vlad and Mikhail also prefer these, let's go with one of the two.

.clang-format
mkhazraee: The tricky part is that corner case is part of our main API. But if not doable, it is what it is. 

Just to clarify, this is my thinking: with pointers I like left align too as it's practically part of the type. With that mindset, then reference is how you pass the value, and not the type, so nice to distinguish it by making it right attached. I see Colin's point that if no name is passed that would look off, not sure how clang handles it, but I think it's less common, mainly used for the examples you gave.

.clang-format
mkhazraee: Agreed it's a corner case, but common in our code :D On the other hand, this is if there is no else, no {, so a single condition and single action. This is also not that common, and the case that both of them are really long and the developer doesn't put any line breaks sounds rare too. Just thinking to give both options to the developer in such a scenario.

.clang-format
w1ldptr: I don't have a strong preference between 0 and -4 here. The reason I was pushing for not indenting access modifiers and namespaces is that I didn't want those to eat up into our limited budget of 100 character line length.

.clang-format
aranadive: Ok I think we can keep this as is for now.

w1ldptr:
> @ColinNV @w1ldptr I have been using https://zed0.co.uk/clang-format-configurator/ to see how the clang formatter looks (this supports not all clang options we use but most). Based on it, I felt that the spaceBeforeParens, shortIfStatementsOnASingleLine deserved a relook and updated their values. But if there is a strong preference otherwise, Ill revert to the earlier PR values.

From my point of view:
- shortIfStatementsOnASingleLine is fine if we also agree that such single-line statements don't need to be wrapper in {}. I think there was a disagreement about this before.
- spaceBeforeParens for function calls is not a usual formatting for neither popular C code bases, nor for Rust. But as I said before a bad code style is much better than no code style so I support merging whatever we have and modify the rules afterwards, if need be.

.clang-format
aranadive: Done.

.clang-format
aranadive: ControlStatements looked weird. We already are using NonEmptyParentheses in most cases so stuck with this.

.clang-format
aranadive: Ok.

w1ldptr (dismissed):

contrib/scripts/apply-format
```diff
@@ -0,0 +1,348 @@
+#! /bin/bash
+#
+# Copyright 2018 Undo Ltd.
+#
+# https://github.com/barisione/clang-format-hooks
```
yosefe, line 5: do we need to add https://github.com/barisione/clang-format-hooks/blob/master/COPYING?

src/api/cpp/backend/backend_engine.h
```diff
@@ -92,9 +92,8 @@ class nixlBackendEngine {
         // *** Pure virtual methods that need to be implemented by any backend *** //
 
         // Register and deregister local memory
-        virtual nixl_status_t registerMem (const nixlBlobDesc &mem,
-                                           const nixl_mem_t &nixl_mem,
-                                           nixlBackendMD* &out) = 0;
+        virtual nixl_status_t
+        registerMem (const nixlBlobDesc &mem, const nixl_mem_t &nixl_mem, nixlBackendMD *&out) = 0;
```
yosefe: why need formatting changes in this PR?
are these the only places in the code that are not according to the new style?

contrib/scripts/apply-format
aranadive: This will be added to the attributions file during release.

src/api/cpp/backend/backend_engine.h
aranadive: No I made these to show the formatter looked. I will remove these files if everyone is okay with the options. Only the clang-format file and the supporting scripts will be added.

aranadive: I removed these files.

yosefe:
replaced by #474 
@dpressle will follow up on adding CI

PR #412, opened 2025-06-02
API/CPP: Replaced classes with only public members with structures.

rakhmets:
Replaced classes with only public members with structures.

### DIFF
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -145,67 +145,68 @@ enum class nixl_cost_t {
 };
 
 /**
- * @class nixlAgentOptionalArgs
- * @brief A class for optional argument that can be provided to relevant agent methods.
+ * @struct nixlAgentOptionalArgs
+ * @brief A structure for optional argument that can be provided to relevant agent methods.
  */
-class nixlAgentOptionalArgs {
-    public:
-        /**
-         * @var backends vector to specify a list of backend handles, to limit the list
-         *      of backends to be considered. Used in registerMem / deregisterMem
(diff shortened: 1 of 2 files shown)


src/api/cpp/nixl_types.h
```diff
@@ -148,64 +148,65 @@ enum class nixl_cost_t {
  * @class nixlAgentOptionalArgs
```
aranadive: Please fix doc comments

rakhmets:
> seems it should not break API, but what is the motivation? I'd suggest to modify API files alltogether once we agree on all proposed changes

The motivation is to remove redundant code. This is a minor change that can be merged without requiring changes to the code using the library.

ColinNV:
> seems it should not break API

I vaguely remember some compiler barfing when a `class` was forward-declared as `struct`, or vice versa, but is that a problem?

yosefe (commented):
LGTM besides Moein's comment

src/api/cpp/nixl_types.h
rakhmets: Fixed.

mkhazraee:
I'm not sure why we put this to class, I guess the idea was if we want to add features to it, let's say a constructor or so, not to block it. We might have run to that issue that Colin mentioned, but I don't think we forward declare them and things are fine for now. And since it's extra params and most prob we add some overloading soon, I think it's fine to proceed with this. Agreed a tad with Mikhail to merge them with API changes like adding overload, but not a big deal for practically 4 lines.

mkhazraee (approved):

brminich (approved):

PR #413, opened 2025-06-02
PLUGINS/UCX: Minor code cleanup related to notification lists.

rakhmets:
`vec.erase(vec.begin(), vec.end())` -> `vec.clear()`
Moved helper function to anonymous namespace.
`lock() + unlock()` -> `std::lock_guard`


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -33,6 +33,16 @@
 
 #endif
 
+namespace {
+    void moveNotifList(notif_list_t &src, notif_list_t &tgt)
```
iyastreb, line 37: Do we really need a separate function?
I think we can just keep a single call that already does everything
```
std::move(src.begin(), src.end(), std::back_inserter(tgt));
```
Maybe no need to check for size and call clear?

brminich: seems checking `src.size() > 0` is a micro optimization, which would be better to keep given that it is called from the progress

rakhmets: `clear()` is called to erase all elements from the container. After `std::move` the elements in the moved-from container will still contain valid values (not necessarily the same values as before the move). This means that the size of the moved-from container will be the same.

yosefe (approved):

PR #415, opened 2025-06-02
UCX: Finer grained progress to improve performance

tvegas1:
When calling postXfer with 120k blocks of 32kB, we want to progress once, not for every block. For instance, on AWS, sending with cuda_ipc on single node from GPU0 to GPU1, we get performance improvement with this fix (must be release build).

Various perf improvements observed from default (0) to pr (1) to retricting transport (2) like: `UCX_TLS=srd,cuda_ipc,cuda_copy` (rc_x would replace srd on other platforms, EFA is coincidence here)

Other ideas if 120k blocks is really a use-case:
- Try not tracking cuda_ipc copies individually
- Improve request tracking ->status() to not be O(n) wrt requests (use request callbacks)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -570,7 +570,6 @@ nixl_status_t nixlUcxWorker::test(nixlUcxReq req)
     if(req == nullptr) {
         return NIXL_SUCCESS;
     }
-    ucp_worker_progress(worker.get());
```
brminich: maybe better do not remove `progress` here, but just call `ucx_status_to_nixl(ucp_request_check_status(req))` instead of `test` in `status()`?

brminich (commented):
nice catch!

src/utils/ucx/ucx_utils.cpp
tvegas1: fixed

yosefe (approved):

tvegas1:
blocked by CI, would be fixed by #417

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -365,11 +365,15 @@ class nixlUcxBackendH : public nixlBackendReqH {
         }
 
         const auto &uw = eng.getWorker(worker_id);
+
+        /* Maximum progress */
+        while (uw->progress());
```
w1ldptr, line 380: It is safe to do when progress thread is enabled? How does progressing a worker here interacts with the same worker being armed and polled on concurrently?

tvegas1: in origin/main branch, progress is already done in `uw->test((nixlUcxReq)req)` below, so I am assuming this is already a working use-case??

w1ldptr: Thanks for clarifying.

PR #417, opened 2025-06-03
TEST/PLUGIN: Add more logs

tvegas1:
Test is failing, CI image might itself contain plugins.


test/nixl/test_plugin.cpp
```diff
@@ -89,25 +91,25 @@ int main(int argc, char** argv) {
         staticPlugs.insert(plugin.name);
     }
 
-    verify_plugin("UCX", plugin_manager);
-    verify_plugin("GDS", plugin_manager);
-    verify_plugin("POSIX", plugin_manager);
-    verify_plugin("UCX_MO", plugin_manager);
-    verify_plugin("MOCK_BASIC", plugin_manager);
-    verify_plugin("MOCK_DRAM", plugin_manager);
+    for (const auto& plugin : plugins) {
```
aranadive, line 95: Can you add a note here why this needs to be done?

test/nixl/test_plugin.cpp
```diff
@@ -89,25 +91,25 @@ int main(int argc, char** argv) {
...
+        plugin_manager.unloadPlugin(plugin);
+    }
+
+    auto preloaded_plugins = plugin_manager.getLoadedPluginNames().size();
```
aranadive: This should be 0 after unloading all above plugins (if there are no static plugins). Have you tried running this test with certain plugins built statically?

tvegas1: Seems there is another plugin named `MOCK_BACKEND`, somehow could not find it, but it is found on CI machine. I think it is fair to assume existence of unknown non-static plugins, although independently, it would be good to eventually remove it.

test/nixl/test_plugin.cpp
tvegas1: fixed

test/nixl/test_plugin.cpp
tvegas1: removed that part since CI seems now properly cleaned

.gitlab/test_cpp.sh
```diff
@@ -61,7 +61,7 @@ cd ${INSTALL_DIR}
 ./bin/ucx_backend_multi
 ./bin/serdes_test
 ./bin/gtest
-./bin/test_plugin
+NIXL_LOG_LEVEL=TRACE ./bin/test_plugin
```
aranadive: Do we still need the TRACE log here?

tvegas1: I think it is useful as it provides infos on plugins it found at class instantiation time and also it only adds ~10 lines in output. Also it was needed to understand the original failure.

brminich: we expect TRACE to grow, I'd remove it now or replace with INFO (afair plugins info loading is going to info)

tvegas1: fixed

PR #418, opened 2025-06-03
Add Nsight markers

rafiw:
This allows us to mark parts of the code to be shown in Nvidia Nsight

### DIFF
--- a/src/plugins/gpunetio/gpunetio_backend.cpp
+++ b/src/plugins/gpunetio/gpunetio_backend.cpp
@@ -24,6 +24,8 @@
 #include "common/nixl_log.h"
 #include <arpa/inet.h>
 
+#include <nvtx3/nvtx3.hpp>
+
 const char info_delimiter = '-';
 
 /****************************************
@@ -1237,10 +1239,16 @@ nixlDocaEngine::postXfer(const nixl_xfer_op_t &operation,
 
         switch (operation) {
         case NIXL_READ:
-            doca_kernel_read(treq->stream, xferReqRingCpu[idx].qp_data, xferReqRingGpu, idx);
+            {
+                nvtx3::mark("nixl_doca_read");
+                doca_kernel_read(treq->stream, xferReqRingCpu[idx].qp_data, xferReqRingGpu, idx);
+            }
             break;
         case NIXL_WRITE:
-            doca_kernel_write(treq->stream, xferReqRingCpu[idx].qp_data, xferReqRingGpu, idx);
+            {
+                nvtx3::mark("nixl_doca_write");
+                doca_kernel_write(treq->stream, xferReqRingCpu[idx].qp_data, xferReqRingGpu, idx);
+            }
             break;
         default:
             return NIXL_ERR_INVALID_PARAM;


meson.build
```diff
@@ -43,15 +43,15 @@ cuda_lib_path = get_option('cudapath_lib')
 cuda_stub_path = get_option('cudapath_stub')
 
 if cuda_lib_path == ''
-    cuda_dep = dependency('cuda', required : false, modules : [ 'cudart', 'cuda' ])
+    cuda_dep = dependency('cuda', required : false, modules : [ 'cudart', 'cuda', 'nvToolsExt'])
```
yosefe: can we make nvToolsExt optional

src/core/nixl_agent.cpp
```diff
@@ -849,12 +850,14 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
     }
 
     // If status is not NIXL_IN_PROG we can repost,
+    NVTXRangePush("NixlAgent::postXferReq");
```
yosefe: core agent code should not use NVTX directly

src/utils/common/util.h
```diff
@@ -21,4 +21,6 @@
 #define CONCAT_0(a, b) a ## b
 #define UNIQUE_NAME(name) CONCAT(name, __COUNTER__)
 
+void NVTXRangePush(const char* const name);
+void NVTXRangePop();
```
yosefe: 1. space line after
2. can we make it a runtime option to avoid function calls overhead for each post xfer

src/utils/common/util.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+#include "util.h"
+
+#ifdef HAVE_NVTX
+#include <nvToolsExt.h>
+#endif
+
+void NVTXRangePush(const char* const name) {
```
yosefe: const char *

yosefe (changes requested):

src/core/nixl_agent.cpp
rafiw: this is an usage example, which functions do you think should be marked with nsight?

meson.build
rafiw: it's already optional as you can see
`required : false`

rafiw:
I updated the code to work with cuda 12.9+
will push a version after i test it still works

brminich:
@rafiw is it still relevant? Can you pls resolve conflicts if it is?

rafiw:
This code now uses nvxt3 so its much more simple, it's header only.
From what i understood no need to ifdef since it does nothing if not run from Nsight instrumentation.

brminich:
This was introduced through [#1845](https://github.com/ai-dynamo/nixl/pull/1845), [#1852](https://github.com/ai-dynamo/nixl/pull/1852), and [#1867](https://github.com/ai-dynamo/nixl/pull/1867); the resulting behavior is documented in [tracing.md](https://github.com/ai-dynamo/nixl/blob/main/docs/tracing.md).

PR #419, opened 2025-06-03
UTILS/SERDES: Optimize and make safer.

ColinNV:
Refactor the serdes facilities to make them more efficient and less error prone, and split into two classes.

### DIFF
--- a/contrib/aws-efa/aws_test.sh
+++ b/contrib/aws-efa/aws_test.sh
@@ -78,6 +78,7 @@ JOB_ID=$(aws batch submit-job \
     --job-definition "NIXL-Ubuntu-JD" \
     --job-queue ucx-nxil-jq \
     --eks-properties-override file://./aws_vars.json \
+    --retry-strategy '{"attempts":3}' \
     --query 'jobId' --output text)
 
 # Function to wait for a specific job status
@@ -104,7 +105,7 @@ wait_for_status() {
 
 # Wait for the job to start running
 echo "Waiting for job to start running (timeout: 30m)..."
-if ! wait_for_status "RUNNING" 1800 180; then
+if ! wait_for_status "RUNNING" 1800 10; then
     echo "Job failed to start"
     exit 1
 fi
@@ -112,11 +113,11 @@ fi
 # Stream logs from the pod
 POD=$(aws batch describe-jobs --jobs "$JOB_ID" --query 'jobs[0].eksProperties.podProperties.podName' --output text)
 echo "Streaming logs from pod: $POD"
-kubectl -n ucx-ci-batch-nodes logs -f "$POD"
+kubectl -n ucx-ci-batch-nodes logs -f "$POD" || kubectl -n ucx-ci-batch-nodes logs "$POD" --previous || true
 
 # Check final job status
 echo "Waiting for job completion (timeout: 10m)..."
-exit_status=$(wait_for_status "SUCCEEDED|FAILED" 600 60)
+exit_status=$(wait_for_status "SUCCEEDED|FAILED" 600 10)
 if [[ "$exit_status" =~ FAILED ]]; then
     echo "Failure running NIXL tests"
     exit 1
--- a/src/api/cpp/backend/backend_aux.h
+++ b/src/api/cpp/backend/backend_aux.h
@@ -28,14 +28,13 @@
 typedef std::vector<std::pair<std::string, std::string>> notif_list_t;
 
 
-class nixlBackendOptionalArgs {
-    public:
-        // During postXfer, user might ask for a notification if supported
-        nixl_blob_t notifMsg;
-        bool        hasNotif = false;
+struct nixlBackendOptionalArgs {
+    // During postXfer, user might ask for a notification if supported
+    nixl_blob_t notifMsg;
+    bool        hasNotif = false;
 };
 
-typedef nixlBackendOptionalArgs nixl_opt_b_args_t;
+using nixl_opt_b_args_t = nixlBackendOptionalArgs;
 
 
 // A base class to point to backend initialization data
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -147,67 +147,68 @@ enum class nixl_cost_t {
 };
 
 /**
- * @class nixlAgentOptionalArgs
- * @brief A class for optional argument that can be provided to relevant agent methods.
+ * @struct nixlAgentOptionalArgs
+ * @brief A structure for optional argument that can be provided to relevant agent methods.
  */
-class nixlAgentOptionalArgs {
-    public:
-        /**
-         * @var backends vector to specify a list of backend handles, to limit the list
-         *      of backends to be considered. Used in registerMem / deregisterMem
-         *      makeConnection / prepXferDlist / makeXferReq / createXferReq / GetNotifs / GenNotif
-         */
-        std::vector<nixlBackendH*> backends;
+struct nixlAgentOptionalArgs {
+    /**
+     * @var backends vector to specify a list of backend handles, to limit the list
+     *      of backends to be considered. Used in registerMem / deregisterMem
+     *      makeConnection / prepXferDlist / makeXferReq / createXferReq / GetNotifs / GenNotif
+     */
+    std::vector<nixlBackendH*> backends;
 
-        /**
-         * @var notifMsg A message to be used in createXferReq / makeXferReq / postXferReq,
-         *               if a notification message is desired
-         */
-        nixl_blob_t notifMsg;
-        /**
-         * @var hasNotif boolean value to indicate that a notification is provided, or to
-         *      remove notification during a repost. If set to false, notifMsg is not checked.
-         */
-        bool hasNotif = false;
+    /**
+     * @var notifMsg A message to be used in createXferReq / makeXferReq / postXferReq,
+     *               if a notification message is desired
+     */
+    nixl_blob_t notifMsg;
 
-        /**
-         * @var makeXferReq boolean to skip merging consecutive descriptors, used in makeXferReq.
-         */
-        bool skipDescMerge = false;
+    /**
+     * @var hasNotif boolean value to indicate that a notification is provided, or to
+     *      remove notification during a repost. If set to false, notifMsg is not checked.
+     */
+    bool hasNotif = false;
 
-        /**
-         * @var includeConnInfo boolean to include connection information in the metadata,
-         *                      used in getLocalPartialMD.
-         */
-        bool includeConnInfo = false;
+    /**
+     * @var makeXferReq boolean to skip merging consecutive descriptors, used in makeXferReq.
+     */
+    bool skipDescMerge = false;
 
-        /**
-         * @var ipAddr Used to specify the IP address of a remote peer for metadata transfer.
-         *                      used in sendLocalMD, fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
-         */
-        std::string ipAddr;
-        /**
-         * @var port Used to specify the port of a remote peer, ipAddr must also be set
-         *                      used in sendLocalMD, fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
-         */
-        int port = default_comm_port;
+    /**
+     * @var includeConnInfo boolean to include connection information in the metadata,
+     *                      used in getLocalPartialMD.
+     */
+    bool includeConnInfo = false;
 
-        /**
-         * @var metadataLabel Used to specify the label of the metadata to be sent/fetched
-         *                    when working with ETCD metadata server. The label will be appended to the
-         *                    agent's key prefix, and the full key will be used to store/fetch
-         *                    the metadata key-value pair from the server.
-         *                    Used in fetchRemoteMD, sendLocalPartialMD.
-         *                    Note that sendLocalMD always uses default_metadata_label and ignores this parameter.
-         *                    Note that invalidateLocalMD invalidates all labels and ignores this parameter.
-         */
-        std::string metadataLabel;
+    /**
+     * @var ipAddr Used to specify the IP address of a remote peer for metadata transfer.
+     *                      used in sendLocalMD, fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
+     */
+    std::string ipAddr;
+
+    /**
+     * @var port Used to specify the port of a remote peer, ipAddr must also be set
+     *                      used in sendLocalMD, fetchRemoteMD, invalidateLocalMD, sendLocalPartialMD.
+     */
+    int port = default_comm_port;
+
+    /**
+     * @var metadataLabel Used to specify the label of the metadata to be sent/fetched
+     *                    when working with ETCD metadata server. The label will be appended to the
+     *                    agent's key prefix, and the full key will be used to store/fetch
+     *                    the metadata key-value pair from the server.
+     *                    Used in fetchRemoteMD, sendLocalPartialMD.
+     *                    Note that sendLocalMD always uses default_metadata_label and ignores this parameter.
+     *                    Note that invalidateLocalMD invalidates all labels and ignores this parameter.
+     */
+    std::string metadataLabel;
 };
 /**
  * @brief A typedef for a nixlAgentOptionalArgs
  *        for providing extra optional arguments
  */
-typedef nixlAgentOptionalArgs nixl_opt_args_t;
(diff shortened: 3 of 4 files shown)


src/api/cpp/nixl_descriptors.h
```diff
@@ -208,12 +208,11 @@ class nixlDescList {
...
         /**
-         * @brief Deserializer constructor for nixlDescList from nixlSerDes object
-         *        which serializes/deserializes our classes into/from blobs
+         * @brief Deserializer constructor for nixlDescList from nixlDeserializer.
          *
-         * @param deserialize nixlSerDes object to construct nixlDescList
+         * @param deserialize nixlDeserializer object to construct nixlDescList
          */
-        nixlDescList(nixlSerDes* deserializer);
+        nixlDescList(nixlDeserializer* deserializer);
```
iyastreb: Is it supposed to call this method with nullptr deserializer? If not, we can pass by ref

src/api/cpp/nixl_descriptors.h
```diff
@@ -324,11 +323,11 @@ class nixlDescList {
...
         int getIndex(const nixlBasicDesc &query) const;
         /**
-         * @brief Serialize a descriptor list with nixlSerDes class
-         * @param serializer nixlSerDes object to serialize nixlDescList
+         * @brief Serialize a descriptor list with nixlSerializer class
+         * @param serializer nixlSerializer object to serialize nixlDescList
          * @return nixl_status_t Error code if serialize was not successful
          */
-        nixl_status_t serialize(nixlSerDes* serializer) const;
+        nixl_status_t serialize(nixlSerializer* serializer) const;
```
iyastreb: pass by ref?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -214,15 +214,15 @@ PYBIND11_MODULE(_bindings, m) {
         .def("print", &nixl_xfer_dlist_t::print)
         .def(py::pickle(
             [](const nixl_xfer_dlist_t& self) { // __getstate__
-                nixlSerDes serdes;
-                self.serialize(&serdes);
-                return py::bytes(serdes.exportStr());
+                nixlSerializer nser;
+                self.serialize(&nser);
+                return py::bytes(nser.exportStr());
```
iyastreb, line 219: will it call move method?

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,154 @@
...
-       return "";
+nixlDeserializer::nixlDeserializer(std::string&& str)
+    : buffer_(std::move(str)),
+      offset_(introString.size())
+{
+    if(buffer_.size() < introString.size() ) {
+        throw std::runtime_error("insufficient deserialization data for intro");
+    }
+    if(std::memcmp(buffer_.data(), introString.data(), introString.size() != 0 )) {
+        throw std::runtime_error("invalid deserializtion intro data");
```
iyastreb:
```suggestion
        throw std::runtime_error("invalid deserialization intro data");
```

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,154 @@
...
-
-    //move past string plus | delimiter
-    des_offset += len + 1;
-
-    return ret;
 }
 
-// Byte buffers serialization
-nixl_status_t nixlSerDes::addBuf(const std::string &tag, const void* buf, ssize_t len){
+nixlDeserializer::nixlDeserializer(const std::string& str)
```
iyastreb, line 85: maybe str can be std::string, then you can std::move it - no extra copies are created in this case for temporary objects
Or it can be rvalue ref for the same purpose

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,154 @@
...
-std::string nixlSerDes::getStr(const std::string &tag){
+const std::string& nixlSerializer::exportStr() const & noexcept
+{
+    return buffer_;
+}
 
-    if(workingStr.compare(des_offset, tag.size(), tag) != 0){
-       //incorrect tag
-       return "";
+nixlDeserializer::nixlDeserializer(std::string&& str)
```
iyastreb, line 73: should deserializer own a copy of the string? Maybe string_view is sufficient?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1173,11 +1172,12 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
     hdr.op = NOTIF_STR;
     flags |= UCP_AM_SEND_FLAG_EAGER;
 
+    nixlSerializer sd;
     sd.addStr("name", localAgent);
     sd.addStr("msg", msg);
     // TODO: replace with mpool for performance
 
-    auto buffer = std::make_unique<std::string>(std::move(sd.exportStr()));
+    auto buffer = std::make_unique<std::string>(std::move(sd).exportStr());
```
w1ldptr: What does the move of nixlSeriealizer instance before exporting a string from it achieves here?

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,63 @@
...
 
 #include <cstring>
 #include <string>
+#include <string_view>
 #include <cstdint>
 
 #include "nixl_types.h"
 
-class nixlSerDes {
+class nixlSerializer
```
rakhmets, line 28: Please follow this recommendation: "... Use the public before protected before private order. ..."

I understand that most of the code that has already been merged does not follow the guideline. But since these classes have been rewritten, I think it's a good time to start.

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,63 @@
...
-    ssize_t des_offset;
-    ser_mode_t mode;
+    void addRaw(const std::string_view&);
+    void addRaw(const void* buf, const std::size_t len);
+
+    template<std::size_t N>
+    void addRaw(const char(&literal)[N])
+    {
+        addRaw(literal, N - 1);
+    }
```
w1ldptr: nice!

src/core/nixl_agent.cpp
```diff
@@ -1024,34 +1023,23 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
...
 
-    nixlSerDes sd;
-    ret = sd.addStr("Agent", data->name);
-    if(ret)
-        return ret;
-
-    ret = sd.addBuf("Conns", &conn_cnt, sizeof(conn_cnt));
-    if(ret)
-        return ret;
+    nixlSerializer sd;
```
rakhmets: Please update the variable name. (Here and in other places).

src/bindings/python/nixl_bindings.cpp
ColinNV: I was assuming that the `py::bytes` object copies the string data anyhow, or does it wrap a `std::string`? The other `exportStr()` returns a const-reference.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1206,7 +1197,7 @@ nixlUcxEngine::notifAmCb(void *arg, const void *header,
                          const ucp_am_recv_param_t *param)
 {
     struct nixl_ucx_am_hdr* hdr = (struct nixl_ucx_am_hdr*) header;
-    nixlSerDes ser_des;
+    nixlDeserializer sd;
```
rakhmets: Please change the variable name. Maybe `deserializer`.

src/utils/serdes/serdes.cpp
ColinNV: It's overloaded for `const std::string&` and `std::string&&`.

src/utils/serdes/serdes.cpp
ColinNV: That's a good question that I considered, but then didn't want to answer in _this_ PR. But I can check now if it's not too much of a hassle.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -798,18 +798,14 @@ nixl_status_t nixlUcxEngine::disconnect(const std::string &remote_agent) {
...
     if(remoteConnMap.count(remote_agent)) {
         return NIXL_ERR_INVALID_PARAM;
     }
 
-    nixlSerDes::_stringToBytes(addr.data(), remote_conn_info, size);
     std::shared_ptr<nixlUcxConnection> conn = std::make_shared<nixlUcxConnection>();
     bool error = false;
     for (auto &uw: uws) {
-        auto result = uw->connect(addr.data(), size);
+        auto result = uw->connect(remote_conn_info.data(), remote_conn_info.size());
```
rakhmets, line 822: :thumbsup:

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,156 @@
...
+    : buffer_(std::move(str)),
+      offset_(introString.size())
+{
+    if(buffer_.size() < introString.size() ) {
+        throw std::runtime_error("insufficient deserialization data for intro");
+    }
+    if(std::memcmp(buffer_.data(), introString.data(), introString.size() != 0 )) {
```
w1ldptr: It looks like '!= 0' here is on the wrong side of the first closing bracket, i.e. it compares string size to 0, not the result of memcmp function.

src/utils/serdes/serdes.h
ColinNV: I prefer public before protected before private myself but adapted to what appeared to be the style preferred here. Happy to change the code here!

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,156 @@
...
 
-    //skip over tag and size, which we assume has been read previously
-    des_offset += tag.size() + sizeof(ssize_t);
-
-    //_stringToBytes(buf, workingStr.data() + des_offset, len);
-    _stringToBytes(buf, workingStr.substr(des_offset, len), len);
-
-    //bytes in string form are twice as long, skip those plus | delimiter
-    des_offset += len + 1;
+nixl_status_t nixlDeserializer::getBuf(const std::string_view& tag, void *buf, const std::size_t len)
```
w1ldptr: This method seem to be used to deserialize POD integer types. I wonder whether it could be trivially "upgraded" to a template which would alleviate the need for passing len.

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,156 @@
...
-std::string nixlSerDes::exportStr() const {
-    return workingStr;
-}
-
-nixl_status_t nixlSerDes::importStr(const std::string &sdbuf) {
-
-    if(sdbuf.compare(0, 11, "nixlSerDes|") != 0){
-       //incorrect tag
-       return NIXL_ERR_MISMATCH;
+nixl_status_t nixlDeserializer::importStr(std::string raw)
```
w1ldptr, line 140: Is there a real use-case for this method? On the cursory glance it seems that all users can just use the constructor directly.

w1ldptr (commented):
Besides other comments this PR seems to be missing the corresponding updates to nixlbench code.

src/plugins/ucx/ucx_backend.cpp
ColinNV: It means that `*this` is an r-value reference so that the `&&` qualified `exportStr()` overload is called, the one that can move from the internal string.

src/utils/serdes/serdes.cpp
ColinNV: Good catch!

src/utils/serdes/serdes.cpp
ColinNV: Yes, absolutely, but the constructor reports errors through exceptions which we can't use everywhere yet...

src/utils/serdes/serdes.cpp
w1ldptr: Perhaps it would be wiser to postpone widespread string_view usage until after CI starts running with address sanitizer enabled.

test/nixl/nixl_test.cpp
```diff
@@ -82,11 +82,11 @@ static void targetThread(nixlAgent &agent, nixl_opt_args_t *extra_params, int th
     dram_for_ucx.print();
 
     /** Only send desc list */
-    nixlSerDes serdes;
-    assert(dram_for_ucx.trim().serialize(&serdes) == NIXL_SUCCESS);
+    nixlSerializer serdes;
```
rakhmets: Please update the name.

test/unit/utils/serdes/serdes_test.cpp
```diff
@@ -14,52 +14,43 @@
...
 
-    nixlSerDes sd;
+    const int i = 0xff;
+    const std::string s = "testString";
+    const std::string t1 = "i";
+    const std::string t2 = "s";
 
-    ret = sd.addBuf(t1, &i, sizeof(i));
-    assert(ret == 0);
+    nixlSerializer sd;
```
rakhmets: Please rename. Here and at line 42.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -468,12 +466,12 @@ std::string nixlUcxWorker::epAddr()
         NIXL_WARN << "Unable to query UCX endpoint address";
         return {};
     }
-    const std::string result = nixlSerDes::_bytesToString(wattr.address, wattr.address_length);
+    const auto result = std::string(reinterpret_cast<const char*>(wattr.address), wattr.address_length);  // Deserialize?
```
rakhmets: Deserialization is not needed. The worker address was obtained using the UCP API of UCX library. So it was not serialized by NIXL.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -521,7 +519,7 @@ std::string nixlUcxContext::packRkey(nixlUcxMem &mem)
         /* TODO: MSW_NET_ERROR(priv->net, "failed to ucp_rkey_pack (%s)\n", ucs_status_string(status)); */
         return {};
     }
-    const std::string result = nixlSerDes::_bytesToString(rkey_buf, size);
+    const auto result = std::string(reinterpret_cast<const char*>(rkey_buf), size);  // Deserialize?
```
rakhmets: Deserialization is not needed. The remote key was packed by UCX library.

w1ldptr (approved):

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -620,10 +620,11 @@ xferBenchNixlWorker::exchangeIOV(const std::vector<std::vector<xferBenchIOV>> &l
                 buffer = (char *)calloc(desc_str_sz, sizeof(*buffer));
                 rt->recvChar((char *)buffer, desc_str_sz, srcrank);
 
-                std::string desc_str(buffer, desc_str_sz);
-                ser_des.importStr(desc_str);
+                const std::string desc_str(buffer, desc_str_sz);
+                nixlDeserializer des;
```
brminich: minor: naming a bit inconsistent, you call serializer **n**ser, but deserializer just des

src/api/cpp/nixl_descriptors.h
```diff
@@ -208,12 +208,11 @@ class nixlDescList {
...
         /**
-         * @brief Deserializer constructor for nixlDescList from nixlSerDes object
-         *        which serializes/deserializes our classes into/from blobs
+         * @brief Deserializer constructor for nixlDescList from nixlDeserializer.
          *
-         * @param deserialize nixlSerDes object to construct nixlDescList
+         * @param deserialize nixlDeserializer object to construct nixlDescList
          */
-        nixlDescList(nixlSerDes* deserializer);
+        explicit nixlDescList(nixlDeserializer& deserializer);
```
brminich, line 215: it is API change

@yosefe, @aranadive is it ok, or can only be done together with other API changes?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
ColinNV: A CI job did not accept `ser` as variable name, "thought" it was a typo of `set`.

src/api/cpp/nixl_descriptors.h
aranadive: Do we plan to have more API changes soon? I would say wait until after the 0.3.1 code freeze to merge this. OTOH, since only the C++ API is changing and we have very few users of that Im okay with this if it doesnt break TRT-LLM, Nixlbench. In general, I think that we should do API changes only at major releases.

ColinNV:
This is being developed further with new requirements, will open new PR.

PR #425, opened 2025-06-04
BACKEND/UCX/IB: convert UCS_ERR_ENDPOINT_TIMEOUT to NIXL_ERR_REMOTE_DISCONNECT

evgeny-leksikov:
+ convert UCS_ERR_ENDPOINT_TIMEOUT to NIXL_ERR_REMOTE_DISCONNECT
 + improve error handling gtests with IB reducing timeouts

fix for:
```
[ RUN      ] UCX/TestErrorHandling.LoadRemoteThenFail/0
set NIXL_PLUGIN_DIR: /workspace/nixl/build-x86_64/src/plugins/ucx
W0604 11:39:17.163769   64637 ucx_utils.cpp:48] Unexpected UCX error: Endpoint timeout
../test/gtest/error_handling.cpp:252: Failure
  NIXL_ERR_REMOTE_DISCONNECT
    Which is: -10
  status
    Which is: -3
```


test/gtest/error_handling.cpp
```diff
@@ -120,7 +120,11 @@ class TestErrorHandling : public testing::TestWithParam<std::string> {
     void exchangeMetaData();
     nixlXferReqH* postXfer(enum nixl_xfer_op_t op, bool target_failure);
 
-    std::pair<bool, std::string> m_plugin_dir_backup = {false, ""};
+    ScopedEnvVar                 m_env_plugin_dir;
```
gleon99: More scoped vars -> more members here?
Maybe instead: vector/list + emplace_back
or separate type (like "scope.add")

gleon99: As it looks like something that can (and will) be reused

test/gtest/common.h
```diff
@@ -39,4 +39,32 @@ class Logger {
...
+        if (m_backup.first) {
+            setenv(m_name.c_str(), m_backup.second.c_str(), 1);
+        } else {
+            unsetenv(m_name.c_str());
+        }
+    }
+
+private:
+    std::string                  m_name;
+    std::pair<bool, std::string> m_backup;
```
gleon99: Why pair?

evgeny-leksikov: to handle case on restore when need to unset

gleon99: That's reason for the bool flag. But what "pair" gives here?

evgeny-leksikov: do you mean separate members like follows?
```
std::string m_backup;
bool        m_unset;
```

gleon99: yes, maybe.
"pair" is used to couple members that have meaning together and passed together between funcs..

evgeny-leksikov: > members that have meaning together

that was the case, but ok IMO not critical here, separated

test/gtest/common.h
```diff
@@ -39,4 +40,56 @@ class Logger {
...
+public:
+    ScopedEnvVar(const std::string &name, const std::string &value)
+    : m_name(name), m_unset(true)
+    {
+        const char* backup = getenv(name.c_str());
+
+        if (backup != nullptr) {
+            m_backup = backup;
+            m_unset  = false;
+        }
```
gleon99:
```
m_backup = getenv(name.c_str());
m_unset  = (m_backup == nullptr);
```
?

test/gtest/common.h
```diff
@@ -39,4 +40,56 @@ class Logger {
...
+
+        setenv(name.c_str(), value.c_str(), 1);
+    }
+
+    // Delete copy constructor and assignment operator
```
gleon99: Redundant comment (adds no info beyond the trivial code below it)

test/gtest/common.h
```diff
@@ -39,4 +40,56 @@ class Logger {
...
+    ScopedEnvVar(const ScopedEnvVar &other) = delete;
+    ScopedEnvVar &operator=(const ScopedEnvVar &other) = delete;
+
+    // Implement move constructor and assignment operator
```
gleon99: Same. Also no impl for assignment

test/gtest/common.h
```diff
@@ -39,4 +40,56 @@ class Logger {
...
+
+    ~ScopedEnvVar()
+    {
+        restore();
+    }
+
+private:
+    void restore()
+    {
+        // If the object is not initialized, do nothing
```
gleon99: Excessive commenting?

test/gtest/common.h
```diff
@@ -39,4 +40,56 @@ class Logger {
...
+
+        if (m_unset) {
+            unsetenv(m_name.c_str());
+        } else {
+            setenv(m_name.c_str(), m_backup.c_str(), 1);
+        }
+    }
+
+    std::string                  m_name;
+    std::string                  m_backup;
```
gleon99: backup -> old_value (or prev)

test/gtest/error_handling.cpp
```diff
@@ -19,6 +19,7 @@
 #include <gtest/gtest.h>
 #include <nixl_types.h>
 #include "nixl.h"
+#include "common.h"
```
gleon99, line 21: order

test/gtest/error_handling.cpp
```diff
@@ -120,7 +120,7 @@ class TestErrorHandling : public testing::TestWithParam<std::string> {
     void exchangeMetaData();
     nixlXferReqH* postXfer(enum nixl_xfer_op_t op, bool target_failure);
 
-    std::pair<bool, std::string> m_plugin_dir_backup = {false, ""};
+    std::vector<ScopedEnvVar>    m_env_vars;
```
gleon99: WDYT of abstracting even more?
vector+emplace_back is an implementation detail. Encapsulate it in an object, i.e `SetEnvVarContext m_context`.
Then in test ctor:
`m_context.push(key, value)` or something like that?

test/gtest/common.h
evgeny-leksikov: 1) AFAIR constructing `std::string` from `nullptr` is not defined by standard
2) `getenv` does not allocate new string

test/gtest/common.h
```diff
@@ -40,56 +41,60 @@ class Logger {
...
 
-        if (m_unset) {
-            unsetenv(m_name.c_str());
-        } else {
-            setenv(m_name.c_str(), m_backup.c_str(), 1);
-        }
+    private:
+        std::string m_name;
+        std::string m_prev_value;
+        bool        m_unset;
```
rakhmets: Please use `std::optional<std::string> m_prev_value;`

gleon99 (dismissed):
👍🏼

test/gtest/common.h
```diff
@@ -39,4 +41,59 @@ class Logger {
     }
 };
 
+class ScopedEnv {
+private:
```
rakhmets, line 42: There is a guideline to use the public before protected before private order. And to avoid multiple blocks of declarations of one access. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl16-use-a-conventional-class-member-declaration-order

Please update the class:
```
class ScopedEnv {
public:
    ...
private:
    ...
};
```

gleon99 (approved):

test/gtest/common.h
evgeny-leksikov: ok, done

rakhmets (approved):

test/gtest/common.h
```diff
@@ -39,4 +41,58 @@ class Logger {
...
+            const char* backup = getenv(name.c_str());
+
+            if (backup != nullptr) {
+                m_prev_value = backup;
+            }
+
+            setenv(name.c_str(), value.c_str(), 1);
+        }
+
+        Variable(Variable &&other)
```
yosefe: can we just delete move semantics?

test/gtest/common.h
```diff
@@ -39,4 +41,58 @@ class Logger {
...
+    {
+        m_vars.emplace(name, value);
+    }
+
+private:
+    class Variable {
+    public:
+        Variable(const std::string &name, const std::string &value)
+        : m_name(name)
+        {
```
yosefe: can we move implementation to cpp file? why does it need to be in header?

test/gtest/common.h
```diff
@@ -39,4 +41,58 @@ class Logger {
...
+            // The moved-from object should be invalidated
+            assert(other.m_name.empty());
+        }
+
+        ~Variable()
+        {
+            if (m_name.empty()) {
+                return;
+            }
```
yosefe: why needed?

test/gtest/common.h
gleon99: Common practice, rule of 5..

test/gtest/common.h
evgeny-leksikov: common.cpp does not exist yet and IMO the class is not so big to introduce this for now.

test/gtest/common.h
evgeny-leksikov: proactive protection for invalidated instance since we have move constructor

test/gtest/common.h
yosefe: we can add it, it's better that putting everything in header file

test/gtest/common.h
yosefe: why do we need to support move ?

test/gtest/common.h
```diff
@@ -39,4 +41,58 @@ class Logger {
...
+
+        Variable(const Variable &other) = delete;
+        Variable &operator=(const Variable &other) = delete;
+
+    private:
+        std::optional<std::string> m_prev_value = std::nullopt;
+        std::string                m_name;
+    };
+
+    std::stack<Variable> m_vars;
```
yosefe, line 57: wouldn't it be simpler to keep std::map<string, string> ?

yosefe (changes requested):

test/gtest/error_handling.cpp
```diff
@@ -209,32 +209,12 @@ bool TestErrorHandling::Agent::dataCmp(const TestErrorHandling::Agent& other) co
...
-    } else {
-        std::cout << "unset NIXL_PLUGIN_DIR" << std::endl;
-        unsetenv("NIXL_PLUGIN_DIR");
-    }
+TestErrorHandling::TestErrorHandling() : m_backend_name(GetParam())
+{
+    m_env.addVar("UCX_RC_TIMEOUT", "100us");
+    m_env.addVar("UCX_RC_RETRY_COUNT", "4");
+    m_env.addVar("UCX_UD_TIMEOUT", "3s");
+    m_env.addVar("NIXL_PLUGIN_DIR", std::string(BUILD_DIR) + "/src/plugins/ucx");
```
yosefe, line 217: is the env vars refactoring necessary to fix the error handling test failure?

yosefe (changes requested):

test/gtest/error_handling.cpp
rakhmets: The failures are fixed by changes in `src/utils/ucx/ucx_utils.cpp`.
Adding UCX environment variables reduces test time.

test/gtest/common.h
gleon99: IMO having the `Variable` class makes sense, as it encapsulates an optional prev value, and is cleaner than handling strings directly.
Also, we don't need random access, thus `std::stack` is reasonable.

yosefe (approved):

PR #426, opened 2025-06-04
UTILS/UCX: Very small refactoring.

ColinNV:
- Eliminate `using namespace std`.
- Consistent use of `nixl_status_t`.
- More use of `const`.

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -525,10 +525,8 @@ nixlUcxContext::memReg(void *addr, const size_t size, nixlUcxMem &mem, nixl_mem_
         }
 
         if (attr.mem_type == UCS_MEMORY_TYPE_HOST) {
-            NIXL_ERROR << "memory is detected as host, check that UCX is configured"
-                          " with CUDA support";
-            ucp_mem_unmap(ctx, mem.memh);
-            return -1;
+            NIXL_WARN << "memory is detected as host, check that UCX is configured"
+                         " with CUDA support";
         }
     }
 


src/utils/ucx/ucx_utils.cpp
```diff
@@ -28,9 +28,7 @@
 #include "ucx_utils.h"
 #include "common/nixl_log.h"
 
-using namespace std;
```
brminich, line 32: maybe it is possible to use something like
`using std::size_t;`
and maybe some other types?

ColinNV: I don't really see the necessity. I'd rather qualify things in other namespaces with their respective namespaces and go int the other direction, put the UCX backend into `namespace nixl::ucx` and then drop all the `nixlUCx` prefixes to shorten the names within the module.

brminich: ok

brminich (approved):

rakhmets (approved):

aranadive:
@yosefe @gleon99 can you review & merge?

brminich:
> @yosefe @gleon99 can you review & merge?

@yosefe has some concerns, pls wait for review

src/utils/ucx/ucx_utils.cpp
```diff
@@ -200,9 +196,9 @@ void nixlUcxEp::rkeyDestroy(nixlUcxRkey &rkey)
  * =========================================== */
 
 nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
-                                void* hdr, size_t hdr_len,
-                                void* buffer, size_t len,
-                                uint32_t flags, nixlUcxReq &req)
+                                void* hdr, std::size_t hdr_len,
```
iyastreb: Since `size_t` and `std::size_t` are the same types, why do we want to type more?
IMO this decreases code readability without adding any benefits.

ColinNV: I usually qualify everything in the standard library with `std::` for consistency, for grep-ability, and because it's more of a "historical accident" that you can just use `size_t` instead of `std::size_t` in C++. If we make an exception for `size_t` I'll change it back.

iyastreb: IMO the fact that `size_t` and `std::size_t` are equal is dictated by interoperability between C/C++. We call underlying UCX C code by passing std::size_t where size_t is expected by C backend. And we implement UCX callbacks that are invoked with size_t (and other fixed size ints). So they must be always the same, and it's guaranteed by standard.
I understand your preference, this is a question of taste to me. My personal preference is to avoid redundant stuff when it brings no benefits. But I agree to do that if the entire team agrees on that, in order to keep the code uniform. It would be nice if we agree (by voting?) on the following:
- explicit `std::` prefix for `size_t` and all fixed-width integers
- const for primitives everywhere. IMO that makes no sense. For example in standard library there are no const for primitives:
```
void* operator new  ( std::size_t count );
void resize( size_type count );
```

ColinNV: Sure, the C type `size_t` is the same as the C++ type `std::size_t`, for backwards compatibility many C++ compilers still allow `size_t` in C++ even when only including C++ headers that define it.

The thing with `const` for primitives that are function arguments is that the two function declarations
```c++
void func( std::size_t );
void func( const std::size_t );
```
are equivalent, two ways of writing the same thing. That's because with call-by-value it doesn't make a difference to the caller whether the callee function treats its copy in a local variable as `const`. So in the function implementation it's just a normal local variable that should be `const` if it can, but there's a choice of whether to add the `const` to the declaration since it doesn't do anything in that place. I usually put it there because I just put `const` "everywhere", but I can also see the logic of omitting it in the prototype and only adding it to the implementation.

iyastreb: Right, if they are equivalent, I always prefer shortest `void func(size_t)` just to keep code shorter (and therefore more readable). Anyway, this is a question of taste. But would be nice to agree on that with the team to stick to the same style

yosefe: Agree. I see most places in the code use size_t. So at least for consistency let's keep it this way. If we get concensus in #nixl-dev to change everything to std::size_t that we can do it

ColinNV: Ok!

src/utils/ucx/ucx_utils.cpp
```diff
@@ -215,10 +211,10 @@ void nixlUcxEp::rkeyDestroy(nixlUcxRkey &rkey)
...
  * =========================================== */
 
-nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
-                                void* hdr, size_t hdr_len,
-                                void* buffer, size_t len,
-                                uint32_t flags, nixlUcxReq &req)
+nixl_status_t nixlUcxEp::sendAm(const unsigned msg_id,
+                                void* hdr, const size_t hdr_len,
+                                void* buffer, const size_t len,
+                                const std::uint32_t flags, nixlUcxReq &req)
```
yosefe: std::uint32_t -> uint32_t

src/utils/ucx/ucx_utils.cpp
```diff
@@ -240,16 +236,16 @@ nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
  * Data transfer
  * =========================================== */
 
-nixl_status_t nixlUcxEp::read(uint64_t raddr, nixlUcxRkey &rk,
+nixl_status_t nixlUcxEp::read(const std::uint64_t raddr, nixlUcxRkey &rk,
```
yosefe: std::uint64_t -> uint64_t

src/utils/ucx/ucx_utils.cpp
```diff
@@ -266,8 +262,8 @@ nixl_status_t nixlUcxEp::read(uint64_t raddr, nixlUcxRkey &rk,
 }
 
 nixl_status_t nixlUcxEp::write(void *laddr, nixlUcxMem &mem,
-                               uint64_t raddr, nixlUcxRkey &rk,
-                               size_t size, nixlUcxReq &req)
+                               const std::uint64_t raddr, nixlUcxRkey &rk,
```
yosefe: std::uint64_t -> uint64_t

src/utils/ucx/ucx_utils.cpp
```diff
@@ -240,16 +236,16 @@ nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
...
 
-nixl_status_t nixlUcxEp::read(uint64_t raddr, nixlUcxRkey &rk,
+nixl_status_t nixlUcxEp::read(const std::uint64_t raddr, nixlUcxRkey &rk,
                               void *laddr, nixlUcxMem &mem,
-                              size_t size, nixlUcxReq &req)
+                              const size_t size, nixlUcxReq &req)
 {
-    nixl_status_t status = checkTxState();
+    const nixl_status_t status = checkTxState();
     if (status != NIXL_SUCCESS) {
```
gleon99, line 236: `if (const auto status = ...; status !=) { ..`

src/utils/ucx/ucx_utils.cpp
```diff
@@ -192,21 +190,19 @@ nixl_status_t nixlUcxEp::disconnect_nb()
...
  * =========================================== */
 
-int nixlUcxEp::rkeyImport(void* addr, size_t size, nixlUcxRkey &rkey)
+nixl_status_t nixlUcxEp::rkeyImport(void* addr, const size_t size, nixlUcxRkey &rkey)
 {
-    ucs_status_t status;
-
-    status = ucp_ep_rkey_unpack(eph, addr, &rkey.rkeyh);
+    const ucs_status_t status = ucp_ep_rkey_unpack(eph, addr, &rkey.rkeyh);
     if (status != UCS_OK)
```
gleon99, line 182: `if (ucp_ep_...(...) != UCS_OK) { ...`

src/utils/ucx/ucx_utils.cpp
```diff
@@ -350,11 +346,11 @@ bool nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t mt_type) noexcept
...
 }
 
-nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
-                               size_t req_size,
+nixlUcxContext::nixlUcxContext(const std::vector<std::string>& devs,
+                               const size_t req_size,
                                nixlUcxContext::req_cb_t init_cb,
                                nixlUcxContext::req_cb_t fini_cb,
-                               bool prog_thread,
+                               const bool prog_thread,
```
gleon99, line 342: There's no meaning in `const`ing a bool (or size_t) func arg..

ColinNV: It's a local variable, if it can be `const` it should be. (It doesn't mean anything in a mere function declaration.)

src/utils/ucx/ucx_utils.cpp
ColinNV: I always feel a bit ambiguous about this style, good: it limits the scope of the variable, bad: it puts multiple things in one line, but I'm happy to use it here...

src/utils/ucx/ucx_utils.cpp
```diff
@@ -263,21 +259,20 @@ nixl_status_t nixlUcxEp::read(const std::uint64_t raddr, nixlUcxRkey &rk,
...
-    ucp_request_param_t param = {
+    const ucp_request_param_t param = {
         .op_attr_mask = UCP_OP_ATTR_FIELD_MEMH |
                         UCP_OP_ATTR_FLAG_MULTI_SEND,
         .memh         = mem.memh,
     };
 
-    ucs_status_ptr_t request = ucp_put_nbx(eph, laddr, size, raddr,
+    const ucs_status_ptr_t request = ucp_put_nbx(eph, laddr, size, raddr,
                                            rk.rkeyh, &param);
```
gleon99: minor: alignment

src/utils/ucx/ucx_utils.cpp
```diff
@@ -163,18 +159,18 @@ nixlUcxEp::nixlUcxEp(ucp_worker_h worker, void* addr,
...
         throw std::runtime_error("failed to create ep");
+    }
+    setState(NIXL_UCX_EP_STATE_CONNECTED);
 }
 
  nixlUcxEp::~nixlUcxEp()
  {
-     nixl_status_t status = disconnect_nb();
-     if (status)
+     if (const nixl_status_t status = disconnect_nb(); status) {
```
yosefe: it looked better before the change

src/utils/ucx/ucx_utils.cpp
```diff
@@ -193,21 +189,19 @@ nixl_status_t nixlUcxEp::disconnect_nb()
  * RKey management
  * =========================================== */
 
-int nixlUcxEp::rkeyImport(void* addr, size_t size, nixlUcxRkey &rkey)
+nixl_status_t nixlUcxEp::rkeyImport(void* addr, const size_t size, nixlUcxRkey &rkey)
```
yosefe: why need to add const for primitive type?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -216,18 +210,17 @@ void nixlUcxEp::rkeyDestroy(nixlUcxRkey &rkey)
...
  * =========================================== */
 
-nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
-                                void* hdr, size_t hdr_len,
-                                void* buffer, size_t len,
-                                uint32_t flags, nixlUcxReq &req)
+nixl_status_t nixlUcxEp::sendAm(const unsigned msg_id,
+                                void* hdr, const size_t hdr_len,
+                                void* buffer, const size_t len,
+                                const uint32_t flags, nixlUcxReq &req)
```
yosefe: why need to add const for primitive types?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -241,22 +234,21 @@ nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
...
 
-nixl_status_t nixlUcxEp::read(uint64_t raddr, nixlUcxRkey &rk,
+nixl_status_t nixlUcxEp::read(const uint64_t raddr, nixlUcxRkey &rk,
                               void *laddr, nixlUcxMem &mem,
-                              size_t size, nixlUcxReq &req)
+                              const size_t size, nixlUcxReq &req)
 {
-    nixl_status_t status = checkTxState();
-    if (status != NIXL_SUCCESS) {
+    if (const nixl_status_t status = checkTxState(); status != NIXL_SUCCESS) {
```
yosefe: IMO it looks better when initializing the status before 'if'

src/utils/ucx/ucx_utils.cpp
ColinNV: It's a local variable, it should be `const` when it's not changed...

src/utils/ucx/ucx_utils.cpp
ColinNV: I'm happy to change it back, for once I have no preference.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -163,18 +159,19 @@ nixlUcxEp::nixlUcxEp(ucp_worker_h worker, void* addr,
     ep_params.err_handler.arg = reinterpret_cast<void*>(this);
     ep_params.address         = reinterpret_cast<ucp_address_t*>(addr);
 
-    status = ucx_status_to_nixl(ucp_ep_create(worker, &ep_params, &eph));
-    if (status == NIXL_SUCCESS)
-        setState(NIXL_UCX_EP_STATE_CONNECTED);
-    else
+    const nixl_status_t status = ucx_status_to_nixl(ucp_ep_create(worker, &ep_params, &eph));
+    if (status != NIXL_SUCCESS) {
```
rakhmets: We don't need to convert UCS status to NIXL status here.
```suggestion
    const auto status = ucp_ep_create(worker, &ep_params, &eph);
    if (status != UCS_OK) {
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -875,18 +875,20 @@ nixlUcxEngine::internalMDHelper (const nixl_blob_t &blob,
     std::vector<char> addr(size);
     nixlSerDes::_stringToBytes(addr.data(), blob, size);
 
-    bool error = false;
+    nixl_status_t status = NIXL_SUCCESS;
     for (size_t wid = 0; wid < uws.size(); wid++) {
         nixlUcxRkey rkey;
-        error = md->conn->getEp(wid)->rkeyImport(addr.data(), size, rkey);
-        if (error)
+        status = md->conn->getEp(wid)->rkeyImport(addr.data(), size, rkey);
```
rakhmets, line 881: CI fails because of this.
```suggestion
        status = md->conn->getEp (wid)->rkeyImport (addr.data(), size, rkey);
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -42,7 +40,6 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
         return NIXL_IN_PROG;
     case UCS_ERR_NOT_CONNECTED:
     case UCS_ERR_CONNECTION_RESET:
-    case UCS_ERR_ENDPOINT_TIMEOUT:
```
evgeny-leksikov: why removed?

ColinNV: Not sure what happened to that line -- it was added to main branch after I forked.

evgeny-leksikov (approved):

brminich (commented):
pls note the conflicts now

PR #427, opened 2025-06-04
CI: Enable arm build

brminich:
arb build was fixed, enable it in CI


Alexey-Rivkin (approved):

.gitlab/test_cpp.sh
dpressle, line 1: you need the same for python script .gitlab/test_python.sh

dpressle (changes requested):

.gitlab/test_cpp.sh
brminich: done

dpressle (approved):

PR #429, opened 2025-06-04
UCX: Fix UCP AM usage (do not duplicate id in the header)

brminich:
- Do not duplicate AM id in the header, am id is enough to ensure proper callback is invoked on the target
- Use correct error code for UCP AM callback

[UCP API doc](https://github.com/openucx/ucx/blob/master/src/ucp/api/ucp_def.h#L670)

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -684,7 +684,7 @@ nixlUcxEngine::connectionCheckAmCb(void *arg, const void *header,
     NIXL_ASSERT(header_length == 0) << "header_length " << header_length;
 
     if(engine->checkConn(remote_agent)) {
-        NIXL_ERROR << "Received connect AM from agent we don't recognize";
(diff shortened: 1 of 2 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -677,25 +677,15 @@ nixlUcxEngine::connectionCheckAmCb(void *arg, const void *header,
...
-        //is this the best way to ERR?
-        return UCS_ERR_INVALID_PARAM;
-    }
+    NIXL_ASSERT(!(param->recv_attr & UCP_AM_RECV_ATTR_FLAG_RNDV));
+    NIXL_ASSERT(header_length == 0) << "header_length " << header_length;
 
     if(engine->checkConn(remote_agent)) {
-        //TODO: received connect AM from agent we don't recognize
-        return UCS_ERR_INVALID_PARAM;
+        NIXL_ERROR << "Received connect AM from agent we don't recognize";
```
yosefe: can we print the remote agent name?

PR #430, opened 2025-06-04
Python: bindings optimizations and support numpy arrays for xfer_dlist

yanivz-nv:
Speedup for 1M descriptor lists:
* get_xfer_descs with numpy input: x3.7 vs list (82ms -> 22ms)
* get_xfer_descs with list input: x1.3 (82ms -> 63ms)
* get_reg_descs with numpy input: x2.7 vs get_reg_descs (146ms -> 54ms)
* get_reg_descs with list input: x1.2 (146ms -> 122ms)
* make_prepped_xfer with list: x1   (74ms -> 74ms)
* make_prepped_xfer with numpy in: x1.7 vs list (74ms -> 44ms)

### DIFF
--- a/src/api/cpp/nixl_descriptors.h
+++ b/src/api/cpp/nixl_descriptors.h
@@ -310,12 +310,6 @@ class nixlDescList {
          */
         nixlDescList<nixlBasicDesc> trim() const;
         /**
-         * @brief Convert a nixlDescList of nixlBasicDesc elements by padding
-         *        it with empty metadata. This is only specialized for nixlBasicDesc
-         *        elements.
-         */
-        nixlDescList<nixlBlobDesc> pad() const;
-        /**
          * @brief  Check if input descriptor `desc` overlaps with any descriptor
          *         within the current object, and returns its index if found.
          *
--- a/src/api/cpp/nixl_descriptors.h
+++ b/src/api/cpp/nixl_descriptors.h
@@ -310,6 +310,12 @@ class nixlDescList {
          */
         nixlDescList<nixlBasicDesc> trim() const;
         /**
+         * @brief Convert a nixlDescList of nixlBasicDesc elements by padding
+         *        it with empty metadata. This is only specialized for nixlBasicDesc
+         *        elements.
+         */
+        nixlDescList<nixlBlobDesc> pad() const;
+        /**
          * @brief  Check if input descriptor `desc` overlaps with any descriptor
          *         within the current object, and returns its index if found.
          *
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -836,7 +836,8 @@ class nixl_agent:
             a) list of 4 element tuples (address, len, device ID, meta info) alongside a mandatory memory type
             b) a tensor
             c) a list of tensors
-            d) a nixlXferDList, which is padded with empty metadata
+            d) a Nx3 2D numpy array, each row defines a single descriptor (address, len, device ID),
+               alongside a mandatory memory type. Empty metadata will be added to each descriptor.
             e) passes along if a reg_dlist is given.
 
     @param descs List of any of the above types
@@ -857,7 +858,8 @@ class nixl_agent:
         if isinstance(descs, nixlBind.nixlRegDList):
             return descs
         elif isinstance(descs, nixlBind.nixlXferDList):
-            new_descs = nixlBind.nixlRegDList(descs)
+            print("XferList type detected for registration, please use RegList")
+            new_descs = None
         elif isinstance(descs[0], tuple):
             if mem_type is not None and len(descs[0]) == 4:
                 new_descs = nixlBind.nixlRegDList(
@@ -868,6 +870,19 @@ class nixl_agent:
                 new_descs = None
             else:
                 print("4-tuple list needed for registration")
+                new_descs = None
+        elif isinstance(descs, np.ndarray):
+            if mem_type is not None and descs.ndim == 2 and descs.shape[1] == 3:
+                new_descs = nixlBind.nixlRegDList(
+                    self.nixl_mems[mem_type], descs, is_sorted
+                )
+            elif mem_type is None:
+                print("Please specify a mem type if not using Tensors")
+                new_descs = None
+            else:
+                print(
+                    "Nx3 shape required for transfer descriptor list from numpy array"
+                )
                 new_descs = None
         elif isinstance(descs, torch.Tensor):
             mem_type = "cuda" if str(descs.device).startswith("cuda") else "cpu"
@@ -883,7 +898,7 @@ class nixl_agent:
             )
         elif isinstance(descs[0], torch.Tensor):  # List[torch.Tensor]:
             tensor_type = descs[0].device
-            dlist = [(0, 0, 0, "")] * len(descs)
+            dlist = np.zeros((len(descs), 3), dtype=np.uint64)
 
             for i in range(len(descs)):
                 if descs[i].device != tensor_type:
@@ -893,7 +908,7 @@ class nixl_agent:
                 gpu_id = descs[i].get_device()
                 if gpu_id == -1:  # DRAM
                     gpu_id = 0
-                dlist[i] = (base_addr, region_len, gpu_id, "")
+                dlist[i, :] = (base_addr, region_len, gpu_id)
             mem_type = "cuda" if str(tensor_type).startswith("cuda") else "cpu"
             new_descs = nixlBind.nixlRegDList(
                 self.nixl_mems[mem_type], dlist, is_sorted
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -14,8 +14,9 @@
 # limitations under the License.
 
 import pickle
-from typing import Optional
+from typing import Optional, Union
 
+import numpy as np
 import torch
 
 import nixl._bindings as nixlBind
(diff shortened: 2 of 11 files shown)


src/bindings/python/nixl_bindings.cpp
```diff
@@ -172,27 +173,53 @@ PYBIND11_MODULE(_bindings, m) {
...
+            if (!py::dtype::of<uint64_t>().equal(descs.dtype()) && !py::dtype::of<int64_t>().equal(descs.dtype()))
+                throw std::invalid_argument("descs must be a Nx3 numpy array of uint64 or int64");
+            if (!(descs.flags() & py::array::c_style)) {
+                throw std::invalid_argument("descs must be a C-contiguous numpy array");
+            }
+            size_t n = descs.shape(0);
+            nixl_xfer_dlist_t new_list(mem, sorted, n);
+            // We assume that the Nx3 array matches the nixlBasicDesc layout so we can simply memcpy
+            std::memcpy(&new_list[0], descs.data(), descs.size() * sizeof(uint64_t));
+            if (sorted) new_list.verifySorted();
```
mkhazraee: Let's do the sorted check only in debug mode. I think we had it somewhere else too.

test/python/test_nixl_bindings.py
```diff
@@ -45,7 +47,7 @@ def test_list():
     test_list.remDesc(1)
     assert test_list.descCount() == 2
 
-    assert test_list[0] == descs[0]
+    assert np.all(test_list[0] == descs[0])
```
mkhazraee: Can we keep both tests?

src/api/cpp/nixl_descriptors.h
```diff
@@ -309,6 +309,12 @@ class nixlDescList {
          *        nixlDescList of nixlBasicDesc elements
          */
         nixlDescList<nixlBasicDesc> trim() const;
+        /**
+         * @brief Convert a nixlDescList of nixlBasicDesc elements by padding
+         *        it with empty metadata. This is only specialized for nixlBasicDesc
+         *        elements.
+         */
+        nixlDescList<nixlBlobDesc> pad() const;
```
mkhazraee: Let's do the loop inside the bindings, instead of adding to API, will see if this becomes a use case, then add it to the API.

src/api/python/_api.py
```diff
@@ -780,6 +783,17 @@ def get_xfer_descs(
...
+        elif isinstance(descs, np.ndarray):
+            if mem_type is not None and descs.shape[1] == 3:
+                new_descs = nixlBind.nixlXferDList(
+                    self.nixl_mems[mem_type], descs, is_sorted
+                )
+            elif mem_type is None:
+                print("Please specify a mem type if not using Tensors")
+                new_descs = None
+            else:
+                print("Nx3 2D numpy array needed for transfer")
```
mkhazraee: for transfer descriptor list.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -172,27 +173,53 @@ PYBIND11_MODULE(_bindings, m) {
...
         .def("__getitem__", [](nixl_xfer_dlist_t &list, unsigned int i) ->
-              std::tuple<uintptr_t, size_t, uint64_t> {
-                    std::tuple<uintptr_t, size_t, uint64_t> ret;
-                    nixlBasicDesc desc = list[i];
-                    std::get<0>(ret) = desc.addr;
-                    std::get<1>(ret) = desc.len;
-                    std::get<2>(ret) = desc.devId;
-                    return ret;
-              })
+            py::array_t<uint64_t> {
```
mkhazraee: This is a rare use case, for now let's keep it tuple to be consistent with the reg one and backward compatible, just add a comment that can be changed to numpy in future.

src/api/python/_api.py
```diff
@@ -747,7 +749,8 @@ def check_remote_metadata(
             a) list of 3 element tuples (address, len, device ID) alongside a mandatory memory type
             b) a tensor
             c) a list of tensors
-            d) passes along if an xfer_dlist is given.
+            d) a Nx3 2D numpy array
```
mkhazraee: Let's explain a tad more, other dimension is addr/len/devID in order.

src/api/python/_api.py
```diff
@@ -313,7 +314,8 @@ def make_connection(self, remote_agent: str, backends: list[str] = []):
                 this is for a loopback (local) transfer to be used for remote_side handle
 
     @param agent_name Name of the agent. It can be "NIXL_INIT_AGENT", local agent name, or remote agent name
-    @param xfer_list List of transfer descriptors, can be list of memory region tuples, tensors, or nixlXferDList.
+    @param xfer_list List of transfer descriptors, can be list of memory region tuples, tensors,
+                     Nx3 numpy array, or nixlXferDList.
```
mkhazraee: Let's also add the order for tuple or numpy is this in the description of prep_xfer_dlist.

test/python/simple_api_perf.py
mkhazraee, line 1: Let's call it desc_prep_perf.py or so to be more specific. Later we should possibly merge this with desc_example, but for now it's ok to be separate.

test/python/simple_api_perf.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def init_agent():
+    agent_config = nixl_agent_config(backends=["UCX"])
+    agent = nixl_agent("agent", agent_config)
+    return agent
+
+
+def prep_handles(agent: nixl_agent, xfer_dlist, reg_dlist, indices):
+    start = time.time()
```
mkhazraee: Maybe we can use time.perf_counter() to be more accurate, but I guess times were long enough already, just a suggestion. We have NIXL time but not exposed to python anyways.

src/api/cpp/nixl_descriptors.h
yanivz-nv: Removed.

src/api/python/_api.py
yanivz-nv: Done

src/api/python/_api.py
yanivz-nv: Done

src/api/python/_api.py
yanivz-nv: Done

src/bindings/python/nixl_bindings.cpp
yanivz-nv: When thinking about it the return value of verifySorted was never checked in the first place. Switched to assert as it feels more appropriate for debug-only checks.

src/bindings/python/nixl_bindings.cpp
yanivz-nv: Done.

test/python/simple_api_perf.py
yanivz-nv: Maybe prep_xfer_perf.py, since it also measures make_prepped_xfer?

test/python/test_nixl_bindings.py
yanivz-nv: Reverted this since we are back to return py::tuple in `__getitem__`

test/python/simple_api_perf.py
yanivz-nv: Done

src/bindings/python/nixl_bindings.cpp
```diff
@@ -172,27 +174,47 @@ PYBIND11_MODULE(_bindings, m) {
...
+                    if (!py::isinstance<py::tuple>(descs[i])) {
+                        throw py::type_error("Each descriptor must be a tuple when provided as a list");
+                    }
+                    auto desc = py::reinterpret_borrow<py::tuple>(descs[i]);
+                    if (desc.size() != 3) {
+                        throw py::value_error("Each descriptor tuple must have exactly 3 elements");
+                    }
+                    new_list[i] = nixlBasicDesc(desc[0].cast<uintptr_t>(), desc[1].cast<size_t>(), desc[2].cast<uint64_t>());
+                }
+                assert(!sorted || new_list.verifySorted());
```
mkhazraee, line 206: No asserts in the code as it can break the flow. For now you can keep it as before. My idea was in release mode we don't need to do this compute, in debug mode we do.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -229,27 +251,36 @@ PYBIND11_MODULE(_bindings, m) {
...
+                    if (!py::isinstance<py::tuple>(descs[i])) {
+                        throw py::type_error("Each descriptor must be a tuple when provided as a list");
+                    }
+                    auto desc = descs[i].cast<py::tuple>();
+                    if (desc.size() != 4) {
+                        throw py::value_error("Each descriptor tuple must have exactly 4 elements");
+                    }
+                    new_list[i] = nixlBlobDesc(desc[0].cast<uintptr_t>(), desc[1].cast<size_t>(), desc[2].cast<uint64_t>(), desc[3].cast<std::string>());
+                }
+                assert(!sorted || new_list.verifySorted());
```
mkhazraee, line 289: same point, no asserts.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -229,27 +251,36 @@ PYBIND11_MODULE(_bindings, m) {
 
     py::class_<nixl_reg_dlist_t>(m, "nixlRegDList")
         .def(py::init<nixl_mem_t, bool, int>(), py::arg("type"), py::arg("sorted")=false, py::arg("init_size")=0)
-        .def(py::init([](nixl_mem_t mem, std::vector<py::tuple> descs, bool sorted) {
+        .def(py::init([](const nixl_xfer_dlist_t &xfer_list) {
```
mkhazraee: I say let's get numpy as input here, not xfer_dlist. So similar code repeated, the conversion might cause confusion. Later we can consider to add .pad to xfer_dlist, not making a reg from xfer input, to be more explicit.

src/bindings/python/nixl_bindings.cpp
mkhazraee: In release mode it gets removed, all good, nvm.

src/api/python/_api.py
```diff
@@ -819,7 +836,9 @@ def get_xfer_descs(
             a) list of 4 element tuples (address, len, device ID, meta info) alongside a mandatory memory type
             b) a tensor
             c) a list of tensors
-            d) passes along if a reg_dlist is given.
+            d) a Nx3 2D numpy array, each row defines a single descriptor (address, len, device ID),
+               alongside a mandatory memory type. Empty metadata will be added to each descriptor.
```
mkhazraee, line 840: instead of metadata say meta info to be consistent with 2 lines above, and not confusing with agent MD.

PR #431, opened 2025-06-04
[CI] run tests on GPU - initial commit

dpressle:


dpressle:
Test run: https://nbuprod.blsm.nvidia.com/swx-media-master/blue/organizations/jenkins/NIXL%2Fnixl-ci-test/detail/nixl-ci-test/51/pipeline

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -0,0 +1,120 @@
...
+      echo "Kernel Version: $(uname -r)"
+      # print ofed info
+      echo "OFED Info: $(ofed_info -s)"
+      # print nvidia drivers info
+      echo "NVIDIA Drivers: $(lsmod | grep nvidia_peermem)"
+      echo "NVIDIA Drivers: $(lsmod | grep gdrdrv)"
+      echo "NVIDIA Drivers: $(lsmod | grep nvidia_fs)"
+      # print nvidia-smi
+      echo "NVIDIA-SMI: $(nvidia-smi)"
+      echo "NVIDIA-SMI Topo: $(nvidia-smi topo -m)"
```
yosefe: can fix in next pr:
IMO the echo XXX prefix is just confusing and it breaks the first line of the output (unless you add\n)
can just run this part with "set -x" and have the command lines and their outputs in the log

.gitlab/build.sh
```diff
@@ -85,7 +94,7 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/v1.18.0" | tar xz
           --enable-mt && \
         make -j && \
         make -j install-strip && \
-        ldconfig \
+        $SUDO ldconfig \
```
yosefe, line 98: i would expect "make -j install-strip" would also need $SUDO

meson.build
```diff
@@ -116,9 +116,16 @@ ucx_path = get_option('ucx_path')
...
   ucx_lib_path = ucx_path + '/lib'
   ucx_inc_path = ucx_path + '/include'
-  ucx_dep = declare_dependency(
-    link_args : ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
-    include_directories : include_directories(ucx_inc_path))
+  # Check if path is absolute
+  if ucx_inc_path.startswith('/')
+    ucx_dep = declare_dependency(
+      link_args : ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
+      compile_args : ['-I' + ucx_inc_path])
```
yosefe, line 123: why include_directories didn't work? is it still relevant and what was the error?

Alexey-Rivkin: Still relevant. Test build without the Meson fix failed with error:
`Tried to form an absolute path to a dir in the source tree.
You should not do that but use relative paths instead`
https://nbuprod.blsm.nvidia.com/swx-media-master/blue/organizations/jenkins/NIXL%2Fnixl-ci-test/detail/nixl-ci-test/53/pipeline/79/


Meson's `include_directories()` doesn't accept absolute paths by design. Switching to a relative pathes would involve a more complex and error-prone solution.

.ci/jenkins/lib/test-matrix.yaml
Alexey-Rivkin: [[CI] Remove echo prefixes from env info step](https://github.com/ai-dynamo/nixl/pull/431/commits/a4a8215476ddc9287d492b39d358ee336c325472)

.gitlab/build.sh
Alexey-Rivkin: It works without sudo as we install into a user-writable dir instead of system-wide.
Adding `$SUDO` for future proofing and consistency. 
[[CI] Add sudo for consistency](https://github.com/ai-dynamo/nixl/pull/431/commits/d33b1180ab33c698eb3768a2c4def05995b12607)

Alexey-Rivkin (approved):

.gitlab/build.sh
Alexey-Rivkin: Reverted.
Adding $SUDO to install-strip causes ownership issues that break subsequent tests.

Alexey-Rivkin:
Latest successful build:
https://nbuprod.blsm.nvidia.com/swx-media-master/blue/organizations/jenkins/NIXL%2Fnixl-ci-test/detail/nixl-ci-test/60/pipeline/

Alexey-Rivkin (approved):

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -0,0 +1,120 @@
...
+  - name: Build GPU Test Environment
+    parallel: false
+    run: |
+      docker build -t "${JOB_BASE_NAME}-${BUILD_ID}-${axis_index}" -f .ci/dockerfiles/Dockerfile.gpu_test --build-arg BASE_IMAGE=${image} .
+    onfail: docker image rm -f "${JOB_BASE_NAME}-${BUILD_ID}-${axis_index}"
+
+  - name: Run GPU Test Environment
+    parallel: false
+    run: |
+      docker run -dt --name "${JOB_BASE_NAME}-${BUILD_ID}-${axis_index}" \
```
Alexey-Rivkin, line 74: Is it possible to follow the `ci-demo` idiomatic approach to build and run Docker containers instead of running Docker commands here?

dpressle: How? we are running on static slave

meson.build
dpressle: Because without this fix you cant use non full path

.gitlab/build.sh
dpressle: Why you need sudo for this if we either run as root or we install to local user owned dir?

meson.build
yosefe: i see. going forward we probably want to try including ucx dependency as pkgconfig dependency since ucx exposes pkgconfig file

yosefe (approved):

PR #439, opened 2025-06-06
Python: bindings optimizations and support numpy arrays for xfer_dlist

mkhazraee:
Speedup for 1M descriptor lists:

get_xfer_descs with numpy input: x3.7 vs list (82ms -> 22ms)
get_xfer_descs with list input: x1.3 (82ms -> 63ms)
get_reg_descs with numpy input: x2.7 vs get_reg_descs (146ms -> 54ms)
get_reg_descs with list input: x1.2 (146ms -> 122ms)
make_prepped_xfer with list: x1 (74ms -> 74ms)
make_prepped_xfer with numpy in: x1.7 vs list (74ms -> 44ms)

### DIFF
--- a/examples/python/nixl_api_example.py
+++ b/examples/python/nixl_api_example.py
@@ -17,6 +17,7 @@
 
 import os
 
+import numpy as np
 import torch
 
 import nixl._utils as nixl_utils
@@ -53,6 +54,19 @@ if __name__ == "__main__":
 
     agent1_reg_descs = nixl_agent1.get_reg_descs(agent1_strings, "DRAM", is_sorted=True)
     agent1_xfer_descs = nixl_agent1.get_xfer_descs(agent1_addrs, "DRAM", is_sorted=True)
+
+    # Prefer numpy arrays for performance
+    agent1_addrs_np = np.array(agent1_addrs)
+    agent1_xfer_descs_np = nixl_agent1.get_xfer_descs(
+        agent1_addrs_np, "DRAM", is_sorted=True
+    )
+    agent1_reg_descs_np = nixl_agent1.get_reg_descs(
+        agent1_addrs_np, "DRAM", is_sorted=True
+    )
+
+    assert agent1_xfer_descs == agent1_xfer_descs_np
+    assert agent1_reg_descs == agent1_reg_descs_np
+    print(agent1_reg_descs, agent1_reg_descs_np)
 
     # Just for tensor test
     tensors = [torch.zeros(10, dtype=torch.float32) for _ in range(2)]
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -837,7 +837,7 @@ class nixl_agent:
             b) a tensor
             c) a list of tensors
             d) a Nx3 2D numpy array, each row defines a single descriptor (address, len, device ID),
-               alongside a mandatory memory type. Empty metadata will be added to each descriptor.
+               alongside a mandatory memory type. Empty meta info will be considered for each descriptor.
             e) passes along if a reg_dlist is given.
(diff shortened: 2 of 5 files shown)


src/api/python/_api.py
```diff
@@ -866,7 +898,7 @@ def get_reg_descs(
             )
         elif isinstance(descs[0], torch.Tensor):  # List[torch.Tensor]:
             tensor_type = descs[0].device
-            dlist = [(0, 0, 0, "")] * len(descs)
+            dlist = np.zeros((len(descs), 3), dtype=np.uint64)
```
aranadive, line 901: Does this need to be uint64 or uint32? Previously uou had mentioned dtype=uint32

test/python/prep_xfer_perf.py
```diff
@@ -0,0 +1,155 @@
...
+        choices=["list", "array"],
+        help="Test mode: list or array based descriptors",
+    )
+    args = parser.parse_args()
+
+    print("Using NIXL Plugins from:")
+    print(os.environ["NIXL_PLUGIN_DIR"])
+
+    # Example using nixl_agent_config
+    agent = init_agent()
```
aranadive, line 145: Pass this agent to the perf_test functions or remove?

aranadive (approved):
Minus the comments

examples/python/nixl_api_example.py
```diff
@@ -54,6 +55,19 @@
...
+    agent1_xfer_descs_np = nixl_agent1.get_xfer_descs(
+        agent1_addrs_np, "DRAM", is_sorted=True
+    )
+    agent1_reg_descs_np = nixl_agent1.get_reg_descs(
+        agent1_addrs_np, "DRAM", is_sorted=True
+    )
+
+    assert agent1_xfer_descs == agent1_xfer_descs_np
+    assert agent1_reg_descs == agent1_reg_descs_np
+    print(agent1_reg_descs, agent1_reg_descs_np)
```
roiedanino, line 69: Can we use a logger instead of `print()` please?

src/api/python/_api.py
```diff
@@ -780,6 +784,19 @@ def get_xfer_descs(
             else:
                 print("3-tuple list needed for transfer")
                 new_descs = None
+        elif isinstance(descs, np.ndarray):
+            if mem_type is not None and descs.ndim == 2 and descs.shape[1] == 3:
+                new_descs = nixlBind.nixlXferDList(
+                    self.nixl_mems[mem_type], descs, is_sorted
+                )
+            elif mem_type is None:
+                print("Please specify a mem type if not using Tensors")
```
roiedanino, line 793: same here, please change to a logger

examples/python/nixl_api_example.py
aranadive: We can clean up separately.

src/api/python/_api.py
aranadive: Same as above.

aranadive (approved):

examples/python/nixl_api_example.py
gleon99: @aranadive  we want to keep the codebase clean & robust. Using a logging library is preferred over `print()`s.
Are you planning to clean up this one?

Also, let's discuss unresolved discussions before merge please.

aranadive: In general, I agree with you. But IMO, the print cleanup should be a separate PR since the entire file needs to be updated rather than piece meal in this PR. We can create a github issue and track it.
Given a two week release schedule - we should try and resolve as soon as possible or move item to separate PR if its a non-blocker.

PR #459, opened 2025-06-12
NIXL: set env vars for best EFA performance

iyastreb:
## What?
Set env vars in NIXL plugin for best EFA performance.

## Why?
The best performance in EFA is achieved by using 4 RMA lanes, but on low message sizes this leads to significant performance degradations. Recent UCX PR (https://github.com/openucx/ucx/pull/10723) fixes this issue by adding MIN_RMA_CHUNK.

## How?
Performance impact of this change is documented here: https://confluence.nvidia.com/pages/viewpage.action?pageId=4004524984

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -58,8 +58,8 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
 
     ucs_status_t status = ucp_config_modify(config, key.data(), value.data());
     if (status != UCS_OK) {
-        NIXL_ERROR << "Failed to modify UCX config: " << key << "=" << value
-                   << ": " << ucs_status_string(status);
+        NIXL_WARN << "Failed to modify UCX config: " << key << "=" << value
+                  << ": " << ucs_status_string(status);
     } else {
         NIXL_DEBUG << "Applied UCX config from " << (env_val ? "env var" : "NIXL")
                    << ": " << key << "=" << value;
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -412,11 +412,12 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
     ucx_modify_config(ucp_config, "ADDRESS_VERSION", "v2");
     ucx_modify_config(ucp_config, "RNDV_THRESH", "inf");
 
-    if (major_version >= 1 && minor_version >= 19) {
+    unsigned ucp_version = UCP_VERSION(major_version, minor_version);
+    if (ucp_version >= UCP_VERSION(1, 19)) {
         ucx_modify_config(ucp_config, "MAX_COMPONENT_MDS", "32");
     }
 
-    if (major_version >= 1 && minor_version >= 20) {
+    if (ucp_version >= UCP_VERSION(1, 20)) {
         ucx_modify_config(ucp_config, "MAX_RMA_RAILS", "4");
     } else {
         ucx_modify_config(ucp_config, "MAX_RMA_RAILS", "2");
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -350,13 +350,6 @@ bool nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t mt_type) noexcept
     std::terminate();
 }
 
-static const char* ucx_get_max_rma_rails()
-{
-    unsigned major_version, minor_version, release_number;
-    ucp_get_version(&major_version, &minor_version, &release_number);
-    return (major_version >= 1 && minor_version >= 20) ? "4" : "2";
-}
-
 nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
                                size_t req_size,
                                nixlUcxContext::req_cb_t init_cb,
@@ -413,10 +406,21 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         ucp_config_modify(ucp_config, "NET_DEVICES", dev_str.c_str());
     }
 
-    ucx_modify_config(ucp_config, "MAX_COMPONENT_MDS", "32");
+    unsigned major_version, minor_version, release_number;
+    ucp_get_version(&major_version, &minor_version, &release_number);
+
     ucx_modify_config(ucp_config, "ADDRESS_VERSION", "v2");
     ucx_modify_config(ucp_config, "RNDV_THRESH", "inf");
-    ucx_modify_config(ucp_config, "MAX_RMA_RAILS", ucx_get_max_rma_rails());
+
+    if (major_version >= 1 && minor_version >= 19) {
+        ucx_modify_config(ucp_config, "MAX_COMPONENT_MDS", "32");
+    }
+
+    if (major_version >= 1 && minor_version >= 20) {
+        ucx_modify_config(ucp_config, "MAX_RMA_RAILS", "4");
+    } else {
+        ucx_modify_config(ucp_config, "MAX_RMA_RAILS", "2");
+    }
 
     status = ucp_init(&ucp_params, ucp_config, &ctx);
     if (status != UCS_OK) {
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -50,6 +50,22 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
     }
 }
 
+void ucx_modify_config(ucp_config_t *config, std::string_view key,
+                       std::string_view value)
+{
+    const char *env_val = std::getenv(absl::StrFormat("UCX_%s", key.data()).c_str());
+    value = env_val ? env_val : value;
+
+    ucs_status_t status = ucp_config_modify(config, key.data(), value.data());
+    if (status != UCS_OK) {
+        NIXL_ERROR << "Failed to modify UCX config: " << key << "=" << value
+                   << ": " << ucs_status_string(status);
+    } else {
+        NIXL_DEBUG << "Applied UCX config from " << (env_val ? "env var" : "NIXL")
+                   << ": " << key << "=" << value;
+    }
+}
+
 static void err_cb_wrapper(void *arg, ucp_ep_h ucp_ep, ucs_status_t status)
 {
     nixlUcxEp *ep = reinterpret_cast<nixlUcxEp*>(arg);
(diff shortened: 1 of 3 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+    // Otherwise, use 4 for UCX 1.20 and above, and 2 otherwise
+    const char *max_rma_rails = (std::getenv("UCX_MAX_RMA_RAILS") ||
+                                 std::getenv("UCX_MAX_RMA_LANES")) ? nullptr :
+                                    (UCP_API_MAJOR >= 1 && UCP_API_MINOR >= 20) ?
+                                        "4" : "2";
+    if (max_rma_rails) {
+        status = ucp_config_modify(ucp_config, "MAX_RMA_RAILS", max_rma_rails);
+        if (status != UCS_OK) {
+            NIXL_WARN << "Failed to modify MAX_RMA_RAILS: " << ucs_status_string(status);
+        }
```
rakhmets: Please put repeated code to a separate helper function.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         ucp_config_modify(ucp_config, "NET_DEVICES", dev_str.c_str());
     }
 
-    status = ucp_config_modify(ucp_config, "MAX_RMA_RAILS", "2");
+    status = ucp_config_modify(ucp_config, "UCX_MAX_COMPONENT_MDS", "32");
```
brminich: can it be a small func for setting config var and print a warning if needed?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+    }
+
+    status = ucp_config_modify(ucp_config, "UCX_ADDRESS_VERSION", "v2");
+    if (status != UCS_OK) {
+        NIXL_WARN << "Failed to modify UCX_ADDRESS_VERSION: " << ucs_status_string(status);
+    }
+
+    // If MAX_RMA_RAILS is set explicitly, then don't overwrite it
+    // Otherwise, use 4 for UCX 1.20 and above, and 2 otherwise
+    const char *max_rma_rails = (std::getenv("UCX_MAX_RMA_RAILS") ||
```
brminich: do we really need to check env vars? I had an impression env var is supposed to overwrite the value set thru API. Can you pls double check?

brminich (dismissed):

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+                                 std::getenv("UCX_MAX_RMA_LANES")) ? nullptr :
+                                    (UCP_API_MAJOR >= 1 && UCP_API_MINOR >= 20) ?
+                                        "4" : "2";
```
rakhmets: Maybe put this into constant values, smth like `ucx_max_rma_rails`, and  `ucx_max_rma_rails_v_1_20`

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+        NIXL_WARN << "Failed to modify UCX_MAX_COMPONENT_MDS: " << ucs_status_string(status);
+    }
+
+    status = ucp_config_modify(ucp_config, "UCX_ADDRESS_VERSION", "v2");
+    if (status != UCS_OK) {
+        NIXL_WARN << "Failed to modify UCX_ADDRESS_VERSION: " << ucs_status_string(status);
+    }
+
+    // If MAX_RMA_RAILS is set explicitly, then don't overwrite it
+    // Otherwise, use 4 for UCX 1.20 and above, and 2 otherwise
```
rakhmets: Maybe add helper functions instead of comments:
```C++
bool is_max_rma_rails_set();
const char* get_ucx_max_rma_lanes_versioned();
```

brminich:
pls also set `UCX_RNDV_THRESH=inf` with the comment, like "Allow to create rma_bw lanes even if error handling is requested, but MD does not support invalidation"

src/utils/ucx/ucx_utils.cpp
iyastreb: Yes
We can also check inside this function whether it's been configured already by an env var, to allow overriding

src/utils/ucx/ucx_utils.cpp
iyastreb: I checked that before creating PR, function call overwrites env var setting

src/utils/ucx/ucx_utils.cpp
iyastreb: Honestly I don't see the point, I would keep it simple

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+    const char *max_rma_rails = (std::getenv("UCX_MAX_RMA_RAILS") ||
+                                 std::getenv("UCX_MAX_RMA_LANES")) ? nullptr :
+                                    (UCP_API_MAJOR >= 1 && UCP_API_MINOR >= 20) ?
```
tvegas1: if we only check include version and not runtime version, maybe dump somewhere the value finally retained?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,9 +390,27 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+
+    status = ucp_config_modify(ucp_config, "UCX_ADDRESS_VERSION", "v2");
+    if (status != UCS_OK) {
+        NIXL_WARN << "Failed to modify UCX_ADDRESS_VERSION: " << ucs_status_string(status);
+    }
+
+    // If MAX_RMA_RAILS is set explicitly, then don't overwrite it
+    // Otherwise, use 4 for UCX 1.20 and above, and 2 otherwise
+    const char *max_rma_rails = (std::getenv("UCX_MAX_RMA_RAILS") ||
+                                 std::getenv("UCX_MAX_RMA_LANES")) ? nullptr :
```
tvegas1: maybe take env value anyways and set it explicitly, which could make things clearer?

src/utils/ucx/ucx_utils.cpp
iyastreb: Yes, I print it in the new function

src/utils/ucx/ucx_utils.cpp
rakhmets: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es45-avoid-magic-constants-use-symbolic-constants

src/utils/ucx/ucx_utils.cpp
rakhmets: I think it is better to check runtime version instead.

src/utils/ucx/ucx_utils.cpp
iyastreb: ok

src/utils/ucx/ucx_utils.cpp
iyastreb: I'm not sure how to we check runtime version, how to do that?

src/utils/ucx/ucx_utils.cpp
iyastreb: I refactored code, no need for these

src/utils/ucx/ucx_utils.cpp
rakhmets: Please use `ucp_get_version`.

src/utils/ucx/ucx_utils.cpp
iyastreb: Yes, in most cases it makes sense, but here I believe it's an overkill.
There is a comment that explains what this block does.
According to this guideline we should also define constant for major_version and minor_version, right?
```
constexpr MIN_REQUIRED_MAJOR_VERSION = 1;
constexpr MIN_REQUIRED_MINOR_VERSION = 20;
const char ucx_max_rma_rails[] = "2";
const char ucx_max_rma_rails_v_1_20[] = "4";
    ucx_modify_config(ucp_config, "MAX_RMA_RAILS",
                      (UCP_API_MAJOR >= MIN_REQUIRED_MAJOR_VERSION &&
                       UCP_API_MINOR >= MIN_REQUIRED_MINOR_VERSION) ? 
                           ucx_max_rma_rails_v_1_20 : ucx_max_rma_rails);
```
It's super verbose and much less readable
Is it really clearer than documented code?
```
    // Set 4 RMA lanes for UCX 1.20 and above, and 2 otherwise
    ucx_modify_config(ucp_config, "MAX_RMA_RAILS",
                      (UCP_API_MAJOR >= 1 && UCP_API_MINOR >= 20) ? "4" : "2");
```
Personally I don't see added value

rakhmets: Right, it is better to define constants for everything, IMO.
But you can keep them.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -390,10 +406,15 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
-    }
+    ucx_modify_config(ucp_config, "MAX_COMPONENT_MDS", "32");
+    ucx_modify_config(ucp_config, "ADDRESS_VERSION", "v2");
+    ucx_modify_config(ucp_config, "RNDV_THRESH", "inf");
+
+    // Set 4 RMA lanes for UCX 1.20 and above, and 2 otherwise
+    unsigned major_version, minor_version, release_number;
+    ucp_get_version(&major_version, &minor_version, &release_number);
+    ucx_modify_config(ucp_config, "MAX_RMA_RAILS",
+                      (major_version >= 1 && minor_version >= 20) ? "4" : "2");
```
rakhmets: Maybe put the logic into a separate function? E.g. `get_ucx_max_ram_rails`, so it will be just `ucx_modify_config(ucp_config, "MAX_RMA_RAILS", get_ucx_max_ram_rails());`

And the comment looks redundant here.

iyastreb: ok

rakhmets (dismissed):

src/utils/ucx/ucx_utils.cpp
```diff
@@ -50,6 +50,22 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
...
 
+void ucx_modify_config(ucp_config_t *config, std::string_view key,
+                       std::string_view value)
+{
+    const char *env_val = std::getenv(absl::StrFormat("UCX_%s", key.data()).c_str());
+    value = env_val ? env_val : value;
+
+    ucs_status_t status = ucp_config_modify(config, key.data(), value.data());
+    if (status != UCS_OK) {
+        NIXL_ERROR << "Failed to modify UCX config: " << key << "=" << value
```
rakhmets: Can we continue to execute the application in case of failure? If so, it should be `NIXL_WARN`.

tvegas1 (approved):

src/utils/ucx/ucx_utils.cpp
iyastreb: ok

brminich (approved):

rakhmets (approved):

src/utils/ucx/ucx_utils.cpp
```diff
@@ -413,10 +406,21 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+
     ucx_modify_config(ucp_config, "ADDRESS_VERSION", "v2");
     ucx_modify_config(ucp_config, "RNDV_THRESH", "inf");
-    ucx_modify_config(ucp_config, "MAX_RMA_RAILS", ucx_get_max_rma_rails());
+
+    if (major_version >= 1 && minor_version >= 19) {
+        ucx_modify_config(ucp_config, "MAX_COMPONENT_MDS", "32");
+    }
+
+    if (major_version >= 1 && minor_version >= 20) {
```
brminich: btw, this check is not valid for e.g. `2.0`
can you pls change that?

iyastreb: ah, right

brminich (approved):

PR #465, opened 2025-06-15
UCX: Check if UCX recognizes GPU mem correctly

brminich:
## What?
Add check that UCX correctly recognizes GPU memory

## Why?
Better debugging experience when UCX is compiled without CUDA support or fails to find some CUDA libs

### DIFF
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -536,23 +536,19 @@ int nixlUcxContext::memReg(void *addr, size_t size, nixlUcxMem &mem, nixl_mem_t 
         return -1;
     }
 
-    if (nixl_mem_type == nixl_mem_t::VRAM_SEG)
-    {
+    if (nixl_mem_type == nixl_mem_t::VRAM_SEG) {
         try {
             ucp_mem_attr_t attr;
             attr.field_mask = UCP_MEM_ATTR_FIELD_MEM_TYPE;
             status = ucp_mem_query(mem.memh, &attr);
             if (status != UCS_OK) {
-                std::ostringstream oss;
-                oss << "Failed to ucp_mem_query: " << ucs_status_string(status);
-                throw std::runtime_error(oss.str());
+                throw std::runtime_error(absl::StrFormat("Failed to ucp_mem_query: %s",
+                                         ucs_status_string(status)));
             }
 
-            if (attr.mem_type != UCS_MEMORY_TYPE_CUDA) {
-                std::ostringstream oss;
-                oss << "memory is not CUDA (" << attr.mem_type
-                    << "), check that UCX is configured with CUDA support";
-                throw std::runtime_error(oss.str());
+            if (attr.mem_type == UCS_MEMORY_TYPE_HOST) {
+                throw std::runtime_error(absl::StrFormat("memory is detected as host "
+                    "check that UCX is configured with CUDA support", attr.mem_type));
             }
         } catch (const std::runtime_error& e) {
             NIXL_ERROR << e.what();
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -516,7 +516,7 @@ absl::StatusOr<std::unique_ptr<nixlUcxEp>> nixlUcxWorker::connect(void* addr, st
  * =========================================== */
 
 
-int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
+int nixlUcxContext::memReg(void *addr, size_t size, nixlUcxMem &mem, nixl_mem_t nixl_mem_type)
 {
     //mem.uw = this;
     mem.base = addr;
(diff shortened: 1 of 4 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -529,12 +529,37 @@ int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
...
 
-    const ucs_status_t status = ucp_mem_map(ctx, &mem_params, &mem.memh);
+    ucs_status_t status = ucp_mem_map(ctx, &mem_params, &mem.memh);
     if (status != UCS_OK) {
         /* TODOL: MSW_NET_ERROR(priv->net, "failed to ucp_mem_map (%s)\n", ucs_status_string(status)); */
         return -1;
     }
 
+    if (nixl_mem_type == nixl_mem_t::VRAM_SEG)
+    {
```
iyastreb:
```suggestion
    if (nixl_mem_type == nixl_mem_t::VRAM_SEG) {
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -529,12 +529,37 @@ int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
...
+        try {
+            ucp_mem_attr_t attr;
+            attr.field_mask = UCP_MEM_ATTR_FIELD_MEM_TYPE;
+            status = ucp_mem_query(mem.memh, &attr);
+            if (status != UCS_OK) {
+                std::ostringstream oss;
+                oss << "Failed to ucp_mem_query: " << ucs_status_string(status);
+                throw std::runtime_error(oss.str());
```
iyastreb: absl::StrFormat

src/utils/ucx/ucx_utils.cpp
```diff
@@ -529,12 +529,37 @@ int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
...
+            }
+
+            if (attr.mem_type != UCS_MEMORY_TYPE_CUDA) {
+                std::ostringstream oss;
+                oss << "memory is not CUDA (" << attr.mem_type
+                    << "), check that UCX is configured with CUDA support";
+                throw std::runtime_error(oss.str());
```
iyastreb: absl::StrFormat

src/utils/ucx/ucx_utils.cpp
```diff
@@ -529,12 +529,37 @@ int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
...
+            ucp_mem_attr_t attr;
+            attr.field_mask = UCP_MEM_ATTR_FIELD_MEM_TYPE;
+            status = ucp_mem_query(mem.memh, &attr);
+            if (status != UCS_OK) {
+                std::ostringstream oss;
+                oss << "Failed to ucp_mem_query: " << ucs_status_string(status);
+                throw std::runtime_error(oss.str());
+            }
+
+            if (attr.mem_type != UCS_MEMORY_TYPE_CUDA) {
```
yosefe: minor: maybe also check ROCM or other GPU memory types. or fail if memory == HOST.

brminich: fixed

src/utils/ucx/ucx_utils.cpp
```diff
@@ -529,12 +529,33 @@ int nixlUcxContext::memReg(void *addr, std::size_t size, nixlUcxMem &mem)
...
         return -1;
     }
 
+    if (nixl_mem_type == nixl_mem_t::VRAM_SEG) {
+        try {
+            ucp_mem_attr_t attr;
+            attr.field_mask = UCP_MEM_ATTR_FIELD_MEM_TYPE;
+            status = ucp_mem_query(mem.memh, &attr);
+            if (status != UCS_OK) {
+                throw std::runtime_error(absl::StrFormat("Failed to ucp_mem_query: %s",
```
iyastreb: Wouldn't it be easier to just log an error here and return?
In fact the only thing which is "reused" is ucp_mem_unmap.

Or maybe:
```
if (nixl_mem_type == nixl_mem_t::VRAM_SEG) {
    // Guard to ensure ucp_mem_unmap is called on early return (error path)
    auto mem_guard = make_scope_guard([&] {
        ucp_mem_unmap(ctx, mem.memh);
    });

    ucp_mem_attr_t attr{.field_mask = UCP_MEM_ATTR_FIELD_MEM_TYPE};
    status = ucp_mem_query(mem.memh, &attr);
    if (status != UCS_OK) {
        NIXL_ERROR << absl::StrFormat("Failed to ucp_mem_query: %s", ucs_status_string(status));
        return -1;
    }

    if (attr.mem_type == UCS_MEMORY_TYPE_HOST) {
        NIXL_ERROR << "memory is detected as host; check that UCX is configured with CUDA support";
        return -1;
    }

    mem_guard.dismiss();
}

iyastreb (approved):

src/utils/ucx/ucx_utils.cpp
brminich: can be done this way, but why is it better?

yosefe: throwing and catching exception in same block seems weird

brminich: fixed

yosefe (approved):

PR #463, opened 2025-06-14
nixlbench: Add GPUNETIO debs to dockerfile

aranadive:
## What?
Updates nixlbench Dockerfile to build GPUNetIO plugin correctly.

## Why?
GPUNetIO plugin is part of NIXL and needs to be tested via nixlbench.

## How?
Add DOCA debians to build GPUNETIO. Add those to keep in line with the nixl container.

PR #466, opened 2025-06-16
contrib: Remove the lib.real from the LD_LIBRARY path

aranadive:
## What?
Remove the lib.real from the LD_LIBRARY path

## Why?
Keeping lib.real in the env var LD_LIBRARY_PATH makes a mess for the CUDA scripts to setup the symlinks for the nvidia runtime. This can cause the error 803 to be displayed.

Unfortunately, we need the libcuda.so.1 when repairing the wheel, so move the ENV command there and restore the old LD_LIBRARY_PATH after finishing.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -76,10 +76,6 @@ RUN wget --tries=3 --waitretry=5 \
 WORKDIR /workspace/nixl
 COPY . /workspace/nixl
 
-# LD_LIBRARY_PATH is needed for auditwheel to find libcuda.so.1
-# Set incorrectly to `compat/lib` in cuda-dl-base image
-ENV LD_LIBRARY_PATH=/usr/local/cuda/compat/lib.real:/usr/local/lib:$LD_LIBRARY_PATH
-
 ENV VIRTUAL_ENV=/workspace/nixl/.venv
 RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
     # pybind11 pip install needed for ubuntu 22.04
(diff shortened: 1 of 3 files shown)


contrib/Dockerfile
```diff
@@ -150,3 +163,7 @@ RUN uv pip install auditwheel && \
     uv run auditwheel repair /tmp/dist/nixl-*cp31*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
 
 RUN uv pip install dist/nixl-*cp${DEFAULT_PYTHON_VERSION//./}*.whl
+
+#Remove the cuda setupp
+RUN rm -f /usr/local/cuda/compat/lib
+ENV LD_LIBRARY_PATH=$OLD_LD_PATH
```
iyastreb, line 157: Should we make the same change in nixlbench Dockerfile?

aranadive: I did already .. see #463

aranadive: Actually since we dont need to build wheels for nixlbench I have removed the lib.real ref there.

contrib/Dockerfile
```diff
@@ -76,8 +76,9 @@ RUN wget --tries=3 --waitretry=5 \
 WORKDIR /workspace/nixl
 COPY . /workspace/nixl
 
-# LD_LIBRARY_PATH is needed for auditwheel to find libcuda.so.1
-# Set incorrectly to `compat/lib` in cuda-dl-base image
+# This is only to find libcuda.so.1 correctly for build, removed once done
```
rakhmets, line 79: I think the comment might be confusing. `<cuda_runtime_install_path>/lib64/stubs` contains stub libraries, including `libcuda.so`. This should be enough for building NIXL and/or UCX. And the real library is required for the package creation using `auditwheel`. It seems that the real issue is that we need a driver level library during the package creation. Current fix is ok as a workaround. However, we should try to avoid that dependency. Because, including the path to `LD_LIBRARY_PATH` may work for one base image, and may not for another.

rakhmets (approved):

contrib/Dockerfile
aranadive: Yeah I agree nut probably more investigation is required on how to avoid the lib.real dependency if possible. Auditwheel repair does ldd to see dependencies, includes them and modifies rpath AFAICT. I will merge this for now then.

PR #472, opened 2025-06-17
nixlbench: Add GPUNETIO debs to dockerfile - 0.3.1

aranadive:
## What?
Updates nixlbench Dockerfile to build GPUNetIO plugin correctly.

## Why?
GPUNetIO plugin is part of NIXL and needs to be tested via nixlbench.

## How?
Add DOCA debians to build GPUNETIO. Add those to keep in line with the nixl container.


yosefe:
@pvijayakrish this is ready for merge

PR #473, opened 2025-06-17
contrib: Remove the lib.real from the LD_LIBRARY path (#466)

aranadive:
## What?
Remove the lib.real from the LD_LIBRARY path

## Why?
Keeping lib.real in the env var LD_LIBRARY_PATH makes a mess for the CUDA scripts to setup the symlinks for the nvidia runtime. This can cause the error 803 to be displayed.

Unfortunately, we need the libcuda.so.1 when repairing the wheel, so move the ENV command there and restore the old LD_LIBRARY_PATH after finishing.

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -90,8 +90,9 @@ COPY --from=nixlbench . /workspace/nixlbench
 
 WORKDIR /workspace/nixl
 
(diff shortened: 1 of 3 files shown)


cliffwoolley:
I recommend closing this PR in favor of  https://github.com/ai-dynamo/nixl/pull/475 -- though note the different target branches.  Seems #475 likely would also need to be CP'd to this release branch.

aranadive:
Closing since a cleaner approach was found and that should be merged.

PR #477, opened 2025-06-18
feat: Enable build with stub api in rust bindings

jthomson04:
In Dynamo, we need to run commands like `cargo check` on the block manager in CI. The block manager requires the `nixl-sys` crate. Because of this, we have to build inside a container where NIXL is installed, even when we don't need to call any of the functions. This MR adds a `stub-api` feature that stubs out all api functions, so we can build the block manager even when NIXL itself isn't installed.

### DIFF
--- a/src/bindings/rust/README.md
+++ b/src/bindings/rust/README.md
@@ -25,6 +25,20 @@ The bindings can be built using Cargo, Rust's package manager:
 cargo build
 ```
 
+### Building with Stubs (No NIXL Library Required)
+
+You can compile the bindings using stub implementations that don't require the actual NIXL library to be installed. This is useful for:
+- Development environments where NIXL isn't available
+- CI/CD pipelines
+- Building documentation
+
+```bash
+# Build with stub API
+cargo build --features stub-api
+```
+
+**Important**: When using stubs, any attempt to actually call NIXL functions at runtime will print an error message and abort the program. The stubs are only meant for compilation, not execution.
+
 ### Environment Variables
 
 - `NIXL_PREFIX`: Path to the NIXL installation (default: `/opt/nvidia/nvda_nixl`)
@@ -46,3 +60,5 @@ Note that multithreading is disabled because NIXL might deadlock.
 ```bash
 cargo test -- --test-threads=1
 ```
+
+**Note**: Tests cannot be run with the `stub-api` feature as they require actual NIXL functionality.
--- a/src/bindings/rust/build.rs
+++ b/src/bindings/rust/build.rs
@@ -65,74 +65,111 @@ fn get_arch() -> String {
 }
 
 fn main() {
-    let nixl_root_path =
-        env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
-    let nixl_include_path = format!("{}/include", nixl_root_path);
+    // Check if we're building with stub API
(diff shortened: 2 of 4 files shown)


src/bindings/rust/README.md
```diff
@@ -25,6 +25,20 @@ The bindings can be built using Cargo, Rust's package manager:
...
+- CI/CD pipelines
+- Building documentation
+
+```bash
+# Build with stub API
+cargo build --features stub-api
+```
+
+**Important**: When using stubs, any attempt to actually call NIXL functions at runtime will print an error message and abort the program. The stubs are only meant for compilation, not execution.
+
```
roiedanino:
```suggestion
**Important**: When using stubs, any attempt to actually call NIXL functions at runtime will print an error message and abort the program.
- The stubs are only meant for compilation, not execution.

```

src/bindings/rust/build.rs
```diff
@@ -65,74 +65,111 @@ fn get_arch() -> String {
...
+        bindgen::Builder::default()
+            .header("wrapper.h")
+            .generate()
+            .expect("Unable to generate bindings")
+            .write_to_file(out_path.join("bindings.rs"))
+            .expect("Couldn't write bindings!");
+
+        // Tell cargo to invalidate the built crate whenever the stubs change
+        println!("cargo:rerun-if-changed=stubs.cpp");
+        println!("cargo:rerun-if-changed=wrapper.h");
```
roiedanino: Lets put this flow in some function for readability

src/bindings/rust/stubs.cpp
```diff
@@ -41,7 +41,7 @@ struct nixl_capi_notif_map_s { /* empty */ };
 nixl_capi_status_t
 nixl_capi_stub_abort()
 {
-  printf("nixl error: detected use of the NIXL C API's stub; please update the LD_LIBRARY_PATH to include the nixl library\n");
+  printf("nixl error: detected use of the NIXL C API's stub; if you want to use NIXL, don't use the stub-api feature.\n");
```
roiedanino: Maybe we can use NIXL_WARN instead of printf?

jthomson04: Hmm. That would require us to to include "common/nixl_log.h", which might not be available if NIXL isn't installed.

roiedanino: Can we at least use `std::cerr` instead of printf?

src/bindings/rust/build.rs
```diff
@@ -64,66 +64,79 @@ fn get_arch() -> String {
...
+        println!("cargo:rustc-link-lib=dylib=ucx_utils");
+
+        // Link against C++ standard library
+        println!("cargo:rustc-link-lib=dylib=stdc++");
+
+        // Tell cargo to invalidate the built crate whenever the wrapper changes
+        println!("cargo:rustc-link-search=native={}", nixl_lib_path);
+        println!("cargo:rerun-if-changed=wrapper.h");
+        println!("cargo:rerun-if-changed=wrapper.cpp");
+        println!("cargo:rerun-if-env-changed=HAVE_ETCD");
```
roiedanino: Can we put each of the if / else branches in a function? The if branch for the least?

roiedanino (approved):

jthomson04:
Can we please re-run CI? Seems like the CI runner ran into some issue

PR #479, opened 2025-06-18
Object storage plugin

w1ldptr:
## What?
Implement OBJ storage plugin for AWS S3 generic objects and accompanying tests.

## Why?
NIXL needs a plugin that can handle OBJ_SEG memory type. AWS S3 seems like an obvious first choice and meets customer demands.

## How?
Use AWS SDK async API for performance. Provide S3Client with executor based on Asio thread pool to establish an upper bound of resource usage (number of threads).

Validate the functionality with new Gtest-based _unit_ tests with mocked S3 wrapper. The tests don't have any external dependencies and can be executed with sanitizers enabled.

### DIFF
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -169,10 +169,6 @@ nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
         const auto &local_desc = local[i];
         const auto &remote_desc = remote[i];
 
-        NIXL_ASSERT (local_desc.len == remote_desc.len)
-            << "Size mismatch for descriptor " << i << ": local=" << local_desc.len
-            << ", remote=" << remote_desc.len;
-
         auto obj_key_search = dev_id_to_obj_key_.find (remote_desc.devId);
         if (obj_key_search == dev_id_to_obj_key_.end()) {
             NIXL_ERROR << "No object key found for device ID: " << remote_desc.devId;
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -40,6 +40,11 @@ isValidPrepXferParams (const nixl_xfer_op_t &operation,
                        const nixl_meta_dlist_t &remote,
                        const std::string &remote_agent,
                        const std::string &local_agent) {
+    if (operation != NIXL_WRITE && operation != NIXL_READ) {
+        NIXL_ERROR << absl::StrFormat ("Error: Invalid operation type: %d", operation);
+        return false;
+    }
+
     if (remote_agent != local_agent) {
         NIXL_ERROR << absl::StrFormat (
             "Error: Remote agent must match the requesting agent (%s). Got %s",
@@ -210,7 +215,7 @@ nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
                                             status_promise->set_value (success ? NIXL_SUCCESS :
                                                                                  NIXL_ERR_BACKEND);
                                         });
-        else if (operation == NIXL_READ)
+        else
             s3_client_->GetObjectAsync (obj_key_search->second,
                                         data_ptr,
                                         data_len,
@@ -219,8 +224,6 @@ nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
                                             status_promise->set_value (success ? NIXL_SUCCESS :
                                                                                  NIXL_ERR_BACKEND);
                                         });
-        else
-            return NIXL_ERR_INVALID_PARAM;
     }
 
     return NIXL_IN_PROG;
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -23,8 +23,16 @@
 #include <future>
 #include <vector>
 #include <chrono>
+#include <algorithm>
 
 namespace {
+
+std::size_t
+getNumThreads (nixl_b_params_t *custom_params) {
+    return custom_params && custom_params->count ("num_threads") > 0 ?
+        std::stoul (custom_params->at ("num_threads")) :
+        std::max (1u, std::thread::hardware_concurrency() / 2);
+}
 
 bool
 isValidPrepXferParams (const nixl_xfer_op_t &operation,
@@ -77,7 +85,6 @@ public:
                 return NIXL_IN_PROG;
             }
         }
-
         return NIXL_SUCCESS;
     }
 };
@@ -104,7 +111,8 @@ public:
 
 nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params)
     : nixlBackendEngine (init_params),
-      executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
+      executor_ (
+          std::make_shared<AsioThreadPoolExecutor> (getNumThreads (init_params->customParams))),
       s3_client_ (std::make_shared<AwsS3Client> (init_params->customParams, executor_)) {
     NIXL_INFO << "Object storage backend initialized with S3 client wrapper";
 }
@@ -115,7 +123,6 @@ nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params,
       executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
       s3_client_ (s3_client) {
     s3_client_->setExecutor (executor_);
-
     NIXL_INFO << "Object storage backend initialized with injected S3 client";
 }
 
@@ -182,7 +189,8 @@ nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
 
         auto obj_key_search = dev_id_to_obj_key_.find (remote_desc.devId);
         if (obj_key_search == dev_id_to_obj_key_.end()) {
-            NIXL_ERROR << "No object key found for device ID: " << remote_desc.devId;
+            NIXL_ERROR << "The object segment key " << remote_desc.devId
+                       << " is not registered with the backend";
             return NIXL_ERR_INVALID_PARAM;
         }
 
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -84,8 +84,11 @@ public:
 
 class nixlObjMetadata : public nixlBackendMD {
 public:
-    nixlObjMetadata(nixl_mem_t nixl_mem, uint64_t dev_id, std::string obj_key)
-        : nixlBackendMD (true), nixl_mem (nixl_mem), dev_id (dev_id), obj_key (obj_key) {}
+    nixlObjMetadata (nixl_mem_t nixl_mem, uint64_t dev_id, std::string obj_key)
+        : nixlBackendMD (true),
+          nixl_mem (nixl_mem),
+          dev_id (dev_id),
+          obj_key (obj_key) {}
     ~nixlObjMetadata() = default;
 
     nixl_mem_t nixl_mem;
@@ -124,13 +127,17 @@ nixl_status_t
 nixlObjEngine::registerMem (const nixlBlobDesc &mem,
                             const nixl_mem_t &nixl_mem,
                             nixlBackendMD *&out) {
-    if (nixl_mem != OBJ_SEG) return NIXL_SUCCESS;
+    auto supported_mems = getSupportedMems();
+    if (std::find (supported_mems.begin(), supported_mems.end(), nixl_mem) == supported_mems.end())
+        return NIXL_ERR_NOT_SUPPORTED;
 
-    std::unique_ptr<nixlObjMetadata> obj_md = std::make_unique<nixlObjMetadata> (
-        nixl_mem, mem.devId, mem.metaInfo.empty() ? std::to_string (mem.devId) : mem.metaInfo);
-    dev_id_to_obj_key_[mem.devId] = obj_md->obj_key;
+    if (nixl_mem == OBJ_SEG) {
+        std::unique_ptr<nixlObjMetadata> obj_md = std::make_unique<nixlObjMetadata> (
+            nixl_mem, mem.devId, mem.metaInfo.empty() ? std::to_string (mem.devId) : mem.metaInfo);
+        dev_id_to_obj_key_[mem.devId] = obj_md->obj_key;
+        out = obj_md.release();
+    }
 
-    out = obj_md.release();
     return NIXL_SUCCESS;
 }
 
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -24,6 +24,8 @@
 #include <vector>
 #include <chrono>
 
+namespace {
+
 bool
 isValidPrepXferParams (const nixl_xfer_op_t &operation,
                        const nixl_meta_dlist_t &local,
@@ -90,6 +92,8 @@ public:
     uint64_t dev_id;
     std::string obj_key;
 };
+
+} // namespace
 
 // -----------------------------------------------------------------------------
 // Obj Engine Implementation
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -45,13 +45,11 @@ isValidPrepXferParams (const nixl_xfer_op_t &operation,
         return false;
     }
 
-    if (remote_agent != local_agent) {
-        NIXL_ERROR << absl::StrFormat (
-            "Error: Remote agent must match the requesting agent (%s). Got %s",
+    if (remote_agent != local_agent)
+        NIXL_WARN << absl::StrFormat (
+            "Warning: Remote agent doesn't match the requesting agent (%s). Got %s",
             local_agent,
             remote_agent);
-        return false;
-    }
 
     if (local.getType() != DRAM_SEG) {
         NIXL_ERROR << absl::StrFormat ("Error: Local memory type must be DRAM_SEG, got %d",
@@ -122,6 +120,7 @@ nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params)
     NIXL_INFO << "Object storage backend initialized with S3 client wrapper";
 }
 
+// Used for testing to inject a mock S3 client dependency
 nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params,
                               std::shared_ptr<IS3Client> s3_client)
     : nixlBackendEngine (init_params),
@@ -206,6 +205,8 @@ nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
         size_t data_len = local_desc.len;
         size_t offset = remote_desc.addr;
 
+        // S3 client interface signals completion via a callback, but NIXL API polls request handle
+        // for the status code. Use future/promise pair to bridge the gap.
         if (operation == NIXL_WRITE)
             s3_client_->PutObjectAsync (obj_key_search->second,
                                         data_ptr,
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -1,0 +1,225 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "obj_backend.h"
+#include "common/nixl_log.h"
+#include "nixl_types.h"
+#include <absl/strings/str_format.h>
+#include <memory>
+#include <future>
+#include <vector>
+#include <chrono>
+
+bool
+isValidPrepXferParams (const nixl_xfer_op_t &operation,
+                       const nixl_meta_dlist_t &local,
+                       const nixl_meta_dlist_t &remote,
+                       const std::string &remote_agent,
+                       const std::string &local_agent) {
+    if (remote_agent != local_agent) {
+        NIXL_ERROR << absl::StrFormat (
+            "Error: Remote agent must match the requesting agent (%s). Got %s",
+            local_agent,
+            remote_agent);
+        return false;
+    }
+
+    if (local.getType() != DRAM_SEG) {
+        NIXL_ERROR << absl::StrFormat ("Error: Local memory type must be DRAM_SEG, got %d",
+                                       local.getType());
+        return false;
+    }
+
+    if (remote.getType() != OBJ_SEG) {
+        NIXL_ERROR << absl::StrFormat ("Error: Remote memory type must be OBJ_SEG, got %d",
+                                       remote.getType());
+        return false;
+    }
+
+    return true;
+}
+
+class nixlObjBackendReqH : public nixlBackendReqH {
+public:
+    nixlObjBackendReqH() = default;
+    ~nixlObjBackendReqH() = default;
+
+    std::vector<std::future<nixl_status_t>> status_futures_;
+
+    nixl_status_t
+    getOverallStatus() {
+        while (!status_futures_.empty()) {
+            if (status_futures_.back().wait_for (std::chrono::seconds (0)) ==
+                std::future_status::ready) {
+                auto current_status = status_futures_.back().get();
+                if (current_status != NIXL_SUCCESS) {
+                    status_futures_.clear();
+                    return current_status;
+                }
+                status_futures_.pop_back();
+            } else {
+                return NIXL_IN_PROG;
+            }
+        }
+
+        return NIXL_SUCCESS;
+    }
+};
+
+class nixlObjMetadata : public nixlBackendMD {
+public:
+    nixlObjMetadata(nixl_mem_t nixl_mem, uint64_t dev_id, std::string obj_key)
+        : nixlBackendMD (true), nixl_mem (nixl_mem), dev_id (dev_id), obj_key (obj_key) {}
+    ~nixlObjMetadata() = default;
+
+    nixl_mem_t nixl_mem;
+    uint64_t dev_id;
+    std::string obj_key;
+};
+
+// -----------------------------------------------------------------------------
+// Obj Engine Implementation
+// -----------------------------------------------------------------------------
+
+nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params)
+    : nixlBackendEngine (init_params),
+      executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
+      s3_client_ (std::make_shared<AwsS3Client> (init_params->customParams, executor_)) {
+    NIXL_INFO << "Object storage backend initialized with S3 client wrapper";
+}
+
+nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params,
+                              std::shared_ptr<IS3Client> s3_client)
+    : nixlBackendEngine (init_params),
+      executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
+      s3_client_ (s3_client) {
+    s3_client_->setExecutor (executor_);
+
+    NIXL_INFO << "Object storage backend initialized with injected S3 client";
+}
+
+nixlObjEngine::~nixlObjEngine() {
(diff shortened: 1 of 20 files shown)


src/plugins/obj/obj_s3_client.cpp
```diff
@@ -0,0 +1,212 @@
...
+        request,
+        [callback, data_ptr, data_len] (
+            const Aws::S3::S3Client *client,
+            const Aws::S3::Model::GetObjectRequest &req,
+            const Aws::S3::Model::GetObjectOutcome &outcome,
+            const std::shared_ptr<const Aws::Client::AsyncCallerContext> &context) {
+            if (outcome.IsSuccess()) {
+                auto &stream = outcome.GetResult().GetBody();
+                stream.seekg (0, std::ios::beg);
+                stream.read (reinterpret_cast<char *> (data_ptr), data_len);
```
barneuman: can we avoid this copy by using SetResponseStreamFactory?

src/plugins/obj/obj_s3_client.h
```diff
@@ -0,0 +1,124 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef OBJ_S3_CLIENT_H
+#define OBJ_S3_CLIENT_H
+
+#include <functional>
```
barneuman, line 21: Please clean up the include section a bit.
Please move what you can to the CPP files and remove unnecessary includes.

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        auto status_promise = std::make_shared<std::promise<nixl_status_t>>();
+        req_h->status_futures_.push_back (status_promise->get_future());
+
+        uintptr_t data_ptr = local_desc.addr;
+        size_t data_len = local_desc.len;
+        size_t offset = remote_desc.addr;
+
+        if (operation == NIXL_WRITE)
```
barneuman, line 210: Can we verify that the operation type is valid in isValidPrepXferParams (READ/WRITE), and assume it's either READ or WRITE here?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,225 @@
...
+                         const std::string &remote_agent,
+                         nixlBackendReqH *&handle,
+                         const nixl_opt_b_args_t *opt_args) const {
+    nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *> (handle);
+
+    for (int i = 0; i < local.descCount(); ++i) {
+        const auto &local_desc = local[i];
+        const auto &remote_desc = remote[i];
+
+        NIXL_ASSERT (local_desc.len == remote_desc.len)
```
barneuman: This check is covered in the agent; you can remove

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,225 @@
...
+    if (!isValidPrepXferParams (operation, local, remote, remote_agent, localAgent))
+        return NIXL_ERR_INVALID_PARAM;
+
+    auto req_h = std::make_unique<nixlObjBackendReqH>();
+    handle = req_h.release();
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlObjEngine::postXfer (const nixl_xfer_op_t &operation,
```
barneuman, line 182: What's the expected behavior in case of failure? Some of the transfers have already been submitted, and we can easily avoid it if we want to.

subprojects/asio
barneuman, line 1: Do we need this? Doesn't the wrap file cover it?

src/plugins/obj/obj_s3_client.h
```diff
@@ -0,0 +1,124 @@
...
+
+    void
+    GetObjectAsync (std::string_view key,
+                    uintptr_t data_ptr,
+                    size_t data_len,
+                    size_t offset,
+                    GetObjectCallback callback) override;
+
+private:
+    std::unique_ptr<Aws::SDKOptions, std::function<void (Aws::SDKOptions *)>> aws_options_;
```
barneuman, line 110: can we store the Aws::SDKOptions as a regular value member to avoid the need to add a custom deleter and just call shutdown in the destructor? IMO it will simplify the code

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,509 @@
...
+    ASSERT_EQ (obj_engine_->registerMem (local_desc, DRAM_SEG, local_metadata), NIXL_SUCCESS);
+    ASSERT_EQ (obj_engine_->registerMem (remote_desc, OBJ_SEG, remote_metadata), NIXL_SUCCESS);
+
+    nixl_meta_dlist_t local_descs (DRAM_SEG, false);
+    nixl_meta_dlist_t remote_descs (OBJ_SEG, false);
+
+    std::vector<char> test_buffer (1024);
+    for (size_t i = 0; i < test_buffer.size(); ++i) {
+        test_buffer[i] = static_cast<char> ('T' + (i % 3));
+    }
```
barneuman: Do we use this data somehow? I think you can remove this for loop

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,509 @@
...
+    testAsyncTransferWithControlledExecution (NIXL_WRITE);
+}
+
+TEST_F (ObjTestFixture, MultiDescriptorWrite) {
+    testMultiDescriptorTransfer (NIXL_READ);
+}
+
+TEST_F (ObjTestFixture, MultiDescriptorRead) {
+    testMultiDescriptorTransfer (NIXL_WRITE);
+}
```
barneuman:
```suggestion
TEST_F (ObjTestFixture, MultiDescriptorWrite) {
    testMultiDescriptorTransfer (NIXL_WRITE);
}

TEST_F (ObjTestFixture, MultiDescriptorRead) {
    testMultiDescriptorTransfer (NIXL_READ);
}
```

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+        nixl_meta_dlist_t local_descs (DRAM_SEG, false);
+        nixl_meta_dlist_t remote_descs (OBJ_SEG, false);
+
+        std::vector<char> test_buffer (1024);
+
+        if (operation == NIXL_WRITE) {
+            for (size_t i = 0; i < test_buffer.size(); ++i) {
+                test_buffer[i] = static_cast<char> ('X' + (i % 3));
+            }
```
barneuman: not sure why this is needed, the test doesn't write to the obj, and therefore we never check it

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,509 @@
...
+        remote_desc.devId = 2;
+        remote_desc.metaInfo = (operation == NIXL_READ) ? "test-read-key" : "test-write-key";
+
+        nixlBackendMD *local_metadata = nullptr;
+        nixlBackendMD *remote_metadata = nullptr;
+
+        ASSERT_EQ (obj_engine_->registerMem (local_desc, DRAM_SEG, local_metadata), NIXL_SUCCESS);
+        ASSERT_EQ (obj_engine_->registerMem (remote_desc, OBJ_SEG, remote_metadata), NIXL_SUCCESS);
+
+        nixl_meta_dlist_t local_descs (DRAM_SEG, false);
```
barneuman: The default is false. Please apply to all.
```suggestion
        nixl_meta_dlist_t local_descs (DRAM_SEG);
```

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,509 @@
...
+#include <vector>
+#include <functional>
+
+#include "obj_s3_client.h"
+#include "obj_backend.h"
+#include "obj_executor.h"
+
+namespace gtest::obj {
+
+class MockS3Client : public IS3Client {
```
barneuman, line 30: Is this supposed to mimic the AwsS3Client? or is it a mock for a general S3 client?

src/plugins/obj/obj_s3_client.h
w1ldptr: Using automatic memory management instead of relying on programmer to properly clean up all resources manually in the destructor _is a simplification_.

Also, SDKOptions must be initialized before using any of the SDK classes (like S3Client on the next line) and destroyed after them. Cleaning up manually in the destructor would cause automatically-destroyed S3Client to outlive SDKOptions.

subprojects/asio
w1ldptr: Apparently we do (same as for Abseil).

test/gtest/unit/obj/obj.cpp
w1ldptr: Not sure what you mean by "mimic". IS3Client is an interface that implements functionality that Obj plugin requires to interact with S3. One of the implementations is AWS SDK wrapper that Obj plugin is using, another one is a mock implementation for tests.

test/gtest/unit/obj/obj.cpp
w1ldptr: removing.

test/gtest/unit/obj/obj.cpp
w1ldptr: ack

src/plugins/obj/obj_backend.cpp
w1ldptr: I think the behavior is as with any other plugins - submitted transfers will run their course. What do you mean by "easily avoid"?

src/plugins/obj/obj_backend.cpp
w1ldptr: removing

src/plugins/obj/obj_backend.cpp
w1ldptr: "operation" is not a part of descriptor lists, but a standalone function parameter. User can pass a value to prepXfer() where isValidPrepXferParams() is called and then another value here to postXfer(). What is the issue with verifying it here?

src/plugins/obj/obj_s3_client.h
w1ldptr: ack

src/plugins/obj/obj_s3_client.cpp
w1ldptr: Good idea, I'll explore that approach and get back to you.

w1ldptr: Yeah, I think I figured out how to implement it.

mamar123 (dismissed):

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+namespace {
+
+bool
+isValidPrepXferParams (const nixl_xfer_op_t &operation,
+                       const nixl_meta_dlist_t &local,
+                       const nixl_meta_dlist_t &remote,
+                       const std::string &remote_agent,
+                       const std::string &local_agent) {
+    if (remote_agent != local_agent) {
```
vvenkates27: Is this check neccessary? Where does this affect a storage plugin?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+public:
+    nixlObjBackendReqH() = default;
+    ~nixlObjBackendReqH() = default;
+
+    std::vector<std::future<nixl_status_t>> status_futures_;
+
+    nixl_status_t
+    getOverallStatus() {
+        while (!status_futures_.empty()) {
+            if (status_futures_.back().wait_for (std::chrono::seconds (0)) ==
```
vvenkates27, line 79: Why nixl::time is not used here?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+                if (current_status != NIXL_SUCCESS) {
+                    status_futures_.clear();
+                    return current_status;
+                }
+                status_futures_.pop_back();
+            } else {
+                return NIXL_IN_PROG;
+            }
+        }
+
```
vvenkates27: Unwanted space

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+// -----------------------------------------------------------------------------
+
+nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params)
+    : nixlBackendEngine (init_params),
+      executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
+      s3_client_ (std::make_shared<AwsS3Client> (init_params->customParams, executor_)) {
+    NIXL_INFO << "Object storage backend initialized with S3 client wrapper";
+}
+
+nixlObjEngine::nixlObjEngine (const nixlBackendInitParams *init_params,
```
vvenkates27, line 124: What is the difference between injected and the previous one?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+                              std::shared_ptr<IS3Client> s3_client)
+    : nixlBackendEngine (init_params),
+      executor_ (std::make_shared<AsioThreadPoolExecutor> (std::thread::hardware_concurrency())),
+      s3_client_ (s3_client) {
+    s3_client_->setExecutor (executor_);
+
```
vvenkates27: Seems no space in previous function, but have space here?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlObjEngine::deregisterMem (nixlBackendMD *meta) {
+    nixlObjMetadata *obj_md = static_cast<nixlObjMetadata *> (meta);
+    if (obj_md) {
+        std::unique_ptr<nixlObjMetadata> obj_md_ptr = std::unique_ptr<nixlObjMetadata> (obj_md);
+        dev_id_to_obj_key_.erase (obj_md->dev_id);
```
vvenkates27, line 160: Why do we have space after erase and before paranthesis?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+nixl_status_t
+nixlObjEngine::registerMem (const nixlBlobDesc &mem,
+                            const nixl_mem_t &nixl_mem,
+                            nixlBackendMD *&out) {
+    auto supported_mems = getSupportedMems();
+    if (std::find (supported_mems.begin(), supported_mems.end(), nixl_mem) == supported_mems.end())
+        return NIXL_ERR_NOT_SUPPORTED;
+
+    if (nixl_mem == OBJ_SEG) {
+        std::unique_ptr<nixlObjMetadata> obj_md = std::make_unique<nixlObjMetadata> (
```
vvenkates27, line 146: Cool way to implement large key mapping!

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+nixl_status_t
+nixlObjEngine::prepXfer (const nixl_xfer_op_t &operation,
+                         const nixl_meta_dlist_t &local,
+                         const nixl_meta_dlist_t &remote,
+                         const std::string &remote_agent,
+                         nixlBackendReqH *&handle,
+                         const nixl_opt_b_args_t *opt_args) const {
+    if (!isValidPrepXferParams (operation, local, remote, remote_agent, localAgent))
```
vvenkates27, line 173: Here we expect the user does't use OBJ_SEG as local dlist?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+                         const nixl_opt_b_args_t *opt_args) const {
+    nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *> (handle);
+
+    for (int i = 0; i < local.descCount(); ++i) {
+        const auto &local_desc = local[i];
+        const auto &remote_desc = remote[i];
+
+        auto obj_key_search = dev_id_to_obj_key_.find (remote_desc.devId);
+        if (obj_key_search == dev_id_to_obj_key_.end()) {
+            NIXL_ERROR << "No object key found for device ID: " << remote_desc.devId;
```
vvenkates27: Add a meaningful error message here. - i.e User failed to register key

src/plugins/obj/obj_executor.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef OBJ_EXECUTOR_H
+#define OBJ_EXECUTOR_H
+
+#include <aws/core/utils/threading/Executor.h>
+#include <asio.hpp>
+#include <functional>
+
+class AsioThreadPoolExecutor : public Aws::Utils::Threading::Executor {
+public:
+    explicit AsioThreadPoolExecutor (std::size_t num_threads) : pool_ (num_threads) {}
```
vvenkates27, line 27: How are we determining num_threads, we should control this through backend parameters.

src/plugins/obj/obj_s3_client.cpp
```diff
@@ -0,0 +1,222 @@
...
+    if (session_token.empty()) return Aws::Auth::AWSCredentials (access_key, secret_key);
+
+    return Aws::Auth::AWSCredentials (access_key, secret_key, session_token);
+}
+
+bool
+getUserVirtualAddressing (nixl_b_params_t *custom_params) {
+    if (!custom_params) return false;
+
+    auto virtual_addressing_it = custom_params->find ("use_virtual_addressing");
```
vvenkates27, line 90: What is the purpose of virtual addressing here?

src/plugins/obj/obj_backend.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        auto status_promise = std::make_shared<std::promise<nixl_status_t>>();
+        req_h->status_futures_.push_back (status_promise->get_future());
+
+        uintptr_t data_ptr = local_desc.addr;
+        size_t data_len = local_desc.len;
+        size_t offset = remote_desc.addr;
+
+        if (operation == NIXL_WRITE)
+            s3_client_->PutObjectAsync (obj_key_search->second,
```
vvenkates27, line 211: Usage of *ObjectAsync is confusing, what is called here? and what is implemented in obj_s3_client.cpp?
Is the future here the callback, how are we using threads to limit this?

src/plugins/obj/obj_backend.cpp
w1ldptr: Because std::future<T>::wait_for() accepts std::chrono::duration, not nixl::time.

src/plugins/obj/obj_backend.cpp
w1ldptr: I used Posix plugin as a reference for sanity checks and don't have a strong opinion about this one. Up to you whether to remote it to leave as-is.

src/plugins/obj/obj_backend.cpp
w1ldptr: removed

src/plugins/obj/obj_backend.cpp
w1ldptr: This ctor implements the dependency injection pattern to allow unit tests to initialize nixlObjEngine with mocked S3 client. Previous one constructs AWS SDK wrapper client for normal operation.

src/plugins/obj/obj_backend.cpp
w1ldptr: removed

src/plugins/obj/obj_backend.cpp
w1ldptr: This is what we have in the current clang-format config. I [wasn't a big fan of the change](https://github.com/ai-dynamo/nixl/pull/410#discussion_r2120389473) and it is being discussed again [here](https://github.com/ai-dynamo/nixl/pull/501/files), but it is not something we can change in this PR.

src/plugins/obj/obj_backend.cpp
w1ldptr: I took both remote!=local check and this assumption is from Posix plugin.

src/plugins/obj/obj_backend.cpp
w1ldptr: done

src/plugins/obj/obj_s3_client.cpp
w1ldptr: Disabling it simplifies container DNS configuration for testing, not relevant for production.

src/plugins/obj/obj_executor.h
w1ldptr: Done.

src/plugins/obj/obj_backend.cpp
w1ldptr: Not sure I follow, so let me try to address the questions one-by-one to the best of my ability:

> Usage of *ObjectAsync is confusing, what is called here?

IS3Client interface is intended to be a thin wrapper over AWS SDK S3 APIs required by the plugin that can easily be substituted with a mock via dependency injection. It tries to adhere to existing AWS SDK in both API (for performance) and naming (for readability). As you can check in obj_s3_client.cpp naming and completion-via-callback implementation of async API is dictated by AWS SDK.

> and what is implemented in obj_s3_client.cpp?

AwsS3Client is a concrete implementation of IS3Client interface that wraps AWS SDK.

> Is the future here the callback

Not sure how a callback can be a future. The callback is a completion callback that is called by AWS SDK on completion of async API operation. This is how AWS SDK implements async. To communicate the completion status from that callback to the nixlBackendReqH instance it relies on future/promise pairs. The callback sets async op status in the promise which then can be read from its corresponding future in checkXfer() -> getOverallStatus().

>  how are we using threads to limit this?

nixlObjEngine ctor initializes executor (a thin wrapper over Asio thread pool that implements interface compatible with AWS SDK) with specified number of threads, then passes it to AwsS3Client instance which instantiates Aws::S3::S3Client with it. The client then uses the executor to schedule all its async operation on the thread pool nixlObjEngine provided.

src/plugins/obj/obj_backend.cpp
barneuman: I'm not sure how it's different from checking the descriptor list in the prepXfer. The user can also pass another value here, and it will not be rechecked.
There is no issue with checking it here.

src/plugins/obj/obj_backend.cpp
barneuman: I meant you can check the parameters before you start calling the put/get functions (not saying that you should, but it can be done).
It's not completely aligned in other backends (in some cases, we cancel the request).

test/gtest/unit/obj/obj.cpp
barneuman: Okay, so I would say we should check writing to offset, as it's only an AWS SDK limitation.
Additionally, we didn't implement the PUT method; however, we did implement the extra copy (which was the first AWS SDK implementation) in the GET method. What's the logic?

test/gtest/unit/obj/obj.cpp
```diff
@@ -0,0 +1,499 @@
...
+        std::vector<char> test_buffer0 (1024);
+        std::vector<char> test_buffer1 (1024);
+        if (operation == NIXL_WRITE) {
+            for (size_t i = 0; i < test_buffer0.size(); ++i) {
+                test_buffer0[i] = static_cast<char> ('X' + (i % 3));
+            }
+            for (size_t i = 0; i < test_buffer1.size(); ++i) {
+                test_buffer1[i] = static_cast<char> ('x' + (i % 3));
+            }
+        }
```
barneuman: not needed

subprojects/asio
barneuman: Looking at Abseil integration [PR.](https://github.com/ai-dynamo/nixl/pull/137/) There is only a wrap file.

src/plugins/obj/obj_backend.cpp
w1ldptr: Okay, moving the check to isValidPrepXferParams.

src/plugins/obj/obj_backend.cpp
w1ldptr: We can't eliminate the possibility of some operations completing while others fail with concurrent async implementation. Even if we move all the checks in this function to a dedicated loop and execute it before calling S3 client the errors in AWS SDK could still happen which would result some of the descriptors being processed concurrently with such failures.

test/gtest/unit/obj/obj.cpp
w1ldptr: Oh, I guess I now understand the confusion. The unit test doesn't check how closely the mocked class mimics behaviour of the original, that would make no sense and provide no value. Instead, it validates that the API of unit under test (nixlBackendEngine) calls lower layer API (IS3Client) on specific input as we expect it to.

subprojects/asio
w1ldptr: Yeah, sorry, you are correct.

test/gtest/unit/obj/obj.cpp
w1ldptr: removed

barneuman (dismissed):

src/plugins/obj/obj_backend.cpp
vvenkates27: We can print out a Error message but we don't have to fail as long as a valid local agent name is provided.

src/plugins/obj/obj_backend.cpp
vvenkates27: Please add a comment to say this is for testing.

src/plugins/obj/obj_backend.cpp
vvenkates27: Its fine to consider OBJ_SEG always as a remote segment - as long as we are able message that with a proper log message to facilitate debugging.

src/plugins/obj/obj_backend.cpp
vvenkates27: Ok I get it now, the call back sets the promise which is used in the status check. This is how the AWS async functions are tied to the internal ASIO thread pool. Its good we have a mock interface, we can add a comment to that end so it gives idea of future plugins to benefits from that.

test/gtest/unit/obj/meson.build
```diff
@@ -0,0 +1,29 @@
...
+obj_unit_test_dep = declare_dependency(
+    sources: [
+        'obj.cpp',
+    ],
+    include_directories: [
+        nixl_inc_dirs,
+        '../../../../src/plugins/obj',
+    ],
+    dependencies: [
+        aws_s3.partial_dependency(compile_args: false, includes: true, link_args: true, links: true),
```
vvenkates27, line 25: Can you tell me the purpose of partial dependency?

src/plugins/obj/obj_s3_client.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+void
+AwsS3Client::PutObjectAsync (std::string_view key,
+                             uintptr_t data_ptr,
+                             size_t data_len,
+                             size_t offset,
+                             PutObjectCallback callback) {
+    // AWS S3 doesn't support partial put operations with offset
+    if (offset != 0) {
```
vvenkates27, line 167: This check is only for Put and not Get, does that mean AWS supports partial read with get?

w1ldptr: > does that mean AWS supports partial read with get?

Yes, it does.

test/gtest/unit/obj/meson.build
w1ldptr: I used it to disable inheriting the C++ compiler flags that AWS SDK uses (compile_args: false). Otherwise NIXL fails to compile because AWS SDK sets C++11 language version.

src/plugins/obj/obj_backend.cpp
w1ldptr: Printing an error without returning one is a bit counterintuitive IMO. Perhaps a warning?

src/plugins/obj/obj_backend.cpp
w1ldptr: We already do:

```
    if (local.getType() != DRAM_SEG) {
        NIXL_ERROR << absl::StrFormat ("Error: Local memory type must be DRAM_SEG, got %d",
                                       local.getType());
        return false;
    }

    if (remote.getType() != OBJ_SEG) {
        NIXL_ERROR << absl::StrFormat ("Error: Remote memory type must be OBJ_SEG, got %d",
                                       remote.getType());
        return false;
    }
```

You suggest to improve these somehow?

src/plugins/obj/obj_backend.cpp
w1ldptr: done

src/plugins/obj/obj_backend.cpp
w1ldptr: The S3 interface doesn't have any assumptions about what the caller does in those callbacks though. Obj backend is the better place for it IMO.

test/gtest/unit/obj/meson.build
vvenkates27: I see good to know. thanks!

src/plugins/obj/obj_backend.cpp
vvenkates27: Ok I missed this. Then its good enough

Issue #334, opened 2025-05-19
NIXL_ERR_BACKEND Seg fault when using device tensors - check if we can detect lack of GPU mem support during UCX plugin init using ucp_context_query


YLGH:
I'm running into

```
  File "/home/yingliu/.local/lib/python3.12/site-packages/nixl/_api.py", line 267, in register_memory
    self.agent.registerMem(reg_descs, handle_list)
nixl._bindings.nixlBackendError: NIXL_ERR_BACKEND
Segmentation fault (core dumped)
```

When running the following
```

import torch

from nixl._api import nixl_agent, nixl_agent_config
import logging

device = torch.device("cuda", 0)
BLOCK_SIZE = 64
agent = nixl_agent("test_agent")
k_cache = torch.randn(BLOCK_SIZE, 4, 128, device=device, dtype=torch.float16)
cache_descs = agent.register_memory(k_cache)
```

It seems to infer the correct memory type of 

`Memory type cuda nixl_mem_t.VRAM_SEG` 

I run into a similar issue when running the example: https://github.com/ai-dynamo/nixl/blob/4f37f074b93d6874ee874ad2f8e317403e84420d/examples/python/nixl_api_example.py#L32 when I change DRAM to VRAM (and corresponding tensors to be on device)

How can I debug this? 

I'm running on version 0.2.0 from` pip install nixl`

Thanks!

mkhazraee:
This is indicating that UCX cannot register GPU memory. Should be coming from either driver access or another UCX being available on the system and causing conflict. Can you detail more about the setup and if you have tested any other gpu-to-gpu tests, ruling out the drive potential.

ovidiusm:
This was fixed in the main branch and will be published in the wheel in the upcoming 0.3.1 release. The issue was due to missing UCX module libraries from the package.

PR #482, opened 2025-06-18
Add missing UCX plugins to python wheel

ovidiusm:
## What?

Fixes to wheel packaging:

* UCX transport modules are now packaged correctly (previously missing)
* libcuda is no longer packaged (should be installed on the system, not part of the wheel)

## Why?
Fixes NIX-100 and issue #334 

## How?

The wheel is a zip archive that contains the nixl code and some dependencies that are also packaged.

The previous wheel bundled by mistake the libcuda.so file from the system used to build the wheel, and did not bundle the UCX transport modules.

The transport modules were not bundled because the bundling step is done by the `auditwheel repair` command, which pulls into the wheel all the libraries linked against from the base system. However, the UCX transport modules are not linked directly, instead loaded at runtime by UCX through dlopen.

To add the transport modules, we copy them into the wheel in a ucx subdirectory relative to the libds directory where the main UCX libs are located (this is important, because that is how UCX is loading the modules) and we patch the binaries to point to the packaged UCX libs instead of the system UCX libs. We use the same lib renaming scheme and rpath patching [as done by auditwheel repair](https://github.com/pypa/auditwheel/blob/112180fbcce8112e8c262d2704c9f625156d59cc/src/auditwheel/repair.py#L35)

## Tests

Tested the python 3.12 wheel using a docker environment based on the nvidia cuda 12.8 base image. Test script included in the PR.

### DIFF
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -160,6 +160,8 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
         uv build --wheel --out-dir /tmp/dist --python $PYTHON_VERSION; \
     done
 RUN uv pip install auditwheel && \
-    uv run auditwheel repair /tmp/dist/nixl-*cp3*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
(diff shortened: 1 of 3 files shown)


ovidiusm:
Please review https://github.com/ai-dynamo/nixl/pull/483 (based on main) instead

PR #483, opened 2025-06-18
Add missing UCX plugins to python wheel

ovidiusm:
## What?

Fixes to wheel packaging:

* UCX transport modules are now packaged correctly (previously missing)
* libcuda is no longer packaged (should be installed on the system, not part of the wheel)

## Why?
Fixes NIX-100 and issue #334 

## How?

The wheel is a zip archive that contains the nixl code and some dependencies that are also packaged.

The previous wheel bundled by mistake the libcuda.so file from the system used to build the wheel, and did not bundle the UCX transport modules.

The transport modules were not bundled because the bundling step is done by the `auditwheel repair` command, which pulls into the wheel all the libraries linked against from the base system. However, the UCX transport modules are not linked directly, instead loaded at runtime by UCX through dlopen.

To add the transport modules, we copy them into the wheel in a ucx subdirectory relative to the libds directory where the main UCX libs are located (this is important, because that is how UCX is loading the modules) and we patch the binaries to point to the packaged UCX libs instead of the system UCX libs. We use the same lib renaming scheme and rpath patching [as done by auditwheel repair](https://github.com/pypa/auditwheel/blob/112180fbcce8112e8c262d2704c9f625156d59cc/src/auditwheel/repair.py#L35)

## Tests

Tested the python 3.12 wheel using a docker environment based on the nvidia cuda 12.8 base image. Test script included in the PR.

### DIFF
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -161,7 +161,9 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
         uv build --wheel --out-dir /tmp/dist --python $PYTHON_VERSION; \
     done
 RUN uv pip install auditwheel && \
-    uv run auditwheel repair /tmp/dist/nixl-*cp3*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
+    uv run auditwheel repair /tmp/dist/nixl-*cp3*.whl --exclude 'libcuda*' --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
+
+RUN /workspace/nixl/contrib/wheel_add_ucx_plugins.py --ucx-lib-dir /usr/lib64 /workspace/nixl/dist/*.whl
 
 RUN uv pip install dist/nixl-*cp${DEFAULT_PYTHON_VERSION//./}*.whl
 
--- a/contrib/wheel-test.sh
+++ b/contrib/wheel-test.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+if [ -n "$TEST_RUN" ]
+then
+  export DEBIAN_FRONTEND=noninteractive
+  apt-get update
+  apt-get install -y python3.12 python3.12-venv python3-pip
+  # Install infiniband libraries
+  apt-get install -y libibverbs1 rdma-core ibverbs-utils ibverbs-providers libibumad3 libnuma1 librdmacm1t64
+  python3.12 -m venv /root/venv
+  source /root/venv/bin/activate
+  pip install --upgrade pip
+  pip install /wheel/nixl*cp312*.whl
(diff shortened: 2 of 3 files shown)


ovidiusm:
This is the same patch from https://github.com/ai-dynamo/nixl/pull/482, I have just tested it on a different base branch.

contrib/Dockerfile.manylinux
```diff
@@ -161,6 +161,7 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
 
 # Exclude libcuda.so.1 due to compatibility issues, should link with cuda driver library on host
 RUN uv pip install auditwheel && \
-    uv run auditwheel repair --exclude libcuda.so.1 /tmp/dist/nixl-*cp3*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
+    uv run auditwheel repair --exclude libcuda.so.1 /tmp/dist/nixl-*cp3*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist && \
```
aranadive: Should we add this to dockerfiles to be consistent?

contrib/Dockerfile.manylinux
```diff
@@ -161,6 +161,7 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
...
+    /workspace/nixl/contrib/wheel_add_ucx_plugins.py --ucx-lib-dir /usr/lib64 /workspace/nixl/dist/*.whl
```
aranadive: Can we make the path relative based on workspace dir?

aranadive:
Can you also fix up the separate [build-wheel.sh](https://github.com/ai-dynamo/nixl/blob/main/contrib/build-wheel.sh) script? Not required for this PR though.

contrib/Dockerfile.manylinux
ovidiusm: Yes. I would do that in a separate PR, since they are not used for creating the public wheel

contrib/Dockerfile.manylinux
ovidiusm: Which path? `/workspace/nixl/contrib/wheel_add_ucx_plugins.py` ?  Or `/workspace/nixl/dist`? Or both?

ovidiusm:
By the way, there are 4 different places we build the nixl wheel in this repository. And each one is slightly different. I think on long term we might want revisit how this is done.

contrib/Dockerfile.manylinux
aranadive: I meant both but we can do it in a cleanup PR.

contrib/wheel-test.sh
```diff
@@ -0,0 +1,54 @@
...
+  apt-get update
+  apt-get install -y python3.12 python3.12-venv python3-pip
+  # Install infiniband libraries
+  apt-get install -y libibverbs1 rdma-core ibverbs-utils ibverbs-providers libibumad3 libnuma1 librdmacm1t64
+  python3.12 -m venv /root/venv
+  source /root/venv/bin/activate
+  pip install --upgrade pip
+  pip install /wheel/nixl*cp312*.whl
+  export UCX_PROTO_INFO=y
+  python /nixl/examples/python/nixl_api_example.py
```
aranadive: Can we confirm the UCX PROTO output contains the correct transport? Or atleast add a note here saying the expected test output?

contrib/Dockerfile.manylinux
yosefe: shall we also update contrib/build-wheel.sh ?

contrib/wheel-test.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
yosefe: i'd remove the test from this PR, to make it easy to cherry pick

contrib/Dockerfile.manylinux
ovidiusm: I would not do it in this PR, because that script is not part of the wheel release process and it's not clear if it is used anywhere and in which environment. I confirmed with @pvijayakrish that she is using the Docker.manylinux environment to generate the wheel.

contrib/wheel-test.sh
ovidiusm: Agreed. But I separated the script from this PR so that we can commit faster.

contrib/wheel-test.sh
ovidiusm: Done.

aranadive (approved):

PR #485, opened 2025-06-18
SRC/BINDINGS/PYTHON: Fix handling of unsorted desc list passed with sorted=True

roiedanino:
## What?
In python bindings - warn when an xferDList / regDList is created with sorted=True while it's not sorted

## Why?
We had an assertion for that, but they are not active in release builds.

## Desired Behavior
Nixl would print a warning message and treat the descriptors list as not sorted

### DIFF
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -35,7 +35,7 @@ export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:/usr/lo
 export LD_LIBRARY_PATH=/usr/local/cuda/compat/lib.real:$LD_LIBRARY_PATH
 export CPATH=${INSTALL_DIR}/include:$CPATH
 export PATH=${INSTALL_DIR}/bin:$PATH
-export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
+export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:${UCX_INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
 export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
 
 pip3 install --break-system-packages .
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -35,7 +35,7 @@ export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:/usr/lo
 export LD_LIBRARY_PATH=/usr/local/cuda/compat/lib.real:$LD_LIBRARY_PATH
 export CPATH=${INSTALL_DIR}/include:$CPATH
 export PATH=${INSTALL_DIR}/bin:$PATH
-export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:${UCX_INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
+export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
 export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
 
 pip3 install --break-system-packages .
--- a/src/bindings/python/nixl_bindings.cpp
+++ b/src/bindings/python/nixl_bindings.cpp
@@ -188,7 +188,9 @@ PYBIND11_MODULE(_bindings, m) {
                 nixl_xfer_dlist_t new_list(mem, sorted, n);
                 // We assume that the Nx3 array matches the nixlBasicDesc layout so we can simply memcpy
                 std::memcpy(&new_list[0], descs.data(), descs.size() * sizeof(uint64_t));
-                assert(!sorted || new_list.verifySorted());
+                if (sorted && !new_list.verifySorted()) {
+                    throw std::invalid_argument("descs must be sorted when 'sorted' passed as True");
+                }
                 return new_list;
             }), py::arg("type"), py::arg("descs").noconvert(), py::arg("sorted")=false)
         .def(py::init([](nixl_mem_t mem, py::list descs, bool sorted) {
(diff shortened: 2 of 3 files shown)


tstamler (commented):
let's also remove the `#include <cassert>` at the beginning, we really shouldn't have any asserts in the bindings

src/bindings/python/nixl_bindings.cpp
```diff
@@ -188,7 +188,9 @@ PYBIND11_MODULE(_bindings, m) {
                 nixl_xfer_dlist_t new_list(mem, sorted, n);
                 // We assume that the Nx3 array matches the nixlBasicDesc layout so we can simply memcpy
                 std::memcpy(&new_list[0], descs.data(), descs.size() * sizeof(uint64_t));
-                assert(!sorted || new_list.verifySorted());
+                if (sorted && !new_list.verifySorted()) {
+                    throw std::invalid_argument("descs must be sorted when 'sorted' passed as True");
```
gleon99: maybe: `...when passing sorted=True` (or "setting")?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -203,7 +205,9 @@ PYBIND11_MODULE(_bindings, m) {
                     }
                     new_list[i] = nixlBasicDesc(desc[0].cast<uintptr_t>(), desc[1].cast<size_t>(), desc[2].cast<uint64_t>());
                 }
-                assert(!sorted || new_list.verifySorted());
+                if (sorted && !new_list.verifySorted()) {
+                    throw std::invalid_argument("descs must be sorted when 'sorted' passed as True");
+                }
```
yosefe: seems the error message is the same in all cases, can we unite common code?

yosefe:
failure in gitlab is relevant:
```
NIXL_ETCD_ENDPOINTS is not set, using socket exchange
Backend UCX was instantiated
Initialized NIXL agent: target
Traceback (most recent call last):
  File "/workspace/nixl/examples/python/partial_md_example.py", line 101, in <module>
    target_reg_descs1 = target_agent.get_reg_descs(target_strs1, "DRAM", is_sorted=True)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nixl/_api.py", line 865, in get_reg_descs
    new_descs = nixlBind.nixlRegDList(
                ^^^^^^^^^^^^^^^^^^^^^^
ValueError: descs must be sorted when 'sorted' passed as True

Running after_script
00:01
Running after script...
$ echo "Cleaning up test container..."
Cleaning up test container...
$ docker rm -f ${CONTAINER_NAME} 2>/dev/null || echo "Container ${CONTAINER_NAME} was already removed or didn't exist"
nixl-test-python-no-gpu-179809971

Cleaning up project directory and file based variables
00:00
ERROR: Job failed: exit status 1
```

roiedanino:
> failure in gitlab is relevant:
> 
> ```
> NIXL_ETCD_ENDPOINTS is not set, using socket exchange
> Backend UCX was instantiated
> Initialized NIXL agent: target
> Traceback (most recent call last):
>   File "/workspace/nixl/examples/python/partial_md_example.py", line 101, in <module>
>     target_reg_descs1 = target_agent.get_reg_descs(target_strs1, "DRAM", is_sorted=True)
>                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.12/dist-packages/nixl/_api.py", line 865, in get_reg_descs
>     new_descs = nixlBind.nixlRegDList(
>                 ^^^^^^^^^^^^^^^^^^^^^^
> ValueError: descs must be sorted when 'sorted' passed as True
> 
> Running after_script
> 00:01
> Running after script...
> $ echo "Cleaning up test container..."
> Cleaning up test container...
> $ docker rm -f ${CONTAINER_NAME} 2>/dev/null || echo "Container ${CONTAINER_NAME} was already removed or didn't exist"
> nixl-test-python-no-gpu-179809971
> 
> Cleaning up project directory and file based variables
> 00:00
> ERROR: Job failed: exit status 1
> ```

It seems the test assumes malloc addresses are monotonically increasing which isn't always the case, I'll sort the lists before passing them

src/infra/nixl_descriptors.cpp
```diff
@@ -312,6 +312,9 @@ bool nixlDescList<T>::verifySorted() {
 
     for (int i=0; i<size-1; ++i) {
         if (descs[i+1] < descs[i]) {
+            if (sorted) {
+                NIXL_WARN << "Descs are not sorted although sorted=True was passed, this may affect performance";
+            }
```
yosefe, line 317: IMO we should not change the behavior of verifySorted, but create another function that warns about wrong usage of sorted=True and called from python bindings

roiedanino: But verifySorted was only made for that exact purpose for python API, its not being called from anywhere else

.gitlab/test_python.sh
```diff
@@ -35,7 +35,7 @@ export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:/usr/lo
 export LD_LIBRARY_PATH=/usr/local/cuda/compat/lib.real:$LD_LIBRARY_PATH
 export CPATH=${INSTALL_DIR}/include:$CPATH
 export PATH=${INSTALL_DIR}/bin:$PATH
-export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
+export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:${UCX_INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
```
gleon99: Unrelated?

gleon99:
> ## What?
> In python bindings - throw an exception when an xferDList / regDList is created with sorted=True while its not actually sorted
> 
> ## Why?
> We had an assertion for that, but they are not active in release builds. Without that exception, NIXL would treat the list as sorted and fail later as he would encounter mismatching descriptors.

"throw an exception"? isn't it only a warning? As discussed, we should go on running in such case, just set `sorted` to True

src/infra/nixl_descriptors.cpp
aranadive: TBH, the behavior was broken on release builds. This is making it clear that perf will be impacted without changing api.

gleon99:
Maybe rephrase the title to smth cleaner, less details. e.g
Improve/fix handling of an unsorted descs list passed with sorted=True

aranadive (approved):

PR #486, opened 2025-06-18
contrib: Remove libssl, libcrypto from wheels

aranadive:
## What?
Remove libssl, libcrypto from wheels when generated. 

## Why?
They are dependencies that can be satisfied with host libraries. Also, there are other licensing implications with older openssl libraries.

## How?
Use the exclude option when repairing the python wheel using auditwheel


aranadive:
Like to hold off on this PR. Trying to see if using openssl3 within the manylinux container is possible.

aranadive:
PR #488 supercedes this.

PR #487, opened 2025-06-19
NIXL/UCX: Added support for self notification

iyastreb:
## What?
Added support for self notification for UCX backend, which is requested by TensorRT-LLM team.

## Why?
Currently NIXL agent supports transfer Request to itself, but doesn't support genNotif to itself.
Without this option they have to use fake transfer request with notification as a workaround.

## How?
- Changed genNotif to check for local name
- Few optimisations to avoid redundant string copies
- Few optimisations to avoid heap allocation/deallocation
- Added test for UCX backend

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -999,14 +999,10 @@ nixlAgent::genNotif(const std::string &remote_agent,
 
     bool localNotif = data->name == remote_agent;
     for (auto & eng: *backend_list) {
-        if (localNotif) {
-            if (eng->supportsLocal()) {
-                return eng->genNotif(remote_agent, msg);
-            }
-        } else {
-            if (data->remoteBackends[remote_agent].count(eng->getType()) != 0) {
-                return eng->genNotif(remote_agent, msg);
-            }
+        if ((localNotif && eng->supportsLocal()) ||
+            (!localNotif &&
+             data->remoteBackends[remote_agent].count(eng->getType()) != 0)) {
+            return eng->genNotif(remote_agent, msg);
         }
     }
 
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -978,41 +978,39 @@ nixlAgent::genNotif(const std::string &remote_agent,
                     const nixl_blob_t &msg,
                     const nixl_opt_args_t* extra_params) const {
 
-    nixlBackendEngine* backend = nullptr;
-    backend_list_t*    backend_list;
+    backend_list_t backend_list_value;
+    backend_list_t* backend_list;
 
     NIXL_SHARED_LOCK_GUARD(data->lock);
-    if (!extra_params || extra_params->backends.size() == 0) {
+    if (!extra_params || extra_params->backends.empty()) {
         backend_list = &data->notifEngines;
         if (backend_list->empty())
             return NIXL_ERR_BACKEND;
     } else {
-        backend_list = new backend_list_t();
-        for (auto & elm : extra_params->backends)
+        backend_list = &backend_list_value;
+        for (auto &elm : extra_params->backends)
             if (elm->engine->supportsNotif())
                 backend_list->push_back(elm->engine);
 
         if (backend_list->empty()) {
-            delete backend_list;
             return NIXL_ERR_BACKEND;
         }
     }
 
+    bool localNotif = data->name == remote_agent;
     for (auto & eng: *backend_list) {
-        if (data->remoteBackends[remote_agent].count(
-                                 eng->getType()) != 0) {
-            backend = eng;
-            break;
+        if (localNotif) {
+            if (eng->supportsLocal()) {
+                return eng->genNotif(remote_agent, msg);
+            }
+        } else {
+            if (data->remoteBackends[remote_agent].count(eng->getType()) != 0) {
+                return eng->genNotif(remote_agent, msg);
+            }
         }
     }
 
-    if (extra_params && extra_params->backends.size() > 0)
-        delete backend_list;
-
-    if (backend)
-        return backend->genNotif(remote_agent, msg);
-    else
-        return NIXL_ERR_NOT_FOUND;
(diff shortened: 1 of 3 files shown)


src/core/nixl_agent.cpp
```diff
@@ -978,41 +978,39 @@ nixlAgent::genNotif(const std::string &remote_agent,
...
         }
     }
 
+    bool localNotif = data->name == remote_agent;
     for (auto & eng: *backend_list) {
-        if (data->remoteBackends[remote_agent].count(
-                                 eng->getType()) != 0) {
-            backend = eng;
-            break;
+        if (localNotif) {
```
tvegas1: minor, either merge calls like below or return early on error and finally `return eng->genNotif()`:
```
if ((localNotif && eng->supportsLocal()) || data...->..count() != 0) {
         return eng->genNotif();
```

test/gtest/test_transfer.cpp
```diff
@@ -464,6 +464,18 @@ TEST_P(TestTransfer, NotificationOnly) {
             getAgent(0), getAgentName(0), getAgent(1), getAgentName(1), repeat, num_threads);
 }
 
+TEST_P(TestTransfer, SelfNotification) {
+    // UCX_MO does not support local communication
+    if (getBackendName() == "UCX_MO") {
```
tvegas1, line 469: can we instead lookup the backend and test with `supportsLocal()`?

src/core/nixl_agent.cpp
iyastreb: I wouldn't merge these calls for two reasons:
1. When `localNotif` mode is used then we should not even check `remoteBackends` map, no matter if current engine `supportsLocal` or not. This OR condition is shorter, but requires extra redundant CPU cycles
2. There is a side effect of this extra check as well: `data->remoteBackends[remote_agent].count` not only lookups but actually **creates** new map entry for `remote_agent` if it does not exist yet. We don't want this when triggering local notification

Return early on error - I think this is not an option, cause we need to check all the engines, therefore we return an error at the end

test/gtest/test_transfer.cpp
iyastreb: I wanted to implement the same approach initially.
But the problem is that nixlAgent does not provide an API to lookup backend by it's name, and there is no way to extract backend from it. The only way that I see (and tried) is to:
- capture backend_handle during creation in `SetUp`
- store this handle inside another vector
- later call `getBackendHandle(0)->supportsLocal`

Why I don't like this approach is because:
- we need to include and depend on "agent_data.h" which is not referenced in gtest so far (cause we test only through API)
- keeping 1:1 mapping between agent:backend is not correct, because agent may have several backends..

So finally I decided not to complicate things and just check it by name. Since UCX_MO will be removed soon, we can just remove this check.

src/core/nixl_agent.cpp
tvegas1: ok, but 2. could be fixed, 1. cpu cycles probably not an issue here.

test/gtest/test_transfer.cpp
tvegas1: ok

src/core/nixl_agent.cpp
iyastreb: ok, right, I've updated the code
Now it have if with OR, and it does not execute remote path for local notif - I believe we don't want 2 lookups in 2 maps just in case. Hope you like it)

tvegas1 (dismissed):

test/gtest/test_transfer.cpp
```diff
@@ -464,6 +464,18 @@ TEST_P(TestTransfer, NotificationOnly) {
...
+        GTEST_SKIP() << "UCX_MO does not support local communication";
+    }
+
+    constexpr size_t repeat = 1;
+    constexpr size_t num_threads = 1;
+    doNotificationTest(
+            getAgent(0), getAgentName(0), getAgent(0), getAgentName(0), repeat, num_threads);
```
brminich: why not just
```suggestion
    constexpr size_t repeat = 1;
    constexpr size_t num_threads = 1;
    doNotificationTest(
            getAgent(0), getAgentName(0), getAgent(0), getAgentName(0), 1, 1);
```

src/core/nixl_agent.cpp
```diff
@@ -978,41 +978,35 @@ nixlAgent::genNotif(const std::string &remote_agent,
...
         }
     }
 
+    bool localNotif = data->name == remote_agent;
     for (auto & eng: *backend_list) {
-        if (data->remoteBackends[remote_agent].count(
-                                 eng->getType()) != 0) {
-            backend = eng;
-            break;
+        if ((localNotif && eng->supportsLocal()) ||
```
brminich, line 1002: is it the fix? so the issue was that some backend not supporting local notif was used?

test/gtest/test_transfer.cpp
iyastreb: In this case I just used the same template as the notification test above. So 1 reason is to keep the style.

In previous PRs there were proposal to make a constant for any literal to avoid magic numbers. Personally I'm not always following this, but in this particular case IMO it makes sense: when we pass multiple integers we clearly indicate what is where.

src/core/nixl_agent.cpp
iyastreb: The primary root cause was that in order to send notification we were checking `data->remoteBackends[remote_agent]` which does not contain local (self) agent name.

Despite that connect to itself is properly established in `createBackend`:
```
        if (backend->supportsLocal()) {
            ret = backend->connect(data->name);

nixl_status_t nixlUcxEngine::connect(const std::string &remote_agent) {
    if(remote_agent == localAgent) {
        return loadRemoteConnInfo(remote_agent, workerAddr);
    }
```
So the connection to itself was properly established, with populating `remoteConnMap` with local worker. But `remoteBackends` are not updated, which totally makes sense.
Now the logic is fixed: we do allow self notification when local backend allows it (and without even checking `remoteBackends` for local case)

PR #488, opened 2025-06-19
Upgrade ssl dependency for wheel

ovidiusm:
## What?
Use newer SSL for building the wheel, but remove the binaries from packaging

## Why?
The base image has a very old version

### DIFF
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -36,9 +36,6 @@ RUN yum groupinstall -y 'Development Tools' &&  \
     gflags \
     glibc-headers \
     gcc-c++ \
-    grpc \
-    grpc-plugins \
-    grpc-devel \
     libaio \
     libaio-devel \
     libtool-ltdl \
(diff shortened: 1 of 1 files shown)


yosefe (approved):

contrib/Dockerfile.manylinux
```diff
@@ -36,9 +36,6 @@ RUN yum groupinstall -y 'Development Tools' &&  \
     gflags \
     glibc-headers \
     gcc-c++ \
-    grpc \
-    grpc-plugins \
-    grpc-devel \
```
aranadive, line 41: Should we remove the openssl, openssl-devel, protobuf packages from here too?

ovidiusm: Not sure if that has any effect, since openssl is sometimes a dependency of the package manager, curl etc.

aranadive: In that case it would be installed in the base image anyway. Ok maybe we can cleanup in separate PR.

aranadive (approved):
I tested with building the manylinux docker image. Then deploying the wheel on an ubuntu24.04. Checked that the libssl, libcrypto dependencies for etcd were linked to host ones. Then ran partial_md_example.py with the NIXL_ETCD_ENDPOINTS set to use ETCD which passed.

PR #499, opened 2025-06-25
BACKEND/UCX: enable error handling mode peer by default

evgeny-leksikov:
## What?
enable error handling mode peer by default in UCX backend

## Why?
better error handling support

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -560,12 +560,20 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
     if (num_workers_iter == custom_params->end() || !absl::SimpleAtoi(num_workers_iter->second, &numWorkers))
         numWorkers = 1;
 
+    ucp_err_handling_mode_t err_handling_mode;
     const auto err_handling_mode_it =
-            custom_params->find("ucx_error_handling_mode");
-    ucp_err_handling_mode_t err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
-    if (err_handling_mode_it != custom_params->end() &&
-        (err_handling_mode_it->second == "peer")) {
+        custom_params->find(std::string(nixl_ucx_err_handling_param_name));
+    if (err_handling_mode_it == custom_params->end()) {
         err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+    } else {
+        try {
+            err_handling_mode = ucx_err_mode_from_string(err_handling_mode_it->second);
+        }
+        catch (const std::invalid_argument &e) {
+            NIXL_ERROR << e.what();
+            initErr = true;
+            return;
+        }
     }
 
     uc = std::make_unique<nixlUcxContext>(devs,
--- a/src/utils/ucx/ucx_utils.cpp
+++ b/src/utils/ucx/ucx_utils.cpp
@@ -14,14 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #include "ucx_utils.h"
 
-#include <exception>
-#include <vector>
-#include <string>
+#include <algorithm>
 #include <cstring>
+#include <exception>
 #include <stdexcept>
-#include <type_traits>
+#include <string>
+#include <vector>
 
 #include <nixl_types.h>
 
@@ -31,6 +32,15 @@
 #include "rkey.h"
 
 using namespace std;
+
+[[nodiscard]] nixl_b_params_t
+get_ucx_backend_common_options() {
+    nixl_b_params_t params = {{"ucx_devices", ""}, {"num_workers", "1"}};
+
+    params.emplace(nixl_ucx_err_handling_param_name,
+                   ucx_err_mode_to_string(UCP_ERR_HANDLING_MODE_PEER));
+    return params;
+}
 
 nixl_status_t ucx_status_to_nixl(ucs_status_t status)
 {
@@ -54,8 +64,46 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
     }
 }
 
-static void err_cb_wrapper(void *arg, ucp_ep_h ucp_ep, ucs_status_t status)
-{
+[[nodiscard]] std::string_view
+ucx_err_mode_to_string(ucp_err_handling_mode_t t) {
+    switch (t) {
+    case UCP_ERR_HANDLING_MODE_NONE:
+        return "none";
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument(std::to_string(t));
+    }
+}
+
+[[nodiscard]] ucp_err_handling_mode_t
+ucx_err_mode_from_string(std::string_view s) {
+    constexpr std::array<ucp_err_handling_mode_t, 2> nixl_ucx_err_handling_modes = {
+        UCP_ERR_HANDLING_MODE_NONE,
+        UCP_ERR_HANDLING_MODE_PEER,
+    };
+
+    for (const auto mode : nixl_ucx_err_handling_modes) {
+        if (ucx_err_mode_to_string(mode) == s) {
+            return mode;
+        }
+    }
+
+    std::stringstream err_msg;
+    err_msg << "Invalid error handling mode: " << s << ". Valid values are: <";
+    for (size_t i = 0; i < nixl_ucx_err_handling_modes.size(); ++i) {
+        err_msg << ucx_err_mode_to_string(nixl_ucx_err_handling_modes[i]);
+        if (i < nixl_ucx_err_handling_modes.size() - 1) {
+            err_msg << "|";
+        }
+    }
+
+    err_msg << ">";
+    throw std::invalid_argument(err_msg.str());
+}
+
+static void
+err_cb_wrapper(void *arg, ucp_ep_h ucp_ep, ucs_status_t status) {
     nixlUcxEp *ep = reinterpret_cast<nixlUcxEp*>(arg);
     ep->err_cb(ucp_ep, status);
 }
--- a/src/utils/ucx/ucx_utils.h
+++ b/src/utils/ucx/ucx_utils.h
@@ -35,6 +35,8 @@ enum class nixl_ucx_mt_t {
     WORKER
 };
 
+constexpr std::string_view nixl_ucx_err_handling_param_name = "ucx_error_handling_mode";
+
 template<typename Enum>
 [[nodiscard]] constexpr auto enumToInteger(const Enum e) noexcept
 {
@@ -209,14 +211,15 @@ private:
     ucp_err_handling_mode_t err_handling_mode_;
 };
 
-[[nodiscard]] static inline nixl_b_params_t get_ucx_backend_common_options() {
-    return {
-        { "ucx_devices", "" },
-        { "ucx_error_handling_mode", "none" }, // or "peer"
-        { "num_workers", "1" }
-    };
-}
+[[nodiscard]] nixl_b_params_t
+get_ucx_backend_common_options();
 
 nixl_status_t ucx_status_to_nixl(ucs_status_t status);
+
+[[nodiscard]] std::string_view
+ucx_err_mode_to_string(ucp_err_handling_mode_t t);
+
+[[nodiscard]] ucp_err_handling_mode_t
+ucx_err_mode_from_string(std::string_view s);
 
 #endif
--- a/test/gtest/error_handling.cpp
+++ b/test/gtest/error_handling.cpp
@@ -24,7 +24,6 @@
 namespace gtest {
 namespace nixl {
     constexpr const char* ucx_err_handling_mode_key  = "ucx_error_handling_mode";
-    constexpr const char* ucx_err_handling_mode_none = "none";
     constexpr const char* ucx_err_handling_mode_peer = "peer";
 
     static nixlBackendH* createUcxBackend(nixlAgent& agent, const std::string& backend_name)
@@ -41,9 +40,7 @@ namespace nixl {
         EXPECT_EQ(NIXL_SUCCESS, status);
 
         nixlBackendH* backend_handle = nullptr;
-        EXPECT_EQ(ucx_err_handling_mode_none,
-                  params[ucx_err_handling_mode_key]);
-        params[ucx_err_handling_mode_key] = ucx_err_handling_mode_peer;
+        EXPECT_EQ(ucx_err_handling_mode_peer, params[ucx_err_handling_mode_key]);
         status = agent.createBackend(*it, params, backend_handle);
         EXPECT_EQ(NIXL_SUCCESS, status);
         EXPECT_NE(nullptr, backend_handle);


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -562,10 +562,19 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+            err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
+        } else if (err_handling_mode_it->second == "peer") {
+            err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+        } else {
+            NIXL_ERROR << "Invalid UCX error handling mode: "
+                       << err_handling_mode_it->second
+                       << ". Valid values are 'none' or 'peer'";
+            initErr = true;
+            return;
+        }
```
rakhmets: I think it would be better to:
1. have the option name, and possible values as constants: `ucx_error_handling_mode`, `none`, `peer`. So you can use them in the error message, and in the test.
2. store option values and corresponding UCP error handling constants in a container. So you can perform find, and if there is no entity in the container, print an error. And this approach is extendable, you can just add a new entity to a container to extend supported modes without changing the logic.
</br></br>
I would even go further, and hide backend option value conversion to UCP type in `utils/ucx`. However, this doesn't look like a scope of this PR, but future enhancement.

evgeny-leksikov: @rakhmets thanks for review , it makes sense except "you can use them ... in the test" since ucx backend specific is not part of nixl API

rakhmets: I see...
It looks like we should show the backend custom parameters to the gtests in order to properly test them. Otherwise, we are creating a strong dependency on the implementation, and we will have to change the tests when the name changes. In addition, we have no way to get valid values for the custom backend parameter.

Maybe remove this code at all from the test: `EXPECT_EQ(ucx_err_handling_mode_peer, params[ucx_err_handling_mode_key]);`
Since the test should fail anyway, if someone changes the default error handling mode in the backend code.

src/utils/ucx/ucx_utils.h
```diff
@@ -35,6 +35,14 @@ enum class nixl_ucx_mt_t {
     WORKER
 };
 
+constexpr std::string_view nixl_ucx_err_handling_param_name =
+        "ucx_error_handling_mode";
```
rakhmets: Please move this constant to `plugins/ucx`. Since it is the backend parameter.

I think `get_ucx_backend_common_options` should be also moved out from this file. Since `utils/ucx` designed to abstract UCX library objects.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -560,12 +560,16 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
         numWorkers = 1;
 
-    const auto err_handling_mode_it =
-            custom_params->find("ucx_error_handling_mode");
-    ucp_err_handling_mode_t err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
-    if (err_handling_mode_it != custom_params->end() &&
-        (err_handling_mode_it->second == "peer")) {
-        err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+    auto err_handling_mode    = UCP_ERR_HANDLING_MODE_PEER;
+    auto err_handling_mode_it = custom_params->find(std::string(nixl_ucx_err_handling_param_name));
```
gleon99: `const it`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -560,12 +560,16 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
+    if (err_handling_mode_it != custom_params->end()) {
```
gleon99: Will init inside `if()` look better or clumsy?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -19,7 +19,7 @@
 #include <string>
 #include <cstring>
 #include <stdexcept>
-#include <type_traits>
+#include <algorithm>
```
gleon99: order

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +67,43 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
     }
 }
 
+[[nodiscard]] std::string to_string(ucp_err_handling_mode_t t) noexcept
```
gleon99: too general name

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +67,43 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+
+    return "invalid";
+}
+
+[[nodiscard]] ucp_err_handling_mode_t err_mode_from_string(const std::string &s)
+{
+    static std::vector<std::string> string_values;
+    if (string_values.empty()) {
+        for (const auto &mode : nixl_ucx_err_handling_modes) {
+            string_values.push_back(to_string(mode));
```
gleon99: `emplace_back`?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +67,43 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+        }
+    }
+
+    auto it = std::find(string_values.begin(), string_values.end(), s);
```
gleon99: `const`?

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: init what?

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: aligned with https://github.com/ai-dynamo/nixl/blob/d385c748742efc8533c03ed89c71f0cd6145f0d1/src/utils/ucx/ucx_utils.h#L45

gleon99: IMO very bad style. Generic filename, generic func name for "to_string"ing of a **very specific type**

src/utils/ucx/ucx_utils.h
evgeny-leksikov: this is common for plugins/ucx and plugins/ucx_mo, probably we need another file or directory like plugins/ucx_common. I would keep it as is for now

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: I added the list of valid values to err message, but this is more likely for users but not gtest..
> Since the test should fail anyway, if someone changes the default error handling mode in the backend code.

I would not be so sure, ucx error handling may indicate remote failures even if peer_failure level is not requested, it's transport and its UCT implementation details

rakhmets: My concern is that we use some string constants in gtest, that should be the same as in implementation.

And I don't see any point in checking the default value in this test. I think it can be removed to avoid defining the same constants in gtests.
```
        EXPECT_EQ(ucx_err_handling_mode_peer,
                  params[ucx_err_handling_mode_key]);
```

P.S.
I think it would be better to print backend parameters only in case of failure instead.

evgeny-leksikov: > My concern is that we use some string constants in gtest, that should be the same as in implementation.

Completely agree but API of custom parameters is defined this way...

> And I don't see any point in checking the default value in this test.

This PR changes default value and I would like to test it

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +79,42 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+        case UCP_ERR_HANDLING_MODE_PEER:
+            return "peer";
+    }
+
+    return "invalid";
+}
+
+[[nodiscard]] ucp_err_handling_mode_t err_mode_from_string(const std::string &s)
+{
+    static std::vector<std::string> string_values;
```
iyastreb: This function is not thread-safe due to using global cache

IMO this function is unnecessary complex in addition to thread safety issues.
Why not just make it simply:
```
// No string allocation, just use a string_view
std::string_view to_string(ucp_err_handling_mode_t t)

// no cache, no allocations
ucp_err_handling_mode_t fromString(std::string_view s) {
       for (const auto &mode : nixl_ucx_err_handling_modes) {
            if (s == to_string(mode)) {
                 return mode;
            }
        }
    return error
}

File: src/utils/ucx/ucx_utils.h [hunk 10]
```diff
@@ -35,6 +35,14 @@ enum class nixl_ucx_mt_t {
...
+
+constexpr std::array<ucp_err_handling_mode_t, 2> nixl_ucx_err_handling_modes = {
```
iyastreb: No need for size:
constexpr ucp_err_handling_mode_t nixl_ucx_err_handling_modes[] = { ... };

IMO it's not worth it, because if your intention is to "automate" enum-string generation, then you still have to change this array.

Also no need to have it in the header file, as this seems to be an impl detail.
Can be just a variable inside of err_mode_from_string

evgeny-leksikov: c-style array does not allow to iterate like:
```
for (const auto &mode : nixl_ucx_err_handling_modes) {
}
```
but `std::array` requires its size in the parameters

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: makes sense since it's not a fast path, refactored

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +81,42 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument (std::to_string (t));
+        ;
+    }
+}
+
+[[nodiscard]] ucp_err_handling_mode_t
+err_mode_from_string (const std::string &s) {
```
iyastreb: s can be also string_view?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +81,42 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+[[nodiscard]] std::string_view
+to_string (ucp_err_handling_mode_t t) {
+    switch (t) {
+    case UCP_ERR_HANDLING_MODE_NONE:
+        return "none";
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument (std::to_string (t));
+        ;
```
iyastreb: remove

src/utils/ucx/ucx_utils.cpp
```diff
@@ -67,6 +81,42 @@ void ucx_modify_config(ucp_config_t *config, std::string_view key,
...
+    catch (const std::invalid_argument &) {
+        std::string err_msg = "Invalid error handling mode: " + s + ". Valid values are: <";
+        for (const auto &mode : nixl_ucx_err_handling_modes) {
+            err_msg += std::string (to_string (mode)) + "|";
+        }
+
+        err_msg.pop_back();
+        err_msg += ">";
+        throw std::invalid_argument (err_msg);
+    }
```
iyastreb: IMO this try-catch block is redundant.
In fact it's only handling exception from line 106, because exception from line 101 is not possible. And even if it happens - no need to catch it here
```suggestion
        for (const auto &mode : nixl_ucx_err_handling_modes) {
            if (to_string (mode) == s) {
                return mode;
            }
        }

        std::string err_msg = "Invalid error handling mode: " + s + ". Valid values are: <";
        for (const auto &mode : nixl_ucx_err_handling_modes) {
            err_msg += std::string (to_string (mode)) + "|";
        }

        err_msg.pop_back();
        err_msg += ">";
        throw std::invalid_argument (err_msg);
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -560,12 +560,20 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
-    if (err_handling_mode_it != custom_params->end() &&
-        (err_handling_mode_it->second == "peer")) {
+        custom_params->find(std::string(nixl_ucx_err_handling_param_name));
+    if (err_handling_mode_it == custom_params->end()) {
         err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+    } else {
+        try {
+            err_handling_mode = err_mode_from_string(err_handling_mode_it->second);
+        }
+        catch (const std::invalid_argument &e) {
```
rakhmets, line 572: I think it would be better to catch the exception in `create_ucx_engine` to avoid using `initErr`. The plan is to eventually abandon this approach in favor of catching exceptions when creating the engine.

evgeny-leksikov: For now it will look inconsistently.

> The plan is to eventually abandon this approach in favor of catching exceptions when creating the engine.

IMO it should be done in separate PR.

rakhmets: We have already started moving in this direction. E.g. exceptions from the creation of `nixlUcxContext`, `nixlUcxContext`,  `nixl::ucx::config` are caught in `create_ucx_engine`. This is a new code, so please catch this exception in `create_ucx_engine`.

evgeny-leksikov: I disagree, `nixlUcxEngine::nixlUcxEngine` should stay consistent, refactoring of entire class and/or its constuctor is not intention of this PR

rakhmets: This is not a refactoring. This doesn't add any extra inconsistency because we already have two ways of error handling in this class: using `initErr`, handling exceptions in `create_ucx_engine`. This is a new added code. Please put this into the suggested location.

evgeny-leksikov: as far as I see, we dont throw any exception from `nixlUcxEngine::nixlUcxEngine`

iyastreb (approved):

src/plugins/ucx/ucx_backend.cpp
rakhmets: As I mentioned previously, we throw exceptions from the creation of `nixlUcxContext` and `nixlUcxWorker`.

evgeny-leksikov: > As I mentioned previously, we throw exceptions from the creation of `nixlUcxContext` and `nixlUcxWorker`.

then why `nixlUcxEngine` was not refactored in the same change? Sorry but IMO you are requesting unrelated change which should be implemented separately.

rakhmets: It was, but it was done in stages, and it is continuing.
The changes are related. A new `catch` block has been added by this PR, but it should be located in a different specific location. Please move it.

evgeny-leksikov: It should be moved in separate PR, the new `catch` block was added for consistency.

rakhmets: I see your point. My point is to be consistent with the changes made by this PR: https://github.com/ai-dynamo/nixl/pull/294.

evgeny-leksikov: let's continue in https://github.com/ai-dynamo/nixl/pull/537

src/utils/ucx/ucx_utils.cpp
```diff
@@ -52,8 +66,39 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
...
+        return "none";
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument(std::to_string(t));
+    }
+}
+
+[[nodiscard]] ucp_err_handling_mode_t
+err_mode_from_string(const std::string_view &s) {
```
rakhmets:
```suggestion
err_mode_from_string(std::string_view s) {
```
`string_view` is non-owning object, passing it as a constant reference is redundant.

src/utils/ucx/ucx_utils.h
```diff
@@ -201,14 +203,15 @@ class nixlUcxWorker {
...
-        { "num_workers", "1" }
-    };
-}
+[[nodiscard]] nixl_b_params_t
+get_ucx_backend_common_options();
 
 nixl_status_t ucx_status_to_nixl(ucs_status_t status);
 
+[[nodiscard]] std::string_view
+to_string(ucp_err_handling_mode_t t);
```
rakhmets: It looks like we don't need this in a header file.

rakhmets (approved):

src/utils/ucx/ucx_utils.h
evgeny-leksikov: defined in pair with `err_mode_from_string` in the same place to avoid potential code duplication if might be needed in other place

src/utils/ucx/ucx_utils.cpp
evgeny-leksikov: fixed

src/utils/ucx/ucx_utils.h
rakhmets: It doesn't look like it will be needed.

The client code needs to convert the string to an option.
If we consider this as a class in the future, the only interface that is needed is to get an option. Smth like this:
```C++
namespace nixl::ucx {
class errHandlingMode
{
public:
    explicit errHandlingMode(std::string_view) {
        // convert string to option
    }

    [[nodiscard]] ucp_err_handling_mode_t
    get () {
        return err_handling_mode_;
    }
private:
    ucp_err_handling_mode_t err_handling_mode_{UCP_ERR_HANDLING_MODE_PEER};
};
}

// in src/utils/ucx/worker.cpp
    worker::worker(..., std::string_view err_handling_mode_str, ...) {
        ...
        errHandlingMode err_handling_mode{err_handling_mode_str};
        err_handling_mode_ = err_handling_mode.get();
    }
```

rakhmets: After posting the comment I realize that class may be redundant here. Anyway the only method that is required is conversion from string to `ucp_err_handling_mode_t`...

evgeny-leksikov: conversion `ucp_err_handling_mode_t` to string is used in `get_ucx_backend_common_options`, yep class looks redundant here

rakhmets (dismissed):

evgeny-leksikov:
@gleon99 could you pls review?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -31,6 +32,19 @@
 
 using namespace std;
 
+constexpr std::array<ucp_err_handling_mode_t, 2> nixl_ucx_err_handling_modes = {
```
yosefe: maybe move it inside err_mode_from_string() ?

src/utils/ucx/ucx_utils.cpp
```diff
@@ -52,8 +66,39 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
     }
 }
 
-static void err_cb_wrapper(void *arg, ucp_ep_h ucp_ep, ucs_status_t status)
-{
+[[nodiscard]] std::string_view
+to_string(ucp_err_handling_mode_t t) {
```
yosefe: ucx_err_mode_to_string

src/utils/ucx/ucx_utils.cpp
```diff
@@ -52,8 +66,39 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
...
+        return "none";
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument(std::to_string(t));
+    }
+}
+
+[[nodiscard]] ucp_err_handling_mode_t
+err_mode_from_string(std::string_view s) {
```
yosefe: ucx_err_mode_from_string

src/utils/ucx/ucx_utils.cpp
```diff
@@ -52,8 +66,39 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
...
+    for (const auto &mode : nixl_ucx_err_handling_modes) {
+        if (to_string(mode) == s) {
+            return mode;
+        }
+    }
+
+    std::string err_msg =
```
yosefe: maybe use std::sstream

src/utils/ucx/ucx_utils.cpp
```diff
@@ -52,8 +66,39 @@ nixl_status_t ucx_status_to_nixl(ucs_status_t status)
...
+    case UCP_ERR_HANDLING_MODE_PEER:
+        return "peer";
+    default:
+        throw std::invalid_argument(std::to_string(t));
+    }
+}
+
+[[nodiscard]] ucp_err_handling_mode_t
+err_mode_from_string(std::string_view s) {
+    for (const auto &mode : nixl_ucx_err_handling_modes) {
```
yosefe: & is redundant here

gleon99 (approved):

evgeny-leksikov:
squashed and rebased

PR #501, opened 2025-06-25
Improve .clang-format

gleon99:
Fix code style

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -12,7 +12,7 @@ AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: Empty
 AllowShortIfStatementsOnASingleLine: WithoutElse
 AllowShortLoopsOnASingleLine: false
-AlwaysBreakAfterReturnType: All
+AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: false
 BinPackArguments: false
@@ -59,7 +59,7 @@ SpaceAfterTemplateKeyword: false
 SpaceBeforeAssignmentOperators: true
 SpaceBeforeCtorInitializerColon: true
 SpaceBeforeInheritanceColon: true
-SpaceBeforeParens: NonEmptyParentheses
+SpaceBeforeParens: ControlStatementsExceptForEachMacros
 SpaceBeforeRangeBasedForLoopColon: true
 SpaceInEmptyParentheses: false
 SpacesBeforeTrailingComments: 1


brminich (dismissed):

yosefe (approved):

yosefe:
@aranadive @w1ldptr can you pls take a look?

aranadive (dismissed):

.clang-format
```diff
@@ -12,7 +12,7 @@ AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: Empty
 AllowShortIfStatementsOnASingleLine: WithoutElse
 AllowShortLoopsOnASingleLine: false
-AlwaysBreakAfterReturnType: All
+AlwaysBreakAfterReturnType: None
```
w1ldptr: Why exactly is this a "fix"? This looks more like a change to me and PR description lacks any kind of motivation for it.

yosefe: According to the [survey](https://nvidia-my.sharepoint.com/:p:/r/personal/etoledano_nvidia_com/_layouts/15/Doc.aspx?sourcedoc=%7BDFEABD11-1565-420F-ABCD-5515D841762D%7D&file=Clang%20and%20CppCoreGuidelines.pptx&wdLOR=c1051F716-9B1D-304C-80E8-676C90172493&fromShare=true&action=edit&mobileredirect=true) it should be None
Do you prefer "All"?

gleon99: This is a fix, because these settings are inconsistent with the actual code style in the repo.
The motivation is, therefore, to make `.clang-format` match the actual style in use.

Specifically, breaking after return type and having spaces after parentheses:
```
void
foo (int arg...
```
is a somewhat rare style.

w1ldptr: > According to the [survey](https://nvidia-my.sharepoint.com/:p:/r/personal/etoledano_nvidia_com/_layouts/15/Doc.aspx?sourcedoc=%7BDFEABD11-1565-420F-ABCD-5515D841762D%7D&file=Clang%20and%20CppCoreGuidelines.pptx&wdLOR=c1051F716-9B1D-304C-80E8-676C90172493&fromShare=true&action=edit&mobileredirect=true) it should be None Do you prefer "All"?

If I remember correctly some stakeholders disagreed with the survey results and it was eventually abandoned. There are multiple other settings from the survey that differ in clang-config that got eventually accepted (like AlignConsecutiveDeclaration, for example) and the survey wasn't considered at all during that PR review. We still have ColumnLimit==100, so breaking long function declaration is a concern...

> is a somewhat rare style.

It seems to be used in some of the most popular C codebases like Linux kernel (not exactly "All" but applied case-by-case) or DPDK in order to break function definitions/declarations that don't fit on a single line. I wonder how the affected code of function definitions spanning multiple lines looks like with this change and the default PenaltyReturnTypeOnItsOwnLine value and whether we should tune it as well for better results.

aranadive: I think the combination of these two options is what makes it look odd. Im fine with both these options but I think we do need to change ColumnLimit to 120 as per previous discussion. @w1ldptr Maybe with increasing columnlimit you are okay with not having a breakafterreturntype?

w1ldptr: Sounds good!

yosefe: > AlignConsecutiveDeclaration, for example) and the survey wasn't considered at all during that PR review. We still have ColumnLimit==100, so breaking long function declaration is a concern...

limit=120 seems too high, would be hard to look at several source files side-by-side.
Instead of 'AlwaysBreakAfterReturnType=None' , we can use 'BreakAfterReturnType=Automatic' which will put the return type in a separate line if the function declaration is long

brminich: both, `none` and `Automatic` sound good to me. IMO, using `all` makes the code harder to read

w1ldptr: BreakAfterReturnType=Automatic is only available starting from clang-format 19. Is this okay?

Moreover, both the old [AlwaysBreakAfterReturnType=None](https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#alwaysbreakafterreturntype) and the new [BreakAfterReturnType=Automatic](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakafterreturntype) rely on PenaltyReturnTypeOnItsOwnLine for their decisions as you can see in their respective manuals which is why I suggested this PR should also include setting that to some reasonable value that looks good for NIXL code.

yosefe: clang format 19 is available on ubuntu 24 so IMO it's ok to use it it.
So - before switching to cf19, let's set `AlwaysBreakAfterReturnType=None` and update PenaltyReturnTypeOnItsOwnLine. @gleon99 @w1ldptr is that ok?

yosefe: Suggest to add:
```diff
--- a/.clang-format
+++ b/.clang-format
@@ -13,6 +13,7 @@ AllowShortFunctionsOnASingleLine: Empty
 AllowShortIfStatementsOnASingleLine: WithoutElse
 AllowShortLoopsOnASingleLine: false
 AlwaysBreakAfterReturnType: None
+PenaltyReturnTypeOnItsOwnLine: 1
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: false
 BinPackArguments: false
```

w1ldptr: What exactly is involved in "switching" to clang-format 19 that we need to do that in multiple steps? If that is the version that will be used by NIXL CI, then let's go with it instead of reformatting the code many times. Especially considering that 'BreakAfterReturnType' is marked as deprecated now.

yosefe: I guess that to run clang-format 19 need to update https://github.com/ai-dynamo/nixl/blob/main/.github/workflows/clang-format.yml to run "clang-format-19" command. Maybe also install the relevant package on the container. @dpressle ?

dpressle: Yes, not sure its installed so might need to install it first

yosefe: Penalty=1 (as in the PR):
image
Penalty=100:
image

yosefe (dismissed):

yosefe (approved):

w1ldptr (approved):

PR #506, opened 2025-06-26
UTILS/UCX: Added ucx::Config class.

rakhmets:
## What?
Added ucx::Config class.

Applied RAII style.
Fixed memory leak in case of the `ucp_init` failure.
Added error handling of `ucp_config_read`.

## Why?
Better resource management and error handling.

## How?
I've introduced new namespaces. And I'd also like to change `class nixlUcxContext`, `class nixlUcxWorker`, etc. to `class Context`, `class Worker`, etc. in `namespace nixl::ucx`

### DIFF
--- a/src/utils/ucx/config.cpp
+++ b/src/utils/ucx/config.cpp
@@ -1,0 +1,62 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "config.h"
+
+#include <stdexcept>
+
+#include <absl/strings/str_format.h>
+
+#include "common/nixl_log.h"
+
+namespace nixl {
+namespace ucx {
+    Config::Config() : config_ (readUcpConfig(), &ucp_config_release) {}
+
+    void
+    Config::modify (std::string_view key, std::string_view value, bool force) {
+        const char *env_val;
+        if (force) {
+            env_val = nullptr;
+        } else {
+            env_val = std::getenv (absl::StrFormat ("UCX_%s", key.data()).c_str());
+        }
+
+        value = env_val ? env_val : value;
+        const auto status = ucp_config_modify (config_.get(), key.data(), value.data());
+        if (status != UCS_OK) {
+            NIXL_WARN << "Failed to modify UCX config: " << key << "=" << value << ": "
+                      << ucs_status_string (status);
+        } else {
+            NIXL_DEBUG << "Applied UCX config from " << (env_val ? "env var" : "NIXL") << ": "
+                       << key << "=" << value;
+        }
+    }
+
+    ucp_config_t *
+    Config::readUcpConfig() {
+        ucp_config_t *config = nullptr;
+        const auto status = ucp_config_read (NULL, NULL, &config);
+        if (status != UCS_OK) {
+            const auto err_str =
+                std::string ("Failed to create UCX config: ") + ucs_status_string (status);
+            NIXL_ERROR << err_str;
+            throw std::runtime_error (err_str);
+        }
+        return config;
+    }
+} // namespace ucx
+} // namespace nixl
--- a/src/utils/ucx/config.cpp
+++ b/src/utils/ucx/config.cpp
@@ -24,7 +24,7 @@
 
 namespace nixl::ucx {
 void
-Config::modify (std::string_view key, std::string_view value) const {
+config::modify (std::string_view key, std::string_view value) const {
     const char *env_val = std::getenv (absl::StrFormat ("UCX_%s", key.data()).c_str());
     if (env_val) {
         NIXL_DEBUG << "UCX env var has already been set: " << key << "=" << env_val;
@@ -34,12 +34,12 @@ Config::modify (std::string_view key, std::string_view value) const {
 }
 
 void
-Config::modifyAlways (std::string_view key, std::string_view value) const {
+config::modifyAlways (std::string_view key, std::string_view value) const {
     modifyUcpConfig (key, value);
 }
 
 ucp_config_t *
-Config::readUcpConfig() {
+config::readUcpConfig() {
     ucp_config_t *config = nullptr;
     const auto status = ucp_config_read (NULL, NULL, &config);
     if (status != UCS_OK) {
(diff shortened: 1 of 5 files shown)


src/utils/ucx/config.h
```diff
@@ -0,0 +1,58 @@
...
+    class Config {
+    public:
+        Config();
+
+        Config (const Config &) = delete;
+        Config &
+        operator= (const Config &) = delete;
+        Config (Config &&) = delete;
+        Config &
+        operator= (Config &&) = delete;
```
iyastreb: this can be moved to private

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        value = env_val ? env_val : value;
+        const auto status = ucp_config_modify (config_.get(), key.data(), value.data());
+        if (status != UCS_OK) {
+            NIXL_WARN << "Failed to modify UCX config: " << key << "=" << value << ": "
+                      << ucs_status_string (status);
+        } else {
+            NIXL_DEBUG << "Applied UCX config from " << (env_val ? "env var" : "NIXL") << ": "
+                       << key << "=" << value;
```
brminich: why do we need to apply env var using config API? I suggest to do that only if no env val is present or force flag is provided

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    ucp_config_t *
+    Config::readUcpConfig() {
+        ucp_config_t *config = nullptr;
+        const auto status = ucp_config_read (NULL, NULL, &config);
+        if (status != UCS_OK) {
+            const auto err_str =
+                std::string ("Failed to create UCX config: ") + ucs_status_string (status);
+            NIXL_ERROR << err_str;
+            throw std::runtime_error (err_str);
```
brminich: do we want to catch it?

src/utils/ucx/config.cpp
rakhmets: I think the code already does what you suggested.

Please note that this code was added here https://github.com/ai-dynamo/nixl/pull/459.

src/utils/ucx/config.cpp
rakhmets: It's caught in `create_ucx_engine`.

src/utils/ucx/config.h
rakhmets: Removed since it handled by the class member.

src/utils/ucx/config.cpp
brminich: it was overlooked there, so i suggest to remove it here, since you are modifying this code anyway

rakhmets: Fixed.

src/utils/ucx/config.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <memory>
+#include <string_view>
+
+extern "C" {
+#include <ucp/api/ucp.h>
+}
+
+namespace nixl {
+namespace ucx {
```
brminich: maybe 
```suggestion
namespace nixl::ucx {
```

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const char *env_val = std::getenv (absl::StrFormat ("UCX_%s", key.data()).c_str());
+    if (!env_val) return false;
+
+    NIXL_DEBUG << "UCX env var has already been set: " << key << "=" << env_val;
+    return true;
+}
+} // namespace
+
+namespace nixl {
+namespace ucx {
```
brminich: maybe
```suggestion
namespace nixl::ucx {
```

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <stdexcept>
+
+#include <absl/strings/str_format.h>
+
+#include "common/nixl_log.h"
+
+namespace {
+[[nodiscard]] bool
+isUcxEnvVarSet (std::string_view key) {
```
evgeny-leksikov: can be private member of `class Config`

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    NIXL_DEBUG << "UCX env var has already been set: " << key << "=" << env_val;
+    return true;
+}
+} // namespace
+
+namespace nixl {
+namespace ucx {
+    void
+    Config::modify (std::string_view key, std::string_view value, bool force) {
```
evgeny-leksikov: maybe add a separated member `Config::force_modify` without env checking instead of the boolean parameter? It could improve readability on usage.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -393,44 +376,42 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
     if (devs.size()) {
         /* TODO: check if this is the best way */
-        string dev_str = "";
+        string dev_str;
         unsigned int i;
-        for(i=0; i < devs.size() - 1; i++) {
+        for (i = 0; i < devs.size() - 1; i++) {
             dev_str = dev_str + devs[i] + ":1,";
         }
         dev_str = dev_str + devs[i] + ":1";
```
evgeny-leksikov: maybe simething like:
```suggestion
        std::string dev_list_str;
        for (const auto &dev : devs) {
            dev_list_str += dev + ":1,";
        }
        dev_list_str.pop_back(); // to remove odd comma after the last device
```

src/utils/ucx/config.cpp
rakhmets: This function doesn't need an access to the representation of the class.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class

rakhmets: Removed the method.

src/utils/ucx/config.cpp
rakhmets: Added.

src/utils/ucx/config.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (env_val) {
+        NIXL_DEBUG << "UCX env var has already been set: " << key << "=" << env_val;
+    } else {
+        modifyUcpConfig (key, value);
+    }
+}
+
+void
+config::modifyAlways (std::string_view key, std::string_view value) const {
+    modifyUcpConfig (key, value);
```
evgeny-leksikov: `modifyUcpConfig` looks redundant, `modifyAlways` can be called after `env_val` check in `modify`

src/utils/ucx/ucx_utils.cpp
```diff
@@ -393,44 +376,41 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+    nixl::ucx::config config;
 
     /* If requested, restrict the set of network devices */
     if (devs.size()) {
         /* TODO: check if this is the best way */
-        string dev_str = "";
-        unsigned int i;
-        for(i=0; i < devs.size() - 1; i++) {
-            dev_str = dev_str + devs[i] + ":1,";
+        std::string dev_str;
```
evgeny-leksikov: minor: rename to `devs_str` or `dev_list_str`

rakhmets: Fixed.

src/utils/ucx/config.cpp
rakhmets: Removed.

PR #508, opened 2025-06-27
Support read/write multiple files for storage backend in nixlbench

Pyngon:
## What?
Add multiple files support in nixlbench when using storage backend

## Why?
To understand the performance of storage backend when read/write multiple files.

## How?
1. In allocateMemory
   - compute the maximum size a file could had and register the file once.
   - compute the maximum size a thread needed for its DRAM, and allocate 1 piece of memory with the max size for each thread
2. When creating the memory xfer_list in createTransferDescLists(), divide the memory to multiple transfer unit depends on the scheme.
   - If scheme=pairwise, each file will be assigned to 1 thread only, so we need to compute number of files per threads and divide the memory to the file per thread number.
   - If scheme=tp, each file will be divided to num_threads blocks and accessed by all threads. So, each thread memory will be divided to num_files block.
3. In exchangeIOV, we need to create the file xfer_list,
   - Based on the scheme, divide the file to the same number of transfer unit as the memory xfer_list, and assigned the fd to the transfer unit based on the scheme.
4. Finally, when computing the stats, current stats does not considered multiple files, so need to add it in too.

### DIFF
--- a/benchmark/nixlbench/src/main.cpp
+++ b/benchmark/nixlbench/src/main.cpp
@@ -37,28 +37,40 @@ static std::pair<size_t, size_t> getStrideScheme(xferBenchWorker &worker, int nu
     initiator_device = xferBenchConfig::num_initiator_dev;
     target_device = xferBenchConfig::num_target_dev;
 
-    // Default value
-    count = 1;
-    buffer_size = xferBenchConfig::total_buffer_size / (initiator_device * num_threads);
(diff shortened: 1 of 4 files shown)


brminich:
@aranadive, @Pyngon is it still relevant? Can you pls resolve conflicts if it is

Pyngon:
> @aranadive, @Pyngon is it still relevant? Can you pls resolve conflicts if it is
This is not relevant anymore, another implementation with the same target is merged. Let me close this.

PR #509, opened 2025-06-27
AGENT: cleanup metadata on failure

evgeny-leksikov:
## What?

1. clean up metadata on failure
2. refactor load and invalidate remote data
3. add reconnect case to error handling suite:
 - exchange metadata
 - data transfer (positive)
 - fail target
 - data transfer (negative)
 - re-init target and re-exchange metadata
 - data transfer

## Why?
auto cleanup for reconnect


test/gtest/error_handling.cpp
```diff
@@ -258,12 +284,24 @@ bool TestErrorHandling::isFailure(size_t iter) {
...
+    case TestType::XFER_FAIL_RESTORE:
+        return iter == 1;
     }
 }
 
-template<TestErrorHandling::TestType test_type> size_t TestErrorHandling::numIter() {
-    return (test_type == TestType::XFER_THEN_FAIL) ? 2 : 1;
+template<TestErrorHandling::TestType test_type>
+size_t
+TestErrorHandling::numIter() {
```
rakhmets, line 294: Why do we need here a template? Can we just pass `test_type` as a parameter?

test/gtest/error_handling.cpp
```diff
@@ -219,8 +237,10 @@ TestErrorHandling::TestErrorHandling() : m_backend_name(GetParam())
 
 template<TestErrorHandling::TestType test_type, enum nixl_xfer_op_t op>
 void TestErrorHandling::testXfer() {
-    m_Initiator.init("initiator", m_backend_name);
-    m_Target.init("target", m_backend_name);
+    std::string initiator_name = "initiator";
+    std::string target_name = "target";
```
rakhmets: const

test/gtest/error_handling.cpp
```diff
@@ -138,11 +143,23 @@ void TestErrorHandling::Agent::init(const std::string& name, const std::string&
...
+    disconnect();
     m_priv->deregisterMem(m_mem.m_dlist, &m_mem.m_params);
     m_priv.reset();
 }
 
+void
+TestErrorHandling::Agent::disconnect() {
+    ASSERT_FALSE (m_MetaRemote.empty());
+
+    nixl_status_t status = m_priv->invalidateRemoteMD (m_MetaRemote);
```
rakhmets: const

test/gtest/error_handling.cpp
evgeny-leksikov: to calculate `numIter()` in compile time

rakhmets: I see. I would assume that the compiler will do the same, since a constant value is passed in the code calling the function.

test/gtest/error_handling.cpp
evgeny-leksikov: fixed

test/gtest/error_handling.cpp
evgeny-leksikov: fixed

test/gtest/error_handling.cpp
```diff
@@ -230,6 +250,11 @@ void TestErrorHandling::testXfer() {
 
         if (isFailure<test_type>(i)) {
             EXPECT_EQ(NIXL_ERR_REMOTE_DISCONNECT, status);
+            if (test_type == TestType::XFER_FAIL_RESTORE) {
+                m_Initiator.disconnect();
+                m_Target.init(target_name, m_backend_name);
```
rakhmets, line 254: It seems that m_backend wasn't properly cleaned before creating a new one.

evgeny-leksikov: actually it's not needed, but added resetting with `nullptr` for cleanliness

rakhmets: `m_Target.init` at line 244 creates a nixlBackendH, and stores a pointer to it. `m_Target.init` at line 256 creates another nixlBackendH, and stores a pointer to it, but the old one is not deleted. You can use `unique_ptr` to automatically clean up resources, or call delete manually.

evgeny-leksikov: Could you point to correct way to delete the old one? AFAIU Agent and PluginManager should handle it automatically, I don't see symmetric API to destroy the handle, examples don't delete it too.

rakhmets: Yes, there is no need to destroy this backend, since the new `nixlAgent` is created, and the old one destroys all created handles.

test/gtest/error_handling.cpp
```diff
@@ -138,11 +143,24 @@ void TestErrorHandling::Agent::init(const std::string& name, const std::string&
...
     m_priv->deregisterMem(m_mem.m_dlist, &m_mem.m_params);
+    m_backend = nullptr;
     m_priv.reset();
 }
 
+void
+TestErrorHandling::Agent::disconnect() {
+    ASSERT_FALSE(m_MetaRemote.empty());
+
+    const nixl_status_t status = m_priv->invalidateRemoteMD(m_MetaRemote);
```
evgeny-leksikov: - need to move to err handler of Agent
- add a case with double invalidation

src/core/nixl_agent.cpp
```diff
@@ -1179,60 +1198,37 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
         return ret;
     }
 
-    for (size_t i=0; i<conn_cnt; ++i) {
+    count = 0;
```
```suggestion
    size_t count = 0;
```
, and remove from beginning of the function.

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+        if (remoteBackends[remote_name][backend] != conn_info) {
+            return NIXL_ERR_NOT_ALLOWED;
+        }
+
+        return NIXL_SUCCESS;
+    }
+
+    nixlBackendEngine *eng = backendEngines[backend];
+    if (!eng->supportsRemote()) {
+        return NIXL_ERR_UNKNOWN;
```
```suggestion
        NIXL_ERROR << backend << " does not support remote operations";
        return NIXL_ERR_NOT_SUPPORTED;
```

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
     req_list = std::move(commQueue);
     commQueue.clear();
 }
+
+nixl_status_t
+nixlAgentData::loadConnInfo(const std::string &remote_name,
+                            const nixl_backend_t &backend,
+                            const nixl_blob_t &conn_info) {
+    // Current agent might not support a remote backend
```
rakhmets: Maybe better to put this into `NIXL_ERROR`

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+
+        return NIXL_SUCCESS;
+    }
+
+    nixlBackendEngine *eng = backendEngines[backend];
+    if (!eng->supportsRemote()) {
+        return NIXL_ERR_UNKNOWN;
+    }
+
+    nixl_status_t ret = eng->loadRemoteConnInfo(remote_name, conn_info);
```
rakhmets: const

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlAgentData::loadRemoteSections(const std::string &remote_name, nixlSerDes &sd) {
+    if (remoteSections.count(remote_name) == 0) {
+        remoteSections[remote_name] = new nixlRemoteSection(remote_name);
+    }
+
+    nixl_status_t ret = remoteSections[remote_name]->loadRemoteData(&sd, backendEngines);
```
rakhmets: const

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+        return ret;
+    }
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlAgentData::invalidateRemoteData(const std::string &remote_name) {
+    if (remote_name == name) {
+        return NIXL_ERR_INVALID_PARAM;
```
rakhmets, line 670: NIXL_ERROR?

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,81 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+    if (it_backends != remoteBackends.end()) {
+        for (auto &it : it_backends->second) {
+            backendEngines[it.first]->disconnect(remote_name);
+        }
+
+        remoteBackends.erase(it_backends);
+        ret = NIXL_SUCCESS;
+    }
+
+    return ret;
```
rakhmets, line 691: Is this a valid case when one of the containers does not contain a key?

src/core/nixl_agent.cpp
```diff
@@ -877,6 +883,12 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
                                       req_hndl->remoteAgent,
                                       req_hndl->backendHandle,
                                       &opt_args);
+    if (ret == NIXL_ERR_REMOTE_DISCONNECT) {
+        data->invalidateRemoteData(req_hndl->remoteAgent);
+        delete req_hndl;
```
rakhmets, line 938: It looks like these two lines should be placed in a separate function to avoid code duplication.

src/core/nixl_agent.cpp
```diff
@@ -1151,25 +1168,27 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
...
     NIXL_LOCK_GUARD(data->lock);
     ret = sd.importStr(remote_metadata);
-    if(ret)
+    if (ret != NIXL_SUCCESS) {
         return ret;
+    }
 
     std::string remote_agent = sd.getStr("Agent");
-    if (remote_agent.size() == 0)
+    if (remote_agent.size() == 0) {
```
iyastreb: empty() is more C++ idiomatic

src/core/nixl_agent.cpp
```diff
@@ -1179,60 +1198,37 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
         return ret;
     }
 
-    for (size_t i=0; i<conn_cnt; ++i) {
+    count = 0;
+    for (size_t i = 0; i < conn_cnt; ++i) {
         nixl_backend = sd.getStr("t");
-        if (nixl_backend.size() == 0)
+        if (nixl_backend.size() == 0) {
```
iyastreb: empty()

src/core/nixl_agent.cpp
```diff
@@ -1179,60 +1198,37 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
...
             return NIXL_ERR_MISMATCH;
+        }
+
         conn_info = sd.getStr("c");
-        if (conn_info.size() == 0)
+        if (conn_info.size() == 0) {
```
iyastreb: empty()

src/core/nixl_listener.cpp
evgeny-leksikov: I don't think so, but the logic was moved "as is", better to fix separately if needed.

src/core/nixl_listener.cpp
```diff
@@ -647,3 +647,83 @@ void nixlAgentData::getCommWork(std::vector<nixl_comm_req_t> &req_list){
...
+            return NIXL_ERR_NOT_ALLOWED;
+        }
+
+        return NIXL_SUCCESS;
+    }
+
+    nixlBackendEngine *eng = backendEngines[backend];
+    if (!eng->supportsRemote()) {
+        NIXL_ERROR << backend << " does not support remote operations";
+        return NIXL_ERR_UNKNOWN;
```
rakhmets: NIXL_ERR_NOT_SUPPORTED?

src/core/transfer_request.h
```diff
@@ -34,6 +38,19 @@ class nixlXferReqH {
...
 
+        nixl_status_t
+        handleFailure(nixlAgentData &data, nixl_status_t status) {
+            NIXL_DEBUG << "Handling failure: " << nixlEnumStrings::statusStr(status);
+
+            if (status == NIXL_ERR_REMOTE_DISCONNECT) {
+                data.invalidateRemoteData(remoteAgent);
+            }
+
+            delete this;
```
rakhmets: I'm afraid it could cause problems.
I suggest to keep `delete` in scope of `nixlAgent`, since the corresponding `new` operator is located there.
I understand that using raw pointers is a problem in itself. But I hope that putting `new` and `delete` in the same scope will make it easier to refactor in the future.

evgeny-leksikov: the change was introduced to resolve https://github.com/ai-dynamo/nixl/pull/509#discussion_r2215602338 (this is the best place I found), so now you are asking to move `delete` back?

rakhmets: I meant a private method in `nixlAgent`.
And I'd suggest to avoid using `delete this;` as it is a rather fragile design with strict requirements:
https://isocpp.org/wiki/faq/freestore-mgmt#delete-this

evgeny-leksikov: > I meant a private method in `nixlAgent`. And I'd suggest to avoid using `delete this;` as it is a rather fragile design with strict requirements: https://isocpp.org/wiki/faq/freestore-mgmt#delete-this

`nixlAgent` is API class and defined in `nixl.h` I would avoid adding helper functions there even in private section

rakhmets: I see. In that case, I would prefer to leave `delete` in `nixlAgent`.

evgeny-leksikov: ok, reverted

rakhmets (approved):

src/core/nixl_agent.cpp
```diff
@@ -1222,88 +1239,66 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
...
+        if (ret == NIXL_SUCCESS) {
+            count++;
+        } else if (ret != NIXL_ERR_NOT_SUPPORTED) {
+            return ret;
         }
     }
 
     // No common backend, no point in loading the rest, unexpected
-    if (count == 0 && conn_cnt > 0)
+    if ((count == 0) && (conn_cnt > 0)) {
```
yosefe: Seems major part (~50%) of this PR is refactoring / style changes that should have been in a separate PR.
It is very important to avoid this. We cannot create a dependency between review of important bugfixes and review of nice-to-have refactoring/style fixes.
Please move the refactoring to another PR and keep only invalidateRemoteData() function since it has to be used from multiple places.

test/gtest/error_handling.cpp
```diff
@@ -124,22 +131,42 @@ class TestErrorHandling : public testing::TestWithParam<std::string> {
...
+void
+TestErrorHandling::Agent::destroy(bool after_failure) {
+    disconnect(after_failure);
     m_priv->deregisterMem(m_mem.m_dlist, &m_mem.m_params);
+    m_backend = nullptr;
     m_priv.reset();
 }
 
+void
+TestErrorHandling::Agent::disconnect(bool after_failure) {
```
yosefe: seems this function is redundant. And the parameter "after_failure" is weird because it describes how the functions should be used and not what the function is doing.
Why need to clear m_MetaRemote string?
And we could move invalidateRemoteMD call to destroy(), there is no need to do it during err handling. If we want to check that MD is actually deleted we better use checkRemoteMD

src/core/nixl_agent.cpp
evgeny-leksikov: >Seems major part (~50%) of this PR is refactoring / style changes that should have been in a separate PR.

This is a side effect of strict code review and continuously changing of code style rules in clang-format. The PR was started as a simple code movement of couple functions + the fix.
In general I like your proposed approach but while most of existing code does not match required code style, the formatting is required to pass CI. Maybe we could apply clang-format to whole code base in single commit to avoid this? 

> keep only invalidateRemoteData() function

do you mean keep it for this PR and move others to another PR with refactoring or do not touch other places at all?

evgeny-leksikov: So, code style PR: https://github.com/ai-dynamo/nixl/pull/678
refactoring PR: https://github.com/ai-dynamo/nixl/pull/683
this PR implements only `invalidateRemoteData`

yosefe (approved):

PR #493, opened 2025-06-23
add mooncake notify support

haobayuxi:
## What?
To support the Notify function in Mooncake backend. GenNofity interface is also supported.
The client can send the data with the notification, and the server can retrieve the notification through the getNotifs interface.
The test is in test/unit/plugins/mooncake/mooncake_backend_test.

The latest main branch of Mooncake is need to downloaded and compiled in order to use the notify.

PR #514, opened 2025-06-30
CORE: backends that support remote must support notifications

tstamler:
## What?
Add a check in backend creation that forces newly created backends that support remote transfers to also support notifications. 

## Why?
Backends with remote transfers that don't have notifications aren't particularly useful, many inference frameworks require them to work properly.


tstamler:
depends on https://github.com/ai-dynamo/nixl/pull/493

tstamler:
also apparently reformat the entire src/core/nixl_agent.cpp?

PR #494, opened 2025-06-23
Integration of GUSLI plugin

danielhe-nvidia:
Change-Id: I05e7b35eb3076142b4dcc1f58a68865d5a5962c0

## What?
Review only: Don't merge, as GUSLI standalone repository still do not exist
Support transfer of 1 element in a list
Support transfer of N element in a list, with/Without scattaer gather extra element
Support the above to multiple block devices in one transfer

## Why?
Allow using gusli access (to server or bdevs) via nixl

PR #515, opened 2025-06-30
DOCS: Add Backend Developer Guide

tstamler:
## What?
Introduce the backend developer API overview to the repo. 

## Why?
Provide a single, standardized document that backend contributors can refer to.


docs/BackendGuide.md
```diff
@@ -0,0 +1,146 @@
...
+
+![NIXL SB API](figures/nixl_sb_api.png)
+
+## The South Bound API
+
+For a backend to be compatible with NIXL, it must implement several key SB API methods including:
+
+### Constructor and Destructor:
+
+* Constructor: A key/value set of parameters alongside Agent name is passed to the backend.
```
w1ldptr, line 44: Can you expand a bit on this? On [recent code reviews](https://github.com/ai-dynamo/nixl/pull/494#discussion_r2167469181) I've seen people confused by how copy of custom params is saved by base class nixlBackendEngine, their lifetime, etc.

docs/BackendGuide.md
```diff
@@ -0,0 +1,146 @@
...
+If a backend requires a progress call, such as UCX, to proceed with the transfers, for both check of transfer status or received notification, they can implement a progress thread, and a frequency of waking up that thread will be passed during backend creation. In addition, each time a user calls to check a transfer status, or check received notifications, this method is called, enabling progress if a progress thread is not implemented.
+
+## Descriptor List Abstraction
+
+A key underlying abstraction for NIXL library is a descriptor list, that is made of a memory space (host/GPU/block/File/Obj-Store) and a list of descriptors. There are 2 types of descriptors used for the SB API.
+
+*For transfers: (addr, len, devID, metadata), where metadata is a pointer to an nixlBackendMD object relevant to the registered memory that this descriptor falls within.
+*For registration, (addr, len, devID, str) where str is an optional byte-array for extra information. The table below shows the meaning of devID for different memory spaces, as well as optional meaning for File and Object-Store.
+
+![NIXL SB API](figures/nixl_desc_table.png)
```
w1ldptr: Does this needs to be an image? Markdown natively supports tables and text-based formats are much easier to edit, scale with resolution, will be rendered for the platform with desired font, size and AA, etc.

docs/BackendGuide.md
```diff
@@ -0,0 +1,146 @@
...
+
+This is achieved through the South Bound API (SB API) that serves as the standardized interface between NIXL's Transfer Agent and various backend plugins. NIXL agent handles bookkeeping for the local registered memories with backend plugins, as well as managing the required metadata for one-sided transfers, i.e., Read and Write operations, for local or remote transfers. The following diagram further illustrates these components:
+
+![Figure of NIXL high level architecture](figures/nixl_high_level.png)
```
w1ldptr, line 24: Can the diagrams be in some format easier to edit? PNGs are not trivial to edit and git diff wouldn't provide any useful info on their history. Can't this be done with simple ASCII-diagram or, even you want it to look prettier, [Mermaid](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/)?

docs/BackendGuide.md
```diff
@@ -0,0 +1,146 @@
...
+
+getPublicData and loadRemoteMD are required if backend supportsRemote, and loadLocalMD is required if backend supportsLocal, and unloadMD is required in all cases to release the deserialized remote identifier object.
+
+### Transfer Operations:
+
+* prepXfer(): Given a descriptor list on each side, read or write operation, and remote agent name (can be loopback to itself if supported), any preparation for a transfer can be performed here, generating a pointer nixlBackendReqH that is a base class to be inherited by the backend for storing state of transfer request.
+* estimateXferCost: Given the same info as prepXfer, as well as the transfer request output from prepXfer, the backend can estimate the time of transfer, with noise margin and method of estimation. This is optional.
+* postXfer(): Posts a transfer request, meaning the backend should start the transfer. This call is asynchronous, meaning it should not wait to finish the transfer. If the transfer is really small, it’s fine to return DONE right after this call.
+* checkXfer(): Checks the status of a transfer request.
```
w1ldptr, line 94: This is very light on details. For example, contributors tend to get confused by the fact that the handle can be deallocated by core after xfer has been scheduled but before it has been completed [over](https://github.com/ai-dynamo/nixl/pull/494#discussion_r2167490627) and [over](https://github.com/ai-dynamo/nixl/pull/461#discussion_r2150040151).

docs/BackendGuide.md
tstamler: yeah, it was ripped from powerpoint, but it's easy enough to recreate in markdown, so I've done it

docs/BackendGuide.md
tstamler: the diagrams are just taken from powerpoints we've been using for customers. At the moment, remaking them all in ASCII or mermaid doesn't seem worth the time. If we get to a point where we need to be updating these diagram regularly, maybe we can reconsider.

docs/BackendGuide.md
tstamler: I'm going to leave some inline suggestions for others to check in a separate review, I'll try to expand on this.

docs/BackendGuide.md
tstamler: Same as above, I'll mention in the destructor that handle can be deallocated but I'm also not looking to add a lot of extra text at the moment as the document has already been reviewed/approved for customer use

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+### Constructor and Destructor:
+
+* Constructor: A key/value set of parameters alongside Agent name is passed to the backend.
+* Destructor: Release the remaining resources.
+
+### Capability Indicators:
+
+* supportsLocal(): Indicates if the backend supports transfers within a node
+* supportsRemote(): Indicates if the backend supports transfers across nodes
+* supportsNotif(): Indicates if the backend supports notifications
```
tstamler, line 53:
```suggestion
* supportsNotif(): Indicates if the backend supports notifications. A backend that supportsRemote() _must_  also support notifications.
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+![NIXL SB API](figures/nixl_sb_api.png)
+
+## The South Bound API
+
+For a backend to be compatible with NIXL, it must implement several key SB API methods including:
+
+### Constructor and Destructor:
+
+* Constructor: A key/value set of parameters alongside Agent name is passed to the backend.
+* Destructor: Release the remaining resources.
```
tstamler, line 45:
```suggestion
* Destructor: Release the remaining resources.

The key/value parameters are a map of strings to byte arrays that are passed from the Agent. These can be whatever the backend needs as its initialization parameters. See the `get_backend_options` plugin API for more detail on how to specify these. 
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+
+getPublicData and loadRemoteMD are required if backend supportsRemote, and loadLocalMD is required if backend supportsLocal, and unloadMD is required in all cases to release the deserialized remote identifier object.
+
+### Transfer Operations:
+
+* prepXfer(): Given a descriptor list on each side, read or write operation, and remote agent name (can be loopback to itself if supported), any preparation for a transfer can be performed here, generating a pointer nixlBackendReqH that is a base class to be inherited by the backend for storing state of transfer request.
+* estimateXferCost: Given the same info as prepXfer, as well as the transfer request output from prepXfer, the backend can estimate the time of transfer, with noise margin and method of estimation. This is optional.
+* postXfer(): Posts a transfer request, meaning the backend should start the transfer. This call is asynchronous, meaning it should not wait to finish the transfer. If the transfer is really small, it’s fine to return DONE right after this call.
+* checkXfer(): Checks the status of a transfer request.
+* releaseReqH(): Releases a transfer request handle.
```
tstamler:
```suggestion
* releaseReqH(): Releases a transfer request handle. Note that if you have extended the nixlBackendReqH to track request state as described above, that the NIXL agent may release that handle at a number of error cases. Because of this, this function should handle proper cancelling and teardown of requests. 
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+
+Note that after this step, the local agent cannot initiate transfers to the remote agent anymore, until its metadata is reloaded again. Through these two methods the dynamicity requirement of NIXL agents are achieved.
+
+Regarding marginal updates, they only require some extra bookkeeping within the agent, and central KV service mode is just the interface between the agent and central metadata server, and does not change how the SB APIs are used.
+
+### Create transfer request:
+
+This API does the preparations on the agent side and does not call the backend SB API. However, it decides which backend to choose (unless optionally specified by the user). If a backend is not specified, the agent will look at the memory types of the request on both sides, the available backend engines on both sides, as well as the memory ranges registered with each backend for that specific memory type. Usually considering all these factors only a single backend can deliver the transfer request, otherwise we select the first match, or use a preference list.
```
tstamler, line 204:
```suggestion
This API does the preparations on both the agent side and backend request through the SB API **prepXfer**. This function must be implemented, but can be left empty in some cases if it doesn't make sense to prepare request handles or state ahead of time for a specific backend.  

At this stage, the NIXL Agent also decides which backend to choose (unless optionally specified by the user). If a backend is not specified, the agent will look at the memory types of the request on both sides, the available backend engines on both sides, as well as the memory ranges registered with each backend for that specific memory type. Usually considering all these factors only a single backend can deliver the transfer request, otherwise we select the first match, or use a preference list.
```

tstamler (commented):
Left a few text changes for @mkhazraee or @vvenkates27 to review.

docs/BackendGuide.md
mkhazraee: I think it is the definition of that flag. So the next line add that point.

docs/BackendGuide.md
mkhazraee: string to byte array

docs/BackendGuide.md
mkhazraee: I'm a tad confused by this, is this what you meant or I'm missing some other aspect?
Note that this method is not just for releasing the resources after the transfer is done. NIXL agent might use it to try to abort a request in case of an error too. So this might be called before end of a request, for instance due to some errors the NIXL agent has called it. So it should support proper way to cancel (or try to and return error) and teardown of requests.

docs/BackendGuide.md
mkhazraee: Good catch.

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+
+* Constructor: A key/value set of parameters alongside Agent name is passed to the backend.
+* Destructor: Release the remaining resources.
+
+### Capability Indicators:
+
+* supportsLocal(): Indicates if the backend supports transfers within a node
+* supportsRemote(): Indicates if the backend supports transfers across nodes
+* supportsNotif(): Indicates if the backend supports notifications
+* supportsProgressThread(): Indicates if the backend requires calls to a progress method, potentially in a separate thread
```
mkhazraee: supportsProgressThread(): Indicates if the backend requires calls to a progress method, in case of central progress thread.

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+* getSupportedMems(): Indicates memory types supported by the backend
+
+Based on the first 4 methods (supports*), the required methods to be implemented change. For instance, UCX backend implements all as it supports all scenarios, while GDS backend only has supportsLocal, detailed more in Example implementations. Note that a network backend should have supportsRemote and supportsNotif to be set to true, and preferably supportsLocal also to true, so another backend doesn’t need to be involved for local transfers. For a storage backend, it should have supportsLocal and supportsNotif is optional. supportsProgressThread is optional for both cases.
+
```
mkhazraee, line 58: Note that supportProgressThread is an indicator whether a backend can support progress thread, but does not indicate a need or lack of need for a progress thread. During init params set to a backend, enablement of progress thread is set, and if a backend cannot work without progress thread while the parameter is set to use, or vice versa, the backend creation would fail. This flag is useful for the NIXL agent in case of running in Multi-threaded mode to know which backends to check if such init parameter was passed to them. Also in future there might be a mode with a central progress method that calls progress of backends that set this flag. However, the timeline for that feature is not clear since there is a potential for contention across different backend progress threads and its benefits need to be studied.

docs/BackendGuide.md
mkhazraee: I see Vlad mentioned some confusion, I guess I didn't quite follow the corner case, so the explanation can be improved.

docs/BackendGuide.md
tstamler, line 58:
```suggestion

Note that supportProgressThread is an indicator whether a backend can support progress thread, but does not indicate a need or lack of need for a progress thread. During init params set to a backend, enablement of progress thread is set, and if a backend cannot work without progress thread while the parameter is set to use, or vice versa, the backend creation would fail. This flag is useful for the NIXL agent in case of running in Multi-threaded mode to know which backends to check if such init parameter was passed to them. Also in future there might be a mode with a central progress method that calls progress of backends that set this flag. However, the timeline for that feature is not clear since there is a potential for contention across different backend progress threads and its benefits need to be studied.
```

docs/BackendGuide.md
tstamler:
```suggestion
* supportsProgressThread(): Indicates if the backend requires calls to a progress method, in case of central progress thread.
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,228 @@
...
+### Capability Indicators:
+
+* supportsLocal(): Indicates if the backend supports transfers within a node
+* supportsRemote(): Indicates if the backend supports transfers across nodes
+* supportsNotif(): Indicates if the backend supports notifications
+* supportsProgressThread(): Indicates if the backend requires calls to a progress method, potentially in a separate thread
+* getSupportedMems(): Indicates memory types supported by the backend
+
+Based on the first 4 methods (supports*), the required methods to be implemented change. For instance, UCX backend implements all as it supports all scenarios, while GDS backend only has supportsLocal, detailed more in Example implementations. Note that a network backend should have supportsRemote and supportsNotif to be set to true, and preferably supportsLocal also to true, so another backend doesn’t need to be involved for local transfers. For a storage backend, it should have supportsLocal and supportsNotif is optional. supportsProgressThread is optional for both cases.
```
tstamler:
```suggestion
Based on the first 4 methods (supports*), the required methods to be implemented change. For instance, UCX backend implements all as it supports all scenarios, while GDS backend only has supportsLocal, detailed more in Example implementations. Note that a network backend should have supportsRemote and supportsNotif to be set to true, and preferably supportsLocal also to true, so another backend doesn’t need to be involved for local transfers. For a storage backend, it should have supportsLocal and supportsNotif is optional. supportsProgressThread is optional for both cases. Additionally, a backend that supportsRemote must also support supportNotifs.
```

docs/BackendGuide.md
tstamler: will add below

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Transfer Operations:
+
+* prepXfer(): Given a descriptor list on each side, read or write operation, and remote agent name (can be loopback to itself if supported), any preparation for a transfer can be performed here, generating a pointer nixlBackendReqH that is a base class to be inherited by the backend for storing state of transfer request.
+* estimateXferCost: Given the same info as prepXfer, as well as the transfer request output from prepXfer, the backend can estimate the time of transfer, with noise margin and method of estimation. This is optional.
+* postXfer(): Posts a transfer request, meaning the backend should start the transfer. This call is asynchronous, meaning it should not wait to finish the transfer. If the transfer is really small, it’s fine to return DONE right after this call.
+* checkXfer(): Checks the status of a transfer request.
+* releaseReqH(): Releases a transfer request handle. Note that if you have extended the nixlBackendReqH to track request state as described above, that the NIXL agent may release that handle at a number of error cases. Because of this, this function should handle proper cancelling and teardown of requests.
```
tstamler:
```suggestion
* releaseReqH(): Releases a transfer request handle. Note that if you have extended the nixlBackendReqH to track request state as described above, that the NIXL agent may release that handle at a number of error cases. Because of this, this function should handle proper cancelling and teardown of requests.
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+## The South Bound API
+
+For a backend to be compatible with NIXL, it must implement several key SB API methods including:
+
+### Constructor and Destructor:
+
+* Constructor: A key/value set of parameters alongside Agent name is passed to the backend.
+* Destructor: Release the remaining resources.
+
+The key/value parameters are a map of strings to byte arrays that are passed from the Agent. These can be whatever the backend needs as its initialization parameters. See the `get_backend_options` plugin API for more detail on how to specify these.
```
tstamler:
```suggestion
The key/value parameters are a map of strings to byte arrays that are passed from the Agent. These can be whatever the backend needs as its initialization parameters. See the `get_backend_options` plugin API for more detail on how to specify these.
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Capability Indicators:
+
+* supportsLocal(): Indicates if the backend supports transfers within a node
+* supportsRemote(): Indicates if the backend supports transfers across nodes
+* supportsNotif(): Indicates if the backend supports notifications
+* supportsProgressThread(): Indicates if the backend requires calls to a progress method, in case of central progress thread.
```
mkhazraee: Indicates if the backend supports progress() method. That method should call the underlying procedure of progressing transfers for this backend.

(UCX doesn't require calls but set this to true, in our slides we mention progress should be implemented if this flag is set, so let it be just that and we provide further details in 2 paragraphs.)

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+
+Note that supportProgressThread is an indicator whether a backend can support progress thread, but does not indicate a need or lack of need for a progress thread. During init params set to a backend, enablement of progress thread is set, and if a backend cannot work without progress thread while the parameter is set to use, or vice versa, the backend creation would fail. This flag is useful for the NIXL agent in case of running in Multi-threaded mode to know which backends to check if such init parameter was passed to them. Also in future there might be a mode with a central progress method that calls progress of backends that set this flag. However, the timeline for that feature is not clear since there is a potential for contention across different backend progress threads and its benefits need to be studied.
```
mkhazraee: Note that supportProgressThread is an indicator whether a backend has implemented the progress() method, but does not imply how the progress thread is implemented. During creation of a backend, the provided init params indicate how the progress thread is intended to be used. For instance, if the enablement of progress thread is set to false, while a backend cannot work without a separate progress thread, the backend creation would fail. This flag is useful for the NIXL agent if we want to provide some agent level guarantees, such as minimum time between calls to progress for backends, or if a central progress method is implemented (for future proofing, not currently implemented).

docs/BackendGuide.md
```diff
@@ -89,7 +92,7 @@ getPublicData and loadRemoteMD are required if backend supportsRemote, and loadL
 * estimateXferCost: Given the same info as prepXfer, as well as the transfer request output from prepXfer, the backend can estimate the time of transfer, with noise margin and method of estimation. This is optional.
 * postXfer(): Posts a transfer request, meaning the backend should start the transfer. This call is asynchronous, meaning it should not wait to finish the transfer. If the transfer is really small, it’s fine to return DONE right after this call.
 * checkXfer(): Checks the status of a transfer request.
-* releaseReqH(): Releases a transfer request handle.
+* releaseReqH(): Releases a transfer request handle. Note that if you have extended the nixlBackendReqH to track request state as described above, that the NIXL agent may release that handle at a number of error cases. Because of this, this function should handle proper cancelling and teardown of requests.
```
mkhazraee: I'm a tad confused, "if you have extended the nixlBackendReqH to track request state as described above", that sounds like a must thing to do, so maybe redundant? And is the issue just in case of abort, due to some error or user request? Or there is another scenario that this method might be called?

docs/BackendGuide.md
tstamler:
```suggestion
* supportsProgressThread(): Indicates if the backend supports progress() method. That method should call the underlying procedure of progressing transfers for this backend.
```

docs/BackendGuide.md
tstamler:
```suggestion
Note that supportProgressThread is an indicator whether a backend has implemented the progress() method, but does not imply how the progress thread is implemented. During creation of a backend, the provided init params indicate how the progress thread is intended to be used. For instance, if the enablement of progress thread is set to false, while a backend cannot work without a separate progress thread, the backend creation would fail. This flag is useful for the NIXL agent if we want to provide some agent level guarantees, such as minimum time between calls to progress for backends, or if a central progress method is implemented (for future proofing, not currently implemented).
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Transfer Operations:
+
+* prepXfer(): Given a descriptor list on each side, read or write operation, and remote agent name (can be loopback to itself if supported), any preparation for a transfer can be performed here, generating a pointer nixlBackendReqH that is a base class to be inherited by the backend for storing state of transfer request.
+* estimateXferCost: Given the same info as prepXfer, as well as the transfer request output from prepXfer, the backend can estimate the time of transfer, with noise margin and method of estimation. This is optional.
+* postXfer(): Posts a transfer request, meaning the backend should start the transfer. This call is asynchronous, meaning it should not wait to finish the transfer. If the transfer is really small, it’s fine to return DONE right after this call.
+* checkXfer(): Checks the status of a transfer request.
+* releaseReqH(): Releases a transfer request handle. Note that if you have extended the nixlBackendReqH to track request state as described above, that the NIXL agent may release that handle at a number of error cases. Because of this, this function should handle proper cancelling and teardown of requests.
```
tstamler:
```suggestion
* releaseReqH(): Releases a transfer request handle, which should be an extension of the nixlBackendReqH. Note that the NIXL agent may release that handle at a number of error cases, expecting this function to handle proper cancellation of requests in addition to freeing resources.
```

docs/BackendGuide.md
```diff
@@ -0,0 +1,231 @@
...
+* checkXfer(): Checks the status of a transfer request.
+* releaseReqH(): Releases a transfer request handle, which should be an extension of the nixlBackendReqH. Note that the NIXL agent may release that handle at a number of error cases, expecting this function to handle proper cancellation of requests in addition to freeing resources.
+
+Within each transfer request, a descriptor list is passed, if there is room for parallelization across different contiguous memory locations, such as across different GPUs (one transfer can expand multiple GPUs). Optionally the user might ask for a notification, which should be sent after all the descriptors within a transfer request are sent. If a backend does not set supportsNotifications, no such notification will be asked.
+
+Note that any transfer request will be prepped only once, but can be posted multiple times, as long as it gets to DONE state before getting reposted. There is no ordering guarantee across transfer requests, and no locking mechanism for any specific memory region; the user is in charge of not corrupting the memory by having two simultaneous transfers to the same location.
+
```
mkhazraee, line 100: Today a question came up regarding releaseReqH, so let's add that too. Please feel free to reword, or move in the text.

Finally, note that a call to releaseXferReq should not block and be asynchronous, but it can return error, meaning it was not successful in cancelling (aborting) the request. If a backend can abort a transfer quickly without a stall, then releaseXferReq can return success right away. Otherwise, one option would be to wait for the transfer to complete. In the meantime checkXferReq should return error if there was a call to abort which was not successful, while calls to releaseXferReq would return error until the transfer is completed and it returns success. If there is a scenario that a blocking call can abort the transfer before it is completed, then that blocking call can start in a separate thread (or within the progress thread of the backend). In other words, use a blocking call under the hood but provide non-blocking APIs to the user.

tstamler, line 100:
```suggestion

Finally, note that a call to releaseXferReq should not block and be asynchronous, especially important to remember when aborting a transfer. This function can return error, meaning it was not successful in aborting the request. If a backend can abort a transfer quickly without a stall, then releaseXferReq can return success right away. Otherwise, one option would be to wait for the transfer to complete. In the meantime, checkXferReq should return error if there was a call to abort which was not successful, while calls to releaseXferReq would return error until the transfer is completed and it returns success. If there is a scenario that a blocking call can abort the transfer before it is completed, then that blocking call can start in a separate thread (or within the progress thread of the backend). In other words, use a blocking call under the hood but provide non-blocking APIs to the user.
```

mkhazraee (approved):

PR #518, opened 2025-07-01
.GITLAB: add UCX pkg-config to env

roiedanino:
## What?
Add UCX pkg-config to path in test_python.sh

## Why?
Otherwise meson can't find UCX when building and running python tests

### DIFF
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -19,9 +19,10 @@ set -x
 
 # Parse commandline arguments with first argument being the install directory.
 INSTALL_DIR=$1
+UCX_INSTALL_DIR=$2
 
-if [ -z "$INSTALL_DIR" ]; then
-    echo "Usage: $0 <install_dir>"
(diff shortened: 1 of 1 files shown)


.gitlab/test_python.sh
```diff
@@ -19,9 +19,10 @@ set -x
 
 # Parse commandline arguments with first argument being the install directory.
 INSTALL_DIR=$1
+UCX_INSTALL_DIR=$2
 
-if [ -z "$INSTALL_DIR" ]; then
-    echo "Usage: $0 <install_dir>"
+if [ -z "$INSTALL_DIR" ] || [ -z "$UCX_INSTALL_DIR" ]; then
```
dpressle, line 24: Default behavior today is to put both Nixl and UCX in same install dir which is how all build/tests scripts works today.
Your change forces us to provide both places even when it same place, i would have it as optional or you need to change all the places that calls this script (gitlab, aws, blossom etc)

PR #519, opened 2025-07-01
GDS multi thread fixes

4D0R:
## What?
Re-opens the **GDS-MT backend** work from PR #461 (originally by @sdimitro) so it can land while the author is OOO.

### Delta vs PR #461
* **Constructor clean-up** – thread-count logic moved to a helper; ctor now uses an initializer list only.
* **RAII fixes** – `deregisterMem()` and `registerMem()` switched to `unique_ptr` for leak-free cleanup.
* **Helper functions** moved to an anonymous namespace; duplicate definitions removed.
* **Interface Cleanup** Interface classes (nixlGdsMtMetadata, nixlGdsMtBackendReqH) and their supporting structs remain in header; pure implementation details moved to cpp

No functional changes; just code-quality tweaks.

---

## Why?
We still want the multi-threaded cuFile backend, but the original PR is stalled waiting on owner approvals. This PR is identical to #461 with the minor polish above.

---

## How?
Same design as #461: Taskflow thread-pool around synchronous cuFile I/O.  
Diff limited to `src/plugins/gds_mt/*` plus a few lines removed in plugin registration as per a comment.


src/plugins/gds_mt/gds_mt_backend.cpp
```diff
@@ -36,7 +36,7 @@ namespace {
 const size_t default_thread_count = std::max (1u, std::thread::hardware_concurrency() / 2);
 
 size_t
-get_thread_count (const nixlBackendInitParams *init_params) {
+getThreadCount (const nixlBackendInitParams *init_params) {
```
w1ldptr, line 39: Also need to rename the call site, otherwise it won't compile

PR #520, opened 2025-07-02
[nixlbench] make large_block_iter_factor a configable option

Pyngon:
## What?
Make the factor to reduce the testing iteration in nixlbench a configuration option and make it default to 1

## Why?
I run some performance test recently and see some weird performance data where throughput of block_size above 1MB is lower than block_size below and equal to 1MB, and spent some time to finally hunted down it is because of the factor that decreased the testing iteration. So to prevent others from running into the same issue, make it an option and default to 1.


Pyngon:
@sb17v @w1ldptr @brminich
I have discussed this with @sb17v offline, please help to review and share your thought. Thank you.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -744,8 +744,8 @@ std::variant<double, int> xferBenchNixlWorker::transfer(size_t block_size,
 
     // Reduce skip by 10x for large block sizes
     if (block_size > LARGE_BLOCK_SIZE) {
-        skip /= LARGE_BLOCK_SIZE_ITER_FACTOR;
-        num_iter /= LARGE_BLOCK_SIZE_ITER_FACTOR;
+        skip /= xferBenchConfig::large_block_iter_factor;
```
sb17v: Why is the following missing here?
```
if (skip < MIN_WARMUP_ITERS) {
    skip = MIN_WARMUP_ITERS;
}
```

Number of skips and iterations in the transfer should match the poll

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -55,6 +55,9 @@ DEFINE_uint64(max_block_size, 64 * (1 << 20), "Max size of block \
 DEFINE_uint64(start_batch_size, 1, "Starting size of batch (Default: 1)");
 DEFINE_uint64(max_batch_size, 1, "Max size of batch (starts from 1)");
 DEFINE_int32(num_iter, 1000, "Max iterations");
+DEFINE_int32 (large_block_iter_factor,
+              1,
+              "factor to reduce test iteration when testing large block size(>1MB)");
```
sb17v: @Pyngon, as we talked about offline, setting the default value for large_block_iter_factor to 1 will enable 1000 iterations for the large block sizes. It consumes too much time for comm benchmarks, which ideally saturates the b/w much before for large sizes.

Let's check with @w1ldptr and @brminich.

sb17v (changes requested):
Overall, looks good to me. Please find the minor suggestion

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: Ops, missed this, will add it.

benchmark/nixlbench/src/utils/utils.cpp
Pyngon: @sb17v, one of the purpose is to make large_block_iter_factor more explicit. For people that is not aware of this option, they may see lower result for block size above 1MB.

I think another way to to set the default to 16 and pointed it out in the README so people are aware. What do you think?

Pyngon: I went ahead to change that. Please help to have a look again. Thanks.

sb17v (dismissed):
Looks good to me

Pyngon:
> Looks good to me

Thanks @sb17v.

@brminich @aranadive @ovidiusm Could you help to give an approval if this looks good. Thanks!

ovidiusm (dismissed):

aranadive (approved):

Issue #516, opened 2025-07-01
[nixlbench + posix] Wrong throughput when benchmarking posix backend with nixlbench


Pyngon:
nixlbench call postXfer() num_iter times and assume that the backend will send the data num_iter times, so the total data transferred used by nixlbench to compute throughput is multiplied by num_iter times.

But posix backend would only send the data once even if postXfer() get called multiple times, (based on observation from tcpdump when using a 3fs folder) so the total data transferred should not be multiplied by num_iter times.

barneuman:
Fixed in https://github.com/ai-dynamo/nixl/pull/523

PR #523, opened 2025-07-02
POSIX: Fixed bugs on re-posting requests and added related test case

dfarge:
## What?
_Fixes bugs on re-posting requests and adds related test case_

## How?
_Reset the counter for num of completed transfers within the posix request handle and moved some necessary work from UringQueue::prepIO to UringQueue::submit, that wouldn't otherwise happen on reposting to the same handle._


barneuman:
#516

PR #527, opened 2025-07-03
chore: Update release version for 0.4.0

pvijayakrish:
## What?
Update release version for 0.4.0

### DIFF
--- a/.ci/jenkins/pipeline/proj-jjb.yaml
+++ b/.ci/jenkins/pipeline/proj-jjb.yaml
@@ -171,7 +171,7 @@
       - string:
           name: "NIXL_VERSION"
           default: "{jjb_branch}"
-          description: "NIXL version to use (tag like 0.3.1, branch name, or commit hash)"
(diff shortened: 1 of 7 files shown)


.ci/jenkins/pipeline/proj-jjb.yaml
```diff
@@ -171,7 +171,7 @@
       - string:
           name: "NIXL_VERSION"
           default: "{jjb_branch}"
-          description: "NIXL version to use (tag like 0.3.1, branch name, or commit hash)"
+          description: "NIXL version to use (tag like 0.4.0, branch name, or commit hash)"
```
dpressle, line 174: Not mandatory its just an example

yosefe (approved):

.ci/jenkins/pipeline/proj-jjb.yaml
yosefe: Indeed seems not mandatory, but since we need this PR for 0.4.0 release i'll merge it as is

PR #529, opened 2025-07-03
UTILS/UCX: Added ucx::rkey class.

rakhmets:
## What?
Added ucx::rkey class.
Applied RAII style.

## Why?
Better resource management and error handling.

### DIFF
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -200,6 +200,6 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
 # Exclude libcuda.so.1 due to compatibility issues, should link with cuda driver library on host
 RUN uv pip install auditwheel && \
     uv run auditwheel repair --exclude libcuda.so.1 --exclude 'libssl*' --exclude 'libcrypto*' /tmp/dist/nixl-*cp3*.whl --plat $WHL_PLATFORM --wheel-dir dist && \
-    contrib/wheel_add_ucx_plugins.py --ucx-lib-dir /usr/lib64 dist/*.whl
+    contrib/wheel_add_ucx_plugins.py --ucx-plugins-dir /usr/lib64/ucx --nixl-plugins-dir /usr/local/nixl/lib64/plugins dist/*.whl
 
 RUN uv pip install dist/nixl-*cp${DEFAULT_PYTHON_VERSION//./}*.whl
--- a/contrib/wheel_add_ucx_plugins.py
+++ b/contrib/wheel_add_ucx_plugins.py
@@ -102,14 +102,15 @@ def get_repaired_lib_name_map(libs_dir):
     (like "libboost_atomic-fb1368c6.so.1.66.0").
     """
     name_map = {}
-    for fname in os.listdir(libs_dir):
+    for fname in sorted(os.listdir(libs_dir)):
         if (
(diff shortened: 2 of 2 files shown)


src/utils/ucx/rkey.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "common/nixl_log.h"
+#include "ucx_utils.h"
+
+namespace nixl::ucx {
+rkey::rkey(const nixlUcxEp &ep, const void *rkey_buffer) noexcept
+    : rkey_{unpackUcpRkey(ep, rkey_buffer), &ucp_rkey_destroy} {}
+
+ucp_rkey *
+rkey::unpackUcpRkey(const nixlUcxEp &ep, const void *rkey_buffer) noexcept {
+    ucp_rkey *rkey = nullptr;
```
brminich: i'd use ucp_rkey_h

rakhmets: Fixed

src/utils/ucx/rkey.h
```diff
@@ -0,0 +1,47 @@
...
+class nixlUcxEp;
+
+namespace nixl::ucx {
+class rkey {
+public:
+    rkey() = delete;
+    rkey(const nixlUcxEp &, const void *rkey_buffer) noexcept;
+
+    [[nodiscard]] ucp_rkey_h
+    getRkey() const noexcept {
```
brminich: maybe just get?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -875,19 +875,12 @@ nixlUcxEngine::internalMDHelper (const nixl_blob_t &blob,
...
-        if (error)
-            // TODO: error out. Should we indicate which desc failed or unroll everything prior
-            break;
-        md->rkeys.push_back(rkey);
-    }
-    if (error) {
-        for (size_t wid = 0; wid < md->rkeys.size(); wid++)
-            md->conn->getEp(wid)->rkeyDestroy(md->rkeys[wid]);
-        return NIXL_ERR_BACKEND;
+        auto rkey = std::make_unique<nixl::ucx::rkey>(*md->conn->getEp(wid), addr.data());
```
iyastreb: Why would we allow to construct an invalid rkey?
In case of an error we should either throw an exception or return unique_ptr(nullptr)
With second scenario we can have a static factory method instead of ctor:
```
static unique_ptr<rkey> createRkey() {
}

rakhmets: I don't want to throw an exception here, as it doesn't happen during initialization...

iyastreb: I don't really get your comment
I was suggesting to modify construction of rkey, so that we don't allow creating semi-valid objects (which is an anti-pattern that we try to get rid of in other places), that we need to validate later before using them. Like we cannot create semi-valid objects with std library. So it does happen during rkey initialization, right? Or we talk about different things
It's also perfectly valid to throw an exception in non-initialization flow..

rakhmets: I meant I tried to avoid throwing an exception here. But I will re-write a code to catch the exception here.

rakhmets: Fixed.

src/utils/ucx/rkey.h
rakhmets: Fixed.

brminich (approved):

PR #531, opened 2025-07-03
queryMem API

mkhazraee:
## What?
Adds queryMem API to NIXL.

## Why?
Users want to query memory, such as available object store keys for KVBM usage.


mkhazraee:
After the base implementation is finalized, Python/C/Rust bindings should be added to this PR as well.

mkhazraee:
After python bindings is added, we can add an exception test as this method returns error by default to test it, as well as the python bindings together.

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,21 @@ enum class nixl_cost_t {
...
+ */
+struct nixlQueryRespArgs {
+    bool accessible;
+    nixl_blob_t info;
+};
+/**
+ * @brief A typedef for nixlQueryArgs
+ *        for querying memory results
+ */
+using nixl_query_resp_t = nixlQueryRespArgs;
```
rakhmets: It looks redundant. `nixlQueryRespArgs` seems to be enough. And `nixl_query_resp_t` contradicts the naming conventions from `docs/CodeStyle.md`.

src/core/nixl_agent.cpp
```diff
@@ -281,6 +281,16 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     return NIXL_ERR_BACKEND;
 }
 
+nixl_status_t
+nixlAgent::queryMem(const nixl_reg_dlist_t &descs,
+                    std::vector<nixl_query_resp_t> *&resp,
+                    const nixl_opt_args_t *extra_params) {
```
rakhmets: If the interface accepts a pointer as a parameter, it can be `nullptr`. We need to check that it is not `nullptr` before using it.

If it is not assumed that it can be `nullptr`, then the parameter must be passed by reference.

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
...
+         * @param  descs         Descriptor list of the buffers to be queried
+         * @param  extra_params  Additional parameters used in querying memory,
+         *                       such as indicating the backend
+         * @param  resp [out]    The output information for the queried descs
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        queryMem(const nixl_reg_dlist_t &descs,
+                 std::vector<nixl_query_resp_t> *&resp,
+                 const nixl_opt_args_t *extra_params);
```
rakhmets: It is not obvious from the interface signature which backend will be used to request information.

The first backend from `nixlAgentOptionalArgs::backends` is used in the implementation. This should be done in a more explicit manner. It might be better to pass a backend handler instead.

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
...
+         *         The backend should be specified via extra_params.
+         *
+         * @param  descs         Descriptor list of the buffers to be queried
+         * @param  extra_params  Additional parameters used in querying memory,
+         *                       such as indicating the backend
+         * @param  resp [out]    The output information for the queried descs
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        queryMem(const nixl_reg_dlist_t &descs,
```
rakhmets: Seems that it should be a `const` method.

src/core/nixl_agent.cpp
```diff
@@ -281,6 +281,16 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
+    if ((extra_params->backends.size() == 0) || (extra_params->backends.size() > 1))
+        return NIXL_ERR_INVALID_PARAM;
+
+    return extra_params->backends[0]->engine->queryMem(descs, resp);
```
rakhmets: The function does not need direct access to the representation of the `nixlAgent` class. So it doesn't look like a part of the class.

src/core/nixl_agent.cpp
```diff
@@ -281,6 +281,16 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     return NIXL_ERR_BACKEND;
 }
 
+nixl_status_t
+nixlAgent::queryMem(const nixl_reg_dlist_t &descs,
+                    std::vector<nixl_query_resp_t> *&resp,
+                    const nixl_opt_args_t *extra_params) {
+    if ((extra_params->backends.size() == 0) || (extra_params->backends.size() > 1))
```
brminich: maybe
```suggestion
    if (extra_params->backends.size() != 1)
```
btw why can't we pass a backend explicitly?

src/api/cpp/backend/backend_engine.h
```diff
@@ -198,5 +184,28 @@ class nixlBackendEngine {
 
         // Force backend engine worker to progress.
         virtual int progress() { return 0; }
+
+
+        // *** Optional virtual methods that is goo to be implemented by any backend *** //
```
rakhmets: Typo `goo` -> `good`.

And maybe the comment is unnecessary. Since the very fact that these functions are not purely virtual suggests that these functions are not required to be implemented.

src/api/cpp/nixl.h
rakhmets: Looking at the parameters and how the function is implemented, it can be concluded that it should not be part of the interface of `nixlAgent`. This function can be an interface of a class, an instance of which is initialized by an object of the corresponding backend (handle).

roiedanino: Consider adding to rust/python bindings + stubs.cpp as well?

src/api/cpp/backend/backend_engine.h
```diff
@@ -198,5 +184,28 @@ class nixlBackendEngine {
...
+        // Estimate the cost (duration) of a transfer operation.
+        virtual nixl_status_t
+        estimateXferCost(const nixl_xfer_op_t &operation,
+                         const nixl_meta_dlist_t &local,
+                         const nixl_meta_dlist_t &remote,
+                         const std::string &remote_agent,
+                         nixlBackendReqH *const &handle,
+                         std::chrono::microseconds &duration,
+                         std::chrono::microseconds &err_margin,
+                         nixl_cost_t &method,
```
iyastreb, line 212: maybe group out fields into a structure, so that we don't need to change API if we need to add new more fields in the future?

rakhmets: I think this is beyond the scope of this PR. This API was added by https://github.com/ai-dynamo/nixl/pull/248. And it's just moved a few lines below by this PR.

src/api/cpp/nixl_types.h
mkhazraee: Mainly following the style of nixlAgentOptionalArgs, but really don't mind any other naming.

src/core/nixl_agent.cpp
mkhazraee: Talking with Yossi, since we will have a API update where most extra_params will be replaced through method overloading, we said let's keep this one similar to the previous ones. Between making it a reference not to be null, or making it default nullptr, the latter is more aligned to what we have now. We might figure out how to aggregate results in the future, not sure, so we can potentially put it to nullptr, and for now just error. Not the best, but when we do the method overloading, this will become clean. First adding with single backend, if we figured it out, add a method with no backend too.

src/api/cpp/nixl.h
mkhazraee: Explained in the previous comment why we stuck to the extra_params mode and not backend as input, in the revised API this issue would go away. So temporary issue, and since KVBM needs it fast, not breaking the overall style of the APIs for now.

src/core/nixl_agent.cpp
mkhazraee: The main issue is the SB API. In case of object store, the backend has to implement this method, and we don't want to ask our collaborators who write plugins to write a plugin for nixl, and then some separate code for this method. We had discussions that maybe someday we would have more control style logic and we would need a separate class, and our collaborators would write a plugin for nixl data path and one for nixl ctrl path, but this is too early to call it, and we try to avoid it. So for a single method, it made more sense to keep it within the same plugin implementation, and then because of that from the nixl agent calling it. Also it's very similar to registerMem call, and the same desc list can be passed to registerMem right away, so nice ease of use there.

src/core/nixl_agent.cpp
mkhazraee: yes, this is better. I explained in a previous comment that Yossi suggested not to break our current API style, as sooner than later we would replace it with method overloading anyways.

src/api/cpp/backend/backend_engine.h
mkhazraee: We already have some comments about the other virtual methods, when they need to be implemented, based on the flags like supportsRemote. So wanted to give a high level idea of these are also virtual, but not bound to any flags. Maybe we can add that notion of not bound to any flags to the comment.

src/api/cpp/nixl.h
mkhazraee: Explained in a previous comment why we want it to be implemented within the same plugin. Technically you can separate it, as in separate instance of the plugin, but it's good to have some state keeping for the backend and I think easier to develop the plugin assuming one place would do the query as well as registration/transfers.

mkhazraee: For sure, waiting for a round of feedback to finalize the API in cpp, and then we need those before merging this PR. Should I assign it to you?

src/api/cpp/backend/backend_engine.h
mkhazraee: There were some long conversations about exactly this, eventually it was concluded to go this way. As Raul said I just moved these lines, as they were next to pure virtual methods, but it isn't.

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,21 @@ enum class nixl_cost_t {
     ANALYTICAL_BACKEND = 0, // Analytical backend cost estimate
 };
 
+/**
+ * @struct nixlQueryArgs
+ * @brief A structure for querying memory results
+ */
+struct nixlQueryRespArgs {
+    bool accessible;
+    nixl_blob_t info;
```
vvenkates27: If we want to use this as a list alternative to fetch all the keys and values?
It would benefit to add two blob's one for key and one for value?

src/core/nixl_agent.cpp
rakhmets: Sorry, I meant if the interface requires the user to pass a valid object, then:
```C++
void foo(const Bar &bar) {
    bar.baz();
}
```
, otherwise
```C++
void foo(const Bar *bar) {
    if (bar == nullptr) {
        // handle nullptr case
    }
    bar->baz();
}
```

If you want to use a pointer, then it should be checked in the implementation before use.

src/api/cpp/backend/backend_engine.h
rakhmets: I see...
So only typo then.

src/core/nixl_agent.cpp
rakhmets: There is no need to add the function to the `nixlAgent`.
It can be just a helper method declared in the same namespace. And since we don't use namespaces in the API, just in the same file.
```C++
nixl_status_t
nixlQueryMem(const nixl_reg_dlist_t &descs,
             std::vector<nixl_query_resp_t> *&resp,
             const nixl_opt_args_t *extra_params);

File: src/api/cpp/nixl_types.h (hunk 1 above)
Comment (mkhazraee) [reply src/api/cpp/nixl_types.h]: Actually it does follow, it's exactly like: 
getAvailPlugins (std::vector<nixl_backend_t> &plugins)

for types we do snake everywhere. And since it's in the header file, we do need the _t version. Or I'm missing some other aspect?

File: src/core/nixl_agent.cpp (hunk 2 above)
Comment (mkhazraee) [reply src/core/nixl_agent.cpp]: oh, didn't realize I missed that if, oops, fixed it.

File: src/core/nixl_agent.cpp (hunk 6 above)
Comment (mkhazraee) [reply src/core/nixl_agent.cpp]: fixed.

File: src/api/cpp/backend/backend_engine.h (hunk 8 above)
Comment (mkhazraee) [reply src/api/cpp/backend/backend_engine.h:212]: Resolving the conversation as not related to this PR, feel free to open a thread on #248 or a new task.

File: src/api/cpp/nixl_types.h (hunk 9 above)
Comment (mkhazraee) [reply src/api/cpp/nixl_types.h]: Replaced it with nixl_b_param_t, the dictionary we had for init params. It's still kind of a backend parameter, so used the same, or we can define a new name. I guess if we end up merging the responses from different backends, we can use this dictionary, backend name being initial part of the key or so.

File: src/core/nixl_agent.cpp (hunk 5 above)
Comment (mkhazraee) [reply src/core/nixl_agent.cpp]: We can translate the user side backend handle to actual backend handle based on the agent info, so at least I have to define this method as friend which is not the best, we tried to minimize it throughout the code. Or you're suggesting another way?

Also personally I'm not fan of utils laying around on the same header file, if we define a separate utils maybe. But as I explained, for a single method we thought not to, and looking at some example libraries like ucx, they have query next to register.

File: src/api/cpp/nixl.h (hunk 4 above)
Comment (mkhazraee) [reply src/api/cpp/nixl.h]: Closing this conversation as we're following on this issue in the other conversation.

File: src/api/cpp/nixl_types.h (hunk 1 above)
Comment (rakhmets) [reply src/api/cpp/nixl_types.h]: I see your point.
We have this guideline in our repo: docs/CodeStyle.md
```
Lower camel style: Data types, Classes, Class Members, Member functions
Snake style: function arguments, local variables
```
And it contradicts what is already used in the API.

Personally, my opinion is that the alias is not needed here.

File: src/core/nixl_agent.cpp (hunk 5 above)
Comment (rakhmets) [reply src/core/nixl_agent.cpp]: No, I didn't mean to use a friend method.
Below I added an example of the code that was meant. And I suggested it because the implementation uses backend from the parameters, not agent info. That is why I suggested to move the function out of the class to follow C++ Core Guidelines https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class

I see your point about UCX. UCX backend (through backend interface) is already a class abstracting the UCX library functionality. And `nixlBackendEngine::queryMem` (added in this PR) is in the right place. However, `nixlAgent` shouldn't just proxy the backend interface. Unless the method uses `nixlAgent` invariants.

```C++
// nixl.h
class nixlAgent {
...
};

nixl_status_t
nixlQueryMem(const nixl_reg_dlist_t &descs,
             std::vector<nixlQueryRespArgs> *&resp,
             const nixl_opt_args_t *extra_params);
```

```C++
// nixl_agent.cpp

nixl_status_t
nixlQueryMem(const nixl_reg_dlist_t &descs,
             std::vector<nixlQueryRespArgs> *&resp,
             const nixl_opt_args_t *extra_params) {
    if (extra_params == nullptr) {
        NIXL_ERROR << "extra_params cannot be nullptr";
    }
    if (extra_params->backends.size() != 1) {
        NIXL_ERROR << "Invalid number of backends (" << backends.size() << "). There must be only one.";
    }

    return extra_params->backends[0]->engine->queryMem(descs, resp);
}
```

mkhazraee: The part that I'm missing is how we can get to the backend pointer? The BackendH class is:
 
```
class nixlBackendH {
    private:
        nixlBackendEngine* engine;
    ...
    friend class nixlAgentData;
    friend class nixlAgent;
};
```

So we had to make nixlAgent a friend to have access to the internal pointer. That new method should be friend of nixlBackendH.

src/api/cpp/nixl_types.h
mkhazraee: @brminich @yosefe what do you think? Reuse the same nixl_b_param_t or define a new one, or keep it nixl_blob_t for now?

src/core/nixl_agent.cpp
rakhmets: I see...
It doesn't seem right to me either... I mean two friends of `nixlBackendH` who already exist. After a little review of the code, it looks like `nixlBackendH` just needs to be a structure defined in nixl_agent.cpp. And it seems to be that `nixlAgentData` should not be a friend.
I'm not sure which side to start fixing.

src/api/cpp/nixl_types.h
mkhazraee: I asked Tim, the CodeStyle file is not updated, it's from the very initial commit and it should've been updated. Maybe we can update it alongside contribution updates, currently that seems more of a mistake as all Data types are snake style.

For the time being I'm going to stick with what we had and follow the API convention, later we can decide to change it across the API or not. At least updating the code style.

rakhmets: But the alias is redundant anyway. It seems that it is an artifact from C-style code:
```C
struct foo { ... ];
typedef struct foo foo_t;
```
Which is then updated changing `typedef` to `using`.
So I think it can be removed. We don't need an alias since in C++ you can just use `foo` without `struct`.

mkhazraee: Regarding code style I mentioned the mistake in the other PR. Regarding alias, I agree with you, that code style itself is part of the issue, struct being camelcase but type being snake case. Something to debate separately and then update things together.

rakhmets: Agree. This is a topic for a separate discussion. And agree again that it should be updated altogether with other names. However, I cannot recommend the better way to do it in this PR. So, please resolve this issue on your own. But at the same time, I would prefer to just delete the alias and use the class name.

rakhmets:
Please add a simple gtest to test the new API:
1. `extra_params=nullptr`, and check that the return value is `NIXL_ERR_INVALID_PARAM`.
2. pass non `nullptr` `extra_params`, and `extra_params->backends.size() != 1`,  and check that the return value is `NIXL_ERR_INVALID_PARAM`.
3. pass the correct  `extra_params`, and check that  the return value is `NIXL_ERR_NOT_SUPPORTED`.
Preferably in this PR, but you can also use a separate PR so as not to slow down the review.

src/core/nixl_agent.cpp
mkhazraee: The idea behind nixlBackendH was to separate user-side handle from the actual backend handle, so user doesn't have any access to the actual backend object. On top of that for nixl agent we wanted to do Pimpl style, so put all the members in nixlAgentData. Since we deallocate the backends in nixlAgentData destructor, that's why we made it a friend, iirc, but that might have changed or can be changed. It was two internal pimpl-related classes, so didn't think too much of it.

rakhmets: I see. Let's follow the existing path then, and update it along with the rest of the API changes.

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,21 @@ enum class nixl_cost_t {
...
+ * @struct nixlQueryArgs
+ * @brief A structure for querying memory results
+ */
+struct nixlQueryRespArgs {
+    bool accessible;
+    nixl_b_params_t info;
+};
+/**
+ * @brief A typedef for nixlQueryArgs
+ *        for querying memory results
```
rakhmets:
```suggestion
 * @brief An alias for nixlQueryResp for querying memory results
```

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,21 @@ enum class nixl_cost_t {
     ANALYTICAL_BACKEND = 0, // Analytical backend cost estimate
 };
 
+/**
+ * @struct nixlQueryArgs
```
rakhmets:
```suggestion
 * @struct nixlQueryResp
```

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,21 @@ enum class nixl_cost_t {
...
+ * @brief A structure for querying memory results
+ */
+struct nixlQueryRespArgs {
```
rakhmets:
```suggestion
struct nixlQueryResp {
```

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
...
                        const nixl_b_params_t &params,
                        nixlBackendH* &backend);
+        /**
+         * @brief  Query information about memory/storage with NIXL.
+         *         The backend should be specified via extra_params.
+         *
+         * @param  descs         Descriptor list of the buffers to be queried
+         * @param  extra_params  Additional parameters used in querying memory,
+         *                       such as indicating the backend
+         * @param  resp [out]    The output information for the queried descs
```
w1ldptr: Order of parameters in comment and in the actual function signature is different here.

src/api/cpp/backend/backend_engine.h
```diff
@@ -198,5 +184,28 @@ class nixlBackendEngine {
 
         // Force backend engine worker to progress.
         virtual int progress() { return 0; }
+
+
+        // *** Optional virtual methods that is good to be implemented by any backend *** //
+
+        // Query information about a list of memory/storage
+        virtual nixl_status_t
+        queryMem(const nixl_reg_dlist_t &descs, std::vector<nixl_query_resp_t> *&resp) const {
```
w1ldptr: Not a major concern, but note that other *Mem() functions receive individual nixlBlobDesc objects one-by-one instead of whole nixl_reg_dlist_t. Not sure whether it is worth it to complicate core for this though.

src/api/cpp/backend/backend_engine.h
```diff
@@ -198,5 +184,28 @@ class nixlBackendEngine {
 
         // Force backend engine worker to progress.
         virtual int progress() { return 0; }
+
+
+        // *** Optional virtual methods that is good to be implemented by any backend *** //
```
w1ldptr: Nit: Optional virtual methods that are good to implement in any backend.

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
...
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        queryMem(const nixl_reg_dlist_t &descs,
+                 std::vector<nixl_query_resp_t> *&resp,
```
w1ldptr: Why is this a pointer? The '*&' thing is used in NIXL agent API for assigning individual 'out' objects, but there are already APIs like getPluginParams that just take 'out' vectors by reference and assign them with new values. IMO making it just a regular reference is much cleaner since it is immediately obvious who creates and owns it (the caller) as opposed to *& (who allocates it? core? target backend?).

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +144,20 @@ enum class nixl_cost_t {
     ANALYTICAL_BACKEND = 0, // Analytical backend cost estimate
 };
 
+/**
+ * @struct nixlQuery
```
rakhmets:
```suggestion
 * @struct nixlQueryResp
```

src/api/cpp/backend/backend_engine.h
roiedanino: Why do we need a reference to a pointer? Wouldn't a reference be enough?

src/api/cpp/backend/backend_engine.h
mkhazraee: Keeping it to be implemented to be consistent with other ones.

src/api/cpp/backend/backend_engine.h
mkhazraee: Actually that is the main use case for this new API. In case of object store, we pass the whole list to backend, it calls its internal list method to the object store, and returns the results. If we split them to individual ones, then we would have atomicity issues. (Actually this is one of the reasons adding a new API was inevitable, between adding a new register which gets a desc_list and handling it with registerMem, this sounded more clean).

mkhazraee:
@roiedanino is going to add the bindings and @vvenkates27 is going to add an example backend implementation, marking it as Draft for now.

test/unit/utils/file/meson.build
```diff
@@ -0,0 +1,26 @@
...
+    'test_file_utils.cpp',
+    include_directories: [include_directories('../../../../src'), nixl_inc_dirs],
+    dependencies: [
+        file_utils_interface,
+        nixl_common_dep
+    ],
+    install: false
+)
+
+test('file_utils', test_file_utils)
```
iyastreb: minor: missing newline

test/unit/utils/file/test_file_utils.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include <iostream>
+#include <fstream>
+#include <cassert>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <filesystem>
+#include "utils/file/file_utils.h"
+
+int main() {
```
iyastreb: why not using gtest?

src/utils/file/file_utils.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <string>
+#include <vector>
+#include <sys/stat.h>
+#include "nixl_types.h"
+
+/**
+ * @brief File utilities for NIXL file backends
+ */
+
```
iyastreb, line 30: minor: maybe wrap them inside nixl namespace?

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,68 @@
...
+    resp.accessible = false;
+    resp.info.clear();
+
+    if (filename.empty()) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    // Check if file exists using stat
+    struct stat stat_buf;
+    bool file_exists = (stat(filename.c_str(), &stat_buf) == 0);
```
iyastreb: this variable seems redundant, can just use resp.accessible

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <errno.h>
+#include <cstring>
+#include <iostream>
+
+nixl_status_t queryFileInfo(const std::string& filename, nixl_query_resp_t& resp) {
+    // Initialize response
+    resp.accessible = false;
+    resp.info.clear();
+
+    if (filename.empty()) {
```
iyastreb, line 30: I'm not sure about this check
I mean, if filename is not empty, but file does not exist - then we still construct a dummy entry with resp.accessible=false
Why not to do the same with empty file then? It's no different than any invalid name IMO

In this case function can be simpler: 
```nixl_query_resp_t queryFileInfo(std::string_view filename)```

File: src/plugins/hf3fs/hf3fs_backend.h [hunk 23]
```diff
@@ -151,5 +151,8 @@ class nixlHf3fsEngine : public nixlBackendEngine {
 
         nixl_status_t checkXfer (nixlBackendReqH* handle) const;
         nixl_status_t releaseReqH(nixlBackendReqH* handle) const;
+
+        nixl_status_t queryMem(const nixl_reg_dlist_t &descs,
+                           std::vector<nixl_query_resp_t> &resp) const override;
```
iyastreb: minor: alignment

src/utils/file/file_utils.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @brief Query file information for a single file
+ * @param filename The filename to query (can be prefixed)
+ * @param resp Output response structure
+ * @return NIXL_SUCCESS on success, error code otherwise
+ */
+nixl_status_t queryFileInfo(const std::string& filename, nixl_query_resp_t& resp);
```
iyastreb: can filename be std::string_view?

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,68 @@
...
+        nixl_query_resp_t query_resp;
+        nixl_status_t status = queryFileInfo(filename, query_resp);
+
+        if (status != NIXL_SUCCESS) {
+            // If one file fails, mark it as inaccessible but continue
+            query_resp.accessible = false;
+            query_resp.info.clear();
+        }
+
+        resp.push_back(query_resp);
```
w1ldptr: This seems like a convoluted way to manually (re)implement `std::optional` here and in `queryFileInfo()`. Why not return `std::optional<nixl_b_params_t>` from `queryFileInfo()` and modify `resp` to be `std::vector<std::optional<nixl_b_params_t>>` instead? That would make all this loop just `resp.emaplace_back(queryFileInfo(filename, query_resp))` and ensure that `info` can only be accessed when it has been correctly initialized.

test/unit/utils/file/meson.build
vvenkates27: Done

test/unit/utils/file/test_file_utils.cpp
vvenkates27: Done

src/utils/file/file_utils.h
vvenkates27: Done

src/utils/file/file_utils.cpp
vvenkates27: Done

src/plugins/hf3fs/hf3fs_backend.h
vvenkates27: Done

src/utils/file/file_utils.h
vvenkates27: Done

src/api/cpp/nixl_types.h
mkhazraee: I'll close this conversation as there wasn't any objections.

src/utils/file/file_utils.cpp
vvenkates27: Changed it as proposed.

src/bindings/python/nixl_bindings.cpp
mkhazraee, line 1: Due to change in indentation, clang didn't accept to only enforce the new line and required the full file to be updated. For reviewers, only py::class_<nixl_query_resp_t> and queryMem were added.

examples/python/query_mem_example.py
mkhazraee, line 1: I'm not sure we need an example for a single API. Maybe it can be part of some larger example, like file example, and if we don't have it rename it to that.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -421,3 +426,15 @@ nixlGdsEngine::~nixlGdsEngine() {
         delete gds_utils;
     }
 }
+
+nixl_status_t
+nixlGdsEngine::queryMem(const nixl_reg_dlist_t &descs, std::vector<nixl_query_resp_t> &resp) const {
+    // Extract metadata from descriptors which are file names
+    // Different plugins might customize parsing of metaInfo to get the file names
+    std::vector<nixl_blob_t> metadata(descs.descCount());
```
mkhazraee, line 434: These 3 lines are repeated across backends, but they might want to parse their metaInfo differently in the future. Moving it to descriptor class removes 1 line of code, while for many descriptors it's not a useful method. And if we move it to file, our file implementation becomes desc dependent, so this sounded like a reasonable middle ground IMO.

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+nixl_status_t
+queryFileInfoList(const std::vector<std::string> &filenames, std::vector<nixl_query_resp_t> &resp) {
+    resp.clear();
+    resp.reserve(filenames.size());
+
+    for (const auto &filename : filenames) {
+        auto file_info = queryFileInfo(filename);
+        resp.emplace_back(nixl_query_resp_t{.accessible = file_info.has_value(),
+                                            .info = file_info.value_or(nixl_b_params_t{})});
```
w1ldptr: So why not change `std::vector<nixl_query_resp_t> &resp` to `std::vector<std::optional<nixl_b_params_t>> &resp` in order to ensure that users can't forget to check `accessible` before using `info`? In fact, the whole structure `nixl_query_resp_t` won't be needed anymore.

src/plugins/posix/posix_backend.cpp
```diff
@@ -311,3 +312,16 @@ nixl_status_t nixlPosixEngine::releaseReqH(nixlBackendReqH* handle) const {
...
+nixl_status_t
+nixlPosixEngine::queryMem(const nixl_reg_dlist_t &descs,
+                          std::vector<nixl_query_resp_t> &resp) const {
+    // Extract metadata from descriptors which are file names
+    // Different plugins might customize parsing of metaInfo to get the file names
+    std::vector<nixl_blob_t> metadata(descs.descCount());
+    for (int i = 0; i < descs.descCount(); ++i)
+        metadata[i] = descs[i].metaInfo;
+
+    // Use file utils to query the files directly with metadata
```
w1ldptr: Please clean up redundant self-describing comments here and in other places.

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+        return std::nullopt;
+    }
+
+    // Check if file exists using stat
+    struct stat stat_buf;
+    if (stat(std::string(filename).c_str(), &stat_buf) != 0) {
+        return std::nullopt;
+    }
+
+    // File exists, return file information
```
w1ldptr: Redundant comments.

examples/python/query_mem_example.py
mkhazraee: For now letting it be, as we have partial md too. Maybe a future todo to merge them.

src/utils/file/file_utils.cpp
vvenkates27: This appears to be an invasive change, which would need to change the API definition also. Not sure we are gaining much from that. We already use optional in the implementation, not sure how that maps to bindings also.

mkhazraee: The initial goal was a vector of booleans, so the main output of this method is actually the boolean values, the nixl_b_params_t is the optional aspect in case in future we want to add some metadata. So the boolean might be true but the nixl_b_params_t be empty, and if we use std::optional for that while there is no value doesn't seem a good fit. Internally within the methods, we can use std::optional, as Vish already did per your comment.

w1ldptr: >So the boolean might be true but the nixl_b_params_t be empty, and if we use std::optional for that while there is no value doesn't seem a good fit.

I'm not following why it is not a good fit: you can still have optional to be nullopt (same as boolean==false) or optional(info) (with empty or non-empty info, same as boolean==true). All it does is preventing the user from accessing `info` field directly without checking whether the value is accessible or not, i.e. improves maintainability and readability.

mkhazraee: Because users want true and false, not nullopt.

src/utils/file/file_utils.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+nixl_status_t
+queryFileInfoList(const std::vector<std::string> &filenames, std::vector<nixl_query_resp_t> &resp) {
+    resp.clear();
+    resp.reserve(filenames.size());
+
+    for (const auto &filename : filenames) {
+        auto file_info = queryFileInfo(filename);
+        resp.emplace_back(file_info);
```
w1ldptr: I think emplace here just calls copy ctor because file_info here is a lvalue. Convert it to rvalue by using `std::move()`, or, even better, just directly emplace the result of `queryFileInfo(filename)` call instead assigning it to a variable first.

test/gtest/unit/query_mem.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    nixl_status_t status = nixl::queryFileInfoList(filenames, resp);
+    EXPECT_EQ(status, NIXL_SUCCESS);
+    EXPECT_EQ(resp.size(), 4);
+    EXPECT_TRUE(resp[0].has_value()); // test_file1 exists
+    EXPECT_FALSE(resp[1].has_value()); // non_existent_file doesn't exist
+    EXPECT_TRUE(resp[2].has_value()); // test_file2 exists
+    EXPECT_FALSE(resp[3].has_value()); // empty filename doesn't exist
+}
+
```
w1ldptr: These are great! But all the following ones don't look like _unit_ tests to me since they use whole NIXL stack. Consider moving them from unit to functional tests in test/gtest.

src/utils/file/file_utils.cpp
mkhazraee: We eventually went for std::optional, as it made the python bindings much cleaner, and with has_value() we achieved the true/false requirement, and value() would give the info.

src/utils/file/file_utils.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace nixl {
+
+/**
+ * @brief Query file information for a single file
+ * @param filename The filename to query (can be prefixed)
+ * @return std::optional<nixl_b_params_t> containing file info if accessible, std::nullopt otherwise
+ */
+std::optional<nixl_b_params_t>
```
w1ldptr: Nit: return nixl_query_resp_t here for consistency. (not mandatory, doesn't change functionality)

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
         createBackend (const nixl_backend_t &type,
                        const nixl_b_params_t &params,
                        nixlBackendH* &backend);
+        /**
```
yosefe: space line before

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,20 @@ class nixlAgent {
...
+         * @param  descs         Descriptor list of the buffers to be queried
+         * @param  resp [out]    The output information for the queried descs
+         * @param  extra_params  Additional parameters used in querying memory,
+         *                       such as indicating the backend
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        queryMem(const nixl_reg_dlist_t &descs,
+                 std::vector<nixl_query_resp_t> &resp,
+                 const nixl_opt_args_t *extra_params) const;
```
yosefe: space line after

src/api/python/_api.py
```diff
@@ -951,3 +951,30 @@ def get_serialized_descs(self, descs) -> bytes:
...
+    @param is_sorted Optional bool for a list of memory regions or tensors for if they are sorted.
+    @return List of query results where each item is either None if not found, or a dictionary with the info
+    """
+
+    def query_memory(
+        self,
+        reg_list,
+        backend: str,
+        mem_type: Optional[str] = None,
+        is_sorted: bool = False,
```
yosefe: why need is_sorted?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -30,100 +30,100 @@ namespace py = pybind11;
...
-            break;
-        case NIXL_ERR_UNKNOWN:
-            throw nixlUnknownError(nixlEnumStrings::statusStr(status).c_str());
-            break;
-        case NIXL_ERR_NOT_SUPPORTED:
-            throw nixlNotSupportedError(nixlEnumStrings::statusStr(status).c_str());
-            break;
-        default:
-            throw std::runtime_error("BAD_STATUS");
+    case NIXL_IN_PROG:
```
yosefe, line 80: much if this PR seems unneeded code formatting, can you pls revert it?

src/core/nixl_agent.cpp
```diff
@@ -325,6 +325,18 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
+nixl_status_t
+nixlAgent::queryMem(const nixl_reg_dlist_t &descs,
+                    std::vector<nixl_query_resp_t> &resp,
+                    const nixl_opt_args_t *extra_params) const {
+
+    if (!extra_params || extra_params->backends.size() != 1) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    return extra_params->backends[0]->engine->queryMem(descs, resp);
```
yosefe, line 337: can we go over all backends and query each of them? skip those that return NOT_IMPLEMENTED, and each backend will add more keys to the nixl_b_params_t map

src/api/cpp/nixl_types.h
```diff
@@ -144,6 +145,14 @@ enum class nixl_cost_t {
     ANALYTICAL_BACKEND = 0, // Analytical backend cost estimate
 };
 
+/**
+ * @brief A typedef for std::optional<nixl_b_params_t> for querying memory results
+ *        Validity of a nixl_query_resp_t can be checked by has_value() method,
+ *        and if true, the dictionary can be accessed by value() method.
+ */
+using nixl_query_resp_t = std::optional<nixl_b_params_t>;
```
yosefe, line 153: why does it have to be optional?
nixl_b_params_t is already a map of key->value so how nullopt is different from an empty map?

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -421,3 +426,14 @@ nixlGdsEngine::~nixlGdsEngine() {
...
+    for (int i = 0; i < descs.descCount(); ++i)
+        metadata[i] = descs[i].metaInfo;
+
+    return nixl::queryFileInfoList(metadata, resp);
```
yosefe, line 438: maybe add helper function that takes nixl_reg_dlist_t and returns file info response? seems it's used in several plugins

test/gtest/query_mem.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "nixl.h"
+#include "nixl_descriptors.h"
+
+class QueryMemTest : public ::testing::Test {
+protected:
+    void
+    SetUp() override {
+        // Create temporary test files
+        test_file1 = "/tmp/test_query_mem_1.txt";
+        test_file2 = "/tmp/test_query_mem_2.txt";
```
yosefe: create files in local directory to not pollute the test env and allow parallel runs

test/gtest/query_mem.cpp
```diff
@@ -0,0 +1,202 @@
...
+    nixl_opt_args_t extra_params;
+    extra_params.backends = {backend_handle};
+
+    std::vector<nixl_query_resp_t> resp;
+    status = agent.queryMem(descs, resp, &extra_params);
+    EXPECT_EQ(status, NIXL_SUCCESS);
+    EXPECT_EQ(resp.size(), 2);
+    EXPECT_TRUE(resp[0].has_value());
+    EXPECT_TRUE(resp[1].has_value());
+    EXPECT_TRUE(resp[0].value().find("size") != resp[0].value().end());
```
yosefe, line 94: maybe check the file size equal to given value?

test/gtest/unit/meson.build
```diff
@@ -30,12 +31,14 @@ unit_test_deps += [agent_unit_test_dep]
...
     sources : [
         'main.cpp',
+        'query_file.cpp',
     ],
     cpp_args : cpp_args,
     dependencies : unit_test_deps,
     include_directories: [
         nixl_inc_dirs,
         utils_inc_dirs,
+        include_directories('../../../src'),
```
yosefe: why needed?

test/gtest/unit/query_file.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "nixl_descriptors.h"
+
+class QueryFileTest : public ::testing::Test {
+protected:
+    void
+    SetUp() override {
+        // Create temporary test files
+        test_file1 = "/tmp/test_query_mem_1.txt";
+        test_file2 = "/tmp/test_query_mem_2.txt";
+        non_existent_file = "/tmp/non_existent_file.txt";
```
yosefe: this seems like duplication of agent-level test, IMO it's redundant, and duplicates the issue of creating files in system-wide directory

test/python/test_nixl_bindings.py
```diff
@@ -156,3 +158,89 @@ def test_agent():
 
     nixl_utils.free_passthru(addr1)
     nixl_utils.free_passthru(addr2)
+
+
+def test_query_mem():
+    """Test basic queryMem functionality"""
+    # Create temporary test files
+    temp_file1 = tempfile.NamedTemporaryFile(delete=False)
```
yosefe: create files in local directory

test/unit/meson.build
```diff
@@ -13,5 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Define gtest_dep for unit tests
+gtest_dep = dependency('gtest', version : '>=1.11.0', required : false)
```
yosefe: why needed here?

test/unit/utils/file/meson.build
```diff
@@ -0,0 +1,27 @@
...
+    include_directories: [include_directories('../../../../src'), nixl_inc_dirs],
+    dependencies: [
+        file_utils_interface,
+        nixl_common_dep,
+        gtest_dep
+    ],
+    install: false
+)
+
+test('file_utils', test_file_utils)
```
yosefe: missing newline

test/unit/utils/file/meson.build
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+test_file_utils = executable('test_file_utils',
+    'test_file_utils.cpp',
+    include_directories: [include_directories('../../../../src'), nixl_inc_dirs],
```
yosefe: can we use some existing variable instead of this weird path? why needed?

test/unit/utils/file/test_file_utils.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <iostream>
+#include <fstream>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <filesystem>
+#include <optional>
+#include <gtest/gtest.h>
+#include "utils/file/file_utils.h"
+
+class FileUtilsTest : public ::testing::Test {
```
yosefe: why need another test that we already have in gtest?

src/api/cpp/nixl_types.h
w1ldptr: For example, Obj will return empty map on success since it doesn't provide any metadata unlike file-based plugins.

src/bindings/python/nixl_bindings.cpp
mkhazraee: I tried a lot to revert it, clang doesn't accept. Since the spacing at the beginning for the new method is not the same! :(

src/api/python/_api.py
mkhazraee: Just kept it to be consistent with others, when we remove the issorted to remove it from all.

src/bindings/python/nixl_bindings.cpp
mkhazraee: Retried now in case without the new python object it goes through, nope!

test/unit/utils/file/meson.build
mkhazraee: As in make it a two liner?

test/unit/utils/file/test_file_utils.cpp
mkhazraee: Just kept the one in gtest.

test/unit/utils/file/meson.build
mkhazraee: Removed the test anyways.

test/unit/meson.build
mkhazraee: Removed.

src/api/python/_api.py
mkhazraee: Removed it.

src/core/nixl_agent.cpp
mkhazraee: For another PR if need rises, very unlikely.

src/api/cpp/nixl_types.h
mkhazraee: Yes, the main output is just the true and false of key/file exists. The dictionary is cherry on top, if any extra information needs to be provided.

src/plugins/cuda_gds/gds_backend.cpp
mkhazraee: We were debating on this. Initially we had metaInfo extractor in desc_list, but that didn't apply to other use cases, and also different backends might want to parse the metaInfo differently, have their special flags and so in it. The other option was to fully move it in the file processing code, but that meant file utils need to know reg_dlist, seemed more modular just to get a list of strings.

If we really want it, we can add a new method in file utils, to do this extraction, but it would save 1 line which later file backends might customize later.

test/gtest/unit/query_file.cpp
mkhazraee: I've fixed the /tmp part across all the tests. I am with you, just this test doesn't have notion of reg_dlist. I think when we decide on adding a new method for that or making the file method to directly get reg_dlist, we can finalize to keep this test or remove it.

w1ldptr (dismissed):

test/gtest/query_mem.cpp
mkhazraee: Ugh missed this, will do in a follow up PR.

mkhazraee: Actually added this. And removed the other test. So all addressed.

vvenkates27 (approved):

aranadive (approved):

test/gtest/query_mem.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include "nixl_descriptors.h"
+
+class QueryMemTest : public ::testing::Test {
+protected:
+    void
+    SetUp() override {
+        // Create temporary test files
+        dir_path = "./files_for_query";
+        if (!(std::filesystem::exists(dir_path) && std::filesystem::is_directory(dir_path)))
+            std::filesystem::create_directory(dir_path);
```
yosefe, line 33: where is removed? also in other tests
and IMO better call the new dir ".test_files"

src/bindings/python/nixl_bindings.cpp
yosefe: next time in such case pls push a preliminary PR to reformat the file

src/plugins/cuda_gds/gds_backend.cpp
yosefe: maybe it makes sense file plugins could inherit from common parent? in addition to the base plugin class. is seems that common case that several plugins have common code.

PR #524, opened 2025-07-02
Extend code style with current practices

w1ldptr:
Make it reflect what we have been suggesting during recent code reviews.

## What?
Make code style document reflect what we have been suggesting during recent code reviews.

## Why?
In order to make the rules explicit.

PR #532, opened 2025-07-03
Add xfer time telemetry.

mkhazraee:
## What?
Add basic telemetry per transfer request.

## Why?
For debugging purposes, it is beneficial to have telemetry per transfer to know if transfers are taking long or some other issue in the system.

## How?
Added 3 fields to the transfer, start and end time, as well as total transfer size.


mkhazraee:
Maybe we should another levels for debug flag, or a separate sys env variable to output telemetry data or not. Later on we can consider having a telemetry output file given in nixl_agent_config, but for now let's start just with debug log as it's the main use case now.

mkhazraee:
This is adding to the debug log, so I'm thinking a specific test can't be written for it, it needs to be tested on a real system and verify its behavior.

src/core/nixl_agent.cpp
```diff
@@ -860,6 +872,18 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
                                       &opt_args);
     req_hndl->status = ret;
+
+    if (req_hndl->status == NIXL_SUCCESS) {
+        req_hndl->endTime = std::chrono::high_resolution_clock::now();
+        std::chrono::duration<double, std::micro> xfer_time =
+            req_hndl->endTime - req_hndl->startTime;
+        NIXL_DEBUG << "Xfer with " << req_hndl->initiatorDescs->descCount()
+                   << " descs of total size: " << req_hndl->totalBytes << " took "
+                   << xfer_time.count() << "us";
```
brminich: could be some common func to avoid code duplication

src/core/nixl_agent.cpp
```diff
@@ -860,6 +872,18 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
                                       req_hndl->backendHandle,
                                       &opt_args);
     req_hndl->status = ret;
+
+    if (req_hndl->status == NIXL_SUCCESS) {
```
brminich: seems some of the users are also interested in just posting times

brminich (commented):
in general maybe compile this code out for the release build?

src/core/nixl_agent.cpp
```diff
@@ -747,10 +755,11 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
     handle->remoteAgent = remote_agent;
-    handle->backendOp   = operation;
-    handle->status      = NIXL_ERR_NOT_POSTED;
-    handle->notifMsg    = opt_args.notifMsg;
-    handle->hasNotif    = opt_args.hasNotif;
+    handle->backendOp = operation;
+    handle->status = NIXL_ERR_NOT_POSTED;
+    handle->notifMsg = opt_args.notifMsg;
+    handle->hasNotif = opt_args.hasNotif;
+    handle->totalBytes = totalBytes;
```
iyastreb: minor: I think we need to init startTime and endTime with zeroes here

mkhazraee:
> in general maybe compile this code out for the release build?

@brminich We can. The thinking was if we want to provide some telemetry in general, this will potentially be the minimum, as each xfer handle is already a list. But for the time being if the debug logs is the only use case, we can remove even the members in the req struct.

src/core/nixl_agent.cpp
```diff
@@ -860,6 +872,18 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
+    if (req_hndl->status == NIXL_SUCCESS) {
+        req_hndl->endTime = std::chrono::high_resolution_clock::now();
+        std::chrono::duration<double, std::micro> xfer_time =
+            req_hndl->endTime - req_hndl->startTime;
+        NIXL_DEBUG << "Xfer with " << req_hndl->initiatorDescs->descCount()
+                   << " descs of total size: " << req_hndl->totalBytes << " took "
+                   << xfer_time.count() << "us";
+    }
+
+    // If post time is important for debugging, we can add a DEBUG print here by getting
```
mkhazraee: @brminich This is why I put this comment, we can add this if we want post time.

src/core/nixl_agent.cpp
mkhazraee: Pointed it out in a  conversation a few lines bellow.

src/core/nixl_agent.cpp
mkhazraee: Hmm, not sure. If you can think of a use case outside post/check let me know. But in post it is set at the beginning, and end shouldn't be used before it is set again in these two. So we're in control. If we have a rule for always initiating the structs, def we can add it to the struct itself, don't recall such a rule though.

iyastreb: Sure, not a big deal.
I was just thinking about use cases when:
- request is created, but not posted yet
- request is posted but not yet completed
- request failed
I thought it's better to have a meaningful state rather than random, just in case. But up to you

brminich:
> > in general maybe compile this code out for the release build?
> 
> @brminich We can. The thinking was if we want to provide some telemetry in general, this will potentially be the minimum, as >each xfer handle is already a list. But for the time being if the debug logs is the only use case, we can remove even the members >in the req struct.

Yes, currently this info is only used by `NIXL_DEBUG`, which is present in debug build only, so it is worth to remove the rest from release build

src/core/nixl_agent.cpp
mkhazraee: After the changes we talked today, the prints got slightly different, and there is single line in common now, so not sure worth defining a method for the print anymore.

src/core/nixl_agent.cpp
mkhazraee: Added the other log anyways as our debugging would use it.

src/core/nixl_agent.cpp
mkhazraee: Set them to chrono min.

mkhazraee:
> > > in general maybe compile this code out for the release build?
> > 
> > 
> > @brminich We can. The thinking was if we want to provide some telemetry in general, this will potentially be the minimum, as >each xfer handle is already a list. But for the time being if the debug logs is the only use case, we can remove even the members >in the req struct.
> 
> Yes, currently this info is only used by `NIXL_DEBUG`, which is present in debug build only, so it is worth to remove the rest from release build

Following our conversation this morning, kept it but added the telemetry flag.

src/core/transfer_request.h
```diff
@@ -17,6 +17,8 @@
 #ifndef __TRANSFER_REQUEST_H_
 #define __TRANSFER_REQUEST_H_
 
+#define MIN_CHRONO_TIME std::chrono::time_point<std::chrono::high_resolution_clock>::min();
```
rakhmets:
```suggestion
constexpr auto min_chrono_time = std::chrono::time_point<std::chrono::high_resolution_clock>::min();
```

src/core/transfer_request.h
```diff
@@ -34,6 +36,11 @@ class nixlXferReqH {
         nixl_xfer_op_t     backendOp;
         nixl_status_t      status;
 
+        // Telemetry info, setting times to 0 in case checkXfer is called before postXfer
+        std::chrono::high_resolution_clock::time_point startTime = MIN_CHRONO_TIME;
+        std::chrono::high_resolution_clock::time_point endTime = MIN_CHRONO_TIME;
```
rakhmets: It seems that we don't need `endTime`. It is assigned to `std::chrono::high_resolution_clock::now()` in both places, and never reused.

src/core/nixl_agent.cpp
```diff
@@ -860,6 +879,24 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
                                       req_hndl->backendHandle,
                                       &opt_args);
     req_hndl->status = ret;
+
+    if (data->telemetryEn) {
+        std::chrono::high_resolution_clock::time_point currTime =
```
rakhmets: auto?

src/core/nixl_agent.cpp
```diff
@@ -860,6 +879,24 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
+            std::chrono::high_resolution_clock::now();
+        std::chrono::duration<double, std::micro> xfer_time = currTime - req_hndl->startTime;
```
rakhmets: std::chrono::microseconds?

src/core/agent_data.h
```diff
@@ -97,6 +97,9 @@ class nixlAgentData {
         void enqueueCommWork(nixl_comm_req_t request);
         void getCommWork(std::vector<nixl_comm_req_t> &req_list);
 
+        // Flag for telemetry
+        bool telemetryEn;
```
rakhmets: I think there's no need to shorten the name. `telemetryEnabled`.
And the comment can be also removed.

src/core/nixl_agent.cpp
```diff
@@ -860,6 +879,24 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
+
+        if (req_hndl->status == NIXL_SUCCESS) {
+            req_hndl->endTime = currTime;
+            NIXL_DEBUG << "Xfer with " << req_hndl->initiatorDescs->descCount()
+                       << " descs of total size: " << req_hndl->totalBytes
+                       << " finished in post and took " << xfer_time.count() << "us to finish";
+        } else {
+            NIXL_DEBUG << "Posting of Xfer with " << req_hndl->initiatorDescs->descCount()
+                       << " descs of total size: " << req_hndl->totalBytes << " took "
+                       << xfer_time.count() << "us";
```
rakhmets: I think we should print this only if `status == NIXL_IN_PROG`.

src/core/nixl_agent.cpp
```diff
@@ -878,6 +915,17 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
     }
 
+    if (data->telemetryEn) {
+        if (req_hndl->status == NIXL_SUCCESS) {
+            req_hndl->endTime = std::chrono::high_resolution_clock::now();
+            std::chrono::duration<double, std::micro> xfer_time =
+                req_hndl->endTime - req_hndl->startTime;
+            NIXL_DEBUG << "Xfer with " << req_hndl->initiatorDescs->descCount()
+                       << " descs of total size: " << req_hndl->totalBytes << " took "
+                       << xfer_time.count() << "us to finish";
```
rakhmets: The output can be unified, e.g. `Completing Xfer ... took ... us`, `Posting Xfer ... took ... us`. And moved to a helper function parameterizing print of `Completing`, and `Posting`.

src/core/transfer_request.h
mkhazraee: I was thinking about this too. In future if we add a full telemetry implementation, outputting to files or so, that becomes handy. As of now in the debug prints, not useful. Eventually thought to keep it, so technically when a transfer is done, we can go and check those values later, so if a separate thread of method is doing telemetry they're decoupled. If you think for full telemetry we will use it right away too, maybe we can remove.

src/core/nixl_agent.cpp
mkhazraee: Throws off compile error, but as the other suggestion went with auto.

src/core/transfer_request.h
rakhmets: I think It is better to add this to the list of class members only when it is really needed.

src/core/nixl_agent.cpp
mkhazraee: Reworded the print and added a helper method. Not sure same file is fine or we want it in a separate file. Not quite general to put it in debug specific utils I think.

src/core/nixl_agent.cpp
mkhazraee: As Raul suggested reworded things to be able to make it a helper method.

src/core/transfer_request.h
mkhazraee: I guess a middle ground would be comment them out, if they become useful in future.

src/core/nixl_agent.cpp
```diff
@@ -860,6 +888,25 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
                                       req_hndl->backendHandle,
                                       &opt_args);
     req_hndl->status = ret;
+
+    if (data->telemetryEnabled) {
+        auto currTime = std::chrono::high_resolution_clock::now();
+        auto xfer_time = currTime - req_hndl->startTime;
```
rakhmets: I'm afraid it will not produce `xfer_time` in microseconds.
```suggestion
        const auto xfer_time = std::chrono::duration_cast<std::chrono::microseconds>(currTime - req_hndl->startTime);
```
The same cast should be done in `getXferStatus`

src/core/nixl_agent.cpp
rakhmets: The same file is fine. And any “helper” class or function that is defined in an implementation source file should be at an unnamed namespace scope.

```C++
namespace {
    inline void
    telemetryPrint(const std::string msg_type,
                   const uint64_t descs,
                   const uint64_t bytes,
                   const uint64_t elapsed) {
        NIXL_DEBUG << msg_type << " Xfer with " << descs << " descs of total size: " << bytes << " in "
                   << elapsed << "us.";
    }
}
```

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed2

src/core/transfer_request.h
rakhmets: I would leave the addition of the class field to someone who will expand the functionality. It is difficult to predict whether this particular field will be needed and with exactly this name, and this type, or whether it will be a new structure, a class.

src/core/nixl_agent.cpp
mkhazraee: I returned it to my original implementation, between auto+cast or just defining it properly, I would go for the latter. And if you meant std::chrono::duration<std::chrono::microseconds>, that causes error too.

src/core/nixl_agent.cpp
mkhazraee: With endTime removed for now, I could simplify the code more, just didn't want to make the helper method to be a friend of nixlXferReqH, so getting the values outside the helper function. If you have a better suggestion please let me know.

src/core/nixl_agent.cpp
rakhmets: Please note that:
```C++
const std::chrono::duration<double, std::micro> xfer_time =
        std::chrono::high_resolution_clock::now() - start_time;
std::cout << xfer_time.count() << std::endl;
```
prints a floating point value. And:
```C++
const auto xfer_time = std::chrono::duration_cast<std::chrono::microseconds>
        (std::chrono::high_resolution_clock::now() - start_time);
std::cout << xfer_time.count() << std::endl
```
prints an integer value.

src/core/transfer_request.h
```diff
@@ -17,6 +17,9 @@
 #ifndef __TRANSFER_REQUEST_H_
 #define __TRANSFER_REQUEST_H_
 
+constexpr auto min_chrono_time = std::chrono::time_point<std::chrono::high_resolution_clock>::min();
+using chrono_point_t = std::chrono::high_resolution_clock::time_point;
```
rakhmets, line 21: We had the conversation in another PR. https://github.com/ai-dynamo/nixl/pull/531#discussion_r2184949313.

And docs/CodeStyle.md has just been updated by @w1ldptr here https://github.com/ai-dynamo/nixl/pull/524.

It seems that we should use lower camel case in this case:
```suggestion
using chronoPoint = std::chrono::high_resolution_clock::time_point;
```

mkhazraee: I double checked with Tim and Vlad, it was carry over from before, but it's not correct, all our types end in _t. A mistake stayed around from the beginning, will open a separate PR to update the code style, later we can decide if we want to change the code style, but for now it was just carrying it over.

rakhmets: I'm not sure that adding `_t` to all our types is a good addition to our naming conventions. This means that every class name should follow this logic... It seems to be an excessive overhead.

mkhazraee: It's not adding, it is the current API (nixl_xfer_dlist_t, nixl_reg_dlist_t, etc). If we want to change that convention or not is a separate PR and discussion.

src/core/nixl_agent.cpp
```diff
@@ -81,6 +82,9 @@ nixlAgentData::nixlAgentData(const std::string &name,
         throw std::invalid_argument("Agent needs a name");
 
     memorySection = new nixlLocalSection();
+
+    const char *telemetry = std::getenv("NIXL_EN_TELEMETRY");
```
brminich: minor: names seems a bit strange
i'd call it something like `NIXL_TELEMETRY_ENABLE` or similar

brminich (approved):

brminich:
also maybe always print telemetry data when new env var is set?
Otherwise user will need to set two env vars (+ NIXL_LOG_LEVEL) and will get some extra output from NIXL_DEBUG logs

src/core/transfer_request.h
rakhmets: Sorry I didn't mean you added a new code style. I meant that in general it is not really required to add `_t` postfix to the type name. Anyway it's not actually related to this PR. This is a topic for a separate discussion.

src/core/nixl_agent.cpp
```diff
@@ -81,6 +83,15 @@ nixlAgentData::nixlAgentData(const std::string &name,
...
 
     memorySection = new nixlLocalSection();
+
+    const char *telemetry = std::getenv("NIXL_TELEMETRY_ENABLE");
+    if (telemetry != nullptr) {
+        if ((!strcmp(telemetry, "y")) || (!strcmp(telemetry, "Y")))
+            telemetryEnabled = true;
+        else
+            NIXL_WARN
+                << "Invalid NIXL_TELEMETRY_ENABLE environment variable, not enabling telemetry.";
```
brminich, line 108: maybe just info? Otherwise better to accept `n` `N` as well

robertgshaw2-redhat:
Could you add an option to expose this back to python optionally? I would like to log this information to prometheus

src/core/nixl_agent.cpp
```diff
@@ -81,6 +83,17 @@ nixlAgentData::nixlAgentData(const std::string &name,
         throw std::invalid_argument("Agent needs a name");
 
     memorySection = new nixlLocalSection();
+
+    const char *telemetry = std::getenv("NIXL_TELEMETRY_ENABLE");
+    if (telemetry != nullptr) {
+        if ((!strcmp(telemetry, "y")) || (!strcmp(telemetry, "Y")))
```
yosefe: strcasecmp

src/core/nixl_agent.cpp
```diff
@@ -556,6 +569,7 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
 
     nixl_meta_dlist_t* local_descs  = local_side->descs.at(backend);
     nixl_meta_dlist_t* remote_descs = remote_side->descs.at(backend);
+    uint64_t totalBytes = 0;
```
yosefe: size_t

src/core/nixl_agent.cpp
```diff
@@ -571,6 +585,8 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
         if ((*local_descs )[local_indices [i]].len !=
             (*remote_descs)[remote_indices[i]].len)
             return NIXL_ERR_INVALID_PARAM;
+        else
```
yosefe: else not needed here

src/core/nixl_agent.cpp
```diff
@@ -681,11 +698,14 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     }
 
     // Check the correspondence between descriptor lists
+    uint64_t totalBytes = 0;
```
yosefe: size_t

src/core/nixl_agent.cpp
```diff
@@ -681,11 +698,14 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
     if (local_descs.descCount() != remote_descs.descCount())
         return NIXL_ERR_INVALID_PARAM;
     for (int i=0; i<local_descs.descCount(); ++i)
         if (local_descs[i].len != remote_descs[i].len)
             return NIXL_ERR_INVALID_PARAM;
+        else
```
yosefe: else not needed but probably better add { } to the for loop

src/core/nixl_agent.cpp
```diff
@@ -818,6 +839,24 @@ nixlAgent::estimateXferCost(const nixlXferReqH *req_hndl,
                                               extra_params);
 }
 
+namespace {
+inline void
+telemetryPrint(const std::string msg_type,
+               const uint64_t descs,
+               const uint64_t bytes,
+               const chrono_point_t start_time) {
```
yosefe: const std::string& msg_type,
size_t num_descs,
size_t total_bytes,
const chrono_point_t &start_time

tstamler (dismissed):

src/core/nixl_agent.cpp
```diff
@@ -878,6 +922,21 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
+                           req_hndl->initiatorDescs->descCount(),
+                           req_hndl->totalBytes,
+                           req_hndl->startTime);
+        else if (req_hndl->status == NIXL_IN_PROG)
+            telemetryPrint("Posted",
+                           req_hndl->initiatorDescs->descCount(),
+                           req_hndl->totalBytes,
+                           req_hndl->startTime);
+        // Errors should show up in debug log separately, not adding a print here
+    }
```
yosefe: can we do smth like this, to keep the data path cleaner:
if (data->telemetryEnabled) {
telemetryUpdate(req_hndl, "post");
}

src/core/transfer_request.h
```diff
@@ -34,6 +37,11 @@ class nixlXferReqH {
         nixl_xfer_op_t     backendOp;
         nixl_status_t      status;
 
+        // Telemetry info, setting startTime to a default min value in case checkXfer is called
+        // before postXfer. If useful, endTime can be added as well.
+        chrono_point_t startTime = min_chrono_time;
```
yosefe: IMO no need to initialize it

src/core/transfer_request.h
```diff
@@ -34,6 +37,11 @@ class nixlXferReqH {
...
+        uint64_t totalBytes;
```
yosefe: size_t

src/core/nixl_agent.cpp
mkhazraee: I actually was going for that first, but realized then I should make this method a friend of nixlXferReqH (not to add the helper method to nixlAgent header file). I was trying to avoid that, but I agree this is cleaner and not a big deal if we make it a friend. Or let me know if you can think of another approach.

src/core/transfer_request.h
mkhazraee: iirc @rakhmets  mentioned it, in case we want to get that value before a xfer is posted. But I think since we are currently the only users of it, we can keep it uninitialized, if a need came up, then we add the initialization.

src/core/nixl_agent.cpp
yosefe: make it a friend. IMO it's a problem caused by "incomplete" implementation of pImpl idiom in nixlAgent.

mkhazraee: I did, just it's not in any namespace now.

aranadive (dismissed):

aranadive (approved):

PR #537, opened 2025-07-07
UCX/BACKEND: refactor nixlUcxEngine::nixlUcxEngine

evgeny-leksikov:
## What?
Refactoring of `nixlUcxEngine::nixlUcxEngine`

~~NOTE: in dependency from https://github.com/ai-dynamo/nixl/pull/499~~

## Why?
to resolve [this](https://github.com/ai-dynamo/nixl/pull/499/files#r2183166510) conversation

## How?
del usage of `initErr` and replaced with exceptions

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -618,9 +618,6 @@ nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
 
 // Through parent destructor the unregister will be called.
 nixlUcxEngine::~nixlUcxEngine() {
-    // initErr is not used in this class and should be removed
-    NIXL_ASSERT(!this->initErr);
-
     progressThreadStop();
     if (pthrOn) {
         for (const auto pthr_control_pipe : pthrControlPipe) {
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -528,7 +528,7 @@ void nixlUcxEngine::progressThreadRestart()
 nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
     : nixlBackendEngine(init_params),
       pthrControlPipe{0, 0} {
-    unsigned long numWorkers;
+    size_t numWorkers;
     std::vector<std::string> devs; /* Empty vector */
     nixl_b_params_t* custom_params = init_params->customParams;
 
@@ -576,14 +576,11 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams *init_params)
                                           numWorkers,
                                           init_params->syncMode);
 
-    for (unsigned int i = 0; i < numWorkers; i++) {
-        auto uw = std::make_unique<nixlUcxWorker>(*uc, err_handling_mode);
-        auto addr = uw->epAddr();
-        uws.push_back(std::move(uw));
-        if (i == 0) {
-            workerAddr = addr;
-        }
+    for (size_t i = 0; i < numWorkers; i++) {
+        uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode));
     }
+
+    workerAddr = uws.front()->epAddr();
 
     if (pthrOn) {
         for (auto &uw: uws) {
@@ -626,12 +623,10 @@ nixlUcxEngine::~nixlUcxEngine() {
 
     progressThreadStop();
     if (pthrOn) {
-        if (pthrControlPipe[0] != 0) {
-            close(pthrControlPipe[0]);
-        }
-
-        if (pthrControlPipe[1] != 0) {
-            close(pthrControlPipe[1]);
+        for (const auto pthr_control_pipe : pthrControlPipe) {
+            if (pthr_control_pipe != 0) {
+                close(pthr_control_pipe);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -584,16 +576,13 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
-    const auto &uw = uws.front();
-    workerAddr = uw->epAddr();
-
-    if (workerAddr.empty()) {
-        NIXL_ERROR << "Failed to get UCX worker address";
-        initErr = true;
-        return;
+    for (unsigned int i = 0; i < numWorkers; i++) {
+        auto uw = std::make_unique<nixlUcxWorker>(*uc, err_handling_mode);
+        auto addr = uw->epAddr();
```
rakhmets: Do you think it would be better to initialize the worker address during the construction of the worker?
In this case, after the creation, we will get a valid `nixlUcxWorker`. And `nixlUcxWorker::epAddr` will not throw an exception, but just return the address (maybe rename to `getAddr` then).
And then you can simplify the code here, e.g.:
```C++
for (unsigned int i = 0; i < numWorkers; i++)
    uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode));

workerAddr = uws[0]->getAddr();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -628,20 +620,21 @@ nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
...
     if (pthrOn) {
-        close(pthrControlPipe[0]);
-        close(pthrControlPipe[1]);
+        if (pthrControlPipe[0] != 0) {
+            close(pthrControlPipe[0]);
+        }
+
+        if (pthrControlPipe[1] != 0) {
+            close(pthrControlPipe[1]);
+        }
```
```suggestion
        for (const auto &pthr_control_pipe : pthrControlPipe) {
            if (pthr_control_pipe != 0) {
                close(pthr_control_pipe);
            }
        }
```

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: the worker does not store its address, I think bcz of memory consumption, so getting address can throw exception anyway.
Basically, checking the address here looks redundant to me, just kept original logic... maybe be remove it all together?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -584,25 +576,22 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
     if (pthrOn) {
         for (auto &uw: uws) {
             pollFds.push_back({uw->getEfd(), POLLIN, 0});
         }
         pollFds.push_back({pthrControlPipe[0], POLLIN, 0});
     }
 
+    // TODO: in case of UCX error handling is enabled, we can clean up AM based connections error
+    //       handling, if user requested disabled error handling, we dont care about it.
+    auto &uw = uws.front();
```
rakhmets, line 594: You can also revert this change to return the line before 583.

evgeny-leksikov: I moved it to here for localization of the following major refactoring, see the TODO comment. `uw` is used only for AM callbacks register which are redundant with native UCX err handling support.

rakhmets: I  mean, previously it was:
```C++
const auto &uw = uws.front();
workerAddr = uw->epAddr();
...
uw->regAmCallback(CONN_CHECK, connectionCheckAmCb, this);
```
, no it's:
```C++
workerAddr = uws.front()->epAddr();
...
auto &uw = uws.front();
uw->regAmCallback(CONN_CHECK, connectionCheckAmCb, this);
```

And you can revert this change, since it's not needed.

evgeny-leksikov: yes, the intention of this change is to remove `uw` variable together with AM callbacks.

rakhmets (dismissed):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -584,25 +576,22 @@ nixlUcxEngine::nixlUcxEngine (const nixlBackendInitParams* init_params)
...
                                           init_params->syncMode);
 
-    for (unsigned int i = 0; i < numWorkers; i++)
+    for (size_t i = 0; i < numWorkers; i++) {
         uws.emplace_back(std::make_unique<nixlUcxWorker>(*uc, err_handling_mode));
-
-    const auto &uw = uws.front();
-    workerAddr = uw->epAddr();
-
-    if (workerAddr.empty()) {
```
gleon99, line 593: This check is not needed anymore?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -628,20 +617,19 @@ nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
...
-    // per registered memory deregisters it, which removes the corresponding metadata too
-    // parent destructor takes care of the desc list
-    // For remote metadata, they should be removed here
-    if (this->initErr) {
-        // Nothing to do
-        return;
-    }
+nixlUcxEngine::~nixlUcxEngine() {
+    // initErr is not used in this class and should be removed
+    NIXL_ASSERT(!this->initErr);
```
gleon99: Maybe remove? It's not used in the class at all..

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: if UCX returns UCS_OK, the address must be valid regardless of its size

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: ok, done

gleon99 (approved):

PR #540, opened 2025-07-07
Add NIXL plugins to the wheel

ovidiusm:
## What?
Bundles NIXL plugins into the packaged python wheel.

## Why?
So that the user can use NIXL plugins when installing nixl via `pip install nixl`.

## How?

The following changes were needed:

* copy NIXL plugins into the wheel, patching their RPATH
* patch the RPATH of all existing libs inside the wheel, so that they can load their dependencies correctly; this was not necessary before because there were no chains like python -> libnixl.so -> dlopen NIXL plugin -> direct-dep.so -> indirect-dep.so

Note that NIXL_PLUGIN_DIR must still be set to a directory inside the installed wheel, e.g. `venv/lib/python3.12/site-packages/nixl.libs/nixl` This is probably not right, but it is a separate issue.

Tested with:

```
docker run -it --privileged --device=/dev/infiniband --net=host --ipc=host --gpus all -e NVIDIA_VISIBLE_DEVICES=all -e UCX_PROTO_INFO=y -v $PWD:/ws nvcr.io/nvidia/cuda-dl-base:25.03-cuda12.8-devel-ubuntu24.04 /bin/bash

apt update
apt install -y python3.12-venv
python3.12 -m venv venv
source venv/bin/activate
pip install /ws/nixl-0.4.0-cp312-cp312-manylinux_2_28_x86_64.whl

python /ws/examples/python/nixl_api_example.py

export NIXL_PLUGIN_DIR=/venv/lib/python3.12/site-packages/nixl.libs/nixl

python /ws/nixl/examples/python/nixl_gds_example.py /tmp
```

### DIFF
--- a/contrib/wheel_add_ucx_plugins.py
+++ b/contrib/wheel_add_ucx_plugins.py
@@ -102,14 +102,15 @@ def get_repaired_lib_name_map(libs_dir):
     (like "libboost_atomic-fb1368c6.so.1.66.0").
     """
     name_map = {}
-    for fname in os.listdir(libs_dir):
+    for fname in sorted(os.listdir(libs_dir)):
         if (
             os.path.isfile(os.path.join(libs_dir, fname))
             and ".so" in fname
             and "-" in fname
         ):
-            base_name = fname.split("-")[0].replace("_", ".")
+            base_name = fname.split("-")[0]
             name_map[base_name] = fname
+            print(f"Found already bundled lib: {base_name} -> {fname}")
     return name_map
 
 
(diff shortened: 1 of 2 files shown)


contrib/wheel_add_ucx_plugins.py
```diff
@@ -102,14 +102,15 @@ def get_repaired_lib_name_map(libs_dir):
...
+    for fname in sorted(os.listdir(libs_dir)):
         if (
             os.path.isfile(os.path.join(libs_dir, fname))
             and ".so" in fname
             and "-" in fname
         ):
-            base_name = fname.split("-")[0].replace("_", ".")
+            base_name = fname.split("-")[0]
             name_map[base_name] = fname
+            print(f"Found already bundled lib: {base_name} -> {fname}")
```
brminich, line 113: in what use case will this output be helpful?

contrib/wheel_add_ucx_plugins.py
```diff
@@ -219,24 +253,34 @@ def add_ucx_plugins(wheel_path, ucx_sys_lib_dir):
...
 
 
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "--ucx-lib-dir",
+        "--ucx-plugins-dir",
+        type=str,
+        help="Path to the UCX plugins directory",
+        default="/usr/lib64/ucx",
```
brminich, line 265: - id assume default path is `/usr/lib64`

contrib/wheel_add_ucx_plugins.py
ovidiusm: This is useful for debugging the log. It is important to know which libraries are loaded from the wheel as opposed to from the system.

contrib/wheel_add_ucx_plugins.py
ovidiusm: The core libraries yes, but plugins (transport modules) are in /usr/lib64/ucx by default

brminich (approved):

PR #544, opened 2025-07-08
[HF3FS] use shared memory to avoid memcpy to improve performance

Pyngon:
## What?
Use shared memory in 3FS plugin to avoid memcpy

## Why?
To get better performance

## How?
To share a memory with the hf3fs backend process, we need to
1. First, create a symlink under {3fs mount point}/3fs-virt/iovs/{uuid} with the target to the shared memory file created with shm_open() and mmap()
2. Initialized an iov with the hf3fs_iovwrap() with the same uuid.
3. Pass the iov to the hf3fs_prep_io() function.


Pyngon:
Work in progress to handle unaligned memory.

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -37,16 +44,29 @@ class nixlHf3fsMetadata : public nixlBackendMD {
         ~nixlHf3fsMetadata() { }
 };
 
+class nixlHf3fsShmMetadata : public nixlBackendMD {
```
zackyoray: I suggest creating a common 3fs MD base class with memtype, that both (file and shm) classes can inherit from.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -95,35 +107,34 @@ nixl_status_t nixlHf3fsEngine::registerMem (const nixlBlobDesc &mem,
...
 
-    out = (nixlBackendMD*) md;
     return status;
 }
 
 nixl_status_t nixlHf3fsEngine::deregisterMem (nixlBackendMD* meta)
 {
-    nixlHf3fsMetadata *md = (nixlHf3fsMetadata *)meta;
-    if (md->type == FILE_SEG) {
+    if (typeid(*meta) == typeid(nixlHf3fsMetadata)) {
```
zackyoray: than you can keep the different handling based on the type

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -61,14 +66,20 @@ nixl_status_t nixlHf3fsEngine::registerMem (const nixlBlobDesc &mem,
...
     int ret;
-    nixlHf3fsMetadata *md = new nixlHf3fsMetadata();
 
     switch (nixl_mem) {
-        case DRAM_SEG:
-            md->type = DRAM_SEG;
-            status = NIXL_SUCCESS;
+        case DRAM_SEG: {
+            try {
+                nixlHf3fsShmMetadata *md = new nixlHf3fsShmMetadata((uint8_t*)mem.addr, mem.len, *hf3fs_utils);
```
zackyoray: As we discussed it will be good to have fallback to the non zero-copy flow for non-align addr

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -61,14 +66,20 @@ nixl_status_t nixlHf3fsEngine::registerMem (const nixlBlobDesc &mem,
     nixl_status_t status;
     int fd;
     int ret;
-    nixlHf3fsMetadata *md = new nixlHf3fsMetadata();
 
     switch (nixl_mem) {
-        case DRAM_SEG:
-            md->type = DRAM_SEG;
-            status = NIXL_SUCCESS;
+        case DRAM_SEG: {
```
zackyoray: i think it might be cleaner to just have another seg type like "SHM_SEG" where we assume the user already allocate the buffer using shm api, and the shm_id will be store under mem.devId. 

like this we can keep both options DRAM_SEG for regular malloc memory, and enhanced flow for shm allocated memory.

@vvenkates27 what do you think?

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -400,3 +387,88 @@ nixlHf3fsEngine::~nixlHf3fsEngine() {
...
+    NIXL_INFO << "Created POSIX shared memory: " << shm_name << " with size: " << len;
+}
+
+nixlHf3fsShmMetadata::~nixlHf3fsShmMetadata()
+{
+    if (unlink(link_path.c_str()) && errno != ENOENT) {
+        NIXL_ERROR << "Failed to remove symlink: " << strerror(errno);
+    }
+
+    if (munmap(mapped_addr, mapped_size) == -1) {
```
zackyoray: Can you verify that user manage to do free(addr) withuot any issue after that?

src/plugins/hf3fs/hf3fs_backend.h
Pyngon: Done.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Done

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Done.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: I added the fall back to regular memory when the shared memory creation failed for now. I think it will be cleaner if added a different type.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Will do

Pyngon: Free is ok, but accessing the address will have seg fault, so looks like munmap will ask the kernel to remove the page from this process.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -212,7 +212,7 @@ static void iovListToNixlXferDlist(const std::vector<xferBenchIOV> &iov_list,
 std::optional<xferBenchIOV> xferBenchNixlWorker::initBasicDescDram(size_t buffer_size, int mem_dev_id) {
     void *addr;
 
-    addr = calloc(1, buffer_size);
+    addr = aligned_alloc(4096, buffer_size);
```
barneuman: let's use sysconf(_SC_PAGESIZE) here too

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -64,21 +103,22 @@ class nixlH3fsThreadStatus {
...
+        ~nixlHf3fsBackendReqH() {}
 };
 
 
 class nixlHf3fsEngine : public nixlBackendEngine {
     private:
-        hf3fsUtil                      *hf3fs_utils;
+        hf3fsUtil *hf3fs_utils;
         std::unordered_set<int> hf3fs_file_set;
+        long page_size;
```
barneuman: do we really need this as a class member? it is only used in one function

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -64,21 +103,22 @@ class nixlH3fsThreadStatus {
...
-       nixlHf3fsBackendReqH() : completed_ios(0), num_ios(0) {}
-       ~nixlHf3fsBackendReqH() {}
+        std::list<nixlHf3fsIO *> io_list;
+        hf3fs_ior ior;
+        uint32_t completed_ios; // Number of completed IOs
+        uint32_t num_ios; // Number of submitted IOs
+        nixlH3fsThreadStatus io_status;
+
+        nixlHf3fsBackendReqH() : completed_ios(0), num_ios(0) {}
+        ~nixlHf3fsBackendReqH() {}
```
barneuman: please use `= default` for the constructor and the destructor and `= 0` for `completed_ios` and `num_ios`

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -25,28 +25,67 @@
...
-        void* orig_addr;  // Original memory address for copying after read
+        void *addr; // Start address to read from/write to
         size_t size;      // Size of the buffer
         bool is_read;     // Whether this is a read operation
         size_t offset;    // Offset in the file
+        nixlHf3fsMemType mem_type;
 
-        nixlHf3fsIO() : fd(-1), orig_addr(nullptr), size(0), is_read(false) {}
+        nixlHf3fsIO() : fd(-1), addr(nullptr), size(0), is_read(false) {}
         ~nixlHf3fsIO() {}
```
barneuman: same as `nixlHf3fsBackendReqH`

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -25,28 +25,67 @@
...
+class nixlHf3fsRegMemMetadata : public nixlHf3fsMetadata {
+public:
+    nixlHf3fsRegMemMetadata() : nixlHf3fsMetadata(NIXL_HF3FS_MEM_TYPE_REG_MEM) {}
+    ~nixlHf3fsRegMemMetadata() {}
+};
+
+class nixlHf3fsShmMetadata : public nixlHf3fsMetadata {
+public:
+    std::string shm_name;
+    std::string shm_path;
```
barneuman: seems like `shm_path` is only in use in the nixlHf3fsShmMetadata constructor, no need to keep it

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -25,28 +25,67 @@
...
+    std::string link_path;
+    void *mapped_addr;
+    size_t mapped_size;
+    boost::uuids::uuid uuid;
+
+    nixlHf3fsShmMetadata(uint8_t *addr, size_t len, hf3fsUtil &utils);
+    ~nixlHf3fsShmMetadata();
```
barneuman: please use `= default` instead of an empty destructor in all the classes

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -51,79 +56,92 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
+                           << mem.addr << ", len: " << mem.len << "): " << e.what();
+                // fall back to using regular memory
+            }
+        }
+        out = new nixlHf3fsRegMemMetadata();
+        NIXL_DEBUG << "HF3FS: Registered regular memory(addr: " << std::hex << mem.addr
+                   << ", len: " << mem.len << ")";
+        break;
+    }
         case FILE_SEG: {
```
barneuman: Can you please align all the cases similar to `DRAM_SEG`? It is very hard to read the code when one case has a different alignment

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -203,31 +222,39 @@ nixl_status_t nixlHf3fsEngine::prepXfer (const nixl_xfer_op_t &operation,
...
+                goto cleanup_handle;
+            }
+
+            // For WRITE operations, copy data from source buffer to IOV buffer
+            // For READ operations, we don't need to copy data now - we'll copy after read completes
+            // TODO: Should the data copy in postXfer? User could still modify the data after
+            // prepXfer
+            if (!is_read) {
+                memcpy(io->iov.base, addr, size);
+            }
```
barneuman, line 275:
```suggestion
            if (!is_read)
                memcpy(io->iov.base, addr, size);
```

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -51,79 +56,92 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
     }
+    delete md;
     return NIXL_SUCCESS;
 }
 
 void nixlHf3fsEngine::cleanupIOList(nixlHf3fsBackendReqH *handle) const
 {
     for (auto prev_io : handle->io_list) {
-        hf3fs_utils->destroyIOV(&prev_io->iov);
+        if (prev_io->mem_type == NIXL_HF3FS_MEM_TYPE_REG_MEM) {
```
barneuman: what about sh_mem? Do we need `hf3fs_iovdestroy` after `hf3fs_iovwrap`?

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -400,3 +422,84 @@ nixlHf3fsEngine::~nixlHf3fsEngine() {
     hf3fs_utils->closeHf3fsDriver();
     delete hf3fs_utils;
 }
+
+nixlHf3fsShmMetadata::nixlHf3fsShmMetadata(uint8_t *addr, size_t len, hf3fsUtil &utils)
+    : nixlHf3fsMetadata(NIXL_HF3FS_MEM_TYPE_SH_MEM) {
+
+    // Generate UUID for shared memory name
+    boost::uuids::random_generator uuidGenerator;
```
barneuman: can we avoid using boost?
@tstamler @aranadive, are we ok with using boost in NIXL?

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -400,3 +422,84 @@ nixlHf3fsEngine::~nixlHf3fsEngine() {
...
+    NIXL_INFO << "Created POSIX shared memory: " << shm_name << " with size: " << len;
+}
+
+nixlHf3fsShmMetadata::~nixlHf3fsShmMetadata() {
+    if (unlink(link_path.c_str()) && errno != ENOENT) {
+        NIXL_ERROR << "Failed to remove symlink: " << strerror(errno);
+    }
+
+    if (shm_unlink(shm_name.c_str()) == -1) {
+        NIXL_ERROR << "Failed to unlink shared memory: " << strerror(errno);
```
barneuman: please use `NIXL_PERROR` when you want to print an error with errno

test/unit/plugins/hf3fs/nixl_hf3fs_test.cpp
```diff
@@ -297,13 +324,16 @@ int main(int argc, char *argv[])
     print_segment_title(phase_title("Allocating and initializing buffers"));
 
     for (int i = 0; i < num_transfers; i++) {
-        void* ptr = malloc(transfer_size);
+        void *ptr = aligned_alloc(page_size, transfer_size); // malloc(transfer_size);
```
barneuman: please remove the comment

barneuman:
Do we have performance numbers from this optimization? If not, can you please run some tests to see the gain?

Pyngon:
> Do we have performance numbers from this optimization? If not, can you please run some tests to see the gain?

Here is the result from my local test, no obvious improvement for small block size, and around 6-11% for large block size. Throughput number in MiB/s

| Read (Block size in KB)                        | 1   | 2   | 4   | 8   | 16  | 32  | 64  | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192 | 16384 | 32768 | 65536 | 131072 | 262144 |
|------------------------------------------|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|------|------|------|------|-------|-------|-------|--------|--------|
| Before                              | 6   | 13  | 22  | 43  | 106 | 177 | 179 | 295 | 450 | 599 | 705  | 774  | 827  | 866  | 829   | 853   | 871   | 882    | 876    |
| After     | 6   | 14  | 22  | 46  | 106 | 179 | 280 | 281 | 498 | 625 | 750  | 829  | 917  | 920  | 920   | 908   | 940   | 945    | 941    |
| diff (%)                                 | 0%  | 8%  | 0%  | 7%  | 0%  | 1%  | 56% | -5% | 11% | 4%  | 6%   | 7%   | 11%  | 6%   | 11%   | 6%    | 8%    | 7%     | 7%     |

src/plugins/hf3fs/hf3fs_backend.h
Pyngon: Try to avoid calling to sysconf() for every call to registerMem. I can make it a static.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Sure, didn't notice the clang format change this.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Are there existing uuid generator in NIXL? Or we need to write one ourselves?

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: iov from hf3fs_iovwrap does not need to be destroy

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: I would prefer to always put {} regardless of number of lines in the block. :smile:

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -260,11 +287,12 @@ nixl_status_t nixlHf3fsEngine::postXfer (const nixl_xfer_op_t &operation,
     if (UINT_MAX - hf3fs_handle->num_ios < hf3fs_handle->io_list.size()) {
         HF3FS_LOG_RETURN(NIXL_ERR_NOT_ALLOWED, "Error: more than UINT_MAX ios");
     }
-
     for (auto it = hf3fs_handle->io_list.begin(); it != hf3fs_handle->io_list.end(); ++it) {
         nixlHf3fsIO* io = *it;
-        status = hf3fs_utils->prepIO(&hf3fs_handle->ior, &io->iov, io->iov.base,
-                                     io->offset, io->size, io->fd, io->is_read, io);
+        void *addr = (io->mem_type == NIXL_HF3FS_MEM_TYPE_REG_MEM) ? io->iov.base : io->addr;
```
zackyoray: in case NIXL_HF3FS_MEM_TYPE_SH_MEM, iovwrap set iov.base to addr, therefore io->iov.base should be equal to io->addr. and you can always use io.base

Pyngon: iovwrap use the mapped_address as base, and addr could be somewhere in the middle of the mapped area, they are not always the same.

src/plugins/hf3fs/hf3fs_backend.cpp
aranadive: Why do we need a uuid gen here specifically? Can you use C++11 random library?

Pyngon: I created a simple class to generate the random id.

Pyngon:
> Do we have performance numbers from this optimization? If not, can you please run some tests to see the gain?

This is the data point from Yoray SSD cluster. (unit is MiB/s)


| Read Block Size (B)    | 1024 | 2048 | 4096 | 8192 | 16384 | 32768 | 65536 | 131072 | 262144 | 524288 | 1048576 | 2097152 | 4194304 | 8388608 |
|------------------------|-----|-----|-----|-----|-----|-----|------|------|------|------|------|------|-------|-------|
| Before                            | 28  | 56  | 112 | 218 | 435 | 894 | 1816 | 3341 | 4884 | 5909 | 8661 | 9903 | 11867 | 12522 |
| After                               | 30  | 58  | 119 | 234 | 453 | 939 | 1880 | 3658 | 6090 | 8523 | 9751 | 12260 | 16016 | 16481 |
| diff (%)                           | 7%  | 4%  | 6%  | 7%  | 4%  | 5%  | 4%   | 9%   | 25%  | 44%  | 13%  | 24%  | 35%   | 32%   |

Pyngon:
@barneuman @zackyoray I have addressed all the feedback so far, do help to take another look to see if it is good to merge. Thanks!

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -473,13 +474,15 @@ void xferBenchUtils::checkConsistency(std::vector<std::vector<xferBenchIOV>> &io
             }
             rc = allBytesAre(addr, len, check_val);
             if (true != rc) {
-                std::cerr << "Consistency check failed\n" << std::flush;
+                std::cerr << "Consistency check failed for iov " << i << ":" << j << std::endl;
```
vvenkates27, line 578: Can we use NIXL_ERROR here?

src/plugins/hf3fs/hf3fs_backend.cpp
vvenkates27: @zackyoray  and @Pyngon I like the approach of a fall back and keeping ZC as the default. I think a better approach would be to also set a backend parameter to help force using DRAM/SHM. And also I would like to name it DRAM_ZC instead of SHM. Helping applications control this will facilitate debugging. We also need to ensure the python and rust bindings support this. I don't think having a SHM segment is good option.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -400,3 +425,79 @@ nixlHf3fsEngine::~nixlHf3fsEngine() {
...
 }
+
+nixlHf3fsShmMetadata::nixlHf3fsShmMetadata(uint8_t *addr, size_t len, hf3fsUtil &utils)
+    : nixlHf3fsMetadata(NIXL_HF3FS_MEM_TYPE_SH_MEM) {
+
+    // Create shared memory name using UUID (POSIX shared memory names start with /)
+    shm_name = "/nixl_hf3fs." + uuid.to_string();
+
+    // Create POSIX shared memory object
+    int shm_fd = shm_open(shm_name.c_str(), O_CREAT | O_RDWR, 0666);
```
vvenkates27, line 470: @Pyngon  what is the cost of doing a large number of registrations like 12K registrations where each opens an SHM?  Would it make sense to have a big region and to carve out from that in such a case? Each registration can go to the kernel in this case, and I am worried about the performance for such an registration call.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -400,3 +425,79 @@ nixlHf3fsEngine::~nixlHf3fsEngine() {
...
+    }
+
+    /**
+     * The purpose of this is to keep the content in the memory.
+     * mmap seems to populate the memory with the content of the shared memory fd,
+     * without the pwrite here the memory will be all 0s after mmap.
+     *
+     * TODO: Is it a valid operation to write to the shared memory fd?
+     */
+    ssize_t written = pwrite(shm_fd, addr, len, 0);
```
vvenkates27, line 493: I don't understand the resoning here, why are doing pwrite here?
It appears with this PR we are just moving the copy to  registration instead of during PostXfer?
I am not convinced about how the mapping is done here.
@Pyngon  @zackyoray  @barneuman

barneuman: mmap creates a mapping between a file/shm object and our application's address space; the initial memory comes from the content of that file/object, not the addr buffer.

We want the shared memory segment to contain the data that's currently pointed to by the addr buffer, so we must copy that data into the file descriptor backing the shared memory (using pwrite). Without this step, the mapped segment will default to all zeros or uninitialized content.

Yes, we moved the copy from postxfer to memreg, but it will happen once per memory and not every post

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: This is what we are doing here, user registers one big memory region, and we will only create one shared memory with this big region. And in prepXfer, the memory for transfer will be from this one big region.

benchmark/nixlbench/src/utils/utils.cpp
Pyngon: Sure

Pyngon: Actually, just found that nixlbench code is using std::cerr everywhere and NIXL_ERROR is not made available to nixlbench yet. So, I will keep the std::cerr.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: I can add the backend parameter to force a certain type of memory.

Pyngon: Added a "mem_config" backend params to force the memory type, please have a look.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -34,96 +39,134 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
-        mount_point = init_params->customParams->at("mount_point");
+    if (init_params && init_params->customParams) {
+        if (init_params->customParams->count("mount_point") > 0) {
+            mount_point = init_params->customParams->at("mount_point");
+        }
+        if (init_params->customParams->count("mem_config") > 0) {
+            std::string mem_config_str = init_params->customParams->at("mem_config");
+            if (mem_config_str == "dram") {
+                mem_config = NIXL_HF3FS_MEM_CONFIG_DRAM;
+            } else if (mem_config_str == "dram_zc") {
```
zackyoray, line 56: can you add the new mem_config in the readme?

Pyngon: Sure

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -22,63 +22,102 @@
...
+    nixlHf3fsShmException(const std::string &message) : std::runtime_error(message) {}
+};
+
+enum nixlHf3fsMemType {
+    NIXL_HF3FS_MEM_TYPE_FILE = 0,
+    NIXL_HF3FS_MEM_TYPE_DRAM = 1,
+    NIXL_HF3FS_MEM_TYPE_DRAM_ZC = 2,
+};
+
+enum nixlHf3fsMemConfig {
```
barneuman: can we remove this config? and have drram_zc instead?

src/plugins/hf3fs/hf3fs_backend.h
```diff
@@ -22,63 +22,102 @@
...
 #include <thread>
 #include "hf3fs_utils.h"
 #include "backend/backend_engine.h"
 
+class nixlHf3fsShmException : public std::runtime_error {
+public:
+    nixlHf3fsShmException(const std::string &message) : std::runtime_error(message) {}
+};
+
+enum nixlHf3fsMemType {
```
barneuman: can we revert to nixl_mem_t? Using the zc parameter as an indication

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -34,96 +39,134 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
     if (hf3fs_utils->openHf3fsDriver() == NIXL_ERR_BACKEND) {
         NIXL_ERROR << "Error opening HF3FS driver";
         this->initErr = true;
+        return;
     }
 
+    mem_config = NIXL_HF3FS_MEM_CONFIG_AUTO;
```
barneuman: please move to initializer list or assign as default value in class definition

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -34,96 +39,134 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
     char mount_point_cstr[256];
     auto ret = hf3fs_extract_mount_point(mount_point_cstr, 256, mount_point.c_str());
     if (ret < 0) {
         this->initErr = true;
+        return;
     }
 
     hf3fs_utils->mount_point = mount_point_cstr;
+
+    page_size = sysconf(_SC_PAGESIZE);
```
barneuman: please move to initializer list

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -34,96 +39,134 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
-                md->handle.fd = *it;
-                md->handle.size = mem.len;
-                md->handle.metadata = mem.metaInfo;
-                md->type = nixl_mem;
-                status = NIXL_SUCCESS;
-                break;
+    case DRAM_SEG: {
+        // mmap requires the memory to be aligned to the page size
+        // and the length to be a multiple of the page size
+        if (mem_config == NIXL_HF3FS_MEM_CONFIG_AUTO ||
```
barneuman, line 89: It's not intuitive that we have a parameter for zero copy, but we do it even if the user doesn't set it.
Maybe we need a parameter to disable, so the default behavior would be to try zc, or not to try to use zc if the user doesn't set it.
@zackyoray, @vvenkates27 WDYT?

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -34,96 +39,134 @@ nixlHf3fsEngine::nixlHf3fsEngine (const nixlBackendInitParams* init_params)
...
-            HF3FS_LOG_RETURN(NIXL_ERR_BACKEND, "Error - type not supported");
+
+        nixlHf3fsFileMetadata *md = new nixlHf3fsFileMetadata();
+        md->handle.fd = fd;
+        md->handle.size = mem.len;
+        md->handle.metadata = mem.metaInfo;
+        out = (nixlBackendMD *)md;
+        break;
+    }
+    case VRAM_SEG:
```
barneuman: can we remove the VRAM case?

barneuman (commented):
please apply clang-tidy to your PR

benchmark/nixlbench/src/utils/utils.cpp
vvenkates27: hmm.. ok, nixlbench should also use the same infra. But it can be in a separate PR.

src/plugins/hf3fs/hf3fs_backend.cpp
vvenkates27: If ZC is the performant path then we should use it always,revert back to using non-zc if not requested.

src/plugins/hf3fs/hf3fs_backend.cpp
vvenkates27: So in theory if the user does a bunch of reg_list entries it will take a long time since its a shm creation operation. The onus is on the user to provide a large region here.

src/plugins/hf3fs/hf3fs_backend.cpp
vvenkates27: Thanks, it is stil a lot done during registration, but as long as the user is able to register the CPU memory region as a whole then its not a problem I assume.

src/plugins/hf3fs/hf3fs_backend.cpp
barneuman: so if the user does
zc = false
Should we try to do ZC copy anyway? because it is the performant path

vvenkates27: If the user is asking for zc = false means the zc has not been performant, so we should disable it IMO.

barneuman: What about cases where the user doesn't pass the parameter at all? Is it considered true or false?

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: If the user provide a bunch of small reg_list entries, it will take longer time, how much longer we will need to measure. Before this patch, the shm creation are done in hf3fs_iovcreate() during prepXfer(). shm is the way to pass data to 3FS now, so it has to be done somewhere.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: If the user do not pass the parameter, the default is AUTO. That's mean we will try zc first, and fallback to non-zc.

If the user pass in DRAM, then we will never try zc.
if the user pass in DRAM_ZC, we will never fallback to non-zc.

I think that should be clear with the default as AUTO?

src/plugins/hf3fs/hf3fs_backend.h
Pyngon: The zc parameter can be AUTO, we will not be able to use that as an indication.

src/plugins/hf3fs/hf3fs_backend.h
Pyngon: Need a way to indicate AUTO, can't use mem_type to do that.

src/plugins/hf3fs/hf3fs_backend.cpp
barneuman: yes great

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -412,3 +459,79 @@ nixlHf3fsEngine::queryMem(const nixl_reg_dlist_t &descs,
...
+nixlHf3fsDramZCMetadata::~nixlHf3fsDramZCMetadata() {
+    if (unlink(link_path.c_str()) && errno != ENOENT) {
+        NIXL_PERROR << "Failed to remove symlink";
+    }
+
+    if (shm_unlink(shm_name.c_str()) == -1) {
+        NIXL_PERROR << "Failed to unlink shared memory";
+    }
+
+    NIXL_INFO << "Cleaned up POSIX shared memory: " << shm_name;
```
barneuman, line 536: I don't think this log is accurate

src/plugins/hf3fs/hf3fs_backend.cpp
aranadive: This isnt different from other plugins (like UCX) where registering multiple smaller entries will add overhead. This seems like an acceptable/known trade-off for such an API.

barneuman (approved):

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Do you mean when the above function failed, it is not actually "cleaned up"?

Pyngon: I can update the log with a followed up patch.

barneuman:
/ok to test [1f25a1e](https://github.com/ai-dynamo/nixl/pull/544/commits/1f25a1e256aa5001b16e6623c754b6a4890b1863)

vvenkates27 (approved):

barneuman:
/ok to test [3895843](https://github.com/ai-dynamo/nixl/pull/544/commits/3895843ec832ab5316139b353f9ee720cc616659)

PR #546, opened 2025-07-08
POSIX: Fixed bug on reposting reqH in aioQueue

dfarge:
## What?
Fixed bug on reposting reqH in aioQueue

## How?
Reset certain fields in aioQueue on each post (num_submitted for example) and changed the way aioQueue registers completions.


src/plugins/posix/aio_queue.cpp
```diff
@@ -93,20 +99,19 @@ nixl_status_t aioQueue::checkCompleted() {
         return NIXL_SUCCESS;
 
     // Check all submitted I/Os
-    for (auto& aiocb : aiocbs) {
-        if (aiocb.aio_fildes == 0 || aiocb.aio_nbytes == 0)
-            continue;  // Skip unused control blocks
+    for (int i = 0; i < num_entries; i++) {
+        if (completed[i] || aiocbs[i].aio_fildes == 0 || aiocbs[i].aio_nbytes == 0)
```
vvenkates27, line 103: Do we still need the aio_fildes and aio_nbytes check give we don't set them anymore?

barneuman: In some places, we don't (this one is one of them), but we wanted to focus on fixing the issue and do a larger refactor later on. relying on aio_fildes is an issue in other places in the aio queue too

vvenkates27: Got it

barneuman (approved):

PR #534, opened 2025-07-07
[CI] add build wheel blossom ci job

dpressle:
## What?
Add wheel build in manylinux

## Why?
New CI build test

## How?
New blossom-ci pipeline

## Summary by CodeRabbit

* **New Features**
  * Added a CI pipeline to build and publish Python wheels across multiple Python versions, CPU architectures, and manylinux targets using matrixed build tasks.

* **Documentation**
  * Added a comprehensive Build Wheel Matrix CI guide covering configuration, build/test/publish flow, artifact naming/versioning, Docker usage, and troubleshooting.

* **Chores**
  * Made container builds respect configurable parallelism.
  * Added optional build-ID tagging for wheel versions and integrated the matrix build job into the dispatcher pipeline.

PR #550, opened 2025-07-09
Unify wheel building

ovidiusm:
## What?
Unify the wheel building scripts across the repository.

## Why?
There is a sequence of commands that must be executed correctly and uniformly to result into creating the wheel correctly, including all deps, plugins and ensuring everything loads correctly on installation.

Therefore we move the logic to `contrib/build-wheel.sh` and invoke this script from all places.

This requires https://github.com/ai-dynamo/nixl/pull/540 to be merged first.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -78,17 +78,7 @@ RUN wget --tries=3 --waitretry=5 \
     rm rustup-init* && \
     chmod -R a+w $RUSTUP_HOME $CARGO_HOME
 
-WORKDIR /workspace/nixl
-COPY . /workspace/nixl
-
-ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
-
-ENV VIRTUAL_ENV=/workspace/nixl/.venv
-RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
-    # pybind11 pip install needed for ubuntu 22.04
-    uv pip install --upgrade meson pybind11 patchelf
-
-    # Add Mellanox repository and install packages
+# Add Mellanox repository and install packages
 RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64-sbsa"; else echo "${ARCH}"; fi) && \
     export PKG_CONFIG_PATH="/opt/mellanox/doca/lib/${ARCH_SUFFIX}-linux-gnu/pkgconfig:/opt/mellanox/dpdk/lib/${ARCH_SUFFIX}-linux-gnu/pkgconfig:$PKG_CONFIG_PATH" && \
     curl -fsSL https://linux.mellanox.com/public/repo/doca/3.0.0/ubuntu24.04/${ARCH_SUFFIX}/GPG-KEY-Mellanox.pub | \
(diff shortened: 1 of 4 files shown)


contrib/Dockerfile
```diff
@@ -102,6 +92,16 @@ RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64-sbsa"; else ech
...
 
+WORKDIR /workspace/nixl
+COPY . /workspace/nixl
+
+ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
+
+ENV VIRTUAL_ENV=/workspace/nixl/.venv
+RUN uv venv $VIRTUAL_ENV --python $DEFAULT_PYTHON_VERSION && \
+    # pybind11 pip install needed for ubuntu 22.04
+    uv pip install --upgrade meson pybind11 patchelf
```
rakhmets: Do we really need to install packages in a virtual environment? It's a docker container anyway.

rakhmets: Perhaps not exactly the scope of this PR. But for future enhancements.

ovidiusm: I think the pip install step was added to have an environment where one could run the python examples out of the box

rakhmets: I mean I expect it will still work without `ENV VIRTUAL_ENV=/workspace/nixl/.venv`, and `venv $VIRTUAL_ENV`.

ovidiusm: For example, the ./contrib/build-wheel.sh requires patchelf, if we remove the venv and the pip install steps above that would fail. Unless we add patchelf to the toml file IIUC which is not something we want to do since it is not an actual dependency of nixl, just of the build and packaging scripts

rakhmets: I meant keep pip install, but not in venv. E.g.:
```
uv pip install --system --upgrade meson pybind11 patchelf
```
, or:
```
uv pip install --python /path/to/python --upgrade meson pybind11 patchelf
```

ovidiusm: Now I see what you mean. The normal way nowadays is to use venv and not system wide installation.

ovidiusm: Are you okay with keeping it like this?

rakhmets: Yes, sure

contrib/Dockerfile
```diff
@@ -102,6 +92,16 @@ RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64-sbsa"; else ech
         doca-sdk-eth doca-sdk-flow doca-sdk-rdma doca-all \
         doca-sdk-gpunetio libdoca-sdk-gpunetio-dev
 
+WORKDIR /workspace/nixl
+COPY . /workspace/nixl
```
aranadive, line 100: After copying lets remove nixl/.venv dir (or sjip copying it) since nixl could be built locally as well then a container created.

contrib/Dockerfile
```diff
@@ -142,17 +141,11 @@ RUN echo "/usr/local/nixl/lib/$ARCH-linux-gnu" > /etc/ld.so.conf.d/nixl.conf &&
...
-    done
-
-# Exclude libcuda.so.1 due to compatibility issues, should link with cuda driver library on host
-RUN uv pip install auditwheel && \
-    uv run auditwheel repair --exclude libcuda.so.1 /tmp/dist/nixl-*cp31*.whl --plat $WHL_PLATFORM --wheel-dir /workspace/nixl/dist
+RUN ./contrib/build-wheel.sh \
+    --python-version $DEFAULT_PYTHON_VERSION \
+    --platform manylinux_2_39_$ARCH \
+    --ucx-plugins-dir /usr/lib/ucx \
+    --nixl-plugins-dir /usr/local/nixl/lib/$ARCH-linux-gnu/plugins \
```
aranadive: Can we make the nixl install dir configurable?

ovidiusm: In what sense?

aranadive: Either we add a new parameter in the build-container script to allow user to specify it or define it at top of file.

ovidiusm: I see. Maybe we should do it in https://github.com/ai-dynamo/nixl/pull/534

aranadive: Not sure since that seems specific to wheel building. This also applies to setting the right prefix on line 131. Maybe a separate PR is okay as well.

contrib/Dockerfile
ovidiusm: Done

contrib/Dockerfile
ovidiusm: I made the paths configurable and cleaned up a bit

aranadive (approved):

PR #551, opened 2025-07-09
[CI] add pylint check

dpressle:
## What?
Add new CI check for pylint

## Why?
Validate python code

## How?
Add workflow yml for running pylint on GHA runner


brminich:
@dpressle is it relevant?

PR #504, opened 2025-06-26
add support in nixlbench for O_DIRECT

cheese-head:
## What?
This PR provides support for O_DIRECT in NIXLBench.
## Why?
To provide a more accurate representation of storage access performance when using the POSIX Plugin
## How?
It aligns memory to the underlying page size using posix_memalign

PR #557, opened 2025-07-09
NIXLBench: Align memory for O_DIRECT

cheese-head:
## What?
This PR provides support for O_DIRECT in NIXLBench.

## Why?
To provide a more accurate representation of storage access performance when using the POSIX Plugin

## How?
It aligns memory to the underlying page size using posix_memalign


cheese-head:
cherry pick of [504](https://github.com/ai-dynamo/nixl/pull/504)

vvenkates27 (approved):
mostly just sytle changes and includes posix_memalign. LGTM

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -368,57 +381,77 @@ createFileFds (std::string name) {
...
-                  << " bytes of memory" << std::endl;
-        return std::nullopt;
+    void *buf;
+    long page_size = sysconf(_SC_PAGESIZE);
+    if (page_size == 0) {
+        std::cerr << "Error: Invalid page size returned by sysconf" << std::endl;
+        exit(EXIT_FAILURE);
+    }
+    if (xferBenchConfig::storage_enable_direct) {
+        int rc = posix_memalign(&buf, (int)page_size, buffer_size);
```
ovidiusm: Not so important, but this call to posix_memalign is dfifferent than the other in this patch: page_size is cast to int and the pointer is not null checked.

ovidiusm: This is a comment regarding the overall procedure to add clang format: It was quite difficult to find the actual changes in the PR due to the clang-format changes. It would be best to do the two steps separately.

ovidiusm (dismissed):

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
cheese-head: I've added a new MR that just handles the clang formatting for nixl_worker.cpp https://github.com/ai-dynamo/nixl/pull/565

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -214,10 +213,26 @@ std::optional<xferBenchIOV>
...
+            return std::nullopt;
+        }
+        memset(addr, 0, buffer_size);
+    } else {
+        addr = calloc(1, buffer_size);
+        if (!addr) {
+            std::cerr << "Failed to allocate " << buffer_size << " bytes of DRAM memory"
+                      << std::endl;
+            return std::nullopt;
+        }
```
aranadive: Lets refactor this into a separate allocation function so you can reuse it in initBasicDescFile as well.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -428,6 +462,15 @@ xferBenchNixlWorker::allocateMemory(int num_lists) {
     }
     buffer_size = xferBenchConfig::total_buffer_size / (num_devices * num_lists);
 
+    if (xferBenchConfig::storage_enable_direct) {
+        long page_size = sysconf(_SC_PAGESIZE);
```
aranadive: We can move page_size to be part of xferBenchConfig and is found during initialization. That way we dont need to compute it every time.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -349,7 +364,8 @@ createFileFds(std::string name) {
 
     for (int i = 0; i < num_files; i++) {
         std::string file_name = file_path + file_name_prefix + name + "_" + std::to_string(i);
-        std::cout << "Creating " << " file: " << file_name << std::endl;
+        std::cout << "Creating "
```
vvenkates27, line 406: @aranadive can we use NIXL logging in here too?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
vvenkates27: +1

cheese-head: Since NIXLBench currently uses a mixture of posix_memalign, calloc, and malloc, I'm thinking of creating a function allocateXferMemory and a new enum type `AllocationType`
```c
#
enum class AllocationType { POSIX_MEMALIGN, CALLOC, MALLOC };
static bool
allocateXferMemory(size_t buffer_size,
                   void **addr,
                   std::optional<AllocationType> allocation_type = std::nullopt)

```
Which then can be called like this. 
```c
if (!allocateXferMemory(buffer_size, &buf, AllocationType::MALLOC) || !buf) {
    std::cerr << "Failed to allocate " << buffer_size << " bytes of memory" << std::endl;
    return std::nullopt;
}
```

vvenkates27: Memory alignment size for POSIX memalign is not an option, can we add that?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: Lets do that separately for entire benchmark.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: Perhaps, change the single malloc use to a calloc, so we dont have to specify the exact alloc function? We can add in future if we need to.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -210,16 +209,69 @@ iovListToNixlXferDlist(const std::vector<xferBenchIOV> &iov_list, nixl_xfer_dlis
     }
 }
 
+
+enum class AllocationType { POSIX_MEMALIGN, CALLOC, MALLOC };
+
+static bool
+allocateXferMemory(size_t buffer_size,
+                   void **addr,
```
aranadive, line 217: Add a num parameter here with default as 1. That we can replace other calloc calls with this one.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -210,16 +209,69 @@ iovListToNixlXferDlist(const std::vector<xferBenchIOV> &iov_list, nixl_xfer_dlis
...
+            std::cerr << "Failed to allocate " << buffer_size << " bytes of DRAM memory"
+                      << std::endl;
+            return false;
+        }
+    } else {
+        std::cerr << "Invalid allocation type" << std::endl;
+        return false;
+    }
+    return true;
+}
```
aranadive, line 262: Add newline here.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: Since PR #565 is merged, resolving this.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: Lets skip this for now.

PR #559, opened 2025-07-10
testing: Add plugin gtests infrastructure

mamar123:
## What?
Introduce a GTest-based infrastructure for testing backend plugins. This includes tests for both local and remote transfers, along with plugin-specific tests for the OBJ backend.

## Why?
A unified and automated testing setup is essential to ensure correctness and stability across different plugin implementations. It also simplifies regression testing and future development.

## How?
Built a modular GTest framework for backend validation, implemented core test cases for transfer functionality, and added dedicated tests for the OBJ plugin using environment-driven configurations.


test/gtest/plugins/memory_handler.h
```diff
@@ -0,0 +1,114 @@
...
+
+    static void
+    set(void *addr, char byte, size_t size) {}
+    
+    static void
+    populateBlobDesc(nixlBlobDesc *desc, void *addr, size_t len, int devId) {
+        desc->addr = 0;
+        desc->len = len;
+        desc->devId = devId;
+        desc->metaInfo = "test-obj-key";
```
w1ldptr: You would like want to test with multiple keys eventually. Hardcoding this is not ideal.

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,488 @@
...
+        NIXL_ERROR << "Failed to prepare transfer";
+        return false;
+    }
+
+    teardownCallback_ = [this]() {
+        return DeregDeallocCallback<LocalMemType, XferMemType>();
+    };
+
+    resetLocalBufCallback_ = [this]() {
+        MemoryHandler<LocalMemType>::set(localMemBuf_, 0x00, BUF_SIZE);
```
w1ldptr: Add reset() method to MemoryHandler.

test/gtest/plugins/memory_handler.h
```diff
@@ -0,0 +1,114 @@
...
+    static void
+    populateBlobDesc(nixlBlobDesc *desc, void *addr, size_t len, int devId) {
+        CHECK(false) << "Unsupported memory type!";
+    }
+
+    static void
+    populateMetaDesc(nixlMetaDesc *desc, void *addr, size_t len, int devId, nixlBackendMD *&md) {
+        CHECK(false) << "Unsupported memory type!";
+    }
+};
```
w1ldptr: Why are all these static? Due to them being static you have to pass "addr" everywhere (including OBJ where it is just a nullptr), use 'teardown' callbacks in test instead of proper memory management with smart pointers, all these weird things. Make this a proper class and encapsulate all the state it needs in it instead of forcing user to pass it around.

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    MemoryHandler<MemType>::deallocate(mem_buf);
+
+    return NIXL_SUCCESS;
+}
+
+void
+SetupBackendTestFixture::ResetLocalBuf() {
+    if (resetLocalBufCallback_) {
```
w1ldptr: I think with our current code style you don't need to wrap simple blocks in curly brackets (here and several other places).

test/gtest/plugins/plugins_common.h
```diff
@@ -0,0 +1,143 @@
...
+#include "backend_engine.h"
+
+namespace plugins_common {
+
+/*
+ * Base class for all plugin tests.
+ * Provides common functionality for all plugin tests.
+ */
+class SetupBackendTestFixture : public testing::TestWithParam<nixlBackendInitParams> {
+private:
```
w1ldptr: We use the following order for class members: public, protected, private.

test/gtest/plugins/plugins_common.h
```diff
@@ -0,0 +1,143 @@
...
+                        void *mem_buf,
+                        nixlBackendMD *&md,
+                        int dev_id);
+    
+    template<nixl_mem_t MemType>
+    void
+    populateDescList(nixl_meta_dlist_t &descs, void *buf, nixlBackendMD *&md, int dev_id);
+
+    bool
+    VerifyConnInfo();
```
w1ldptr: Lower camel case for class fields. BTW I think it is also the same for class names which didn't name accordingly in our recent PRs unfortunately, but let's now do it correctly.

test/gtest/plugins/memory_handler.h
mamar123: Implemented multi-bufs support; object keys are now indexed per buffer.

Note: There are two approaches to testing "multi-segments":
1. **Multi-descs** approach – Register one large buffer and split it into multiple chunks, creating a separate descriptor for each chunk.
2. **Multi-bufs** approach – Create multiple independent buffers, register each one separately, and create a descriptor for each.

The Multi-descs method is more "natural" and aligns with real-world usage, but it's not feasible with the OBJ plugin, as AWS does not support writing to specific offsets.

After discussing with Bar, we decided to implement both approaches, and:
- For OBJ, we will test only the Multi-bufs approach.
- For other backends, if supported, we will test the Multi-descs approach (only).

Currently, I’ve implemented the Multi-bufs approach (which suits OBJ), and I’m now working on the Multi-descs one.

test/gtest/plugins/memory_handler.h
mamar123: Refactored into a proper class that encapsulates all related memory data

test/gtest/plugins/plugins_common.cpp
mamar123: I’ll polish the formatting completely at the end

test/gtest/plugins/plugins_common.h
mamar123: Done

test/gtest/plugins/plugins_common.h
mamar123: Are you sure about class names?

aranadive:
@mamar123 @glevnv Does this supersede PR #338? Please make sure that gtest infra for testing plugins is in single PR.

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+void
+MemoryHandler::set(char byte) {
+    switch (memType_) {
+    case DRAM_SEG:
+        for (size_t i = 0; i < len_; i++)
+            ((char *)addr_)[i] = byte + i;
+        break;
+    case OBJ_SEG:
+        break;
```
w1ldptr: Doesn't look like this operation is actually supported for OBJ_SEG.

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,139 @@
...
+            uint8_t expected_byte = (uint8_t)byte + i;
+            if (((char *)addr_)[i] != expected_byte) {
+                NIXL_ERROR << "Verification failed at index " << i
+                           << "! local: " << ((char *)addr_)[i] << ", expected: " << expected_byte;
+                return false;
+            }
+        }
+        break;
+    case OBJ_SEG:
+        break;
```
w1ldptr: ditto

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "absl/log/check.h"
+#include "common/nixl_log.h"
+#include "backend/backend_aux.h"
+#include "memory_handler.h"
+
+void
+MemoryHandler::allocate(size_t len) {
+    switch (memType_) {
+    case DRAM_SEG:
+        addr_ = new char[len];
```
w1ldptr: Any reason not to make this a vector instead of manually managing arrays on heap?

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,139 @@
...
+        break;
+    default:
+        CHECK(false) << "Unsupported memory type!";
+        break;
+    }
+    len_ = len;
+}
+
+void
+MemoryHandler::deallocate() {
```
w1ldptr: Why aren't allocate/deallocate just ctor/dtor? Having semi-constructed/seme-destructed objects is a code smell.

src/plugins/obj/obj_backend.cpp
```diff
@@ -146,6 +146,8 @@ nixlObjEngine::registerMem(const nixlBlobDesc &mem,
             nixl_mem, mem.devId, mem.metaInfo.empty() ? std::to_string(mem.devId) : mem.metaInfo);
         devIdToObjKey_[mem.devId] = obj_md->objKey;
         out = obj_md.release();
+    } else {
+        out = nullptr;
```
w1ldptr: Is this a fix? Make it a standalone PR.

test/gtest/plugins/plugins_common.h
```diff
@@ -0,0 +1,138 @@
...
+namespace plugins_common {
+
+/*
+ * Base class for all plugin tests.
+ * Provides common functionality for all plugin tests.
+ */
+class SetupBackendTestFixture : public testing::TestWithParam<nixlBackendInitParams> {
+protected:
+    std::unique_ptr<nixlBackendEngine> remote_backend_engine_;
+    std::unique_ptr<nixlBackendEngine> backend_engine_;
```
w1ldptr: Naming.

test/gtest/plugins/plugins_common.h
w1ldptr: Please consult [the doc](https://github.com/ai-dynamo/nixl/blob/main/docs/CodeStyle.md).

test/gtest/plugins/memory_handler.h
w1ldptr: It is not either specialized template class with bunch of static function vs regular non-generic class with dynamic functions. Make it a template, create specializations per-memory type, have it properly encapsulate all the data you need instead of manually passing data around to the static functions.

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,460 @@
...
+              << " and device ID " << mem_handler->getDevId();
+
+    nixlBackendMD *md;
+    nixl_status_t ret;
+    ret = engine->registerMem(desc, mem_handler->getMemType(), md);
+    if (ret != NIXL_SUCCESS) {
+        NIXL_ERROR << "Failed to register memory: " << ret;
+        return ret;
+    }
+    mem_handler->setMD(md);
```
w1ldptr: I wonder if it will be more correct to store shared_ptr<nixlBackendEngine> in MemoryHandler and move all the code from backendAllocReg() to MemoryHandler() and from backendDeregDealloc() to ~MemoryHandler(). After all these:

- Don't use any SetupBackendTestFixture() which is strong indicator that the code doesn't belong to this class' methods.
- Implement manual resource management (in disguise) which should be automated by RAII.

WDYT?

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+SetupBackendTestFixture::backendDeregDealloc(nixlBackendEngine *engine,
```
w1ldptr: Why are you passing raw pointer around for no good reason?

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include "backend_engine.h"
+#include "plugins_common.h"
+
+namespace plugins_common {
+
+nixl_status_t
+SetupBackendTestFixture::backendAllocReg(nixlBackendEngine *engine,
+                                         nixl_mem_t mem_type,
+                                         size_t len,
+                                         std::unique_ptr<MemoryHandler> &mem_handler,
```
w1ldptr: Don't hide output parameter in the middle of parameter list. Return std::Optional or absl::StatusOr, etc.

test/gtest/plugins/obj_plugin.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include "plugins_common.h"
+#include "obj/obj_backend.h"
+
+namespace gtest {
+namespace obj_plugin {
```
w1ldptr: namespace gtest::obj_plugin

test/gtest/plugins/obj_plugin.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    const nixlBackendInitParams obj_test_params = {.localAgent = "Agent1",
+                                                   .type = "OBJ",
+                                                   .customParams = &obj_params,
+                                                   .enableProgTh = false,
+                                                   .pthrDelay = 0,
+                                                   .syncMode =
+                                                       nixl_thread_sync_t::NIXL_THREAD_SYNC_RW};
+
+    class SetupObjTestFixture : public plugins_common::SetupBackendTestFixture {
```
w1ldptr: I think most of the convoluted parts of design here stems from implementing bunch of xfer handling logic in parent class of the test fixture. Instead, extract methods like `setupLocalXfer()` and co into dedicated class (`transferHandler` perhaps?), which would be constructed in every test with test-specific options. Followup in 20 lines down.

test/gtest/plugins/plugins_common.cpp
```diff
@@ -0,0 +1,460 @@
...
+        num_notifs = target_notifs.size();
+        if (backend_engine_->supportsProgTh()) {
+            backend_engine_->progress();
+        }
+    }
+
+    NIXL_INFO << "\nNotification transfer complete";
+
+    if (num_notifs != 1) {
+        NIXL_ERROR << "Expected 1 notification, got " << num_notifs;
```
w1ldptr: These all should be ASSERT, EXPECT, etc which are specifically defined by Gtest for such checks.

test/gtest/plugins/obj_plugin.cpp
```diff
@@ -0,0 +1,73 @@
...
+            backend_engine_ = std::make_unique<nixlObjEngine>(&GetParam());
+        }
+    };
+
+    TEST_P(SetupObjTestFixture, SimpleLifeCycleTest) {
+        EXPECT_TRUE(isLoaded());
+    }
+
+    TEST_P(SetupObjTestFixture, XferTest) {
+        EXPECT_TRUE(isLoaded());
```
w1ldptr: What is the point of continue running if this returned false? Shouldn't most of these be ASSERT_TRUE instead?

test/gtest/plugins/obj_plugin.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    TEST_P(SetupObjTestFixture, XferMultiBufTest) {
+        EXPECT_TRUE(isLoaded());
+        EXPECT_TRUE((setupLocalXfer(DRAM_SEG, OBJ_SEG, false, 3)));
+        EXPECT_TRUE(testLocalXfer(NIXL_WRITE));
+        resetLocalBuf();
+        EXPECT_TRUE(testLocalXfer(NIXL_READ));
+        EXPECT_TRUE(checkLocalBuf());
+        EXPECT_TRUE(teardownXfer());
```
w1ldptr: So this would become:
```
transferHandler<DRAM_SEG, OBJ_SEG> handler(localAgent_, removeAgent_, false, 3);
handler.testLocalXfer(NIXL_WRITE)
handler.resetLocalBuf();
handler.testLocalXfer(NIXL_READ);
handler.checkLocalBuf()
```

This would allow you to easily setup complex multi-threaded scenarios in the future where you spawn bunch of threads, each having its own `transferHandler` instance. Not clear how you would do this with current architecture where xfers/bufs are basically 'global' for the test fixture.

test/gtest/plugins/plugins_common.cpp
mamar123: Neither EXPECT nor ASSERT fully meets our needs here: EXPECT continues execution on failure, while ASSERT terminates the test immediately which prevents us from performing cleanup or rollback logic.
WDYT?

test/gtest/plugins/transfer_handler.h
```diff
@@ -0,0 +1,93 @@
...
+    std::unique_ptr<nixl_meta_dlist_t> dst_descs_;
+    nixlBackendEngine *src_backend_engine_;
+    nixlBackendEngine *dst_backend_engine_;
+    nixl_opt_b_args_t xfer_opt_args_;
+    nixlBackendMD *xfer_loaded_md_;
+    std::string src_agent_name_;
+    std::string dst_agent_name_;
+    int src_dev_id_;
+    int dst_dev_id_;
+    int num_bufs_;
```
w1ldptr: Please run clang-tidy on all new files and fix the naming.

test/gtest/plugins/transfer_handler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    int num_bufs);
+template transferHandler<DRAM_SEG, OBJ_SEG>::~transferHandler();
+template void
+transferHandler<DRAM_SEG, OBJ_SEG>::setLocalMem();
+template void
+transferHandler<DRAM_SEG, OBJ_SEG>::resetLocalMem();
+template void
+transferHandler<DRAM_SEG, OBJ_SEG>::checkLocalMem();
+template void
+transferHandler<DRAM_SEG, OBJ_SEG>::testTransfer(nixl_xfer_op_t op);
```
w1ldptr: You are doing this because other translation units that use transferHandler fail to compile when they instantiate it with template parameters not listed here, right? Just move the class functions to the header file since fully defining templates in header files is the idiomatic approach in C++.

test/gtest/plugins/memory_handler.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "backend/backend_aux.h"
+#include "backend_engine.h"
+#include "common/nixl_log.h"
+#include "nixl.h"
+
+template<nixl_mem_t memType> class memoryHandler {
+public:
+    memoryHandler(size_t len, int dev_id) {
+        CHECK(false) << "memoryHandler() is not implemented for <" << memType << "> memory type";
```
w1ldptr: What is CHECK? Something from Abseil? Use Gtest macros in tests (ASSERT*/EXPECT*).

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,118 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <absl/strings/str_format.h>
+#include "memory_handler.h"
+
+// DRAM_SEG specific functions
+template<>
+memoryHandler<DRAM_SEG>::memoryHandler(size_t len, int dev_id) : len_(len),
```
w1ldptr: Move specializations into the header file. Declare full specialized classes instead of just manually initializing specific functions one-by-one. This would allow to get rid of the default memoryHandler method implementations that all just fail.

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,118 @@
...
+                                                                 dev_id_(dev_id) {
+    addr_ = malloc(len);
```
w1ldptr: No manual memory management please. Use std::array or some other appropriate container instead.

test/gtest/plugins/memory_handler.cpp
```diff
@@ -0,0 +1,118 @@
...
+template<>
+void
+memoryHandler<DRAM_SEG>::set(char byte) {
+    for (size_t i = 0; i < len_; i++)
+        ((char *)addr_)[i] = byte + i;
+}
+
+template<>
+bool
+memoryHandler<DRAM_SEG>::check(char byte) {
```
w1ldptr: set/check a single char is a very misleading naming since a user would assume these just set/validate that the memory with the same byte value. setIncreasing/verifyIncreasing or something like that would be much more descriptive.

test/gtest/plugins/transfer_handler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    std::unique_ptr<nixlBackendEngine> &src_engine,
+    std::unique_ptr<nixlBackendEngine> &dst_engine,
+    bool split_buf,
+    int num_bufs)
+    : num_bufs_(num_bufs) {
+
+    CHECK(num_bufs_ <= (int)MAX_NUM_BUFS) << "Number of buffers exceeds maximum number of buffers";
+    src_backend_engine_ = src_engine.get();
+    src_agent_name_ = LOCAL_AGENT_NAME;
+    src_dev_id_ = 0;
```
w1ldptr: Why is num_bufs_ in initialized list but these aren't?

test/gtest/plugins/transfer_handler.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+    bool remote_xfer = src_engine.get() != dst_engine.get();
+    if (remote_xfer) {
+        CHECK(src_engine->supportsRemote()) << "Local engine does not support remote transfers";
+        dst_backend_engine_ = dst_engine.get();
```
w1ldptr: What is the point of using automatic memory management if you just extract the pointer from it and store it in class field while the caller still owns the unique_ptr and can destroy it concurrently whenever they want? You should either assume ownership of unique_ptr (remove & from ctor parameter) or use shared_ptr instead.

test/gtest/plugins/transfer_handler.cpp
```diff
@@ -0,0 +1,348 @@
...
+    } else {
+        CHECK(src_engine->supportsLocal()) << "Local engine does not support local transfers";
+        dst_backend_engine_ = src_engine.get();
+        dst_agent_name_ = LOCAL_AGENT_NAME;
+        dst_dev_id_ = src_dev_id_;
+    }
+
+    for (int i = 0; i < num_bufs_; i++) {
+        src_mem_[i] = std::make_unique<memoryHandler<srcMemType>>(BUF_SIZE, src_dev_id_ + i);
+        dst_mem_[i] = std::make_unique<memoryHandler<dstMemType>>(BUF_SIZE, dst_dev_id_ + i);
```
w1ldptr: Refactor these to be vectors instead of C-arrays. No need to check that `num_bufs_ <= (int)MAX_NUM_BUFS` then and you can use more convenient loops instead of `for (int i = 0; i < num_bufs_; i++)` everywhere.

test/gtest/plugins/plugins_common.h
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+#ifndef __PLUGIN_TEST_H
+#define __PLUGIN_TEST_H
+
+#include <gtest/gtest.h>
+#include <absl/log/check.h>
+#include "backend_engine.h"
+
+namespace plugins_common {
```
w1ldptr: Namespacing in this PR is inconsistent:

- Global namespace here.
- gtest::obj_plugin in object plugin.
- transferHandler is not namespaced at all.

I don't have strong opinion either way, but please decide on the approach you like and unify all new code accordingly.

test/gtest/plugins/transfer_handler.h
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+private:
+    static constexpr uint8_t LOCAL_BUF_BYTE = 0x11;
+    static constexpr uint8_t XFER_BUF_BYTE = 0x22;
+    static constexpr size_t NUM_ENTRIES = 4;
+    static constexpr size_t ENTRY_SIZE = 16;
+    static constexpr size_t BUF_SIZE = NUM_ENTRIES * ENTRY_SIZE;
+    static constexpr std::string_view LOCAL_AGENT_NAME = "Agent1";
+    static constexpr std::string_view REMOTE_AGENT_NAME = "Agent2";
```
w1ldptr: These are now hardcoded both here in the infrastructure code and also in tests code. Consider making agent names constructor parameters for this class.

w1ldptr (dismissed):
Good job!

barneuman:
/ok to test [ef9639f](https://github.com/ai-dynamo/nixl/pull/559/commits/ef9639f547e903ce078bfd657a8a080177900d97)

.github/workflows/aws_s3_validation.yml
```diff
@@ -0,0 +1,78 @@
...
+
+jobs:
+  run_aws_s3_tests:
+    name: Run AWS S3 Tests
+    environment: SWX_AWS
+    runs-on: ubuntu-latest
+    env:
+      AWS_DEFAULT_REGION: eu-central-1
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
dpressle: Wrong Keys:
      AWS_ACCESS_KEY_ID: ${{ secrets.NIXL_AWS_ACCESS_KEY_ID }}
      AWS_SECRET_ACCESS_KEY: ${{ secrets.NIXL_AWS_SECRET_ACCESS_KEY }}

.github/workflows/aws_s3_validation.yml
```diff
@@ -0,0 +1,78 @@
...
+        working-directory: ./.gitlab
+        timeout-minutes: 180
+        run: |
+          echo "Running AWS S3 Tests on bucket $AWS_DEFAULT_BUCKET"
+          ./test_plugins.sh ${NIXL_INSTALL_DIR}
+      - name: Cleanup AWS resources
+        if: always()
+        run: |
+          # Delete S3 bucket
+          aws s3 rb s3://${AWS_DEFAULT_BUCKET} --force
```
dpressle: missing new line

.gitlab/build.sh
```diff
@@ -114,6 +116,15 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/${UCX_VERSION}" | tar xz
   $SUDO ldconfig \
 )
 
+( \
```
dpressle, line 119: I would say we need this also in docker files if we want users to be able to run s3 plugin in them

dpressle (changes requested):

barneuman:
/ok to test [d428994](https://github.com/ai-dynamo/nixl/pull/559/commits/d4289943227e6bb8807f393ae93ab1ff674be883)

barneuman:
/ok to test [f97535f](https://github.com/ai-dynamo/nixl/pull/559/commits/f97535f6417498c386f3c15c8454388fdddbb041)

.gitlab/build.sh
barneuman: already there

w1ldptr (approved):

barneuman:
/ok to test [1389d05](https://github.com/ai-dynamo/nixl/pull/559/commits/1389d053bdaeb8b2aef1fa0406b5c6d46a3f12e3)

.gitlab/build.sh
dpressle: i dont see it in benchmark/nixlbench/contrib/Dockerfile and in manylinux docker file (not sur we need there)

barneuman: It's on another PR, which adds obj plugin nixl bench support
[NixlBench OBJ Support](https://github.com/ai-dynamo/nixl/pull/607)

barneuman:
/ok to test [9083986](https://github.com/ai-dynamo/nixl/pull/559/commits/9083986529aa54578079f12661c1e6ead782df91)

dpressle (approved):

PR #560, opened 2025-07-10
API: Applied naming conventions according to NIXL code style.

rakhmets:
## What?
Applied naming conventions according to NIXL code style.
Deprecated old names keeping them as aliases to not break the API.


yosefe:
IMO we should not make name convention changes

rakhmets:
> IMO we should not make name convention changes

Agree. But this is not a change, it is an alignment with the one described in `docs/CodeStyle.md`. As an another option we can try to adjust `docs/CodeStyle.md` to the existing code (https://github.com/ai-dynamo/nixl/pull/561).

PR #561, opened 2025-07-10
Fixed the Data types code style to be consistent with the current API.

mkhazraee:
## What?
Fixed the Data types code style to be consistent with the current API.

## Why?
There was an inconsistency with API and code style.

### DIFF
--- a/docs/CodeStyle.md
+++ b/docs/CodeStyle.md
@@ -19,8 +19,8 @@ limitations under the License.
 
 ## Naming Conventions
 
-* **Lower camel case** (e.g., `myVariable`): classes, class members, member functions, structs
-* **Snake case** (e.g., `my_variable`): function arguments, local variables, aliases, enums, unions
+* **Lower camel case** (e.g., `myVariable`): compound data types (class, struct, union), class members, member functions
+* **Snake case** (e.g., `my_variable`): function arguments, local variables, enums and aliases (with _t suffix)
 
 ## Class Design
 


docs/CodeStyle.md
```diff
@@ -19,8 +19,8 @@ limitations under the License.
 
 ## Naming Conventions
 
-* **Lower camel case** (e.g., `myVariable`): Data types, Classes, Class Members, Member functions
-* **Snake case** (e.g., `my_variable`): function arguments, local variables
+* **Lower camel case** (e.g., `myVariable`): Classes, Class Members, Member functions
+* **Snake case** (e.g., `my_variable`): function arguments, local variables, Data types
```
rakhmets: This contradicts the fact that the class is in `Lower camel case`. Should we just remove `Data types`?

rakhmets: I prepared a PR: https://github.com/ai-dynamo/nixl/pull/560. If we agree to merge these changes, then we can simply delete `Data types` from the list.

brminich: imo, id better update code style, than changing names all around the code but still keeping old names as aliases

rakhmets: I don't mind updating the code style. But it should be consistent and not contradictory. I don't understand what `Data types` means in this context. Is `Aliases` (`using foo_bar = baz;`) meant here? Or is it `Any type other than a class (struct)`? Or both? Because `enum`, `enum class` are in snake case at the moment. If so, let's write it down explicitly.

mkhazraee: The idea was the data types that are used in API methods definition to be snake case. I think the confusion and typo exactly happened for structs, that they're lower camel case. How about we put structs instead of Data types in the first line, and keep data types in the second line?

rakhmets: I don't mind adding `Structures` to the first line.
What is missing here, is `Type aliases`(or just `Aliases`). Please add it to the second line. Otherwise, it is not defined what to use as an alias name `using ??? = ...;`.
And if it is obvious to everyone else that `Data types` are all types except those listed in the first line (which is essentially `enums`, `unions`, and it seems that's it), then ok to leave it in the second line. But I would prefer to just put `Enumerations` (which includes `enum`, `enum class`, `enum struct` names) to the second line.

yosefe: @rakhmets @mkhazraee @w1ldptr are we good with the current version of this PR?

rakhmets: IIUC the purpose of the change is to describe the current state of the code. At the moment, the names of the classes are in **Lower camel case**, the aliases and the names of enumerations (`enum`, `enum class`) are in **Snake case**.

As I wrote in a previous comment, this version has a couple of issues.
1. It does not answer the question of what to use in this case:
```C++
class myClass;

using my_class_alias = myClass; 
// or
using myClassAlias = myClass;
```
The discussion of the code style started with this case.

2. Data types includes classes (a user-defined type). So we have the same entity in two lists.

-
1. I suggest to add `Aliases` to **Snake case** line to fix the first issue.

2. To solve the second problem, any of the following ways can be used:
a. Add `Data types (excluding classes)` to **Lower camel case**. Meaning that a structure is the same as a class (except that its members are public by default).
b. Add `Structures` to **Lower camel case**, add `Enumerations` to **Snake case**.
c. Add `Data types` to **Snake case**. Meaning that since we added classes to the first list, all other types fall here.

mkhazraee: I'm going with 1 and 2.b, I think data types is too general and might be confusing, so better to be specific. Also added unions that you mentioned in case.

Making all the words lower case to be consistent, as Class doesn't need to be capitalized, or Member functions looked weird. And then using short form struct/union/enum (instad of structures, enumerations).

yosefe (dismissed):

docs/CodeStyle.md
```diff
@@ -19,8 +19,8 @@ limitations under the License.
 
 ## Naming Conventions
 
-* **Lower camel case** (e.g., `myVariable`): Data types, Classes, Class Members, Member functions
-* **Snake case** (e.g., `my_variable`): function arguments, local variables
+* **Lower camel case** (e.g., `myVariable`): classes, class members, member functions, structs
+* **Snake case** (e.g., `my_variable`): function arguments, local variables, aliases, enums, unions
```
w1ldptr: Just wondering, what is the point of making classes and structs camelCase but union (the only other compound data type) snake_case? Looking at the code it seems we don't even have any unions currently so it is not needed for any kind of backward compatibility but just introduces a weird inconsistency into the code style.

PS Feel free to resolve the comment and ignore, not a major concern for me.

PR #564, opened 2025-07-11
Load plugins relative to the install directory

ovidiusm:
## What?
Load plugins relative to the install directory

## Why?
So that it works out of the box regardless where nixl is installed. Including in virtual python environments.

## How?
The core library determines its own path and loads plugins from a relative path.


src/core/nixl_plugin_manager.cpp
```diff
@@ -185,8 +197,16 @@ nixlPluginManager::nixlPluginManager() {
...
-        plugin_dirs_.insert(plugin_dirs_.begin(), plugin_dir);  // Insert at the beginning for priority
+        plugin_dirs_.insert(plugin_dirs_.begin(),
+                            plugin_dir); // Insert at the beginning for priority
         discoverPluginsFromDir(plugin_dir);
+    } else {
+        std::string plugin_dir = getDefaultPluginDir();
+        if (!plugin_dir.empty()) {
+            NIXL_DEBUG << "Loading plugins from default path: " << plugin_dir;
+            plugin_dirs_.insert(plugin_dirs_.begin(), plugin_dir);
+            discoverPluginsFromDir(plugin_dir);
```
rakhmets: Maybe to avoid code duplication put `getenv("NIXL_PLUGIN_DIR")` to `getPluginDir` method, which is:
```C++
std::string
getPluginDir() {
    const char* plugin_dir = getenv("NIXL_PLUGIN_DIR");
    if (plugin_dir) {
        return plugin_dir;
    }

    // code from getDefaultPluginDir
}
```
, and then here:
```C++
const auto plugin_dir = getPluginDir();
if (!plugin_dir.empty()) {
    NIXL_DEBUG << "Loading plugins from: " << plugin_dir;
    plugin_dirs_.insert(plugin_dirs_.begin(), plugin_dir);
    discoverPluginsFromDir(plugin_dir);
}
```

src/core/nixl_plugin_manager.cpp
```diff
@@ -170,6 +171,17 @@ void nixlPluginManager::loadPluginsFromList(const std::string& filename) {
     }
 }
 
+static std::string
```
rakhmets, line 175: Please put the method to unnamed namespace.
```C++
namespace {
    std::string
    getDefaultPluginDir() {
    ...
    }
}
```
Because any “helper” class or function that is defined in an implementation source file should be at an unnamed namespace scope.

src/core/nixl_plugin_manager.cpp
ovidiusm: Thanks, makes sense

src/core/nixl_plugin_manager.cpp
```diff
@@ -182,10 +202,10 @@ nixlPluginManager::nixlPluginManager() {
 #endif
 
     // Check for NIXL_PLUGIN_DIR environment variable
-    const char* plugin_dir = getenv("NIXL_PLUGIN_DIR");
-    if (plugin_dir) {
-        NIXL_DEBUG << "Loading plugins from directory: " << plugin_dir;
-        plugin_dirs_.insert(plugin_dirs_.begin(), plugin_dir);  // Insert at the beginning for priority
+    std::string plugin_dir = getPluginDir();
```
rakhmets, line 204: The comment above this line can be removed now.

rakhmets (dismissed):

src/core/nixl_plugin_manager.cpp
```diff
@@ -170,6 +171,25 @@ void nixlPluginManager::loadPluginsFromList(const std::string& filename) {
...
+        return plugin_dir;
+    }
+    // By default, use the plugin directory relative to the binary
+    Dl_info info;
+    int ret = dladdr(reinterpret_cast<void *>(&getPluginDir), &info);
+    if (ret != 0) {
+        NIXL_ERROR << "Failed to get plugin directory from dladdr";
+        return "";
+    }
+    return std::filesystem::path(info.dli_fname).parent_path().string() + "/plugins";
```
yosefe: can we use operator/ to concat the paths?

yosefe (approved):

rakhmets (approved):

PR #568, opened 2025-07-14
.CI/SCRIPTS: Avoid tcp port collisions in tests

roiedanino:
## What?
Added `common.sh` script to avoid code duplication across CI shell scripts,
which will provide macros for the following:
1. Setting common environment variables such as LD_LIBRARY_PATH and PKG_CONFIG_PATH
2. Usage checking
3. Getting (hopefully) the next available TCP port 

## Why?
1. Avoiding code duplication - easier maintenance.
2. TCP Ports - avoid collisions as some pipeline jobs may run concurrently.


.ci/scripts/common.sh
```diff
@@ -0,0 +1,66 @@
...
+    export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
+}
+
+
+#
+# Set initial port number for client/server applications to be updated with
+# function below
+#
+server_port_range=1000
+EXECUTOR_NUMBER=$(($RANDOM % 52)) # TODO: set using environment variable
```
gleon99: Why 52? Why is it an "executor number"? How randomizing across [0..51] solves anything?

.ci/scripts/common.sh
```diff
@@ -0,0 +1,66 @@
...
+
+#
+# Common functions for CI scripts
+#
+
+#
+# Check test script usage
+# Parse commandline arguments with first argument being the install directory.
+#
+check_usage_install_dir() {
```
gleon99: Maybe `check_install_dir`?

.ci/scripts/common.sh
```diff
@@ -0,0 +1,66 @@
...
+# Set initial port number for client/server applications to be updated with
+# function below
+#
+server_port_range=1000
+EXECUTOR_NUMBER=$(($RANDOM % 52)) # TODO: set using environment variable
+server_port_min=$((10500 + EXECUTOR_NUMBER * server_port_range))
+server_port_max=$((server_port_min + server_port_range))
+server_port=${server_port_min}
+
+step_server_port() {
```
gleon99: Why have the caller responsible to "step" each time?
Better couple getter() + "step" in a single func here.

dpressle:
Tested 2 parallel runs on GPU with this PR and it failed, please check if its something your PR should solve or we have other issues:
https://nbuprod.blsm.nvidia.com/swx-media-master/blue/organizations/jenkins/NIXL%2Fnixl-ci-test/detail/nixl-ci-test/65/pipeline/

roiedanino:
/ok to test [e4b810a](https://github.com/ai-dynamo/nixl/pull/568/commits/e4b810a66579b3da5d81f927f7d519e811cd038b)

.ci/scripts/common.sh
```diff
@@ -0,0 +1,80 @@
...
+elif [ -z "$EXECUTOR_NUMBER" ]; then
+    # Fallback to random number if both CI_CONCURRENT_ID and EXECUTOR_NUMBER are not set
+    EXECUTOR_NUMBER=$(($RANDOM % $(((max_port_number - min_port_number) / server_port_range))))
+fi
+
+echo EXECUTOR_NUMBER=$EXECUTOR_NUMBER
+
+server_port_min=$((min_port_number + EXECUTOR_NUMBER * server_port_range))
+server_port_max=$((server_port_min + server_port_range))
+server_port=${server_port_min}
```
dpressle: why you need all this lines outside the actual function that needs it? it means it is run each time someone sources this file and you override system environment variable EXECUTER_NUMBER which is coming from Jenkins we dont know who else need it so i would change the name you are using

.gitlab/build.sh
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+source $(dirname $0)/../.ci/scripts/common.sh
```
dpressle: This is not the common way to do this source and it is issuing a shell check error we want to keep our script cleaned from shell check errors

.ci/scripts/common.sh
```diff
@@ -0,0 +1,80 @@
...
+    if [ -z "$INSTALL_DIR" ]; then
+        echo "Usage: $0 <install_dir>"
+        exit 1
+    fi
+}
+
+#
+# Set environment variables for the build
+#
+set_env() {
```
yosefe: pls revert the unification of env vars, because:
1. this PR is indented for fixing TCP ports
2. the common varsity are not exactly as they were before the change (i.e cuda dirs) and we don't want to block review of the main fix because of other unrelated changes
3. we may want to remove these scripts altogether

.ci/scripts/common.sh
yosefe: yes, should use a new var instead of overriding EXECUTOR_NUMBER, such as "local_executor_id". but server_port_min/server_port_max have to stay global because they keep the state of last used TCP port

.ci/scripts/common.sh
```diff
@@ -0,0 +1,80 @@
...
+    EXECUTOR_NUMBER=$(($RANDOM % $(((max_port_number - min_port_number) / server_port_range))))
+fi
+
+echo EXECUTOR_NUMBER=$EXECUTOR_NUMBER
+
+server_port_min=$((min_port_number + EXECUTOR_NUMBER * server_port_range))
+server_port_max=$((server_port_min + server_port_range))
+server_port=${server_port_min}
+
+get_next_server_port() {
```
yosefe: get_next_tcp_port

.ci/scripts/common.sh
```diff
@@ -0,0 +1,80 @@
...
+    EXECUTOR_NUMBER=$CI_CONCURRENT_ID
+# Jenkins CI
+elif [ -z "$EXECUTOR_NUMBER" ]; then
+    # Fallback to random number if both CI_CONCURRENT_ID and EXECUTOR_NUMBER are not set
+    EXECUTOR_NUMBER=$(($RANDOM % $(((max_port_number - min_port_number) / server_port_range))))
+fi
+
+echo EXECUTOR_NUMBER=$EXECUTOR_NUMBER
+
+server_port_min=$((min_port_number + EXECUTOR_NUMBER * server_port_range))
```
yosefe: tcp_port_min
tcp_port_max
tcp_port_next

.gitlab/test_cpp.sh
```diff
@@ -74,9 +63,11 @@ cd ${INSTALL_DIR}
 ./bin/test_plugin
 
 # Run NIXL client-server test
-./bin/nixl_test target 127.0.0.1 1234&
+nixl_test_port=$(get_next_server_port)
+
+./bin/nixl_test target 127.0.0.1 $nixl_test_port&
```
yosefe: from quick check there are places in gtest that also open TCP server socket:
1. search for getRandomPort
2. search for getPort 

1 and 2 eventually affect nixlMetadataStream::setupStream

yosefe (changes requested):

.gitlab/build.sh
yosefe: @dpressle what is the right way?

roiedanino: It seems `shellcheck` wants me to use `.` instead of `source` and probably has issues with using variables as part of the path, should be worked around with some comment above it

dpressle: we usualy do it like this
```bash
# shellcheck disable=SC1091
. "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")/../.ci/scripts/common.sh"
```

test/gtest/common.cpp
```diff
@@ -72,4 +76,34 @@ ScopedEnv::Variable::~Variable()
     }
 }
 
+uint16_t
+get_available_tcp_port() {
```
yosefe: but we want to get TCP port from the given range of the executor, not just any available port

yosefe (changes requested):

test/gtest/common.cpp
roiedanino: Why is it matter as binding to 0 just gives you an available tcp port? isn't that much easier?

yosefe: what if it after we select some avail port, another executor will try to use same port number - because it is from the calculated tcp port range?

roiedanino: Fixed it by allocating the second half of each range used by common.sh

.ci/scripts/common.sh
```diff
@@ -0,0 +1,49 @@
...
+#
+tcp_port_range=1000
+min_port_number=10500
+max_port_number=65535
+
+# GITLAB CI
+if [ -n "$CI_CONCURRENT_ID" ]; then
+    nixl_concurrent_id=$CI_CONCURRENT_ID
+# Jenkins CI
+elif [ -z "$EXECUTOR_NUMBER" ]; then
```
roiedanino: check -n and then set nixl_concurrent_id = EXECUTOR_NUMBER

test/gtest/common.cpp
```diff
@@ -72,4 +78,61 @@ ScopedEnv::Variable::~Variable()
...
+PortAllocator &PortAllocator::instance() {
+    static PortAllocator _instance;
+    return _instance;
+}
+
+bool PortAllocator::_is_port_available(uint16_t port) {
+    struct sockaddr_in addr;
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(port);
+    addr.sin_addr.s_addr = INADDR_ANY;
```
roiedanino: Initialization list

test/gtest/metadata_exchange.cpp
```diff
@@ -136,11 +120,9 @@ class MetadataExchangeTestFixture : public testing::Test {
 
     void SetUp() override
     {
-        int port_base = getRandomPort();
-
         // Create two agents
         for (int i = 0; i < AGENT_COUNT_; i++) {
-            int port = port_base + i;
+            int port = PortAllocator::next_tcp_port();
```
roiedanino: const auto

test/gtest/common.cpp
```diff
@@ -76,34 +78,65 @@ ScopedEnv::Variable::~Variable()
...
+uint16_t
+PortAllocator::_get_first_port(int concurrent_id) {
+    return MIN_PORT + concurrent_id * PORT_RANGE + OFFSET;
+}
 
-    ret = getsockname(sock_fd, (struct sockaddr *)&ret_addr, &len);
+int
+PortAllocator::_get_concurrent_id() {
+    char *jenkins_executor_number = getenv("EXECUTOR_NUMBER");
+    char *gitlab_concurrent_id = getenv("CI_CONCURRENT_ID");
```
yosefe: IMO we need to add this as parameter to gtest executable that is passed from test_cpp.sh
for example:
`./gtest ... --tcp-port-range="${tcp_port_min}-${tcp_port_max}"`

test/gtest/common.cpp
```diff
@@ -76,34 +78,65 @@ ScopedEnv::Variable::~Variable()
...
 
-    do {
-        addr_in.sin_port = htons(0);
-        /* Ports below 1024 are considered "privileged" (can be used only by
-         * user root). Ports above and including 1024 can be used by anyone */
-        ret = bind(sock_fd, (struct sockaddr *)&addr_in, sizeof(struct sockaddr_in));
-    } while (ret);
+uint16_t
+PortAllocator::_get_first_port(int concurrent_id) {
+    return MIN_PORT + concurrent_id * PORT_RANGE + OFFSET;
```
yosefe: use ( )

test/gtest/common.h
```diff
@@ -64,8 +66,39 @@ class ScopedEnv {
     std::stack<Variable> m_vars;
 };
 
-uint16_t
-get_available_tcp_port();
```
yosefe: remove also function body?

test/gtest/common.h
```diff
@@ -63,6 +66,40 @@ class ScopedEnv {
...
+public:
+    static constexpr uint16_t MIN_PORT = 10500;
+    static constexpr uint16_t MAX_PORT = 65535;
+    static constexpr uint16_t PORT_RANGE = 1000;
+    // Offset to avoid conflicts with common.sh allocated ports
+    static constexpr uint16_t OFFSET = PORT_RANGE / 2;
+
+private:
+    PortAllocator();
+    ~PortAllocator() = default;
```
yosefe, line 76: do we really need this line?

test/gtest/common.h
```diff
@@ -63,6 +66,40 @@ class ScopedEnv {
...
+
+public:
+    static uint16_t
+    next_tcp_port();
+    static PortAllocator &
+    instance();
+
+private:
+    static uint16_t
+    _get_first_port(int concurrent_id);
```
yosefe: no need for _ prefix

test/gtest/common.cpp
```diff
@@ -72,4 +78,65 @@ ScopedEnv::Variable::~Variable()
...
+uint16_t
+PortAllocator::next_tcp_port() {
+    PortAllocator &instance = PortAllocator::instance();
+    std::lock_guard<std::mutex> lock(instance._mutex);
+    int max_port = MIN_PORT + (instance._concurrent_id + 1) * PORT_RANGE - 1;
+
+    while (!_is_port_available(++instance._port) && (instance._port <= max_port))
+        ;
+
+    if (instance._port >= max_port) {
```
yosefe: maybe wrap around to first port? and fail only if we really scanned all range and not just started from MAX_PORT-1 for example

test/gtest/common.h
roiedanino: "need" not, I think in C++ it's a good practice to have everything explicit, multiple classes are doing so in the repo:
image

roiedanino:
Failure in blossom should be fixed by: https://github.com/ai-dynamo/nixl/pull/624
Failure in gitlab seems unrelated

test/gtest/common.h
yosefe: some need it because they have virtual destructor, here it is not the case

test/gtest/meson.build
```diff
@@ -67,7 +68,7 @@ test_exe = executable('gtest',
     sources : gtest_sources,
     include_directories: [nixl_inc_dirs, utils_inc_dirs],
     cpp_args : cpp_flags,
-    dependencies : [nixl_dep, cuda_dep, gtest_dep, gmock_dep, absl_strings_dep, absl_time_dep, file_utils_interface],
+    dependencies : [nixl_dep, cuda_dep, gtest_dep, gmock_dep, absl_strings_dep, absl_time_dep, absl_flags_dep, file_utils_interface],
```
yosefe: where is absl::Flags being used in this PR?

.ci/scripts/common.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+min_gtest_port() {
+    # Half of the port range is used for gtest, to avoid conflicts with ports allocated by shell script
+    gtest_port_offset=$((tcp_port_range / 2))
+    echo $((tcp_port_min + nixl_concurrent_id * tcp_port_range + gtest_port_offset))
+}
+
+max_gtest_port() {
+    echo $((tcp_port_min + (nixl_concurrent_id + 1) * tcp_port_range - 1))
```
yosefe: can we unite common logic with line 42-43?

.ci/scripts/common.sh
```diff
@@ -39,23 +39,23 @@ fi
...
     echo $tcp_port
 }
 
+# Second half of the port range is used for gtest
+gtest_offset=$((tcp_port_range / 2))
 min_gtest_port() {
-    # Half of the port range is used for gtest, to avoid conflicts with ports allocated by shell script
-    gtest_port_offset=$((tcp_port_range / 2))
-    echo $((tcp_port_min + nixl_concurrent_id * tcp_port_range + gtest_port_offset))
+    echo $((tcp_port_min + gtest_offset))
```
yosefe: btw, why are these functions? can they be just env vars?

.gitlab/test_plugins.sh
```diff
@@ -45,4 +48,4 @@ uname -a || true
 
 echo "==== Running Plugins Gtest tests ===="
 cd ${INSTALL_DIR}
-./bin/plugins_gtest
+./bin/plugins_gtest --min-tcp-port="$(min_gtest_port)" --max-tcp-port="$(max_gtest_port)"
```
yosefe: need to use ` --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port"`

yosefe (changes requested):

yosefe (dismissed):

yosefe:
Unrelated failure:
```
[2025-07-29T16:28:44.903Z] + python3 examples/python/partial_md_example.py
[2025-07-29T16:28:48.274Z] W0729 16:28:47.668908   84323 nixl_descriptors.cpp:316] Descs are not sorted although sorted=True was passed, this may affect performance
[2025-07-29T16:28:48.274Z] W0729 16:28:48.019478   84323 nixl_descriptors.cpp:316] Descs are not sorted although sorted=True was passed, this may affect performance
```

dpressle (changes requested):
should also change for etcd
export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:2379"

dpressle (dismissed):

.gitlab/test_cpp.sh
```diff
@@ -61,8 +61,13 @@ ibv_devinfo || true
...
 echo "==== Running ETCD server ===="
-export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:2379"
-etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} &
+etcd_port=$(get_next_tcp_port)
+etcd_peer_port=$(get_next_tcp_port)
+export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:${etcd_port}"
+export NIXL_ETCD_PEER_URLS="http://127.0.0.1:${etcd_peer_port}"
+etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} \
+     --listen-peer-urls ${NIXL_ETCD_PEER_URLS} --initial-advertise-peer-urls ${NIXL_ETCD_PEER_URLS} \
+     --initial-cluster default=${NIXL_ETCD_PEER_URLS} &
```
yosefe, line 70: can we add starting etcd server to common.sh?

.ci/scripts/common.sh
```diff
@@ -42,12 +42,30 @@ echo nixl_concurrent_id="$nixl_concurrent_id"
 # First half of the port range is used for shell script tests
 tcp_port_min=$((min_port_number + nixl_concurrent_id * tcp_port_range))
 tcp_port_max=$((tcp_port_min + tcp_port_range / 2))
-tcp_port=${tcp_port_min}
 
 get_next_tcp_port() {
-    tcp_port=$((tcp_port + 1))
-    tcp_port=$((tcp_port >= tcp_port_max ? tcp_port_min : tcp_port))
-    echo $tcp_port
+    local port_file="/tmp/nixl_tcp_port_${nixl_concurrent_id}"
```
yosefe, line 47: why need to use a file and not as before when port number is taken from executor id?

yosefe (approved):

.gitlab/test_cpp.sh
roiedanino: In that case, we need to do the same for the rest of the common logic. Can we address it in a separate PR that resolves all code duplications in those scripts?

.ci/scripts/common.sh
roiedanino: It seems that without a file, the port number is not shared between scripts

PR #573, opened 2025-07-15
NIXL/UCX threadpool

iyastreb:
## What?
Enable threadpool in NIXL/UCX to parallelise posting of large requests (in terms of batch size) .

To enable this feature client needs to set "num_threads" in UCX backend params (close to "num_workers").
This essentially specifies the number of **dedicated** workers and threads (1 worker per thread) that are used to post requests and progress.

Progress thread logic remains the same as before: if configured, a single progress thread is used for all **shared** workers. There must be at least one shared worker configured in any use case.

## Why?
Clients report long posting time with large batch sizes (> 1000).
For example, posting of 32k batch size with cuda_ipc protocol being used takes up to 0.25s. 
With this fix we can drastically reduce the time needed to post the request.

## How?
This PR depends on refactoring PR: https://github.com/ai-dynamo/nixl/pull/606 (merged)
It depends also on PR https://github.com/ai-dynamo/nixl/pull/673, which bumps UCX from 1.18 to 1.19
[The performance evaluation](https://confluence.nvidia.com/pages/viewpage.action?pageId=4130093961)

Summary:
With communication over **IB** devices, using multiple threads for posting+progress makes a huge difference in terms of request posting time, and also improves BW.
E.g. on 64k message size, the posting time is reduced by 8x, and BW is 10% better

With **cuda_ipc** communication I see not so huge decrease of posting time, it's basically around 30%, but BW is also 30% better.


benchmark/kvbench/README.md
```diff
@@ -137,6 +137,7 @@ These arguments are used by both `plan` and `profile` commands:
 | `--num_initiator_dev` | Number of devices in initiator processes (default: 1) |
 | `--num_target_dev` | Number of devices in target processes (default: 1) |
 | `--enable_pt` | Enable progress thread |
+| `--progress_threads` |  Number of progress threads (default: 0) |
```
mkhazraee, line 140: I'm thinking number of prog threads, or generally number of threads used per backend, should be backend specific. Maybe we can change it to max_threads_per_bknd, and if not provided any per backend value, reuse this.

That enable_pt was also to say usage of prog thread is fine or we want to run single threaded. One idea is setting max_threads_per_bknd to 0 means enforce single threaded, and 1 means the backend can have a single thread, and so on. And not call it progress anymore. If we do that, maybe we can remove the enable_pt flag.

mkhazraee: I see this matches your agent_config and python change better too. I wasn't a fan of having number of worker threads in agent_config, as its for ucx specifically. I think if we want number of workers as well as number of threads (maybe not anymore after this PR), they should call createBackend and in python we expose that somehow. Still the renaming from numthread to maxthread might be better, so it's not unique for a backend.

src/plugins/ucx/meson.build
```diff
@@ -23,15 +24,15 @@ endif
 if 'UCX' in static_plugins
     ucx_backend_lib = static_library('UCX',
                'ucx_backend.cpp', 'ucx_backend.h', 'ucx_plugin.cpp',
-               dependencies: [nixl_infra, ucx_utils_dep, serdes_interface, cuda_dep, ucx_dep, thread_dep, nixl_common_dep],
+               dependencies: [nixl_infra, ucx_utils_dep, serdes_interface, cuda_dep, ucx_dep, thread_dep, nixl_common_dep, asio_dep],
```
mkhazraee, line 27: GDS_MT is using taskflow, but I realized S3 obj is using asio. Maybe in future we should merge them. I guess a separate PR, for now pushing this PR ASAP.

benchmark/kvbench/README.md
iyastreb: Yes, my idea was to remote `enable_pt` flag (and I have a TODO comment for that)
But this affects quite many places, so can be done later in a separate PR

I also agree that this is backend specific, and can be configured in backend params (together with num_workers)

src/plugins/ucx/meson.build
iyastreb: The important difference between taskflow and asio is that taskflow does not let us control thread routine, register handlers for poll and use internal MPMC queue. Taskflow is very suitable for workflow with multiple independent tasks/threads, but our use case is very different.

We need to control precisely thread routine, we need to register worker EFD for polling, and we need to give priority to post tasks over progress task. All of that is hard to achieve with taskflow (I tried that, spent quite some time on that).

On contrary asio gives us fine-grained control, and therefore I use it here

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -483,75 +552,344 @@ void nixlUcxEngine::progressFunc()
...
+        std::vector<nixlUcxBackendH> m_chunks;
+        size_t m_chunkSize;
+};
+
+class nixlUcxThreadContext {
+    public:
+        nixlUcxThreadContext(asio::io_context &io, nixlUcxWorker &worker,
+                             size_t workerId) :
+            m_io(io), m_worker(worker), m_workerId(workerId),
+            m_ev(new asio::posix::stream_descriptor(m_io, m_worker.getEfd())) {}
```
Pyngon: Need a destructor to clean this up.

benchmark/kvbench/README.md
mkhazraee: I think this line can get confusing:
numThreads(use_prog_thread ? 1 : num_threads),

if it's set to false, then we want it to be 0, otherwise use defaults. I say either we shouldn't touch the agent_config and only add it to ucx init_params, or go all in and introduce max_threads_per_bknd or something like that.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -483,75 +552,344 @@ void nixlUcxEngine::progressFunc()
...
+    }
+
+    m_io = std::make_unique<asio::io_context>();
+
+    m_numDedicatedWorkers = std::min(init_params.numThreads, uws.size() - 1);
+    m_threadContexts.reserve(m_numDedicatedWorkers);
+    m_numSharedWorkers = uws.size() - m_numDedicatedWorkers;
+
+    for (size_t i = 0; i < m_numDedicatedWorkers; i++) {
+        // Shared workers come first in the array
```
Pyngon: Shared workers or dedicated worker?

benchmark/kvbench/README.md
iyastreb: Ok, I changed this to:
`numThreads(std::max(use_prog_thread ? 1u : 0u, num_threads))`
to keep "relaxed" semantics of use_prog_thread, so that when it's set more than one progress thread may be configured

I think we can also configure "num_threads" in backend params, close to "num_workers"

src/plugins/ucx/ucx_backend.cpp
iyastreb: Sure, this is noted in TODO. The problem is that this descriptor must be valid when we destroy parent object, will resolve it in the full PR

iyastreb: Fixed

src/plugins/ucx/ucx_backend.cpp
iyastreb: Fixed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,143 +465,546 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    size_t batchSize = local.descCount();
+    if (batchSize <= MIN_CHUNK_SIZE) {
+        return nixlUcxEngine::prepXfer(operation, local, remote, remote_agent, handle, opt_args);
+    }
+
+    size_t chunkSize =
+        std::clamp(batchSize / m_dedicatedThreads.size(), MIN_CHUNK_SIZE, MAX_CHUNK_SIZE);
+    size_t numChunks = (batchSize + chunkSize - 1) / chunkSize;
+
+    size_t workerId = getWorkerId();
```
Pyngon: getWorkerId() could return both dedicated worker and shared worker, so dedicated worker could be used by more than 1 thread(user threads + dedicated thread from thread pool).

Need to set the mt_type of the worker to nixl_ucx_mt_t::WORKER or make sure only 1 thread will use the dedicated worker.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,143 +465,546 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+    // TODO: find the best split strategy based on batchSize and numThreads
+    // Maybe make it configurable
+    const char *minEnv = getenv("NIXL_MIN_CHUNK_SIZE");
+    const char *maxEnv = getenv("NIXL_MAX_CHUNK_SIZE");
+    size_t MIN_CHUNK_SIZE = (minEnv != NULL) ? atoi(minEnv) : 1024;
+    size_t MAX_CHUNK_SIZE = (maxEnv != NULL) ? atoi(maxEnv) : 8192;
+
+    size_t batchSize = local.descCount();
+    if (batchSize <= MIN_CHUNK_SIZE) {
+        return nixlUcxEngine::prepXfer(operation, local, remote, remote_agent, handle, opt_args);
```
Pyngon, line 991: nixlUcxEngine::prepXfer() get worker by hashing the thread id. The problem is it will distribute load on worker unevenly and performance will be affected.

src/plugins/ucx/ucx_backend.cpp
iyastreb: As you can see function `nixlUcxEngine::getWorkerId` is declared as virtual, so when `nixlUcxThreadpoolEngine` is instantiated then `nixlUcxThreadpoolEngine::getWorkerId` override is invoked. And it handles the concern you raised: only shared workers are accessible:
```
    size_t
    getWorkerId() const override {
        std::thread::id id = std::this_thread::get_id();
        return (std::hash<std::thread::id>{}(id) % m_numSharedWorkers);
    }
```
So `nixlUcxThreadpoolEngine::getWorkerId` may return only shared worker ID.

src/plugins/ucx/ucx_backend.cpp
iyastreb: Yes, here we are talking only about the case when batch size is relatively small, so we don't split it.
You're right, I see the problem.
I think in this case we can still execute this job asynchronously via dedicated worker, but without split.
Maybe we can have another threshold, so that for really small batches (below 16-64?) we always prefer shared worker (sync post) even if distribution is not ideal. Otherwise (for batches above 16-64) we still execute them async without any split.
WDYT would it solve the problem?

src/plugins/ucx/ucx_backend.cpp
Pyngon: oh right, I missed this override. Then, I think one more thing is to set mt_type of shared workers to always be nixl_ucx_mt_t::WORKER.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Yes, i think it would be good to use the dedicated worker without splitting it.
Would there be a problem to always use the dedicated worker regardless of the batch size?

src/plugins/ucx/ucx_backend.cpp
iyastreb: With existing logic, we set `nixl_ucx_mt_t::WORKER`:
```
    // With strict synchronization model nixlAgent serializes access to backends, with more
    // permissive models backends need to account for concurrent access and ensure their internal
    // state is properly protected. Progress thread creates internal concurrency in UCX backend
    // irrespective of nixlAgent synchronization model.
    mt_type = (sync_mode == nixl_thread_sync_t::NIXL_THREAD_SYNC_RW || prog_thread) ?
        nixl_ucx_mt_t::WORKER : nixl_ucx_mt_t::SINGLE;
```
And in principle the change that you request is not directly related to this PR. Because you propose to change lock type for shared workers to be always protected by mutex. Currently there is a way to enforce it by using a `sync` option.
I'm not sure if we want to always unconditionally force `nixl_ucx_mt_t::WORKER` for threadpool use case, because in this sense it's not different from a regular use case without threadpool. If there are users who post from a single thread, then we don't strictly need this lock.

src/plugins/ucx/ucx_backend.cpp
iyastreb: Running post asynchronously adds a bit of overhead, let me quickly measure it.

If this overhead is noticeable, maybe I can make this threshold configurable via ucx backend options? So that by default it's a small value like 16-64, but client can override it to be 0, and in this case all posting are done by dedicated workers

iyastreb: I measured the impact of async post for batch size = 1, and message size 4K, DRAM
Sync latency = 2us, Async latency = 10us.
So async post can be a few times slower than sync one in certain conditions.
I think the best option is to make it configurable, maybe "min_chunk_size" option is enough

src/plugins/ucx/ucx_backend.cpp
Pyngon: Yes, this is the existing behavior.
But i think this add some nuances that for the application to handle. An application with several threads posting variable batch size traffic has to be aware that they have to hold a lock when they are posting a batch smaller than MIN_CHUNK_SIZE. (They can also always hold a lock, but it is not needed when using the threadpool)

iyastreb: `An application with several threads posting variable batch size traffic has to be aware that they have to hold a lock` - correct, this statement is TRUE no matter if threadpool is used or not. So if client posts from multiple threads - he must specify the desired `sync` mode (RW or STRICT).

IMO threadpool is completely orthogonal to that. Moreover, despite we post async with threadpool, we still use shared worker to monitor the status of the composite handle. So if client posts from multiple threads - he **must** set `sync` mode, even if batch is posted asynchronously.

src/plugins/ucx/ucx_backend.cpp
iyastreb: I have just implemented even distribution of shared workers across threads, so maybe it solves this problem.
Basically if your application is using 4 threads, and you want to have a threadpool of 8 threads, then you just need to configure nixlAgent with `num_workers=12, num_threads=8`
Then we have 4 shared workers and 8 dedicated ones. But we guarantee that 4 shared workers are split evenly between 4 app threads, so there is no intersection anymore

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,143 +465,547 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+            } else {
+                m_io.run_one();
+            }
+
+            if (m_requests.empty()) {
+                continue;
+            }
+
+            for (auto it = m_requests.begin(); it != m_requests.end();) {
+                if ((*it)->status() == NIXL_SUCCESS) {
```
iyastreb: From Vlad Buslov:
note that request handle can be arbitrary deleted by nixlAgent on any error, so directly reference any of its fields from another thread will likely result user-after-free in such cases (like "std::atomic<size_t> m_pendingReqs;")

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,143 +465,547 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+
+protected:
+    virtual void
+    run() {
+        // Set timeout event so that the main loop would progress all workers on first iteration
+        bool timeout = true;
+        bool pthrStop = false;
+        while (!pthrStop) {
+            for (size_t i = 0; i < m_pollFds.size() - 1; i++) {
+                if (!(m_pollFds[i].revents & POLLIN) && !timeout) continue;
```
w1ldptr: Note that the timeout handling was added as a temporary workaround for issues with UCX FDs. You can simplify and improve performance of this code by removing the workaround assuming it is no longer needed.

iyastreb: Sure, I write this down and will definitely address this in another PR, this one is already overloaded

src/plugins/ucx/ucx_backend.cpp
iyastreb: In addition to round-robin attachment of shared workers, I also added "split_batch_size" option to backend params, mainly for testing purposes - to enforce split and parallel execution of chunks below the default limit (1024).

src/plugins/ucx/ucx_backend.cpp
iyastreb: Addressed this comment by using `shared_ptr<SharedState>` between main request and all of its chunks. Also added error handling tests to verify the behaviour

iyastreb:
Tests fail because of using UCX 1.18, should be fixed by https://github.com/ai-dynamo/nixl/pull/673

src/plugins/ucx/ucx_backend.h
```diff
@@ -332,4 +345,60 @@ class nixlUcxThreadEngine : public nixlUcxEngine {
     notif_list_t notifPthr_;
 };
 
+namespace asio {
```
tvegas1, line 348: why are those needed?

tvegas1, line 348: any include to use?

src/api/python/_api.py
```diff
@@ -117,6 +119,8 @@ def __init__(
 
         # TODO: populate init from default parameters, or define a set of params in python
         init: dict[str, str] = {}
+        if nixl_conf.num_threads > 0:
```
mkhazraee: Let's move this to be per plugin, and only for the ones that support it. So from line 120 forward it would be:

```
if instantiate_all:
    nixl_conf.backends = self.plugin_list

for bknd in nixl_conf.backends:
    if bknd not in self.plugin_list:
        print(
            "Skipping backend registration",
            bknd,
            "due to the missing plugin.",
        )
    else:    
        # TODO: improve populatation of init from nixl_conf
        init: dict[str, str] = {}
        if nixl_conf.num_threads > 0:
            if (bknd=="UCX" or bknd=="OBJ"):
                init["num_threads"] = str(nixl_conf.num_threads)
            elif (bknd=="GDS_MT"):
                init["thread_count"] = str(nixl_conf.num_threads)
        self.create_backend(bknd, init)
```

mkhazraee: The 
if instantiate_all: 
can be moved further up too, where it's more appropriate.

iyastreb: Thanks, done

mkhazraee (dismissed):

test/gtest/error_handling.cpp
```diff
@@ -41,6 +44,10 @@ namespace nixl {
 
         nixlBackendH* backend_handle = nullptr;
         EXPECT_EQ(ucx_err_handling_mode_peer, params[ucx_err_handling_mode_key]);
+        params["num_workers"] = std::to_string(num_workers);
+        params["num_threads"] = std::to_string(num_threads);
+        // If threadpool is configured always force split
```
brminich, line 49: is it todo comment?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +667,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+    }
+};
+
+void
+nixlUcxChunkBackendH::complete(nixl_status_t status) {
+    NIXL_ASSERT(sharedState_.get() != nullptr);
+    if (status != NIXL_SUCCESS) {
+        nixlUcxBackendH::release();
+        sharedState_->status.store(status);
+        NIXL_TRACE << *this << " completed with error: " << status;
```
brminich: maybe no need as it will be printed again below

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +667,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+    }
+
+    bool
+    isComposite() const override {
+        return true;
+    }
+
+    nixl_status_t
+    release() override {
+        NIXL_DEBUG << "composite " << *this << " releasing";
```
brminich: lets change to TRACE as it is data path

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +667,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+        return nixlUcxEngine::prepXfer(operation, local, remote, remote_agent, handle, opt_args);
+    }
+
+    size_t chunk_size = std::max(batch_size / dedicatedThreads_.size(), splitBatchSize_);
+    size_t num_chunks = (batch_size + chunk_size - 1) / chunk_size;
+
+    size_t worker_id = getWorkerId();
+    nixlUcxCompositeBackendH *comp_handle =
+        new nixlUcxCompositeBackendH(getWorker(worker_id).get(), worker_id, chunk_size, num_chunks);
+    NIXL_DEBUG << "created " << *comp_handle;
```
brminich: maybe trace?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +667,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+    nixlUcxBackendH *int_handle = (nixlUcxBackendH *)handle;
+    if (!int_handle->isComposite()) {
+        return nixlUcxEngine::sendXferRange(
+            operation, local, remote, remote_agent, handle, start_idx, end_idx);
+    }
+
+    nixlUcxCompositeBackendH *comp_handle = (nixlUcxCompositeBackendH *)int_handle;
+    comp_handle->startXfer();
+    size_t chunk_size = comp_handle->getChunkSize();
+    NIXL_DEBUG << "sending " << *comp_handle;
```
brminich: trace

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +667,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+            }
+
+            if (remaining.fetch_sub(1) == 1) {
+                promise.set_value();
+            }
+        });
+    }
+
+    future.wait();
+    NIXL_DEBUG << "sent " << *comp_handle << " with status: " << status.load();
```
brminich: trace

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -991,6 +1468,18 @@ static nixl_status_t _retHelper(nixl_status_t ret,  nixlUcxBackendH *hndl, nixlU
     return NIXL_SUCCESS;
 }
 
+size_t
+nixlUcxEngine::getWorkerId() const {
+    auto it = tlsSharedWorkerMap().find(this);
+    if (it == tlsSharedWorkerMap().end()) {
+        size_t index = sharedWorkerIndex_.fetch_add(1) % getSharedWorkersSize();
+        it = tlsSharedWorkerMap().emplace(this, index).first;
+        NIXL_DEBUG << "engine " << this << " bound shared worker " << index << " to thread "
```
brminich, line 1506: trace?

test/gtest/error_handling.cpp
iyastreb: No, this is an explanation for the line below the comment: why we need to split at 0 batch size.
Basically current test is using a batch of a single element, but I still want to trigger threadpool split behavior to check that error is properly handled in this case

src/plugins/ucx/ucx_backend.h
iyastreb: This is forward declaration for a very heavy asio. I don't want to expose asio dependency to all the consumers of UCX backend

iyastreb: `#include <asio.hpp>`
There is an include for this, but I don't want to expose it in the header file, because this is an impl detail

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: The difference here is that we print the error, which is important to understand why the entire composite request fails. So I would keep it

src/plugins/ucx/ucx_backend.cpp
iyastreb: This is supposed to be printed a very few times, only when we bind a shared worker to a new thread. I would keep this as debug, it seems to be quite important in some scenarios

src/plugins/ucx/ucx_backend.cpp
iyastreb: will change

src/plugins/ucx/ucx_backend.h
```diff
@@ -332,4 +345,60 @@ class nixlUcxThreadEngine : public nixlUcxEngine {
...
+class io_context;
+}
+
+class nixlUcxThreadPoolEngine : public nixlUcxEngine {
```
rakhmets, line 352: Please move the class declaration to separate .h file, and the definition of methods to a separate .cpp file.

src/plugins/ucx/ucx_backend.h
```diff
@@ -332,4 +345,60 @@ class nixlUcxThreadEngine : public nixlUcxEngine {
...
+
+    size_t
+    getSharedWorkersSize() const override {
+        return numSharedWorkers_;
+    }
+
+    nixl_status_t
+    getNotifs(notif_list_t &notif_list) override;
+
+protected:
```
rakhmets, line 378: Why do we need protected members in this class?

src/plugins/ucx/ucx_backend.h
rakhmets, line 352: Do you plan to delete `nixlUcxThreadEngine` in a separate PR?
https://github.com/ai-dynamo/nixl/pull/606#discussion_r2223245240?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +670,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
  * Constructor/Destructor
  *****************************************/
 
 std::unique_ptr<nixlUcxEngine>
 nixlUcxEngine::create(const nixlBackendInitParams &init_params) {
     nixlUcxEngine *engine;
-    if (init_params.enableProgTh) {
+    size_t num_threads = nixl_b_params_get(init_params.customParams, "num_threads", 0);
+    if (num_threads > 0) {
+        engine = new nixlUcxThreadPoolEngine(init_params);
```
rakhmets, line 1114: I'd recommend to use `make_unique` instead,
as I suggested here https://github.com/ai-dynamo/nixl/pull/606#discussion_r2224767823.

src/plugins/ucx/ucx_backend.h
```diff
@@ -332,4 +345,60 @@ class nixlUcxThreadEngine : public nixlUcxEngine {
...
+                  size_t start_idx,
+                  size_t end_idx) const override;
+
+private:
+    std::unique_ptr<asio::io_context> io_;
+    std::unique_ptr<nixlUcxThread> sharedThread_;
+    std::vector<std::unique_ptr<nixlUcxThread>> dedicatedThreads_;
+    size_t numSharedWorkers_;
+    std::mutex notifMutex_;
+    notif_list_t notifThread_;
```
rakhmets, line 400: What does this name mean? Is it "thread notifications"?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +670,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+    size_t batch_size = local.descCount();
+    if (batch_size < splitBatchSize_) {
+        return nixlUcxEngine::prepXfer(operation, local, remote, remote_agent, handle, opt_args);
+    }
+
+    size_t chunk_size = std::max(batch_size / dedicatedThreads_.size(), splitBatchSize_);
+    size_t num_chunks = (batch_size + chunk_size - 1) / chunk_size;
+
+    size_t worker_id = getWorkerId();
+    nixlUcxCompositeBackendH *comp_handle =
```
rakhmets: `auto` since the type is already mentioned in the rhs.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +670,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+
+nixl_status_t
+nixlUcxThreadPoolEngine::sendXferRange(const nixl_xfer_op_t &operation,
+                                       const nixl_meta_dlist_t &local,
+                                       const nixl_meta_dlist_t &remote,
+                                       const std::string &remote_agent,
+                                       nixlBackendReqH *handle,
+                                       size_t start_idx,
+                                       size_t end_idx) const {
+    nixlUcxBackendH *int_handle = (nixlUcxBackendH *)handle;
```
rakhmets: cast?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +670,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+}
+
+nixl_status_t
+nixlUcxThreadPoolEngine::prepXfer(const nixl_xfer_op_t &operation,
+                                  const nixl_meta_dlist_t &local,
+                                  const nixl_meta_dlist_t &remote,
+                                  const std::string &remote_agent,
+                                  nixlBackendReqH *&handle,
+                                  const nixl_opt_b_args_t *opt_args) const {
+    size_t batch_size = local.descCount();
```
rakhmets, line 989: Please add `const`...

and for other variables in this method.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -617,14 +670,427 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
...
+        sharedThread_->start();
+    }
+
+    if (num_threads > 0) {
+        io_.reset(new asio::io_context());
+        dedicatedThreads_.reserve(num_threads);
+        for (size_t i = 0; i < num_threads; ++i) {
+            size_t worker_id = numSharedWorkers_ + i;
+            dedicatedThreads_.emplace_back(
+                std::make_unique<nixlUcxDedicatedThread>(this, init, *io_));
```
rakhmets, line 962: The fact that we construct another object in the object constructor and pass `this` as one of the parameters suggests that there might be a problem.
IIUC the only reason to pass `this` is to check that `nixlUcxThreadEngine::appendNotif` is invoked in progress thread.

That's not an issue of this PR, but something that can be simplified in a separate PR.

src/plugins/ucx/ucx_backend.h
iyastreb: In general I like this idea, but the problem is that there are several classes defined inside ucx_backend.cpp that are used by all engines: nixlUcxBackendH, nixlUcxThread, nixlUcxSharedThread and some common functions.
So if we split this class into a separate file, then we need to split&move these common entities too.. This is gonna be a large refactoring, better be done in a separate PR (if we decide to go that way). Given upcoming code freeze I don't think we can afford that now

iyastreb: Yes, nixlUcxThreadPoolEngine implements everything that nixlUcxThreadEngine does, so it must be removed in the future PR. This one is already quite large, no place for further refactoring

src/plugins/ucx/ucx_backend.h
iyastreb: As you can see these methods override virtual methods from the base class. Protected - because they are not supposed to be called from API, but only from the internal impl

src/plugins/ucx/ucx_backend.h
iyastreb: Yes, I just kept the existing name from nixlUcxThreadEngine

src/plugins/ucx/ucx_backend.cpp
iyastreb: Sure, this can be improved in the future, but I don't see a real issue for now

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: I believe we do not have such a guideline for primitives, and it's not enforced either by clang-format nor by clang-tidy. So I think this is a question of taste, and personally I prefer not to use it for primitives to keep the code shorter

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: I see your point, but this is not public API, we fully control the creation of these objects. One of the NIXL guideline says `Prefer clarity over cleverness`, I think is applied here.

brminich (approved):

src/plugins/ucx/ucx_backend.cpp
rakhmets: The suggestion is base on https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction

src/plugins/ucx/ucx_backend.cpp
rakhmets: Sorry I can't agree with you that this is about cleverness. The recommendation is based on C++ Core Guidelines. There are several topics with recommendation to avoid new/delete operators. And here is one of them https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r11-avoid-calling-new-and-delete-explicitly

src/plugins/ucx/ucx_backend.h
rakhmets: Ok, please address it in a separate PR. Because the .cpp file has become too large, it is becoming increasingly difficult to keep the code clean, avoiding code duplication and spreading out the interfaces of individual classes.

PR #576, opened 2025-07-15
Add support for parallel nixlbench tests

ovidiusm:
## What?
Allows multiple benchmarks to be executed at the same time.

## How?
Adds benchmark name to use separate namespaces for all metadata keys.

### DIFF
--- a/benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
+++ b/benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
@@ -28,15 +28,16 @@
 #define ETCD_EP_DEFAULT "http://localhost:2379"
 
 // ETCD Runtime implementation
-xferBenchEtcdRT::xferBenchEtcdRT(const std::string& etcd_endpoints, const int size,
-                                 int *terminate_input) {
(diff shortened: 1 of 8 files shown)


benchmark/nixlbench/src/runtime/etcd/etcd_rt.cpp
```diff
@@ -28,15 +28,16 @@
...
+                                 const std::string& etcd_endpoints,
+                                 const int size, int *terminate_input) {
     // Store parameters for later use in setup
     stored_etcd_endpoints = etcd_endpoints.empty() ? ETCD_EP_DEFAULT : etcd_endpoints;
     global_size = size;
     terminate = terminate_input;
 
     // Initialize other members
-    namespace_prefix = "xferbench/"; // Namespace for XFER benchmark
+    namespace_prefix = "xferbench/" + benchmark_name + "/"; // Namespace for XFER benchmark
```
aranadive: Call this benchmark_group_name instead? Benchmark_name seems too generic.

yosefe (commented):
can we add this to CI?

PR #579, opened 2025-07-16
BACKEND/UCX: disable error handling by default

evgeny-leksikov:
## What?
BACKEND/UCX: disable error handling by default

## Why?
reported perf regression with LLM

### DIFF
--- a/test/gtest/error_handling.cpp
+++ b/test/gtest/error_handling.cpp
@@ -23,8 +23,8 @@
 
 namespace gtest {
 namespace nixl {
-    constexpr const char* ucx_err_handling_mode_key  = "ucx_error_handling_mode";
-    constexpr const char* ucx_err_handling_mode_peer = "peer";
+    constexpr const char *ucx_err_handling_mode_key = "ucx_error_handling_mode";
+    constexpr const char *ucx_err_handling_dflt_mode = "none";
 
     static nixlBackendH* createUcxBackend(nixlAgent& agent, const std::string& backend_name)
     {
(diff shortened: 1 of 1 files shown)


gleon99 (dismissed):

iyastreb:
There was just another manifestation of this issue: https://github.com/ai-dynamo/nixl/issues/610
We confirmed that error handling works well with UCX 1.19+, but it doesn't work with UCX1.18 (which people are still using a lot).
In order to avoid these performance issues, can we disable error handling only when runtime UCX version is below 1.19.
It should be really trivial, we already have similar code in ucx_utils:
```
    unsigned major_version, minor_version, release_number;
    ucp_get_version(&major_version, &minor_version, &release_number);

    unsigned ucp_version = UCP_VERSION(major_version, minor_version);
    if (ucp_version >= UCP_VERSION(1, 19)) {
        config.modify ("MAX_COMPONENT_MDS", "32");
    }
```

evgeny-leksikov:
> There was just another manifestation of this issue: #610 We confirmed that error handling works well with UCX 1.19+, but it doesn't work with UCX1.18 (which people are still using a lot).

It works with <= 1.19 but with performance regression. The solution which you propose changes functionality as well not only performance. IMO in use case when no care about fault tolerance but needs the best performance, need to disable err handling explicitly (ucx backend has such custom parameter). To have the full functionality + perf, the best option is to upgrade UCX.

Otherwise, in addition to your proposal we heed to document somewhere non-obvious difference in behavior.

iyastreb:
> It works with <= 1.19 but with performance regression. The solution which you propose changes functionality as well not only performance. 

IMO 10x perf degradation means it essentially does not work

Is it critical functionality that is being changed?
Yes, it's a tradeoff. I think for majority of the cases users will tolerate if some error handling functionality is missing, but they can't tolerate such a huge perf degradation. So it's not ideal solution, but rather a better alternative IMO.

What's also important is that NIXL used to work with UCX 1.18 before that change (without error handling), so in fact we keep the existing behaviour for 1.18

evgeny-leksikov:
> I think for majority of the cases users will tolerate if some error handling functionality is missing

usually, users are not happy with unresolved issues even if they rarely happen, err handling bugs are hardly reproducible and may have unpredictable symptoms.

> they can't tolerate such a huge perf degradation. So it's not ideal solution, but rather a better alternative IMO.

I don't agree, for this know issue we have well defined and simple solution - just upgrade UCX.

PR #588, opened 2025-07-17
TEST/GTEST: TestTransfer/RandomSizes - use correct memory type instead of just DRAM

roiedanino:
## What?
Modified test to set VRAM_SEG as memory type when CUDA devices are being used

## Why?
1. RandomSizes test was failing with:
```cpp
cuMemHostRegister_v2(address, length, 0x01) failed: part or all of the requested memory range is already mapped
```
When trying to register CUDA memory as DRAM

2. Consistency with `remoteMDFromSocket` test

### DIFF
--- a/test/gtest/test_transfer.cpp
+++ b/test/gtest/test_transfer.cpp
@@ -282,8 +282,7 @@ protected:
     void
     deregisterMem(nixlAgent &agent,
                   const std::vector<MemBuffer> &buffers,
-                  nixl_mem_t mem_type) const
-    {
+                  nixl_mem_t mem_type) const {
         const auto desc_list = makeDescList<nixlBlobDesc>(buffers, mem_type);
         agent.deregisterMem(desc_list);
     }


test/gtest/test_transfer.cpp
```diff
@@ -430,18 +430,27 @@ TEST_P(TestTransfer, RandomSizes)
...
     };
+    nixl_mem_t mem_type = m_cuda_device ? VRAM_SEG : DRAM_SEG;
 
     for (const auto &[size, count, repeat, num_threads] : test_cases) {
         std::vector<MemBuffer> src_buffers, dst_buffers;
 
-        createRegisteredMem(getAgent(0), size, count, DRAM_SEG, src_buffers);
-        createRegisteredMem(getAgent(1), size, count, DRAM_SEG, dst_buffers);
+        createRegisteredMem(getAgent(0), size, count, mem_type, src_buffers);
+        createRegisteredMem(getAgent(1), size, count, mem_type, dst_buffers);
```
rakhmets, line 446: I'm not sure that I understand the change. Here we passed `mem_type`, so we allocated and registered host memory. And then this type was passed to `doTransfer`. What was the issue here?

roiedanino: We were getting:
1751540971.575733] [cascade02:1    :0]    cuda_copy_md.c:168  UCX  ERROR cuMemHostRegister_v2(address, length, 0x01) failed: part or all of the requested memory ran

ge is already mapped                                                                                                                                                 

[1751540971.575746] [cascade02:1    :0]          ucp_mm.c:76   UCX  ERROR failed to register address 0x555556ce8cb0 (host) length 40 on md[4]=cuda_cpy: Input/output 

error (md supports: host|cuda|cuda-managed)                                                                                                                          

../test/gtest/test_transfer.cpp:357: Failure                                                                                                                         

  status                                                                                                                                                             
    Which is: -4                                                                                                                                                     
  NIXL_SUCCESS                                                                                                                                                       
    Which is: 0                                                                                                                                                      
    
    
    
    When UCX v1.19.0 is used

rakhmets: But I don't see that there is an issue. We create memory - register - do transfer. it should work with host and cuda memory. Maybe we don't properly unregister memory in this test?

roiedanino: Sounds right, I didn't see any deregistrations in that test, added them, that solved the issue as well

rakhmets (approved):

test/gtest/test_transfer.cpp
```diff
@@ -430,18 +436,29 @@ TEST_P(TestTransfer, RandomSizes)
          {1000000, 100, 3, 4},
          {40, 1000, 1, 4}}
     };
+    nixl_mem_t mem_type = DRAM_SEG;
```
yosefe: minor: const

test/gtest/test_transfer.cpp
```diff
@@ -280,6 +280,12 @@ class TestTransfer : public testing::TestWithParam<std::string> {
         registerMem(agent, out, mem_type);
     }
 
+    void
+    deregisterMem(nixlAgent &agent, const std::vector<MemBuffer> &buffers, nixl_mem_t mem_type) {
```
```suggestion
    deregisterMem(nixlAgent &agent, const std::vector<MemBuffer> &buffers, nixl_mem_t mem_type) const {
```

test/gtest/test_transfer.cpp
```diff
@@ -280,6 +280,12 @@ class TestTransfer : public testing::TestWithParam<std::string> {
...
+        auto desc_list = makeDescList<nixlBlobDesc>(buffers, mem_type);
```
```suggestion
        const auto desc_list = makeDescList<nixlBlobDesc>(buffers, mem_type);
```

rakhmets (dismissed):

yosefe (approved):

PR #590, opened 2025-07-18
Disable prog thread to enable ucx workers

praveingk:
## What?
Disable prog thread by default to enable multi-threaded workers

## Why?
For NIXL read/write launch overhead


praveingk:
Closing this, as its handled.

PR #591, opened 2025-07-18
Nixlbench: measure and display tail latency

ovidiusm:
## What?
Collect fine grained time measurements. Display breakdown in different categories (prepare, post, transfer) and tail latency.

## Example

```
****************************************************************************************************************************************************************
NIXLBench Configuration
****************************************************************************************************************************************************************
Runtime (--runtime_type=[etcd])                             : ETCD
ETCD Endpoint                                               : http://localhost:2379
Worker type (--worker_type=[nixl,nvshmem])                  : nixl
Backend (--backend=[UCX,UCX_MO,GDS,GDS_MT,POSIX,Mooncake,HF3FS,OBJ]): UCX
Enable pt (--enable_pt=[0,1])                               : 0
Device list (--device_list=dev1,dev2,...)                   : all
Enable VMM (--enable_vmm=[0,1])                             : 0
Initiator seg type (--initiator_seg_type=[DRAM,VRAM])       : VRAM
Target seg type (--target_seg_type=[DRAM,VRAM])             : DRAM
Scheme (--scheme=[pairwise,manytoone,onetomany,tp])         : pairwise
Mode (--mode=[SG,MG])                                       : SG
Op type (--op_type=[READ,WRITE])                            : WRITE
Check consistency (--check_consistency=[0,1])               : 0
Total buffer size (--total_buffer_size=N)                   : 8589934592
Num initiator dev (--num_initiator_dev=N)                   : 1
Num target dev (--num_target_dev=N)                         : 1
Start block size (--start_block_size=N)                     : 4096
Max block size (--max_block_size=N)                         : 67108864
Start batch size (--start_batch_size=N)                     : 1
Max batch size (--max_batch_size=N)                         : 1
Num iter (--num_iter=N)                                     : 1008
Warmup iter (--warmup_iter=N)                               : 112
Large block iter factor (--large_blk_iter_ftr=N)            : 16
Num threads (--num_threads=N)                               : 1

4096                1              0.941137       4.4            7.0            7.0            0.9            1.0            3.4            4.0
8192                1              1.767074       4.6            7.0            7.0            0.9            1.0            3.7            5.0
16384               1              3.375935       4.9            7.0            7.0            0.9            1.0            3.9            5.0
32768               1              6.090751       5.4            7.0            7.0            0.9            1.0            4.5            5.0
65536               1              9.971364       6.6            7.0            7.0            0.9            1.0            5.7            6.0
131072              1              11.998962      10.9           7.0            7.0            0.9            1.0            10.0           14.0
262144              1              15.186273      17.3           7.0            7.0            0.9            1.0            16.3           22.0
524288              1              17.508110      29.9           7.0            7.0            0.9            1.0            29.0           38.0
1048576             1              20.454476      51.3           7.0            7.0            0.9            1.0            50.3           72.0
2097152             1              18.579746      112.9          7.0            7.0            1.0            4.0            111.7          119.0
4194304             1              19.106374      219.5          7.0            7.0            1.0            5.0            218.4          259.0
8388608             1              20.104322      417.3          7.0            7.0            0.9            4.0            416.1          478.0
16777216            1              20.291902      826.8          7.0            7.0            1.0            4.0            825.7          973.0
33554432            1              20.036104      1674.7         7.0            7.0            1.0            4.0            1673.6         1863.0
67108864            1              19.086277      3516.1         7.0            7.0            1.0            4.0            3514.9         3764.0
```


michal-shalev:
It looks really good!

A few small suggestions to consider:
1. To be consistent with "Avg Lat. (us)" I would change:
    Post avg (us) -> Avg Post
    Post max (us)  -> Max Post
    Tx avg (us) -> Avg Tx
    Tx max (us) -> Max Tx
2. Maybe it would help to clarify whether "Avg Lat." includes all phases (prepare + post + tx) or is measured independently.
3. Why do we need MiB/s, GiB/s, and GB/s? Can we drop one to simplify the table?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -788,27 +798,29 @@ execTransfer(nixlAgent *agent,
...
+                thread_stats.transfer_duration.add(transfer_duration);
             }
         }
-
         agent->releaseXferReq(req);
         if (error) {
             std::cout << "NIXL releaseXferReq failed" << std::endl;
             ret = -1;
         }
+        stats.add(thread_stats);
```
iyastreb: Is it thread-safe to modify common object from multiple threads?

ovidiusm: Good point, I fixed this and also prevented and latency introduced by memory allocation in the object

iyastreb (dismissed):

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -137,15 +138,78 @@ class xferBenchConfig {
...
+        static void
+        printSeparator(const char sep = '-');
+        static std::vector<std::string>
+        parseDeviceList();
         static bool
         isStorageBackend();
 };
 
+// Timer class for measuring durations at high resolution
+class xferBenchTimer {
```
aranadive, line 184: Why not use nixlTime? It uses chrono as well..

ovidiusm: It is not visible here, since we are outside nixl and it is not public API

aranadive: Hmm what do you mean? You should be able to include utils/common/nixl_time.h since its installed with nixl. Thats all you need for using the nixlTime APIs I think - it just relies on chrono.

ovidiusm: I changed to use nixlTime to measure the time, but I still use the timer class here for the lap() functionality.

aranadive (approved):

benchmark/nixlbench/src/worker/nvshmem/nvshmem_worker.cpp
```diff
@@ -172,30 +186,30 @@ std::variant<double, int> xferBenchNvshmemWorker::transfer(size_t block_size,
         num_iter /= xferBenchConfig::large_blk_iter_ftr;
     }
 
-    ret = execTransfer(local_trans_lists, remote_trans_lists, skip, stream);
+    ret = execTransfer(local_trans_lists, remote_trans_lists, skip, stream, stats);
     if (ret < 0) {
         return std::variant<double, int>(ret);
```
iyastreb: When I build with nvshmem, this line fails, need to use proper type

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -799,3 +810,111 @@ xferBenchUtils::rmObjS3(const std::string &name) {
...
+    samples.insert(samples.end(), other.samples.begin(), other.samples.end());
+}
+
+void
+xferMetricStats::reserve(size_t n) {
+    samples.reserve(n);
+}
+
+void
+xferMetricStats::reset() {
```
ColinNV: Call it `clear()`, too?

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -162,15 +164,78 @@ class xferBenchConfig {
...
+    reserve(size_t n);
+    void
+    reset();
+
+private:
+    std::vector<double> samples;
+};
+
+// Stats class for measuring benchmark metrics
+class xferBenchStats {
```
ColinNV: "It's a `struct`."

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -853,8 +859,13 @@ execTransfer(nixlAgent *agent,
         CHECK_NIXL_ERROR(agent->createXferReq(op, local_desc, remote_desc, target, req, &params),
                          "createTransferReq failed");
 
+        long long prepare_duration = timer.lap();
```
ColinNV: If the type is not machine dependent it's usually better to use a sized type like `std::int64_t`. In this case `std::uint64_t` since that is what `timer.lap()` returns, or rather it's type alias `nixlTime::us_t`. Also, should be `const`.

ovidiusm: Thanks, fixed

benchmark/nixlbench/src/worker/nvshmem/nvshmem_worker.cpp
ovidiusm: nvshmem build fixed

PR #592, opened 2025-07-18
UTILS/SERDES: New serdes.

ColinNV:
## What?

- More compact wire format.
- Provisions for forward/backward medatata compatibility.
- Serdes class based on separate new serializer and deserializer.
- New deserializer does not copy to-be-deserialized data.
- New (de)serializer have improved interfaces.

## Why?
Requirements "backward compatible metadata" and "metadata loading optimizations".

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1056,36 +1056,45 @@ nixlAgent::getNotifs(nixl_notifs_t &notif_map,
 nixl_status_t
 nixlAgent::genNotif(const std::string &remote_agent,
                     const nixl_blob_t &msg,
-                    const nixl_opt_args_t* extra_params) const {
+                    const nixl_opt_args_t *extra_params) const {
 
     backend_list_t backend_list_value;
-    backend_list_t* backend_list;
+    backend_list_t *backend_list;
 
-    NIXL_SHARED_LOCK_GUARD(data->lock);
     if (!extra_params || extra_params->backends.empty()) {
         backend_list = &data->notifEngines;
-        if (backend_list->empty())
-            return NIXL_ERR_BACKEND;
     } else {
         backend_list = &backend_list_value;
-        for (auto &elm : extra_params->backends)
-            if (elm->engine->supportsNotif())
+        for (auto &elm : extra_params->backends) {
+            if (elm->engine->supportsNotif()) {
                 backend_list->push_back(elm->engine);
-
-        if (backend_list->empty()) {
-            return NIXL_ERR_BACKEND;
+            }
         }
     }
 
-    bool localNotif = data->name == remote_agent;
-    for (auto & eng: *backend_list) {
-        if ((localNotif && eng->supportsLocal()) ||
-            (!localNotif &&
-             data->remoteBackends[remote_agent].count(eng->getType()) != 0)) {
-            return eng->genNotif(remote_agent, msg);
-        }
+    if (backend_list->empty()) {
+        return NIXL_ERR_BACKEND;
     }
 
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    if (data->name == remote_agent) {
+        for (const auto &eng : *backend_list) {
+            if (eng->supportsLocal()) {
+                return eng->genNotif(remote_agent, msg);
+            }
+        }
+        return NIXL_ERR_NOT_FOUND;
+    }
+    const auto iter = data->remoteBackends.find(remote_agent);
+
+    if (iter != data->remoteBackends.end()) {
+        for (const auto &eng : *backend_list) {
+            if (iter->second.count(eng->getType()) != 0) {
+                return eng->genNotif(remote_agent, msg);
+            }
+        }
+    }
     return NIXL_ERR_NOT_FOUND;
 }
 


test/unit/utils/serdes/serdes_test.cpp
```diff
@@ -52,14 +52,397 @@ int main() {
...
+        assert(status == NIXL_SUCCESS);
+        assert(get == "555");
+        assert(des.empty());
+    }
+}
+
+} // namespace
 
+int
+main() {
```
iyastreb, line 441: why not gtest?

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,280 @@
...
+
+} // namespace internal
+
+class serializer {
+public:
+    serializer()
+        : serializer(4000) // OK?
+    {}
+
+    explicit serializer(const size_t preAlloc);
```
iyastreb: maybe keep a single ctor with default param value (preAlloc = 4000)

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,280 @@
...
+
+    void
+    addString(const std::string_view &tag, const std::string_view &str) {
```
iyastreb, line 41: Design question.
Here we use TLV serialization, which I like. But does it make sense to keep also 1 byte for TYPE of the entry? Because with current impl it's possible to store string and read it as integer later, right?
Benefits:
- we can make serdes always consistent wrt load/stores, so that client can't misinterpret the value
- enum type {string, int8, int16, int32, int64, binary }
- For plain integers we may even skip Length, because you can take it form the type

This way each type can be an array type (meaning Length is provided), or a primitive type, meaning Length is known from type itself

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,280 @@
...
+    void
+    addRawImpl(const std::string_view &str);
+    void
+    addRawImpl(const void *buf, size_t len);
+
+    template<typename Int>
+    void
+    addIntegralImpl(const Int value) {
+        static_assert(std::is_integral_v<Int> || std::is_enum_v<Int>);
+        if (value < 16) {
```
iyastreb, line 84: IMO this can be solved in a more elegant way with explicit TYPE byte

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,473 @@
...
+serializer::serializer(const size_t preAlloc) {
+    buffer_.reserve(preAlloc);
+    buffer_.append(introString);
 }
 
-std::string nixlSerDes::_bytesToString(const void *buf, ssize_t size) {
-    std::string ret_str = std::string(reinterpret_cast<const char*>(buf), size);
-    return ret_str;
+void
+serializer::addLongLength(size_t len) {
```
iyastreb, line 34: IMO it looks a bit ugly.
Maybe we just always store length as 4 byte integer?
Do we really need to store more elements?
With TYPE enum we don't even need to store a length for plain integers
I believe that CPU overhead of this calculation overweights smaller footprint benefit

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,280 @@
...
+
+    void
+    addBinary(const std::string_view &tag, const void *buf, size_t len) {
+        addRawImpl(tag);
+        addRawImpl(buf, len);
+    }
+
+    template<typename Int>
+    void
+    addIntegral(const std::string_view &tag, const Int value) {
```
iyastreb, line 54: Maybe we also support integer array types?
It seems pretty trivial, maybe worth it

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,473 @@
...
+void
+serializer::addRawImpl(const std::string_view &str) {
+    addRawImpl(str.data(), str.size());
 }
 
-// Strings serialization
-nixl_status_t nixlSerDes::addStr(const std::string &tag, const std::string &str){
+void
+serializer::addRawImpl(const void *buf, const size_t len) {
+    if (len < 128) /* [[likely]] -- C++20 */ {
```
iyastreb, line 53: I'm not sure this microoptimization is worth it
For short data types TYPE enum is superior alternative
For arrays maybe we can tolerate unique size or 4-8 bytes?
Or we could have a dedicated marker in the TYPE enum to specify that array is short (length below 256):
```C
typedef enum {
    TYPE_INT8 = 0, // suitable for strings and binary
    TYPE_INT16,
    TYPE_INT32,
    TYPE_INT64,
} entry_type;

// 2. Define the compact 1-byte structure using bitfields
typedef struct {
    unsigned int type         : 6; // 6 bits for the entry_type (up to 64 types)
    unsigned int is_primitive : 1; // 1 bit for the 'primitive' flag
    unsigned int is_short_array : 1; // 1 bit for the 'short' flag
} TYPE;

```

src/utils/serdes/serdes.h
```diff
@@ -14,41 +14,280 @@
...
+    size() const noexcept {
+        return buffer_.size();
+    }
+
+    [[nodiscard]] const std::string_view &
+    rawBuffer() const noexcept {
+        return buffer_;
+    }
+
+    // MUST (only) be used for values serialized with addString (also addBinary?)
```
iyastreb, line 134: With TYPE the consistency is guaranteed by design -> return error on type mismatch

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,473 @@
...
-nixl_status_t nixlSerDes::addBuf(const std::string &tag, const void* buf, ssize_t len){
+bool
+deserializer::findTag(const std::string_view &tag) {
+    deserializer des(*this);
 
-    workingStr.append(tag);
-    workingStr.append(_bytesToString(&len, sizeof(ssize_t)));
-    workingStr.append(_bytesToString(buf, len));
-    workingStr.append("|");
+    while (!des.empty()) {
```
iyastreb, line 164: findTag seems to be critical deserialization lookup routine, and I see that it does just a linear lookup, each time comparing many strings. IMO we must drastically improve this routine, so that it takes constant time
I can think of several ways achieve that either:
- by storing tags apart from payload, so that all tags are stored at the beginning and have data offset. This way we may load all tags at the des construction, and then access elements in constant time
- cache tags->value offset on first load, later reuse this cache
- cache tags on first lookup lazily
- store index in different format?

test/unit/utils/serdes/serdes_test.cpp
ColinNV: The serdes test that I found didn't use it (and I didn't need more than `assert()` for these tests.)

src/utils/serdes/serdes.h
ColinNV: The "store as string, retrieve as integer" (and vice versa) was a "feature" of the original implementation and I kept it in the name of strict backwards compatibility. (And because `nixlSerDes` does not offer dedicated functions for integers.) If that's not necessary we can do something like you suggest, possibly going beyond by omitting the original size for the "compressed" integers.

src/utils/serdes/serdes.h
ColinNV: If that's a better trade-off for the distribution of values we encounter in practice, sure.

src/utils/serdes/serdes.cpp
ColinNV: My impression was that most (all?) tag strings are shorter than 127 which is why I chose to optimize that case with the "short" length embedded in the leading byte. For the values, do we know which distribution of lengths to optimize for?

src/utils/serdes/serdes.h
ColinNV: Yes, it's easy to add convenience for other types, what exactly do you suggest, `std::array` and `std::vector` of integer types?

src/utils/serdes/serdes.cpp
ColinNV: It does a linear lookup starting from the beginning of the string view, **but** every time a tag-value-pair is deserialized it, and everything before it, are removed from the beginning of the string view!

test/unit/utils/serdes/serdes_test.cpp
iyastreb: gtest is a framework that provides much more than just assertions
It allows you to filter tests, produce output in many formats, integrate nicely with CI, run tests in a loop and many more. I propose to stick to gtest, so that we use a unique test engine across C++ code

src/utils/serdes/serdes.h
iyastreb: My personal preference is to use simplest form: pointer + size
std::vector means that we need to construct an object on deserialization, which we don't want to do
We can just return starting pointer + size

src/utils/serdes/serdes.cpp
iyastreb: hmm, so deserialization is "destructive"?
Does it mean that I can't call `des.getString("aaa")` multiple times? Because first time it returns a value, and second time it returns nullptr?
If so, I find it too restrictive, I believe we can do better

src/utils/serdes/serdes.cpp
iyastreb: Honestly I have no idea, looking in the code I see that we mostly use short strings.
Anyway, with TYPE structure we can have this optimization for free (for arrays shorter than 256 elements)

test/unit/utils/serdes/serdes_test.cpp
ColinNV: I can update the test to use gtest.

src/utils/serdes/serdes.cpp
ColinNV: I'd be very happy to switch to a more strict typing of the serialized data, the problem is the strict backward compatibility of the API.

src/utils/serdes/serdes.h
ColinNV: I was thinking of the input/serializing part where convenience overloads for other types make the interface less error prone.

src/utils/serdes/serdes.cpp
ColinNV: There are 100 ways of doing this, we should first nail down the compatibility issues before possibly changing the encoding.

src/utils/serdes/serdes.h
ColinNV: See above comment on strict typing.

src/utils/serdes/serdes.cpp
ColinNV: That is correct, it can't be called multiple times, just like it can't at the moment; changing this would break API behaviour.

PR #594, opened 2025-07-20
refactor: replace bool with XferStatus enum in get_xfer_status method

### LINKED ISSUE #447: Modify get_xfer_status response to use enum instead of a bool
open issue to change the nixl-sys api to avoid using a bool and use a more intuitive enum instead.

_Originally posted by @ryanolson in https://github.com/ai-dynamo/dynamo/pull/1363#discussion_r2136500011_

da-x:
Fixes #447

### DIFF
--- a/src/bindings/rust/src/agent.rs
+++ b/src/bindings/rust/src/agent.rs
@@ -21,6 +21,18 @@ pub struct Agent {
     inner: Arc<RwLock<AgentInner>>,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum XferStatus {
+    Success,
+    InProgress,
+}
+
+impl XferStatus {
+    pub fn is_success(&self) -> bool {
+        return *self == XferStatus::Success;
+    }
+}
+
 impl Agent {
     /// Creates a new agent with the given name
     pub fn new(name: &str) -> Result<Self, NixlError> {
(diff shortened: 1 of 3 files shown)


src/bindings/rust/src/agent.rs
```diff
@@ -21,6 +21,18 @@ pub struct Agent {
     inner: Arc<RwLock<AgentInner>>,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum XferStatus {
+    Success,
+    InProgress,
+}
```
roiedanino, line 29: Wouldn't be better to just add all nixl statuses?

src/bindings/rust/src/agent.rs
```diff
@@ -21,6 +21,18 @@ pub struct Agent {
     inner: Arc<RwLock<AgentInner>>,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum XferStatus {
```
roiedanino, line 26: Maybe it would be better to name it `NixlXferStatus`?

src/bindings/rust/src/agent.rs
da-x: `nixl_capi_status_t` only has `NIXL_CAPI_IN_PROG` and `NIXL_CAPI_SUCCESS` as non-errors, and so does `nixl_status_t` in the C++ side that only has `NIXL_IN_PROG` and `NIXL_SUCCESS` as non-errors. The rust-side `NixlError` takes cares of the other values.

src/bindings/rust/src/agent.rs
da-x: It's more consistent this way - the other publicly exported types such as `MemType` and `XferOp` do not contain `Nixl` as prefix.

roiedanino (approved):

yosefe (approved):

PR #599, opened 2025-07-21
obj: Add request checksum config toggle

w1ldptr:
## What?
Allow users to configure AWS Request checksum setting.

## Why?
Some users encountered issues with using Obj plugin with endpoints that don't support STREAMING-UNSIGNED-PAYLOAD-TRAILER.

## How?
Via plugin custom params.

Fixes NIX-205


jgoldsch12:
Changes look good to me.  Thank you.

barneuman (approved):

PR #600, opened 2025-07-21
Fix issue in registerMem for OBJ plugin

mamar123:
## What?
Fixed a bug in the registerMem function of the OBJ plugin.

## Why?
The previous implementation didn’t handle the metadata pointer correctly for memory types other than OBJ, potentially leaving it uninitialized. This could result in a segmentation fault during nixlObjEngine::deregisterMem when attempting to access it.

## How?
Explicitly set metadata output to nullptr for non-OBJ memory types to avoid undefined behavior.


barneuman:
please add what&why&how

mamar123:
> please add what&why&how

@barneuman  Done

barneuman:
/ok to test [635d5dc](https://github.com/ai-dynamo/nixl/pull/600/commits/635d5dc8de22ba1cd5ddcc5ad670f83a1951f821)

PR #605, opened 2025-07-22
Build 3fs

dfarge:
## What?
Installs 3fs's dependencies and installs it in .gitlab/build.sh.

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -99,11 +99,12 @@ fi
 echo "Detected OS version: $OS_VERSION"
 
 INSTALL_3FS=false
-if [ "$OS_VERSION" = "22.04" ]; then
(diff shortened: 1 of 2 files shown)


.gitlab/build.sh
```diff
@@ -83,7 +82,446 @@ $SUDO apt-get -qq install -y curl \
...
-                             libcurl4-openssl-dev zlib1g-dev # aws-sdk-cpp dependencies
+                             libclang-dev \
+                             libgflags-dev \
+                             libcpprest-dev \
+                             etcd-client \
+                             libcurl4-openssl-dev \
+                             zlib1g-dev
+
+# Detect OS version using /etc/os-release
+OS_VERSION="unknown"
```
aranadive, line 93: It would be better to create a separate build_3fs.sh script and call it here.

PR #606, opened 2025-07-22
NIXL UCX progress thread refactoring

iyastreb:
## What?
Split nixlUcxThreadEngine into a separate subclass of nixlUcxEngine.

## Why?
This is a preparatory step to implement threadpool engine (https://github.com/ai-dynamo/nixl/pull/573).
The idea is to implement progress thread functionality in a separate subclass to split the concerns and follow single responsibility principle. With threadpool approach we don't need to use progress thread in it's current form, so it's desirable to keep them separate.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -566,9 +566,7 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     if (custom_params->count("device_list")!=0)
         devs = str_split((*custom_params)["device_list"], ", ");
 
-    const auto num_workers_iter = custom_params->find("num_workers");
-    if (num_workers_iter == custom_params->end() || !absl::SimpleAtoi(num_workers_iter->second, &numWorkers))
-        numWorkers = 1;
+    numWorkers = init_params.getOrDefault("num_workers", 1);
 
     ucp_err_handling_mode_t err_handling_mode;
     const auto err_handling_mode_it =
@@ -924,6 +922,8 @@ static nixl_status_t _retHelper(nixl_status_t ret,  nixlUcxBackendH *hndl, nixlU
     /* if transfer wasn't immediately completed */
     switch(ret) {
         case NIXL_IN_PROG:
+            // TODO: this cast does not look safe
+            // We need to allocate a vector of nixlUcxIntReq and set nixlUcxReqt
             hndl->append((nixlUcxIntReq*)req);
         case NIXL_SUCCESS:
             // Nothing to do
@@ -943,6 +943,11 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        nixlBackendReqH* &handle,
                                        const nixl_opt_b_args_t* opt_args) const
 {
+    if (local.descCount() == 0 || remote.descCount() == 0) {
+        NIXL_ERROR << "Local or remote descriptor list is empty";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
     /* TODO: try to get from a pool first */
     size_t workerId = getWorkerId();
     handle = new nixlUcxBackendH(getWorker(workerId).get(), workerId);
@@ -1023,17 +1028,22 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
     size_t workerId = intHandle->getWorkerId();
 
     if (lcnt != rcnt) {
+        NIXL_ERROR << "Local (" << lcnt << ") and remote (" << rcnt
+                   << ") descriptor lists differ in size";
         return NIXL_ERR_INVALID_PARAM;
     }
+
+    // TODO: assert that handle is empty/completed, as we can't post request before completion
 
     for(i = 0; i < lcnt; i++) {
         void *laddr = (void*) local[i].addr;
         size_t lsize = local[i].len;
-        void *raddr = (void*) remote[i].addr;
+        uint64_t raddr = (uint64_t) remote[i].addr;
         size_t rsize = remote[i].len;
 
         lmd = (nixlUcxPrivateMetadata*) local[i].metadataP;
         rmd = (nixlUcxPublicMetadata*) remote[i].metadataP;
+        auto &ep = rmd->conn->getEp(workerId);
 
         if (lsize != rsize) {
             return NIXL_ERR_INVALID_PARAM;
@@ -1041,10 +1051,10 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
 
         switch (operation) {
         case NIXL_READ:
-            ret = rmd->conn->getEp(workerId)->read((uint64_t) raddr, rmd->getRkey(workerId), laddr, lmd->mem, lsize, req);
+            ret = ep->read(raddr, rmd->getRkey(workerId), laddr, lmd->mem, lsize, req);
             break;
         case NIXL_WRITE:
-            ret = rmd->conn->getEp(workerId)->write(laddr, lmd->mem, (uint64_t) raddr, rmd->getRkey(workerId), lsize, req);
+            ret = ep->write(laddr, lmd->mem, raddr, rmd->getRkey(workerId), lsize, req);
             break;
         default:
             return NIXL_ERR_INVALID_PARAM;
@@ -1058,6 +1068,7 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
     /*
      * Flush keeps intHandle non-empty until the operation is actually
      * completed, which can happen after local requests completion.
+     * TODO: should we flush all distinct endpoints?
      */
     rmd = (nixlUcxPublicMetadata*) remote[0].metadataP;
     ret = rmd->conn->getEp(workerId)->flushEp(req);
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -313,17 +313,19 @@ static void _internalRequestReset(nixlUcxIntReq *req) {
 *****************************************/
 
 class nixlUcxBackendH : public nixlBackendReqH {
-private:
+protected:
+    // TODO: use std::vector here for a single allocation and cache friendly
+    // traversal
     nixlUcxIntReq head;
-    const nixlUcxEngine &eng;
+    nixlUcxWorker *worker;
     size_t worker_id;
 
     // Notification to be sent after completion of all requests
     struct Notif {
-	    std::string agent;
-	    nixl_blob_t payload;
-	    Notif(const std::string& remote_agent, const nixl_blob_t& msg)
-		    : agent(remote_agent), payload(msg) {}
+        std::string agent;
+        nixl_blob_t payload;
+        Notif(const std::string& remote_agent, const nixl_blob_t& msg) :
+            agent(remote_agent), payload(msg) {}
     };
     std::optional<Notif> notif;
 
@@ -332,7 +334,8 @@ public:
         return notif;
     }
 
-    nixlUcxBackendH(const nixlUcxEngine &eng_, size_t worker_id_): eng(eng_), worker_id(worker_id_) {}
+    nixlUcxBackendH(nixlUcxWorker *worker_, size_t worker_id_) :
+        worker(worker_), worker_id(worker_id_) {}
 
     void append(nixlUcxIntReq *req) {
         head.link(req);
@@ -346,23 +349,21 @@ public:
             return NIXL_SUCCESS;
         }
 
-        const auto &uw = eng.getWorker(worker_id);
         // TODO: Error log: uncompleted requests found! Cancelling ...
-        while(req) {
+        while (req) {
             nixlUcxIntReq *cur = req;
             bool done = cur->is_complete();
             req = cur->unlink();
             if (!done) {
                 // TODO: Need process this properly.
                 // it may not be enough to cancel UCX request
-                uw->reqCancel((nixlUcxReq)cur);
+                worker->reqCancel((nixlUcxReq)cur);
             }
             _internalRequestReset(cur);
-            uw->reqRelease((nixlUcxReq)cur);
+            worker->reqRelease((nixlUcxReq)cur);
         }
         return NIXL_SUCCESS;
     }
-
 
     nixl_status_t status()
     {
@@ -374,13 +375,11 @@ public:
             return NIXL_SUCCESS;
         }
 
-        const auto &uw = eng.getWorker(worker_id);
-
         /* Maximum progress */
-        while (uw->progress());
+        while (worker->progress());
 
         /* Go over all request updating their status */
-        while(req) {
+        while (req) {
             nixl_status_t ret;
             if (!req->is_complete()) {
                 ret = ucx_status_to_nixl(ucp_request_check_status((nixlUcxReq)req));
@@ -403,11 +402,11 @@ public:
         /* Remove completed requests keeping the first one as
         request representative */
         req = head.unlink();
-        while(req) {
+        while (req) {
             nixlUcxIntReq *next_req = req->unlink();
             if (req->is_complete()) {
                 _internalRequestReset(req);
-                uw->reqRelease((nixlUcxReq)req);
+                worker->reqRelease((nixlUcxReq)req);
             } else {
                 /* Enqueue back */
                 append(req);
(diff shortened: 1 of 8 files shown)


src/plugins/ucx/ucx_backend.h
```diff
@@ -258,10 +243,45 @@ class nixlUcxEngine
...
         }
+};
 
-        size_t getWorkerId() const {
-            return std::hash<std::thread::id>{}(std::this_thread::get_id()) % uws.size();
+/**
+ * Represents an engine with a single progress thread for all shared workers
+ * TODO: can it be merged with nixlUcxThreadPoolEngine?
+ */
+class nixlUcxThreadEngine : public nixlUcxEngine {
```
rakhmets, line 309: I suggest to put the class declaration into a separate header, and the definition into a separate source file.

src/plugins/ucx/ucx_backend.h
```diff
@@ -258,10 +243,45 @@ class nixlUcxEngine
...
+        bool isProgressThread() const noexcept {
+            return std::this_thread::get_id() == pthr.get_id();
         }
+
+        int vramApplyCtx() override;
+        void appendNotif(std::string remote_name, std::string msg) override;
+        void notifProgress();
+        void notifProgressCombineHelper(notif_list_t &src, notif_list_t &tgt);
+
+    public:
```
rakhmets: Please put public section first.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,11 +426,44 @@ class nixlUcxBackendH : public nixlBackendReqH {
  * Progress thread management
 *****************************************/
 
-void nixlUcxEngine::progressFunc()
+nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
+    : nixlUcxEngine(init_params) {
+    if (!nixlUcxMtLevelIsSupported(nixl_ucx_mt_t::WORKER)) {
+        NIXL_ERROR << "UCX library does not support multi-threading";
+        this->initErr = true;
```
rakhmets: Please take into account this PR https://github.com/ai-dynamo/nixl/pull/537.
We try to get rid of using of `initErr`, and throw/catch exceptions instead.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -483,81 +515,94 @@ void nixlUcxEngine::progressFunc()
...
+nixlUcxEngine::create(const nixlBackendInitParams &init_params)
+{
+    nixlUcxEngine *engine;
+    if (init_params.enableProgTh) {
+        engine = new nixlUcxThreadEngine(init_params);
     } else {
-        pthrOn = false;
+        engine = new nixlUcxEngine(init_params);
     }
+    return std::unique_ptr<nixlUcxEngine>(engine);
```
rakhmets, line 632:
```C++
if (init_params.enableProgTh) {
    return std::make_unique<nixlUcxThreadEngine>(init_params);
}
return std::make_unique<nixlUcxEngine>(init_params);

File: src/plugins/ucx/ucx_backend.cpp [hunk 5]
```diff
@@ -1148,6 +1186,11 @@ nixl_status_t nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
     return ret;
 }
 
+void nixlUcxEngine::appendNotif(std::string remote_name, std::string msg)
+{
+    notifMainList.push_back(std::make_pair(std::move(remote_name), std::move(msg)));
```
rakhmets: emplace_back?

src/plugins/ucx/ucx_backend.h
```diff
@@ -105,22 +105,12 @@ using nixlUcxCudaDevicePrimaryCtxPtr = std::shared_ptr<nixlUcxCudaDevicePrimaryC
 
 class nixlUcxEngine
     : public nixlBackendEngine {
-    private:
+    protected:
```
rakhmets: It doesn't look like an optimal solution. Why do we need to give inheritors access to all private class members?

src/plugins/ucx/ucx_plugin.cpp
```diff
@@ -26,12 +26,7 @@ namespace
    const char* ucx_plugin_version = "0.1.0";
 
    [[nodiscard]] nixlBackendEngine* create_ucx_engine(const nixlBackendInitParams* init_params) {
-        try {
-            return new nixlUcxEngine(init_params);
-        } catch (const std::exception &e) {
-            NIXL_ERROR << "Failed to create UCX engine: " << e.what();
-            return nullptr;
-        }
+       return nixlUcxEngine::create(*init_params).release();
```
rakhmets: We have to catch exceptions here.

src/plugins/ucx/ucx_backend.h
```diff
@@ -105,92 +105,98 @@ using nixlUcxCudaDevicePrimaryCtxPtr = std::shared_ptr<nixlUcxCudaDevicePrimaryC
...
+    getWorkerId() const {
+        return std::hash<std::thread::id>{}(std::this_thread::get_id()) % uws.size();
+    }
 
     public:
-        nixlUcxEngine(const nixlBackendInitParams* init_params);
+        static std::unique_ptr<nixlUcxEngine>
+        create(const nixlBackendInitParams &init_params);
+
+        nixlUcxEngine(const nixlBackendInitParams &init_params);
```
rakhmets: Now we have two ways to create the engine. Maybe add protected struct `Accessor`?
```C++
class nixlUcxEngine {
public:
    nixlUcxEngine(const nixlBackendInitParams &, Accessor);
    static std::unique_ptr<nixlUcxEngine> create(const nixlBackendInitParams &);
protected:
    struct Accessor {};
    ...
};
```

iyastreb: Right, I better make ctor protected

src/plugins/ucx/ucx_backend.h
iyastreb: sure

src/plugins/ucx/ucx_backend.h
iyastreb: I think no need, IMO this class should disappear once we introduce nixlUcxThreadPoolEngine (https://github.com/ai-dynamo/nixl/pull/573)

src/plugins/ucx/ucx_backend.cpp
iyastreb: Thanks for spotting this, my bad - I rebased on older version without this commit, and overlooked it

src/plugins/ucx/ucx_backend.cpp
iyastreb: Since I made ctor protected, I can't construct new entry with make_unique

src/plugins/ucx/ucx_plugin.cpp
iyastreb: ok

iyastreb: ok

src/plugins/ucx/ucx_backend.h
iyastreb: let me try to make private as much as I can
But subclass IS-A base class, so in principle I don't see a big issue if it has access to some fields.

src/plugins/ucx/ucx_backend.cpp
rakhmets: Yes, so I suggested to make ctor public, but disable it from the usage via protected structure:
```C++
class nixlUcxEngine {
public:
    nixlUcxEngine(const nixlBackendInitParams &, Accessor);
    static std::unique_ptr<nixlUcxEngine> create(const nixlBackendInitParams &);
protected:
    struct Accessor {};
    ...
};
```

The same approach is used in the example in this topic https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c50-use-a-factory-function-if-you-need-virtual-behavior-during-initialization

src/plugins/ucx/ucx_backend.h
rakhmets: I prefer to avoid using `protected` data for the reasons described here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data. But for well-known design patterns that use `protected` data, e.g. decorator.

I think it's ok to have some protected fields, considering that this class will disappear with https://github.com/ai-dynamo/nixl/pull/573. But it would be better if you move at least some of them to private scope.

iyastreb: I put all of the fields to private section, keeping only some methods in protected

mkhazraee (dismissed):

src/plugins/ucx/ucx_backend.h
```diff
@@ -103,165 +103,233 @@ class nixlUcxCudaCtx;
...
+        return true;
+    }
+
+    bool
+    supportsNotif() const override {
+        return true;
+    }
+
+    bool
+    supportsProgTh() const override {
```
brminich, line 129: is it possible to avoid unrelated changes to keep PR smaller?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -932,10 +1001,14 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
...
-    /* TODO: try to get from a pool first */
-    nixlUcxBackendH *intHandle = new nixlUcxBackendH(*this, getWorkerId());
+    if (local.descCount() == 0 || remote.descCount() == 0) {
+        NIXL_ERROR << "Local or remote descriptor list is empty";
+        return NIXL_ERR_INVALID_PARAM;
+    }
 
-    handle = (nixlBackendReqH*)intHandle;
+    /* TODO: try to get from a pool first */
+    size_t workerId = getWorkerId();
```
Pyngon: getWorkerId() could return both dedicated worker and shared worker, so dedicated worker could be used by more than 1 thread(user threads + dedicated thread from thread pool).

Need to set the mt_type of the worker to nixl_ucx_mt_t::WORKER or make sure only 1 thread will use the dedicated worker.

Pyngon: ops, wrong PR. This comment is for the #573 UCX threadpool

iyastreb: Yes, let me respond there

src/api/cpp/nixl_types.h
```diff
@@ -21,7 +21,6 @@
 #include <unordered_map>
 #include <optional>
 
-
```
brminich: pls revert - this is a single unrelated change in this file

src/utils/ucx/ucx_utils.h
```diff
@@ -57,6 +59,23 @@ template<typename Enum>
...
+        return defaultValue;
+    }
+
+    auto it = customParams->find(key);
+    if (it == customParams->end()) {
+        return defaultValue;
+    }
+
+    T result;
+    if constexpr (std::is_same_v<T, int>)
```
brminich: do we really need a template here then?

test/gtest/test_transfer.cpp
```diff
@@ -245,28 +260,6 @@ class TestTransfer : public testing::TestWithParam<std::string> {
         }
     }
 
-    void waitForXfer(nixlAgent &from, const std::string &from_name,
```
brminich: pls remove in a separate PR if it is unused

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -427,137 +431,219 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-    const char signal = 'X';
-    int ret = write(pthrControlPipe[1], &signal, sizeof(signal));
-    if (ret < 0)
-        NIXL_PERROR << "write to progress thread control pipe failed";
-    pthr.join();
+    // This will ensure that the resulting delay is at least 1ms and fits into int in order for
+    // it to be compatible with poll()
+    auto pthrDelay = std::chrono::ceil<std::chrono::milliseconds>(std::chrono::microseconds(
+        init_params.pthrDelay < std::numeric_limits<int>::max() ? init_params.pthrDelay :
+                                                                  std::numeric_limits<int>::max()));
```
brminich: minor:
can split for better readability

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -932,10 +1001,14 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        nixlBackendReqH* &handle,
                                        const nixl_opt_b_args_t* opt_args) const
 {
-    /* TODO: try to get from a pool first */
-    nixlUcxBackendH *intHandle = new nixlUcxBackendH(*this, getWorkerId());
+    if (local.descCount() == 0 || remote.descCount() == 0) {
+        NIXL_ERROR << "Local or remote descriptor list is empty";
+        return NIXL_ERR_INVALID_PARAM;
+    }
```
brminich, line 1004: is it unrelated fix?

test/gtest/test_transfer.cpp
iyastreb: Ok, separate PR:
https://github.com/ai-dynamo/nixl/pull/652

src/plugins/ucx/ucx_backend.cpp
iyastreb: I'm not sure it brings much value, I would prefer to keep it simpler, it's not a public API anyway

src/plugins/ucx/ucx_backend.cpp
iyastreb: Well, it's related, because I use descCount to calculate the split size and need to make sure it's more than 0

src/plugins/ucx/ucx_backend.h
iyastreb: I didn't want to bring all these changes, but clang formatter gives errors and requires these changes. This is because this class was mis-formatted and when I touched a few places, clang-format requires to format lines around my change and that leads to reformatting of the entire class. I didn't manage to workaround that..

src/utils/ucx/ucx_utils.h
iyastreb: Yes, for example we can reuse this method to handle `ucp_err_handling_mode_t`, which is handled this way:
```
    ucp_err_handling_mode_t err_handling_mode;
    const auto err_handling_mode_it =
        custom_params->find(std::string(nixl_ucx_err_handling_param_name));
    if (err_handling_mode_it == custom_params->end()) {
        err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
    } else {
        err_handling_mode = ucx_err_mode_from_string(err_handling_mode_it->second);
    }
```
after corresponding change it would be:
```
err_handling_mode = nixl_b_params_get(custom_params, nixl_ucx_err_handling_param_name, UCP_ERR_HANDLING_MODE_PEER);
```
Could be also useful for strings (`T=const char *`)

src/plugins/ucx/ucx_backend.cpp
iyastreb: done

src/utils/ucx/ucx_utils.h
brminich: would not it fail this check on line 75?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -313,17 +314,21 @@ static void _internalRequestReset(nixlUcxIntReq *req) {
...
-	    std::string agent;
-	    nixl_blob_t payload;
-	    Notif(const std::string& remote_agent, const nixl_blob_t& msg)
-		    : agent(remote_agent), payload(msg) {}
+        std::string agent;
+        nixl_blob_t payload;
+
+        Notif(const std::string &remote_agent, const nixl_blob_t &msg)
+            : agent(remote_agent),
+              payload(msg) {}
```
brminich: seems no need to change formatting as nothing new is added here

brminich (dismissed):
👍 besides minor comments

src/utils/ucx/ucx_utils.h
iyastreb: Currently it handles only int.
With each new type supported we need to extend this function as well, e.g. for string we would add:
```
    if constexpr (std::is_same_v<T, int>)
        return absl::SimpleAtoi(it->second, &result) ? result : defaultValue;
    if constexpr (std::is_same_v<T, const char*>)
        return it->second.data();
```

src/plugins/ucx/ucx_backend.cpp
iyastreb: There was no real need, but it was the only place in the code where TABs were used for indentation. Reverted

iyastreb:
From Vlad Buslov:
a lot of code doesn't conform to code style guide. You check your commits by clang-tidy for this, if you like:
`git diff -U0 HEAD^1 HEAD | clang-tidy-diff-19.py -clang-tidy-binary=clang-tidy-19  -p1  -path=.`

iyastreb:
> From Vlad Buslov: a lot of code doesn't conform to code style guide. You check your commits by clang-tidy for this, if you like: `git diff -U0 HEAD^1 HEAD | clang-tidy-diff-19.py -clang-tidy-binary=clang-tidy-19 -p1 -path=.`

Addressed

PR #607, opened 2025-07-22
Nixlbench obj

glevnv:
## What?
Add OBJ plugin support for Nixlbench

## Why?
to test new added plugin

## How?
Updates to Nixlbench Dockerfile
Nixlbench support for OBJ plugin
AWS CLI commands to verify S3 objects integrity


barneuman:
/ok to test [3105b48](https://github.com/ai-dynamo/nixl/pull/607/commits/3105b4850460d013bdb1692e00ff7ec2adfeaddd)

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -580,3 +655,130 @@ void xferBenchUtils::printStats(bool is_target, size_t block_size, size_t batch_
...
+    if (!xferBenchConfig::obj_region.empty()) {
+        env_setup += "AWS_DEFAULT_REGION=" + xferBenchConfig::obj_region + " ";
+    }
+
+    return env_setup;
+}
+
+bool
+xferBenchUtils::putObjS3(size_t buffer_size, const std::string &name) {
+    std::string filename = std::filesystem::current_path().string() + "/" + name;
```
ovidiusm: Could we use a temp file in /tmp instead?

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -580,3 +655,130 @@ void xferBenchUtils::printStats(bool is_target, size_t block_size, size_t batch_
...
+    std::string aws_cmd = "aws s3 cp " + name + " s3://" + bucket_name;
+    if (!xferBenchConfig::obj_endpoint_override.empty()) {
+        aws_cmd += " --endpoint-url " + xferBenchConfig::obj_endpoint_override;
+    }
+
+    std::string full_cmd = buildAwsCredentials() + aws_cmd;
+    std::cout << "Putting S3 object: " << name << " in bucket: " << bucket_name
+              << " (size: " << buffer_size << " bytes)" << std::endl;
+
+    int result = system(full_cmd.c_str());
```
ovidiusm, line 736: system() on manually crafted string is a bit fragile, but I am not strongly against it if you think there is no risk

ovidiusm:
I have a few questions related on how this should be benchmarked and tested.

How does dynamo store the data in S3 in practice? Small chunks? Or some compacted format? Is the format the same as for the other storage plugins?

How would one choose some values to be tested (buffer sizes, number of buffers)? Are the sizes the same as for the other plugins?

Should we also add support in kvbench plan and profile commands for this backend?

Do we need to cover this in verification? Do they have a test account? What should they expect from the test in terms of success and performance?

benchmark/nixlbench/src/utils/utils.cpp
glevnv: the command is simple and I don't see a failure point other than the filepath and the credentials (which would be used in other solutions as well)

glevnv:
> I have a few questions related on how this should be benchmarked and tested.
> 
> How does dynamo store the data in S3 in practice? Small chunks? Or some compacted format? Is the format the same as for the other storage plugins?
> 
> How would one choose some values to be tested (buffer sizes, number of buffers)? Are the sizes the same as for the other plugins?
> 
> Should we also add support in kvbench plan and profile commands for this backend?
> 
> Do we need to cover this in verification? Do they have a test account? What should they expect from the test in terms of success and performance?

@vvenkates27 could you shed some light?

barneuman:
/ok to test [80fed9c](https://github.com/ai-dynamo/nixl/pull/607/commits/80fed9cd541893f37b86037602cb48fd987a8af7)

ovidiusm (dismissed):

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -100,6 +102,16 @@ DEFINE_string (posix_api_type,
...
+// OBJ options - only used when backend is OBJ
+DEFINE_string(obj_access_key, "", "Access key for S3 backend");
+DEFINE_string(obj_secret_key, "", "Secret key for S3 backend");
+DEFINE_string(obj_session_token, "", "Session token for S3 backend");
+DEFINE_string(obj_bucket_name, XFERBENCH_OBJ_BUCKET_NAME_DEFAULT, "Bucket name for S3 backend");
+DEFINE_string(obj_scheme, XFERBENCH_OBJ_SCHEME_HTTP, "HTTP scheme for S3 backend [http, https]");
+DEFINE_string(obj_region, XFERBENCH_OBJ_REGION_EU_CENTRAL_1, "Region for S3 backend");
+DEFINE_bool(obj_use_virtual_addressing, false, "Use virtual addressing for S3 backend");
+DEFINE_string(obj_endpoint_override, "", "Endpoint override for S3 backend");
+
```
jgoldsch12, line 117: There's a new option added to the OBJ plugin called req_checksum, which is optional and the two possible values are 'supported' and 'required'

barneuman:
/ok to test [c77e5d6](https://github.com/ai-dynamo/nixl/pull/607/commits/c77e5d68f98d3af252b47e8ffd3c2a65d5a49ea4)

aranadive:
> > I have a few questions related on how this should be benchmarked and tested.
> > How does dynamo store the data in S3 in practice? Small chunks? Or some compacted format? Is the format the same as for the other storage plugins?
> > How would one choose some values to be tested (buffer sizes, number of buffers)? Are the sizes the same as for the other plugins?
> > Should we also add support in kvbench plan and profile commands for this backend?
> > Do we need to cover this in verification? Do they have a test account? What should they expect from the test in terms of success and performance?
> 
> @vvenkates27 could you shed some light?

Yes kvbench support would be good but as a separate PR. We can discuss the other items offline.

@w1ldptr can address the verification and account info questions.

barneuman:
/ok to test [1146dbb](https://github.com/ai-dynamo/nixl/pull/607/commits/1146dbbcd3cd77316eb46d7d4d1bfc4769a948be)

aranadive (approved):
@glevnv Just curious why the aws cli was used directly vs the SDK API to access buckets?

glevnv:
> @glevnv Just curious why the aws cli was used directly vs the SDK API to access buckets?

The interface nixlbench needs with AWS is simple (just: get, put, remove). So it seemed to me easier and cleaner to use the AWS CLI rather than to import AWS SDK, instantiate an S3 client, store it somewhere and maintain it.

vvenkates27:
> > @glevnv Just curious why the aws cli was used directly vs the SDK API to access buckets?
> 
> 
> 
> The interface nixlbench needs with AWS is simple (just: get, put, remove). So it seemed to me easier and cleaner to use the AWS CLI rather than to import AWS SDK, instantiate an S3 client, store it somewhere and maintain it.

Bucket creation should be handled using nixl backend parameter, there shouldn't be a need to make awss3 client calls from nixlbench, am I missing something?

glevnv:
> > > @glevnv Just curious why the aws cli was used directly vs the SDK API to access buckets?
> > 
> > 
> > The interface nixlbench needs with AWS is simple (just: get, put, remove). So it seemed to me easier and cleaner to use the AWS CLI rather than to import AWS SDK, instantiate an S3 client, store it somewhere and maintain it.
> 
> Bucket creation should be handled using nixl backend parameter, there shouldn't be a need to make awss3 client calls from nixlbench, am I missing something?

There is a need for another, independent, AWS client in a few places. Before a test of the read op - nixlbench needs to put the object to AWS, for cleaning the uploaded files and for data consistency checks - nixlbench needs to get the uploaded file

PR #611, opened 2025-07-23
Allow UCX as backend in kvbench plan and profile commands

ovidiusm:
## What?
Allow running networking benchmarks as well directly from plan and profile commands.


benchmark/kvbench/commands/nixlbench.py
```diff
@@ -158,11 +158,28 @@ def _configure_posix(self, source: str, destination: str):
...
+            self.target_seg_type = "VRAM"
+        else:
+            raise ValueError(f"Invalid source for UCX: {source}")
+
+        if destination == "memory":
+            self.target_seg_type = "DRAM"
+        elif destination == "gpu":
+            self.initiator_seg_type = "VRAM"
+        else:
+            raise ValueError(f"Invalid destination for UCX: {destination}")
```
yosefe: 1. shall we also set the op_type?
2. seens there is mixup, we check source==gpu and then set the self.target_seg_type?
3. can we reduce code duplication of source/target?

ovidiusm: Thanks, done.

There is a separate --op_type parameter that already works.

brminich (approved):

PR #620, opened 2025-07-25
RUST: Implement queryMem binding

w1ldptr:


src/bindings/rust/src/descriptors/query.rs
```diff
@@ -0,0 +1,172 @@
...
+        let size = self.len()?;
+        if index >= size {
+            return Err(NixlError::InvalidParam);
+        }
+
+        Ok(QueryResponse { list: self, index })
+    }
+
+    /// Iterates over all responses in the list
+    pub fn iter(&self) -> Result<QueryResponseIterator<'_>, NixlError> {
```
roiedanino:
```suggestion
    /// Returns an iterator
    pub fn iter(&self) -> Result<QueryResponseIterator<'_>, NixlError> {
```

src/bindings/rust/wrapper.cpp
```diff
@@ -1715,4 +1727,111 @@ nixl_capi_notif_map_clear(nixl_capi_notif_map_t map)
...
+
+    try {
+        nixl_opt_args_t *args = opt_args ? &opt_args->args : nullptr;
+        nixl_status_t ret = agent->inner->queryMem(*descs->dlist, resp->responses, args);
+        return ret == NIXL_SUCCESS ? NIXL_CAPI_SUCCESS : NIXL_CAPI_ERROR_BACKEND;
+    }
+    catch (...) {
+        return NIXL_CAPI_ERROR_BACKEND;
+    }
+}
```
roiedanino, line 1835: Need to add empty implementations in stubs.cpp as well

src/bindings/rust/src/lib.rs
```diff
@@ -57,7 +57,9 @@ use bindings::{
     nixl_capi_opt_args_get_skip_desc_merge, nixl_capi_opt_args_set_has_notif,
     nixl_capi_opt_args_set_notif_msg, nixl_capi_opt_args_set_skip_desc_merge,
     nixl_capi_params_create_iterator, nixl_capi_params_destroy_iterator, nixl_capi_params_is_empty,
-    nixl_capi_params_iterator_next, nixl_capi_post_xfer_req, nixl_capi_reg_dlist_add_desc,
+    nixl_capi_params_iterator_next, nixl_capi_post_xfer_req, nixl_capi_query_mem,
+    nixl_capi_create_query_resp_list, nixl_capi_destroy_query_resp_list, nixl_capi_query_resp_list_size,
+    nixl_capi_query_resp_list_has_value, nixl_capi_query_resp_list_get_params, nixl_capi_reg_dlist_add_desc,
```
roiedanino: Can you please add new C APIs to the end of the list so it would be easier to fix conflicts?

src/bindings/rust/src/agent.rs
```diff
@@ -243,6 +244,38 @@ impl Agent {
...
+        let resp = QueryResponseList::new()?;
+
+        let status = unsafe {
+            nixl_capi_query_mem(
+                self.inner.write().unwrap().handle.as_ptr(),
+                descs.handle(),
+                resp.handle(),
+                opt_args.map_or(std::ptr::null_mut(), |args| args.inner.as_ptr()),
+            )
+        };
```
roiedanino: write() lock should be created outside the scope to prevent race condition
```rust
let inner_guard = self.inner.write().unwrap();
let status = unsafe {
    nixl_capi_query_mem(
        inner_guard.handle.as_ptr(),
        descs.handle(),
        resp.handle(),
        opt_args.map_or(std::ptr::null_mut(), |args| args.inner.as_ptr()),
    )
};
drop(inner_guard);
```

src/bindings/rust/src/descriptors/query.rs
```diff
@@ -0,0 +1,172 @@
...
+    inner: NonNull<bindings::nixl_capi_query_resp_list_s>,
+}
+
+/// Represents a single query response which may or may not contain parameters
+pub struct QueryResponse<'a> {
+    list: &'a QueryResponseList,
+    index: usize,
+}
+
+impl QueryResponseList {
```
roiedanino, line 30: Do we have to implement List ourselves or can we just use rust built-in `Vec<QueryResponse>` ?
It seems like a basic List API we can probably avoid, maybe we can do some aliasing to the type

src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -48,8 +48,42 @@ impl<'a> RegDescList<'a> {
...
-            nixl_capi_reg_dlist_add_desc(self.inner.as_ptr(), addr as uintptr_t, len, dev_id)
+            nixl_capi_reg_dlist_add_desc(
+                self.inner.as_ptr(),
+                addr as uintptr_t,
+                len,
+                dev_id,
+                std::ptr::null(),
+                0,
+            )
+        };
```
roiedanino: code duplication, just call `add_desc_with_meta` and pass null for metadata

src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -48,8 +48,42 @@ impl<'a> RegDescList<'a> {
...
+        dev_id: u64,
+        metadata: &[u8],
+    ) -> Result<(), NixlError> {
+        let status = unsafe {
+            nixl_capi_reg_dlist_add_desc(
+                self.inner.as_ptr(),
+                addr as uintptr_t,
+                len,
+                dev_id,
+                metadata.as_ptr() as *const _,
```
roiedanino:
```suggestion
                metadata.as_ptr() as *const c_void,
```

src/bindings/rust/tests/tests.rs
```diff
@@ -432,3 +432,163 @@ fn test_basic_agent_lifecycle() {
...
+    // Add blob descriptors with filenames as metadata
+    descs
+        .add_desc_with_meta(0, 1024, 0, test_file1.to_string_lossy().as_bytes())
+        .expect("Failed to add descriptor for test file 1");
+    descs
+        .add_desc_with_meta(0, 1024, 0, non_existent_file.to_string_lossy().as_bytes())
+        .expect("Failed to add descriptor for non-existent file");
+    descs
+        .add_desc_with_meta(0, 1024, 0, test_file2.to_string_lossy().as_bytes())
+        .expect("Failed to add descriptor for test file 2");
```
roiedanino: Can we use a loop here? Can we put magic numbers in consts?

src/bindings/rust/tests/tests.rs
```diff
@@ -432,3 +432,163 @@ fn test_basic_agent_lifecycle() {
...
+        .query_mem(&descs, Some(&opt_args))
+        .expect("Failed to query mem");
+
+    // Verify results
+    let num_responses = resp.len().expect("Failed to get response count");
+    assert_eq!(
+        num_responses, 0,
+        "Expected 0 responses for empty descriptor list"
+    );
+}
```
roiedanino: Do we have to create backends and all of that for empty list? Why would that test fail without POSIX?

src/bindings/rust/tests/tests.rs
```diff
@@ -432,3 +432,163 @@ fn test_basic_agent_lifecycle() {
     assert!(storage1.as_slice().iter().all(|&x| x == 0xbb));
     assert!(storage2.as_slice().iter().all(|&x| x == 0xbb));
 }
+
+#[test]
+fn test_query_mem_with_files() {
```
roiedanino, line 471: I think we can create a setup function for those tests to avoid code duplication as much as possible, they are quite long and its a bit hard to follow that way

src/bindings/rust/wrapper.cpp
```diff
@@ -1715,4 +1727,111 @@ nixl_capi_notif_map_clear(nixl_capi_notif_map_t map)
...
+nixl_capi_status_t
+nixl_capi_query_resp_list_has_value(nixl_capi_query_resp_list_t list,
+                                    size_t index,
+                                    bool *has_value) {
+    if (!list || !has_value || index >= list->responses.size()) {
+        return NIXL_CAPI_ERROR_INVALID_PARAM;
+    }
+
+    try {
+        *has_value = list->responses[index].has_value();
```
roiedanino, line 1786: please check size before accessing at index.

src/bindings/rust/wrapper.cpp
```diff
@@ -1715,4 +1727,111 @@ nixl_capi_notif_map_clear(nixl_capi_notif_map_t map)
...
+nixl_capi_status_t
+nixl_capi_query_resp_list_get_params(nixl_capi_query_resp_list_t list,
+                                     size_t index,
+                                     nixl_capi_params_t *params) {
+    if (!list || !params || index >= list->responses.size()) {
+        return NIXL_CAPI_ERROR_INVALID_PARAM;
+    }
+
+    try {
+        if (!list->responses[index].has_value()) {
```
roiedanino, line 1803: check size

src/bindings/rust/wrapper.cpp
```diff
@@ -1715,4 +1727,111 @@ nixl_capi_notif_map_clear(nixl_capi_notif_map_t map)
...
+            return NIXL_CAPI_ERROR_INVALID_PARAM;
+        }
+
+        auto param_list = new nixl_capi_params_s;
```
roiedanino, line 1807: Who is responsible for deleting param_list? the caller? maybe we need some "destroy_params" function similar to whats happening with response list

src/bindings/rust/wrapper.cpp
```diff
@@ -85,6 +85,10 @@ struct nixl_capi_notif_map_s {
   nixl_notifs_t notif_map;
 };
 
+struct nixl_capi_query_resp_list_s {
+    std::vector<nixl_query_resp_t> responses;
```
roiedanino, line 89: This definitely justifies using Rust built-in Vector as well

src/bindings/rust/wrapper.cpp
roiedanino: I saw that underline its an `unordered_map`, maybe the user can pass it to avoid calling `new` here?

src/bindings/rust/src/agent.rs
w1ldptr: Hmm, why is the race condition only manifests in case of query API when all existing code in this file use that handle exactly the same way? (register_memory, get_local_md, load_remote_md, post_xfer_req, etc.).

roiedanino: I know and agree, we should fix the other APIs as well, I'll open a ticket for it (NIX-257), and it will be handled eventually, big parts of the code were migrated as-is from dynamo, we are trying to implement better practices now

src/bindings/rust/tests/tests.rs
w1ldptr: nixlAgent::queryMem() just forwards the message to target backend. Without providing it with some backend instance it will just instantly return an error:

```
nixl_status_t
nixlAgent::queryMem(const nixl_reg_dlist_t &descs,
                    std::vector<nixl_query_resp_t> &resp,
                    const nixl_opt_args_t *extra_params) const {

    if (!extra_params || extra_params->backends.size() != 1) {
    }

    return extra_params->backends[0]->engine->queryMem(descs, resp);
}
```

src/bindings/rust/wrapper.cpp
w1ldptr: How is check on the line 1781 not enough?

src/bindings/rust/wrapper.cpp
w1ldptr: How is check on the line 1798 not enough?

roiedanino: Missed that somehow, my bad

src/bindings/rust/src/descriptors/reg.rs
w1ldptr: I might be missing something, but how exactly do you propose to pass 'null' to a function that accepts &[u8] parameter?

src/bindings/rust/tests/tests.rs
w1ldptr: It is written in exactly the same style as all other tests in this file, but fine.

src/bindings/rust/src/agent.rs
w1ldptr: NIXL agent already implements its own locking internally. What is the point of duplicating it on the Rust binding level?

src/bindings/rust/src/descriptors/query.rs
w1ldptr: Not sure I'm following. The internals of this is a pointer to opaque C++ object that is created by capi:

```
    pub fn new() -> Result<Self, NixlError> {
        let mut list = ptr::null_mut();
        let status = unsafe { nixl_capi_create_query_resp_list(&mut list) };

        match status {
            NIXL_CAPI_SUCCESS => {
                // SAFETY: If status is NIXL_CAPI_SUCCESS, list is non-null
                let inner = unsafe { NonNull::new_unchecked(list) };
                Ok(Self { inner })
            }
            NIXL_CAPI_ERROR_INVALID_PARAM => Err(NixlError::InvalidParam),
            _ => Err(NixlError::BackendError),
        }
    }
```

And then it is passed directly to other capi calls like this:

```
    pub fn query_mem(
        &self,
        descs: &RegDescList,
        opt_args: Option<&OptArgs>,
    ) -> Result<QueryResponseList, NixlError> {
        let resp = QueryResponseList::new()?;

        let status = unsafe {
            nixl_capi_query_mem(
                self.inner.write().unwrap().handle.as_ptr(),
                descs.handle(),
                resp.handle(),
                opt_args.map_or(std::ptr::null_mut(), |args| args.inner.as_ptr()),
            )
        };

        match status {
            NIXL_CAPI_SUCCESS => Ok(resp),
            NIXL_CAPI_ERROR_INVALID_PARAM => Err(NixlError::InvalidParam),
            _ => Err(NixlError::BackendError),
        }
    }
```

As other containers in NIXL Rust bindings (RegDescList, XfreDescList) is implemented similarly I assumed that the motivation for this is achieving zero-copy for them. How do you pass Rust Vec instead to capi? What would be the lifetime of QueryResponse instances it holds and which object would own resulting bindings::nixl_capi_query_resp_list_s instance that those QueryResponse objects point to?

src/bindings/rust/src/agent.rs
roiedanino: The internal locking mechanism within NIXL won't protect access to AgentInner fields, maybe we could use some re-design of the internal implementation to avoid it but it would also be on a separate PR

@ryanolson What do you think?

src/bindings/rust/wrapper.cpp
w1ldptr: Can you suggest the exact implementation you have in mind? I'm not getting how to just pass Rust Vec into capi that requires an instance of std::vector to pass it to nixlAgent API.

src/bindings/rust/src/descriptors/reg.rs
roiedanino: So maybe an empty metadata array/sequence?

src/bindings/rust/wrapper.cpp
w1ldptr: The result of this function is used to instantiate Params struct:
```
    pub fn get_params(&self) -> Result<Option<Params>, NixlError> {
        if !self.has_value()? {
            return Ok(None);
        }

        let mut params = ptr::null_mut();
        let status = unsafe {
            nixl_capi_query_resp_list_get_params(self.list.inner.as_ptr(), self.index, &mut params)
        };

        match status {
            NIXL_CAPI_SUCCESS => {
                // SAFETY: If status is NIXL_CAPI_SUCCESS, params is non-null
                let inner = unsafe { NonNull::new_unchecked(params) };
                Ok(Some(Params::new(inner)))
            }
            NIXL_CAPI_ERROR_INVALID_PARAM => Err(NixlError::InvalidParam),
            _ => Err(NixlError::BackendError),
        }
    }
```

Then, the resulting pointer deallocation by corresponding drop by using generic nixl_capi_destroy_params() function:
```
impl Drop for Params {
    fn drop(&mut self) {
        // SAFETY: self.inner is guaranteed to be valid by NonNull
        unsafe {
            nixl_capi_destroy_params(self.inner.as_ptr());
        }
    }
}
```

src/bindings/rust/src/descriptors/query.rs
roiedanino: I was thinking passing `Vec::as_mut_ptr()` of a `Vec` created before calling the function, the difference between `RegDescList`, `XferDescList`, and this case is that they expose NIXL-specific APIs matching inner objects while here its just a simple vector of responses which I'm not sure why we should implement it ourselves

w1ldptr: Pass Vec::as_mut_ptr() to capi? How would it work? nixlAgent::queryMem() expects a std::vector instance that it will push responses into, so how do you pass result of Vec::as_mut_ptr() instead?

src/bindings/rust/src/agent.rs
w1ldptr: > The internal locking mechanism within NIXL won't protect access to AgentInner fields

Indeed, but this function doesn't access other AgentInner fields besides nixlAgent handle.

src/bindings/rust/src/descriptors/query.rs
roiedanino: Went over the PR again, it seems doable, but only when passing non-opaque pointers, or change how this API works, for example querying for memory size then allocating the vector within rust, then passing a pointer and size to query memory, but at this point it seems like its not saving a lot of effort compared to your current implementation, I saw there's a library that does bindings for std::vector but thats probably also out of scope, I guess we can move on with rest of the  requested changes and merge this

src/bindings/rust/src/agent.rs
w1ldptr: Anyway, extended the scope of the lock to the whole nixl_capi_query_mem() call.

roiedanino (approved):

PR #622, opened 2025-07-27
Plugin manager enhancements

aranadive:
## What?
Simplify the way plugins are registered with the plugin manager

## Why?
Reduces a lot of the boilerplate code for the plugins. Since this is internal to plugin manager, it doesnt affect functionality. Also, helps to commonize the interface for all plugins.

## How?
Template added for BackendPlugin class. Added macros to easily register a plugin.


src/core/nixl_plugin_manager.cpp
```diff
@@ -390,43 +390,46 @@ const std::vector<nixlStaticPluginInfo>& nixlPluginManager::getStaticPlugins() {
     return static_plugins_;
 }
 
+#define REGISTER_STATIC_PLUGIN(name)                        \
```
yosefe: NIXL_REGISTER_STATIC_PLUGIN

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,107 @@ class nixlBackendPlugin {
 // Macro to define exported C functions for the plugin
 #define NIXL_PLUGIN_EXPORT __attribute__((visibility("default")))
 
+// Template for creating backend plugins with minimal boilerplate
+template<typename EngineType> class nixlBackendPluginTemplate {
```
yosefe: where is this class used?

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,107 @@ class nixlBackendPlugin {
...
+                create_engine_impl,
+                destroy_engine_impl,
+                []() { return plugin_name; },
+                []() { return plugin_version; },
+                get_options,
+                get_mems};
+    }
+};
+
+#define NIXL_DEFINE_PLUGIN(PluginName, EngineType, PluginVersion, GetOptionsFunc, GetMemsFunc) \
```
yosefe: GetOptionsFunc -> PluginOptions
GetMemsFunc -> PluginMems
They can be either a function or value like {}, both will work

src/plugins/cuda_gds/gds_plugin.cpp
```diff
@@ -18,71 +18,22 @@
...
+nixl_b_params_t
+get_gds_options() {
     nixl_b_params_t params;
     return params;
 }
 
-// Function to get supported backend mem types
-static nixl_mem_list_t get_backend_mems() {
+nixl_mem_list_t
+get_gds_mems() {
```
yosefe: instead we could lass {DRAM_SEG, VRAM_SEG, FILE_SEG} directly to NIXL_DEFINE_PLUGIN macro

src/plugins/cuda_gds/gds_plugin.cpp
```diff
@@ -18,71 +18,22 @@
...
-}
-
-// Plugin cleanup function
-extern "C" NIXL_PLUGIN_EXPORT void nixl_plugin_fini() {
-    // Cleanup any resources if needed
-}
-
-#endif
+// Define the complete GDS plugin using the template
+NIXL_DEFINE_PLUGIN(GDS, nixlGdsEngine, "0.1.1", get_gds_options, get_gds_mems)
```
yosefe: we can probably remove get_gds_options func and just pass {}

src/plugins/gpunetio/gpunetio_plugin.cpp
```diff
@@ -18,83 +18,24 @@
...
     params["gpu_devices"] = "";
     params["cuda_streams"] = "";
     return params;
 }
 
-// Function to get supported backend mem types
-static nixl_mem_list_t
-get_backend_mems() {
+nixl_mem_list_t
+get_gpunetio_mems() {
```
yosefe: pass {DRAM, VRAM} directly to the macro

yosefe (commented):
can we also use this macro in tests?

src/api/cpp/backend/backend_plugin.h
rakhmets: Please avoid using macros. This is one of the C++ Core Guidelines.
**Reason** Macros are a major source of bugs. Macros don’t obey the usual scope and type rules. Macros ensure that the human reader sees something different from what the compiler sees. Macros complicate tool building. (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)

You can use templates to have static (compile-time) polymorphism.
For me personally, using inheritance/dynamic polymorphism is also fine in this case.

src/api/cpp/backend/backend_plugin.h
aranadive: This was from a previous templated change I had for plugins. From Raul's comment, this would be preferred than the macro.

src/api/cpp/backend/backend_plugin.h
aranadive: I had the template style before I moved to the macro. I guess I can move back to that.

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,49 @@ class nixlBackendPlugin {
...
+// Template for creating backend plugins with minimal boilerplate
+template<typename EngineType> class nixlBackendPluginTemplate {
+public:
+    [[nodiscard]] static nixlBackendEngine *
+    create_engine_impl(const nixlBackendInitParams *init_params) {
+        try {
+            return new EngineType(init_params);
+        }
+        catch (const std::exception &e) {
+            return nullptr;
```
rakhmets:
```suggestion
            NIXL_ERROR << "Failed to create engine: " << e.what();
            return nullptr;
```

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,49 @@ class nixlBackendPlugin {
...
+        }
+    }
+
+    static void
+    destroy_engine_impl(nixlBackendEngine *engine) {
```
rakhmets: Can we put create and destroy engine methods to private scope?

src/plugins/ucx/ucx_plugin.cpp
```diff
@@ -17,74 +17,33 @@
...
-       create_ucx_engine,
-       destroy_ucx_engine,
-       get_plugin_name,
-       get_plugin_version,
-       get_backend_options,
-       get_backend_mems
-   };
+namespace {
+nixl_b_params_t
+get_ucx_options() {
```
rakhmets: Can we just use `get_ucx_backend_common_options` instead?

src/plugins/ucx/ucx_plugin.cpp
```diff
@@ -17,74 +17,33 @@
 
 #include "backend/backend_plugin.h"
 #include "ucx_backend.h"
-
 #include "nixl_log.h"
```
rakhmets: Please remove, this header is no longer needed.

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,49 @@ class nixlBackendPlugin {
 // Macro to define exported C functions for the plugin
 #define NIXL_PLUGIN_EXPORT __attribute__((visibility("default")))
 
+// Template for creating backend plugins with minimal boilerplate
+template<typename EngineType> class nixlBackendPluginTemplate {
```
rakhmets: I think `nixlBackendPluginFactory` or `nixlBackendPluginCreator` will better reflect the purpose of the class.

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,49 @@ class nixlBackendPlugin {
...
+public:
+    [[nodiscard]] static nixlBackendEngine *
+    create_engine_impl(const nixlBackendInitParams *init_params) {
```
rakhmets: Maybe just: `createEngine` and `destroyEngine`.

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +50,49 @@ class nixlBackendPlugin {
...
+        }
+    }
+
+    static void
+    destroy_engine_impl(nixlBackendEngine *engine) {
+        delete engine;
+    }
+
+    static nixlBackendPlugin *
+    initialize_plugin(int api_version,
```
rakhmets: Maybe just `create`, so the usage in the client code may look like:
```C++
using my_plugin_factory = nixlBackendPluginFactory<myEngine>;
...
extern "C" nixlBackendPlugin *
createMyPlugin() {
    return my_plugin_factory::create( ... );
}
```

src/plugins/cuda_gds/gds_plugin.cpp
```diff
@@ -18,71 +18,26 @@
...
-    NIXL_PLUGIN_API_VERSION,
-    create_gds_engine,
-    destroy_gds_engine,
-    get_plugin_name,
-    get_plugin_version,
-    get_backend_options,
-    get_backend_mems
-};
+// Plugin type alias for convenience
+using GdsPlugin = nixlBackendPluginTemplate<nixlGdsEngine>;
```
rakhmets: Please take into account the latest state of https://github.com/ai-dynamo/nixl/blob/main/docs/CodeStyle.md#naming-conventions

src/api/cpp/backend/backend_plugin.h
aranadive: Done

src/plugins/ucx/ucx_plugin.cpp
aranadive: Done

src/api/cpp/backend/backend_plugin.h
aranadive: Went with nixlBackendPluginCreator

src/plugins/cuda_gds/gds_plugin.cpp
aranadive: I changed the names to be in format of gds_plugin_t.

src/plugins/cuda_gds/gds_plugin.cpp
```diff
@@ -18,71 +18,25 @@
...
-};
+// Plugin type alias for convenience
+using gds_plugin_t = nixlBackendPluginCreator<nixlGdsEngine>;
 
 #ifdef STATIC_PLUGIN_GDS
-
-nixlBackendPlugin* createStaticGdsPlugin() {
-    return &plugin; // Return the static plugin instance
+// Function for static loading
+extern "C" nixlBackendPlugin *
```
rakhmets: Why do we need to support C linkage?

src/core/nixl_plugin_manager.cpp
```diff
@@ -390,43 +390,46 @@ const std::vector<nixlStaticPluginInfo>& nixlPluginManager::getStaticPlugins() {
     return static_plugins_;
 }
 
+#define NIXL_REGISTER_STATIC_PLUGIN(name)                   \
+    extern nixlBackendPlugin *createStatic##name##Plugin(); \
+    registerStaticPlugin(#name, createStatic##name##Plugin);
```
rakhmets, line 395: It looks like we duplicate the same names in two different places.
Is it possible to update `nixlPluginManager::registerStaticPlugin`, and use name returned by `nixlBackendPlugin::get_plugin_name`?

src/core/nixl_plugin_manager.cpp
```diff
@@ -390,43 +390,46 @@ const std::vector<nixlStaticPluginInfo>& nixlPluginManager::getStaticPlugins() {
     return static_plugins_;
 }
 
+#define NIXL_REGISTER_STATIC_PLUGIN(name)                   \
```
rakhmets, line 393: I would suggest to avoid using macros here either.
Maybe keep `registerBuiltinPlugins` unchanged, and do this in a separate PR, if the changes are too large for a single PR.

src/api/cpp/backend/backend_plugin.h
```diff
@@ -50,6 +54,58 @@ class nixlBackendPlugin {
...
+            return nullptr;
+        }
+    }
+
+    static void
+    destroyEngine(nixlBackendEngine *engine) {
+        delete engine;
+    }
+
+public:
```
rakhmets, line 59: Please put `public` before `private`.
https://github.com/ai-dynamo/nixl/blob/main/docs/CodeStyle.md#member-declaration-order

src/plugins/cuda_gds/gds_plugin.cpp
aranadive: I can remove this in separate PR since this is a bit big already.

src/core/nixl_plugin_manager.cpp
aranadive: I can update this in separate PR since this will require testing changes as well.

src/api/cpp/backend/backend_plugin.h
aranadive: Done

src/plugins/cuda_gds/gds_plugin.cpp
rakhmets: `extern "C"` wasn't here (in `#ifdef STATIC_PLUGIN_GDS`) before these changes. It's the same with other plugins.
Please remove if it's not needed.

src/core/nixl_plugin_manager.cpp
aranadive: It would be easier doing in a separate PR with the other changes above.

src/plugins/cuda_gds/gds_plugin.cpp
aranadive: Ah you're right. Will revert for static plugins.

src/core/nixl_plugin_manager.cpp
rakhmets: Then maybe it's better not to change `nixlPluginManager::registerBuiltinPlugins`.
(except for adding previously skipped backends: GDS_MT, HF3FS, and fixing name of GPUNETIO backend).
Because along with `NIXL_REGISTER_STATIC_PLUGIN` macro, names have been added that do not comply with our naming conventions:
`createStaticGdsMtPlugin` -> `createStaticGDS_MTPlugin`
`createStaticUcxMoPlugin` -> `createStaticUCX_MOPlugin`

Other function names have also changed in an unusual way:
`createStaticGdsPlugin` -> `createStaticGDSPlugin`
`createStaticMooncakePlugin` -> `createStaticMOONCAKEPlugin`
...

brminich (approved):

src/core/nixl_plugin_manager.cpp
aranadive: These are based on the plugin name defined by plugin authors. It cleans up most of the static plugin registration code. I think its okay to make an exception in such cases.

brminich (approved):

PR #627, opened 2025-07-28
[CI] Add new CI test to detect large PRs

dpressle:
Limit for single PR should be 500 added lines, except subprojects and new files.

## What?
Add CI check for PR changed line number

## Why?
We want to avoid very large PRs

## How?
Add github workflow to detect large PRs


.github/workflows/pr-size-check.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2  # Need full history to compare with base branch
+
+    - name: Check PR size
+      run: |
+        # Calculate total lines changed (additions + deletions)
+        LINES_CHANGED=$(git diff --numstat HEAD^1 HEAD | awk '{sum += $1 + $2} END {print sum+0}')
```
yosefe: can we use the following file pattern for checking:
`grep -E '\.(c|cc|cpp|h|cu|cuh|rs|py|build)$'`
and exclude "subprojects" directory
also count only added lines (not deleted lines)

dpressle:
Test PR less then 500: https://github.com/ai-dynamo/nixl/actions/runs/16573268758/job/46870994231
Test PR more then 500 in subprojects: https://github.com/ai-dynamo/nixl/actions/runs/16587841238/job/46916585073
Test PR more then 500 not in subprojects: https://github.com/ai-dynamo/nixl/actions/runs/16587841238/job/46916585073

.github/workflows/pr-size-check.yml
dpressle: done see test results in comments

yosefe: I don't see where specific extensions are checked?

dpressle: why we need to check extensions? i thought we only want to ignore subprojects folder changes

.github/workflows/pr-size-check.yml
```diff
@@ -0,0 +1,67 @@
...
+        fi
+
+        echo "📈 Change Summary:"
+        echo "  • Lines added to existing files: $LINES_CHANGED"
+        echo "  • Lines added in new files: $LINES_NEW_FILES"
+        echo "  • Lines added in subprojects/: $LINES_SUBPROJECTS"
+        echo "  • Total lines (excluding subprojects): $TOTAL_LINES_EXCLUDING_SUBPROJECTS"
+        echo "  • Total lines (including subprojects): $TOTAL_LINES_ALL"
+        echo "  • Maximum allowed lines: 500"
+        echo ""
```
yosefe: The command:
`git show --numstat --pretty="" --diff-filter=M -- . ':(exclude)subprojects/*' | awk '{s+=$1} END {print s}'`

The prompt for Gemini Pro 2.5:
"pls write a git diff based command that shows how many lines were added in the last commit, excluding new files and excluding files in "subprojects" directory"

yosefe:
@aranadive @barneuman @w1ldptr can you pls take a look if have agree with this approach in principle ?

.github/workflows/pr-size-check.yml
dpressle: What do you mean i am using git diff and i am showing all the data:
        echo "  • Lines added to existing files: $LINES_CHANGED"
        echo "  • Lines added in new files: $LINES_NEW_FILES"
        echo "  • Lines added in subprojects/: $LINES_SUBPROJECTS"
        echo "  • Total lines (excluding subprojects): $TOTAL_LINES_EXCLUDING_SUBPROJECTS"
        echo "  • Total lines (including subprojects): $TOTAL_LINES_ALL"

w1ldptr:
> @aranadive @barneuman @w1ldptr can you pls take a look if have agree with this approach in principle ?

I don't see much value in this automation since code owners are already free to reject all PRs marked as XXL, for example. But I'm not opposed to this either, _assuming the check won't become "Required"_.

.github/workflows/pr-size-check.yml
yosefe: IMO It would simpler and less error-prone to use the simple command above, rather than counting all the types of changed lines and doing calculations. Also the grep command assumes specific output format and better avoid this if possible.

PR #633, opened 2025-07-29
PYTHON-PROJECT-WIDE: replaced prints with a logger, fixed backtraces

roiedanino:
## What?
Fixed https://github.com/ai-dynamo/nixl/pull/512 by removing hostname from logger format

## Why?
PR was merged and reverted due to traceback prints flooding python examples output


src/api/python/logging.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if _logging_configured:
+        return
+
+    # Find python_logging.ini in the same directory as this file (project root)
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    config_file_path = os.path.join(current_dir, "python_logging.ini")
+    config_file = os.path.abspath(config_file_path)
+
+    logging.config.fileConfig(config_file, disable_existing_loggers=False)
```
ovidiusm: As discussed, we should not apply any kind of config changes to the root logger. The ini file touches the root logger.

roiedanino: Yes, agree, I'm currently just testing if raiseExceptions fails CI on those issues

ovidiusm: Thanks! Please also take a look at my changes from https://github.com/ai-dynamo/nixl/pull/629/files

I found a few calls like `logger.info(var)` which should be `logger.info("%s", var)`

roiedanino:
@ovidiusm @dpressle can take a look please?

benchmark/kvbench/test/custom_traffic_perftest.py
```diff
@@ -377,7 +388,9 @@ def run(
                     total_size_gb,
                 ]
             ]
-            print(tabulate(data, headers=headers, floatfmt=".6f"))
+            logger.info(
+                f"Performance metrics:\n{tabulate(data, headers=headers, floatfmt='.6f')}"
```
ovidiusm: Why not "%s", ... ?

src/api/python/meson.build
```diff
@@ -18,3 +18,5 @@ py = import('python').find_installation('python3', pure: false)
 py.install_sources('_api.py', subdir: ('nixl'))
 py.install_sources('__init__.py', subdir: ('nixl'))
 py.install_sources('py.typed', subdir: ('nixl'))
+py.install_sources('logging.py', subdir: ('nixl'))
+py.install_sources('python_logging.ini', subdir: ('nixl'))
```
ovidiusm: Let's remove the ini file please

src/api/python/python_logging.ini
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ovidiusm: This is the old file that was touching the root logger, let's remove it since it is not used

o90o (approved):

ovidiusm (approved):

dpressle (approved):

PR #635, opened 2025-07-29
Fix nixlbench: properly implement --num_files, fix READ mode support

da-x:
So far, giving `--num_files <n>` with a value about 1 resulted in one file with the total buffer size, while the other files with ending up with file size 0.

The PR fixes this along with other small related fixes.

- Allow dividing M buffers to N files, checking that M is divisible by N, and that each file contains at least one buffer.
- READ mode does not need to write to the files if they are big enough.
- `buffer_size` is already made sure to be page aligned for direct IO in `allocateMemory`, no need to do the alignment again in `initBasicDescFile`.
- No need to recall `initBasicDescFile` in `exchangeIOV`, as we don't want to do writes to the file for the second time. We only need to changes `len` from the `remote_iovs` that were already computed. The offset in the file will remain the same, only len will be smaller than the maximum.
- Remove `gds_running_ptr` global. Allocate a `buffer_size` from each file,      so we need to track what offsets we already used using the newly added `xferFileState`.
- Remove unused globals `gds_remote_iovs`, `storage_remote_iovs`, `gds_remote_iovs`.
- Fix `pwrite` call for short write handling.


aranadive:
@da-x can you check if some of this is covered by PR #508?

da-x:
@aranadive PR 508 ties the number of files to the number of threads. This PR divides individual buffers to files, instead of having all buffers in a single file. It does so irrespective of threads. I was under the impression that `num_threads` is more highlevel in the nixlbench envelope design. What is more desired in the larger scope?

benchmark/nixlbench/src/worker/nixl/nixl_worker.h
```diff
@@ -63,7 +70,10 @@ class xferBenchNixlWorker: public xferBenchWorker {
         std::optional<xferBenchIOV> initBasicDescVram(size_t buffer_size, int mem_dev_id);
         void cleanupBasicDescVram(xferBenchIOV &basic_desc);
 #endif
-        std::optional<xferBenchIOV> initBasicDescFile(size_t buffer_size, int fd, int mem_dev_id);
+        void
+        advanceGdsRunningPtr(size_t buffer_size, int mem_dev_id);
```
aranadive: Where is this defined?

da-x: Removing - it's artifact from an earlier version of the PR.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -428,26 +424,48 @@ createFileFds(std::string name) {
...
+
+    fstate.offset = end_offset;
+
+    // If in READ mode, only write if the region is not already present in the file
+    if (XFERBENCH_OP_READ == xferBenchConfig::op_type && end_offset <= fstate.file_size) {
+        return ret;
+    }
+
     // Fill up with data
     void *buf;
```
aranadive, line 470: Move buffer allocation to createFileFds so we allocate only once.

aranadive (approved):

aranadive:
> @aranadive PR 508 ties the number of files to the number of threads. This PR divides individual buffers to files, instead of having all buffers in a single file. It does so irrespective of threads. I was under the impression that `num_threads` is more highlevel in the nixlbench envelope design. What is more desired in the larger scope?

The threads ensures that we try to push more data as possible and have multiple ways to access data for block managers. We want to be able to provide a broad platform for testing various scenarios. PR 508 uses the scheme to decide on the access patterns. That would be an add-on to this PR.

aranadive:
@da-x looks like nixlbench is failing for storage plugins with multiple batch sizes. Can you check?
```

4096                1              0.188296       21.8           13.0           13.0           2.4            3.0            19.3           26.0           
NIXL: createTransferReq failed (Error code: -2)
```

The nixlbench command is:
```
/usr/local/nixlbench/bin/nixlbench --etcd-endpoints http://127.0.0.1:10511 --initiator_seg_type DRAM --target_seg_type DRAM --filepath /tmp --total_buffer_size 80000000 --start_block_size 4096 --max_block_size 16384 --start_batch_size 1 --max_batch_size 4 --benchmark_group JHb168NkuDIYj5lVHODRtP6ybbhNgOzF --backend POSIX --op_type READ
```

da-x:
The issue was with `batch_size > 1` mode.

I found and fix two things.

1)

First, in my original change I should have retained the creation of `remote_iovs` from `local_iovs` to make sure the amount of bytes to transfer in both lists is identical (otherwise `createTransferReq` rightfully fails).

2)

Also, I have found that even now in `main` branch, there is a pre-existing bug that if you pass a batch size larger than the one, with a single file, all IOs go to offset 0, because the file size was calculated to allow for all buffers having their own offset.

By printing in `nixlXferDlistToIOVList` we can see this:

```
# main branch
local_iov
[0]: addr=0x7f38df3b4010 len=0x1000 devId=0x0
[1]: addr=0x7f38df3b5010 len=0x1000 devId=0x0
[2]: addr=0x7f38df3b6010 len=0x1000 devId=0x0
[3]: addr=0x7f38df3b7010 len=0x1000 devId=0x0

remote_iovs
[0]: addr=0x0 len=0x1000 devId=0x6
[1]: addr=0x0 len=0x1000 devId=0x6
[2]: addr=0x0 len=0x1000 devId=0x6
[3]: addr=0x0 len=0x1000 devId=0x6
```

(Run of `./benchmark/nixlbench/build/nixlbench --initiator_seg_type DRAM --target_seg_type DRAM --filepath /tmp/nixl --total_buffer_size 80000000 --start_block_size 4096 --max_block_size 4096 --start_batch_size 4 --max_batch_size 4 --benchmark_group JHb168NkuDIYj5lVHODRtP6ybbhNgOzF --backend POSIX --op_type READ`)

After the fix there are non-overlapping offsets:

```
# PR branch
local_iov
[0]: addr=0x7fe0873b4010 len=0x1000 devId=0x0
[1]: addr=0x7fe0873b5010 len=0x1000 devId=0x0
[2]: addr=0x7fe0873b6010 len=0x1000 devId=0x0
[3]: addr=0x7fe0873b7010 len=0x1000 devId=0x0

remote_iovs
[0]: addr=0x0 len=0x1000 devId=0x6
[1]: addr=0x1000 len=0x1000 devId=0x6
[2]: addr=0x2000 len=0x1000 devId=0x6
[3]: addr=0x3000 len=0x1000 devId=0x6
```

The `--start_block_size 4096 --max_block_size 16384 --start_batch_size 1 --max_batch_size 4` run should work well now. Also, should consider adding optional debug prints such as the above for making sure see that the transfers lists make sense.

PR #639, opened 2025-07-30
NIXLBENCH: Add test in CI

ovidiusm:
## What?
Execute basic nixlbench test in CI

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -137,7 +137,11 @@ export CMAKE_PREFIX_PATH=${INSTALL_DIR}:${CMAKE_PREFIX_PATH}
 export UCX_TLS=^cuda_ipc
 
 meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true ${EXTRA_BUILD_ARGS}
-cd nixl_build && ninja && ninja install
+cd nixl_build && ninja && ninja install && cd ..
 
 # TODO(kapila): Copy the nixl.pc file to the install directory if needed.
 # cp ${BUILD_DIR}/nixl.pc ${INSTALL_DIR}/lib/pkgconfig/nixl.pc
+
+cd benchmark/nixlbench
+meson setup nixlbench_build -Dnixl_path=${INSTALL_DIR} -Dprefix=${INSTALL_DIR}
+cd nixlbench_build && ninja && ninja install && cd ..
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -84,6 +84,12 @@ cd ${INSTALL_DIR}
 sleep 1
 ./bin/nixl_test initiator 127.0.0.1 1234
 
+# Run NIXLBench test
+benchmark_group=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+./bin/nixlbench --etcd-endpoints http://127.0.0.1:2379 --initiator_seg_type DRAM --target_seg_type DRAM --total_buffer_size 80000000 --benchmark_group $benchmark_group &
+sleep 1
+./bin/nixlbench --etcd-endpoints http://127.0.0.1:2379 --initiator_seg_type DRAM --target_seg_type DRAM --total_buffer_size 80000000 --benchmark_group $benchmark_group
+
 echo "${TEXT_YELLOW}==== Disabled tests==="
 echo "./bin/md_streamer disabled"
 echo "./bin/p2p_test disabled"


.gitlab/test_cpp.sh
```diff
@@ -84,6 +84,12 @@ cd ${INSTALL_DIR}
 sleep 1
 ./bin/nixl_test initiator 127.0.0.1 1234
 
+# Run NIXLBench test
+benchmark_group=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+./bin/nixlbench --etcd-endpoints http://127.0.0.1:2379 --initiator_seg_type DRAM --target_seg_type DRAM --total_buffer_size 80000000 --benchmark_group $benchmark_group &
+sleep 1
+./bin/nixlbench --etcd-endpoints http://127.0.0.1:2379 --initiator_seg_type DRAM --target_seg_type DRAM --total_buffer_size 80000000 --benchmark_group $benchmark_group
```
brminich: may we try running with different params and possibly different backends?

ovidiusm: Agreed. I have added more tests. We are limited to DRAM though here since the system running the unit tests might not have a GPU.

.gitlab/build.sh
```diff
@@ -137,7 +138,11 @@ export CMAKE_PREFIX_PATH=${INSTALL_DIR}:${CMAKE_PREFIX_PATH}
...
 meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true ${EXTRA_BUILD_ARGS}
-cd nixl_build && ninja && ninja install
+cd nixl_build && ninja && ninja install && cd ..
 
 # TODO(kapila): Copy the nixl.pc file to the install directory if needed.
 # cp ${BUILD_DIR}/nixl.pc ${INSTALL_DIR}/lib/pkgconfig/nixl.pc
+
+cd benchmark/nixlbench
+meson setup nixlbench_build -Dnixl_path=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dprefix=${INSTALL_DIR}
+cd nixlbench_build && ninja && ninja install && cd ..
```
aranadive: This nixlbench build/install also needs to be added to the nixl dockerfile. The build.sh script does not seem to run in the Gitlab CI. The gitlab CI will build nixl docker containers and then use them for testing with the test scripts.

.gitlab/build.sh
```diff
@@ -137,7 +138,11 @@ export CMAKE_PREFIX_PATH=${INSTALL_DIR}:${CMAKE_PREFIX_PATH}
 export UCX_TLS=^cuda_ipc
 
 meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true ${EXTRA_BUILD_ARGS}
-cd nixl_build && ninja && ninja install
+cd nixl_build && ninja && ninja install && cd ..
```
brminich: imo better: `ninja -C nixl_build && ninja -C nixl_build install`

benchmark/nixlbench/meson.build
```diff
@@ -84,7 +84,23 @@ endif
 
 
 # UCX
-ucx_dep = dependency('ucx')
+ucx_path = get_option('ucx_path')
+if ucx_path != ''
+  ucx_lib_path = ucx_path + '/lib'
+  ucx_inc_path = ucx_path + '/include'
+  # Check if path is absolute
+  if ucx_inc_path.startswith('/')
```
brminich: why need this `if/else`?
can it be just 
```
ucx_path = get_option('ucx_path')
if ucx_path != ''
  ucx_lib_path = ucx_path + '/lib'
  ucx_inc_dir = include_directories(ucx_path + '/include')

  ucx_dep = declare_dependency(
    link_args: ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
    include_directories: ucx_inc_dir
  )
```

.gitlab/build.sh
ovidiusm: We should not add it to the nixl dockerfile, it will slow it down even more.

Instead I have added a separate script to run the nixlbench test and added it to the CI config.

.gitlab/build.sh
ovidiusm: Actually it should be ninja && $SUDO ninja install but I will do it in a separate commit

benchmark/nixlbench/meson.build
ovidiusm: Done, also cleaned up nixl meson.build file (which is where this snippet was coming from) and tested with absolute and relative paths

.gitlab/test_nixlbench.sh
```diff
@@ -0,0 +1,89 @@
...
+export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
+
+echo "==== Show system info ===="
+env
+nvidia-smi topo -m || true
+ibv_devinfo || true
+uname -a || true
+
+echo "==== Running ETCD server ===="
+export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:2379"
```
dpressle: you cant have fixed port we are running CI in parallel - use Roei's solution

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -108,6 +108,14 @@ steps:
       docker rm -f "${JOB_BASE_NAME}-${BUILD_ID}-${axis_index}"
```
dpressle, line 111: always can only be on the last stage in non last stage you need onfail:

dpressle (changes requested):

.gitlab/build.sh
ovidiusm: Fixed

.ci/jenkins/lib/test-matrix.yaml
ovidiusm: Fixed

benchmark/nixlbench/meson.build
ovidiusm: Actually I reverted it, because it was causing a CI failure. The problem is that include_directories() with relative paths fails if the dependency was pulled inside the source tree :(

benchmark/nixlbench/meson.build
```diff
@@ -84,7 +84,23 @@ endif
 
 
 # UCX
-ucx_dep = dependency('ucx')
+ucx_path = get_option('ucx_path')
```
aranadive: Is the ucx_dep dependency used anywhere in nixlbench? I couldnt find any instance. If not we can remove it.

aranadive (approved):

dpressle (changes requested):
What about build-matrix.yaml why you did not add this test there as well?

ovidiusm:
> What about build-matrix.yaml why you did not add this test there as well?

I added them. Why do we have tests in build matrix?

.gitlab/test_nixlbench.sh
```diff
@@ -0,0 +1,89 @@
...
+if [ -z "$INSTALL_DIR" ]; then
+    echo "Usage: $0 <install_dir>"
+    exit 1
+fi
+
+ARCH=$(uname -m)
+[ "$ARCH" = "arm64" ] && ARCH="aarch64"
+
+export LD_LIBRARY_PATH=${INSTALL_DIR}/lib:${INSTALL_DIR}/lib/$ARCH-linux-gnu:${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins:/usr/local/lib:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs:/usr/local/cuda-12.8/compat:$LD_LIBRARY_PATH
```
brminich, line 36: do we really need/expect dependency on cuda12.8?

benchmark/nixlbench/meson.build
```diff
@@ -82,10 +82,6 @@ if cuda_available
     endif
 endif
 
-
-# UCX
-ucx_dep = dependency('ucx')
```
brminich, line 87: is it unrealted fix? In CI we rely on UCX backend

ovidiusm: We do not need to *build* against ucx because nixlbench does not include anything from UCX directly. It only uses the nixl api

.gitlab/test_nixlbench.sh
ovidiusm: All tests are set up the same way, that's why we have these variables. See test_cpp.sh and test_python.sh. Let's not address env changes in this commit, since it would be better to do it when we reorganize the CI using a fixed base image. For now it is better if the tests are set up in a consistent way.

brminich (approved):

benchmark/nixlbench/meson_options.txt
```diff
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-option('cudapath_inc', type: 'string', value: '/usr/local/cuda/include', description: 'Include path for CUDA')
-option('cudapath_lib', type: 'string', value: '/usr/local/cuda/lib64/', description: 'Library path for CUDA')
-option('cudapath_stub', type: 'string', value: '/usr/local/cuda/lib64/stubs', description: 'Extra Stub path for CUDA')
+option('cudapath_inc', type: 'string', value: '', description: 'Include path for CUDA')
+option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
+option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
```
brminich, line 18: why is this needed?

ovidiusm: It was a mistake to hardcode the paths here. Because then we cannot autodetect where cuda is installed. And we now need to build nixlbench and nixl in the same script (.gitlab/build.sh) so the options should be aligned.

brminich (dismissed):

dpressle (dismissed):

.gitlab/build.sh
```diff
@@ -148,8 +150,12 @@ export UCX_TLS=^cuda_ipc
 
 # shellcheck disable=SC2086
 meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true -Drust=false ${EXTRA_BUILD_ARGS}
-cd nixl_build && ninja && ninja install
+meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true ${EXTRA_BUILD_ARGS}
```
rakhmets: Please delete one of the two `meson setup`.

dpressle (approved):

aranadive (approved):

PR #640, opened 2025-07-30
EXAMPLES/PYTHON: partial md example - fix test logic and warnings

roiedanino:
## What?
Fixing partial_md_example.py

## Why?
The test had 2 main issues:
1. Sending target_xfer_descs1 instead of target_xfer_descs2
2. Setting `is_sorted=True` while descriptors are not sorted by address, I set the flag to False. We can alternatively sort descriptors by address


gleon99 (approved):
+ @yosefe

roiedanino:
/ok to test [19b66e1](https://github.com/ai-dynamo/nixl/pull/640/commits/19b66e169e945dce0b425c56ead7fc1a682d7ace)

PR #642, opened 2025-07-30
RUST: Fix cargo package

ovidiusm:
## What?

The rust build scripts assume that NIXL_PREFIX env var is present. We should export it in the Dockerfile.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -134,6 +134,7 @@ RUN cd /usr/local/src && \
      make -j install-strip &&        \
      ldconfig
 
+ENV NIXL_PREFIX=$NIXL_PREFIX
 RUN rm -rf build && \
     mkdir build && \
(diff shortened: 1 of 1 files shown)


contrib/Dockerfile
```diff
@@ -134,6 +134,7 @@ RUN cd /usr/local/src && \
      make -j install-strip &&        \
      ldconfig
 
+ENV NIXL_PREFIX=$NIXL_PREFIX
```
aranadive, line 137: Do we need this in the manylinux file too? Which one do we release cargo packages from?

pvijayakrish: Currently cargo builds are done from the image build out of this Dockerfile. Can we also add it to the manylinux Dockerfiels?

ovidiusm: If we do not need it let's not add it, to keep the fix small.

PR #646, opened 2025-07-31
OBJ: Add option to create S3 bucket

w1ldptr:
## What?
Instead of just assuming that the target S3 bucket already exists during the backend initialization add capability to create it.

## Why?
To not force users to duplicate AWS S3 control path in their app in order to just create an S3 bucket for NIXL.

## How?
Add optional backend parameter 'create_bucket' that user can set to 'true' in order to request Obj backend to do it.

### DIFF
--- a/.github/workflows/aws_s3_validation.yml
+++ b/.github/workflows/aws_s3_validation.yml
@@ -36,31 +36,6 @@ jobs:
 
           # Verify AWS credentials
           aws sts get-caller-identity >/dev/null
-
-          # Create S3 bucket
-          aws s3 mb s3://${AWS_DEFAULT_BUCKET}
-
-          # Loop to check if S3 bucket exists with retries
-          max_attempts=5
-          attempt=1
-          while [ $attempt -le $max_attempts ]; do
-            if aws s3 ls s3://${AWS_DEFAULT_BUCKET} >/dev/null 2>&1; then
-              echo "S3 bucket exists successfully on attempt $attempt"
-              break
-            else
-              echo "S3 bucket check failed on attempt $attempt/$max_attempts"
-              if [ $attempt -eq $max_attempts ]; then
-                echo "S3 bucket creation failed after $max_attempts attempts"
-                exit 1
-              fi
-              sleep 2
-              attempt=$((attempt + 1))
-            fi
-          done
-
-          # validate upload file to s3
-          echo test > testfile.txt
-          aws s3 cp testfile.txt s3://${AWS_DEFAULT_BUCKET}/testfile.txt
       - name: Build NIXL
         timeout-minutes: 30
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/aws_s3_validation.yml
```diff
@@ -36,31 +36,6 @@ jobs:
...
-                exit 1
-              fi
-              sleep 2
-              attempt=$((attempt + 1))
-            fi
-          done
-
-          # validate upload file to s3
-          echo test > testfile.txt
-          aws s3 cp testfile.txt s3://${AWS_DEFAULT_BUCKET}/testfile.txt
```
mamar123, line 63: Do we still expect the CI to delete the bucket (down below), given that it no longer creates it?

w1ldptr: Yes. Otherwise, the next CI invocation won't be able to validate that the bucket has been successfully re-created by Obj.

mamar123: The bucket is unique per run, but cleaning it up may still be worthwhile to keep the cloud clean

src/plugins/obj/obj_plugin.cpp
```diff
@@ -55,6 +55,7 @@ get_backend_options() {
     params["secret_key"] = "AWS secret access key (required)";
     params["session_token"] = "AWS session token (optional)";
     params["bucket"] = "S3 bucket name (optional)";
+    params["create_bucket"] = "Create bucket if it doesn't exist (true/false) (default: false)";
```
barneuman, line 58: Please add `optional` similar to other params.

src/plugins/obj/obj_s3_client.cpp
```diff
@@ -251,3 +281,50 @@ awsS3Client::checkObjectExists(std::string_view key) {
...
+    // Can be empty for us-east-1, which is the default region
+    if (!config_.region.empty()) {
+        Aws::S3::Model::CreateBucketConfiguration configuration;
+        configuration.SetLocationConstraint(
+            Aws::S3::Model::BucketLocationConstraintMapper::GetBucketLocationConstraintForName(
+                config_.region));
+        request.SetCreateBucketConfiguration(std::move(configuration));
+    }
+
+    auto outcome = s3Client_->CreateBucket(request);
```
barneuman, line 314: In the case we create the bucket, shouldn't we also delete it on teardown?

w1ldptr: Why should we?

barneuman: Didn't you say we are trying to help the user avoid handling the AWS SDK? How does it achieve the goal if the user needs to clean up?

w1ldptr: Yeah, but why do we assume that the user wants to delete it all on shutdown?

PR #652, opened 2025-08-01
NIXL/GTEST: remove dead code

iyastreb:
## What?
Removed dead code

### DIFF
--- a/test/gtest/test_transfer.cpp
+++ b/test/gtest/test_transfer.cpp
@@ -245,28 +245,6 @@ protected:
         }
     }
 
-    void waitForXfer(nixlAgent &from, const std::string &from_name,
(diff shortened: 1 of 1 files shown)


yosefe:
Maybe revive this function? Move lines 369-382 from doTransfer() (the loop that waits for xfer status) inside it?

brminich (approved):

iyastreb:
> Maybe revive this function? Move lines 369-382 from doTransfer() (the loop that waits for xfer status) inside it?

I slightly modified doTransfer() in https://github.com/ai-dynamo/nixl/pull/573, to handle the case without progress thread.
So let me consider this refactoring after threadpool code is merged, to not introduce even more conflicts

PR #655, opened 2025-08-01
NIXLBENCH: Install arch specific AWS CLI - 0.5.0

ovidiusm:
Cherry pick of #650

## What?
Fixes a dependency installation which was incorrectly hardcoded for x86_64.


yosefe (approved):
maybe grep for "x86_64" in the whole code base to find more such places?

aranadive (approved):

PR #658, opened 2025-08-03
SRC/API/PYTHON: add pylint checks in CI

roiedanino:
## What?
Add pylint checks to CI, fix existing code according to warnings

## Why?
Pylint is an essential tool for finding common bugs, code duplication, and different kinds of errors that current tools won't find


examples/python/__init__.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yosefe, line 1: why need to run with python -m ?

src/api/python/_api.py
```diff
@@ -47,9 +47,12 @@ def __init__(
         enable_prog_thread: bool = True,
         enable_listen_thread: bool = False,
         listen_port: int = 0,
-        backends: list[str] = ["UCX"],
+        backends: Optional[list[str]] = None,
```
yosefe, line 55: why need to make it optional instead of default value?

pyproject.toml
```diff
@@ -35,3 +35,29 @@ profile = "black"
...
+    "R0912", # Too many branches
+    "R0913", # Too many arguments
+    "R0914", # Too many local variables
+    "R0915", # Too many statements
+    "R0917"  # Too many positional arguments,
+]
+
+[tool.pylint.typecheck]
+extension-pkg-allow-list = ["nixl._bindings", "nixl._utils"]
+ignored-modules = ["nixl._bindings", "nixl._utils", "nixl"]
```
yosefe: why need to ignore these?

src/api/python/_api.py
roiedanino: Lists in python are mutable objects and can't be set as a default value, as the same list will be shared across different calls, so the common solution is to set the default to None and then set it within the function

examples/python/__init__.py
roiedanino: Otherwise relative imports won't work - `from . import util`

yosefe: seems wait_for_transfer_completion is used only in 2 place so maybe just keep it as is without introducing utilities module?

roiedanino: But we would need to import the other module in the same way to avoid that code duplication. If you meant leaving the code duplication as it was, I don't think we want to suppress those pylint warnings.

Maybe I'll try using full path import instead

examples/python/query_mem_example.py
```diff
@@ -111,8 +115,6 @@
 
     except Exception as e:
         print(f"Error in example: {e}")
```
ovidiusm: This would be nicer with logging.exception

pyproject.toml
```diff
@@ -35,3 +35,29 @@ profile = "black"
 
 [tool.meson-python.args]
 setup = ['-Dinstall_headers=false']
+
+[tool.pylint]
```
ovidiusm, line 39: I'm wondering if we should consider using https://github.com/astral-sh/ruff instead of pylint

examples/python/query_mem_example.py
roiedanino: Yes it is fixed in my other PR which has a higher priority IMO

pyproject.toml
roiedanino: It seems like it's mostly covering flake8 checks

brminich:
@roiedanino is it relevant? can you rebase it pls?

PR #614, opened 2025-07-23
Nixlbench: build in release mode by default in container

ovidiusm:
## What?
Add option for build mode in the container, build nixl and nixlbench in release mode by default

## Why?
Seeing up to 1-10% better performance in DRAM tests (more for smaller buffers)

PR #661, opened 2025-08-04
Fix not deleted\initialized  memory

rafiw:
## What?
fix memory issues

## Why?
since it's needed


src/core/nixl_agent.cpp
```diff
@@ -91,6 +91,8 @@ nixlAgentData::nixlAgentData(const std::string &name,
         useEtcd = false;
         NIXL_DEBUG << "NIXL ETCD is disabled";
     }
+#else
```
brminich: can we do
```
#if HAVE_ETCD
    if (getenv("NIXL_ETCD_ENDPOINTS")) {
        useEtcd = true;
        NIXL_DEBUG << "NIXL ETCD is enabled";
    } else
#endif
     {
        useEtcd = false;
        NIXL_DEBUG << "NIXL ETCD is disabled";
    }
```

src/plugins/gpunetio/gpunetio_backend.cpp
```diff
@@ -404,6 +405,8 @@ nixlDocaEngine::nixlDocaInitNotif (const std::string &remote_agent,
     delete notif->recv_mmap;
     delete notif->recv_barr;
 
+    delete notif;
```
brminich: seems aldo need to free `notif->send_addr`

rafiw: also recv_addr

src/plugins/obj/obj_backend.cpp
```diff
@@ -159,7 +159,6 @@ nixl_status_t
 nixlObjEngine::deregisterMem(nixlBackendMD *meta) {
     nixlObjMetadata *obj_md = static_cast<nixlObjMetadata *>(meta);
     if (obj_md) {
-        std::unique_ptr<nixlObjMetadata> obj_md_ptr = std::unique_ptr<nixlObjMetadata>(obj_md);
```
w1ldptr, line 162: Why are you removing this line? Who's deleting the metadata instance then?

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -89,6 +90,7 @@ RUN echo "INFO: Starting custom UCX build..." && \
     ./autogen.sh && \
     echo "INFO: Building UCX..." && \
     ./contrib/configure-release --with-cuda=/usr/local/cuda \
+                                --enable-debug \
```
ovidiusm: I see why you need it but this is not the right way to do it. We want nixlbench container builds to happen in release mode by default, so that we can do performance testing. If we need debug mode, it should be a parameter.

I would rather add the changes in this file to https://github.com/ai-dynamo/nixl/pull/614/ Would that be OK with you?

ovidiusm (changes requested):

benchmark/nixlbench/contrib/Dockerfile
rafiw: this was by mistake

src/plugins/obj/obj_backend.cpp
rafiw: your right delete is needed

w1ldptr: Again, is there a problem with this code? Why are you substituting RAII-style memory management with inferior manual object deletion?

rafiw:
will be fixed as part of #733

PR #663, opened 2025-08-04
UTILS/SERDES: New implementation (for previous API)

ColinNV:
## What?
 - First PR for new wire format:
 - Basic new wire format.

## Why?
Requirement for more compact wireformat and backward compatible metadata...

## Note
This is just the first PR, planned next are
- more compact encoding of integers,
- "skip" feature for backwards compatibility,
- modest interface improvents.


test/unit/utils/serdes/serdes_test.cpp
```diff
@@ -36,30 +41,501 @@ int main() {
...
+        char dummy;
+        assert(sd1.getBuf("l", &dummy, 1) != NIXL_SUCCESS);
+        assert(sd1.remainingSize() == size);
+    }
+}
+
+} // namespace
 
+int
+main() {
```
iyastreb, line 270: As before, I suggest to use gtest framework

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
-nixlSerDes::nixlSerDes() {
-    workingStr = "nixlSerDes|";
-    des_offset = 11;
+namespace nixl {
+namespace {
+    inline constexpr size_t varLenBias = 127;
 
-    mode = SERIALIZE;
-}
+    const std::string_view header1 = "N1XL";
```
iyastreb, line 24: maybe just `header`? Or we plan to add `header2` and more?

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
+    template<typename...> inline constexpr bool dependentFalse = false;
 
-void nixlSerDes::_stringToBytes(void* fill_buf, const std::string &s, ssize_t size){
-    s.copy(reinterpret_cast<char*>(fill_buf), size);
-}
+    template<typename Int>
+    [[nodiscard]] Int
+    reconstruct(const void *buf) noexcept {
+        Int result;
+        std::memcpy(&result, buf, sizeof(result));
```
iyastreb: I'm wondering if it's possible to do the same without memcpy, by just reinterpret_cast?

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
+            tmp[++used] = uint8_t(size);
+            size >>= 8;
+        }
+        tmp[0] = (used | char(0xC0));
+        addRaw(buffer, tmp, 1 + used);
+    }
+
+    void
+    addLen(std::string &buffer, const size_t size) {
+        if (size < 128) {
```
iyastreb: should we use varLenBias constant here?

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
 
-    ssize_t len;
+size_t
+nixlSerDes::getTagLen() {
+    const uint8_t first = getFirst();
 
-    //get len
-    //_stringToBytes(&len, workingStr.data() + des_offset + tag.size(), sizeof(ssize_t));
-    _stringToBytes(&len, workingStr.substr(des_offset + tag.size(), sizeof(ssize_t)), sizeof(ssize_t));
+    if (first < 128) {
```
iyastreb: maybe use that constant here as well

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
+        uint8_t tmp[sizeof(value) + 1];
+        uint8_t used = 0;
+
+        for (Int v = value; v > 0; v >>= 8) {
+            tmp[++used] = uint8_t(v);
+        }
+        if (used == sizeof(Int)) {
+            addRawInt(buffer, value);
+        } else if constexpr (sizeof(Int) == 2) {
+            tmp[0] = 0xD0; // used == 1 -> D0
```
iyastreb: Since 0xDO used in several places, better add a constant with descriptive name
And same for other constants like C0..

src/utils/serdes/serdes.cpp
```diff
@@ -14,121 +14,483 @@
...
+nixlSerDes::skipValue() {
+    const uint8_t first = getFirst();
+
+    if (first < 128) {
+        require(first);
+        consume(first);
+        return;
+    }
+    switch (first & 0xF0) {
+    case 0x80:
```
iyastreb: I would appreciate some comments, it's not very clear what these constance are and what is the meaning

src/utils/serdes/serdes.cpp
ColinNV: AFAIK the `memcpy` is the only standard compliant way and it will work on all architectures, even those that don't, or don't efficiently, support unaligned memory access. And since the size is known at compile-time we can assume modern compilers are smart enough to "do the right thing" on architectures that do support efficient unaligned access.

test/unit/utils/serdes/serdes_test.cpp
ColinNV: Will be done at the end.

src/utils/serdes/serdes.cpp
ColinNV: As they say, forever is a long time...

PR #671, opened 2025-08-07
NIXLBENCH: Add support for GDS_MT backend

aranadive:
## What?
GDS_MT backend for NIXLBench. Fixes #667.

## Why?
Testing GDS_MT backend using NIXLBench. Add thread_count backend parameter support as well.

PR #672, opened 2025-08-07
fix: nixlbench support for GDS_MT

### LINKED ISSUE #667: nixlbench: add support for GDS_MT
--backend=GDS_MT is currently not supported in nixlbench.
https://github.com/ai-dynamo/nixl/blob/160a4349ccc4979dcaa3a94be9845beaece3b8d8/benchmark/nixlbench/src/utils/utils.cpp#L44-L47
Would be great if this can be added.
Thanks

da-x:
Also clarify some prints regarding invalid backend specification.

Fixes #667.


aranadive:
#671 does same.

PR #656, opened 2025-08-01
Handle UCS_ERR_CANCELED as NIXL_ERR_REMOTE_DISCONNECT

ovidiusm:
## What?
Unrecognized error code in test failure for error handling on x86_64 + GPU.

## Why?

Failing test:

```
docker run --privileged  --device=/dev/infiniband  --net=host --ipc=host --gpus all  -e NVIDIA_VISIBLE_DEVICES=all -e NIXL_LOG_LEVEL=debug -e UCX_LOG_LEVEL=info db92b0bc4073  /usr/local/nixl/bin/gtest --gtest_filter=UCX/TestErrorHandling.LoadRemoteThenFail/0

==========
== CUDA ==
==========

NVIDIA Release  (build )
CUDA Version 12.8.1.012
Container image Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

Various files include modifications (c) NVIDIA CORPORATION & AFFILIATES.  All rights reserved.

GOVERNING TERMS: The software and materials are governed by the NVIDIA Software License Agreement
(found at https://www.nvidia.com/en-us/agreements/enterprise-software/nvidia-software-license-agreement/)
and the Product-Specific Terms for NVIDIA AI Products
(found at https://www.nvidia.com/en-us/agreements/enterprise-software/product-specific-terms-for-ai-products/).

Note: Google Test filter = UCX/TestErrorHandling.LoadRemoteThenFail/0
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from UCX/TestErrorHandling
[ RUN      ] UCX/TestErrorHandling.LoadRemoteThenFail/0
I0731 18:18:03.454256       1 nixl_agent.cpp:92] NIXL ETCD is disabled
I0731 18:18:03.454535       1 nixl_plugin_manager.cpp:197] Loading plugins from file: /workspace/nixl/build/pluginlist
I0731 18:18:03.552022       1 nixl_plugin_manager.cpp:206] Loading plugins from: /workspace/nixl/build/src/plugins/ucx
I0731 18:18:03.552188       1 nixl_plugin_manager.cpp:316] Discovered and loaded plugin: UCX
I0731 18:18:03.552534       1 config.cpp:43] Modified UCX config: ADDRESS_VERSION=v2
I0731 18:18:03.552558       1 config.cpp:43] Modified UCX config: RNDV_THRESH=inf
[...]

PR #677, opened 2025-08-11
Handle Xfer canсeled error accordingly to connection status

evgeny-leksikov:
## What?
Handle Xfer canсeled error accordingly to connection status

## Why?
https://github.com/ai-dynamo/nixl/pull/656

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -61,6 +61,8 @@ std::string nixlEnumStrings::statusStr (const nixl_status_t &status) {
         case NIXL_ERR_UNKNOWN:           return "NIXL_ERR_UNKNOWN";
         case NIXL_ERR_NOT_SUPPORTED:     return "NIXL_ERR_NOT_SUPPORTED";
         case NIXL_ERR_REMOTE_DISCONNECT: return "NIXL_ERR_REMOTE_DISCONNECT";
+        case NIXL_ERR_CANCELED:
+            return "NIXL_ERR_CANCELED";
         default:                         return "BAD_STATUS";
     }
 }


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1115,6 +1131,10 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
             return NIXL_ERR_INVALID_PARAM;
         }
 
+        if (ret == NIXL_IN_PROG) {
+            nixlUcxReqSetConnection(req, rmd->conn);
+        }
```
evgeny-leksikov: probably should be part of `_retHelper`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -279,17 +279,43 @@ void nixlUcxEngine::vramFiniCtx()
 class nixlUcxIntReq : public nixlLinkElem<nixlUcxIntReq> {
     private:
         int _completed;
+        ucx_connection_ptr_t _conn;
```
iyastreb: According to rules, it should be `conn_`
You can check with clang-tidy

evgeny-leksikov: I made only that code style changes which required to pass CI to avoid comments like this: https://github.com/ai-dynamo/nixl/pull/509#discussion_r2265137307

IMO it's weird to have `conn_` and `_completed` in the same class

iyastreb: I agree, it's weird, same happened to my PR.
But in this particular case it's a good opportunity to rename `_completed` and change its type to `bool`:
`bool completed_;`
Anyway you changed is_complete and completed functions

iyastreb (dismissed):
LGTM
Is there an easy way to test these changes?

evgeny-leksikov:
> LGTM Is there an easy way to test these changes?

I don't think so, the issue very sensitive to UCX timings and protocols, stable reproducible on "soul" setup but I dont know simple way to test it on nixl level.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1159,7 +1186,8 @@ nixl_status_t nixlUcxEngine::checkXfer (nixlBackendReqH* handle) const
         nixlUcxReq req;
         status = notifSendPriv(notif->agent, notif->payload, req, workerId);
         notif.reset();
-        if (_retHelper(status, intHandle, req)) {
+        // TODO: conn lookup
```
ovidiusm, line 1189: Is the lookup missing here?

evgeny-leksikov: I going to add this in separate PR, I touched this place when decided to reuse `_retHelper`. Also `notifSendPriv` already does that lookup, so need to see how we can avoid double lookup in this particular case. So, I decided to leave `TODO` for now and don't grow the PR with unrelated changes.

ovidiusm (approved):

PR #681, opened 2025-08-13
Bugfix: Replace select() with poll() in connectToIP to prevent stack smashing

michal-shalev:
## What?
Fixed stack smashing detected crash in connectToIP() by replacing `select()` with `poll()`.

## Why?
The `select()` system call crashes when file descriptors exceed `FD_SETSIZE (1024)`. The crash occurred when `connectToIP()` received `fd>1024`, causing `FD_SET()` to write beyond the `fd_set` structure bounds.
According to the Linux manual: "All modern applications should instead use poll(2) or epoll(7), which do not suffer this limitation."
https://man7.org/linux/man-pages/man2/select.2.html

## How?
Replaced `select()` with `poll()` using` POLLOUT event` and `1000ms` timeout. Maintains identical functionality while supporting unlimited file descriptor numbers.

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -27,6 +27,7 @@
 #include <future>
 #endif // HAVE_ETCD
 #include <absl/strings/str_format.h>
+#include <poll.h>
 
 const std::string default_metadata_label = "metadata";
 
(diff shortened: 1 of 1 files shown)


src/core/nixl_listener.cpp
```diff
@@ -22,7 +22,7 @@
 #include "agent_data.h"
 #include "common/nixl_log.h"
 #if HAVE_ETCD
-#include <etcd/Client.hpp>
```
ovidiusm: Could you please apply the patch on top of recent main? Then you will not see all these etcd changes.

ovidiusm:
Could we also add the return error value check for inet_ntop?

src/core/nixl_listener.cpp
michal-shalev: done

ovidiusm (approved):

tvegas1 (approved):

src/core/nixl_listener.cpp
```diff
@@ -447,7 +451,12 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
                 socklen_t client_addrlen = sizeof(client_address);
                 if (getpeername(new_fd, (sockaddr*)&client_address, &client_addrlen) == 0) {
                     char client_ip[INET_ADDRSTRLEN];
-                    inet_ntop(AF_INET, &client_address.sin_addr, client_ip, INET_ADDRSTRLEN);
+                    if (inet_ntop(AF_INET, &client_address.sin_addr, client_ip, INET_ADDRSTRLEN) ==
+                        nullptr) {
+                        NIXL_ERROR << "inet_ntop failed for client address";
```
brminich: pls use `NIXL_PERROR` to print also errno

michal-shalev: done

brminich (dismissed):

ovidiusm (approved):

PR #685, opened 2025-08-13
BACKEND/UCX: cleanup AM based control path

evgeny-leksikov:
## What?
cleanup AM based control path

## Why?
since the UCX backend supports native UCX peer failure error handling, AM based connections management is redundant.


evgeny-leksikov:
> maybe keep it, given that error handling mode can be changed with a parameter?

looking into the current code I don't see cases when it could guarantee handling of peer failure with disabled peer failure err handling support in UCX level.

yosefe (approved):

Issue #636, opened 2025-07-29
Crashed when running nixlbench using ucx backend with multiple threads


Jye-525:
The program crashed when I ran `nixlbench` using UCX backend with multiple threads.
I enabled `--enable-mt` when installing UCX.

Here are the commands:
On node 1: 
```
$ ./nixlbench --worker_type nixl --runtime_type=ETCD --etcd-endpoints http://10.52.3.209:2379 --backend UCX -enable_pt --initiator_seg_type VRAM --target_seg_type VRAM --max_batch_size 1 --start_batch_size 1 --num_threads 4 --total_buffer_size 8589934592
```

On node 2: 
```
$ ./nixlbench --worker_type nixl --runtime_type=ETCD --etcd-endpoints http://10.52.3.209:2379 --backend UCX -enable_pt --initiator_seg_type VRAM --target_seg_type VRAM --max_batch_size 1 --start_batch_size 1 --num_threads 4 --total_buffer_size 8589934592
```
Here are the errors when nixlbench crashed:
```
        [nixl-2:1886795:0:1886972] proto_select.c:549  Assertion `khret == UCS_KH_PUT_BUCKET_EMPTY' failed
	[nixl-2:1886795:1:1886973] proto_select.c:549  Assertion `khret == UCS_KH_PUT_BUCKET_EMPTY' failed
	[nixl-2:1886795:2:1886974] proto_select.c:549  Assertion `khret == UCS_KH_PUT_BUCKET_EMPTY' failed
	==== backtrace (tid:1886973) ====
	 0  /home/cc/shared_cc/deps/ucx-1.18.0/install/lib/libucs.so.0(ucs_handle_error+0x304) [0x74ea67d83074]
	 1  /home/cc/shared_cc/deps/ucx-1.18.0/install/lib/libucs.so.0(ucs_fatal_error_message+0xc2) [0x74ea67d7ff22]
	 2  /home/cc/shared_cc/deps/ucx-1.18.0/install/lib/libucs.so.0(ucs_fatal_error_format+0x122) [0x74ea67d80052]
	 3  /home/cc/shared_cc/deps/ucx-1.18.0/install/lib/libucp.so.0(ucp_proto_select_lookup_slow+0x7e3) [0x74ea4c6ede83]
	 4  /home/cc/shared_cc/deps/ucx-1.18.0/install/lib/libucp.so.0(ucp_put_nbx+0x14c3) [0x74ea4c7046c3]
[...]

PR #686, opened 2025-08-13
NIXLBENCH: Set RW sync mode for UCX when num_threads > 1

iyastreb:
## What?
nixlbench [crashes](https://github.com/ai-dynamo/nixl/issues/636) with UCX backend when `--num_threads` > 1

## Why?
NIXL sets proper UCX worker lock mode only when progress thread is set (`--enable_pt` option), but not when multiple threads are being used.

### DIFF
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -63,19 +63,18 @@ class nixlAgentConfig {
          * @param pthr_delay_us      Optional delay for pthread in us
          * @param lthr_delay_us      Optional delay for listener thread in us
          */
-        nixlAgentConfig (const bool use_prog_thread,
-                         const bool use_listen_thread=false,
-                         const int port=0,
-                         nixl_thread_sync_t sync_mode=nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
-                         unsigned int num_workers = 1,
-                         const uint64_t pthr_delay_us=0,
-                         const uint64_t lthr_delay_us = 100000) :
-                         useProgThread(use_prog_thread),
-                         useListenThread(use_listen_thread),
-                         listenPort(port),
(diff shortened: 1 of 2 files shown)


src/api/cpp/nixl_params.h
```diff
@@ -63,19 +63,18 @@ class nixlAgentConfig {
...
-                         useListenThread(use_listen_thread),
-                         listenPort(port),
-                         syncMode(sync_mode),
-                         pthrDelay(pthr_delay_us),
-                         lthrDelay(lthr_delay_us) { }
+        nixlAgentConfig(const bool use_prog_thread,
+                        const bool use_listen_thread = false,
+                        const int port = 0,
+                        nixl_thread_sync_t sync_mode = nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
+                        const uint64_t pthr_delay_us = 0,
```
brminich: pls do not change the API

brminich (changes requested):

src/api/cpp/nixl_params.h
iyastreb: ok, reverted

test/gtest/test_transfer.cpp
```diff
@@ -103,6 +103,7 @@ class TestTransfer :
                                listen_port > 0,
                                listen_port,
                                nixl_thread_sync_t::NIXL_THREAD_SYNC_RW,
+                               1,
```
brminich, line 106: it is unrelated bugfix right?

aranadive (approved):

test/gtest/test_transfer.cpp
iyastreb: Yes. Just found it when checking nixlAgentConfig

PR #688, opened 2025-08-14
when ucx post op and return in prog, notify info should be saved

Jazel-Z:
## What?
avoid notify msg drop when ucx op in async 

## How?
save notify msg info when return, let checkxfer func send notify msg


brminich:
closing, as the code was significantly refactored

PR #690, opened 2025-08-15
BACKEND/UCX: fix req initialization in checkXfer

evgeny-leksikov:
## What?
- initialize `req->conn` in `checkXfer`
- added a test
- more few fixes with the new test
- Fix for https://nvbugspro.nvidia.com/bug/5472171

## Why?
 - Close the debt https://github.com/ai-dynamo/nixl/pull/677#discussion_r2277670415
 - https://github.com/ai-dynamo/nixl/pull/690#pullrequestreview-3127128611

## How?
move connection lookup to a helper method `getConnection` and pass the result to `notifSendPriv`

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1364,25 +1364,23 @@ nixl_status_t nixlUcxEngine::unloadMD (nixlBackendMD* input) {
 *****************************************/
 
 static nixl_status_t
-_retHelper(nixl_status_t ret,
-           nixlUcxBackendH *hndl,
-           nixlUcxReq &req,
-           ucx_connection_ptr_t conn = nullptr) {
+_retHelper(nixl_status_t ret, nixlUcxBackendH *hndl, nixlUcxReq &req, ucx_connection_ptr_t conn) {
     /* if transfer wasn't immediately completed */
     switch(ret) {
-        case NIXL_IN_PROG:
-            // TODO: this cast does not look safe
-            // We need to allocate a vector of nixlUcxIntReq and set nixlUcxReqt
-            hndl->append((nixlUcxIntReq *)req);
-            nixlUcxReqSetConnection(req, conn);
-        case NIXL_SUCCESS:
-            // Nothing to do
-            break;
-        default:
-            // Error. Release all previously initiated ops and exit:
-            hndl->release();
-            return ret;
+    case NIXL_IN_PROG:
+        // TODO: this cast does not look safe
+        // We need to allocate a vector of nixlUcxIntReq and set nixlUcxReqt
+        hndl->append((nixlUcxIntReq *)req);
+        nixlUcxReqSetConnection(req, conn);
+    case NIXL_SUCCESS:
+        // Nothing to do
+        break;
+    default:
+        // Error. Release all previously initiated ops and exit:
+        hndl->release();
+        return ret;
     }
+
     return NIXL_SUCCESS;
 }
 
@@ -1564,7 +1562,8 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
         if (ret == NIXL_SUCCESS) {
             nixlUcxReq req;
             auto rmd = (nixlUcxPublicMetadata *)remote[0].metadataP;
-            ret = notifSendPriv(remote_agent, opt_args->notifMsg, req, int_handle->getWorkerId());
+            ret = notifSendPriv(
+                remote_agent, opt_args->notifMsg, req, rmd->conn->getEp(int_handle->getWorkerId()));
             if (_retHelper(ret, int_handle, req, rmd->conn)) {
                 return ret;
             }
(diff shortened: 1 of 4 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1301,7 +1301,7 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
     nixlUcxReq req;
     size_t wid = getWorkerId();
 
-    ret = notifSendPriv(remote_agent, msg, req, wid);
+    ret = notifSendPriv(remote_agent, msg, req, getConnection(remote_agent)->getEp(wid));
```
yosefe: in this case if not found there will be segv instead of error mesage

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1178,23 +1176,28 @@ nixl_status_t nixlUcxEngine::postXfer (const nixl_xfer_op_t &operation,
...
 
-        status = intHandle->status();
+    if ((handle_status != NIXL_SUCCESS) || !notif.has_value()) {
+        return handle_status;
     }
 
-    return status;
+    ucx_connection_ptr_t conn = getConnection(notif->agent);
+    if (!conn) {
+        return NIXL_ERR_NOT_FOUND;
```
yosefe, line 1599: previously there was notif.reset(); in this flow

evgeny-leksikov: hmm... should we then `notif.reset()` in check above? for case if handle_status == NIXL_ERR_* and notif.has_value() (the last can be skipped for simplicity). smth like 
```
    if ((handle_status != NIXL_SUCCESS) || !notif.has_value()) {
        if (handle_status != NIXL_IN_PROG) { // check error flow
            notif.seret();
        }
        return handle_status;
    }
```

evgeny-leksikov:
> is this PR a bug fix? If yes can we add gtest?

yes, but seems like this flow is not covered yet, I'll see how to improve

iyastreb (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -790,7 +790,7 @@ class nixlUcxCompositeBackendH : public nixlUcxBackendH {
 
     size_t
     getNumChunks() const {
-        return sharedState_->chunks.size();
+        return sharedState_.get() ? sharedState_->chunks.size() : 0;
```
yosefe:
```suggestion
        return sharedState ? sharedState_->chunks.size() : 0;
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -816,12 +816,14 @@ class nixlUcxCompositeBackendH : public nixlUcxBackendH {
     release() override {
         NIXL_TRACE << *this << " releasing";
         nixl_status_t status = nixlUcxBackendH::release();
-        // Set failed status to stop progress chunks
-        sharedState_->status.store(NIXL_ERR_NOT_FOUND);
+        if (sharedState_.get()) {
```
yosefe:
```suggestion
        if (sharedState) {
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1581,23 +1582,33 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
...
+    nixlUcxReq req;
+    nixl_status_t status =
+        notifSendPriv(notif->agent, notif->payload, req, conn->getEp(intHandle->getWorkerId()));
+    notif.reset();
+    status = _retHelper(status, intHandle, req, conn);
+    if (status != NIXL_SUCCESS) {
+        return status;
+    }
+
+    return intHandle->status();
```
yosefe, line 1611: isn't this supposed to be NIXL_SUCCESS here always, because of line 1588?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1702,9 +1710,14 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
     nixlUcxReq req;
-    size_t wid = getWorkerId();
 
-    ret = notifSendPriv(remote_agent, msg, req, wid);
+    auto conn = getConnection(remote_agent);
+    if (!conn) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    size_t wid = getWorkerId();
```
yosefe: IMO the wid var is redundant

src/plugins/ucx/ucx_backend.cpp
yosefe: also, what changed in this PR that need to check sharedState_ is valid?

iyastreb: This problem existed even before this PR.
The problem is that we may call `operator<<` on a released (but not destroyed) object of this class, which leads to a crash, because `operator<<` calls `getNumChunks()` => access sharedState(nullptr)

src/plugins/ucx/ucx_backend.cpp
evgeny-leksikov: nope, `intHandle->status();` is a bit complicated than just a getter, I tried this change and it leads to regression in tests. You can find this call at line 1597 before this PR.

roiedanino (approved):

PR #694, opened 2025-08-18
NIXL/UCX: Replace homemade list of requests with vector

iyastreb:
## What?
Replace homemade list of requests with vector

## Why?
to make iteration cache-friendly and reduce allocations, and make code simpler

### DIFF
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -78,7 +78,7 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=1 ./bin/agent_example &
+NIXL_TELEMETRY_ENABLE=y ./bin/agent_example &
 sleep 1
 ./bin/telemetry_reader /tmp/Agent001 &
 telePID=$!
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -79,7 +79,7 @@ python3 partial_md_example.py --etcd
 python3 query_mem_example.py
 
 # Running telemetry for the last test
-export NIXL_TELEMETRY_ENABLE=1
+export NIXL_TELEMETRY_ENABLE=y
 blocking_send_recv_port=$(get_next_tcp_port)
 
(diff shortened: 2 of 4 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -278,34 +278,34 @@ void nixlUcxEngine::vramFiniCtx()
...
 
 
-class nixlUcxIntReq : public nixlLinkElem<nixlUcxIntReq> {
-    private:
-        bool completed_ = false;
-        ucx_connection_ptr_t conn_;
+class nixlUcxIntReq {
+private:
+    bool completed_ = false;
+    ucx_connection_ptr_t conn_;
```
rakhmets: You can move the private fields after the public section, since all the lines in the class have been modified anyway.

rakhmets (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -428,24 +423,18 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-            nixlUcxIntReq *next_req = req->unlink();
+        size_t incomplete_reqs = 0;
+        for (nixlUcxIntReq *req : requests_) {
             if (req->is_complete()) {
                 _internalRequestReset(req);
                 worker->reqRelease((nixlUcxReq)req);
             } else {
-                /* Enqueue back */
-                append(req);
+                requests_[incomplete_reqs++] = req;
```
evgeny-leksikov, line 434: should we care about ordering?

iyastreb: not really, but still the original insertion ordering is maintained

evgeny-leksikov (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -428,24 +423,18 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
             }
-            req = next_req;
         }
-
+        requests_.resize(incomplete_reqs);
```
brminich, line 437: can it affect perf? or we assume new size is always smaller?

iyastreb: it's always smaller, therefore this call just decrements the size variable

PR #695, opened 2025-08-18
[HF3FS] add mempool

Pyngon:
## What?
Use mempool in HF3FS plugin

## Why?
Avoid allocating memory in datapath

### DIFF
--- a/src/plugins/hf3fs/hf3fs_backend.cpp
+++ b/src/plugins/hf3fs/hf3fs_backend.cpp
@@ -71,7 +71,7 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
                     iopool_size = size;
                 } else {
                     iopool_size = HF3FS_MAX_IOPOOL_SIZE;
-                    NIXL_INFO << size << " exceeded max iopool size " << iopool_size
+                    NIXL_WARN << size << " exceeded max iopool size " << iopool_size
                               << ", set it to max";
                 }
             }
@@ -97,7 +97,7 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
         iopool.push_back(io);
     }
 
-    NIXL_DEBUG << "HF3FS: Page size: " << page_size;
+    NIXL_DEBUG << "HF3FS: page size " << page_size << " iopool_size " << iopool_size;
 }
 
 nixlHf3fsIO *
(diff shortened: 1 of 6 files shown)


src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -72,9 +87,46 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
...
+nixlHf3fsIO*
+nixlHf3fsEngine::getFromIOPool() const
+{
+    const std::lock_guard<std::mutex> lock(iopool_lock);
+    if (!iopool.empty()) {
+        auto io = iopool.front();
+        iopool.pop_front();
+        return io;
+    }
+    return nullptr;
```
zackyoray, line 111: should we support fallback for empty pool?
it seems that with default configuration if we use batch size larger than 64 we will fail.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -237,7 +289,7 @@ nixl_status_t nixlHf3fsEngine::prepXfer (const nixl_xfer_op_t &operation,
         offset = (size_t) (*file_list)[i].addr;  // Offset in file
         auto mem_md = (nixlHf3fsMetadata *)(*mem_list)[i].metadataP;
 
-        nixlHf3fsIO *io = new nixlHf3fsIO();
+        nixlHf3fsIO *io = getFromIOPool();
         if (io == nullptr) {
             nixl_err = NIXL_ERR_BACKEND;
             nixl_mesg = "Error: Failed to create IO";
```
zackyoray: maybe use more specific error msg, regarding that io pool is empty.

Pyngon: Update the message.

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Modified to alloc a new io when pool is empty.

zackyoray (approved):
Looks good to me.
Thanks.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -61,6 +64,18 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
                 return;
             }
         }
+        if (init_params->customParams->count("iopool_size") > 0) {
```
aranadive, line 67: Can you add these parameters to nixlbench and 3fs test example?

Pyngon: Sure

Pyngon: Done

aranadive (approved):

PR #701, opened 2025-08-19
Removed progress method and its indicator from SB API.

mkhazraee:
## What?
Removes progress() and supportsProgTh() from SB API. They can be still used within the plugin.

## Why?
As ucx and gds_mt are already using multi-threads to deliver performance, idea of shared progress thread became less useful. Shared progress thread can also cause contention between the backends, while we want the isolation in modularity as much as possible. So team decided to remove it from the SB API. Plugins internally can still use it, such as current ucx implementation, just not a mandatory method in SB API anymore.


docs/BackendGuide.md
```diff
@@ -213,20 +205,20 @@ Note that inside a transfer, a backend might provide methods for network resilie
 
 ### Get transfer status:
 
-The agent will call the backend specific transfer handle that is stored within the agent transfer handle, and check the status of the transfer. This is achieved through a call to **checkXfer** in the SB API. Internal to the backend, they can call the **progress** method in SB API, if that’s necessary to get the latest status of the transfers. If the agent is run in progress thread mode, the agent will call that periodically, and therefore reduce the load on this internal call.
+The agent will call the backend specific transfer handle that is stored within the agent transfer handle, and check the status of the transfer. This is achieved through a call to **checkXfer** in the SB API. Internal to the backend, they can call their internall progress method, if that’s necessary to get the latest status of the transfers.
```
iyastreb: typo: internal

docs/BackendGuide.md
```diff
@@ -213,20 +205,20 @@ Note that inside a transfer, a backend might provide methods for network resilie
...
+The agent will iterate over all the backends that support notification, and call their **getNotifs** from the SB API, which will return a list of notifications received from each remote node between the previous call to this method and this time. Then the agent will merge the results from all such backends, and append them to the map that the user has provided. Similar to get transfer status, Internal to the backend, they can call ther internall progress method, if that’s necessary to get the latest notifications received from the transfers initiated by the other agents towards them.
```
iyastreb: typo: their internal

test/unit/plugins/ucx/ucx_backend_multi.cpp
```diff
@@ -52,7 +52,7 @@ void test_thread(int id)
 
     ucx = nixlUcxEngine::create(init_params).release();
 
-    if(!USE_PTHREAD) ucx->progress();
+    if (!USE_PTHREAD) ((nixlUcxEngine *)ucx)->progress();
```
iyastreb: Instead of casting we can change the variable type to be:
`nixlUcxEngine *ucx`

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -414,7 +414,7 @@ void performTransfer(nixlBackendEngine *ucx1, nixlBackendEngine *ucx2,
         while(ret3 == NIXL_IN_PROG) {
             ret3 = ucx1->checkXfer(handle);
             if(progress){
-                ucx2->progress();
+                ((nixlUcxEngine *)ucx2)->progress();
```
iyastreb: same here, since this test is for ucx only

test/unit/plugins/ucx/ucx_backend_multi.cpp
mkhazraee: The idea back then was we want to make sure the methods work from base backendEngine class. With override keyword that should be kind of tested so less of a big deal. 
That being said, the progress should be called from checkXfer or getNotifs within the backend, not externally. So not sure if doing this test makes sense anymore. But if you think good to keep it around, I suggest to make this one just exception downcasting.

yosefe: This is UCX only test, only UCX now has "progress" method, it should be
`auto ucx_engine = nixlUcxEngine::create(init_params);`
And use nixlUcxEngine type it places that need to call progress() method

brminich (approved):

PR #702, opened 2025-08-20
Agent: Add getXferTelemetry API.

mkhazraee:
## What?
Add getXferTelemetry API to nixl Agent.

## Why?
Another mode of outputting telemetry, getting information per xfer_req.

### DIFF
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -78,7 +78,7 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=y ./bin/agent_example &
+NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=/tmp ./bin/agent_example &
 sleep 1
 ./bin/telemetry_reader /tmp/Agent001 &
 telePID=$!
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -78,9 +78,9 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=/tmp ./bin/agent_example &
+NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=$PWD/telemetry ./bin/agent_example &
 sleep 1
-./bin/telemetry_reader /tmp/Agent001 &
+./bin/telemetry_reader $PWD/telemetry/Agent001 &
 telePID=$!
 sleep 6
 kill -s SIGINT $telePID
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -78,7 +78,7 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-mkdir /tmp/telemetry_test
+mkdir -p /tmp/telemetry_test
 NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=/tmp/telemetry_test ./bin/agent_example &
 sleep 1
 ./bin/telemetry_reader /tmp/telemetry_test/Agent001 &
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -78,9 +78,10 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=$PWD/telemetry ./bin/agent_example &
+mkdir /tmp/telemetry_test
+NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=/tmp/telemetry_test ./bin/agent_example &
 sleep 1
-./bin/telemetry_reader $PWD/telemetry/Agent001 &
+./bin/telemetry_reader /tmp/telemetry_test/Agent001 &
 telePID=$!
 sleep 6
 kill -s SIGINT $telePID
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -289,6 +289,17 @@ class nixlAgent {
         nixl_status_t
         getXferStatus (nixlXferReqH* req_hndl) const;
 
+
+        /**
+         * @brief  Get the telemetry data associated with `req_hndl`.
+         *
+         * @param  req_hndl        Transfer request handle obtained from makeXferReq/createXferReq
+         * @param  telemetry [out] Output telemetry information
+         * @return nixl_status_t   Error code if call was not successful
+         */
+        nixl_status_t
+        getXferTelemetry(const nixlXferReqH *req_hndl, nixl_xfer_telem_t &telemetry) const;
+
         /**
          * @brief  Query the backend associated with `req_hndl`. E.g., if for genNotif
          *         the same backend as a transfer is desired.
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -264,13 +264,13 @@ struct nixlXferTelemetry {
     /**
      * @var totalBytes Amount of bytes transferred in the request
      */
-    size_t totalBytes = 0;
+    size_t totalBytes;
 
     /**
      * @var descCount Number of descriptors in the transfer request.
      *      If any merging of descriptors were performed, it will be reflected here.
      */
-    size_t descCount = 0;
+    size_t descCount;
 };
 
 /**
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -241,7 +241,7 @@ using chrono_point_t = std::chrono::steady_clock::time_point;
 constexpr auto min_chrono_time = std::chrono::steady_clock::time_point::min();
 
 /**
- * @brief A typedefs for a period of time
+ * @brief A typedefs for a period of time in microseconds
  */
 using chrono_period_t = std::chrono::microseconds;
 
@@ -258,29 +258,29 @@ struct nixlXferTelemetry {
     /**
      * @var startTime Time that the transfer was posted
      */
-    chrono_point_t startTime;
+    chrono_point_t startTime = min_chrono_time;
 
     /**
      * @var postDuration Time it took to do the post operation
      */
-    chrono_period_t postDuration;
+    chrono_period_t postDuration = chrono_period_t(0);
 
     /**
      * @var xferDuration Time it took to complete the transfer
      *      if checkXferReq is called late, that might impact this result
      */
-    chrono_period_t xferDuration;
+    chrono_period_t xferDuration = chrono_period_t(0);
 
     /**
      * @var totalBytes Amount of bytes transferred in the request
      */
-    size_t totalBytes;
+    size_t totalBytes = 0;
 
     /**
      * @var descCount Number of descriptors in the transfer request.
      *      If any merging of descriptors were performed, it will be reflected here.
      */
-    size_t descCount;
+    size_t descCount = 0;
 };
 
 /**
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -243,18 +243,13 @@ constexpr auto min_chrono_time = std::chrono::steady_clock::time_point::min();
 /**
  * @brief A typedefs for a period of time in microseconds
  */
-using chrono_period_t = std::chrono::microseconds;
+using chrono_period_us_t = std::chrono::microseconds;
 
 /**
  * @struct nixlXferTelemetry
  * @brief A structure for telemetry output from agent API
  */
 struct nixlXferTelemetry {
-    /**
-     * @var backendType Type of the backend performing the transfer
-     */
-    nixl_backend_t backendType;
-
     /**
      * @var startTime Time that the transfer was posted
      */
@@ -263,13 +258,13 @@ struct nixlXferTelemetry {
     /**
      * @var postDuration Time it took to do the post operation
      */
-    chrono_period_t postDuration = chrono_period_t(0);
+    chrono_period_us_t postDuration = chrono_period_us_t(0);
 
     /**
      * @var xferDuration Time it took to complete the transfer
      *      if checkXferReq is called late, that might impact this result
      */
-    chrono_period_t xferDuration = chrono_period_t(0);
+    chrono_period_us_t xferDuration = chrono_period_us_t(0);
 
     /**
      * @var totalBytes Amount of bytes transferred in the request
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -236,11 +236,6 @@ using nixlGpuXferReqH = void;
 using chrono_point_t = std::chrono::steady_clock::time_point;
 
 /**
- * @brief A constant indicating min chrono_point_t value
- */
-constexpr auto min_chrono_time = std::chrono::steady_clock::time_point::min();
-
-/**
  * @brief A typedefs for a period of time in microseconds
  */
 using chrono_period_us_t = std::chrono::microseconds;
@@ -253,18 +248,18 @@ struct nixlXferTelemetry {
     /**
      * @var startTime Time that the transfer was posted
      */
-    chrono_point_t startTime = min_chrono_time;
+    chrono_point_t startTime;
 
     /**
      * @var postDuration Time it took to do the post operation
      */
-    chrono_period_us_t postDuration = chrono_period_us_t(0);
+    chrono_period_us_t postDuration;
 
     /**
      * @var xferDuration Time it took to complete the transfer
      *      if checkXferReq is called late, that might impact this result
      */
-    chrono_period_us_t xferDuration = chrono_period_us_t(0);
+    chrono_period_us_t xferDuration;
 
     /**
      * @var totalBytes Amount of bytes transferred in the request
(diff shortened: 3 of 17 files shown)


mkhazraee:
For now just the initial commit to discuss the API, still 4 items remaining:
- Add Python bindings
- Add Rust bindings
- Test performance and if necessary do optimizations (unlikely)
- Add a mode where python file is not generated based on NIXL_TELEMETRY_ENABLE env var, e.g., being set to "a" instead of "y"

src/core/nixl_agent.cpp
```diff
@@ -1049,6 +1049,23 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
     return req_hndl->status;
 }
 
+nixl_status_t
+nixlAgent::getXferTelemetry(const nixlXferReqH *req_hndl, nixl_xfer_telem_t &telemetry) const {
+
+    if (!data->telemetry_) return NIXL_ERR_NO_TELEMETRY;
```
roiedanino: Maybe add some `NIXL_WARN << "An xfer telemetry requested by user but telemetry configuration was not provided, consider using: <ENV NAME>"`

Otherwise, I think users may struggle with this

src/core/nixl_agent.cpp
```diff
@@ -1049,6 +1049,23 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    if (req_hndl->status < 0) return req_hndl->status;
+
+    // NIXL_SUCCESS or NIXL_IN_PROG, values are initialized
+    telemetry.backendType = req_hndl->engine->getType();
+    telemetry.startTime = req_hndl->telemetry.startTime_;
+    telemetry.postDuration = req_hndl->telemetry.postDuration_;
+    telemetry.xferDuration = req_hndl->telemetry.xferDuration_;
+    telemetry.totalBytes = req_hndl->telemetry.totalBytes_;
```
roiedanino: Any reason `telemetry` and `req_hndl->telemetry` are different objects? (startTime vs startTime_ and so on)

src/core/transfer_request.h
```diff
@@ -52,11 +48,12 @@ class nixlXferReqH {
...
         struct {
-            chrono_point_t startTime;
-            microseconds postDuration_ = microseconds(0);
-            microseconds xferDuration_ = microseconds(0);
-            size_t totalBytes;
+            chrono_point_t startTime_ = min_chrono_time;
+            chrono_period_t postDuration_ = chrono_period_t(0);
+            chrono_period_t xferDuration_ = chrono_period_t(0);
+            size_t totalBytes_;
         } telemetry;
```
roiedanino: Can we use `nixl_xfer_telem_t` here instead of redefining it? much easier to maintain IMO

src/core/nixl_agent.cpp
```diff
@@ -941,7 +941,7 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
         return NIXL_ERR_INVALID_PARAM;
     }
 
-    if (data->telemetry_) req_hndl->telemetry.startTime = std::chrono::steady_clock::now();
+    if (data->telemetry_) req_hndl->telemetry.startTime_ = std::chrono::steady_clock::now();
```
roiedanino:
```suggestion
    if (data->telemetry_) {
        req_hndl->telemetry.startTime_ = std::chrono::steady_clock::now();
    }
```

src/api/cpp/nixl_types.h
```diff
@@ -223,6 +225,65 @@ struct nixlAgentOptionalArgs {
...
+
+/**
+ * @brief A constant indicating min chrono_point_t value
+ */
+constexpr auto min_chrono_time = std::chrono::steady_clock::time_point::min();
+
+/**
+ * @brief A typedefs for a period of time
+ */
+using chrono_period_t = std::chrono::microseconds;
```
roiedanino: maybe `chrono_time_us_t` or even just `chrono_us_t`?

src/core/nixl_agent.cpp
mkhazraee: That's the main optimization I was thinking of. backendType and desc_count are not in the internal telemetry, but we can totally add them during xferReq generation if telem is enabled. Still we have to do a copy in this method, not exposing internal xferReq memory to the user, but would be more efficient.

src/core/nixl_agent.cpp
mkhazraee: There is a task to add NIXL_WARN or NIXL_DEBUG to all the error cases in this file. But can add for this anyways.

src/api/cpp/nixl_types.h
mkhazraee: I think the notion of period or duration being here is good, chose period to be shorter and not say dur. chrono_period_us_t would be the best, but a tad long, so just added a comment on the definition as it's only used within that scope. But open to chrono_period_us_t. WDYT @brminich ?

roiedanino: Explicit is better than implicit even if longer, especially when describing time IMO

src/core/nixl_agent.cpp
roiedanino: What about using inheritance?

mkhazraee: I've removed the backend, it wasn't quite a telemetry and anyways we have getXferBackend. Now adding a single descCount value is not a big deal, so no need.

mkhazraee:
Requires perf test before merge, and a separate PR to add Rust bindings for the new API.

src/api/cpp/nixl.h
```diff
@@ -289,6 +289,17 @@ class nixlAgent {
...
+
+        /**
+         * @brief  Get the telemetry data associated with `req_hndl`.
+         *
+         * @param  req_hndl        Transfer request handle obtained from makeXferReq/createXferReq
+         * @param  telemetry [out] Output telemetry information
+         * @return nixl_status_t   Error code if call was not successful
+         */
+        nixl_status_t
+        getXferTelemetry(const nixlXferReqH *req_hndl, nixl_xfer_telem_t &telemetry) const;
```
ovidiusm, line 301: Just curious, what are the main usecases for this? Dynamo? Or other projects? Do they want to get telemetry for all requests?

ovidiusm: Wondering if there is a usecase where we want to enable it only for some requests or toggle it on/off

mkhazraee: vLLM already asked for it and they even have a PR for it in nixl repo. The idea is they want to have their own telemetry log, instead of our log file.

.gitlab/test_cpp.sh
```diff
@@ -78,7 +78,7 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=y ./bin/agent_example &
+NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=/tmp ./bin/agent_example &
```
yosefe: 1. /tmp may not be a good place if have multiple instances of test running on same machine , also not clear where are these files removed?
2. why in python tests NIXL_TELEMETRY_DIR is set globally, and here only for one test?
3. why need to change file-writing related things in this PR, that adds API to query telemetry? I'd assume that if we get the info using the API, no need to write telemetry to a file at all?

src/api/cpp/nixl_types.h
```diff
@@ -223,6 +225,60 @@ struct nixlAgentOptionalArgs {
...
+
+/**
+ * @struct nixlXferTelemetry
+ * @brief A structure for telemetry output from agent API
+ */
+struct nixlXferTelemetry {
+    /**
+     * @var startTime Time that the transfer was posted
+     */
+    chrono_point_t startTime = min_chrono_time;
```
yosefe: why need to define default values to the fields? since nixlXferTelemetry is an output value from the API?

src/api/cpp/nixl_types.h
```diff
@@ -223,6 +225,60 @@ struct nixlAgentOptionalArgs {
...
+
+    /**
+     * @var postDuration Time it took to do the post operation
+     */
+    chrono_period_us_t postDuration = chrono_period_us_t(0);
```
yosefe: maybe define postTime, xferTime, etc?
so all fields are points in time? to make it consistent

src/bindings/python/nixl_bindings.cpp
```diff
@@ -161,6 +162,22 @@ PYBIND11_MODULE(_bindings, m) {
         .value("NIXL_ERR_NOT_SUPPORTED", NIXL_ERR_NOT_SUPPORTED)
         .export_values();
 
+    py::class_<nixl_xfer_telem_t>(m, "nixlXferTelemetry")
```
yosefe, line 189: 1. do we need rust bindings?
2. need python tests in CI

src/core/nixl_agent.cpp
```diff
@@ -31,14 +31,13 @@
...
 
 // Macro to safely call telemetry methods only if telemetry_ is not null
-#define UPDATE_TELEMETRY_DATA(telemetry_ptr, method_call) \
-    do {                                                  \
-        if (telemetry_ptr) {                              \
-            telemetry_ptr->method_call;                   \
-        }                                                 \
+#define ADD_ERR_TELEMETRY(status)                                         \
+    do {                                                                  \
+        if (data->telemetry_) data->telemetry_->updateErrorCount(status); \
```
yosefe: 1. what is "data" here?
2. can it be inline function instead of macro?

src/core/nixl_agent.cpp
```diff
@@ -1040,18 +1052,33 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    if (!data->telemetryEnabled) {
+        NIXL_ERROR << "getXferTelemetry cannot return values when telemetry is not enabled.";
+        return NIXL_ERR_NO_TELEMETRY;
+    }
+
+    if (req_hndl->status < 0) return req_hndl->status;
+
+    // NIXL_SUCCESS or NIXL_IN_PROG, values are initialized
+    telemetry = req_hndl->telemetry;
```
yosefe, line 1206: IMO it's weird getXferTelemetry returns NIXL_IN_PROG
request status is reported by getXferStatus already

yosefe (dismissed):

.gitlab/test_cpp.sh
mkhazraee: 1. There was an issue with some release checks that we assumed /tmp/ at least would have better access. It didn't even have /tmp support, so can become local. Here it was just replacing the default in the original code.
2. Good idea to make python one only for this, it was just an example, but this is better.
3. Exactly for that, previously when enabling telemetry, file was used. Now we're disabling file output unless this env var is set.

src/api/cpp/nixl_types.h
mkhazraee: yes

src/api/cpp/nixl_types.h
mkhazraee: We were thinking of this, but postTime and xferTime is mainly what we're interested, and users have to do the extra compute. We could remove the initial Time, but thought useful to be around.

src/bindings/python/nixl_bindings.cpp
mkhazraee: 1. yes, a separate PR, rust bindings become large and this PR is already long.
2. I added tests to current tests ran in .gitlab/ , do we need something else for that?

src/core/nixl_agent.cpp
mkhazraee: data is agentData object, as we call this from nixlAgent. We can make it a method of data, but mainly it was a single if statement that Rafi didn't want to repeat. That being said, now that I simplified the use case, maybe we can just remove the define, as there is a 1liner if and call.

src/core/nixl_agent.cpp
mkhazraee: I was debating to return error if it's prog, or just return the currently available values like post time, but clarify that this is not NIXL_SUCCESS status.

src/api/cpp/nixl.h
NickLucche: exactly, vllm already has a built-in log system that aggregates at different granularities.
Providing a code-only api would enable proper integration with the rest of the system

src/api/python/_api.py
```diff
@@ -513,6 +513,18 @@ def check_xfer_state(self, handle: nixl_xfer_handle) -> str:
         else:
             return "ERR"
 
+    """
+    @brief Get telemetry information of a transfer request.
+
+    @param handle Handle to the transfer operation, from make_prepped_xfer or initialize_xfer.
+    @return nixlXferTelemetry object
+    """
```
NickLucche, line 628: nit: can we document the struct fields here along with the unit of measure?

.gitlab/test_cpp.sh
yosefe: so why not make it NIXL_TELEMETRY_DIR=$PWD/telemetry?

src/api/cpp/nixl_types.h
yosefe: Don't understand the answer .. can we remove the default values?

src/api/cpp/nixl_types.h
yosefe: so startTime is actually postTime?

test/gtest/test_transfer.cpp
```diff
@@ -403,6 +416,18 @@ class TestTransfer :
                              << "(" << bandwidth << " GB/s)";
                 }
 
+                if (test_telemetry) {
```
yosefe: why not always test telemetry if (expected_telem_status == NIXL_SUCCESS)

test/gtest/test_transfer.cpp
```diff
@@ -528,6 +554,113 @@ TEST_P(TestTransfer, ListenerCommSize) {
     deregisterMem(getAgent(1), buffers, DRAM_SEG);
 }
 
+TEST_P(TestTransfer, GetXferTelemetryFile) {
+    gtest::ScopedEnv env;
+    env.addVar("NIXL_TELEMETRY_ENABLE", "y");
```
yosefe, line 562: how can we enable telemetry, but not write it to a file?

mkhazraee: Enabling telemetry here means recording the times a transfer started and posted and finished. If it's not enabled, those measurement overheads will not be there. Now whether user wants to only use our API, or actually have a telemetry thread to write to a file is what the DIR var determines.

test/gtest/test_transfer.cpp
mkhazraee: Sure, didn't want to modify the previous test, will set them with false as the env var might not be set.

src/api/cpp/nixl_types.h
mkhazraee: If the user calls getTelemetry before posting it, wanted to have some values. Or we can say returning error is sufficient. Also not sure if they call getTelemetry between the post and before the completion, should be error or let this be?

src/api/cpp/nixl_types.h
mkhazraee: startTime is when the call to postXferReq happened, postTime is how much that method took.

src/core/nixl_agent.cpp
mkhazraee: Following the initialization comment, if we remove this, maybe we can avoid doing the initialization altogether. The returned output is garbage, but we said error.

src/api/cpp/nixl_types.h
mkhazraee: I've made that to be error for now, it's a weird use case anyways, later if needed we can put this back.

src/core/nixl_agent.cpp
mkhazraee: Still repeating the 1 liner code is not the best, and can't inline it as it's getting called from the agent method, and if make it inline for agent_data, that doesn't have access to agent's telemetry_enabled_ field. For now keeping this PR to just simplify it, maybe in another PR we can find a better method.

.gitlab/test_cpp.sh
```diff
@@ -78,9 +78,9 @@ cd ${INSTALL_DIR}
 ./bin/nixl_etcd_example
 ./bin/ucx_backend_test
 ./bin/ucx_mo_backend_test
-NIXL_TELEMETRY_ENABLE=y ./bin/agent_example &
+NIXL_TELEMETRY_ENABLE=y NIXL_TELEMETRY_DIR=$PWD/telemetry ./bin/agent_example &
```
mkhazraee: @yosefe  after doing this the CI fails as I suspected:
Failed to open a file for shared memory: /opt/nixl/telemetry/Agent001
Any suggestions or revert back to /tmp for now?

src/core/telemetry.cpp
```diff
@@ -34,13 +34,13 @@ namespace fs = std::filesystem;
 constexpr std::chrono::milliseconds DEFAULT_TELEMETRY_RUN_INTERVAL = 100ms;
 constexpr size_t DEFAULT_TELEMETRY_BUFFER_SIZE = 4096;
 
-nixlTelemetry::nixlTelemetry(const std::string &name, backend_map_t &backend_map)
+nixlTelemetry::nixlTelemetry(const std::string &file_path, backend_map_t &backend_map)
```
ovidiusm, line 37: Maybe out of scope, but it would be good to rename this class to something more obvious like telemetryWriter. Can be done in a separate PR.

.gitlab/test_cpp.sh
ovidiusm: That's right, we should use /tmp in CI containers, the other mount points do not support files opened for shared memory

ovidiusm (approved):

src/api/cpp/nixl.h
brminich, line 301: can we pass `req_hndl` by ref?

brminich: nvm, probably not consistent

mkhazraee: yeah, tried to be consistent.

src/core/telemetry.cpp
mkhazraee: Agreed, let's ask Rafi on the small telemetry fix PR, #750

brminich (approved):

PR #704, opened 2025-08-20
API/GPU/UCX: Create GPU-side Device API

michal-shalev:
## What?
Add GPU device function stubs for UCX backend with template-based coordination levels (thread, warp, block, grid) for memory transfers, signal transfers, and status checking.

## Why?
To establish foundation for GPU-initiated transfers.
See also: https://github.com/ai-dynamo/nixl/pull/705

## How?
Added 5 templated device functions in `nixl_device.cuh` that return `NIXL_ERR_NOT_SUPPORTED` as placeholders:
`nixlGpuPostSingleWriteXferReq` - memory transfers
`nixlGpuPostSignalXferReq` - signal transfers
`nixlGpuPostPartialWriteXferReq` - partial transfers
`nixlGpuPostWriteXferReq` - full transfers
`nixlGpuGetXferStatus` - status checking

### DIFF
--- a/src/api/gpu/ucx/nixl_device.cuh
+++ b/src/api/gpu/ucx/nixl_device.cuh
@@ -1,0 +1,157 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef _NIXL_DEVICE_CUH
+#define _NIXL_DEVICE_CUH
+#include <nixl_types.h>
+
+struct nixlGpuXferStatusH;
+
+struct nixlGpuSignal {
+    uint64_t inc = 0;
+    uint64_t remote_addr = 0;
+};
+
+/**
+ * @enum  nixl_gpu_level_t
+ * @brief An enumeration of different levels for GPU transfer requests.
+ */
+enum class nixl_gpu_level_t : uint64_t {
+    THREAD,
+    WARP,
+    BLOCK,
+    GRID
+};
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl    [in]  Request handle.
+ * @param addr        [in]  Local address of the memory to be transferred.
+ * @param remote_addr [in]  Remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode.
+ * @param xfer_status [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH *req_hndl,
+                              const void *addr,
+                              uint64_t remote_addr,
+                              bool is_no_delay = true,
+                              nixlGpuXferStatusH *xfer_status = nullptr)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Post a signal transfer request to the GPU.
+ *
+ * @param req_hndl           [in]  Request handle.
+ * @param signal             [in]  Signal to be sent.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostSignalXferReq(nixlGpuXferReqH *req_hndl,
+                         const nixlGpuSignal &signal,
+                         bool is_no_delay = true,
+                         nixlGpuXferStatusH *xfer_status = nullptr)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Post a partial memory transfer request to the GPU.
+ *
+ * @param req_hndl           [in]  Request handle.
+ * @param count              [in]  Number of blocks to send. This is also the length of the arrays
+ *                                 @a indices, @a sizes, @a addrs, and @a remote_addrs.
+ * @param indices            [in]  Indices of the blocks to send.
+ * @param sizes              [in]  Sizes of the blocks to send.
+ * @param addrs              [in]  Addresses of the blocks to send.
+ * @param remote_addrs       [in]  Remote addresses of the blocks to send to.
+ * @param signal             [in]  Signal to be sent.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostPartialWriteXferReq(nixlGpuXferReqH *req_hndl,
+                               size_t count,
+                               const int *indices,
+                               const size_t *sizes,
+                               void *const *addrs,
+                               const uint64_t *remote_addrs,
+                               const nixlGpuSignal &signal,
+                               bool is_no_delay = true,
+                               nixlGpuXferStatusH *xfer_status = nullptr)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl           [in]  Request handle.
+ * @param sizes              [in]  Sizes of the blocks to send.
+ * @param addrs              [in]  Addresses of the blocks to send.
+ * @param remote_addrs       [in]  Remote addresses of the blocks to send to.
+ * @param signal             [in]  Signal to be sent.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @note The arrays @a sizes, @a addrs, and @a remote_addrs must have the same length, which
+ *       corresponds to the number of blocks to transfer as specified in @a req_hndl.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostWriteXferReq(nixlGpuXferReqH *req_hndl,
+                        const size_t *sizes,
+                        void *const *addrs,
+                        const uint64_t *remote_addrs,
+                        const nixlGpuSignal &signal,
+                        bool is_no_delay = true,
+                        nixlGpuXferStatusH *xfer_status = nullptr)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Get the status of the transfer request.
+ *
+ * @param xfer_status [in]  Status of the transfer.
+ *
+ * @return NIXL_SUCCESS  The request has completed, no more operations are in progress.
+ * @return NIXL_IN_PROG  One or more operations in the request have not completed.
+ * @return Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuGetXferStatus(const nixlGpuXferStatusH &xfer_status)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+#endif
--- a/src/api/cpp/nixl_types.h
+++ b/src/api/cpp/nixl_types.h
@@ -224,6 +224,11 @@ struct nixlAgentOptionalArgs {
 using nixl_opt_args_t = nixlAgentOptionalArgs;
 
 /**
+ * @brief A typedef for a nixlGpuXferReqH
+ */
+using nixlGpuXferReqH = void;
+
+/**
  * @brief A define for an empty string, that indicates the descriptor list is being
  *        prepared for the local agent as an initiator in prepXferDlist method.
  */


src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
+/*
```
brminich: i think we do not need `ucx` subfolder in the path, i.e.:
`src/api/cuda/ucx/` -> `src/api/cuda`

src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+    NIXL_GPU_COORDINATION_BLOCK,
+    NIXL_GPU_COORDINATION_GRID
+};
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl    [in]  The request handle.
+ * @param address     [in]  The local address of the memory to be transferred.
+ * @param remote_addr [in]  The remote address of the memory to be transferred.
```
brminich:
```suggestion
 * @param remote_addr [in]  The remote address of the memory to be transferred to.
```

src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param xfer_status [out] The status of the transfer. If null the status is not reported.
```
brminich: msybe better make `out` parameter to be the last one?
(relevant for other APIs too)

src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostPartialMemXferReq(nixlGpuXferReqH* req_hndl,
+                             size_t count,
+                             const int* indices,
+                             const size_t* sizes,
+                             const void** addresses,
+                             const uint64_t* remote_addrs,
+                             uint64_t signal_remote_addr,
+                             uint64_t signal_inc,
```
brminich: in the previous API it was `signal_inc` and then `signal_remote_addr`, better to be consistent with the order of args

src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param signal_remote_addr [in]  The remote address of the signal. If address is 0, no signal.
+ * @param signal_inc         [in]  The increment of the signal.
+ * @param xfer_status        [out] The status of the transfer. If null the status is not reported.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlPostGpuXferReq(nixlGpuXferReqH* req_hndl,
```
brminich:
```suggestion
nixlGpuPostXferReq(nixlGpuXferReqH* req_hndl,
```

src/api/cuda/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl           [in]  The request handle.
+ * @param sizes              [in]  The sizes of the blocks to send.
+ * @param addresses          [in]  The addresses of the blocks to send.
+ * @param remote_addrs       [in]  The remote addresses of the blocks to send.
```
brminich:
```suggestion
 * @param remote_addrs       [in]  The remote addresses of the blocks to send to.
```

src/api/cuda/ucx/nixl_device.cuh
ofirfarjun7: After further discussion I think the path should be `src/api/gpu` to fit other GPU vendors.

michal-shalev: We said `src/api/gpu/ucx`, no? @ofirfarjun7

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: Would it be convenient? `is_no_delay` has a default param.

brminich: xfer_status as well, we typically try to keep out params in the very end

src/api/cuda/ucx/nixl_device.cuh
ofirfarjun7: yes

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
brminich: why?

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param remote_addr [in]  The remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode. True by default.
+ * @param xfer_status [out] The status of the transfer. If null the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostMemXferReq(nixlGpuXferReqH* req_hndl,
+                      void *address,
```
rakhmets:
```suggestion
                      const void *address,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostPartialMemXferReq(nixlGpuXferReqH* req_hndl,
+                             size_t count,
+                             const int* indices,
+                             const size_t* sizes,
+                             const void** addresses,
```
rakhmets:
```suggestion
                             void *const *addresses,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param is_no_delay [in]  Whether to use no-delay mode. True by default.
+ * @param xfer_status [out] The status of the transfer. If null the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostMemXferReq(nixlGpuXferReqH* req_hndl,
+                      void *address,
+                      uint64_t remote_addr,
```
rakhmets: `addr` and `remote_addr`, or `address` and `remote_address`.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] The status of the transfer. If null the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostXferReq(nixlGpuXferReqH* req_hndl,
+                   const size_t* sizes,
+                   const void** addresses,
```
rakhmets:
```suggestion
                   void *const *addresses,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param signal_inc         [in]  The increment of the signal.
+ * @param signal_remote_addr [in]  The remote address of the signal to be sent to. If address is 0, no signal.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] The status of the transfer. If null the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
+nixlGpuPostXferReq(nixlGpuXferReqH* req_hndl,
```
rakhmets: If there is a correspondence between the size of `sizes`, `addresses`, `remote_addrs` and the content of `req_hndl`, then this should be mentioned at least in the description of the method.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+ * @param req_hndl    [in]  The request handle.
+ * @param address     [in]  The local address of the memory to be transferred.
+ * @param remote_addr [in]  The remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode. True by default.
+ * @param xfer_status [out] The status of the transfer. If null the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+__device__ static inline nixl_status_t
```
rakhmets: Why do we need `static` keyword?

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: To be more generic, and to be aligned with the APIs that are named "Gpu" and not "Cuda"

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * @brief A struct for a nixlGpuXferStatusH
+ */
+struct nixlGpuXferStatusH;
+
+/**
+ * @enum  nixl_gpu_coordination_level_t
+ * @brief An enumeration of different coordination levels for GPU transfer requests.
+ */
+enum nixl_gpu_coordination_level_t {
```
ColinNV: Can this be written as follows?
```c++
enum class nixl_gpu_coordination_level_t : uint8_t {
   THREAD,
   ...
};
```

src/api/cuda/ucx/nixl_device.cuh
brminich: i mean why do we need `ucx` name in the path?

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,153 @@
...
+    NIXL_GPU_COORDINATION_THREAD,
```
yosefe: suggest to rename "COORDINATION" to "COOPERATION" everywhere, since it's the right notation
https://developer.nvidia.com/blog/cooperative-groups/

src/api/gpu/ucx/nixl_device.cuh
yosefe: @rakhmets why const void*?
this could also be GET operation that modifies local buffer

yosefe: also pls rename to "addr"

src/api/gpu/ucx/nixl_device.cuh
yosefe: `void *const *addrs`,

src/api/gpu/ucx/nixl_device.cuh
yosefe: `void *const *addrs,`

michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: renamed to addr

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
michal-shalev: Since this API is inlined and can't be a full OOP solution, this is basically a specific UCX implementation for the device API.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: I changed the `count` description in `nixlGpuPostPartialMemXferReq` and also added a note in `nixlGpuPostXferReq`

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -24,31 +24,31 @@
...
  * @param xfer_status [out] The status of the transfer. If null the status is not reported.
  *
  * @return nixl_status_t    Error code if call was not successful
  */
-template<nixl_gpu_coordination_level_t level = NIXL_GPU_COORDINATION_BLOCK>
+template<nixl_gpu_cooperation_level_t level = NIXL_GPU_COOPERATION_BLOCK>
 __device__ static inline nixl_status_t
 nixlGpuPostMemXferReq(nixlGpuXferReqH* req_hndl,
-                      void *address,
+                      const void *addr,
```
yosefe: IMO remove const, @rakhmets ?

michal-shalev: removed `const`

src/api/gpu/ucx/nixl_device.cuh
rakhmets: @yosefe but the description says that `addr` is the local address of the memory to be transferred.
If we want to have the same API for `put` and `get`, then maybe change semantics to use `source` and `destination` to describe the buffers. So that the source will be `const void* src`, and the destination - `void *dst`.
Or use the overloading, so that `put` will have `const void *addr`, and `get` - `void *addr`.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+    NIXL_GPU_LEVEL_THREAD,
+    NIXL_GPU_LEVEL_WARP,
+    NIXL_GPU_LEVEL_BLOCK,
+    NIXL_GPU_LEVEL_GRID
+};
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl    [in]  The request handle.
```
rakhmets:
```suggestion
 * @param req_hndl    [in]  Request handle.
```
You can omit `the` at the beginning of the parameter description.
The same is true for other parameters.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Post a partial memory transfer request to the GPU.
+ *
+ * @param req_hndl           [in]  The request handle.
+ * @param count              [in]  The number of blocks to send. This is also the length of the arrays
+ *                                 @ref indices, @ref sizes, @ref addrs, and @ref remote_addrs.
```
rakhmets:
```suggestion
 *                                 @a indices, @a sizes, @a addrs, and @a remote_addrs.
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+ * @param indices            [in]  The indices of the blocks to send.
+ * @param sizes              [in]  The sizes of the blocks to send.
+ * @param addrs              [in]  The addresses of the blocks to send.
+ * @param remote_addrs       [in]  The remote addresses of the blocks to send to.
+ * @param signal_inc         [in]  The increment of the signal.
+ * @param signal_remote_addr [in]  The remote address of the signal to be sent to. If address is 0, no signal.
```
rakhmets: It might be worth making 0 the default value for these parameters in this case. Then the user will not have to think about which value to use for `signal_inc` if `signal_remote_addr` is 0.
Moreover, since these parameters are paired, it is better to define a structure.
```C++
struct nixlGpuPostSignal {
    uint64_t inc = 0;
    uint64_t remote_addr = 0;
};
...
nixlGpuPostPartialMemXferReq(nixlGpuXferReqH* req_hndl,
                             size_t count,
                             const int* indices,
                             const size_t* sizes,
                             void *const *addrs,
                             const uint64_t* remote_addrs,
                             nixlGpuPostSignal signal = {},
                             bool is_no_delay = true,
                             nixlGpuXferStatusH* xfer_status = nullptr);
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+ * @param req_hndl           [in]  The request handle.
+ * @param sizes              [in]  The sizes of the blocks to send.
+ * @param addrs              [in]  The addresses of the blocks to send.
+ * @param remote_addrs       [in]  The remote addresses of the blocks to send to.
+ * @param signal_inc         [in]  The increment of the signal.
+ * @param signal_remote_addr [in]  The remote address of the signal to be sent to. If address is 0, no signal.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] The status of the transfer. If null the status is not reported.
+ *
+ * @note The arrays @ref sizes, @ref addrs, and @ref remote_addrs must have the same length, which
```
rakhmets:
```suggestion
 * @note The arrays @a sizes, @a addrs, and @a remote_addrs must have the same length, which
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+ *       corresponds to the number of blocks to transfer as specified in @ref req_hndl.
```
rakhmets:
```suggestion
 *       corresponds to the number of blocks to transfer as specified in @a req_hndl.
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+                   uint64_t signal_inc,
+                   uint64_t signal_remote_addr,
+                   bool is_no_delay = true,
+                   nixlGpuXferStatusH* xfer_status = nullptr)
+{
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+/**
+ * @brief Get the status of a transfer request.
```
rakhmets:
```suggestion
 * @brief Get the status of the transfer request.
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * @param xfer_status [in]  The status of the transfer.
+ *
+ * @return nixl_status_t    Error code if call was not successful
```
rakhmets: It would be better to add smth like:
```C++
 * @return NIXL_SUCCESS  The request has completed, no more operations are in progress.
 * @return NIXL_IN_PROG  One or more operations in the request have not completed.
```

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: Since these are inline template functions defined in a header file, without static, each translation unit that includes this header would create its own definition of these functions, leading to "multiple definition" errors during linking. The static keyword ensures each function has internal linkage, making it local to each compilation unit.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+    NIXL_GPU_LEVEL_GRID
+};
+
+/**
+ * @brief Post a memory transfer request to the GPU.
+ *
+ * @param req_hndl    [in]  The request handle.
+ * @param addr        [in]  The local address of the memory to be transferred.
+ * @param remote_addr [in]  The remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode. True by default.
```
rakhmets: Please remove `True by default` as it repeats what is declared in the function signature.
The same for `nixlGpuPostSignalXferReq`.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,161 @@
...
+ */
+#ifndef _NIXL_DEVICE_CUH
+#define _NIXL_DEVICE_CUH
+#include <nixl_types.h>
+
+#define NIXL_DEVICE_INLINE __device__ static inline
+
+/**
+ * @brief A struct for a nixlGpuXferStatusH
+ */
```
rakhmets: Please delete this comment, as it doesn't make any additional sense.

src/api/gpu/ucx/nixl_device.cuh
ColinNV: Since this is a function template the `inline` is implied and can be removed.

And since this is C++ the `static` can also be removed because for `inline` functions the linker will silently get rid of multiple copies from different translation units (in the case that not all calls are actually inlined).

rakhmets: It seems that `inline` is enough in your case, because with `inline` each translation unit uses its own copy of the function's definition.

ColinNV: But the `inline` is not needed either because it is implicit for function templates.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: It depends on the implementation, `NIXL_IN_PROG` will not necessarily return from this function, IMO it could be modified during implementation.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
rakhmets: Then I don't quite understand the purpose of this method.
`nixlGpuPost*` returns `nixl_status_t`, and `nixlGpuXferStatusH*`. I thought the second return value was to track the progress. Why do we need two statuses?

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: I'll remove both the `inline` and `static` then

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done, used uint64_t since the flags being passed to UCX is uint64_t

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: fixed, done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/cuda/ucx/nixl_device.cuh
brminich: ok

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @enum  nixl_gpu_level_t
+ * @brief An enumeration of different levels for GPU transfer requests.
+ */
+enum class nixl_gpu_level_t : uint64_t {
+    NIXL_GPU_LEVEL_THREAD,
+    NIXL_GPU_LEVEL_WARP,
+    NIXL_GPU_LEVEL_BLOCK,
+    NIXL_GPU_LEVEL_GRID
```
brminich: no need in `NIXL_GPU_LEVEL_` prefix with scoped enum class, because you are supposed to access it like `nixl_gpu_level_t::THREAD`

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @param req_hndl    [in]  Request handle.
+ * @param addr        [in]  Local address of the memory to be transferred.
+ * @param remote_addr [in]  Remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode.
+ * @param xfer_status [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = NIXL_GPU_LEVEL_BLOCK>
```
brminich: i guess it is missing `nixl_gpu_level_t::`

src/api/gpu/ucx/nixl_device.cuh
yosefe: in nixl api the terminology is always "local" and "remote" w.r.t memory (for example local descs / remote descs) so IMO better keep same naming here.
We can change it to "nixlGpuPostWriteXferReq" and later add "nixlGpuPostReadXferReq
@michal-shalev WDYT?

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: Maybe `nixlGpuWriteXferReq` and later add `nixlGpuReadXferReq`?
@yosefe @rakhmets

michal-shalev: or `nixlGpuPutXferReq` and `nixlGpuGetXferReq`

rakhmets: We have `nixlAgent::postXferReq`. So I suggest to keep `post` for GPU-side API:
`nixlGpuPostSingleWriteXferReq`
`nixlGpuPostSignalXferReq`
`nixlGpuPostScatteredWriteXferReq`, or `nixlGpuPostPartialWriteXferReq`
`nixlGpuPostWriteXferReq` (IIUC this is the main API, so we use this name)

rakhmets: I prefer Write/Read because it is already used in NIXL, e.g. `enum nixl_xfer_op_t {NIXL_READ, NIXL_WRITE};`.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+ * @param addr        [in]  Local address of the memory to be transferred.
+ * @param remote_addr [in]  Remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode.
+ * @param xfer_status [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostMemXferReq(nixlGpuXferReqH* req_hndl,
```
yosefe:
```suggestion
nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH* req_hndl,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+ * @param remote_addrs       [in]  Remote addresses of the blocks to send to.
+ * @param signal             [in]  Signal to be sent.
+ * @param is_no_delay        [in]  Whether to use no-delay mode.
+ * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostPartialMemXferReq(nixlGpuXferReqH* req_hndl,
```
yosefe:
```suggestion
nixlGpuPostPartialWriteXferReq(nixlGpuXferReqH* req_hndl,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+ * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @note The arrays @a sizes, @a addrs, and @a remote_addrs must have the same length, which
+ *       corresponds to the number of blocks to transfer as specified in @a req_hndl.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostXferReq(nixlGpuXferReqH* req_hndl,
```
yosefe:
```suggestion
nixlGpuPostWriteXferReq(nixlGpuXferReqH* req_hndl,
```

yosefe (dismissed):

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -0,0 +1,157 @@
...
+ * @param addr        [in]  Local address of the memory to be transferred.
+ * @param remote_addr [in]  Remote address of the memory to be transferred to.
+ * @param is_no_delay [in]  Whether to use no-delay mode.
+ * @param xfer_status [out] Status of the transfer. If null, the status is not reported.
+ *
+ * @return nixl_status_t    Error code if call was not successful
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ nixl_status_t
+nixlGpuPostWriteXferReq(nixlGpuXferReqH *req_hndl,
```
yosefe:
```suggestion
nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH *req_hndl,
```

michal-shalev: done

PR #705, opened 2025-08-20
Create host-side GPU transfer request API

michal-shalev:
## What?
Adds host-side API for GPU transfer request.

## Why?
To establish foundation for GPU-initiated transfers.
See also: https://github.com/ai-dynamo/nixl/pull/704

## How?
 - Added two new virtual methods to nixlBackendEngine interface:
     - `createGpuXferReq()`: Creates a GPU transfer request handle to GPU memory
     - `releaseGpuXferReq()`: Releases a GPU transfer request from GPU memory
 - Implemented public API methods in `nixlAgent` with parameter validation and thread safety
 - Added UCX backend stubs for compatibility
 - Included API documentation

### DIFF
--- a/src/api/cpp/backend/backend_engine.h
+++ b/src/api/cpp/backend/backend_engine.h
@@ -156,6 +156,15 @@ class nixlBackendEngine {
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
 
+        // Create a GPU transfer request to GPU memory for GPU transfer.
+        virtual nixl_status_t
+        createGpuXferReq(const nixlBackendReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
+
+        // Release a GPU transfer request from GPU memory
+        virtual void
+        releaseGpuXferReq(nixlGpuXferReqH *gpu_req_hndl) const {}
 
         // *** Needs to be implemented if supportsRemote() is true *** //
 
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -312,6 +312,24 @@ class nixlAgent {
         releaseXferReq (nixlXferReqH* req_hndl) const;
 
         /**
+         * @brief  Create a GPU transfer request from a transfer request.
+         *
+         * @param  req_hndl     [in]  Transfer request obtained from makeXferReq/createXferReq
+         * @param  gpu_req_hndl [out] GPU transfer request handle
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const;
+
+        /**
+         * @brief  Release transfer request from GPU memory
+         *
+         * @param  gpu_req_hndl  [in] GPU transfer request handle to be released
+         */
+        void
+        releaseGpuXferReq(nixlGpuXferReqH *gpu_req_hndl) const;
+
+        /**
          * @brief  Release the prepared descriptor list handle `dlist_hndl`
          *
          * @param  dlist_hndl    Prepared descriptor list handle to be released
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1088,6 +1088,41 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
 }
 
 nixl_status_t
+nixlAgent::createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
+    if (!req_hndl.engine) {
+        NIXL_ERROR << "Invalid request handle[" << &req_hndl << "]: engine is null";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    if (!req_hndl.backendHandle) {
+        NIXL_ERROR << "Invalid request handle[" << &req_hndl << "]: backendHandle is null";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+    const auto status = req_hndl.engine->createGpuXferReq(*req_hndl.backendHandle, gpu_req_hndl);
+    if (status == NIXL_SUCCESS) {
+        data->gpuReqToEngine.emplace(&gpu_req_hndl, req_hndl.engine);
+    }
+
+    return status;
+}
+
+void
+nixlAgent::releaseGpuXferReq(nixlGpuXferReqH *gpu_req_hndl) const {
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+    auto it = data->gpuReqToEngine.find(gpu_req_hndl);
+    if (it == data->gpuReqToEngine.end()) {
+        NIXL_WARN << "Invalid gpu_req_hndl[" << gpu_req_hndl << "] ";
+        return;
+    }
+
+    it->second->releaseGpuXferReq(gpu_req_hndl);
+
+    data->gpuReqToEngine.erase(it);
+}
+
+nixl_status_t
 nixlAgent::releasedDlistH (nixlDlistH* dlist_hndl) const {
     NIXL_LOCK_GUARD(data->lock);
     delete dlist_hndl;
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -190,6 +190,12 @@ public:
     nixl_status_t
     releaseReqH(nixlBackendReqH *handle) const override;
 
+    nixl_status_t
+    createGpuXferReq(const nixlBackendReqH &handle, nixlGpuXferReqH *&gpu_req_hndl) const override;
+
+    void
+    releaseGpuXferReq(nixlGpuXferReqH *gpu_req_hndl) const override;
+
     int
     progress();
 
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -78,6 +78,9 @@ class nixlAgentData {
         std::unordered_map<nixl_backend_t, nixlBackendH*> backendHandles;
         std::unordered_map<nixl_backend_t, nixl_blob_t>   connMD;
 
+        // Bookkeeping from GPU request handles to backend engines
+        std::unordered_map<nixlGpuXferReqH *, nixlBackendEngine *> gpuReqToEngine;
+
         // Local section, and Remote sections and their available common backends
         nixlLocalSection*                                        memorySection;
 
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1622,6 +1622,15 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
     return status;
 }
 
+nixl_status_t
+nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &handle,
+                                nixlGpuXferReqH *&gpu_req_hndl) const {
+    return NIXL_ERR_NOT_SUPPORTED;
+}
+
+void
+nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH *gpu_req_hndl) const {}
+
 int nixlUcxEngine::progress() {
     // TODO: add listen for connection handling if necessary
     int ret = 0;


src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,17 @@ class nixlBackendEngine {
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
 
+        // Export transfer request to GPU memory for GPU transfer.
+        virtual nixl_status_t
+        exportXferReqtoGPU(nixlBackendReqH *req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
+            return NIXL_ERR_NOT_SUPPORTED; // Default implementation for file backends
```
brminich: why file?

src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,17 @@ class nixlBackendEngine {
...
+        }
+
+        // Release transfer request from GPU memory
+        virtual nixl_status_t
+        releaseXferReqtoGPU(nixlBackendReqH *req_hndl) const {
+            return NIXL_ERR_NOT_SUPPORTED; // Default implementation for file backends
```
brminich: why file?

src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,17 @@ class nixlBackendEngine {
...
 
+        // Export transfer request to GPU memory for GPU transfer.
+        virtual nixl_status_t
+        exportXferReqtoGPU(nixlBackendReqH *req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
+            return NIXL_ERR_NOT_SUPPORTED; // Default implementation for file backends
+        }
+
+        // Release transfer request from GPU memory
+        virtual nixl_status_t
+        releaseXferReqtoGPU(nixlBackendReqH *req_hndl) const {
```
brminich: should not it be `nixlGpuXferReqH *gpu_req_hndl` instead?

src/api/cpp/nixl.h
```diff
@@ -311,6 +311,26 @@ class nixlAgent {
...
+
+        /**
+         * @brief  Release transfer request from GPU memory
+         *
+         * @param  req_hndl      [in] Transfer request handle to release from GPU
+         *
+         * @return nixl_status_t      Error code if call was not successful
+         */
+        nixl_status_t
+        releaseXferReqtoGPU(nixlXferReqH *req_hndl) const;
```
brminich: `gpu_req_hndl`? 
if it is `req_hndl` shall we also invoke `releaseXferReq` for it?
Pls add to the description of this API

src/api/cpp/nixl.h
```diff
@@ -311,6 +311,26 @@ class nixlAgent {
         nixl_status_t
         releaseXferReq (nixlXferReqH* req_hndl) const;
 
+        /**
+         * @brief  Export transfer request to GPU memory for GPU transfer.
+         *
+         * @param  req_hndl     [in]  Transfer request obtained from makeXferReq/createXferReq
+         * @param  gpu_req_hndl [out] GPU transfer request handle output
```
brminich: i'd remove `output`

src/core/nixl_agent.cpp
```diff
@@ -1087,6 +1087,26 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::exportXferReqtoGPU(nixlXferReqH *req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
```
rakhmets: I suggest passing a constant reference instead of a pointer so that we can get rid of checking `req_hndl` that it is not nullptr.

src/core/nixl_agent.cpp
```diff
@@ -1087,6 +1087,26 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
...
+    if (!req_hndl || !req_hndl->engine) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
```
rakhmets, line 1232: It doesn't look like it's being used for its intended purpose. There is no access to `data` here.

src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,17 @@ class nixlBackendEngine {
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
 
+        // Export transfer request to GPU memory for GPU transfer.
+        virtual nixl_status_t
+        exportXferReqtoGPU(nixlBackendReqH *req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
```
rakhmets: From what I see I think it would be better to have only one method which returns smart pointer. E.g.:
```C++
virtual std::unique_ptr<nixlGpuXferReqH>
exportXferReqtoGPU(nixlBackendReqH *req_hndl) const;
```

P.S.
Moreover, I think we have a design issue here. Ideally, it would be better to have a separate interface so as not to overload the main one. E.g.,
```C++
namespace nixl {
class iGpuXferRequestFactory {
public:
    virtual std::unique_ptr<gpuXferReq>
    create(const nixlBackendReqH &req_hndl) const = 0;
};
}
```
This is not something that should be addressed in this request, but a tip on how to improve the API in the future.

src/api/cpp/backend/backend_engine.h
michal-shalev: removed

src/api/cpp/backend/backend_engine.h
michal-shalev: removed

src/api/cpp/backend/backend_engine.h
michal-shalev: According to the current implementation, we pass the request and the exported batch to UCX, and the exported batch is saved internally in NIXL during export, at EP level.
Now when I think about, it's better to pass `releaseXferReqtoGPU` both the request and the exported gpu handle, the user will have to save it.

src/api/cpp/nixl.h
michal-shalev: Please see https://github.com/ai-dynamo/nixl/pull/705#discussion_r2298640847
Regarding invoking `releaseXferReq` AFAIU the lifecycle of the GPU-specific request is different from the lifecycle of the overall transfer handle, which is why in the current implementation I kept `releaseXferReqtoGPU` and `releaseReqH` as two distinct calls.

src/api/cpp/nixl.h
michal-shalev: done

src/api/cpp/backend/backend_engine.h
michal-shalev: done

src/core/nixl_agent.cpp
michal-shalev: `data` isn't accessed after the lock is acquired in this function, but AFAIU the lock is intended to protect against race conditions where another thread could modify the agent's state, such as invalidating the remote connection or destroying the backend engine that `req_hndl->engine` refers to. 
I checked the same pattern in `releaseXferReq`. It acquires a shared lock on data and then all operations are performed on `req_hndl->engine`.

src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,17 @@ class nixlBackendEngine {
...
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
+
+        // Release transfer request from GPU memory
+        virtual nixl_status_t
+        releaseXferReqtoGPU(nixlBackendReqH *req_hndl, nixlGpuXferReqH *gpu_req_hndl) const {
```
brminich: do we really need `nixlBackendReqH *req_hndl` in this func?

src/api/cpp/backend/backend_engine.h
yosefe: I think that in current state of the API better something that is aligned with the current functions:
createXferReq
releaseXferReq

->
exportXferReqtoGPU
releaseXferReqtoGPU

src/api/cpp/backend/backend_engine.h
yosefe: yeah hopefully it would be enough to get releaseXferReqtoGPU*

src/api/cpp/backend/backend_engine.h
michal-shalev: `exportXferReqtoGPU` and `releaseXferReqtoGPU` are the current functions names

src/api/cpp/backend/backend_engine.h
michal-shalev: Yes, it's how the current implementation works.

src/api/cpp/nixl.h
```diff
@@ -311,6 +311,27 @@ class nixlAgent {
...
+        /**
+         * @brief  Release transfer request from GPU memory
+         *
+         * @param  req_hndl      [in] Transfer request handle to release from GPU
+         * @param  gpu_req_hndl  [in] Exported GPU transfer request handle to be released
+         *
+         * @return nixl_status_t      Error code if call was not successful
+         */
+        nixl_status_t
+        releaseXferReqtoGPU(nixlXferReqH *req_hndl, nixlGpuXferReqH *gpu_req_hndl) const;
```
rakhmets: The most transparent and obvious API (excluding smart pointers) would be as follows:
```C++
nixl_status_t
createXferReqtoGPU(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const;
```
The method returns a pointer to a valid object. And the parameter used to create the object is passed as a constant reference.

```C++
void
releaseXferReqtoGPU(nixlGpuXferReqH *gpu_req_hndl) const;
```
Since we have passed the pointer into the possession of the client code. It should be enough to pass the pointer to the method that destroys the object. Otherwise, we implicitly place a responsibility on the client code to maintain mapping between `nixlXferReqH` and `nixlGpuXferReqH` objects used during creation. If you need mapping between them, it should be done in the implementation of the methods.

src/core/nixl_agent.cpp
rakhmets: I see, although it looks unobvious. This, of course, is not the issue of this PR.

src/api/cpp/backend/backend_engine.h
rakhmets: It's better to do something like the one described here https://github.com/ai-dynamo/nixl/pull/705/files#r2301557404:
```C++
virtual nixl_status_t
createXferReqtoGPU(const nixlBackendReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const;
virtual void
releaseXferReqtoGPU(nixlGpuXferReqH *gpu_req_hndl) const;
```
Please note that the release function returns `void`.

src/api/cpp/nixl.h
yosefe: agree with @rakhmets here

michal-shalev: done

src/api/cpp/backend/backend_engine.h
michal-shalev: done

brminich (dismissed):

src/core/nixl_agent.cpp
```diff
@@ -1087,6 +1087,28 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
...
+        NIXL_ERROR << "Invalid request handle: engine is null";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+    return req_hndl.engine->createGpuXferReq(req_hndl.backendHandle, gpu_req_hndl);
+}
+
+void
+nixlAgent::releaseGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *gpu_req_hndl) const {
```
yosefe: didn't we want to remove req_hndl from function parameters?
though, i would requre smth like std::unordered_map from gpu_req_hndl to engine

src/api/cpp/nixl.h
```diff
@@ -311,6 +311,24 @@ class nixlAgent {
...
+         * @param  gpu_req_hndl [out] GPU transfer request handle
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const;
+
+        /**
+         * @brief  Release transfer request from GPU memory
+         *
+         * @param  gpu_req_hndl  [in] Exported GPU transfer request handle to be released
```
rakhmets:
```suggestion
         * @param  gpu_req_hndl  [in] GPU transfer request handle to be released
```

src/core/nixl_agent.cpp
```diff
@@ -1090,23 +1090,34 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
...
     }
 
-    NIXL_SHARED_LOCK_GUARD(data->lock);
-    req_hndl.engine->releaseGpuXferReq(gpu_req_hndl);
+    if (!it.second) {
+        NIXL_WARN << "Invalid gpu_req_hndl[" << gpu_req_hndl << "]: engine is null";
+        return;
+    }
+
+    it.second->releaseGpuXferReq(gpu_req_hndl);
```
yosefe: need to remove gpu_req_hndl from the map

src/api/cpp/backend/backend_engine.h
```diff
@@ -159,6 +159,16 @@ class nixlBackendEngine {
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
 
+        // Export transfer request to GPU memory for GPU transfer.
```
rakhmets: `Create a GPU transfer request`?

src/core/nixl_agent.cpp
michal-shalev: done

src/api/cpp/nixl.h
michal-shalev: done

src/api/cpp/backend/backend_engine.h
michal-shalev: done

src/core/nixl_agent.cpp
```diff
@@ -1087,6 +1087,41 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
...
+    }
+
+    if (!it->second) {
+        NIXL_WARN << "Invalid gpu_req_hndl[" << gpu_req_hndl << "]: engine is null";
+        data->gpuReqToEngine.erase(it);
+        return;
+    }
+
+    it->second->releaseGpuXferReq(gpu_req_hndl);
+    data->gpuReqToEngine.erase(it);
```
```suggestion
    if (it->second) {
        it->second->releaseGpuXferReq(gpu_req_hndl);
    } else {
        NIXL_WARN << "Invalid gpu_req_hndl[" << gpu_req_hndl << "]: engine is null";
    }

    data->gpuReqToEngine.erase(it);
```

src/core/nixl_agent.cpp
```diff
@@ -1087,6 +1087,40 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlAgent::createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_req_hndl) const {
+    if (!req_hndl.engine) {
+        NIXL_ERROR << "Invalid request handle[" << &req_hndl << "]: engine is null";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
```
rakhmets, line 1226:
```suggestion

    if (!req_hndl.backendHandle) {
        NIXL_ERROR << "Invalid request handle[" << &req_hndl << "]: backendHandle is null";
        return NIXL_ERR_INVALID_PARAM;
    }

```

rakhmets (dismissed):

src/api/cpp/nixl.h
```diff
@@ -318,13 +318,13 @@ class nixlAgent {
          * @param  gpu_req_hndl [out] GPU transfer request handle
          * @return nixl_status_t Error code if call was not successful
          */
-        nixl_status_t
+        [[nodiscard]] nixl_status_t
```
yosefe: why need [[nodiscard]] ? it doesn't exist in other API functions

src/core/nixl_agent.cpp
```diff
@@ -1106,18 +1111,19 @@ nixlAgent::createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH *&gpu_
...
     NIXL_SHARED_LOCK_GUARD(data->lock);
-    const auto it = data->gpuReqToEngine.find(gpu_req_hndl);
+    auto it = data->gpuReqToEngine.find(gpu_req_hndl);
     if (it == data->gpuReqToEngine.end()) {
         NIXL_WARN << "Invalid gpu_req_hndl[" << gpu_req_hndl << "] ";
         return;
     }
 
-    if (!it.second) {
+    if (it->second) {
```
yosefe: why would it->second be null, assuming it was actually found in the map?

src/plugins/ucx/ucx_backend.h
```diff
@@ -195,7 +195,7 @@ class nixlUcxEngine : public nixlBackendEngine {
     nixl_status_t
     releaseReqH(nixlBackendReqH *handle) const override;
 
-    nixl_status_t
+    [[nodiscard]] nixl_status_t
```
yosefe: can we remove nodiscard?

src/core/nixl_agent.cpp
rakhmets: `it->second` has type `nixlBackendEngine *`. We need to check the pointer before dereferencing it.

yosefe: but we checked it in line 1907 before adding to the map

src/api/cpp/nixl.h
ColinNV: Thaty's my fault, I suggested adding it to new functions to establish best practice going forward.

yosefe: IMO for now better keep the existing practice, because such inconsistency doesn't look good in the API

rakhmets: The absence of `[[nodiscard]]` means that API users can ignore the returned status. But that's not what we mean by using the function execution status return approach. Do not check the return status like not handling exceptions.

Not having this specifier for existing methods is also not a good practice.

rakhmets: This way we protect the user from unintentional ignoring errors.

src/core/nixl_agent.cpp
rakhmets: The map doesn't own the object. And we are not protected from deleting the object from another location.

yosefe: even if object is deleted from another location, it would not become nullptr in the map

src/api/cpp/nixl.h
yosefe: perhaps, but this PR is not aimed to fix this practice, and since this is API it better be aligned.

src/core/nixl_agent.cpp
rakhmets: Yes, right. This should be removed.

src/api/cpp/nixl.h
michal-shalev: Removed nodiscard

src/core/nixl_agent.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.h
michal-shalev: done

PR #708, opened 2025-08-21
[nixlbench]pass num_workers to ucx

Pyngon:
## What?
create ucp_workers based on the number of thread

## Why?
Reduce contention when using ucp_worker


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1122,7 +1122,7 @@ nixlUcxEngine::create(const nixlBackendInitParams &init_params) {
 
 nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     : nixlBackendEngine(&init_params),
-      sharedWorkerIndex_(0) {
+      sharedWorkerIndex_(1) {
```
iyastreb, line 1116: This variable is an index of the first shared worker, not the count of workers
No need to set it to 1

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -131,6 +131,8 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
             exit(EXIT_FAILURE);
         }
 
+        backend_params["num_workers"] = std::to_string(xferBenchConfig::num_threads + 1);
```
iyastreb, line 134: This is redundant, because in nixlUcxEngine we already handle that:
```
    if (num_workers <= num_threads) {
        /* There must be at least one shared worker */
        num_workers = num_threads + 1;
    }

iyastreb (changes requested):

src/plugins/ucx/ucx_backend.cpp
Pyngon: The first worker will be used to handle notification, so start from 1 to assign first worker last

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: This is not redundant. The main reason for this patch is because people seeing low performance result when running nixlbench with multiple thread. This will make sure the nixlbench get a reasonable result regardless if the user specify "progress thread" or not.

iyastreb: Hmm, but this case must be already handled by the code in nixlUcxEngine (that I shared), that does exactly the same. If it's not working for some reason, can we fix it there?

src/plugins/ucx/ucx_backend.cpp
iyastreb: Ok, could you please clarify your use case then?
On the sender side for sending notifications we use the same worker used to post the request. On the receiver side worker0 handles all the incoming notifications, this is correct. So I'm assuming that your concern is when you use the same agents both for send and receive? Otherwise I don't see an issue

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: Let me provide more details. The problem is user specify "num_threads=8" in nixlbench and do not specify "progress_thread". Then in the UCX code, num_threads=0 and  num_workers=1, so 8 threads shared 1 ucp_workers and the performance is bad.

src/plugins/ucx/ucx_backend.cpp
Pyngon: Yes, i think the agent could be both send and receive, right?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
yosefe: Can we somehow set num_workers for UCX backend from nixlBench command line? Or env var?

Pyngon: > Can we somehow set num_workers for UCX backend from nixlBench command line? Or env var?

We can. But the point is to get the good result by default. When user does not set the progress_thread and num_workers, what is the num_workers we should use? Based on the current test, having multiple threads using 1 worker will cause a lot of lock contention and affect performance. So, i think it make senses to match the number of worker with the number of threads. (of course, we could provide num_workers option for user to experiment with different kind of config, 16 threads+4 workers, or 4 threads+16 workers, etc...)

Having said that, I feel some kind resistance and I am probably missing something. What is your concern to set the "num_workers" here?

iyastreb: Ah, ok, now I see, I mixed up progress_threads with num_threads.
Then your change makes sense!

> Can we somehow set num_workers for UCX backend from nixlBench command line? Or env var?

I don't think we can do this now
Maybe we could have a generic config for backend params map, so that we don't have to introduce a dedicated variable for each and every backend param? Something like:
`--backend_params=num_workers:5,split_batch_size:1024`

iyastreb (approved):

src/plugins/ucx/ucx_backend.cpp
iyastreb: ok

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: @Pyngon @yosefe Do we need to set for target or only initiator? Target side is still single-threaded when polling it is not doing active transfers anyway.

Pyngon: > Ah, ok, now I see, I mixed up progress_threads with num_threads. Then your change makes sense!

ok, got it. The "num_threads" in UCX backend does make some confusion, probably we can rename it to "progress_thread" to match the param name.

> @Pyngon @yosefe Do we need to set for target or only initiator? Target side is still single-threaded when polling it is not doing active transfers anyway.

Target side is only polling for the number of notification, does not look like it will affect the performance now. So, I think single threaded is good enough.

aranadive (approved):

PR #711, opened 2025-08-21
Dump backtrace on crash

ovidiusm:
Extra logs

### DIFF
--- a/src/core/signalhandler.cpp
+++ b/src/core/signalhandler.cpp
@@ -1,0 +1,88 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


src/core/signalhandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+    int nptrs = backtrace(buffer, MAX_BACKTRACE_DEPTH);
+    backtrace_symbols_fd(buffer, nptrs, STDERR_FILENO);
+}
+
+void
+gdb_signal_handler(int sig) {
+    signal(sig, SIG_DFL);
+
+    const char *header = "\n!!! Caught signal. Generating backtrace: !!!\n";
+    ssize_t ignored __attribute__((unused)) = write(STDERR_FILENO, header, strlen(header));
```
ColinNV, line 40: The common way to ignore a function return value is
```c++
(void)write(...);
```
or the modern C++ way is
```c++
#include <utility>
std::ignore = write(...);
```

ovidiusm: Thanks! Actually I did not send this for review since I was just using the PR to diagnose a crash in a test seen in CI.

Do you think it would have value, maybe hidden behind an option and enabled for test runs?

ColinNV: I find it useful to print a backtrace in the terminate handler and for some signals, though it's quite ... courageous to use `fork` and `exec` to invoke `gdb` for it.

PR #715, opened 2025-08-21
Cherry pick to fix telemetry test issue 

mkhazraee:
## What?
Moved files generated for test to /tmp, and improved env var parsing.
(Previously if "n" was passed there would be still warning, or if user passed a bad string which started with y, it would still enable. Improved the checks and add all 4 mode of y/yes/on/1 vs n/no/off/0).

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -132,12 +132,13 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
 
     if (telemetry_env_val != nullptr) {
         if (!strcasecmp(telemetry_env_val, "y") || !strcasecmp(telemetry_env_val, "1") ||
-            !strcasecmp(telemetry_env_val, "on"))
+            !strcasecmp(telemetry_env_val, "yes") || !strcasecmp(telemetry_env_val, "on")) {
             telemetry_ = std::make_unique<nixlTelemetry>(name, backendEngines);
-        else if (strcasecmp(telemetry_env_val, "n") && strcasecmp(telemetry_env_val, "0") &&
-                 strcasecmp(telemetry_env_val, "off"))
+        } else if (strcasecmp(telemetry_env_val, "n") && strcasecmp(telemetry_env_val, "0") &&
+                   strcasecmp(telemetry_env_val, "no") && strcasecmp(telemetry_env_val, "off")) {
             NIXL_WARN
                 << "Invalid NIXL_TELEMETRY_ENABLE environment variable, not enabling telemetry.";
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


test/gtest/telemetry_test.cpp
```diff
@@ -54,13 +54,13 @@ class telemetryTest : public ::testing::Test {
 protected:
     void
     SetUp() override {
-        testDir_ = "./telemetry_test_files";
+        testDir_ = "/tmp/telemetry_test_files";
         testFile_ = "test_telemetry";
         if (!fs::exists(testDir_)) {
             fs::create_directory(testDir_);
```
aranadive: Should there be a check that directory was created? Or a try catch block ..

src/core/nixl_agent.cpp
```diff
@@ -128,14 +128,14 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
...
-                              (telemetry_env_val[0] == 'y' || telemetry_env_val[0] == 'Y' ||
-                               telemetry_env_val[0] == '1'));
-    if (telemetry_enabled) {
-        telemetry_ = std::make_unique<nixlTelemetry>(name, backendEngines);
-    } else if (telemetry_env_val != nullptr) {
-        NIXL_WARN << "Invalid NIXL_TELEMETRY_ENABLE environment variable, not enabling telemetry.";
+    const char *telemetry_env_val = std::getenv(TELEMETRY_ENABLED_VAR);
+
+    if (telemetry_env_val != nullptr) {
+        if (!strcasecmp(telemetry_env_val, "y"))
```
ovidiusm: Would it be possible to accept "y", "1" or "on" as in UCX?

ovidiusm: https://github.com/openucx/ucx/blob/6df8f3c515065a0e4033c026baaae7f553be895d/src/ucs/config/parser.c#L246

test/gtest/telemetry_test.cpp
```diff
@@ -54,13 +54,18 @@ class telemetryTest : public ::testing::Test {
...
         testFile_ = "test_telemetry";
-        if (!fs::exists(testDir_)) {
-            fs::create_directory(testDir_);
+        try {
+            if (!fs::exists(testDir_)) {
+                fs::create_directory(testDir_);
+            }
+        }
+        catch (const fs::filesystem_error &e) {
+            throw std::runtime_error("Could not create the directory for telemetry test.")
```
ovidiusm:
```suggestion
            throw std::runtime_error("Could not create the directory for telemetry test.");
```

ovidiusm (approved):

aranadive (approved):

PR #719, opened 2025-08-24
BUILD: Add configurable CUDA architecture targeting with meson built-in options

michal-shalev:
## What?
Add support for configurable CUDA architecture targeting using Meson's built-in `cuda_args` and `cuda_link_args` options.

## Why?
Replaces hardcoded CUDA architecture flags with configurable defaults using meson's built-in CUDA argument handling for better flexibility.

## How?
- Leverage Meson's built-in `cuda_args`/`cuda_link_args` compiler options
- Apply sensible defaults (compute_80, compute_90) only when user hasn't specified custom values
- Updated README with comprehensive CUDA architecture configuration section
- Simplified code by removing custom flag handling logic

## Usage
```bash
# Uses defaults (Ampere & Hopper: compute_80, compute_90)
meson setup build

# Target specific architecture
meson setup build \
    -Dcuda_args="-gencode=arch=compute_75,code=sm_75" \
    -Dcuda_link_args="-gencode=arch=compute_75,code=sm_75"

# Target multiple architectures
meson setup build \
    -Dcuda_args="-gencode=arch=compute_75,code=sm_75 -gencode=arch=compute_80,code=sm_80" \
    -Dcuda_link_args="-gencode=arch=compute_75,code=sm_75 -gencode=arch=compute_80,code=sm_80"
```

See `README.md` for complete documentation and additional examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Common CUDA compute capabilities:
 - `compute_80,code=sm_80`: NVIDIA Ampere (A100, RTX 30xx)
 - `compute_86,code=sm_86`: NVIDIA Ampere (RTX 30xx consumer)
 - `compute_89,code=sm_89`: NVIDIA Ada Lovelace (RTX 40xx)
-- `compute_90,code=sm_90`: NVIDIA Hopper (H100)
+- `compute_90,code=sm_90`: NVIDIA Hopper (H100, H800, H200)
 
 ### Building Documentation
 
--- a/meson.build
+++ b/meson.build
@@ -92,22 +92,29 @@ if cuda_dep.found()
     cuda_args_option = get_option('cuda_args')
     cuda_link_args_option = get_option('cuda_link_args')
 
-    cuda_args = [
+    # Define default architectures
+    default_cuda_args = [
         '-gencode=arch=compute_80,code=sm_80',
         '-gencode=arch=compute_90,code=sm_90'
     ]
-    cuda_link_args = [
+    default_cuda_link_args = [
         '-gencode=arch=compute_80,code=sm_80',
         '-gencode=arch=compute_90,code=sm_90'
     ]
 
     # Apply defaults only if user hasn't specified custom args
     if cuda_args_option == []
+        cuda_args = default_cuda_args
         add_project_arguments(cuda_args, language: 'cuda')
+    else
+        cuda_args = []
     endif
 
     if cuda_link_args_option == []
+        cuda_link_args = default_cuda_link_args
         add_project_link_arguments(cuda_link_args, language: 'cuda')
+    else
+        cuda_link_args = []
     endif
 
     # Refer to https://mesonbuild.com/Cuda-module.html


meson.build
```diff
@@ -67,20 +67,20 @@ if cuda_dep.found()
     cuda = import('unstable-cuda')
     nvcc = meson.get_compiler('cuda')
 
-    # Please extend with your arch if not present in the list
     nvcc_flags = []
-    nvcc_flags += ['-gencode', 'arch=compute_80,code=sm_80']
-    nvcc_flags += ['-gencode', 'arch=compute_90,code=sm_90']
+    nvcc_flags_link = []
+    if get_option('nvcc_gencode') != ''
```
yosefe: why need this check? wouldn't the for loop below be empty any way if options not set?

meson_options.txt
```diff
@@ -23,6 +23,8 @@ option('gds_path', type: 'string', value: '/usr/local/cuda/', description: 'Path
 option('cudapath_inc', type: 'string', value: '', description: 'Include path for CUDA')
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
+option('nvcc_gencode', type: 'string', value: '-gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_90,code=sm_90',
```
yosefe: It seems like can provide meson cuda build flags like this:
```
meson setup -Dcuda_args="-gencode=arch=compute_90,code=sm_90" \
            -Dcuda_link_args="-gencode=arch=compute_90,code=sm_90"
```
Can you pls check it? so maybe no need to add specific option

meson.build
michal-shalev: Because with a string option, ''.split(' ') returns [''], so the loop runs once and adds an empty flag. I changed `nvcc_gencode` to an array and removed this check and split to avoid empty args.

meson_options.txt
michal-shalev: I tested `-Dcuda_args` and -`Dcuda_link_args` and it's working.
I've updated the PR to use meson's built-in `cuda_args` and `cuda_link_args` instead of adding a custom option.

README.md
```diff
@@ -114,6 +114,11 @@ $ meson setup <name_of_build_dir> \
     -Ducx_path=/path/to/ucx \     # Custom UCX installation path
     -Dinstall_headers=true \      # Install development headers
     -Ddisable_gds_backend=false   # Enable GDS backend
+
+# CUDA GPU architecture targeting (example)
+$ meson setup <name_of_build_dir> \
+    -Dcuda_args="-gencode=arch=compute_80,code=sm_80,-gencode=arch=compute_90,code=sm_90" \
```
ovidiusm: Is this comma separated or space separated?

michal-shalev: space, fixed and tested

ovidiusm (approved):

meson.build
```diff
@@ -82,24 +82,44 @@ else
...
+    # See README.md for details and examples
+    cuda_args_option = get_option('cuda_args')
+    cuda_link_args_option = get_option('cuda_link_args')
+
+    # Define default architectures
+    default_cuda_args = [
+        '-gencode=arch=compute_80,code=sm_80',
+        '-gencode=arch=compute_90,code=sm_90'
+    ]
+    default_cuda_link_args = [
```
tvegas1, line 100: remove and use `default_cuda_args` instead

meson.build
```diff
@@ -82,24 +82,44 @@ else
...
+        '-gencode=arch=compute_80,code=sm_80',
+        '-gencode=arch=compute_90,code=sm_90'
+    ]
+
+    # Apply defaults only if user hasn't specified custom args
+    if cuda_args_option == []
+        cuda_args = default_cuda_args
+        add_project_arguments(cuda_args, language: 'cuda')
+    else
+        cuda_args = []
```
tvegas1, line 110: i guess we need to `add_project_argument` with `cuda_args_option`?

meson.build
```diff
@@ -82,24 +82,44 @@ else
...
+    endif
+
+    if cuda_link_args_option == []
+        cuda_link_args = default_cuda_link_args
+        add_project_link_arguments(cuda_link_args, language: 'cuda')
+    else
+        cuda_link_args = []
```
tvegas1, line 117: same as above?

meson.build
michal-shalev: No, meson has built-in cuda_args and cuda_link_args options, but if it's not clear I can add another comment here

meson.build
michal-shalev: notice that previously we had `nvcc_flags` and `nvcc_flags_link`, I did not won't to change that logic, only to use the built-in meson option

README.md
```diff
@@ -123,6 +128,32 @@ Common build options:
...
+    -Dcuda_args="-gencode=arch=compute_75,code=sm_75 -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_90,code=sm_90" \
+    -Dcuda_link_args="-gencode=arch=compute_75,code=sm_75 -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_90,code=sm_90"
+```
+
+Common CUDA compute capabilities:
+- `compute_75,code=sm_75`: NVIDIA Turing (RTX 20xx, Tesla T4)
+- `compute_80,code=sm_80`: NVIDIA Ampere (A100, RTX 30xx)
+- `compute_86,code=sm_86`: NVIDIA Ampere (RTX 30xx consumer)
+- `compute_89,code=sm_89`: NVIDIA Ada Lovelace (RTX 40xx)
+- `compute_90,code=sm_90`: NVIDIA Hopper (H100, H800, H200)
```
ovidiusm, line 156: Question 1: do we support Blackwell? Why is it not listed?
Question 2: what should be the value when packaging the wheel, since that needs to cover all platforms where the user may do `pip install nixl`

meson.build
tvegas1: i mean that `default_cuda_link_args` and `default_cuda_args` are the same, so we can remove the first one and use `default_cuda_args` line 114 too

meson.build
tvegas1: makes sense, but also why `cuda_args=[]`? i would have assumed we need to pass it to the `cuda.compiles()` below? or is it implictly passed already? in that case we could remove the `args: cuda_args`?

meson.build
michal-shalev: `nvcc_flags` and `nvcc_flags_link` were the same, I can set `default_cuda_link_args` to `default_cuda_args`

tvegas1: i am suggesting to remove `default_cuda_link_args = {}` at line 100, and replace all other occurences of `default_cuda_link_args` by `default_cuda_args`.

README.md
michal-shalev: 1. I didn't test it on a cluster with Blackwell, and I don't think we'll have time to test it for this release.
2. IMO it should stay the same, there are still defaults

PR #720, opened 2025-08-24
BUILD: detect UCX GPU device API

michal-shalev:
## What?
Add Meson detection for UCX GPU device API.

## Why?
Ensure build enables GPU device API features only when supported.

## How?
Check for `ucp/api/ucp.cuh` and `ucp_gpu_dlist_export` in `ucp.h`,  
define `HAVE_UCX_GPU_DEVICE_API` if found, with option to disable explicitly.

### DIFF
--- a/src/plugins/ucx/meson.build
+++ b/src/plugins/ucx/meson.build
@@ -20,16 +20,26 @@ cpp = meson.get_compiler('cpp')
 compile_flags = []
 if cuda_dep.found()
     compile_flags = [ '-DHAVE_CUDA' ]
+    cuda = meson.get_compiler('cuda')
 
     # UCX GPU device API detection
-    if cpp.has_function('ucp_mem_list_create',
-                        prefix: '#include <ucp/api/device/ucp_host.h>',
-                        dependencies: ucx_dep)
+    have_gpu_side = cuda.compiles('''
+            #include <ucp/api/device/ucp_device_impl.h>
+            int main() { return 0; }
+        ''', dependencies : ucx_dep)
+
+    have_host_side = cpp.has_function(
+        'ucp_mem_list_create',
+        prefix: '#include <ucp/api/device/ucp_host.h>',
+        dependencies: ucx_dep)
+
+    if have_gpu_side and have_host_side
         compile_flags += ['-DHAVE_UCX_GPU_DEVICE_API']
         message('UCX GPU device API support detected and enabled')
     else
         message('UCX GPU device API support not available - GPU device API features will be disabled')
     endif
+
 endif
 
 if 'UCX' in static_plugins


meson.build
```diff
@@ -130,6 +130,23 @@ else
   ucx_dep = dependency('ucx', modules: ['ucx::ucs', 'ucx::ucp', 'ucx::uct'])
 endif
 
+# UCX GPU device API detection
```
aranadive: All of this should go under the src/plugins/ucx/meson.build file

meson_options.txt
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 option('ucx_path', type: 'string', value: '', description: 'Path to UCX install')
+option('enable_ucx_gpu_device_api', type: 'boolean', value: true,
```
aranadive: Is there a need for this? Shouldnt we always use the device API if available?

michal-shalev: This option allows us to easily isolate issues related to the GPU device API and provides a fallback for environments where it might be unstable.

meson.build

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,44 @@
 
 ucx_utils_dep = declare_dependency(link_with: ucx_utils_lib, include_directories: utils_inc_dirs )
 asio_dep = [dependency('asio', required: true)]
+cpp  = meson.get_compiler('cpp')
```
brminich: minor:
```suggestion
cpp = meson.get_compiler('cpp')
```

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,44 @@
...
+    # UCX GPU device API detection
+    ucx_gpu_device_api_available = false
+    if get_option('enable_ucx_gpu_device_api')
+        have_ucp_cuh = cuda.compiles('''
+            #include <ucp/api/device/inline/cuda/ucp.cuh>
+            int main() { return 0; }
+        ''', dependencies : ucx_dep)
+
+        have_gpu_dlist_export = cpp.has_function(
+            'ucp_gpu_dlist_export',
```
brminich: it is `ucp_gpu_mem_list_create` now in `src/ucp/api/device/ucp_host.h`

michal-shalev: done

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,44 @@
...
+        )
+
+        if have_ucp_cuh and have_gpu_dlist_export
+            ucx_gpu_device_api_available = true
+            compile_flags += ['-DHAVE_UCX_GPU_DEVICE_API']
+            message('UCX GPU device API support detected and enabled')
+        endif
+    endif
+else
+    ucx_gpu_device_api_available = false
```
aranadive: Move this to before the if condition and remove the else

aranadive: Actually you dont need this bool. Just print the message in an else part for if have_ucp_cuh ...

meson_options.txt
aranadive: This should be renamed to use_ucx_gpu_device_api since the other compile options actually enable the API use.

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,44 @@
...
 if cuda_dep.found()
     compile_flags = [ '-DHAVE_CUDA' ]
+
+    cuda = meson.get_compiler('cuda')
+
+    # UCX GPU device API detection
+    ucx_gpu_device_api_available = false
+    if get_option('enable_ucx_gpu_device_api')
+        have_ucp_cuh = cuda.compiles('''
+            #include <ucp/api/device/inline/cuda/ucp.cuh>
```
ofirfarjun7: I'm not sure we will use .cuh extension.

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,44 @@
...
+            int main() { return 0; }
+        ''', dependencies : ucx_dep)
+
+        have_gpu_dlist_export = cpp.has_function(
+            'ucp_gpu_mem_list_create',
+            prefix: '#include <ucp/api/device/ucp_host.h>',
+            dependencies: ucx_dep
+        )
+
+        if have_ucp_cuh and have_gpu_dlist_export
```
ofirfarjun7: Why we need both checks?

meson_options.txt
michal-shalev: done

src/plugins/ucx/meson.build
michal-shalev: done

src/plugins/ucx/meson.build
michal-shalev: removed have_ucp_cuh

src/plugins/ucx/meson.build
michal-shalev: removed have_ucp_cuh

meson_options.txt
aranadive: Wasnt renamed here ..

michal-shalev: Can you please re-check?

meson_options.txt
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 option('ucx_path', type: 'string', value: '', description: 'Path to UCX install')
+option('use_ucx_gpu_device_api', type: 'boolean', value: true,
```
yosefe: The name too long.. do we really need this option ? why not always enable if it is detected?

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,29 @@
...
 asio_dep = [dependency('asio', required: true)]
+cpp = meson.get_compiler('cpp')
 
 compile_flags = []
 if cuda_dep.found()
     compile_flags = [ '-DHAVE_CUDA' ]
+
+    # UCX GPU device API detection
+    if get_option('use_ucx_gpu_device_api')
+        if cpp.has_function('ucp_gpu_mem_list_create',
```
yosefe: can we also test actual device-side API?

michal-shalev: I did but then I removed it following https://github.com/ai-dynamo/nixl/pull/720#discussion_r2308337621
In UCX initially it was a cuda header file, so I checked if it compiles, now it's a regular header, should I check it anyways?

meson_options.txt
michal-shalev: I asked specifically if this should be optional and you said yes, but I can change it

yosefe: optional meaning that if the header/function is not found, the build will continue without ucx device api. not that we need explicit meson option for it

src/plugins/ucx/meson.build
yosefe: ic, let's discuss it offline

src/plugins/ucx/meson.build
```diff
@@ -15,10 +15,29 @@
...
+            compile_flags += ['-DHAVE_UCX_GPU_DEVICE_API']
+            message('UCX GPU device API support detected and enabled')
+        else
+            message('UCX GPU device API support not available - GPU device API features will be disabled')
+        endif
+    else
+        message('UCX GPU device API support explicitly disabled by user')
+    endif
+else
+    if get_option('use_ucx_gpu_device_api')
```
ovidiusm: This looks a bit strange, why are we chaining both dependencies here instead of checking separately?

michal-shalev: removed it

meson_options.txt
michal-shalev: removed

src/plugins/ucx/meson.build
michal-shalev: re-added it

ovidiusm (approved):

yosefe (approved):

PR #722, opened 2025-08-25
Fix Error in barrier when testing block size above 1M

Pyngon:
## What?
Fix an issue where target stuck at waiting for more data from initiator due to wrong expectation

## Why?
warm up iterations used in initiator is per thread, so target have to multiply that by thread.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -989,10 +989,12 @@ xferBenchNixlWorker::transfer(size_t block_size,
         num_iter /= xferBenchConfig::large_blk_iter_ftr;
     }
 
-    ret = execTransfer(
-        agent, local_iovs, remote_iovs, xfer_op, skip, xferBenchConfig::num_threads, stats);
-    if (ret < 0) {
-        return std::variant<xferBenchStats, int>(ret);
+    if (skip > 0) {
+        ret = execTransfer(
+            agent, local_iovs, remote_iovs, xfer_op, skip, xferBenchConfig::num_threads, stats);
+        if (ret < 0) {
+            return std::variant<xferBenchStats, int>(ret);
+        }
     }
 
     // Synchronize to ensure all processes have completed the warmup (iter and polling)
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -1024,8 +1024,8 @@ xferBenchNixlWorker::poll(size_t block_size) {
     // Reduce skip by 10x for large block sizes
     if (block_size > LARGE_BLOCK_SIZE) {
         skip /= xferBenchConfig::large_blk_iter_ftr;
-        if (skip < MIN_WARMUP_ITERS) {
-            skip = MIN_WARMUP_ITERS;
+        if (skip < MIN_WARMUP_ITERS * xferBenchConfig::num_threads) {
+            skip = MIN_WARMUP_ITERS * xferBenchConfig::num_threads;
         }
         num_iter /= xferBenchConfig::large_blk_iter_ftr;
     }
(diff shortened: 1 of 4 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1006,8 +1006,8 @@ xferBenchNixlWorker::poll(size_t block_size) {
     // Reduce skip by 10x for large block sizes
     if (block_size > LARGE_BLOCK_SIZE) {
         skip /= xferBenchConfig::large_blk_iter_ftr;
-        if (skip < MIN_WARMUP_ITERS) {
-            skip = MIN_WARMUP_ITERS;
+        if (skip < MIN_WARMUP_ITERS * xferBenchConfig::num_threads) {
```
aranadive: If warmup_iter is set to 0 no need to poll for these iters. Similarly should be addressed in transfer function.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1006,8 +1006,8 @@ xferBenchNixlWorker::poll(size_t block_size) {
...
+            skip = MIN_WARMUP_ITERS * xferBenchConfig::num_threads;
```
yosefe: maybe use std::max?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: In this case, i guess we do not need to have min_warmp_iterations. I can remove that.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: It looks like we don't really need to have minimum warmup iterations, I went ahead to remove this check.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: Removed the MIN_WARMUP_ITERS and allow it to be 0

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -968,9 +968,6 @@ xferBenchNixlWorker::transfer(size_t block_size,
     // Reduce skip by 10x for large block sizes
     if (block_size > LARGE_BLOCK_SIZE) {
         skip /= xferBenchConfig::large_blk_iter_ftr;
-        if (skip < MIN_WARMUP_ITERS) {
-            skip = MIN_WARMUP_ITERS;
-        }
         num_iter /= xferBenchConfig::large_blk_iter_ftr;
     }
```
aranadive, line 991: Value of skip can be 0 here so no need to call execTransfer for skip==0.

benchmark/nixlbench/src/worker/nvshmem/nvshmem_worker.cpp
```diff
@@ -182,9 +182,6 @@ xferBenchNvshmemWorker::transfer(size_t block_size,
     // Reduce skip by 10x for large block sizes
     if (block_size > LARGE_BLOCK_SIZE) {
         skip /= xferBenchConfig::large_blk_iter_ftr;
-        if (skip < MIN_WARMUP_ITERS) {
-            skip = MIN_WARMUP_ITERS;
-        }
         num_iter /= xferBenchConfig::large_blk_iter_ftr;
     }
```
aranadive, line 187: Value of skip can be 0 here so no need to call execTransfer for skip==0.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
Pyngon: Added a check for that.

benchmark/nixlbench/src/worker/nvshmem/nvshmem_worker.cpp
Pyngon: Added a check for that.

PR #725, opened 2025-08-25
CI: Download rustup in a more reliable way

ovidiusm:
## What?
Download rustup with multiple retries and from static.rust-lang.org instead of sh.rustup.rs

## Why?
Many failures in CI are due to this download failing with connection timeout.

## How?
Use wget with retry mechanism as in the Dockerfiles

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -51,6 +51,7 @@ $SUDO rm -rf /usr/lib/cmake/grpc /usr/lib/cmake/protobuf
 
 $SUDO apt-get -qq update
 $SUDO apt-get -qq install -y curl \
+                             wget \
                              libnuma-dev \
                              numactl \
                              autotools-dev \


.gitlab/build.sh
```diff
@@ -93,7 +94,9 @@ $SUDO apt-get -qq install -y curl \
                              clang \
                              libcurl4-openssl-dev zlib1g-dev # aws-sdk-cpp dependencies
 
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.86.0
+wget --tries=3 --waitretry=5 https://static.rust-lang.org/rustup/dist/${ARCH}-unknown-linux-gnu/rustup-init
```
brminich, line 97: maybe better to use `--retry` and `--retry-delay` options or curl?

ovidiusm: That's how I did it initially, but then I noticed that we are downloading rustup in 2 more places in these Dockerfiles:

* https://github.com/ai-dynamo/nixl/blob/main/contrib/Dockerfile#L87
* https://github.com/ai-dynamo/nixl/blob/main/contrib/Dockerfile.manylinux#L114

It's better to have everything working the same way

brminich (approved):

PR #729, opened 2025-08-26
Improve populate method performance for not fully sorted inputs.

mkhazraee:
## What?
Improve populate method performance for not fully sorted inputs.

## Why?
In some use cases the input is not fully sorted, but has chunks of sorted regions. Also this removes the need from user indicating and making sure the input is sorted, as benefits become very marginal.

## How?
- Find the first element in LogN search time. Then assume it's sorted unless it is not, and search for that non-conforming element in LogN search time.
- Replaced check for nixl_sec_dlist_t to be sorted with assert as it should always be.
- For fully sorted lists this would add one extra check.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 #include <map>
+#include <cassert>
+#include <algorithm>
 #include <iostream>
 #include "nixl.h"
 #include "nixl_descriptors.h"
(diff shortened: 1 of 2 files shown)


ovidiusm (dismissed):
Seems correct but it would be great if we had some unit tests for this

mkhazraee:
@ovidiusm regarding testing the new method, ./test/nixl/agent_example is testing it. I added some extra prints in local test to further confirm:
- If list is sorted (which we have 8 sorted lists of 2M entries in this test), it always runs the sorted path in the alg (since it's doing the check on query, it was impossible for it not to happen. Still did an assert check and it was fine).
- If the list is not sorted, every single time I ran the test, even those 8 entries in the sideXferTest function ends up with some consecutive elements being in order and some not to be, at different indices. And there is a check for those transfers in the test, so they were fine. I increased the n further to make sure (where I caught a const value bug in the test and pushed the fix), and added a print that says when list was not sorted. It happens every time. 
- The second bullet covers the first element LogN search too, which is working as expected. 

So that test really covered all the corner cases, and I further double checked with extra prints (e.g., to see some elements are just out of order), I say good to merge. Complicating the code to make sure it always becomes unsorted is not that valuable, as I tested several times it was not sorted and worked fine.

ovidiusm (approved):

PR #731, opened 2025-08-26
Remove sorted flag from user facing nixlDescList

mkhazraee:
## What?
Sorted is only needed in nixl_sec_dlist_t, so using inheritance to move it there and not exposed to the user.

## Why?
Sorted flag caused confusion while unused. It removes some related methods as well.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 #include <map>
-#include <cassert>
 #include <algorithm>
 #include <iostream>
 #include "nixl.h"
@@ -50,7 +49,6 @@ nixl_status_t nixlMemSection::populate (const nixl_xfer_dlist_t &query,
         return NIXL_ERR_NOT_FOUND;
 
     nixl_sec_dlist_t* base = it->second;
-    // assert(base->verifySorted());
     resp.resize(query.descCount());
 
     int size = base->descCount();
--- a/test/nixl/agent_example.cpp
+++ b/test/nixl/agent_example.cpp
@@ -411,9 +411,6 @@ nixl_status_t sideXferTest(nixlAgent* A1, nixlAgent* A2, nixlXferReqH* src_handl
     src_list = mem_list1.trim();
     dst_list = mem_list2.trim();
 
-    if (!src_list.verifySorted() || !dst_list.verifySorted())
-        std::cout << "src_list or dst_list is not sorted, covering corner cases.";
-
     status = A1->registerMem(mem_list1, &extra_params1);
     assert (status == NIXL_SUCCESS);
 


mkhazraee:
Note that this PR includes PR 729.

roiedanino (dismissed):

src/api/cpp/nixl_descriptors.h
```diff
@@ -211,35 +206,38 @@ class nixlDescList {
      * @brief Parametrized Constructor for nixlDescList
      *
      * @param type         NIXL memory type of descriptor list
-     * @param sorted       Flag to set sorted option (default = false)
      * @param init_size    initial size for descriptor list (default = 0)
      */
-    nixlDescList(const nixl_mem_t &type, const bool &sorted = false, const int &init_size = 0);
+    nixlDescList(const nixl_mem_t &type, const int &init_size = 0);
```
ovidiusm, line 211: Can we make such API changes?

src/infra/nixl_descriptors.cpp
```diff
@@ -403,3 +343,62 @@ template bool operator==<nixlBlobDesc>(const nixlDescList<nixlBlobDesc> &lhs,
                                        const nixlDescList<nixlBlobDesc> &rhs);
 template bool operator==<nixlSectionDesc>(const nixlDescList<nixlSectionDesc> &lhs,
                                           const nixlDescList<nixlSectionDesc> &rhs);
+
+// nixlSecDescList keeps the elements sorted
+void
+nixlSecDescList::addDesc(const nixlSectionDesc &desc) {
```
ovidiusm, line 349: Adding one by one in random order will have n^2 * logn performance because of the shifts

Would be better to add an extra method to add a whole list in a single shot, in case that is a usecase that matters

src/infra/nixl_memory_section.cpp
```diff
@@ -39,109 +41,52 @@ nixl_status_t nixlMemSection::populate (const nixl_xfer_dlist_t &query,
...
         return NIXL_ERR_INVALID_PARAM;
 
     section_key_t sec_key = std::make_pair(query.getType(), backend);
     auto it = sectionMap.find(sec_key);
     if (it==sectionMap.end())
         return NIXL_ERR_NOT_FOUND;
 
-    nixlBasicDesc *p;
     nixl_sec_dlist_t* base = it->second;
+    // assert(base->verifySorted());
```
ovidiusm: Is this needed?

ovidiusm: I suppose that you want to merge #729 separately

mkhazraee: Yeah it's a redundant assert, I just added that for 729 as I was removing the redundant if, but totally fine to remove it.

src/api/cpp/nixl_descriptors.h
mkhazraee: We wanted to do this for a while, and why I'm pushing to do it sooner than later. No one is using it as of now as far as we know (other than our own code in SGLang set it to false which I'll remove), and it's just an optional part getting removed.

src/infra/nixl_descriptors.cpp
mkhazraee: This practically only happens once per registration. So the n is small here, and we register the descriptors with the any backend one by one. If that becomes a performance bottleneck (even though just during setup), def something to consider.

src/infra/nixl_memory_section.cpp
mkhazraee: Removed.

ovidiusm (approved):

brminich (approved):

gleon99 (approved):

PR #732, opened 2025-08-26
[HF3FS] use threadpool to poll for completion

Pyngon:
## What?
Create a threadpool to poll for completion

## Why?
To not create a thread for every single postXfer.

### DIFF
--- a/src/plugins/hf3fs/hf3fs_backend.cpp
+++ b/src/plugins/hf3fs/hf3fs_backend.cpp
@@ -30,6 +30,8 @@
 #define NUM_CQES 1024
 #define HF3FS_DEFAULT_IOPOOL_SIZE 64
 #define HF3FS_MAX_IOPOOL_SIZE (1 << 20)
+#define HF3FS_THREADPOOL_SIZE 4
+#define HF3FS_THREADPOOL_MAX 1024
 
 long nixlHf3fsEngine::page_size = sysconf(_SC_PAGESIZE);
 
@@ -37,6 +39,7 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
     : nixlBackendEngine(init_params),
       mem_config(NIXL_HF3FS_MEM_CONFIG_AUTO),
       iopool_size(HF3FS_DEFAULT_IOPOOL_SIZE) {
+    int threadpool_size = HF3FS_THREADPOOL_SIZE;
     hf3fs_utils = new hf3fsUtil();
 
     this->initErr = false;
@@ -64,6 +67,7 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
                 return;
             }
         }
+
         if (init_params->customParams->count("iopool_size") > 0) {
             int size = atoi(init_params->customParams->at("iopool_size").c_str());
             if (size > 0) {
@@ -76,6 +80,19 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
                 }
             }
         }
+
+        if (init_params->customParams->count("threadpool_size") > 0) {
+            int size = atoi(init_params->customParams->at("threadpool_size").c_str());
+            if (size > 0) {
+                if (size <= HF3FS_THREADPOOL_MAX) {
+                    threadpool_size = size;
+                } else {
+                    threadpool_size = HF3FS_THREADPOOL_MAX;
+                    NIXL_WARN << "threadpool_size param " << threadpool_size << " exceeded max "
+                              << HF3FS_THREADPOOL_MAX << ", set to max";
+                }
+            }
+        }
     }
 
     char mount_point_cstr[256];
@@ -84,7 +101,6 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
         this->initErr = true;
         return;
     }
-
     hf3fs_utils->mount_point = mount_point_cstr;
 
     for (unsigned int i = 0; i < iopool_size; i++) {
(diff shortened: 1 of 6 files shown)


src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -61,6 +65,19 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
                 return;
             }
         }
+
+        if (init_params->customParams->count("threadpool_size") > 0) {
```
aranadive, line 84: Please add to nixlbench and 3fs test

aranadive:
@Pyngon how does this improve performance and cpu utilization? Can you measure?

src/plugins/hf3fs/hf3fs_backend.cpp
Pyngon: Done

Pyngon:
> @Pyngon how does this improve performance and cpu utilization? Can you measure?

@aranadive  Performance is similar in my setup. I have measure the time taken for transfer, creating thread and threadpool, I can see that the waiting time for transfer 1K block is around 650us. Creating thread takes around 15us and posting asio task take around 2-5us. So the performance improvement is not obvious here. So this patch is more on limiting resource usage (do not create thread for every transfer).

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -482,39 +499,52 @@ nixl_status_t nixlHf3fsEngine::checkXfer(nixlBackendReqH* handle) const
...
     return NIXL_SUCCESS;
 }
 
 nixl_status_t nixlHf3fsEngine::releaseReqH(nixlBackendReqH* handle) const
 {
     nixlHf3fsBackendReqH *hf3fs_handle = (nixlHf3fsBackendReqH *) handle;
 
-    cleanupIOThread(hf3fs_handle);
+    hf3fs_handle->stopped = true;
+    while (hf3fs_handle->is_posted) {
```
ovidiusm, line 526: Does is_posted need synchronization, since it is accessed from multiple threads?

Pyngon: Do you mean make it an atomic variable? I can do that.

vvenkates27:
> > @Pyngon how does this improve performance and cpu utilization? Can you measure?
> 
> @aranadive Performance is similar in my setup. I have measure the time taken for transfer, creating thread and threadpool, I can see that the waiting time for transfer 1K block is around 650us. Creating thread takes around 15us and posting asio task take around 2-5us. So the performance improvement is not obvious here. So this patch is more on limiting resource usage (do not create thread for every transfer).

@Pyngon what is the SOL of the platform (i.e Max possible with something like FIO on 3FS) and
what is the performance achieved without thread pool? 
and what is the number with thread pool.  Also why are we trying for 1K blocks? I would try with larger blocks atleast like 128k etc.

Pyngon:
> @Pyngon what is the SOL of the platform (i.e Max possible with something like FIO on 3FS) and what is the performance achieved without thread pool? and what is the number with thread pool. Also why are we trying for 1K blocks? I would try with larger blocks atleast like 128k etc.

Using small block size because any performance improvement should be most obvious on small block size as the transfer time is the lowest. What is the reason to use big block size? What are you trying to see from that?

The performance achieved with and without thread pool is quiet similar, to give some example from a run.
For 1K block,
- non-thread pool is 0.000946GB/s
- thread pool is 0.001038GB/s

For 8M block,
- non thread pool is 2.846611GB/s
- thread pool is 2.779871

The number will have some variation, and sometimes non-thread pool will be better.

I don't have the number for FIO, but that is not too relevant here.

vvenkates27:
> > @Pyngon what is the SOL of the platform (i.e Max possible with something like FIO on 3FS) and what is the performance achieved without thread pool? and what is the number with thread pool. Also why are we trying for 1K blocks? I would try with larger blocks atleast like 128k etc.
> 
> Using small block size because any performance improvement should be most obvious on small block size as the transfer time is the lowest. What is the reason to use big block size? What are you trying to see from that?
> 
> The performance achieved with and without thread pool is quiet similar, to give some example from a run. For 1K block,
> 
> * non-thread pool is 0.000946GB/s
> * thread pool is 0.001038GB/s
> 
> For 8M block,
> 
> * non thread pool is 2.846611GB/s
> * thread pool is 2.779871
> 
> The number will have some variation, and sometimes non-thread pool will be better.
> 
> I don't have the number for FIO, but that is not too relevant here.

The FIO number will help us determine the max b/w possible in the system without NIXL. FIO also allows to use multiple threads to increase number of inflight IOs. But from your numbers even with 8M block we are just seeing 2.8GB/s which is too little bandwidth to see any benefit from threadpool. This configuration is bottlenecked by the storage bandwidth and adding more threads will only add overhead. I would suggest to test in a larger configuration where there is unused bandwidth to saturate while running with a single thread and try there.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -97,7 +113,14 @@ nixlHf3fsEngine::nixlHf3fsEngine(const nixlBackendInitParams *init_params)
         iopool.push_back(io);
     }
 
-    NIXL_DEBUG << "HF3FS: page size " << page_size << " iopool_size " << iopool_size;
+    io_ctx = new asio::io_context();
+    wg = new asio::executor_work_guard<asio::io_context::executor_type>(io_ctx->get_executor());
```
vvenkates27, line 117: @w1ldptr can you review for asio usage?

w1ldptr, line 117: Why do you need to use io_context here at all? Looking on how you use it just creating a simple asio::thread_pool without manually managing io_context, executor_work_guard and threads should be enough to just schedule tasks on it.

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -398,73 +407,81 @@ nixl_status_t nixlHf3fsEngine::postXfer (const nixl_xfer_op_t &operation,
...
+        // Repost to handle the remaining io
+        asio::post(
+            *ctx, [ctx, handle, utils]() { nixlHf3fsEngine::waitForIOsAsync(ctx, handle, utils); });
+    } else {
+        // It is possible that the user thread post new request at this point
+        // checkXfer has to detect that and post this async task if needed.
+        handle->is_posted = false;
     }
 
+async_out:
```
w1ldptr, line 484: Don't use goto to clean up resources in C++ code, it has far too many pitfalls. This simple scenario can be served much more cleanly with wrapping `cqes` into a smart pointer.

PR #737, opened 2025-08-28
Added pythonic objects for xfer and dlist handles

mkhazraee:
## What?
Added pythonic objects for xfer and dlist handles

## Why?
For some of the current use cases, users create the xfer_req and check for it until it's done, call the release, and then remove it from a list. To make it more pythonic, made the xfer_req to be a python object that has del method, so when removed from the list, would automatically be called. Also helps with type checking in compile time.

## How?
Added two wrapper classes for the handles, and some optimizations to make the post and check times minimal on critical path. Also added some protection code for when a handle cannot be released (the transfer is not complete), both in the delete method of the handle, and then delete of the agent to retry to release those elements.

### DIFF
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -27,10 +27,96 @@ logger = get_logger(__name__)
 
 DEFAULT_COMM_PORT = nixlBind.DEFAULT_COMM_PORT
 
-# Opaque nixl handle types
+
+"""
+@brief Opaque handle wrapper for a prepared transfer descriptor list.
+       Use release() to explicitly free resources; __del__ performs best-effort cleanup.
+@param agent Owning nixl_agent used to perform release operations.
+@param value Internal handle
+"""
+
+
+class nixl_prepped_dlist_handle:
+    __slots__ = ("_handle", "_agent", "_released")
+
+    def __init__(self, agent, value: int):
+        self._handle = int(value)
+        self._agent = agent
+        self._released = False
+
+    def __repr__(self) -> str:
+        return (
+            f"nixl_prepped_dlist_handle(0x{self._handle:x}, released={self._released})"
+        )
+
+    def release(self):
+        if not self._released:
+            self._agent.releasedDlistH(self._handle)
+            self._released = True
+
+    def __del__(self):
+        if not self._released:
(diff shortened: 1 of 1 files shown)


src/api/python/_api.py
```diff
@@ -489,7 +503,8 @@ def initialize_xfer(
     """
 
     def transfer(self, handle: nixl_xfer_handle, notif_msg: bytes = b"") -> str:
-        status = self.agent.postXferReq(handle, notif_msg)
+        # Using h and walrus operator to enable caching for subsequent calls
+        status = self.agent.postXferReq((h := handle._handle), notif_msg)  # noqa: F841
```
tstamler: what is the check we're disabling here?

src/api/python/_api.py
```diff
@@ -984,3 +1000,93 @@ def get_serialized_descs(self, descs) -> bytes:
...
+                    pass
+                try:
+                    self._agent._leaked_xfer_handles.append(self._handle)
+                except Exception:
+                    pass
+                return
+
+
+# Opaque handle for backend can be just int, as it's not passed to the user
+nixl_backend_handle = int
```
tstamler: maybe we shouldn't just sneak these in at the bottom. I wonder if we're reaching the point where we should split into multiple files

src/api/python/_api.py
mkhazraee: It says h is not used anywhere else. I thought it'll be used for caching across calls, which was a surprise but cursor said, asking more, nope. Will revert, unless I find a @cache way for it, although maybe __slots__ is sufficient already.

src/api/python/_api.py
mkhazraee: I was debating on this too. One idea is to bring this one up, and keep the other 2 here, or just move all 3 up.

src/api/python/_api.py
mkhazraee: Confirmed slots is sufficient, removed.

src/api/python/_api.py
mkhazraee: Moved up.

ovidiusm (approved):

PR #738, opened 2025-08-28
Added error logs throughout nixl_agent.

mkhazraee:
## What?
Added error logs throughout nixl_agent.

## Why?
To help with debugging.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -42,6 +42,18 @@
 /* ERROR log prefixed with current function name and a colon */
 #define NIXL_ERROR_FUNC NIXL_ERROR << __FUNCTION__ << ": "
 
+#define SERIALIZE_ERR(ret)                         \
+    do {                                           \
+        NIXL_ERROR_FUNC << "serialization failed"; \
+        return ret;                                \
+    } while (0)
+
+#define DESERIALIZE_ERR()                            \
+    do {                                             \
+        NIXL_ERROR_FUNC << "deserialization failed"; \
+        return NIXL_ERR_MISMATCH;                    \
+    } while (0)
+
 const char TELEMETRY_ENABLED_VAR[] = "NIXL_TELEMETRY_ENABLE";
 static const std::vector<std::vector<std::string>> illegal_plugin_combinations = {
     {"GDS", "GDS_MT"},
@@ -1347,31 +1359,27 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
 
     nixlSerDes sd;
     ret = sd.addStr("Agent", data->name);
-    if (ret) goto getLocalMDerr;
+    if (ret) SERIALIZE_ERR(ret);
 
     ret = sd.addBuf("Conns", &conn_cnt, sizeof(conn_cnt));
-    if (ret) goto getLocalMDerr;
+    if (ret) SERIALIZE_ERR(ret);
 
     for (auto &c : data->connMD) {
         nixl_backend = c.first;
         ret = sd.addStr("t", nixl_backend);
-        if (ret) goto getLocalMDerr;
+        if (ret) SERIALIZE_ERR(ret);
         ret = sd.addStr("c", c.second);
-        if (ret) goto getLocalMDerr;
+        if (ret) SERIALIZE_ERR(ret);
     }
 
     ret = sd.addStr("", "MemSection");
-    if (ret) goto getLocalMDerr;
+    if (ret) SERIALIZE_ERR(ret);
 
     ret = data->memorySection->serialize(&sd);
-    if (ret) goto getLocalMDerr;
+    if (ret) SERIALIZE_ERR(ret);
 
     str = sd.exportStr();
     return NIXL_SUCCESS;
-
-getLocalMDerr:
-    NIXL_ERROR_FUNC << "serialization failed";
-    return ret;
 }
 
 nixl_status_t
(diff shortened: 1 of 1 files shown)


aranadive:
Why not readjust the nixl log macros to log the function name always? You can do the following in the agent and then just update the message appropriately without needing to write the function name every time. Other option is to define a new macros in utils/common/nixl_log.h.

```
#undef NIXL_ERROR
#undef NIXL_WARN
#undef NIXL_INFO
#undef NIXL_DEBUG
#undef NIXL_TRACE

#define NIXL_ERROR LOG(ERROR) << "[" << __FUNCTION__ << "]: "
#define NIXL_WARN LOG(WARNING) << "[" << __FUNCTION__ << "]: "
#define NIXL_INFO LOG(INFO) << "[" << __FUNCTION__ << "]: "
#define NIXL_DEBUG VLOG(1) << "[" << __FUNCTION__ << "]: "
#define NIXL_TRACE DVLOG(2) << "[" << __FUNCTION__ << "]: "
```

src/utils/common/nixl_log.h
```diff
@@ -43,6 +43,9 @@
 /* Logs messages unconditionally (maps to Abseil ERROR level) */
 #define NIXL_ERROR LOG(ERROR)
 
+/* ERROR log prefixed with current function name and a colon */
```
aranadive: Just keep in nixl_agent.cpp for now since thats the only place where its used?

src/core/nixl_agent.cpp
```diff
@@ -241,15 +248,18 @@ nixlAgent::getPluginParams (const nixl_backend_t &type,
...
     return NIXL_ERR_NOT_FOUND;
 }
 
 nixl_status_t
 nixlAgent::getBackendParams (const nixlBackendH* backend,
                              nixl_mem_list_t &mems,
                              nixl_b_params_t &params) const {
-    if (!backend)
+    if (!backend) {
+        NIXL_ERROR_FUNC << "backend handle is null";
```
aranadive: "backend handle not provided"

src/core/nixl_agent.cpp
```diff
@@ -304,25 +316,29 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
             if (!backend->supportsNotif()) {
                 delete backend;
+                NIXL_ERROR_FUNC << "backend '" << type << "' supportsRemote but not notifications";
                 return NIXL_ERR_BACKEND;
             }
 
             ret = backend->getConnInfo(str);
             if (ret != NIXL_SUCCESS) {
                 delete backend;
+                NIXL_ERROR_FUNC << "getting connection info failed for '" << type
```
aranadive: "failed to get connection for .. type .. with status: "

src/core/nixl_agent.cpp
```diff
@@ -1338,53 +1481,56 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
...
-    // No common backend, no point in loading the rest, unexpected
     if ((count == 0) && (conn_cnt > 0)) {
+        NIXL_ERROR_FUNC << "no common backend found";
         return NIXL_ERR_BACKEND;
     }
 
-    if (sd.getStr("") != "MemSection") {
-        return NIXL_ERR_MISMATCH;
-    }
+    if (sd.getStr("") != "MemSection") goto loadRemoteMDerr;
```
aranadive: why goto?

src/core/nixl_agent.cpp
```diff
@@ -1321,15 +1460,19 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
...
 
     std::string remote_agent = sd.getStr("Agent");
     if (remote_agent.empty()) {
+        NIXL_ERROR_FUNC << "error in deserializing remote agent name";
         return NIXL_ERR_MISMATCH;
     }
 
     if (remote_agent == data->name) {
+        NIXL_ERROR_FUNC << "remote agent name same as local agent, "
+                           "this does not need to be loaded";
```
aranadive: "no need to load MD"

src/core/nixl_agent.cpp
```diff
@@ -1321,15 +1460,19 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
     NIXL_LOCK_GUARD(data->lock);
     ret = sd.importStr(remote_metadata);
     if (ret != NIXL_SUCCESS) {
+        NIXL_ERROR_FUNC << "failed to deserialize remote metadata";
         return ret;
     }
 
     std::string remote_agent = sd.getStr("Agent");
     if (remote_agent.empty()) {
+        NIXL_ERROR_FUNC << "error in deserializing remote agent name";
```
aranadive, line 1476: "error in deserializing remote agent name or it is empty"

src/core/nixl_agent.cpp
mkhazraee: For any of deserialization issues, didn't want to have the same exact print repeated.

src/core/nixl_agent.cpp
mkhazraee: Empty is the way our deserializer indicates errors. Specially we have a check on sending side, not to be empty, so this can only be error.

aranadive (approved):

src/core/nixl_agent.cpp
```diff
@@ -38,6 +39,21 @@
...
+    do {                                           \
+        NIXL_ERROR_FUNC << "serialization failed"; \
+        return ret;                                \
+    } while (0)
+
+#define DESERIALIZE_ERR()                            \
+    do {                                             \
+        NIXL_ERROR_FUNC << "deserialization failed"; \
+        return NIXL_ERR_MISMATCH;                    \
+    } while (0)
```
yosefe, line 55: The serialize errors should be moved to serializer class, and NIXL_ERROR_FUNC should be moved to the common log.h header

mkhazraee: Agreed, it was there initially, Adit suggested to make that a separate PR, so this PR touches less files. And for moving the logger to serializer, we can do and add a comment here that logging is happening there. Because IMO in nixl agent for any error we should have a log, or say where the log is.

PR #740, opened 2025-08-28
BENCH/NIXL: error handling related fixes

evgeny-leksikov:
## What?
 - handle errors on data and control path
 - fix memory leak in `xferBenchNixlWorker::exchangeIOV`

## Why?
bugfixes in scenarios with failure/killing of target


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -917,37 +929,38 @@ execTransfer(nixlAgent *agent,
...
-                    }
-                } while (NIXL_SUCCESS != rc);
-                const nixlTime::us_t transfer_duration = timer.lap();
-                thread_stats.transfer_duration.add(transfer_duration);
+            while (NIXL_IN_PROG == rc) {
+                /* XXX agent isn't const because the getXferStatus() is not const  */
+                rc = agent->getXferStatus(req);
+            }
+
+            const nixlTime::us_t transfer_duration = timer.lap();
```
aranadive: Only log the transfer duration if it was successful

evgeny-leksikov: fixed

PR #743, opened 2025-08-28
Fallback to manual CUDA detection when meson detection fails

ovidiusm:
## What?
Add manual CUDA detection as a fallback in meson build script for NIXL.

## Why?
Seeing that on lego and some other ARM hosts (but not all images), the builtin CUDA detection of meson does not find libcudart. This is due to missing pkg-config files in the base image.

Before:

```
#26 24.03 Run-time dependency CUDA (modules: cudart, cuda) found: NO (tried system)
```

After:
```
#26 24.03 Run-time dependency CUDA (modules: cudart, cuda) found: NO (tried system) <<<< THIS IS MESON
#26 24.03 WARNING: You should add the boolean check kwarg to the run_command call.
#26 24.03          It currently defaults to false,
#26 24.03          but it will default to true in meson 2.0.
#26 24.03          See also: https://github.com/mesonbuild/meson/issues/9300
#26 24.03 Message: CUDA root: /usr/local/cuda
#26 24.03 ../meson.build:62: WARNING: Project targets '>= 0.64.0' but uses feature introduced in '1.2.0': str.splitlines.
#26 24.03 Message: CUDA libdirs: ['/usr/local/cuda/targets/sbsa-linux/lib', '/usr/local/cuda/targets/sbsa-linux/lib/stubs']
#26 24.03 Library cudart found: YES
#26 24.03 Library cuda found: YES
#26 24.03 Message: Found CUDA installation: /usr/local/cuda <<<< THIS IS MANUAL FALLBACK
```

This is particularly important for the wheel packaging.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -23,6 +23,7 @@ project('nixl', 'CPP', version: '0.5.1',
 
 # set up some global vars for compiler, platform, configuration, etc.
 cpp = meson.get_compiler('cpp')
+fs = import('fs')
 
 dl_dep = cpp.find_library('dl', required: true)
 rt_dep = cpp.find_library('rt', required: true)
--- a/meson.build
+++ b/meson.build
@@ -64,6 +64,9 @@ if cuda_lib_path == ''
         cuda_dep = declare_dependency(
           link_args : ['-L' + cuda_lib, '-L' + cuda_stub, '-lcuda', '-lcudart'],
           include_directories : include_directories(cuda_inc))
+        if cuda_dep.found()
+            message('Found CUDA installation through fallback method:', cuda_home)
+        endif
     endif
 else
     message('cuda lib path ', cuda_lib_path)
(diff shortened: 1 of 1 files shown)


aranadive (approved):

meson.build
```diff
@@ -52,6 +53,24 @@ cuda_stub_path = get_option('cudapath_stub')
 
 if cuda_lib_path == ''
     cuda_dep = dependency('cuda', required : false, modules : [ 'cudart', 'cuda' ])
+    if not cuda_dep.found()
+        # Meson is not detecting CUDA reliably on ARM, fallback to default
+        cuda_home = run_command('bash', '-c', 'echo $CUDA_HOME').stdout().strip()
+        if cuda_home == ''
+            cuda_home = '/usr/local/cuda'
+        endif
+        cuda_lib = cuda_home + '/lib64'
```
yosefe, line 62: is it always lib64?

PR #749, opened 2025-08-31
API/GPU/UCX: Add signal read device API

michal-shalev:
## What?
Added `nixlGpuReadSignalValue` device function to read signal values from GPU memory.

## Why?
Provides GPU kernels with the ability to read signal values directly from device code, completing the signal operations API alongside the existing `nixlGpuPostSignalXferReq`

## How?
Added a templated `nixlGpuReadSignalValue` function in `nixl_device.cuh` that return `NIXL_ERR_NOT_SUPPORTED` as placeholder.


yosefe (dismissed):

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -154,4 +154,17 @@ nixlGpuGetXferStatus(const nixlGpuXferStatusH &xfer_status)
...
+ * @brief Read the value of a signal.
+ *
+ * @param signal_addr [in]  Address of the signal.
+ *
+ * @return The value of the signal.
+ */
+template<nixl_gpu_level_t level = nixl_gpu_level_t::BLOCK>
+__device__ uint64_t
+nixlGpuReadSignalValue(const void *signal_addr) {
+    return NIXL_ERR_NOT_SUPPORTED;
```
rakhmets, line 167:
```suggestion
    return 0;
```
The method returns value, not status.

rakhmets: The method returns value, not status.
```suggestion
    return 0;
```

yosefe (approved):

PR #750, opened 2025-08-31
Telemetry small fixes

rafiw:
## What?

## Why?

## How?
architecture, approach, etc._


examples/python/telemetry_reader.py
```diff
@@ -106,10 +105,6 @@ def __init__(self, file_path, version=1):
     def _open_file(self):
         """Open existing file"""
         self.file_fd = os.open(self.file_path, os.O_RDWR)
-        if self.file_fd == -1:
```
ovidiusm, line 109: Why was this error check and message removed?

rafiw: turns out if os.open fails it throw an exception

ovidiusm: How will the reader program work if the file failed to open?

rafiw: it will throw an exception and application will exit

ovidiusm: Then why not raise it here? I don't understand why this change was made.

src/core/telemetry.h
```diff
@@ -35,10 +35,15 @@ struct periodicTask {
     asio::steady_timer timer_;
     std::function<bool()> callback_;
     std::chrono::milliseconds interval_;
+    bool enabled_;
```
ovidiusm: IIUC the timer executes in a thread pool, which means that this variable needs to be protected, since we write it from ~nixlTelemetry(). Either mutex or atomic

src/core/telemetry.cpp
```diff
@@ -130,24 +134,27 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
+            }
 
             // Schedule next execution with adjusted interval
             auto next_interval = std::chrono::milliseconds(task.interval_) - execution_time;
             if (next_interval.count() < 0) {
                 next_interval = std::chrono::milliseconds(0);
             }
-
-            // Schedule the next operation
             task.interval_ = next_interval;
```
ovidiusm: IIUC this keeps shrinking at each execution, is it correct?

src/core/telemetry.cpp
```diff
@@ -46,14 +46,17 @@ nixlTelemetry::nixlTelemetry(const std::string &name, backend_map_t &backend_map
 }
 
 nixlTelemetry::~nixlTelemetry() {
+    writeTask_.enabled_ = false;
```
ovidiusm, line 49: Do I understand correctly that the goal here is to wait for the timer to finish executing and stop, before destroying the telemetry object?

src/core/telemetry.h
rafiw: only main thread writes to it, the pool only reads

src/core/telemetry.cpp
rafiw: no it makes sure next iteration in X millis from the **end** of operation since operation it self takes time.

src/core/telemetry.h
ovidiusm: Then it needs synchronization

src/core/telemetry.cpp
rafiw: a few line below a call to write events is done here to drain last events. I don't want to add a lock to that function since besides here it's always called from the pool thread.

src/core/telemetry.cpp
ovidiusm: Let's say that we configure the interval at 100 ms

Let's say that execution_time is 10 ms

Then next_interval is 90 ms

Then we write task.interval_ as 90 ms

Next run it will become 80 ms then 70 ms etc.

src/core/telemetry.cpp
ovidiusm: You can make it `std::atomic<bool>`, we don't care about perf impact on timer thread, and main thread does not touch it except on destruction, so there is no perf impact on main thread

ovidiusm (approved):

src/core/telemetry.cpp
```diff
@@ -128,25 +132,16 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
-            if (!task.callback_ || !task.callback_()) {
-                // if return false, stop the task
+            if (!task.enabled_) {
                 return;
             }
 
-            auto end_time = std::chrono::steady_clock::now();
-            auto execution_time =
-                std::chrono::duration_cast<std::chrono::milliseconds>(end_time - start_time);
+            if (task.callback_) task.callback_();
```
w1ldptr: Check that `callback_` pointer once before scheduling the timer instead of on ever tick (perhaps with assert).

src/core/telemetry.cpp
```diff
@@ -128,25 +132,16 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
     task.timer_.expires_after(task.interval_);
     task.timer_.async_wait([this, &task](const asio::error_code &ec) {
         if (ec != asio::error::operation_aborted) {
-            auto start_time = std::chrono::steady_clock::now();
-
-            if (!task.callback_ || !task.callback_()) {
-                // if return false, stop the task
+            if (!task.enabled_) {
```
w1ldptr: When the timer is canceled `ec` value is `asio::error::operation_aborted`, so this code block won't be executed anyway making this check before calling `task.callback_()` redundant. Consider removing this conditional (leave second one after `task.callback_()` as-is to handle concurrent cancellation when this callback is executing).

src/core/telemetry.cpp
rafiw: I will remove the if entirely since adding it once is not possible

ovidiusm (approved):

PR #755, opened 2025-09-02
BINDINGS: Package the OBJ plugin

ovidiusm:
## What?
Adds the OBJ plugin to the python wheel.

## Why?
So far it was not packaged because of missing dependencies in the manylinux image.

### DIFF
--- a/.github/workflows/blossom-ci.yml
+++ b/.github/workflows/blossom-ci.yml
@@ -3,6 +3,11 @@ name: Blossom-CI
 on:
   issue_comment:
     types: [created]
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - 'release/*'
+      - main
   workflow_dispatch:
       inputs:
           platform:
--- a/ATTRIBUTIONS-Rust.md
+++ b/ATTRIBUTIONS-Rust.md
@@ -10475,3 +10475,1042 @@ END OF TERMS AND CONDITIONS
 
 ```
 
+## errno - 0.3.13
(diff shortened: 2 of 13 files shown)


src/plugins/obj/meson.build
```diff
@@ -22,10 +22,12 @@ obj_sources = [
 ]
 
 aws_s3 = dependency('aws-cpp-sdk-s3', static: false, required: false)
-if aws_s3.found()
+aws_core = dependency('aws-cpp-sdk-core', required: false, static: false)
```
aranadive, line 25: Why is aws_core required now as a dependency?

src/plugins/obj/meson.build
```diff
@@ -22,10 +22,12 @@ obj_sources = [
...
+if aws_s3.found() and aws_core.found()
     # By default aws-cpp-sdk sets c++11 compile flag for the whole project
     partial_aws_s3 = aws_s3.partial_dependency(compile_args: false, includes: true, link_args: true, links: true)
-    plugin_deps += [partial_aws_s3]
+    partial_aws_core = aws_core.partial_dependency(compile_args: false, includes: true, link_args: true, links: true)
+    plugin_deps += [partial_aws_s3, partial_aws_core]
 else
     subdir_done()
```
aranadive, line 33: Lets add a message here about missing deps for Obj plugin.

src/plugins/obj/meson.build
ovidiusm: S3 links against core, and on RHEL there is a linking issue. On Ubuntu this has no effect.

aranadive (approved):

PR #761, opened 2025-09-02
Removed the macros in nixl_agent.

mkhazraee:
## What?
Removed the macros in nixl_agent.

## Why?
Better to avoid such macros in general.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1394,21 +1394,22 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
 
     nixlSerDes sd;
     ret = sd.addStr("Agent", data->name);
-    if (ret) return ret; // serdes class logs errors if necessary
+    // Always returns SUCCESS, serdes class logs errors if necessary
+    if (ret) return NIXL_ERR_UNKNOWN;
 
     ret = sd.addBuf("Conns", &conn_cnt, sizeof(conn_cnt));
-    if (ret) return ret; // serdes class logs errors if necessary
+    if (ret) return NIXL_ERR_UNKNOWN;
 
     for (auto &c : data->connMD) {
         nixl_backend = c.first;
         ret = sd.addStr("t", nixl_backend);
-        if (ret) return ret; // serdes class logs errors if necessary
+        if (ret) return NIXL_ERR_UNKNOWN;
         ret = sd.addStr("c", c.second);
-        if (ret) return ret; // serdes class logs errors if necessary
+        if (ret) return NIXL_ERR_UNKNOWN;
     }
 
     ret = sd.addStr("", "MemSection");
-    if (ret) return ret; // serdes class logs errors if necessary
+    if (ret) return NIXL_ERR_UNKNOWN;
 
     ret = data->memorySection->serialize(&sd);
     if (ret) {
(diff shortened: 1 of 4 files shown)


src/core/nixl_agent.cpp
```diff
@@ -1476,19 +1464,19 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
 
     nixlSerDes sd;
```
aranadive, line 1472: Maybe just a refactor of the serialize stuff would make this less messy?

mkhazraee: Hmm, refactor in what way? There is this other PR
https://github.com/ai-dynamo/nixl/pull/663
But it's not doing such refactoring.

brminich: @ColinNV already suggested nice refactoring, but it was not accepted, because it changes serdes API
in #592

src/utils/common/nixl_log.h
```diff
@@ -48,6 +48,11 @@
  */
 #define NIXL_PERROR NIXL_ERROR.WithPerror()
 
+/*
+ * Like NIXL_ERROR, but prefixed with current function name and a colon
+ */
+#define NIXL_ERROR_FUNC NIXL_ERROR << __FUNCTION__ << ": "
```
brminich, line 54: do we really need func name? Maybe always print it then to  have consistent log formatting?

src/core/nixl_agent.cpp
brminich: i'd keep this PR small as it addresses a certain comment to the previous PR

src/utils/common/nixl_log.h
mkhazraee: At first I was thinking of doing that, but didn't want to break the current code, and maybe there is a case that they don't want it. 

Also since we have that PERR one, it would be easier with a base one to be extended in both modes. (otherwise I think we need to do "do ... while" or so to use abseil log and add __ FUCNTION __ to it). Still we can rename and there might be another way.

src/core/nixl_agent.cpp
mkhazraee: Agreed

aranadive: My suggestion was to just move the serializing parts to a diff function to make it more readable without changing API.

aranadive: If it makes sense to do it in a different PR then sure.

src/core/nixl_agent.cpp
```diff
@@ -1476,19 +1464,19 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
...
     ret = sd.addStr("Agent", data->name);
-    if (ret) SERIALIZE_ERR(ret);
+    if (ret) return ret; // serdes class logs errors if necessary
```
yosefe: IMO no need to comment // serdes class logs errors if necessary in every line ..

src/core/nixl_agent.cpp
mkhazraee: Following up with Adit, I looked into the code, about 10 lines are repeated between getLocalMD and getPartialLocalMD, and they have some different parameters, not worth refactoring.

mkhazraee:
@brminich @yosefe @aranadive Moving the logging only to serdes meant the original calling method is not there anymore and hence not consistent. So I added the agent level logging, with some code clean up to reduce the code repeat (especially on serializer most paths just return success in serdes). I was thinking user doesn't need to know details of serdes failure, like what part of the blob didn't work, but that's so rare that I think it's fine.

src/core/nixl_agent.cpp
```diff
@@ -1474,33 +1464,38 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
         selected_engines.insert(backend);
     }
 
+    if (selected_engines.size() == 0 && descs.descCount() > 0) {
```
mkhazraee, line 1467: Just moved it up, as no point in serializing if this check doesn't pass.

PR #764, opened 2025-09-03
Add a flag to agent config to captureTelemetry

mkhazraee:
## What?
Add a flag to agent config to captureTelemetry

## Why?
Enabling telemetry capture through both config and env vars. For output file, the env var is still needed as it's more runtime, but just enabling to capture the telemetry data to be used by getXferTelemetry can be set in agent config.

### DIFF
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -70,8 +70,8 @@ class nixlAgentConfig {
          * @param num_workers        Optional number of shared workers per backend
          * @param pthr_delay_us      Optional delay for pthread in us
          * @param lthr_delay_us      Optional delay for listener thread in us
-         * @param etcd_watch_timeout Optional timeout for etcd watch operations in microseconds
          * @param capture_telemetry  Optional flag to enable telemetry capture
+         * @param etcd_watch_timeout Optional timeout for etcd watch operations in microseconds
          */
         nixlAgentConfig(const bool use_prog_thread,
                         const bool use_listen_thread = false,
@@ -80,9 +80,9 @@ class nixlAgentConfig {
                         unsigned int num_workers = 1,
                         const uint64_t pthr_delay_us = 0,
                         const uint64_t lthr_delay_us = 100000,
+                        const bool capture_telemetry = false,
                         const std::chrono::microseconds &etcd_watch_timeout =
-                            std::chrono::microseconds(5000000),
-                        const bool capture_telemetry = false)
+                            std::chrono::microseconds(5000000))
             : useProgThread(use_prog_thread),
               useListenThread(use_listen_thread),
               listenPort(port),


src/api/python/_api.py
```diff
@@ -137,6 +137,7 @@ def __init__(
         enable_prog_thread: bool = True,
         enable_listen_thread: bool = False,
         listen_port: int = 0,
+        capture_telemetry: bool = False,
```
mkhazraee, line 141: Didn't put it at the end, since still agent_config, the other two are backend related.

test/nixl/nixl_test.cpp
```diff
@@ -204,7 +204,7 @@ static void initiatorThread(nixlAgent &agent, nixl_opt_args_t *extra_params,
 }
 
 static void runTarget(const std::string &ip, int port, nixl_thread_sync_t sync_mode) {
-    nixlAgentConfig cfg(true, true, port, sync_mode);
+    nixlAgentConfig cfg(true, true, port, sync_mode, false);
```
mkhazraee: Keeping it in one of the main tests as example.

src/api/cpp/nixl_params.h
```diff
@@ -37,6 +37,8 @@ class nixlAgentConfig {
         int      listenPort;
         /** @var synchronization mode for multi-threaded environment execution */
         nixl_thread_sync_t syncMode;
+        /** @var Capture telemetry info regardless of environment variables*/
+        bool captureTelemetry;
```
mkhazraee, line 41: @yosefe @brminich What do you think of the name? Meaning it can only enable the capture part, but irrelevant for file output which is through env var only.

mkhazraee:
> Clean api! Should this flag also appear in https://github.com/ai-dynamo/nixl/blob/main/examples/python/nixl_api_example.py for reference?

Agreed, I haven't got to python tests, first finalizing the name and so. The bindings and so should be tested.

src/api/cpp/nixl_params.h
```diff
@@ -63,19 +65,21 @@ class nixlAgentConfig {
...
-                         useListenThread(use_listen_thread),
-                         listenPort(port),
-                         syncMode(sync_mode),
-                         pthrDelay(pthr_delay_us),
-                         lthrDelay(lthr_delay_us) { }
+        nixlAgentConfig(const bool use_prog_thread,
+                        const bool use_listen_thread = false,
+                        const int port = 0,
+                        nixl_thread_sync_t sync_mode = nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
+                        const bool capture_telemetry = false,
```
ovidiusm: Given that this is an API change, safer to add this at the end since there may be clients using the other optional arguments

src/api/cpp/nixl_params.h
```diff
@@ -63,19 +65,21 @@ class nixlAgentConfig {
...
+                        unsigned int num_workers = 1,
```
ovidiusm, line 80: We also need to update the doc comment

src/core/nixl_agent.cpp
```diff
@@ -154,14 +154,21 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
             } else {
                 NIXL_DEBUG << "NIXL telemetry is enabled without an output file";
             }
+        } else if (cfg.captureTelemetry) {
```
ovidiusm, line 142: So this takes precedence only if true? Would it make sense to always take precedence?

src/api/cpp/nixl_params.h
mkhazraee: fair, I missed those as in examples the lock was the latest. Will move.

src/api/cpp/nixl_params.h
mkhazraee: you mean the md file?

ovidiusm: No, I mean the comment above:

```
/**
 * @brief  Agent configuration constructor for enabling various features.
 * @param use_prog_thread    flag to determine use of progress thread
 * @param use_listen_thread  flag to determine use of listener thread
 * @param port               specify port for listener thread to listen on
 * @param sync_mode          Thread synchronization mode
 * @param lthr_delay_us      Optional delay for listener thread in us
 */
```

src/core/nixl_agent.cpp
mkhazraee: It's more of an "either" scenario. So we separated telemetry capture from file writer, so there are 3 modes: no capture, capture only (can be read using api), capture + file output. 

Now considering the 3 desired modes, based on agent config flag and env vars:
- no capture: both are disabled (env var not set, or set to n, or set to an unknown value)
- capture only: If either of them is set, that means we want the capture.
- capture and file: we only support this in env var mode for now, because fixing a file path during compile time is not a good practice.

src/api/cpp/nixl_params.h
mkhazraee: honestly the booleans being first, or flags being next to their corresponding values would have been much better, but for now keeping it backward compatible until we do a refactor on this class, using some overloading or so to simplify things. Right now if someone wants to set the capture telemetry, they have to set some non-intuitive values which is not the best.

mkhazraee: Wait, now I remember, I did this based on nixl_bindings, will update that one too ...

ovidiusm: Let's please move it to the end, let's not introduce a breaking change.
We can also add a setter so that the user does not have to set everything:

cfg = nixlAgentConfig().setCaptureTelemetry(enable)...;

Or builder pattern

mkhazraee: Yeah something for future, for now moved it to the end.

src/api/cpp/nixl_params.h
```diff
@@ -53,29 +55,32 @@ class nixlAgentConfig {
...
-
         /**
          * @brief  Agent configuration constructor for enabling various features.
          * @param use_prog_thread    flag to determine use of progress thread
-         * @param use_listen_thread  flag to determine use of listener thread
-         * @param port               specify port for listener thread to listen on
-         * @param sync_mode          Thread synchronization mode
+         * @param use_listen_thread  Optional flag to determine use of listener thread
+         * @param port               Optional port for listener thread to listen on
+         * @param sync_mode          OptionalThread synchronization mode
```
ovidiusm:
```suggestion
         * @param sync_mode          Optional Thread synchronization mode
```

ovidiusm (approved):

aranadive (dismissed):

src/api/cpp/nixl_params.h
```diff
@@ -53,29 +55,32 @@ class nixlAgentConfig {
...
+                        nixl_thread_sync_t sync_mode = nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
+                        unsigned int num_workers = 1,
+                        const uint64_t pthr_delay_us = 0,
+                        const uint64_t lthr_delay_us = 100000,
+                        const bool capture_telemetry = false)
+            : useProgThread(use_prog_thread),
+              useListenThread(use_listen_thread),
+              listenPort(port),
+              syncMode(sync_mode),
+              captureTelemetry(capture_telemetry),
```
brminich, line 90: pls move to the end

mkhazraee: Causes compile error, as it's private, the later ones are public. Not sure why the later ones are public though!

ovidiusm (approved):

PR #766, opened 2025-09-03
Make ETCD watch timeout configurable

michal-shalev:
## What?
Add configurable etcd watch timeout parameter to `nixlAgentConfig`, replacing hardcoded 5-second timeout.

## Why?
Users are experiencing etcd watch timeouts, causing metadata fetch failures. The hardcoded 5-second timeout is insufficient for high-latency networks or under heavy load conditions. This change allows users to tune the timeout based on their specific network conditions and etcd performance characteristics.

## How?
- Added `etcdWatchTimeout` member to `nixlAgentConfig` with default value of 5 seconds
- Updated `nixlEtcdClient` constructor to accept timeout parameter
- Modified `waitForMetadataFromEtcd` to use configurable timeout instead of hardcoded value
- Updated Python API documentation to include new parameter
- Maintains backward compatibility through default value

### DIFF
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -57,7 +57,7 @@ class nixlAgentConfig {
          * @var ETCD watch timeout in seconds
          *      Timeout for waiting for metadata changes when watching etcd keys.
          */
-        std::chrono::seconds etcdWatchTimeoutSec;
+        std::chrono::seconds etcdWatchTimeout;
 
         /**
          * @brief  Agent configuration constructor for enabling various features.
@@ -67,7 +67,7 @@ class nixlAgentConfig {
          * @param sync_mode          Thread synchronization mode
          * @param pthr_delay_us      Optional delay for pthread in us
          * @param lthr_delay_us      Optional delay for listener thread in us
-         * @param etcd_watch_timeout_sec Optional timeout for etcd watch operations in seconds
+         * @param etcd_watch_timeout Optional timeout for etcd watch operations in seconds
          */
         nixlAgentConfig(const bool use_prog_thread,
                         const bool use_listen_thread = false,
@@ -76,14 +76,14 @@ class nixlAgentConfig {
                         unsigned int num_workers = 1,
                         const uint64_t pthr_delay_us = 0,
                         const uint64_t lthr_delay_us = 100000,
-                        const std::chrono::seconds etcd_watch_timeout_sec = std::chrono::seconds(5))
+                        const std::chrono::seconds etcd_watch_timeout = std::chrono::seconds(5))
             : useProgThread(use_prog_thread),
               useListenThread(use_listen_thread),
               listenPort(port),
               syncMode(sync_mode),
               pthrDelay(pthr_delay_us),
               lthrDelay(lthr_delay_us),
-              etcdWatchTimeoutSec(etcd_watch_timeout_sec) {}
+              etcdWatchTimeout(etcd_watch_timeout) {}
 
         /**
          * @brief Copy constructor for nixlAgentConfig object
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -57,7 +57,7 @@ class nixlAgentConfig {
          * @var ETCD watch timeout in seconds
          *      Timeout for waiting for metadata changes when watching etcd keys.
          */
-        uint64_t etcdWatchTimeoutSec;
+        std::chrono::seconds etcdWatchTimeoutSec;
 
         /**
          * @brief  Agent configuration constructor for enabling various features.
--- a/src/api/cpp/nixl_params.h
+++ b/src/api/cpp/nixl_params.h
@@ -59,7 +59,6 @@ class nixlAgentConfig {
          */
         uint64_t etcdWatchTimeoutSec;
 
-
         /**
          * @brief  Agent configuration constructor for enabling various features.
          * @param use_prog_thread    flag to determine use of progress thread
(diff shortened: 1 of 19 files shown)


src/api/cpp/nixl_params.h
```diff
@@ -53,6 +53,11 @@ class nixlAgentConfig {
          */
         uint64_t lthrDelay;
 
+        /**
+         * @var ETCD watch timeout in seconds
+         *      Timeout for waiting for metadata changes when watching etcd keys.
+         */
+        uint64_t etcdWatchTimeoutSec;
```
yosefe:
```suggestion
       std::chrono::seconds etcdWatchTimeout;
```
since there are already APIs that use std::chrono such as estimateXferCost

michal-shalev: done

src/api/cpp/nixl_params.h
```diff
@@ -53,6 +53,11 @@ class nixlAgentConfig {
          */
         uint64_t lthrDelay;
 
+        /**
+         * @var ETCD watch timeout in seconds
+         *      Timeout for waiting for metadata changes when watching etcd keys.
+         */
+        std::chrono::seconds etcdWatchTimeoutSec;
```
yosefe: can remove the suffix Sec

src/api/cpp/nixl_params.h
```diff
@@ -62,20 +67,23 @@ class nixlAgentConfig {
...
-                         pthrDelay(pthr_delay_us),
-                         lthrDelay(lthr_delay_us) { }
+        nixlAgentConfig(const bool use_prog_thread,
+                        const bool use_listen_thread = false,
+                        const int port = 0,
+                        nixl_thread_sync_t sync_mode = nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
+                        unsigned int num_workers = 1,
+                        const uint64_t pthr_delay_us = 0,
+                        const uint64_t lthr_delay_us = 100000,
+                        const std::chrono::seconds etcd_watch_timeout_sec = std::chrono::seconds(5))
```
yosefe: etcd_watch_timeout_sec -> etcd_watch_timeout
since the type already contains the time unit

src/api/cpp/nixl_params.h
michal-shalev: done

src/api/cpp/nixl_params.h
michal-shalev: done

src/core/nixl_listener.cpp
```diff
@@ -197,7 +198,9 @@ class nixlEtcdClient {
     }
 
 public:
-    nixlEtcdClient(const std::string& my_agent_name) {
+    nixlEtcdClient(const std::string &my_agent_name,
+                   std::chrono::seconds timeout_sec = std::chrono::seconds(5))
+        : watch_timeout_sec(timeout_sec) {
```
yosefe: watch_timeout_sec -> watch_timeout

src/api/cpp/nixl_params.h
```diff
@@ -62,20 +67,23 @@ class nixlAgentConfig {
...
-                         pthrDelay(pthr_delay_us),
-                         lthrDelay(lthr_delay_us) { }
+        nixlAgentConfig(const bool use_prog_thread,
+                        const bool use_listen_thread = false,
+                        const int port = 0,
+                        nixl_thread_sync_t sync_mode = nixl_thread_sync_t::NIXL_THREAD_SYNC_DEFAULT,
+                        unsigned int num_workers = 1,
+                        const uint64_t pthr_delay_us = 0,
+                        const uint64_t lthr_delay_us = 100000,
+                        const std::chrono::seconds etcd_watch_timeout = std::chrono::seconds(5))
```
rakhmets:
```suggestion
                        const std::chrono::seconds &etcd_watch_timeout = std::chrono::seconds(5))
```

src/core/nixl_listener.cpp
```diff
@@ -197,7 +198,9 @@ class nixlEtcdClient {
     }
 
 public:
-    nixlEtcdClient(const std::string& my_agent_name) {
+    nixlEtcdClient(const std::string &my_agent_name,
+                   std::chrono::seconds timeout_sec = std::chrono::seconds(5))
```
```suggestion
                   const std::chrono::seconds &timeout = std::chrono::seconds(5))
```

michal-shalev: done

src/api/cpp/nixl_params.h
michal-shalev: done

src/core/nixl_listener.cpp
michal-shalev: done

yosefe (approved):

src/api/cpp/nixl_params.h
rakhmets: Please push the changes fixing the issue.

michal-shalev: weird. I'm having a lot of issues with my VScode during the last few days. checking.

src/core/nixl_listener.cpp
```diff
@@ -187,6 +187,7 @@ class nixlEtcdClient {
     std::mutex invalidated_agents_mutex;
     std::unordered_map<std::string, std::unique_ptr<etcd::Watcher>,
                         std::hash<std::string>, strEqual> agentWatchers;
+    std::chrono::seconds watch_timeout;
```
```suggestion
    std::chrono::seconds watchTimeout_;
```
Сlass members must be in lower camel case according to the NIXL code style (docs/CodeStyle.md).

rakhmets (approved):

src/core/nixl_listener.cpp
michal-shalev: but the other class members above it aren't

rakhmets: Only one previously added class member in the correct style: `agentWatchers`.
I suggest following NIXL code style on the new code.
I believe we'll be adding clang-tidy soon to catch these issues automatically. The config has already been merged. (https://github.com/ai-dynamo/nixl/blob/main/.clang-tidy)

rakhmets: after looking at the config again, it should be `watchTimeout_`, since it is a private member.

src/api/python/_api.py
```diff
@@ -139,13 +140,15 @@ def __init__(
         listen_port: int = 0,
         num_threads: int = 0,
         backends: list[str] = ["UCX"],
+        etcd_watch_timeout: int = 5,
```
ovidiusm: Could we allow sub-second precision in the Python API? Either us or float?

aranadive: Any reason for this?

src/api/python/_api.py
```diff
@@ -139,13 +140,15 @@ def __init__(
         listen_port: int = 0,
         num_threads: int = 0,
         backends: list[str] = ["UCX"],
+        etcd_watch_timeout: float = 5000000.0,
```
ovidiusm: It should be int here since we use microseconds.

Earlier I meant that either use microseconds, either use float (seconds).

src/api/python/_api.py
```diff
@@ -186,6 +189,12 @@ def __init__(
             nixl_conf.enable_listen,
             nixl_conf.port,
             thread_config,
+            nixl_conf.num_threads,
+            0,  # pthr_delay_us
+            100000,  # lthr_delay_us
+            int(
```
ovidiusm: Let's use int, no need to cast then

michal-shalev: done

src/api/python/_api.py
michal-shalev: done

ovidiusm (approved):

yosefe:
Python tests failures seems related to change in _api.py

yosefe (approved):

ovidiusm:
Why was the change not exposed in the Python API?

michal-shalev:
> Why was the change not exposed in the Python API?

The Python API exposure will be addressed in a separate PR. Following the review yesterday, we decided to keep only the C++ changes because of concerns with the time units in the Python API part.

PR #767, opened 2025-09-03
Implement user metadata support

mamar123:
## What?
This PR introduces support for user metadata (userMD) in Nixl.

## Why?
User metadata support enables attaching custom key–value data to agents and retrieving it across nodes. This is required to extend Nixl’s flexibility in handling agent-specific information and to support upcoming features that depend on user-defined metadata.

## How?
- Adds userMD storage to nixlAgentData for both local and remote metadata.
- Implements addLocalUserMD() to store key–value pairs locally.
- Implements getRemoteUserMD() to retrieve user metadata from remote agents.
- Extends metadata serialization so getLocalMD() includes user metadata.
- Adds filtering support (userMdKeys) to control selective user md retrieval in getLocalPartialMD().
- Provides comprehensive unit tests covering user metadata functionality.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -24,6 +24,29 @@
 #include "plugin_manager.h"
 #include "common/nixl_log.h"
 
+namespace {
+nixl_status_t
+serializeUserMD(nixlSerDes &sd, const std::unordered_map<std::string, std::string> &user_md) {
+    size_t user_md_cnt = user_md.size();
+    nixl_status_t ret;
+
+    ret = sd.addBuf("UserMDs", &user_md_cnt, sizeof(user_md_cnt));
+    if(ret)
+        return ret;
+
+    for (const auto &umd : user_md) {
+        ret = sd.addStr("k", umd.first);
+        if(ret)
+            return ret;
+        ret = sd.addStr("v", umd.second);
+        if(ret)
+            return ret;
+    }
+
+    return NIXL_SUCCESS;
+}
+} // namespace
+
 static const std::vector<std::vector<std::string>> illegal_plugin_combinations = {
     {"GDS", "GDS_MT"},
 };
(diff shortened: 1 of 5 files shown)


src/core/nixl_agent.cpp
```diff
@@ -1261,6 +1285,33 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
     if(ret)
         return ret;
 
+    // Add userMD if requested via extra_params->userMdKeys
+    if (extra_params && extra_params->userMdKeys.has_value()) {
```
etoledano: Is `has_value()` needed? Won't `extra_params->userMdKeys` be `false` if there is no value and `true` otherwise?

src/core/nixl_agent.cpp
```diff
@@ -1261,6 +1285,33 @@ nixlAgent::getLocalPartialMD(const nixl_reg_dlist_t &descs,
...
+            return ret;
+
+        for (const auto &it : found_user_mds) {
+            ret = sd.addStr("k", it->first);
+            if(ret)
+                return ret;
+            ret = sd.addStr("v", it->second);
+            if(ret)
+                return ret;
+        }
```
etoledano: Code duplication? (`user_md` -> `it`)

src/core/nixl_agent.cpp
```diff
@@ -1458,6 +1533,20 @@ nixlAgent::fetchRemoteMD (const std::string remote_name,
...
+nixlAgent::getRemoteUserMD(const std::string &remote_agent,
+                           const std::string &key,
+                           std::string &value) {
+    NIXL_LOCK_GUARD(data->lock);
+    if (data->remoteUserMD.count(remote_agent) != 0) {
+        if (data->remoteUserMD[remote_agent].count(key) != 0) {
+            value = data->remoteUserMD[remote_agent][key];
+            return NIXL_SUCCESS;
+        }
+    }
```
etoledano:
```
auto it = data->remoteUserMD.find(key);
if (it == data->remoteUserMD.end()) {
  return NIXL_ERR_NOT_FOUND;
}

return it->second;
```

brminich:
can you pls give an example of the feature requiring custom metadata?

PR #781, opened 2025-09-08
Reproduce CI issue with telemetry

mkhazraee:
## What?
Verify CI issue with telemetry is gone

## Why?
Previously for short python tests the telemetry didn't generate the output, checking if it was fixed.


brminich:
@mkhazraee can we close?

ovidiusm:
Let's close for tidyness

PR #789, opened 2025-09-15
Implement host device API

michal-shalev:
## What?
Implements host side device API for UCX backend enabling direct GPU-initiated network transfers via CUDA device functions.

## Why?
Allows GPU kernels to initiate network transfers without CPU involvement, reducing latency and eliminating CPU-GPU synchronization overhead for AI inference workloads.

## How?
- Implemented `createGpuXferReq` and `releaseGpuXferReq`
- Updated `getGpuSignalSize` and `prepGpuSignal` to get the backend from `extra_params`
- Build fixes
- Added GTEST for single API

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1666,7 +1666,8 @@ nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
     }
 
     try {
-        auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(*ep, local_mems, remote_rkeys);
+        auto device_mem_list =
+            std::make_unique<nixl::ucx::deviceMemList>(*ep, local_mems, remote_rkeys);
         gpu_req_hndl = device_mem_list->get();
         device_mem_list.release();
         return NIXL_SUCCESS;


meson.build
```diff
@@ -140,11 +140,11 @@ if ucx_path != ''
   # Check if path is absolute
   if ucx_inc_path.startswith('/')
     ucx_dep = declare_dependency(
-      link_args : ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
+      link_args : ['-L' + ucx_lib_path, '-Wl,-rpath,' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
       compile_args : ['-I' + ucx_inc_path])
   else
     ucx_dep = declare_dependency(
-      link_args : ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
+      link_args : ['-L' + ucx_lib_path, '-Wl,-rpath,' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
```
yosefe: why needed?

meson.build
```diff
@@ -156,23 +156,27 @@ nvcc_prog = find_program('nvcc', required: false)
...
-    else
-        message('UCX GPU device API support not available - GPU device API features will be disabled')
     endif
+
+    summary({
+    'UCX GPU Device API' : ucx_gpu_device_api_available,
+    'GPU-side compile'   : have_gpu_side,
+    'Host-side compile'  : have_host_side,
+    'nvcc available'     : nvcc_prog.found(),
+  }, section: 'UCX GPU Device API', bool_yn: true)
```
yosefe, line 179: maybe remove for now since we don't have summary for other parts of the build process

src/api/cpp/nixl.h
```diff
@@ -344,12 +344,13 @@ class nixlAgent {
...
          *
-         * @param  backend       [in]  Backend handle to get the size of the GPU signal
          * @param  signal_size   [out] Size required for the GPU signal
+         * @param  extra_params  Extra parameters used in getting the size of the GPU signal.
+         *                       The backend must be specified in extra_params.
          * @return nixl_status_t Error code if call was not successful
          */
         nixl_status_t
-        getGpuSignalSize(const nixlBackendH &backend, size_t &signal_size) const;
+        getGpuSignalSize(size_t &signal_size, const nixl_opt_args_t *extra_params) const;
```
yosefe, line 353: extra_params = nullptr

src/api/cpp/nixl.h
```diff
@@ -359,10 +360,13 @@ class nixlAgent {
...
          *
          * @param  signal_descs  [in] Registered descriptor list for the signal memory
+         * @param  extra_params  Extra parameters used in preparing the GPU signal.
+         *                       The backend must be specified in extra_params.
          * @return nixl_status_t Error code if call was not successful
          */
         nixl_status_t
-        prepGpuSignal(const nixl_reg_dlist_t &signal_descs) const;
+        prepGpuSignal(const nixl_reg_dlist_t &signal_descs,
+                      const nixl_opt_args_t *extra_params) const;
```
yosefe, line 369: extra_params = nullptr

src/core/nixl_agent.cpp
```diff
@@ -1252,47 +1252,53 @@ nixlAgent::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
...
 nixl_status_t
-nixlAgent::prepGpuSignal(const nixl_reg_dlist_t &signal_descs) const {
+nixlAgent::prepGpuSignal(const nixl_reg_dlist_t &signal_descs,
+                         const nixl_opt_args_t *extra_params) const {
     if (signal_descs.descCount() == 0) {
         NIXL_ERROR_FUNC << "signal descriptor list is empty";
         return NIXL_ERR_INVALID_PARAM;
     }
 
+    if (!extra_params || extra_params->backends.empty()) {
```
yosefe, line 1274: check if extra_params is null and if yes take the first available backend than implements prepGpuSignal

src/api/cpp/nixl.h
rakhmets, line 353: I'm not sure that it should be `const nixl_opt_args_t *extra_params`. I don't see an example of using this API when this parameter is `nullptr`. This is not an extra parameter, but the main and only one.
I suggest `const nixl_opt_args_t &params`. And you can get rid of checking the parameter for being `nullptr`.

The same for `prepGpuSignal`.

src/core/nixl_agent.cpp
```diff
@@ -1252,47 +1252,53 @@ nixlAgent::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
 }
 
 nixl_status_t
-nixlAgent::getGpuSignalSize(const nixlBackendH &backend, size_t &signal_size) const {
+nixlAgent::getGpuSignalSize(size_t &signal_size, const nixl_opt_args_t *extra_params) const {
     NIXL_SHARED_LOCK_GUARD(data->lock);
-    return backend.engine->getGpuSignalSize(signal_size);
+    return extra_params->backends[0]->engine->getGpuSignalSize(signal_size);
```
rakhmets, line 1263: Please check that `backends` is not empty.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1649,112 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
 }
 
 nixl_status_t
-nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &handle,
+nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
                                 nixlGpuXferReqH &gpu_req_hndl) const {
+#ifdef HAVE_UCX_GPU_DEVICE_API
+    auto *intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
```
```suggestion
    auto intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1649,112 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    size_t workerId = intHandle->getWorkerId();
+
+    if (remoteConnMap.empty()) {
+        NIXL_ERROR << "No remote connections found";
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    std::string remote_agent = remoteConnMap.begin()->second->getRemoteAgent();
+    auto search = remoteConnMap.find(remote_agent);
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1649,112 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
 }
 
 nixl_status_t
-nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &handle,
+nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
                                 nixlGpuXferReqH &gpu_req_hndl) const {
+#ifdef HAVE_UCX_GPU_DEVICE_API
+    auto *intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
+    size_t workerId = intHandle->getWorkerId();
```
rakhmets: Please place declaration of the variables to a place before the first use.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1649,112 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+
+    std::vector<nixlUcxDeviceMemElem> elements;
+    elements.reserve(localDescs->descCount());
+
+    for (size_t i = 0; i < static_cast<size_t>(localDescs->descCount()); i++) {
+        const auto &localDesc = (*localDescs)[i];
+        const auto &remoteDesc = (*remoteDescs)[i];
+
+        nixlUcxPrivateMetadata *localMd = (nixlUcxPrivateMetadata *)localDesc.metadataP;
+        nixlUcxPublicMetadata *remoteMd = (nixlUcxPublicMetadata *)remoteDesc.metadataP;
```
rakhmets: static cast

src/utils/ucx/ucx_utils.h
```diff
@@ -84,6 +84,9 @@ class rkey;
 }
 class nixlUcxMem;
 
+typedef void *nixlUcxDeviceMemListH;
```
rakhmets: using

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1649,112 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+
+        nixlUcxDeviceMemElem elem;
+        elem.mem = localMd->mem;
+        elem.rkey = &remoteMd->getRkey(workerId);
+        elem.local_addr = (void *)localDesc.addr;
+        elem.remote_addr = remoteDesc.addr;
+        elem.length = localDesc.len;
+
+        elements.push_back(elem);
```
```suggestion
        elements.emplace_back(localMd->mem, &remoteMd->getRkey(workerId),
                              static_cast<void *>(localDesc.addr), remoteDesc.addr, localDesc.len);
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -715,3 +719,72 @@ nixlUcxWorker::getEfd() const {
...
+
+    if (elements.empty()) {
+        throw std::invalid_argument("Empty memory list elements provided");
+    }
+
+    std::vector<ucp_device_mem_list_elem_t> ucp_elements;
+    ucp_elements.reserve(elements.size());
+
+    for (const auto &elem : elements) {
+        ucp_device_mem_list_elem_t ucp_elem = {};
```
rakhmets:
```suggestion
        ucp_device_mem_list_elem_t ucp_elem;
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -715,3 +719,72 @@ nixlUcxWorker::getEfd() const {
...
+        ucp_elem.field_mask =
+            UCP_DEVICE_MEM_LIST_ELEM_FIELD_MEMH | UCP_DEVICE_MEM_LIST_ELEM_FIELD_RKEY;
+        ucp_elem.memh = elem.mem.memh;
+        ucp_elem.rkey = elem.rkey->get();
+        ucp_elements.push_back(ucp_elem);
+    }
+
+    ucp_device_mem_list_params_t params = {};
```
rakhmets:
```suggestion
    ucp_device_mem_list_params_t params;
```

src/utils/ucx/ucx_utils.cpp
```diff
@@ -715,3 +719,72 @@ nixlUcxWorker::getEfd() const {
...
+    mem_list_handle = reinterpret_cast<nixlUcxDeviceMemListH>(ucx_handle);
+
+    NIXL_DEBUG << "Created device memory list handle with " << elements.size() << " elements";
+#else
+    throw std::runtime_error(std::string(ucxGpuDeviceApiUnsupported));
+#endif
+}
+
+void
+nixlUcxEp::releaseMemList(nixlUcxDeviceMemListH mem_list_handle) {
```
rakhmets: This shouldn't be a method of class `nixlUcxEp`.
Please add a separate class.

```C++
// src/utils/ucx/device_mem_list.h
class nixlUcxEp;

namespace nixl::ucx {
struct deviceMemElem {
    nixlUcxMem mem;
    const nixl::ucx::rkey *rkey;
    void *local_addr;
    uint64_t remote_addr;
    size_t length;
};

class deviceMemList {
public:
    deviceMemList() = delete;
    deviceMemList(const nixlUcxEp &, const std::vector<nixlUcxDeviceMemElem> &);
    explicit deviceMemList(const ucp_device_mem_list_handle_h) noexcept;

    [[nodiscard]] ucp_device_mem_list_handle_h
    get() const noexcept {
        return deviceMemList_.get();
    }

private:
    [[nodiscard]] static ucp_device_mem_list_handle_h
    createDeviceMemList(const nixlUcxEp &, const std::vector<nixlUcxDeviceMemElem> &);

    const std::unique_ptr<ucp_device_mem_list_handle, void (*)(ucp_device_mem_list_handle_h)> deviceMemList_;
};
} // namespace nixl::ucx

// src/utils/ucx/device_mem_list.cpp
namespace nixl::ucx {
deviceMemList::deviceMemList(const nixlUcxEp &ep, const std::vector<nixlUcxDeviceMemElem> &elements)
    : deviceMemList_{createDeviceMemList(ep, elements), &ucp_device_mem_list_release} {}

deviceMemList::deviceMemList(const ucp_device_mem_list_handle_h device_mem_list)
    : deviceMemList_{device_mem_list, &ucp_device_mem_list_release} {}

ucp_device_mem_list_handle_h
deviceMemList ::createDeviceMemList(const nixlUcxEp &ep, const std::vector<nixlUcxDeviceMemElem> &elements) {
    // the content of nixlUcxEp::createMemList
}
}

// src/plugins/ucx/ucx_backend.cpp

nixl_status_t
nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
                                nixlGpuXferReqH &gpu_req_hndl) const {
...
    try {
        auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(ep, elements);
        gpu_req_hndl = std::reinterpret_cast<nixlGpuXferReqH>(device_mem_list.get());
        device_mem_list.release();
        return NIXL_SUCCESS;
    }
...
}

void
nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
    const auto ucp_device_mem_list = std::reinterpret_cast<ucp_device_mem_list_handle_h>(gpu_req_hndl);
    const auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(ucp_device_mem_list);
}
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1554,6 +1578,8 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
 
     // TODO: assert that handle is empty/completed, as we can't post request before completion
 
+    int_handle->storeDescriptorLists(local, remote);
```
ofirfarjun7: Why we need it here?

src/api/cpp/nixl.h
michal-shalev: `nixl_opt_args_t` is always `extra_params` in `nixl.h`, so I used `extra_params` for consistency but I don't mind changing it. WDYT? @brminich @ofirfarjun7 @yosefe

src/core/nixl_agent.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/utils/ucx/ucx_utils.h
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/utils/ucx/ucx_utils.cpp
michal-shalev: done

src/utils/ucx/ucx_utils.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: It should be only in prepXfer, fixed.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -352,6 +352,9 @@ class nixlUcxBackendH : public nixlBackendReqH {
     };
     std::optional<Notif> notif;
 
+    std::unique_ptr<nixl_meta_dlist_t> storedLocalDescs_;
+    std::unique_ptr<nixl_meta_dlist_t> storedRemoteDescs_;
```
yosefe: IMO it's not needed since nixlXferReqH already saves the descriptors list. We can assign them from nixlAgent::createGpuXferReq to ucx backend's req_hndl.engine->createGpuXferReq as additional parameter

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1647,111 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+#ifdef HAVE_UCX_GPU_DEVICE_API
+    auto intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
+
+    if (remoteConnMap.empty()) {
+        NIXL_ERROR << "No remote connections found";
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    std::string remote_agent = remoteConnMap.begin()->second->getRemoteAgent();
+    auto search = remoteConnMap.find(remote_agent);
```
yosefe: seems we are always taking the first connection which is not the right thing, we need to get the ep from remote dlist metadata , see postXfer in ucx backend as example

src/utils/ucx/ucx_utils.h
```diff
@@ -165,6 +175,14 @@ class nixlUcxMem {
     friend class nixlUcxEp;
 };
 
+struct nixlUcxDeviceMemElem {
+    nixlUcxMem mem;
+    const nixl::ucx::rkey *rkey;
+    void *local_addr;
+    uint64_t remote_addr;
+    size_t length;
+};
```
yosefe: do we really need this structure, maybe just pass local,remote dlists to uct_utils layer?

src/utils/ucx/ucx_utils.h
```diff
@@ -152,6 +155,13 @@ class nixlUcxEp {
     getEp() const noexcept {
         return eph;
     }
+
+    void
+    createMemList(const std::vector<nixlUcxDeviceMemElem> &elements,
+                  nixlUcxDeviceMemListH &mem_list_handle);
+
+    void
+    releaseMemList(nixlUcxDeviceMemListH mem_list_handle);
```
yosefe: this does not need to be method of the ep. it could be static function or in nixlUcxContext. then would not need to calculate the ep during destroy.

test/gtest/device_api/utils.cu
```diff
@@ -0,0 +1,157 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "utils.cuh"
+
+const char *GetGpuXferLevelStr(nixl_gpu_level_t level) {
```
yosefe, line 11: can we move them to 
namespace gtest {
namespace gpu {
...
}
}

src/api/cpp/nixl.h
ofirfarjun7: I agree that `extra_params` is misleading but this is not the first function that use this term even thou it expect it to be specified.
`nixlAgent::queryMem` is a good example so I'm guess if we want to be consistence it should be `extra_param`, if not we should keep in mind to change it in places it is required.

Another option is to have a fallback plan but I'm not sure if we want to make things more complicated..

ofirfarjun7: Discussed with Yossi. We will add fallback in next PR so this param will become optional.
For now leave it as is and add TODO.

meson.build
michal-shalev: image

From recent experience, users often struggle for hours with UCX GPU Device API dependencies and then reach out for help. This summary helps users quickly diagnose what's missing instead of parsing verbose build logs. It should reduce those support requests, though I won't insist if you prefer consistency.

yosefe: ok

src/api/cpp/nixl.h
michal-shalev: The API documentation specifies that "The backend must be specified in extra_params", because we wanted to avoid searching for it in the function implementation. So we have 2 options:
1. Keep the current explicitness, and then `extra_params` cannot be nullptr for this function, the user must specify the backend.
2. Switch to implicit backend selection, change to `extra_params = nullptr`, search for the first available backend that supports GPU signals, and update the documentation accordingly.

In any case, it should be consistent with `prepGpuSignal` API.

src/plugins/ucx/ucx_backend.cpp
michal-shalev: The `initiatorDescs` and `targetDescs` fields in `nixlXferReqH` are private and only accessible to `nixlAgent` via friend class. In order to pass them as additional parameters to `createGpuXferReq` it would require breaking encapsulation by exposing private fields.
image

yosefe: See how postXfer is passing desc list from handle to backend: https://github.com/ai-dynamo/nixl/blob/6798da4af3d506a1d9b9a6debd8995f43172cc1a/src/core/nixl_agent.cpp#L1095

src/api/cpp/nixl.h
ofirfarjun7: Option 2 in next PR.
For now just add TODO.

meson.build
michal-shalev: It ensures that executables can locate UCX shared libraries at runtime without requiring LD_LIBRARY_PATH to be set manually.
Not crucial for this PR. Reverted.

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

test/gtest/device_api/utils.cu
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: pushed a fix

src/utils/ucx/ucx_utils.cpp
michal-shalev: Added a separate class for deviceMemList

src/utils/ucx/ucx_utils.h
michal-shalev: It's now in a separate class

src/utils/ucx/ucx_utils.h
michal-shalev: It's now in a separate class, similar to `rkey.h`, a wrapper to `ucp_device_mem_list_handle_h`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -19,6 +19,9 @@
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
 #include "common/nixl_log.h"
+#ifdef HAVE_UCX_GPU_DEVICE_API
+#include "ucx/device_mem_list.h"
+#endif
```
yosefe: next pr: move these 3 lines to ucx_utils.h

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1412,7 +1415,10 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
 
     /* TODO: try to get from a pool first */
     size_t worker_id = getWorkerId();
-    handle = new nixlUcxBackendH(getWorker(worker_id).get(), worker_id);
+    auto *ucx_handle = new nixlUcxBackendH(getWorker(worker_id).get(), worker_id);
```
yosefe, line 1416: nextpr: can we remove the helper var ucx_handle?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    if (!remoteMd || !remoteMd->conn) {
+        NIXL_ERROR << "No connection found in remote metadata";
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    size_t workerId = intHandle->getWorkerId();
+    nixlUcxEp *ep = remoteMd->conn->getEp(workerId).get();
+
+    std::vector<nixl::ucx::deviceMemElem> elements;
+    elements.reserve(local_descs.descCount());
```
yosefe: next pr: try to get rid of deviceMemElem class, IMO it's redundant

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+
+        elements.emplace_back(nixl::ucx::deviceMemElem{&localMd->mem,
+                                                       &remoteDescMd->getRkey(workerId),
+                                                       reinterpret_cast<void *>(localDesc.addr),
+                                                       remoteDesc.addr,
+                                                       localDesc.len});
+    }
+
+    try {
+        auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(*ep, elements);
```
yosefe: next pr: pass  local_descs,remote_descs directly to deviceMemList constructor

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
-nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {}
+nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
+#ifdef HAVE_UCX_GPU_DEVICE_API
+    if (gpu_req_hndl == nullptr) {
+        NIXL_WARN << "Attempting to release null GPU transfer request handle";
+        return;
+    }
+
+    const auto ucp_device_mem_list = reinterpret_cast<ucp_device_mem_list_handle_h>(gpu_req_hndl);
+    const auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(ucp_device_mem_list);
```
yosefe: how does it release gpu_req_hndl?
maybe we need to pass CPU xfer req from above?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+#else
+    NIXL_WARN << "UCX GPU device API not supported";
```
yosefe: can remove the else part

src/plugins/ucx/ucx_backend.h
```diff
@@ -50,6 +50,11 @@ class nixlUcxConnection : public nixlBackendConnMD {
             return eps[ep_id];
         }
 
+        [[nodiscard]] const std::string &
+        getRemoteAgent() const noexcept {
```
yosefe: where is it used?

src/utils/ucx/ucx_utils.h
```diff
@@ -160,6 +160,11 @@ class nixlUcxMem {
     size_t size;
     ucp_mem_h memh;
 public:
+    [[nodiscard]] ucp_mem_h
+    getMemh() const noexcept {
+        return memh;
+    }
```
yosefe, line 166: where is it used?

test/gtest/device_api/single_write_test.cu
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    void
+    registerMem(nixlAgent &agent, const std::vector<MemBuffer> &buffers, nixl_mem_t mem_type) {
+        auto reg_list = makeDescList<nixlBlobDesc>(buffers, mem_type);
+        agent.registerMem(reg_list);
+    }
+
+    void
+    completeWireup(size_t from_agent, size_t to_agent) {
```
yosefe, line 198: nextpr: can we move this somehow to exportXferReqToGpu?

test/gtest/meson.build
```diff
@@ -66,8 +70,8 @@ gtest_sources = [
     ]
 
 if ucx_gpu_device_api_available
-    gtest_sources += 'device_api_test.cu'
-    device_api_inc = nixl_gpu_inc_dirs
+    gtest_sources += device_api_test_sources
```
yosefe, line 73: where is this defined?

test/gtest/meson.build
```diff
@@ -66,8 +70,8 @@ gtest_sources = [
...
+    device_api_inc = [nixl_gpu_inc_dirs, include_directories('device_api')]
```
yosefe, line 74: why do we need this part if already have device_api subdirectories?
can we move device_api_test.cu to device_api subfolder?

test/gtest/test_transfer.cpp
```diff
@@ -703,7 +703,8 @@ TEST_P(TestTransfer, PrepGpuSignal) {
     GTEST_SKIP() << "UCX GPU device API not available, skipping test";
 #else
     size_t gpu_signal_size = 0;
-    nixl_status_t size_status = getAgent(0).getGpuSignalSize(*backend_handles[0], gpu_signal_size);
+    nixl_opt_args_t extra_params = {.backends = {backend_handles[0]}};
```
yosefe, line 706: Need to move this test - `TEST_P(TestTransfer, PrepGpuSignal)` - to device_api_test.cu

src/utils/ucx/device_mem_list.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace nixl::ucx {
+class rkey;
+
+struct deviceMemElem {
+    const nixlUcxMem *mem;
+    const nixl::ucx::rkey *rkey;
+    void *local_addr;
+    uint64_t remote_addr;
+    size_t length;
```
rakhmets: These fields are unused.

meson.build
brminich, line 179: 👍

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
 }
 
 nixl_status_t
-nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &handle,
+nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
+                                const nixl_meta_dlist_t &local_descs,
+                                const nixl_meta_dlist_t &remote_descs,
                                 nixlGpuXferReqH &gpu_req_hndl) const {
+#ifdef HAVE_UCX_GPU_DEVICE_API
```
rakhmets: We can get rid of this `ifdef` in this file. We can place the `ifdef` into `deviceMemList::createDeviceMemList` to use the UCX device API on `utils/ucx` level (as we did for other UCX device APIs).

src/utils/ucx/ucx_utils.h
michal-shalev: `device_mem_list.cpp`
`createDeviceMemList`

src/plugins/ucx/ucx_backend.h
michal-shalev: removed

src/core/nixl_agent.cpp
brminich: can it be next PR? Otherwise should the same be relevant to getGpuSignalSize?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    auto intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
+
+    if (local_descs.descCount() != remote_descs.descCount()) {
+        NIXL_ERROR << "Mismatch between local and remote descriptor counts";
```
brminich, line 1637: minor, count values are worth to be printed (next pr)

src/utils/ucx/meson.build
```diff
@@ -21,8 +21,15 @@ endif
 
 ucx_utils_inc_dirs = include_directories('.')
 
+ucx_utils_sources = ['ucx_utils.cpp', 'ucx_utils.h', 'config.h', 'config.cpp', 'rkey.cpp', 'rkey.h']
+
+# Add device_mem_list files only if UCX GPU Device API is available
+if ucx_gpu_device_api_available
+    ucx_utils_sources += ['device_mem_list.cpp', 'device_mem_list.h']
+endif
```
rakhmets:
```suggestion
ucx_utils_sources = ['ucx_utils.cpp', 'ucx_utils.h', 'config.h', 'config.cpp', 'rkey.cpp', 'rkey.h', 'device_mem_list.cpp', 'device_mem_list.h']
```

I recommend controlling the functionality at one level: using ifdef in the code, or at the meson level by including various files. Since the file-level control approach is already being used, it is better to stick to this path in this case.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,80 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+        elements.emplace_back(nixl::ucx::deviceMemElem{&localMd->mem,
+                                                       &remoteDescMd->getRkey(workerId),
+                                                       reinterpret_cast<void *>(localDesc.addr),
+                                                       remoteDesc.addr,
+                                                       localDesc.len});
+    }
+
+    try {
+        auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(*ep, elements);
+        gpu_req_hndl = reinterpret_cast<nixlGpuXferReqH>(device_mem_list.get());
```
rakhmets: Sorry, that's my oversight. `reinterpret_cast` should be inside `nixl::ucx::deviceMemList` to hide `ucp_device_mem_list_handle_h`. So that `#ifdef HAVE_UCX_GPU_DEVICE_API` can be placed only in `utils/ucx` files.

```C++
// device_mem_list.h
class deviceMemList {
public:
    deviceMemList() = delete;
    deviceMemList(const nixlUcxEp &, const std::vector<deviceMemElem> &);
    explicit deviceMemList(nixlGpuXferReqH) noexcept;

    [[nodiscard]] nixlGpuXferReqH
    get() const noexcept;

private:
    [[nodiscard]] static ucp_device_mem_list_handle_h
    createDeviceMemList(const nixlUcxEp &, const std::vector<deviceMemElem> &);

    const std::unique_ptr<ucp_device_mem_list_handle, void (*)(ucp_device_mem_list_handle_h)>
        deviceMemList_;
#endif
};

// device_mem_list.cpp
deviceMemList::deviceMemList(const nixlUcxEp &ep, const std::vector<deviceMemElem> &elements)
    : deviceMemList_{createDeviceMemList(ep, elements), &ucp_device_mem_list_release} {}

deviceMemList::deviceMemList(nixlGpuXferReqH) noexcept
    : deviceMemList_{reinterpret_cast<ucp_device_mem_list_handle_h>(nixlGpuXferReqH), &ucp_device_mem_list_release} {}

nixlGpuXferReqH
get() const noexcept {
    return reinterpret_cast<nixlGpuXferReqH>(deviceMemList_.get());
}

ucp_device_mem_list_handle_h
deviceMemList::createDeviceMemList(const nixlUcxEp &ep,
                                   const std::vector<deviceMemElem> &elements) {
    ...
}
#else
deviceMemList::deviceMemList(const nixlUcxEp &ep, const std::vector<deviceMemElem> &elements) {
    throw std::runtime_error("UCX was not compiled with GPU device API support");
}

deviceMemList::deviceMemList(nixlGpuXferReqH) noexcept {}

nixlGpuXferReqH
get() const noexcept {
    return nullptr;
}
#endif
```

src/plugins/ucx/ucx_backend.cpp
michal-shalev: check now

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,73 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+        auto *localMd = static_cast<nixlUcxPrivateMetadata *>(local_descs[i].metadataP);
+        auto *remoteMdDesc = static_cast<nixlUcxPublicMetadata *>(remote_descs[i].metadataP);
+
+        local_memhs.push_back(localMd->mem.getMemh());
+        remote_rkeys.push_back(&remoteMdDesc->getRkey(workerId));
+    }
+
+    try {
+        auto device_mem_list = std::make_unique<nixl::ucx::deviceMemList>(*ep, local_memhs, remote_rkeys);
+        gpu_req_hndl = reinterpret_cast<nixlGpuXferReqH>(device_mem_list.release());
```
rakhmets: This will not work.
```C++
//nixl_device.cuh
nixlGpuXferReqParams(nixlGpuXferReqH req_hndl,
                         ...)
        : mem_list{static_cast<ucp_device_mem_list_handle_h>(req_hndl)},
          ...
    ucp_device_mem_list_handle_h mem_list;
```

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Removed deviceMemElem

src/plugins/ucx/ucx_backend.cpp
michal-shalev: removed deviceMemElem

src/utils/ucx/device_mem_list.h
michal-shalev: I removed deviceMemElem

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,73 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    auto remoteMd = static_cast<nixlUcxPublicMetadata *>(remote_descs[0].metadataP);
+    if (!remoteMd || !remoteMd->conn) {
+        NIXL_ERROR << "No connection found in remote metadata";
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    size_t workerId = intHandle->getWorkerId();
+    nixlUcxEp *ep = remoteMd->conn->getEp(workerId).get();
+
+        std::vector<ucp_mem_h> local_memhs;
```
```suggestion
    std::vector<nixlUcxMem> local_memhs;
```

src/plugins/ucx/ucx_backend.cpp
rakhmets: It releases using deleter of `unique_ptr`.
```C++
deviceMemList::deviceMemList(const ucp_device_mem_list_handle_h device_mem_list) noexcept
    : deviceMemList_{device_mem_list, &ucp_device_mem_list_release} {}
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,73 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    std::vector<const nixl::ucx::rkey*> remote_rkeys;
+    local_memhs.reserve(local_descs.descCount());
+    remote_rkeys.reserve(remote_descs.descCount());
+
+    for (size_t i = 0; i < static_cast<size_t>(local_descs.descCount()); i++) {
+        auto *localMd = static_cast<nixlUcxPrivateMetadata *>(local_descs[i].metadataP);
+        auto *remoteMdDesc = static_cast<nixlUcxPublicMetadata *>(remote_descs[i].metadataP);
```
```suggestion
        auto localMd = static_cast<nixlUcxPrivateMetadata *>(local_descs[i].metadataP);
        auto remoteMdDesc = static_cast<nixlUcxPublicMetadata *>(remote_descs[i].metadataP);
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1629,73 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
-nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {}
+nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
+#ifdef HAVE_UCX_GPU_DEVICE_API
+    if (gpu_req_hndl == nullptr) {
+        NIXL_WARN << "Attempting to release null GPU transfer request handle";
+        return;
+    }
+
+    auto *device_mem_list = reinterpret_cast<nixl::ucx::deviceMemList *>(gpu_req_hndl);
+    delete device_mem_list;
```
rakhmets: This is incorrect. `nixl::ucx::deviceMemList` allocates and frees device memory via `ucp_device_mem_list_create`, and `ucp_device_mem_list_release`. Please revert to previously suggested approach. It should be:
```C++
void
nixlUcxEngine::releaseGpuXferReq(nixlGpuXferReqH gpu_req_hndl) const {
    nixl::ucx::deviceMemList device_mem_list{gpu_req_hndl};
    gpu_req_hndl = nullptr;
}
```

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

test/gtest/test_transfer.cpp
michal-shalev: In the next PR there will be a test for signal and I'll remove this one

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: done

test/gtest/device_api/single_write_test.cu
```diff
@@ -0,0 +1,585 @@
...
+            printf("Thread %d: nixlGpuPostSingleWriteXferReq failed iteration %lu: status=%d (0x%x)\n",
+                   threadIdx.x,
+                   (unsigned long)i,
+                   status,
+                   static_cast<unsigned int>(status));
+            return;
+        }
+
+        status = nixlGpuGetXferStatus<level>(*xfer_status_ptr);
+        if (status != NIXL_SUCCESS && status != NIXL_IN_PROG) {
```
ofirfarjun7, line 59: Can remove this check in next PR.

test/gtest/device_api/single_write_test.cu
```diff
@@ -0,0 +1,585 @@
...
+                      uint64_t remote_addr,
+                      size_t size,
+                      size_t num_iters,
+                      bool is_no_delay,
+                      unsigned long long *start_time_ptr,
+                      unsigned long long *end_time_ptr) {
+    nixl_status_t ret = NIXL_SUCCESS;
+    cudaError_t err;
+
+    TestSingleWriteKernel<level><<<1, num_threads>>>(req_hdnl,
```
ofirfarjun7, line 108: Need to support also number of blocks or run kernel by dividing threads according to max threads in block.
can do in next PR.

test/gtest/device_api/single_write_test.cu
```diff
@@ -0,0 +1,585 @@
...
+    logResultsPublic(size_t size,
+                     size_t count,
+                     size_t num_iters,
+                     unsigned long long start_time_cpu,
+                     unsigned long long end_time_cpu) {
+        logResults(size, count, num_iters, start_time_cpu, end_time_cpu);
+    }
+};
+
+TEST_P(SingleWriteTest, BasicSingleWriteTest) {
```
ofirfarjun7, line 382: Improve code reuse in Next PRs

test/gtest/device_api/single_write_test.cu
```diff
@@ -0,0 +1,585 @@
...
+        printf("Failed to launch kernel: %s\n", cudaGetErrorString(err));
+        ret = NIXL_ERR_BACKEND;
+    }
+
+    return ret;
+}
+
+class SingleWriteTest : public DeviceApiTestBase {
+protected:
+    std::string getBackendName() const { return "UCX"; }
```
ofirfarjun7, line 135: We can probably move some of this code to a common place, in next PRs

brminich (approved):

src/utils/ucx/device_mem_list.cpp
```diff
@@ -0,0 +1,114 @@
...
+    NIXL_DEBUG << "Created device memory list handle with " << local_mems.size() << " elements";
+    return ucx_handle;
+}
+
+#else
+
+deviceMemList::deviceMemList(const nixlUcxEp &ep,
+                             const std::vector<nixlUcxMem> &local_mems,
+                             const std::vector<const nixl::ucx::rkey *> &remote_rkeys) {
+    NIXL_ERROR << "UCX GPU device API not supported";
```
rakhmets, line 99:
```suggestion
```

michal-shalev: next PR

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1627,62 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+
+    try {
+        auto device_mem_list =
+            std::make_unique<nixl::ucx::deviceMemList>(*ep, local_mems, remote_rkeys);
+        gpu_req_hndl = device_mem_list->get();
+        device_mem_list.release();
+        return NIXL_SUCCESS;
+    }
+    catch (const std::exception &e) {
+        NIXL_ERROR << "Failed to create device memory list for GPU transfer: " << e.what();
```
rakhmets, line 1676:
```suggestion
        NIXL_ERROR << e.what();
```
It's enough to `e.what();` here.

michal-shalev: next PR

test/gtest/device_api/single_write_test.cu
michal-shalev: next PR

test/gtest/device_api/single_write_test.cu
michal-shalev: next PR

test/gtest/device_api/single_write_test.cu
michal-shalev: next PR

test/gtest/device_api/single_write_test.cu
michal-shalev: next PR

src/core/nixl_agent.cpp
michal-shalev: next PR

src/api/cpp/nixl.h
michal-shalev: next PR

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1623,13 +1627,62 @@ nixl_status_t nixlUcxEngine::releaseReqH(nixlBackendReqH* handle) const
...
+    local_mems.reserve(local_descs.descCount());
+    remote_rkeys.reserve(remote_descs.descCount());
+
+    for (size_t i = 0; i < static_cast<size_t>(local_descs.descCount()); i++) {
+        auto localMd = static_cast<nixlUcxPrivateMetadata *>(local_descs[i].metadataP);
+        auto remoteMdDesc = static_cast<nixlUcxPublicMetadata *>(remote_descs[i].metadataP);
+
+        local_mems.push_back(localMd->mem);
+        remote_rkeys.push_back(&remoteMdDesc->getRkey(workerId));
+    }
```
rakhmets, line 1666:
```C++
    auto intHandle = static_cast<const nixlUcxBackendH *>(&req_hndl);
    size_t workerId = intHandle->getWorkerId();
    auto remoteMd = static_cast<nixlUcxPublicMetadata *>(remote_descs[0].metadataP);
    if (!remoteMd || !remoteMd->conn) {
        NIXL_ERROR << "No connection found in remote metadata";
        return NIXL_ERR_NOT_FOUND;
    }

    nixlUcxEp *ep = remoteMd->conn->getEp(workerId).get();

    std::vector<nixlUcxMem> local_mems;
    local_mems.reserve(local_descs.descCount());
    for (size_t i = 0; i < static_cast<size_t>(local_descs.descCount()); i++) {
        auto localMd = static_cast<nixlUcxPrivateMetadata *>(local_descs[i].metadataP);
        local_mems.push_back(localMd->mem);
    }

    std::vector<const nixl::ucx::rkey *> remote_rkeys;
    remote_rkeys.reserve(remote_descs.descCount());
    for (size_t i = 0; i < static_cast<size_t>(local_descs.descCount()); i++) {
        auto remoteMdDesc = static_cast<nixlUcxPublicMetadata *>(remote_descs[i].metadataP);
        remote_rkeys.push_back(&remoteMdDesc->getRkey(workerId));
    }
```
removed duplicated checks.

PR #793, opened 2025-09-16
CI: update pytorch base image version

dpressle:
Update to newer version to have cuda 12.9 alligned with build servers

## What?

## Why?

## How?
architecture, approach, etc._

### DIFF
--- a/.ci/jenkins/lib/build-matrix.yaml
+++ b/.ci/jenkins/lib/build-matrix.yaml
@@ -6,7 +6,7 @@
 # Key Components:
 # - Job Configuration: Defines timeout, failure behavior, and Kubernetes resources
 # - Docker Images: Specifies the container images used for different build stages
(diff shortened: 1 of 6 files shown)


.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -6,7 +6,7 @@
 # Key Components:
 # - Job Configuration: Defines timeout, failure behavior, and Kubernetes resources
 # - Docker Images: Specifies the container images used for different build stages
-#   - PyTorch images (24.10 and 25.02) for building and testing
+#   - PyTorch images (24.10 and 25.06) for building and testing
```
ovidiusm, line 9: 24.10 is based on CUDA 12.6 which we officially do not support anymore, minimum is 12.8. It would be great if we could find another base image for ubuntu 22.04. nvcr.io/nvidia/cuda:12.8.0-devel-ubuntu22.04 might work

PR #798, opened 2025-09-16
DEVICE/API: Remove deviceMemList class and simplify create/release GPU request handle

michal-shalev:
## What?
Decided to revert to https://github.com/ai-dynamo/nixl/pull/789/commits/247e211746aaeed89cfefdad76366ac437e002fd since it's simpler and the `deviceMemList` class requires to add mapping to gpu req handle to release it.


src/utils/ucx/gpu_xfer_req_h.h
```diff
@@ -35,26 +35,24 @@ class nixlUcxMem;
 namespace nixl::ucx {
 class rkey;
 
-class deviceMemList {
+class gpuXferReqH {
```
rakhmets: No need to add a class. It should be just:
```C++
namespace nixl::ucx {
nixlGpuXferReqH createGpuXferReq(const nixlUcxEp &,
                                 const std::vector<nixlUcxMem> &,
                                 const std::vector<const nixl::ucx::rkey *> &);

void releaseGpuXferReq(nixlGpuXferReqH) noexcept;
}
```

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -28,25 +25,30 @@ namespace nixl::ucx {
...
+    if (gpu_req == nullptr) {
+        NIXL_WARN << "Attempting to release null GPU transfer request handle";
+        return;
+    }
 
-nixlGpuXferReqH
-deviceMemList::get() const noexcept {
-    return reinterpret_cast<nixlGpuXferReqH>(deviceMemList_.get());
+    ucp_device_mem_list_handle_h ucx_handle =
+        reinterpret_cast<ucp_device_mem_list_handle_h>(gpu_req);
```
rakhmets:
```suggestion
    auto ucx_handle = reinterpret_cast<ucp_device_mem_list_handle_h>(gpu_req);
```

yosefe (dismissed):

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -28,25 +25,30 @@ namespace nixl::ucx {
...
-deviceMemList::deviceMemList(const nixlUcxEp &ep,
-                             const std::vector<nixlUcxMem> &local_mems,
-                             const std::vector<const nixl::ucx::rkey *> &remote_rkeys)
-    : deviceMemList_{createDeviceMemList(ep, local_mems, remote_rkeys),
-                     &ucp_device_mem_list_release} {}
+nixlGpuXferReqH
+gpuXferReqH::create(const nixlUcxEp &ep,
+                    const std::vector<nixlUcxMem> &local_mems,
+                    const std::vector<const nixl::ucx::rkey *> &remote_rkeys) {
+    ucp_device_mem_list_handle_h ucx_handle = createDeviceMemList(ep, local_mems, remote_rkeys);
```
rakhmets: `createDeviceMemList` is not needed. The implementation of `createDeviceMemList` can be moved here.

src/utils/ucx/gpu_xfer_req_h.h
```diff
@@ -35,26 +35,24 @@ class nixlUcxMem;
...
+    create(const nixlUcxEp &ep,
+           const std::vector<nixlUcxMem> &local_mems,
+           const std::vector<const nixl::ucx::rkey *> &remote_rkeys);
+
+    static void
+    release(nixlGpuXferReqH gpu_req);
 
-#ifdef HAVE_UCX_GPU_DEVICE_API
 private:
+#ifdef HAVE_UCX_GPU_DEVICE_API
```
rakhmets: `ifdef` can be removed from the header.
And `#include <memory>`.

The following header is not needed in `.h`, but in `.cpp`.
```C++
extern "C" {
#include <ucp/api/device/ucp_host.h>
#endif
}
```

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: done

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: `ucp_device_mem_list_handle_h` is in the header
removed #include <memory>

rakhmets: `ucp_device_mem_list_handle_h` is not needed in this header.
Only these functions are needed:
```C++
namespace nixl::ucx {
nixlGpuXferReqH
createGpuXferReq(const nixlUcxEp &, const std::vector<nixlUcxMem> &,
                 const std::vector<const nixl::ucx::rkey *> &);
void
releaseGpuXferReq(nixlGpuXferReqH) noexcept;
}
```

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: copied your comment here:
No need to add a class. It should be just:

```
namespace nixl::ucx {
nixlGpuXferReqH createGpuXferReq(const nixlUcxEp &,
                       const std::vector<nixlUcxMem> &,
                       const std::vector<const nixl::ucx::rkey *> &);

void releaseGpuXferReq(nixlGpuXferReqH gpu_req) noexcept;
}
```

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -88,25 +70,33 @@ deviceMemList::createDeviceMemList(const nixlUcxEp &ep,
...
-    return ucx_handle;
+    return reinterpret_cast<nixlGpuXferReqH>(ucx_handle);
+}
+
+void
+releaseGpuXferReq(nixlGpuXferReqH gpu_req) noexcept {
+    if (gpu_req == nullptr) {
+        NIXL_WARN << "Attempting to release null GPU transfer request handle";
+        return;
+    }
```
rakhmets: Not sure that we need this warning message.
It doesn't seem to be wrong to pass `nullptr` to this function.
Maybe remove these lines.

rakhmets (approved):

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: yes it was before I removed the class

michal-shalev: done

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: done

rakhmets (dismissed):

rakhmets (approved):

yosefe (approved):

brminich (approved):

PR #800, opened 2025-09-16
BUILD: Add configurable CUDA device debug info flag

michal-shalev:
## What?
Add configurable CUDA device debug info flag with `cuda_enable_debug` meson option.

## Why?
Meson automatically adds the -G flag for CUDA debug builds, which significantly degrades performance. The -G flag generates device debug info and turns off optimizations unless -dopt=on is specified. 
Reference: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/ 4.2.3.3

## How?
- Add cuda_enable_debug boolean option (default: false for performance)
- Conditionally add -dopt=on only when debug is disabled to prevent automatic -G inclusion
- When enabled, allows meson to add -G flag for device debugging/profiling

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -100,6 +100,10 @@ if cuda_dep.found()
     nvcc_flags_link += ['-gencode=arch=compute_80,code=sm_80']
     nvcc_flags_link += ['-gencode=arch=compute_90,code=sm_90']
     add_project_link_arguments(nvcc_flags_link, language: 'cuda')
+
+    if not get_option('cuda_enable_debug')
+        add_project_arguments('-dopt=on', language: 'cuda')
+    endif
 else
     warning('CUDA not found. UCX backend will be built without CUDA support, and some plugins will be disabled.')
 endif
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -24,6 +24,7 @@ option('gds_path', type: 'string', value: '/usr/local/cuda/', description: 'Path
 option('cudapath_inc', type: 'string', value: '', description: 'Include path for CUDA')
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
+option('cuda_enable_debug', type: 'boolean', value: false, description: 'Enable CUDA device debug info (-G flag) for performance analysis')
 option('static_plugins', type: 'string', value: '', description: 'Plugins to be built-in, comma-separated')
 option('build_docs', type: 'boolean', value: false, description: 'Build Doxygen documentation')
 option('log_level', type: 'combo', choices: ['trace', 'debug', 'info', 'warning', 'error', 'fatal', 'auto'], value: 'auto', description: 'Log Level (auto: auto-detect based on build type: trace for debug builds, info for release builds)')


meson_options.txt
```diff
@@ -24,6 +24,7 @@ option('gds_path', type: 'string', value: '/usr/local/cuda/', description: 'Path
 option('cudapath_inc', type: 'string', value: '', description: 'Include path for CUDA')
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
+option('cuda_enable_debug', type: 'boolean', value: false, description: 'Enable CUDA device debug info (-G flag) for performance analysis')
```
yosefe: the option description seem not match what it's doing - adding -dopt=on

michal-shalev: changed the description

meson.build
```diff
@@ -100,6 +100,10 @@ if cuda_dep.found()
     nvcc_flags_link += ['-gencode=arch=compute_80,code=sm_80']
     nvcc_flags_link += ['-gencode=arch=compute_90,code=sm_90']
     add_project_link_arguments(nvcc_flags_link, language: 'cuda')
+
+    if not get_option('enable_cuda_debug')
+        add_project_arguments('-dopt=on', language: 'cuda')
+    endif
```
rakhmets, line 106:
```suggestion
    if get_option('enable_cuda_debug')
        add_project_arguments('-G', language: 'cuda')
    endif
```
When `-G` is not specified, `-dopt=on` is implicit.

meson_options.txt
```diff
@@ -24,6 +24,7 @@ option('gds_path', type: 'string', value: '/usr/local/cuda/', description: 'Path
 option('cudapath_inc', type: 'string', value: '', description: 'Include path for CUDA')
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
+option('enable_cuda_debug', type: 'boolean', value: false, description: 'Enable CUDA debug mode (disables -dopt=on optimization)')
```
rakhmets, line 27:
```suggestion
option('enable_cuda_debug', type: 'boolean', value: false, description: 'Enable CUDA debug mode')
```

rakhmets (dismissed):

meson.build
michal-shalev: I checked here:
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/ 4.2.3.3
And also compiled with and without `add_project_arguments('-dopt=on', language: 'cuda')` to verify.

yosefe (dismissed):

meson.build
brminich: but as @rakhmets  mentioned 
```
4.2.3.12. --dopt kind (-dopt)
Enable device code optimization. When specified along with -G, enables limited debug information generation for optimized device code (currently, only line number information). When -G is not specified, -dopt=on is implicit.
```
So, should we just add -G for debug mode and do not add it for release?

rakhmets: I think the issue is that `-G` is added by meson.
So, in debug build it will be `-G`. But I think the purpose of the PR is:
- debug build `-G  -dopt=on`
- debug + cuda debug: `-G`

brminich: why do we need a separate debug option for cuda? I'd just make sure we have -G for debug and we do not have it for release

michal-shalev: debug is the default build type, and it's not documented anywhere, out of the box users can get very bad performance because of this "-G" flag

meson_options.txt
ovidiusm, line 27: Could we tie this to the buildtype (debug/release) instead of adding a new option?

brminich: yes and maybe we need to build NIXL in release mode by default

michal-shalev:
Closing this PR following https://github.com/ai-dynamo/nixl/pull/869

PR #801, opened 2025-09-16
Fix build when libfabric is not installed

ovidiusm:
## What?
Fix build regression.

## Why?
Missing check for libfabric dependency.

### DIFF
--- a/test/unit/utils/meson.build
+++ b/test/unit/utils/meson.build
@@ -14,7 +14,9 @@
 # limitations under the License.
 
 subdir('common')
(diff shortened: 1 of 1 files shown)


aranadive (approved):

yosefe:
@ovidiusm we didn't see this in CI because we always have libbfabric ?

ovidiusm:
> @ovidiusm we didn't see this in CI because we always have libbfabric ?

That's right. We only hit this problem on individual dev environments that do not have the dependency installed.

PR #802, opened 2025-09-16
Replace full EFA installer with minimal EFA + custom libfabric build

akkart-aws:
## What?

Replace full EFA installer with minimal EFA installation and custom libfabric build from source. This change switches from using `efa_installer.sh -y -g` (which installs OpenMPI, NCCL, and other unwanted packages) to `efa_installer.sh -y --minimal` (which installs only RDMA core and EFA drivers) combined with a custom libfabric v2.3.0 build configured specifically for EFA, CUDA, and GDRCopy support.

## Why?

The original EFA installer was installing unwanted components like OpenMPI and NCCL that were causing package conflicts and bloating the installation. The user specifically needed only EFA drivers and libfabric components without the additional packages that come with the full EFA installer.

## How?

1. __Modified EFA installer usage__: Changed from `-y -g` to `-y --minimal` flag to install only essential RDMA core and EFA driver components

2. __Added custom libfabric build__: Downloads libfabric v2.3.0 from GitHub releases and builds with specific configure options:

   - `--enable-efa` for EFA provider support
   - `--disable-verbs/psm3/opx/usnic/rstream` to disable unused providers
   - `--with-cuda=/usr/local/cuda --enable-cuda-dlopen` for CUDA support
   - `--with-gdrcopy --enable-gdrcopy-dlopen` for GDRCopy support

3. __Updated build system__: Added `LIBFABRIC_VERSION` and `LIBFABRIC_INSTALL_DIR` parameters following the existing UCX pattern, updated environment variables and meson configuration to use custom libfabric paths

4. __Applied consistently__: Updated `.gitlab/build.sh`, `contrib/Dockerfile`, and `benchmark/nixlbench/contrib/Dockerfile` with the same approach


yosefe (approved):

.gitlab/build.sh
```diff
@@ -22,21 +22,27 @@ set -o pipefail
...
 # UCX_VERSION is the version of UCX to build override default with env variable.
 UCX_VERSION=${UCX_VERSION:-v1.19.0}
 # EFA_INSTALLER_VERSION is the version of EFA installer to use, defaults to "latest"
 EFA_INSTALLER_VERSION=${EFA_INSTALLER_VERSION:-latest}
-
+# LIBFABRIC_VERSION is the version of libfabric to build override default with env variable.
+LIBFABRIC_VERSION=${LIBFABRIC_VERSION:-v2.3.0}
 if [ -z "$INSTALL_DIR" ]; then
-    echo "Usage: $0 <install_dir> <ucx_install_dir>"
+    echo "Usage: $0 <install_dir> <ucx_install_dir> <libfabric_install_dir>"
```
ovidiusm: I'm not convinced that adding an extra argument here is the best option, and it is not used in callers anyways.

Could it be an env variable?

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -128,10 +129,29 @@ COPY --from=nixlbench . /workspace/nixlbench
 # Install AWS SDK C++ dependencies and build
 RUN apt-get update && apt-get install -y libcurl4-openssl-dev libssl-dev uuid-dev zlib1g-dev hwloc libhwloc-dev
 
-# Install EFA (Elastic Fabric Adapter)
+# Install EFA (Elastic Fabric Adapter) - minimal installation
 RUN curl -fsSL "https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz" | tar xz && \
```
ovidiusm: Could we have wget with retries here and below instead of curl | ? Prevents flaky tests due to transient network errors.

contrib/Dockerfile
```diff
@@ -74,10 +75,29 @@ RUN git clone --depth 1 https://github.com/etcd-cpp-apiv3/etcd-cpp-apiv3.git &&
     mkdir build && cd build && \
     cmake .. -DBUILD_ETCD_CORE_ONLY=ON -DCMAKE_BUILD_TYPE=Release && make -j${NPROC:-$(nproc)} && make install
 
-# Install EFA (Elastic Fabric Adapter)
+# Install EFA (Elastic Fabric Adapter) - minimal installation
 RUN curl -fsSL "https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz" | tar xz && \
```
ovidiusm: Same here, should have retries in the download

benchmark/nixlbench/contrib/Dockerfile
akkart-aws: I dont think we need efa-installer or libfabric to install nixlbench. I think it was incorrect to add the EFA installer and libfabric here to begin with. Hence I have removed it.

.gitlab/build.sh
akkart-aws: Done

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -104,11 +106,38 @@ ARG DEFAULT_PYTHON_VERSION
...
+                --disable-verbs \
+                --disable-psm3 \
+                --disable-opx \
+                --disable-usnic \
+                --disable-rstream \
+                --enable-efa \
+                --with-cuda=/usr/local/cuda \
+                --enable-cuda-dlopen \
+                --with-gdrcopy \
+                --enable-gdrcopy-dlopen && \
```
aranadive, line 136: Do you need to add these to the other libfabric build in .gitlab/build.sh?

akkart-aws: yes. for gpu direct rdma they are needed.

aranadive: Ok can you add in separate PR then?

PR #804, opened 2025-09-17
CI: Pin GHA runner ver to fix CMake compatibility error

Alexey-Rivkin:
## What?
Pin GitHub Actions runner to ubuntu-24.04 in AWS S3 validation workflow.

## Why?
Fix CMake compatibility error where ubuntu-latest runner includes newer CMake that drops support for CMake < 3.5, causing builds to fail with "Compatibility with CMake < 3.5 has been removed" error.

### DIFF
--- a/.github/workflows/aws_s3_validation.yml
+++ b/.github/workflows/aws_s3_validation.yml
@@ -10,7 +10,7 @@ jobs:
   run_aws_s3_tests:
     name: Run AWS S3 Tests
     environment: SWX_AWS
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     env:
       AWS_DEFAULT_REGION: eu-central-1
       AWS_ACCESS_KEY_ID: ${{ secrets.NIXL_AWS_ACCESS_KEY_ID }}


michal-shalev (dismissed):

dpressle (dismissed):

yosefe (approved):

ovidiusm (dismissed):

brminich (dismissed):

.github/workflows/aws_efa_validation.yml
```diff
@@ -44,7 +45,7 @@ jobs:
           sudo apt-get -qq install -y curl gettext git gzip jq tar unzip
 
           # Install kubectl
-          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
```
dpressle: i dont think this will help, how is downloading stable version more error prone then downloading specific version, i think we also need to add retry mechanism here

dpressle (changes requested):

.github/workflows/aws_efa_validation.yml
Alexey-Rivkin: The error was fetching the version, not downloading the binary

dpressle: yes it comes before the download curl so it failed on that but its same symptom - the url is unavailable, and now you will get it in the second curl

Alexey-Rivkin: aa899178 Use curl’s built-in retry option and keep the kubectl version pinned

brminich (approved):

PR #806, opened 2025-09-18
OBJ: add ca_bundle option in obj plugin and nixlbench

ziruiliu:
## What?
Add a new param 'ca_bundle' for obj plugin and corresponding option 'obj_ca_bundle' in nixlbench

## Why?
AWS cpp sdk does not honor environment variable AWS_CA_BUNDLE when initializing ClientConfiguration. 
When testing with a S3-compatible storage with self-signed certificates, the plugin and nixlbench are not working

## How?
pass certificate path to the plugin, and set caFile when initializing ClientConfiguration


benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -120,6 +120,7 @@ DEFINE_string(obj_endpoint_override, "", "Endpoint override for S3 backend");
 DEFINE_string(obj_req_checksum,
               XFERBENCH_OBJ_REQ_CHECKSUM_SUPPORTED,
               "Required checksum for S3 backend [supported, required]");
+DEFINE_string(obj_ca_bundle, "", "CA bundle for S3 backend");
```
ovidiusm:
```suggestion
DEFINE_string(obj_ca_bundle, "", "Path to CA bundle for S3 backend");
```

ovidiusm (dismissed):

ovidiusm:
It also needs to be added to benchmark/kvbench/commands/args.py (see obj_endpoint_override for example)

ziruiliu:
> It also needs to be added to benchmark/kvbench/commands/args.py (see obj_endpoint_override for example)

Hi @ovidiusm ,  I have updated benchmark/kvbench/commands/args.py according to you suggestion

aranadive (approved):

ovidiusm (approved):

ziruiliu:
Hi @w1ldptr and @barneuman, would you please review this change that add ca_bundle in obj plugin?
It is critical to pass self-signed certificate into the obj plugin, when testing with private S3-compatible storage.
Thanks!

barneuman (approved):
LGTM

PR #809, opened 2025-09-18
Add same-agent transfer optimization with memcpy fallback

akkart-aws:
# Implement direct memcpy for same-agent transfers and add debug logging

## What?
This PR implements direct memory copy (memcpy) for transfers within the same agent to avoid unnecessary network operations. It also adds comprehensive debug logging for transfer path selection and agent comparison to help diagnose transfer routing decisions.

## Why?
Currently, same-agent transfers (when both source and destination are within the same agent process) are failing with `agent_example.sh` because the system attempts to use network-based transfers through libfabric even when both memory regions are local to the same process. This results in:

- Connection handshake failures between local endpoints
- Transfer failures with errors like "Unresponsive receiver" and "Software caused connection abort"

The root cause is that the current transfer path selection logic in `nixlAgent::createXferReq()` and related functions doesn't detect when both source and destination belong to the same agent, leading to inefficient network-based transfers instead of simple memory copies.

## How?
The implementation adds:

1. **Same-agent detection logic**: Check if the remote agent name matches the local agent name before attempting network transfers
2. **Direct memcpy path**: When same-agent transfer is detected, perform direct memory copy operations instead of going through the network backend
3. **Debug logging**: Add comprehensive logging to show:
   - Transfer path selection decisions (same-agent vs. network)
   - Agent name comparisons 
   - Memory copy operation details
4. **Fallback safety**: Maintain existing network path as fallback if direct copy fails

This ensures that intra-agent transfers use the most efficient path (direct memory copy) while preserving the existing network-based functionality for true inter-agent transfers.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1067,7 +1067,39 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
         NIXL_DEBUG << "Processing descriptor " << desc_idx << " GPU " << gpu_id
                    << " addr: " << transfer_addr << " size: " << transfer_size;
 
-        // Prepare and submit transfer
+        NIXL_DEBUG << "DEBUG: remote_agent='" << remote_agent << "' localAgent='" << localAgent
+                   << "'";
+
+        // Check for same-agent (local) transfer - handle with direct memcpy
+        if (remote_agent == localAgent) {
+            NIXL_DEBUG << "Same-agent transfer detected for descriptor " << desc_idx
+                       << ", using memcpy fallback for " << transfer_size << " bytes";
+
+            // For same-agent transfers, we need to copy directly between the descriptor addresses
+            // The remote[desc_idx].addr should be the target address for the transfer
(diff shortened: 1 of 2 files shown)


src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1067,7 +1067,39 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
         NIXL_DEBUG << "Processing descriptor " << desc_idx << " GPU " << gpu_id
                    << " addr: " << transfer_addr << " size: " << transfer_size;
 
-        // Prepare and submit transfer
+        NIXL_DEBUG << "DEBUG: remote_agent='" << remote_agent << "' localAgent='" << localAgent
```
mkhazraee, line 1070: @akkart-aws  Are the debugs temporarily, or you want to keep them? Both are fine, just double checking before we merge the PR.

akkart-aws: Thanks @mkhazraee. This was redundant message. I have fixed it now.

mkhazraee (approved):

PR #815, opened 2025-09-22
DEVICE/API: Change addresses to offsets and add channel_id

michal-shalev:
## What?
Implement support for passing local and remote addresses to UCX device memory lists, enabling GPU-initiated transfers with pre-configured address pairs and offset-based operations, and add channel ID to the APIs.

## Why?
Improve user experience by allowing GPU kernels to work with simple offsets relative to pre-configured base addresses, rather than requiring users to manage and pass absolute addresses for every transfer operation.
Adding channel ID to the device APIs would enable pre-resolved QP/channel selection at runtime, allowing users to register memory once for all experts and then dynamically choose the correct channel during execution, which reduces connection setup time and initialization overhead.

## How?
- Extended `createGpuXferReq` to accept local memory objects, remote rkeys, and remote addresses
- Added `getBase()` method to `nixlUcxMem` to access local addresses
- Added `getSize()` method to `nixlUcxMem` to access descriptor size
- Modified `createGpuXferReq` implementation to populate `local_addr` and `remote_addr` fields in UCX memory list elements
- Updated UCX backend to extract addresses from descriptors and pass them to the device memory list creation
- Added `channel_id` to NIXL device APIs


src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -193,17 +194,17 @@ template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
 nixlGpuPostWriteXferReq(nixlGpuXferReqH req_hndl,
                         const size_t *sizes,
-                        void *const *addrs,
-                        const uint64_t *remote_addrs,
+                        void *const *offsets,
```
itayalroy: I think it makes more sense to use `uint64_t` for local offsets instead of `void *`, since they represent byte offsets, not actual pointers.

itayalroy:
This PR is missing a change to nixlGpuSignal:

```
struct nixlGpuSignal {
    uint64_t inc = 0;
    uint64_t remote_addr = 0;
};
```
should be changed to:
```
struct nixlGpuSignal {
    uint64_t inc = 0;
    uint64_t offset = 0;
};
```

src/api/gpu/ucx/nixl_device.cuh
yosefe: use ptrdiff_t or size_t

michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -115,14 +118,15 @@ nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH req_hndl,
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
 nixlGpuPostSignalXferReq(nixlGpuXferReqH req_hndl,
-                         unsigned index,
+                         uint16_t channel_id,
                          const nixlGpuSignal &signal,
+                         unsigned signal_index,
```
itayalroy: Does `signal_index` mean descriptor index? If so then:
1. name is misleading
2. `signal_index` is tied to `signal_offset` which is inside `nixlGpuSignal` (IIUC they mean which descriptor to use and what offset inside the descriptor), I believe either they both go into `nixlGpuSignal` or better- remove this `nixlGpuSignal`, why do we need it?

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -131,13 +135,15 @@ nixlGpuPostSignalXferReq(nixlGpuXferReqH req_hndl,
  * @brief Post a partial memory transfer request to the GPU.
  *
  * @param req_hndl           [in]  Request handle.
+ * @param channel_id         [in]  Channel ID to use for the transfer.
  * @param count              [in]  Number of blocks to send. This is also the length of the arrays
- *                                 @a indices, @a sizes, @a addrs, and @a remote_addrs.
+ *                                 @a indices, @a sizes, @a local_offsets, and @a remote_offsets.
  * @param indices            [in]  Indices of the blocks to send.
```
itayalroy: Indices of the descriptors*, right? these are indices to the dlist

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -74,29 +74,31 @@ nixlGpuConvertUcsStatus(ucs_status_t status) {
...
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
 nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH req_hndl,
+                              uint16_t channel_id,
                               unsigned index,
-                              const void *addr,
-                              uint64_t remote_addr,
+                              uint64_t local_offset,
+                              uint64_t remote_offset,
```
yosefe:
```suggestion
                              size_t local_offset,
                              size_t remote_offset,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -74,29 +74,31 @@ nixlGpuConvertUcsStatus(ucs_status_t status) {
...
+ * @param size          [in]  Size of the memory to be transferred.
+ * @param is_no_delay   [in]  Whether to use no-delay mode.
+ * @param xfer_status   [out] Status of the transfer. If null, the status is not reported.
  *
  * @return nixl_status_t    Error code if call was not successful
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
 nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH req_hndl,
+                              uint16_t channel_id,
```
yosefe:
```suggestion
                              unsigned channel_id,
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -131,13 +132,16 @@ nixlGpuPostSignalXferReq(nixlGpuXferReqH req_hndl,
...
+ *                                 @a desc_indices, @a sizes, @a local_offsets, and @a remote_offsets.
+ * @param desc_indices       [in]  Indices of the memory descriptors to send.
  * @param sizes              [in]  Sizes of the blocks to send.
- * @param addrs              [in]  Addresses of the blocks to send.
- * @param remote_addrs       [in]  Remote addresses of the blocks to send to.
- * @param signal             [in]  Signal to be sent.
+ * @param local_offsets      [in]  Local offsets of the blocks to send.
+ * @param remote_offsets     [in]  Remote offsets of the blocks to send to.
+ * @param signal_desc_index  [in]  Index of the signal descriptor to be sent.
+ * @param signal_inc         [in]  Increment value for the signal.
```
ofirfarjun7: Need to tell user that signal will be posted only if `signal_inc` != 0.
Same for multi.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -177,36 +184,41 @@ nixlGpuPostPartialWriteXferReq(nixlGpuXferReqH req_hndl,
...
+                        size_t signal_offset,
                         bool is_no_delay = true,
                         nixlGpuXferStatusH *xfer_status = nullptr) {
     const nixlGpuXferReqParams params{req_hndl, is_no_delay, xfer_status};
 
     ucs_status_t status =
         ucp_device_put_multi<static_cast<ucs_device_level_t>(level)>(params.mem_list,
-                                                                     addrs,
-                                                                     remote_addrs,
+                                                                     channel_id,
```
ofirfarjun7: Check for put_multi latest API and update accordingly.

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -34,29 +34,35 @@ namespace nixl::ucx {
...
 
-    if (local_mems.empty() || remote_rkeys.empty()) {
-        throw std::invalid_argument("Empty memh or rkey lists provided");
+    if (local_mems.empty() || remote_rkeys.empty() || remote_addrs.empty()) {
+        throw std::invalid_argument("Empty memory, rkey, or address lists provided");
     }
 
-    if (local_mems.size() != remote_rkeys.size()) {
-        throw std::invalid_argument("Local memh and remote rkey lists must have same size");
+    if (local_mems.size() != remote_rkeys.size() || local_mems.size() != remote_addrs.size()) {
```
ofirfarjun7, line 48: Maybe just add `local_mems.empty() ||` here and remove the previous check?

michal-shalev: I'll change the checks anyways in the next PR

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -115,14 +114,16 @@ nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH req_hndl,
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
 nixlGpuPostSignalXferReq(nixlGpuXferReqH req_hndl,
-                         unsigned index,
-                         const nixlGpuSignal &signal,
+                         unsigned signal_desc_index,
+                         uint64_t signal_inc,
+                         size_t signal_offset,
```
rakhmets, line 119: Why was structure `nixlGpuSignal` deleted?
Maybe:
```C++
struct nixlGpuSignal {
    uint64_t inc = 0;
    size_t offset = 0;
};
```

michal-shalev: It was discussed and agreed with @ofirfarjun7 and @yosefe following @itayalroy's request to change it.

itayalroy (approved):

src/api/gpu/ucx/nixl_device.cuh
rakhmets: What is the reason for this change? The advantages of this solution are not obvious to me. These two parameters are always passed together and describe a single entity. From the point of view of the NIXL API, it is more natural to use a structure with two fields.

itayalroy: IMO the advantage is that it's just cleaner- wrapping these two arguments in a struct doesn’t add any value and only complicates the API usage by forcing users to stack allocate a `nixlGpuSignal` and fill it before every submit call

michal-shalev: Also, there are inconsistencies across APIs in how the signal parameters are used.
Both `nixlGpuPostPartialWriteXferReq` and `nixlGpuPostSignalXferReq` include a `signal_desc_index` argument, while `nixlGpuPostWriteXferReq` does not.
Because of this, the `nixlGpuSignal` struct wouldn’t actually encapsulate all signal-related fields consistently across the API surface - it would still leave one of them outside.
So keeping these parameters explicit in each function avoids partial abstraction and keeps the interface uniform.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: done

src/api/gpu/ucx/nixl_device.cuh
rakhmets: > IMO the advantage is that it's just cleaner- wrapping these two arguments in a struct doesn’t add any value and only complicates the API usage by forcing users to stack allocate a `nixlGpuSignal` and fill it before every submit call

This adds value by reducing the number of arguments. `Keep the number of function arguments low` - is a general recommendation. One of the two most common reasons why functions have too many parameters is missing an abstraction.

> Also, there are inconsistencies across APIs in how the signal parameters are used.
Both nixlGpuPostPartialWriteXferReq and nixlGpuPostSignalXferReq include a signal_desc_index argument, while nixlGpuPostWriteXferReq does not.
Because of this, the nixlGpuSignal struct wouldn’t actually encapsulate all signal-related fields consistently across the API surface - it would still leave one of them outside.
So keeping these parameters explicit in each function avoids partial abstraction and keeps the interface uniform.

It looks like a separate problem. `signal_desc_index` is related to `nixlGpuXferReqH` (`ucp_device_mem_list_handle_h`). `nixlGpuPostWriteXferReq` calls `ucp_device_put_multi`. The last entry in the descriptor list contains the remote memory registration descriptors to be used for the increment operation in case of `ucp_device_put_multi`. I think we can get rid of this parameter altogether if we impose the same condition on other UCX APIs.

michal-shalev: Actually, `signal_desc_index` refers to the signal `nixlGpuXferReqH`, not the data one.
That’s why it’s required in `nixlGpuPostPartialWriteXferReq` and `nixlGpuPostSignalXferReq`, but not in `nixlGpuPostWriteXferReq`.
In `nixlGpuPostWriteXferReq`, the signal descriptors are already part of the same request handle used for the data transfer, while in the other two APIs, the signal request handle is separate, therefore the descriptor must be specified explicitly.

PR #819, opened 2025-09-22
UCX/BACKEND: Add internal connection establishment

michal-shalev:
## What?
Add internal connection establishment to UCX backend to prevent `UCS_ERR_NOT_CONNECTED` errors during data transfers.

## Why?
UCX requires connection establishment before data transfers, but the current implementation relies on manual workarounds in test code using `genNotif`/`getNotifs` polling. This should be handled automatically at the backend level.

## How?
- Add `performConnectionEstablishment()` method that flushes all endpoints using UCX blocking pattern
- Call it automatically in `loadRemoteConnInfo()` after connection setup
- Remove `completeWireup` workarounds from all test files


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+nixlUcxEngine::performConnectionEstablishment(
+    const std::string &remote_agent,
+    const std::shared_ptr<nixlUcxConnection> &conn) const {
+    NIXL_DEBUG << "Establishing connection with " << remote_agent;
+
+    // Flush all endpoints to ensure connection establishment
+    // and avoid UCS_ERR_NOT_CONNECTED errors during data transfers
+    for (size_t i = 0; i < conn->eps.size(); ++i) {
+        nixlUcxReq req;
+        nixl_status_t ret = conn->eps[i]->flushEp(req);
```
tstamler, line 1833: I'm sure just flushing the endpoint is fine, but I want to make sure this doesn't a) make the makeConnection API redundant and b) slow down loading the metadata too much.

michal-shalev: a) Why would it make `makeConnection` API redundant?
b) But we end up doing this anyways, this substitute users workaround of sending a notification and waiting for it to make the connection *after* using `makeConnection` API

tstamler (approved):

src/plugins/ucx/ucx_backend.cpp
tstamler: yes, my mistake. In older versions of the UCX backend, the makeConnection path would send an active message and trigger a callback on the remote host to do this forced connection establishment but it looks like we don't do that anymore

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+
+void
+nixlUcxEngine::performConnectionEstablishment(
+    const std::string &remote_agent,
+    const std::shared_ptr<nixlUcxConnection> &conn) const {
+    NIXL_DEBUG << "Establishing connection with " << remote_agent;
+
+    // Flush all endpoints to ensure connection establishment
+    // and avoid UCS_ERR_NOT_CONNECTED errors during data transfers
+    for (size_t i = 0; i < conn->eps.size(); ++i) {
```
tvegas1, line 1831: i think flush without any previous request could complete without endpoint being connected. if confirmed, we could first send dummy op, then start flush

src/plugins/ucx/ucx_backend.cpp
tvegas1, line 1833: since it is ucx engine api, i think it would be safer to start all the flush operations, then progress all flush requests in another loop, for the case where we would have inter-dependency somehow.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+            nixlUcxWorker *worker = getWorker(i).get();
+            do {
+                ret = worker->test(req);
+            } while (ret == NIXL_IN_PROG);
+
+            worker->reqRelease(req);
+        }
+
+        if (ret != NIXL_SUCCESS) {
+            NIXL_WARN << "Failed to flush endpoint " << i << " for " << remote_agent
```
tvegas1, line 1845: return error

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
     }
     return NIXL_SUCCESS;
 }
+
+void
+nixlUcxEngine::performConnectionEstablishment(
+    const std::string &remote_agent,
+    const std::shared_ptr<nixlUcxConnection> &conn) const {
```
rakhmets, line 1826:
```suggestion
    const nixlUcxConnection &conn) const {
```

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers

Or pass just endpoints.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
+    NIXL_DEBUG << "Establishing connection with " << remote_agent;
+
+    // Flush all endpoints to ensure connection establishment
+    // and avoid UCS_ERR_NOT_CONNECTED errors during data transfers
+    for (size_t i = 0; i < conn->eps.size(); ++i) {
+        nixlUcxReq req;
+        nixl_status_t ret = conn->eps[i]->flushEp(req);
+
+        if (ret == NIXL_IN_PROG) {
+            nixlUcxWorker *worker = getWorker(i).get();
```
rakhmets, line 1836: `nixlUcxWorker &` ?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1817,3 +1819,33 @@ nixl_status_t nixlUcxEngine::genNotif(const std::string &remote_agent, const std
...
 }
+
+void
+nixlUcxEngine::performConnectionEstablishment(
+    const std::string &remote_agent,
+    const std::shared_ptr<nixlUcxConnection> &conn) const {
+    NIXL_DEBUG << "Establishing connection with " << remote_agent;
+
+    // Flush all endpoints to ensure connection establishment
+    // and avoid UCS_ERR_NOT_CONNECTED errors during data transfers
```
evgeny-leksikov, line 1830: To me it looks like trying to move workaround from user level to ucx backend instead of fixing UCP API, UCP EP should not return NOT_CONNECTED to avoid blocking on any level. Instead, the request should go on pending until completion as any other operation posted on UCP EP.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1248,6 +1248,8 @@ nixl_status_t nixlUcxEngine::loadRemoteConnInfo (const std::string &remote_agent
 
     remoteConnMap.insert({remote_agent, conn});
 
+    performConnectionEstablishment(remote_agent, conn);
```
evgeny-leksikov, line 1251: handle possible failure

michal-shalev:
Decided on a different solution

PR #820, opened 2025-09-22
NIXLBENCH: Add CUDA 13 support

ovidiusm:
## What?

## Why?

## How?
architecture, approach, etc._


benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -159,45 +155,47 @@ RUN git clone --recurse-submodules --depth 1 --shallow-submodules https://github
...
-    # pybind11 pip install needed for ubuntu 22.04
-    uv pip install --upgrade meson pybind11 patchelf pyYAML click tabulate
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+COPY --from=nixl . /workspace/nixl
+COPY --from=nixlbench . /workspace/nixlbench
+
+WORKDIR /workspace/nixl
 
+# Create a new virtual environment
+ENV VIRTUAL_ENV=/workspace/nixl/.venv
```
aranadive: Can we create a separate directory in /workspace for the virtualenv? So whenever we modify things in nixl dir and build container it doesnt need to recreate and download dependencies again.
You can prefix uv commands with UV_NO_SYNC=1 to avoid redownloading packages as well.
Then the COPY commands would come after torch is installed.

ovidiusm: Good point. I moved the virtualenv installation before the source code copying, and ensured that we do all the pip install steps there.

PR #825, opened 2025-09-25
Make UCX dependency optional

ovidiusm:
## What?
Make UCX dependency optional

## Why?
Amazon would like to be able to build NIXL without UCX

### DIFF
--- a/examples/cpp/meson.build
+++ b/examples/cpp/meson.build
@@ -13,12 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-nixl_example = executable('nixl_example',
-           'nixl_example.cpp',
-           dependencies: [nixl_dep, nixl_infra, nixl_common_deps, ucx_backend_dep, ucx_dep],
-           include_directories: [nixl_inc_dirs, utils_inc_dirs],
-           link_with: [serdes_lib],
-           install: true)
+if ucx_dep.found()
+    nixl_example = executable('nixl_example',
+            'nixl_example.cpp',
+            dependencies: [nixl_dep, nixl_infra, nixl_common_deps, ucx_backend_dep, ucx_dep],
+            include_directories: [nixl_inc_dirs, utils_inc_dirs],
+            link_with: [serdes_lib],
+            install: true)
+endif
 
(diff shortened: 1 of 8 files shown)


examples/cpp/meson.build
```diff
@@ -13,17 +13,22 @@
...
+endif
+
 nixl_example = executable('nixl_example',
-           'nixl_example.cpp',
-           dependencies: [nixl_dep, nixl_infra, nixl_common_deps, ucx_backend_dep, ucx_dep],
-           include_directories: [nixl_inc_dirs, utils_inc_dirs],
-           link_with: [serdes_lib],
-           install: true)
+        'nixl_example.cpp',
+        dependencies: [nixl_dep, nixl_infra, nixl_common_deps] + backend_deps,
```
aranadive: Can we make it so that there is atleast one backend available to run nixl_example? Either ucx/libfabric?

ovidiusm: There are many backends and I would rather not add the logic here.

Moein opened https://github.com/ai-dynamo/nixl/pull/822 which will run an arbitrary plugin at runtime

ovidiusm: @mkhazraee

mkhazraee: Hmm, I say what Adit is saying is rather orthogonal to that, the idea being users won't build nixl without any network backend and then say why nothing works. But not sure if we can easily handle that here, and also since we have storage backends, they're not mandatory either, right? If so maybe fine as we can't mandate a storage backend for each type of memory either.

mkhazraee (approved):

examples/cpp/meson.build
ovidiusm: That logic does not belong in meson. It's up to the user to build the plugins they want to load in the example, and it's fine to display an error at runtime if they cannot be loaded. And in theory, the user could also load plugins coming from outside the build tree.

mkhazraee (approved):

examples/cpp/meson.build
ovidiusm: I removed backend_deps completely

PR #826, opened 2025-09-25
Refactor libfabric request management and improve rail handling

akkart-aws:
## What?

This PR introduces expandable request pools for the libfabric backend and implementing dynamic pool expansion when pools become exhausted.

## Why?

nixlbench was getting hung because it kept hitting the maximum request limit on instances that only support EFA (not EFA-Direct). This occurs because simulated writes are much slower than RDMA writes, causing request pool exhaustion when the system cannot keep up with the allocation demand. The fixed-size pools (originally 2048 requests per rail) were insufficient for sustained workloads on EFA-only instances.

## How?

- __Implemented dynamic pool expansion__: Added `expandPool()` methods in both `ControlRequestPool` and `DataRequestPool` that double the pool size when exhausted, with proper buffer chunk management for control pools
- __Reduced initial pool sizes for debugging__: Reduced the initial pool sizes to 1024 requests per rail (down from 2048) to avoid initial memory overhead.
- __Added comprehensive error handling__: Enhanced allocation failure paths with detailed logging and graceful degradation
- __Improved resource management__: Better cleanup and validation of pool indices with bounds checking

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -19,41 +19,83 @@
 #include "libfabric_rail.h"
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
+#include "libfabric_common.h"
 
 #include <cstring>
 #include <stdexcept>
+#include <stack>
 
 // RequestPool Base Class Implementation
 
-RequestPool::RequestPool(size_t pool_size, size_t rail_id) : rail_id_(rail_id) {
-    requests_.resize(pool_size);
+RequestPool::RequestPool(size_t pool_size, size_t rail_id)
+    : rail_id_(rail_id),
+      initial_pool_size_(pool_size) {
+    initializeBasePool(pool_size);
+}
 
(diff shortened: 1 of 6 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -751,6 +927,7 @@ nixlLibfabricRail::processRecvCompletion(struct fi_cq_data_entry *comp) {
     nixl_status_t status = postRecv(new_req);
     if (status != NIXL_SUCCESS) {
         NIXL_ERROR << "Failed to post subsequent receive on rail " << rail_id;
+        // Don't release request for efa provider, release for all other providers
```
yexiangd: which logic in `releaseRequest` or `RequestPool::release` is checking EFA provider?

akkart-aws: sorry this was a stale comment. I have removed it.

yexiang-aws (approved):

PR #827, opened 2025-09-25
Fix Rust query_xfer_backend API

evgeny-leksikov:
## What?
Fix Rust query_xfer_backend API

## Why?
return a string, as rest of the API doesn't return the raw pointer

## How?
 - add `nixl_capi_query_xfer_backend_type` and use it in Rust implementation
 - add C++ `Agent::getBackendType` helper API and use it in C wrapper

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -106,6 +106,15 @@ class nixlAgent {
                        const nixl_b_params_t &params,
                        nixlBackendH* &backend);
         /**
+         * @brief  Get the type of a backend
+         *
+         * @param  backend Backend handle
+         * @param  type [out] Backend type
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        getBackendType(const nixlBackendH *backend, nixl_backend_t &type) const;
+        /**
          * @brief  Register a memory/storage with NIXL. If a list of backends hints is provided
          *         (via extra_params), the registration is limited to the specified backends.
          *
--- a/src/bindings/rust/src/agent.rs
+++ b/src/bindings/rust/src/agent.rs
@@ -931,23 +931,45 @@ impl Agent {
     /// * `req` - Transfer request handle after `post_xfer_req`
     ///
     /// # Returns
-    /// A handle to the backend used for the transfer
+    /// Name of the backend used for the transfer
     ///
     /// # Errors
     /// Returns a NixlError if the operation fails
-    pub fn query_xfer_backend(&self, req: &XferRequest) -> Result<Backend, NixlError> {
-        let mut backend = std::ptr::null_mut();
+    pub fn query_xfer_backend(&self, req: &XferRequest) -> Result<String, NixlError> {
+        let mut backend_type_ptr: *mut std::ffi::c_void = std::ptr::null_mut();
+        let mut backend_type_size: usize = 0;
+
         let inner_guard = self.inner.write().unwrap();
         let status = unsafe {
-            nixl_capi_query_xfer_backend(
+            nixl_capi_query_xfer_backend_type(
                 inner_guard.handle.as_ptr(),
                 req.handle(),
-                &mut backend
+                &mut backend_type_ptr,
+                &mut backend_type_size,
             )
         };
         match status {
             NIXL_CAPI_SUCCESS => {
-                Ok(Backend{ inner: NonNull::new(backend).ok_or(NixlError::FailedToCreateBackend)? })
+                if backend_type_ptr.is_null() || backend_type_size == 0 {
+                    return Err(NixlError::BackendError);
+                }
+
+                // Extract the backend type string from the binary data
+                let backend_type = unsafe {
+                    let slice = std::slice::from_raw_parts(
+                        backend_type_ptr as *const u8,
+                        backend_type_size
+                    );
+                    // Handle potential embedded nulls and convert to String
+                    String::from_utf8_lossy(slice).to_string()
+                };
+
+                // Verify this backend type exists in our backends map
+                if inner_guard.backends.contains_key(&backend_type) {
+                    Ok(backend_type)
+                } else {
+                    Err(NixlError::BackendError)
+                }
             }
             NIXL_CAPI_ERROR_INVALID_PARAM => Err(NixlError::InvalidParam),
             _ => Err(NixlError::BackendError),
--- a/src/bindings/rust/tests/tests.rs
+++ b/src/bindings/rust/tests/tests.rs
@@ -1324,10 +1324,10 @@ fn test_query_xfer_backend_success() {
             None
         ).expect("Failed to create transfer request");
         // Query which backend will be used for this transfer
-        let result: Result<Backend, NixlError> = agent1.query_xfer_backend(&xfer_req);
+        let result: Result<String, NixlError> = agent1.query_xfer_backend(&xfer_req);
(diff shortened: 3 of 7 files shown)


src/api/cpp/nixl.h
mkhazraee, line 1: We should not add APIs to nixl just for this. I don't understand why in our cpp wrapper we can't call getType on the nixlBackendH.

evgeny-leksikov: Because nixl API defines `nixlBackendH` as an opaque type.
Dependency of the wrapper on internal `core/agent_data.h` may help here but IMO "get backend type" public API looks useful.

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,15 @@ class nixlAgent {
...
                        nixlBackendH* &backend);
+        /**
+         * @brief  Get the type of a backend
+         *
+         * @param  backend Backend handle
+         * @param  type [out] Backend type
+         * @return nixl_status_t Error code if call was not successful
+         */
+        nixl_status_t
+        getBackendType(const nixlBackendH *backend, nixl_backend_t &type) const;
```
gleon99, line 116: Line break

src/api/cpp/nixl.h
```diff
@@ -105,6 +105,15 @@ class nixlAgent {
         createBackend (const nixl_backend_t &type,
                        const nixl_b_params_t &params,
                        nixlBackendH* &backend);
+        /**
+         * @brief  Get the type of a backend
+         *
+         * @param  backend Backend handle
+         * @param  type [out] Backend type
```
gleon99, line 112: Alignment

src/api/cpp/nixl.h
gleon99: I think @evgeny-leksikov  is right here, I don't see how it can be done w/o adding a new API, because of the opaque handle type.

src/core/nixl_agent.cpp
```diff
@@ -396,6 +396,22 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
 }
 
+nixl_status_t
+nixlAgent::getBackendType(const nixlBackendH *backend, nixl_backend_t &type) const {
+    if (!backend) {
+        NIXL_ERROR_FUNC << "backend handle is not provided";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    try {
```
gleon99, line 406: Let's remove the `try`.
It is doesn't exist anywhere else in this file.
And it is never used for `getType()`..

src/bindings/rust/wrapper.h
```diff
@@ -243,6 +243,12 @@ nixl_capi_query_xfer_backend(nixl_capi_agent_t agent,
                              nixl_capi_xfer_req_t req_hndl,
                              nixl_capi_backend_t *backend);
 
+nixl_capi_status_t
+nixl_capi_query_xfer_backend_type(nixl_capi_agent_t agent,
```
gleon99, line 247: `query` -> `get`?

src/bindings/rust/wrapper.cpp
```diff
@@ -1507,6 +1507,32 @@ nixl_capi_query_xfer_backend(nixl_capi_agent_t agent,
...
+    }
+
+    nixl_capi_backend_t backend_handle;
+    auto ret = nixl_capi_query_xfer_backend(agent, req_hndl, &backend_handle);
+    if (ret != NIXL_CAPI_SUCCESS) {
+        return ret;
+    }
+
+    static thread_local nixl_backend_t type;
+    auto nixl_ret = agent->inner->getBackendType(backend_handle->backend, type);
```
gleon99, line 1526: `try { ...` here

src/bindings/rust/wrapper.cpp
```diff
@@ -1507,6 +1507,32 @@ nixl_capi_query_xfer_backend(nixl_capi_agent_t agent,
...
+        return NIXL_CAPI_ERROR_INVALID_PARAM;
+    }
+
+    nixl_capi_backend_t backend_handle;
+    auto ret = nixl_capi_query_xfer_backend(agent, req_hndl, &backend_handle);
+    if (ret != NIXL_CAPI_SUCCESS) {
+        return ret;
+    }
+
+    static thread_local nixl_backend_t type;
```
gleon99, line 1525: Is `static thread_local` really needed? It doesn't appear anywhere else in this file..

src/bindings/rust/tests/tests.rs
```diff
@@ -1324,10 +1324,10 @@ fn test_query_xfer_backend_success() {
             None
         ).expect("Failed to create transfer request");
         // Query which backend will be used for this transfer
-        let result: Result<Backend, NixlError> = agent1.query_xfer_backend(&xfer_req);
+        let result: Result<String, NixlError> = agent1.query_xfer_backend(&xfer_req);
         assert!(result.is_ok(), "query_xfer_backend failed with error: {:?}", result.err());
-        let backend = result.unwrap();
-        println!("Transfer will use backend: {:?}", backend);
+        let backend_name = result.unwrap();
```
gleon99, line 1329: `name` -> `type`? (consistency)

src/bindings/rust/src/agent.rs
```diff
@@ -931,23 +931,45 @@ impl Agent {
     /// * `req` - Transfer request handle after `post_xfer_req`
     ///
     /// # Returns
-    /// A handle to the backend used for the transfer
+    /// Name of the backend used for the transfer
```
gleon99, line 934: `Type`?

src/bindings/rust/src/agent.rs
```diff
@@ -931,23 +931,45 @@ impl Agent {
...
         match status {
             NIXL_CAPI_SUCCESS => {
-                Ok(Backend{ inner: NonNull::new(backend).ok_or(NixlError::FailedToCreateBackend)? })
+                if backend_type_ptr.is_null() || backend_type_size == 0 {
+                    return Err(NixlError::BackendError);
+                }
+
+                // Extract the backend type string from the binary data
+                let backend_type = unsafe {
+                    let slice = std::slice::from_raw_parts(
```
gleon99, line 959: Maybe
```
CStr::from_ptr(backend_type_ptr as *const 8)
                        .to_string_lossy()
                        .into_owned()
```?

src/api/cpp/nixl.h
mkhazraee: Ok, let's ignore this one API then for now, as it's not used by users we know of as of today. Adding an API to C++ and Python just to handle Rust doesn't sound the best, while they can easily handle it. Also we might do a large refactor and add methods to the xferReqH object itself, will see, and we can decide on it then.

src/bindings/rust/wrapper.cpp
evgeny-leksikov: it's needed to avoid memory management in rust layer and thread safety in wrapper

src/bindings/rust/src/agent.rs
evgeny-leksikov: c-str may lose data because it's null-teminated. Need to keep c++ like string with length

src/api/cpp/nixl.h
gleon99: @mkhazraee do you have any users of this API?
Otherwise we'll close this PR..

mkhazraee: Not as of now specifically, if we have to add API to main agent for this, it's not wroth it. We can circle back to this when doing the minor API revisioning. Closing the PR.

PR #828, opened 2025-09-26
RUST: RegDescList and XferDescList APIs

evgeny-leksikov:
## What?
RegDescList and XferDescList missed APIs:
- equality operators
- set/get
- get index

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -31,6 +31,6 @@ CODEOWNERS @ai-dynamo/Devops @ai-dynamo/nixl-maintainers
 /benchmark @aranadive @ovidiusm @brminich
 
 # Libfabric Plugins
-/src/plugins/libfabric* @yexiang-aws @akkart-aws
-/src/utils/libfabric @yexiang-aws @akkart-aws
-/test/unit/utils/libfabric @yexiang-aws @akkart-aws
+/src/plugins/libfabric* @yexiang-aws @akkart-aws @fengjica
+/src/utils/libfabric @yexiang-aws @akkart-aws @fengjica
+/test/unit/utils/libfabric @yexiang-aws @akkart-aws @fengjica
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -825,22 +825,30 @@ nixlLibfabricEngine::getPublicData(const nixlBackendMD *meta, std::string &str) 
 }
 
 nixl_status_t
-nixlLibfabricEngine::loadLocalMD(nixlBackendMD *input, nixlBackendMD *&output) {
-    nixlLibfabricPrivateMetadata *input_md = static_cast<nixlLibfabricPrivateMetadata *>(input);
+nixlLibfabricEngine::loadMetadataHelper(const std::vector<uint64_t> &rail_keys,
+                                        void *buffer,
+                                        std::shared_ptr<nixlLibfabricConnection> conn,
+                                        nixlBackendMD *&output) {
     auto pub_md = std::make_unique<nixlLibfabricPublicMetadata>();
-    // Store all rail keys instead of just the first one
-    pub_md->rail_remote_key_list_.reserve(input_md->rail_key_list_.size());
-    for (size_t rail_id = 0; rail_id < input_md->rail_key_list_.size(); ++rail_id) {
-        pub_md->rail_remote_key_list_.push_back(input_md->rail_key_list_[rail_id]);
-        NIXL_DEBUG << "Added rail " << rail_id << " key: " << input_md->rail_key_list_[rail_id];
-    }
 
-    pub_md->remote_buf_addr_ = reinterpret_cast<uint64_t>(input_md->buffer_);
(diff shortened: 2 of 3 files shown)


src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -14,11 +14,26 @@
...
+    pub dev_id: u64,
+    pub metadata: Vec<u8>,
+}
 
 /// A safe wrapper around a NIXL registration descriptor list
 pub struct RegDescList<'a> {
     inner: NonNull<bindings::nixl_capi_reg_dlist_s>,
     _phantom: PhantomData<&'a dyn NixlDescriptor>,
+    // Track descriptors internally for comparison purposes
+    descriptors: Vec<RegDescriptor>,
```
gleon99: Why do we need to maintain duplicate storage, with all the implications?
Instead of simply exposing operators from C++...
Am I missing something?
[patch.txt](https://github.com/user-attachments/files/22584916/patch.txt)

evgeny-leksikov: Index and IndexMut

roiedanino: It looks like a nice optimization, but I'm afraid some of those APIs may have (or will have in the future) some side effects and checks that we are skipping that way, IMO RegDescList should call the bindings directly, and maybe have another struct called "CachedRegDescList" or make those implementations interchangeable with a configuration flag, making it less error prone, giving a simple workaround in case some incosistencies happen on users end because of this.

evgeny-leksikov: I added `sync_mgr` to automate consistency and avoid accidental mutation in future changes

src/bindings/rust/src/descriptors/sync_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn new(data: T, backend: T::Backend) -> Self {
+        Self {
+            data,
+            backend,
+            dirty: Cell::new(true),
+        }
+    }
+
+    /// Mutates the frontend data (marks as dirty)
+    pub fn mutate<F, R>(&mut self, f: F) -> R
```
roiedanino: Minor: Maybe we should rename this function to `modify`, as semantically `mutate` may be associated mistakenly with a modification of a replica/copy.

Although we can think of it as a copy of the C++ object, it may not be clear to the next developer working on this

src/bindings/rust/tests/tests.rs
```diff
@@ -1585,3 +1585,135 @@ fn test_get_xfer_telemetry_before_posting() {
...
+
+        xfer_list9.add_desc(0x1000, 0x100, 0).unwrap();
+        xfer_list9.add_desc(0x2000, 0x200, 1).unwrap();
+
+        xfer_list10.add_desc(0x2000, 0x200, 1).unwrap();
+        xfer_list10.add_desc(0x1000, 0x100, 0).unwrap();
+
+        assert_ne!(xfer_list9, xfer_list10);
+        assert!(!(xfer_list9 == xfer_list10));
+    }
```
roiedanino, line 1718: This is a separate test; it would be better to move it to a separate test function

src/bindings/rust/tests/tests.rs
```diff
@@ -1585,3 +1585,135 @@ fn test_get_xfer_telemetry_before_posting() {
...
+
+        // 3. Lists with same descriptors should be equal
+        let mut reg_list3 = RegDescList::new(MemType::Dram).unwrap();
+        let mut reg_list4 = RegDescList::new(MemType::Dram).unwrap();
+
+        reg_list3.add_desc(0x1000, 0x100, 0).unwrap();
+        reg_list3.add_desc(0x2000, 0x200, 1).unwrap();
+
+        reg_list4.add_desc(0x1000, 0x100, 0).unwrap();
+        reg_list4.add_desc(0x2000, 0x200, 1).unwrap();
```
roiedanino, line 1613: It feels like 1, 2, 3, 4, 5, 6, 7 can each be a test on its own. It would also be much easier to look at and understand what failed once one of those scenarios fails.

Also, there are a few lines here and there that look like code duplication that we can avoid

src/bindings/rust/tests/tests.rs
```diff
@@ -1585,3 +1585,135 @@ fn test_get_xfer_telemetry_before_posting() {
         println!("Successfully tested telemetry before posting - got expected error");
     }
 }
+
+// Tests for equality operators on RegDescList and XferDescList
+#[test]
+fn test_descriptor_list_equality() {
```
roiedanino, line 1591: Maybe we should add some tests that trigger the synchronization manager and make sure it works properly. Maybe we can do it directly by creating a class that implements the trait just to assert that the synchronization happened

src/bindings/rust/src/descriptors/sync_manager.rs
evgeny-leksikov: Fixed

evgeny-leksikov:
@roiedanino I would address all test related comments in separate PR, this one is close to hard limit in 500 lines.

PR #829, opened 2025-09-26
RUST: RegDescList and XferDescList serialization 

evgeny-leksikov:
## What?
Rust RegDescList and XferDescList serialization 

NOTE: in dependency from https://github.com/ai-dynamo/nixl/pull/828


src/bindings/rust/src/descriptors.rs
```diff
@@ -20,9 +20,50 @@ mod reg;
 mod xfer;
 mod xfer_dlist_handle;
 
+mod codec {
```
gleon99: Can we use smth like `serde` instead?

src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -207,11 +181,159 @@ impl<'a> RegDescList<'a> {
...
+        Ok(new_list)
+    }
+
+    /// Serialize the descriptor list to a byte vector using custom binary format
+    pub fn serialize(&self) -> Result<Vec<u8>, NixlError> {
+        // Precompute capacity: 1 (mem_type) + 8 (count) + per-desc (24 + 8 + meta.len())
+        let mut total_size = 1usize + 8usize;
+        for d in &self.descriptors {
+            total_size = total_size
+                .saturating_add(24) // addr,len,dev_id
```
gleon99: Can avoid hardcoding?

src/bindings/rust/tests/tests.rs
```diff
@@ -1717,3 +1717,188 @@ fn test_descriptor_list_equality() {
...
+        ($list_type:ty, $type_name:expr) => {{
+            let empty_list = <$list_type>::new(MemType::Dram).unwrap();
+            let serialized = empty_list.serialize().unwrap();
+            assert!(!serialized.is_empty(), "Serialized empty {} should contain metadata", $type_name);
+            println!("Empty {} serialized to {} bytes", $type_name, serialized.len());
+        }};
+    }
+
+    test_empty!(RegDescList, "RegDescList");
+    test_empty!(XferDescList, "XferDescList");
```
roiedanino: Consider using std::any::type_name::<$list_type>() instead of passing 2 parameters:

```suggestion
macro_rules! test_empty {
    ($list_type:ty) => {
        let empty_list = <$list_type>::new(MemType::Dram).unwrap();
        let serialized = empty_list.serialize().unwrap();
        assert!(
            !serialized.is_empty(),
            "Serialized empty {} should contain metadata",
            std::any::type_name::<$list_type>()
        );
        println!(
            "Empty {} serialized to {} bytes",
            std::any::type_name::<$list_type>(),
            serialized.len()
        );
    };
}

test_empty!(RegDescList);
test_empty!(XferDescList);
```

We could also use the + syntax in macro_rules, but I think it's less readable and an overkill here

src/bindings/rust/tests/tests.rs
```diff
@@ -1717,3 +1717,188 @@ fn test_descriptor_list_equality() {
...
+}
+
+// Test: List with descriptors serializes to larger size than empty list
```
roiedanino, line 1937: Can we add test cases using `create_storage_list()`? which is an important use case

src/bindings/rust/tests/tests.rs
evgeny-leksikov: fixed

src/bindings/rust/tests/tests.rs
evgeny-leksikov: done

evgeny-leksikov:
@gleon99 @roiedanino pls take a look, your approvals were dismissed after resolving conflicts

PR #830, opened 2025-09-26
Flush out commQueue before stopping listener

2dm:
## What?
Verify that all enqueued work has completed. 

## Why?
There is a race between the agent destructor and jobs on `commQueue` since the background thread might finish without checking for existing work in the queue. In particular, the missed job is probably `invalidateLocalMD` that is called right before, which will keep the MD intact and prevent the re-addition of that rank later. 

## How?
1. Stop enqueueing
2. Check queue is empty (the thread is processing or completed the last work that was enqueued)
3. Only then set `commThreadStop`


src/core/agent_data.h
```diff
@@ -50,94 +50,118 @@ enum nixl_comm_t {
...
+        remoteSections;
+
+    // State/methods for listener thread
+    nixlMDStreamListener *listener;
+    std::map<nixl_socket_peer_t, int> remoteSockets;
+    std::thread commThread;
+    std::vector<nixl_comm_req_t> commQueue;
+    std::mutex commLock;
+    bool commThreadStop;
+    bool agentShutdown;
```
mkhazraee: This line is the only new line in this file. The rest are clang format. We can bite the bullet and let this clang format be done, or do just a few surrounding lines.

2dm: @mkhazraee I've returned to the original change and added a commit to fix format of only the lines surrounding the new agentShutdown. It's a smaller PR now and the changes are more clear. clang-format test passed.

src/core/nixl_agent.cpp
```diff
@@ -194,13 +194,19 @@ nixlAgent::nixlAgent(const std::string &name, const nixlAgentConfig &cfg) :
...
+        data->agentShutdown = false;
         data->commThread =
             std::thread(&nixlAgentData::commWorker, data.get(), this);
     }
 }
 
 nixlAgent::~nixlAgent() {
     if (data && (data->useEtcd || data->config.useListenThread)) {
+        data->agentShutdown = true;
+        while (!data->commQueue.empty()) {
```
aranadive, line 205: What is the purpose for this while loop? We clear out the commQueue after copying all elements out in getCommWork. Shouldnt we wait till all those requests are processed, otherwise we could cause issues for other agents?

2dm: Yes, that's the point. Any work that is added from the time of `getCommWork` until the `commThreadStop` is evaluated in the following iteration on nixl_listener won't be handled. This PR stops getting new job once destructor begins, and sets `commThreadStop` only when all the work that was posted is done.

src/core/nixl_listener.cpp
```diff
@@ -662,6 +662,10 @@ void nixlAgentData::commWorker(nixlAgent* myAgent){
 }
 
 void nixlAgentData::enqueueCommWork(nixl_comm_req_t request){
+    if (agentShutdown) {
+        NIXL_WARN << "Agent shutting down, rejecting new work";
```
aranadive:
```suggestion
        NIXL_WARN << "Agent shutting down, unable to accept new requests";
```

src/core/nixl_agent.cpp
aranadive: > Yes, that's the point. Any work that is added from the time of `getCommWork` until the `commThreadStop` is evaluated in the following iteration on nixl_listener won't be handled. This PR stops getting new job once destructor begins, and sets `commThreadStop` only when all the work that was posted is done.

I mean the work isnt done until the thread joins which is after commThreadStop is set to true. But I see what its doing now. The listener thread will always complete stuff it pulled out of the commQueue. And the commQueue will be empty once the listener thread has joined.

aranadive (approved):

src/core/nixl_agent.cpp
```diff
@@ -194,13 +194,19 @@ nixlAgent::nixlAgent(const std::string &name, const nixlAgentConfig &cfg) :
...
         data->commThreadStop = false;
+        data->agentShutdown = false;
         data->commThread =
             std::thread(&nixlAgentData::commWorker, data.get(), this);
     }
 }
 
 nixlAgent::~nixlAgent() {
     if (data && (data->useEtcd || data->config.useListenThread)) {
+        data->agentShutdown = true;
```
ovidiusm, line 204: IIUC agentShutdown is accessed from different threads, it should be atomic

ovidiusm:
I still see 2 tests failing with SIGPIPE during cleanup, so I would try to stress test this locally to make sure there is no issue.

src/core/nixl_agent.cpp
2dm: @ovidiusm Same for commThreadStop then?

2dm: @ovidiusm - addressed in new commit

PR #832, opened 2025-09-28
CI: Add EFA security group validation and pre-flight checks

Alexey-Rivkin:
AWS EFA requires self-referencing security group rules for inter-node communication. Add verification commands to README and integrate fi_pingpong validation into aws_test.sh to fail fast if EFA is misconfigured.

## What?
- Integrate fi_pingpong pre-flight check into aws_test.sh.
- Add EFA security group validation commands to README.

## Why?
AWS reported that EFA requires specific security group configurations with self-referencing rules for inter-node communication. Current CI only uses a single node, yet the config might still be needed.

## How?
- Integrate `fi_pingpong -p efa` validation into aws_test.sh after build but before tests to fail fast if EFA is not working
- Add verification commands to README that check for required EFA security group rules (self-referencing inbound/outbound + SSH access)
- Commands are read-only for safe troubleshooting without infrastructure modifications


contrib/aws-efa/aws_test.sh
```diff
@@ -64,14 +64,15 @@ setup_cmd="set -x && \
     git clone ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY} && \
     cd nixl && \
     ${GIT_CHECKOUT_CMD}"
+efa_validation_cmd="fi_info -p efa"
```
brminich, line 67: what does this check do? Will it fail CI if config is not correct?

Alexey-Rivkin: `fi_info -p efa` checks local EFA provider availability (driver/hardware). 
CI will fail if EFA is not available.

Local test with no EFA fabric:
```
[arivkin@arivkin-mlt /]$ fi_info -p efa
fi_getinfo: -96 (No message available on STREAM)
[arivkin@arivkin-mlt /]$ echo $?
96
```

mkhazraee (approved):

PR #835, opened 2025-09-29
Add a change

ovidiusm:
## What?

## Why?

## How?
architecture, approach, etc._

### DIFF
--- a/remove-me.txt
+++ b/remove-me.txt
@@ -1,0 +1,1 @@
+random change


.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -48,6 +48,7 @@ env:
   NIXL_INSTALL_DIR: /opt/nixl
   TEST_TIMEOUT: 30
   NPROC: "16"
+  UCX_TLS: "^mm,shm"
```
brminich: sm or shm is enough it also includes mm

brminich (dismissed):
lets merge it to enable CI and continue debugging

.ci/jenkins/lib/build-matrix.yaml
ovidiusm: Ok I do another run with shm

brminich (approved):

saturley-hall (approved):

PR #853, opened 2025-10-02
Add New VRAM Example

bloodeagle40234:
This commit provides a Python example for VRAM (GPU memory) transfer. The block-based transfer model is inspired by vLLM and demonstrates transfer bandwidth under specific configurations.

## What?

This commit adds a new Python example for VRAM (GPU memory) transfer. The example follows vLLM's block-based memory management and shows how to configure NIXL using the `make_prepped_xfer` API.

## Why?

The current NIXL upstream does not include a VRAM-based example, which may make it unclear for users how to configure and debug data transfers between GPUs. To address this, this PR provides a working example. It also demonstrates index-based transfer, as used in vLLM, along with bandwidth performance results. This helps users understand KV cache transfer in disaggregated inference scenarios.

## How?

This PR includes three components and one documentation file. The example is based on a server-client model: the client reads specific tensor blocks from the running server and outputs performance results to the console.

### DIFF
--- a/examples/python/vram_example/README.md
+++ b/examples/python/vram_example/README.md
@@ -1,0 +1,83 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2025 Kota Tsuyuzaki <bloodeagle40234@gmail.com>
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# NIXL VRAM Transfer Example
+
+This is an example of VRAM transfer using NIXL, inspired by the [vLLM](https://github.com/vllm-project/vllm) 0.10.0 KV cache transfer algorithm. It demonstrates a basic server-client model for transferring VRAM memory. The server process allocates tensor memory filled with ones, and the client copies these tensors into locally allocated tensors on the client side.
+
+## Memory Alignment
+
+As described above, this example follows vLLM's KV cache memory management. The transfer unit is controlled by a specific memory block size within each tensor.
+
+The memory layout is designed as follows:
+
+- Let **N** be the number of layers. The process creates **N** tensors.
+- Each tensor maintains contiguous memory alignment but is logically divided into blocks of `block_size` tokens.
+- The memory size of each block is calculated using the following formula:
+
+`KV * Heads * Dimensions * Tokens per block * Precision`
+
+For example, considerating [Mistral-Small-3.1](https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/tree/main) for default vLLM config, with:
+
+- N (Attention layers) = 40
+- KV = 2
+- Heads = 8
+- Dimensions = 128
+- Tokens per block = 16
+- Precision = bf16 (2 bytes)
+
+The block size in bytes is:
+
+`2 * 8 * 128 * 16 * 2 = 65536 (64KB)`
+
+The memory alignment looks like this:
+
+```
+tensor 0  [(64KB)|(64KB)|(64KB)| ... |(64KB)]
+tensor 1  [(64KB)|(64KB)|(64KB)| ... |(64KB)]
+...
+tensor 39 [(64KB)|(64KB)|(64KB)| ... |(64KB)]
+```
+
+If 4GB of VRAM is available for Mistral-Small-3.1, the number of blocks per tensor is calculated as:
+
+`((4 * 1024 * 1024 * 1024) / (64 * 1024)) // 40 = 1638 (blocks)`
+
+
+## Transfer Model
+
+This example uses the memory alignment described above. NIXL has a capability to manage memory blocks using indices. Following vLLM's approach, the process first reserves blocks for KV cache. These reserved blocks maintain consistent indices across all layer tensors. Therefore, this example transfers data from the head block up to the number of blocks required for the given input tokens.
+
+## Usage
+Start the server process to wait for incoming requests:
+
+```
+python server.py
+```
+
+Then, launch the client process:
+
+```
+python client.py
+```
+
+Note that, the server process will be running with while-loop so use `Ctrl-C` or kill the process directly to terminate server process at the end.
+
+If you want to use a different GPU for each process, pass `CUDA_VISIBLE_DEVICES` environment variable to pin the GPU index.
+
+And the variables of the tranfer model can be configured by the argument. Please use `--help` to confirm available arguments.
+
+
+> [!NOTE]
+> The server runs in a `while` loop. Use `Ctrl-C` or terminate the process manually to stop the server.
+
+> [!TIP]
+> To use different GPUs for each process, set the `CUDA_VISIBLE_DEVICES` environment variable to specify the GPU index.
+
+You can configure the transfer model parameters using command-line arguments. Use `--help` to view all available options.
+
+> [!NOTE]
+> The parameters must be identical between the server and client processes to ensure successful transfer.
+
--- a/examples/python/vram_example/README.md
+++ b/examples/python/vram_example/README.md
@@ -13,8 +13,9 @@ As described above, this example follows vLLM's KV cache memory management. The 
 
 The memory layout is designed as follows:
 
-- Let **N** be the number of layers. The process creates **N** tensors.
-- Each tensor maintains contiguous memory alignment but is logically divided into blocks of `block_size` tokens.
+- The process creates single contiguous.
+- Let **N** be the number of layers. The process emurates the single tensor as **N** tensors.
+- Each layer tensor block maintains memory alignment but is logically divided into blocks of `block_size` tokens.
 - The memory size of each block is calculated using the following formula:
 
 `KV * Heads * Dimensions * Tokens per block * Precision`
--- a/examples/python/vram_example/README.md
+++ b/examples/python/vram_example/README.md
@@ -1,5 +1,5 @@
 <!--
-SPDX-FileCopyrightText: Copyright (c) 2025 Kota Tsuyuzaki <bloodeagle40234@gmail.com>
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 SPDX-License-Identifier: Apache-2.0
 -->
 
--- a/examples/python/vram_example/client.py
+++ b/examples/python/vram_example/client.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# SPDX-FileCopyrightText: Copyright (c) 2025 Kota Tsuyuzaki <bloodeagle40234@gmail.com>
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/examples/python/vram_example/client.py
+++ b/examples/python/vram_example/client.py
@@ -119,22 +119,18 @@ def main():
     agent = nixl_agent("client", config)
 
     # Allocate memory and register with NIXL
-    tensors = [
-        torch.zeros(tensor_size, dtype=torch.bfloat16, device=device)
-        for x in range(args.layers)
-    ]
+    tensors = torch.zeros(
+        tensor_size * args.layers, dtype=torch.bfloat16, device=device
+    )
 
-    size_in_bytes = tensors[0].nelement() * tensors[0].element_size() * len(tensors)
+    # size_in_bytes = tensors[0].nelement() * tensors[0].element_size() * len(tensors)
+    size_in_bytes = tensors.nelement() * tensors.element_size()
     logger.info("Client Tensors in MB: %d", size_in_bytes / 1024 / 1024)
 
-    block_len = shape_len * tensors[0].element_size()  # bytes of tensor
+    block_len = shape_len * tensors.element_size()  # bytes of tensor
     logger.debug("block_len: %d", block_len)
 
-    reg_addrs = []
-    for t in tensors:
-        reg_addrs.append((t[0].data_ptr(), tensor_size * t.element_size(), 0, ""))
-
-    reg_descs = agent.get_reg_descs(reg_addrs, "VRAM")
+    reg_descs = agent.get_reg_descs(tensors, "VRAM")
 
     success = agent.register_memory(reg_descs)
     if not success:  # Same as reg_descs if successful
@@ -143,18 +139,17 @@ def main():
 
     # Create data block chunk to emulate vllm 0.10.0 data transfer
     xfer_addrs = []
-    for t in tensors:
-        base_addr = t.data_ptr()
-        for block_id in range(num_blocks):
-            offset = block_id * block_len
-            addr = base_addr + offset
-            xfer_addrs.append((addr, block_len, 0))
+    base_addr = tensors[0].data_ptr()
+    for block_id in range(num_blocks * args.layers):
+        offset = block_id * block_len
+        addr = base_addr + offset
+        xfer_addrs.append((addr, block_len, 0))
 
     logger.info(
-        "addrs info: layers: %d, elements: %d, shape: %d, block_len: %d"
+        "addrs info: layers: %d, elements: %d, shape: %d, block_len: %d, "
         "addrs_len: %d",
-        len(tensors),
-        tensors[0].nelement(),
+        args.layers,
+        tensor_size,
         shape_len,
         block_len,
         len(xfer_addrs),
@@ -204,10 +199,12 @@ def main():
         agent.send_notif("server", msg.encode())
 
     # Verify data after read.
-    for i, tensor in enumerate(tensors):
+    # for i, tensor in enumerate(tensors):
+    for i in range(args.layers):
+        layer_base = i * tensor_size
         check_blocks = trans_blocks * shape_len
         if not torch.allclose(
-            tensor[:check_blocks],
+            tensors[layer_base : layer_base + check_blocks],
             torch.ones(check_blocks, dtype=torch.bfloat16, device=device),
         ):
             logger.error("Data verification failed for tensor %d.", i)
--- a/examples/python/vram_example/client.py
+++ b/examples/python/vram_example/client.py
@@ -36,11 +36,6 @@ def transfer(
     layers,
     num_blocks,
 ):
-    # Ensure remote metadata has arrived.
-    ready = False
-    while not ready:
-        ready = agent.check_remote_metadata("server")
-
     # Handshake to server.
     agent.send_notif("server", "SYN")
 
@@ -57,11 +52,6 @@ def transfer(
 
     assert local_prep_handle != 0
     assert remote_prep_handle != 0
-
-    # Ensure remote metadata, again.
-    ready = False
-    while not ready:
-        ready = agent.check_remote_metadata("server")
 
     start = time.monotonic()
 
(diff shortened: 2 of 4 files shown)


bloodeagle40234:
For reviewers:

To pass the failed copyright check, I added commit bcba809044bef49ef415416c74b3f058f21355d1, which modifies the copyright notice from my own to the NVIDIA template.

However, I'm not sure if it's appropriate to use the NVIDIA copyright template, as I'm not an NVIDIA employee. If it's acceptable to update the check to allow more general copyright notices, including non-NVIDIA ones. In that case, I can remove the above commit.

mkhazraee:
Hi Kota,

Thanks a lot for your contribution, this is really valuable. I have few comments and noticed some issues, let's clarify/improve them before we proceed with merging the PR:

- I see you create the tensors, then per tensor get the address and len from the tensor to create the descriptor. Why don't you create a tensor with 1 more dimension, register that full tensor, and then for xfer_descs use the slices of that tensor? NIXL API already supports tensors directly. Also if your registration is back to back, better to do it once (iirc similar to vllm as they have a big pool of KV blocks). 

- In client code, you call `agent.fetch_remote_metadata("server", args.ip, args.port)` in main, without verifying the metadata was received, instead you do `agent.check_remote_metadata("server")` in each transfer, and actually do a double check there. As long as you don't call `remote_remote_agent` the metadata will not be removed, which you never do. So I suggest to add that check in the main loop instead of each transfer, could be few operations later before the transfer calls.

- I see you practically are doing blocking operations, without always the fixed notification of "UUID", and after the blocking transfer is done, you send another notification, which can be "KEEPALIVE" or "COMPLETE". So it would make more sense to use that msg for the notification of the transfer itself, which would be automatically sent when the notification is complete, as opposed to wait for it to finish and then send a second notification.

- I see you're doing blocking transfers one after another. One great feature of nixl is all of the APIs are non-blocking. And you can start all the transfer together. So I think being closer to what vLLM also does, you can start all the transfers, add the handles into a list, and remove them from the list when they're complete (again similar to vLLM implementation). When they're all done, then you can use the send_msg to indicate all of them are complete. I also suggest putting something more informative in the notifications instead of keepalive or complete, let's say the block ID. That being said, there is one point to consider here, we don't provide guarantees between transfer ordering. So you should check all of them are done, before sending that final notification. Or if you really don't care, you can skip assigning any notifications to the transfers, and just have one at the end, after checking all the transfers where complete.

- One point of having multiple transfers per agent in vLLM is they go to different agents, but here all of them go to the same agent. So you could've made a single larger transfer, and the underlying backend might have been optimized that transfer in a better way. It's good to have this as an example, but let's clarify that with a comment.

I have some more nitpicking comments, such as some syntax (like VRAM is not needed if xfer_desc_list is passed) or the name of the example being VRAM example is downplaying it as it has more important capabilities (also examples like blocking_send_recv.py is also VRAM already).

Moein

bloodeagle40234:
Thanks Moein (@mkhazraee ) for your valuable comments and for taking a deep look at my code.
Before I start making changes, I’d like to make sure I fully understand your suggestions.
I’ve added responses to each of your comments, so I’d appreciate it if you could check whether my understanding aligns with your intent.

> - I see you create the tensors, then per tensor get the address and len from the tensor to create the descriptor. Why don't you create a tensor with 1 more dimension, register that full tensor, and then for xfer_descs use the slices of that tensor? NIXL API already supports tensors directly. Also if your registration is back to back, better to do it once (iirc similar to vllm as they have a big pool of KV blocks).

Perhaps I’m misunderstanding your suggestion regarding the NIXL API. For reference, I reviewed [the vLLM KV cache allocation code](https://github.com/vllm-project/vllm/blob/v0.10.0/vllm/v1/worker/gpu_model_runner.py#L2583-L2588), and from that, I assumed vLLM allocates multiple tensors—each corresponding to a layer. However, I may also be missing your point about how tensors are actually structured, such as whether you meant “one more dimension for multiple tensors” or “a single-dimensional tensor with slices,” or something else.
I agree with your point that vLLM manages a large pool of KV blocks. That said, in the example I referred to, the blocks are also managed across multiple tensors, with a single index list used by make_prepped_xfer. If you could show me an example or document reference on your point, it could be helpful for me to get the idea.

> - In client code, you call agent.fetch_remote_metadata("server", args.ip, args.port) in main, without verifying the metadata was received, instead you do agent.check_remote_metadata("server") in each transfer, and actually do a double check there. As long as you don't call remote_remote_agent the metadata will not be removed, which you never do. So I suggest to add that check in the main loop instead of each transfer, could be few operations later before the transfer calls.

OK, this was due to my misunderstanding of the NIXL API.
It seems preferable to call check_remote_metadata only once after fetch_remote_metadata, rather than calling check_remote_metadata repeatedly in a loop.

> - I see you practically are doing blocking operations, without always the fixed notification of "UUID", and after the blocking transfer is done, you send another notification, which can be "KEEPALIVE" or "COMPLETE". So it would make more sense to use that msg for the notification of the transfer itself, which would be automatically sent when the notification is complete, as opposed to wait for it to finish and then send a second notification.

Ah, does this mean we can use "KEEPALIVE" or "COMPLETE" directly as the notification to confirm the transfer is done, instead of using a "UUID"?
This was the part I struggled with during implementation, and it’s one of the reasons I was motivated to write this example.
If that’s the case, let me take some time to verify whether your recommended status synchronization logic between the server and client works well without relying on UUID-based notifications.

> - I see you're doing blocking transfers one after another. One great feature of nixl is all of the APIs are non-blocking. And you can start all the transfer together. So I think being closer to what vLLM also does, you can start all the transfers, add the handles into a list, and remove them from the list when they're complete (again similar to vLLM implementation). When they're all done, then you can use the send_msg to indicate all of them are complete. I also suggest putting something more informative in the notifications instead of keepalive or complete, let's say the block ID. That being said, there is one point to consider here, we don't provide guarantees between transfer ordering. So you should check all of them are done, before sending that final notification. Or if you really don't care, you can skip assigning any notifications to the transfers, and just have one at the end, after checking all the transfers where complete.

At this point, I’d like to clarify your suggestion further.
[In the current example](https://github.com/ai-dynamo/nixl/pull/853/files#diff-76a885c3e0f39b216014fa5ee93240dee3f379b5518d413e84f9e721301c4190R68-R84), the code emulates transferring multiple blocks across tensors at once using make_prepped_xfer, and I assume this follows your suggestion for asynchronous transfer.
The reason I placed a for loop in the main function is simply to allow users to perform multiple attempts for throughput experiments. That’s why I believe each step should run in a blocking manner.

> - One point of having multiple transfers per agent in vLLM is they go to different agents, but here all of them go to the same agent. So you could've made a single larger transfer, and the underlying backend might have been optimized that transfer in a better way. It's good to have this as an example, but let's clarify that with a comment.

This comment seems to be related to the same context I responded to earlier, so I’m wondering if it would be acceptable to simply say:
“This example loops multiple attempts to evaluate the transfer rate for each step. Each step transfers multiple blocks asynchronously.”
Does this align with your thought?

I really appreciate your comprehensive review. I'd like to do my best to improve my work based on your feedback. Thanks!

bloodeagle40234:
I’ve fixed the check_remote_metadata call based on the comments.
The other changes still need further discussion with the reviewers, in my opinion.

If we need more enhancement even if it's in "example", perhaps add more examples but it could be in different pull requests to reduce the pull request size. Sorry this pr is already XL though.

mkhazraee:
Hi Kota,

Really sorry for the delayed reply, I was giving a talk at the PyTorch conference, then some work related to GTC, then attending Ray Summit, and I lost track of this PR.

- Regarding using tensors, I see you do: 
tensors = [torch.zeros(tensor_size, dtype=torch.bfloat16, device=device) for x in range(args.layers)] 
First of all, you can already pass a list of tensors directly to NIXL API, and it would get the addr/len/devID parts. No need to extract them yourself. My other point was good chance these sensors end up back to back in memory, so you can make them a single tensor, and then have blocks within it as individual tensor. So let's say you wanted 2D tensors, you can define a single 3D tensor, use the single tensor for registration, and then use slicing during prep. That becomes more similar to vLLM. That being said, that's what they're doing per layer, different layer tensors are not necessarily back to back, so this was a suggestion based on your current code.
   
- Regarding notifications, yes you can put whatever string (actually byte array) you want. That being said, unless you block between transfers (which you're due to while 1), there is no guarantee of ordering between transfers. Let's say you make 10 transfers, 9 of them being "KEEPALIVE" and one being "COMPLETE", and you start them in non-blocking manner, then the transfer with COMPLETE notification might finish before an earlier KEEPALIVE. So I think this might be a tad confusing as an example. Maybe you can put a transfer number (like 1 to 10) as the notification, and wait until you receive all of them. (similar to vLLM).

- Regarding asynchronous aspect, there are two levels of async, one being within each xfer_req, and one across xfer_req. If you want to only benefit from the internal one, and have a blocking mechanism across xfer_req (to make them ordered for example), that's fine to do what you're doing in that aspect, but let's add a comment and explanation in README, this is how to enforce ordering across Xfers as an example, otherwise they could've been started together and based on different notification messages understand when all the transfers are done.

Moein

bloodeagle40234:
Thanks @mkhazraee for further clarifications. 

For the first item, I tried using a single contiguous tensor instead of a list of tensors at https://github.com/ai-dynamo/nixl/pull/853/commits/e85a9be24fc2fa5e1adadb1106582e966b317f6b. This change actually reduces and simplifies the code, so I hope it aligns with your suggestion.

For the second and third items, I created another commit at https://github.com/ai-dynamo/nixl/pull/853/commits/0d60e5cd505ae93c03d1b46e96d442f719b7421d. With this update, the client sends the total number of sequences (the counts in the code), and the server checks the transferred sequence number in the message during the transfers. This seems to eliminate the blocking while loop with check_remote_transfer_done, so it may address the third comment—unless I’m missing something.

I’m really glad you got back to this, so please don’t worry about the delay in review. I hope your work on PyTorch Con and GTC went well.

If you notify anything to update, please let me know. Thanks, again.

tstamler:
Hi @bloodeagle40234, Moein is on vacation but I talked to him and I see you addressed his comments. I'll take one more look and try to get this merged.

bloodeagle40234:
Thanks @tstamler for progressing this PR.

Is there anything I can do to help with the CI failure?
It seems to be related to NVIDIA's internal CI process, so I couldn't find any details about the error for now.

bloodeagle40234:
@mkhazraee @tstamler Can I have any update to move forward this pull request before Christmas vacation?

mkhazraee:
Thanks Kota for the reminder, I lost track of this. 

Nitpicks first: the first commit to use a single tensor is what I meant, only minor thing is when you want to register a tensor, you don't need to pass the memory type. It's a tensor, NIXL python wrapper will figure it internally. In second commit, I see again VRAM is passed to prepXferDlist, but if it's already a dlist, then the type is included, that VRAM will be ignored. I see pycache is included in the files, let's remove it. 

Also let's clarify what makes this example unique now, to name it based on that. I think these 2 examples are the closest ones:
https://github.com/ai-dynamo/nixl/blob/main/examples/python/basic_two_peers.py
https://github.com/ai-dynamo/nixl/blob/main/examples/python/expanded_two_peers.py

Then we can come up with a better name. (I think it might need one more round of clean up, it's a tad hard to follow the full logic of the example, but first I want to know what is the core of challenges you had to figure and clean it up accordingly).

Sorry again for the delays and thanks for your contribution,
Moein

bloodeagle40234:
Thanks @mkhazraee 

> Nitpicks first: the first commit to use a single tensor is what I meant, only minor thing is when you want to register a tensor, you don't need to pass the memory type. It's a tensor, NIXL python wrapper will figure it internally. In second commit, I see again VRAM is passed to prepXferDlist, but if it's already a dlist, then the type is included, that VRAM will be ignored. I see pycache is included in the files, let's remove it.

Got it. I’ll remove the memory type when the input is a tensor and also remove the VRAM parameter passed within prepXferDlist.
I’ll rebase and force-push this branch to remove the unintended __pycache__ binary. While doing that, I’ll also rebase onto the latest master to keep this pull request clean.

> Also let's clarify what makes this example unique now, to name it based on that. I think these 2 examples are the closest ones:
> https://github.com/ai-dynamo/nixl/blob/main/examples/python/basic_two_peers.py
> https://github.com/ai-dynamo/nixl/blob/main/examples/python/expanded_two_peers.py
> Then we can come up with a better name. (I think it might need one more round of clean up, it's a tad hard to follow the full logic of the example, but first I want to know what is the core of challenges you had to figure and clean it up accordingly).

I didn’t realize you had recently worked on examples like [this](https://github.com/ai-dynamo/nixl/pull/1072).

I really like those examples—they help me understand the nixl API behavior more deeply. It may cover a part of this pull request as you said.
However, I’d still believe this one as a strong example of capability and performance considerations for memory block transfers. Users of this example can simulate KV cache transfers with configurable parameters (such as LLM model dimensions, layers, etc.), and then measure throughput for specific transfer block sizes.

bloodeagle40234:
Thinking about a new name for this example, perhaps something like kv_cache_transfer_example.py would be straightforward. What do you think?

mkhazraee:
Hi Kota,

Glad to hear the new examples helped clarify things! In conversations with users, I’d noticed recurring confusion around these concepts, and improving the examples had been on the list for a while. Your PR and our back-and-forth helped highlight those gaps and gave the right push to prioritize and refine the basic examples.

I see what you mean about using this example to simulate KV cache transfers and measure performance. We actually have a dedicated tool for that — [KV Bench](https://github.com/ai-dynamo/nixl/tree/main/benchmark/kvbench) — which provides a more complete setup for benchmarking and testing transfer configurations across different model sizes.

I’ve actually been contemplating this a bit, and since the main KV-cache concept here focuses on how we split the tensor (basically an evolved version of the expanded_two_peers example), I’m thinking it might make more sense to enhance that one rather than add another with overlapping logic. What do you think might still be missing from the current version that we could integrate?

Moein

bloodeagle40234:
Hi @mkhazraee,

> I’ve actually been contemplating this a bit, and since the main KV-cache concept here focuses on how we split the tensor (basically an evolved version of the expanded_two_peers example), I’m thinking it might make more sense to enhance that one rather than add another with overlapping logic. What do you think might still be missing from the current version that we could integrate?

Telling about the expanded_two_peers example, following things are what I struggled with the code as users.

1. Giant if/else Condition for Initiator and Target.
For now, I understand the example you provided, but at first glance, it was difficult to grasp what was happening. I also found the negotiation process between the initiator and the target a bit confusing because those steps are included in a single file with a giant if/else block. By dividing the process into server and client (a.k.a. target and initiator), newcomers may better understand the transfer steps required by the NIXL API. This point is also what I’d like to address in the VRAM example by adding a more detailed README and creating a utils.py file to separate common logic. This should make the understanding process more straightforward for users.

2. Description of NIXL Memory Management
This might be partially addressed by the expanded_two_peers example but it seems still difficult to understand what's the "block" and the index ("idx") on the way of memory transportation. In this example, I used similar words (e.g. block_len, shape_len, etc...) of vllm implementation as possible so more comments or step description with those technical terms would be helpful for users to emulate the real transfer model with NIXL for disaggregated inference. 

3. Configurable Transfer Parameters to Emulate Models
From my brief review of KV_Bench, the configuration of supported models appears to be limited to a few open-weight models. While that may be sufficient in some cases, I actually needed to emulate an unsupported model (gpt-oss and mistral) last August for my NVIDIA AI Day presentation held in Japan. In addition, I had to investigate the KV transfer bottleneck using this example—specifically, the number of tokens in a block. The --block-size argument in this example (i.e., how much KV data is actually transferred) was the missing piece in my investigation at that time. This point is where I strongly push my example to help users to get better investigation on their research and development works.

Regarding future work, I’m fine with either approach—merging this into the enhancement of expanded_two_peers or reviewing it independently. However, I’m not entirely sure how I can contribute to that effort so I'd like to hear what you imagine the next step of this PR.

mkhazraee:
Hi Kota,

1) I see your point, right now that if/else is confusing. We didn't want to split it to two files, specially for the basic_two_peers one, so the user just reads through a single simple file. But fair point, if we make them two methods within the file, and based on parsing of the arguments we call each, that would be more clear. For such a simple example, doing external utils.py is not the best, user has to go back and forth between the files. I'm thinking expanded_two_peers is borderline long to see if it wants a split and such utils file, but on the other hand it builds on the same logic as basic_two_peers, so user can compare them back and forth, a cons for splitting. If that makes sense to you, can you add those methods for each side to make it easier to understand without the split? Def fresh eyes can help there.

2) These examples were simpler and more generic, so didn't exactly follow the kv-cache notion of vLLM. But I see what you mean, def we can link this to such usage for more clarity. I'm thinking we can expand on the README to clarify this, the current readme is rather minimal, how it maps to kv-cache, and even better we mention how it maps to weight transfer, two examples for more clarity. Keeping the benefit of a simple generic one, while helping the understanding with more examples of how it maps to that. Having those as comments in the file becomes too crowded and hard to read, but in README we can even add diagrams. WDYT? If you agree, can you also help with updating that readme in examples/python, maybe some extra comments if necessary in the file too.

3) This is actually an invaluable feedback, I say we should do the proper thing and improve the KVBench then, it is what it's supposed to deliver. Tagging @aranadive who knows more about the KVBench to see how we can enable this. If you can open an issue about it with more details and tag Adit and me, we will follow up on that separately. If by any chance you already have some insights what needs to be added and can open a PR, even better.

Let me know what you think about my suggestions. There are no rules about who can contribute to examples, so if you open a PR changing the readme and those 2 examples that would be great. The points you brought up are the points that were in my blindspot, another reason it would be great if you can make those changes as there might be more. I think this approach will make the examples much better for first time users.

Thanks for your effort and follow ups,
Moein

bloodeagle40234:
Thanks Moein (@mkhazraee ) for kindly responding my PR comment even it's in the world Chistmas vacation week.

My comments for your response are as follows:

> 1. If condition in example improvement.

That’s a fair point. Splitting the file may cause the back and forth between the files for users. As you suggested, define the initiator and the target method inside basic_two_peers and expand would be a resolution on this issue. If you prefer this way, I’d try to make another pull request something like

current:
```
    if args.mode == "target":
        <target code>
    else:
        <initiator code>
```
will be:
```
def target(some args):
   <target code>

def initiator(some args):
   <initiator code> 

<snip>

    if args.mode == "target":
        target(some args)
    else:
        initiator(some args)
```

I believe this is a simple way to keep a single file for comparing differences between the basic and expanded versions, while also making it easier to understand each process in a straightforward manner.

After that, we can continue our current discussion on whether we should implement another kv_cache mechanism inside the expanded version or create an additional example like this pull request.


> 2. Memory management and NIXL transfer steps description

Including a description of the required steps in the documentation (e.g., README) will help users understand the process and accelerate writing new code with NIXL. As you suggested, diagrams would also be helpful—I agree on that point, and we may need more documents with examples. From another perspective, I’d like to confirm whether you intended this conversation to address not only this kv_cache example but also the basic/expand_two_peers examples. At the moment, expand_two_peers seems as complex as what this PR aims to achieve (index-based transfer), so I’m concerned about which case I should focus on to tackle this problem.

Assuming you recommended, we work on basic/expand_two_peers examples first with this comment. Actually, my point 1 and 2 could be resolved with those examples. > Let me know what you think about my suggestions. There are no rules about who can contribute to examples, so if you open a PR changing the readme and those 2 examples that would be great. 

If that’s correct, I’ll create separate PRs for those changes (smaller ones for easier review). Does that align with your idea?

> 3. Consideration on KV cache emulation and how we can work with KV Bench.

Actually, I haven’t deep-dived into KV Bench yet, so I’ll start by creating an issue first. In my view, working on both example improvements and KV Bench enhancements in parallel might be tough for me, so I’d prefer to complete the example improvements first, since this PR still works for my case for now.


If you are OK on item 1 and 2, I'll take a time to work on basic/exapand_two_peers example in the new year.
Thanks again, and best wishes for the new year.

mkhazraee:
Hi Kota, 

Yes, agreed let's start with those basic examples in 2 separate PRs, they're the introductory examples and the cleaner they are the better. Then we can take it from there.

Thanks,
Moein

PR #857, opened 2025-10-03
BACKEND/UCX: Multi-device support: Allow setting CUDA context to match device of the memory

tvegas1:
## What?
Allow changing the CUDA context to match the address.

## Why?
Allow multi-device support.

## How?
```
$ nixlbench --backend UCX --initiator_seg_type VRAM --target_seg_type VRAM \
    --num_target_dev 8 --num_initiator_dev 8 --mode MG -op_type WRITE
...
16777216            32             357.716629     46.9           220.0          220.0          914.0          942.0          11089.0        11281.0
```


tvegas1:
Any input on testing it? We might even be able to remove it altogether since UCX will set it internally if needed?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1052,7 +1042,6 @@ nixlUcxThreadPoolEngine::sendXferRange(const nixl_xfer_op_t &operation,
 
 int
 nixlUcxThreadPoolEngine::vramApplyCtx() {
-    // TODO: Check if UCX can handle context change at runtime
```
brminich, line 1055: do we need to do thread restarts now? i think we do not, but @rakhmets  can address in next PRs

brminich: we can merge now to unblock nixlbench testing and improve later during ucx_mo removal

tvegas1: we do not need afaiu

PR #863, opened 2025-10-06
README: Document default buildtype as debug

michal-shalev:
## What?
Documents the default buildtype (debug) in the Build Options section of the README.

## Why?
The default `buildtype` was not documented anywhere, leaving users unaware that NIXL builds in debug mode by default without optimizations. This could lead to performance issues in production deployments.

## How?
Added `buildtype` to the Common build options list with the available values (debug and release), clarified the default in the basic build example, and included a release build example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,15 @@ $ ninja install
 NIXL supports several build options that can be specified during the meson setup phase:
 
 ```bash
-# Basic build setup with default options
+# Basic build setup with default options (debug build)
 $ meson setup <name_of_build_dir>
+
+# Setup for release build with optimizations
+$ meson setup <name_of_build_dir> --buildtype=release
 
 # Setup with custom options (example)
 $ meson setup <name_of_build_dir> \
+    --buildtype=release \         # Build with optimizations
     -Dbuild_docs=true \           # Build Doxygen documentation
     -Ducx_path=/path/to/ucx \     # Custom UCX installation path
     -Dinstall_headers=true \      # Install development headers
@@ -117,6 +121,7 @@ $ meson setup <name_of_build_dir> \
 ```
 
 Common build options:
+- `buildtype`: Build type `debug` or `release` (default: debug)
 - `build_docs`: Build Doxygen documentation (default: false)
 - `ucx_path`: Path to UCX installation (default: system path)
 - `install_headers`: Install development headers (default: true)


tvegas1 (approved):

README.md
```diff
@@ -105,18 +105,23 @@ $ ninja install
...
 $ meson setup <name_of_build_dir> \
+    --buildtype=release \         # Build with optimizations
     -Dbuild_docs=true \           # Build Doxygen documentation
     -Ducx_path=/path/to/ucx \     # Custom UCX installation path
     -Dinstall_headers=true \      # Install development headers
     -Ddisable_gds_backend=false   # Enable GDS backend
 ```
 
 Common build options:
+- `buildtype`: Build type `debug` or `release` (default: debug)
```
rakhmets: I think it's better to make `release` the default build type.

ovidiusm (dismissed):

README.md
michal-shalev: me too, @ovidiusm @rakhmets I'd like to do it in a different PR and merge this one, okay?

rakhmets: I don't think any changes to the documentation are needed in this case, except this line:
```
- `buildtype`: Build type `debug` or `release` (default: release)
```

michal-shalev: Please see: https://github.com/ai-dynamo/nixl/blob/35403b32c91cc49dbf2e90fa545333f822a187d1/src/plugins/gpunetio/README.md?plain=1#L53

michal-shalev: Also, @karya0 initially made `debug` the default build type, maybe he'll have some comments.

rakhmets: I meant that we don't need almost all changes presented in this PR if we change the mode to `release`.
If you are going to change the mode in the next PR, it turns out that you will first add changes to this PR and immediately delete the changes in the next one.

rakhmets: > Please see:
> 
> https://github.com/ai-dynamo/nixl/blob/35403b32c91cc49dbf2e90fa545333f822a187d1/src/plugins/gpunetio/README.md?plain=1#L51

and this one can be deleted along with changing the default build type.

michal-shalev: This PR is for documenting build type, because it's missing from the main README (along with it's default)
I don't know what was the original reason that made debug the default build type, which is why I wanted to change it in a separate PR, and give other maintainers a chance to explain/object/approve.

rakhmets: `buildtype` is a meson built-in option. It shouldn't be listed in NIXL common build options.
The only thing that needs to be specified at the moment is that NIXL is built in debug mode by default.

I suggest to fix default value instead, and delete all the changes introduced by this PR.

michal-shalev: For the past 4 months of integration, helping dozens of engineers, I discovered that build type meson built-in option isn't common knowledge, in UCX you we configure-devel/release, we're making the user choose a build type but here it's "hidden". IMO we should document how to build in release/debig modes just like we do it in UCX.

rakhmets: Agree. Let's put this into sub-sections: `Release build`, `Debug build`.

karya0: > Also, @karya0 initially made `debug` the default build type, maybe he'll have some comments.

I think I got it from the internal git repo. @aranadive might have better insights on whether we want to move to release permanently.

PR #869, opened 2025-10-06
BUILD: Change default buildtype from debug to release

michal-shalev:
## What?
Changes the default build type from debug to release.

## Why?
Users should get optimized, production-ready builds by default without needing to specify build flags. Debug builds remain available with `--buildtype=debug`.

## How?
Updated `meson.build` default_options and documentation (README.md, gpunetio plugin README) to reflect release as the new default.


michal-shalev:
@aranadive @karya0, what was the original reason for NIXL's default build type being debug?

brminich (approved):

README.md
```diff
@@ -102,16 +102,16 @@ $ ninja install
 
 ### Build Options
 
-#### Release build
+#### Release build (default)
 
 ```bash
-$ meson setup <name_of_build_dir> --buildtype=release
+$ meson setup <name_of_build_dir>
```
ovidiusm, line 121: I would put also the explicit command here

brminich:
maybe this CI issue is related, because we build release now
`./.gitlab/test_cpp.sh: line 62: ./bin/desc_example: No such file or directory`

src/plugins/gpunetio/README.md
```diff
@@ -50,5 +50,5 @@ NIXL bench also has the option to specify the DOCA GPUNetIO backend. An example
 
 ## Caveats
 
-By default NIXL is built with `buildtype=debug` option. This is ok for correctness and debugging.
-To run for performace (e.g. with NIXL bench) t's hightly recommended to build NIXL with `buildtype=release`.
+By default NIXL is built with `buildtype=release` option for optimal performance.
```
tvegas1: maybe keep the explicit recommendation for performance

tvegas1:
probably need to explicitly set debug for CI builds now.

rakhmets:
> maybe this CI issue is related, because we build release now `./.gitlab/test_cpp.sh: line 62: ./bin/desc_example: No such file or directory`

There is the following code in `meson.build`:
```
if get_option('buildtype') != 'release'
  subdir('test')
  subdir('examples')
endif
```
I'm not sure that this is the correct approach. I think it would be better to have separate options to build tests and examples. It makes sense to build examples in release mode.

ovidiusm:
CI should cover at least release mode, otherwise we will miss issues. Maybe there would be side effects for debugging, but we do not support core dump collection right now anyway.

So I think we should add tests and examples compilation in release builds.

But with care not to package them into wheels/crates.

ovidiusm:
A drawback is that asserts would be compiled out in release mode, so they will not trigger in CI if we build in release mode.

ovidiusm:
Please see related https://github.com/ai-dynamo/nixl/pull/872

src/plugins/gpunetio/README.md
michal-shalev: fixed

README.md
michal-shalev: fixed

ovidiusm (dismissed):

src/plugins/gpunetio/README.md
```diff
@@ -51,5 +51,5 @@ NIXL bench also has the option to specify the DOCA GPUNetIO backend. An example
 
 ## Caveats
 
-By default NIXL is built with `buildtype=debug` option. This is ok for correctness and debugging.
-To run for performace (e.g. with NIXL bench) t's hightly recommended to build NIXL with `buildtype=release`.
+To run for performace (e.g. with NIXL bench) it's highly recommended to build NIXL with `buildtype=release`.
```
iyastreb:
```suggestion
To run for performance (e.g. with NIXL bench) it's highly recommended to build NIXL with `buildtype=release`.
```

michal-shalev: fixed

mkhazraee (approved):

README.md
```diff
@@ -102,16 +102,16 @@ $ ninja install
 
 ### Build Options
 
-#### Release build
+#### Release build (default)
 
 ```bash
 $ meson setup <name_of_build_dir> --buildtype=release
```
aranadive: Remove buildtype arg here

aranadive (approved):

contrib/build-container.sh
```diff
@@ -81,6 +82,14 @@ get_options() {
         --no-cache)
             NO_CACHE=" --no-cache"
             ;;
+        --build-type)
```
iyastreb, line 89: minor Q: maybe better have the same naming? `buildtype`
Up to you

iyastreb (approved):

contrib/build-container.sh
michal-shalev: I followed the same naming pattern like the above code "--no-cache"

README.md
brminich: seems unresolved

contrib/build-container.sh
brminich: @iyastreb can we resolve this thread?

michal-shalev: @iyastreb @brminich still relevant?

README.md
michal-shalev: fixed

PR #871, opened 2025-10-07
chore: update rust and python attributions files

nv-nmailhot:
## What?

## Why?

## How?
architecture, approach, etc._


ATTRIBUTIONS-CPP.md
```diff
@@ -19,6 +19,48 @@ limitations under the License.
 
 This project uses the following third-party libraries. Each library is open-source and licensed under the terms indicated below.
 
+## LibFabric - 2.3.0
```
dmitry-tokarev-nv: Can you add that Dynamo project selected BSD license to use and redistribute LibFabric?

PR #872, opened 2025-10-07
chore: Include tests and examples in release builds

aranadive:
Lets build these for release builds. In release builds there is some optimization added that might help with running tests and seeing performance numbers. The asserts are removed in release builds so a new macro was added to check for return status' of the API calls.

## What?
Add tests and examples to release builds

## Why?
Its useful to have these tests run in release builds when looking at performance.

## How?
Remove the buildtype flag. Use a check macro verify that the function calls return correctly.


examples/cpp/nixl_etcd_example.cpp
```diff
@@ -22,6 +22,15 @@
 
 #include "nixl.h"
 
+#define CHECK_NIXL_ERROR(result, message, agent)                         \
```
ovidiusm: Since this is duplicated everywhere, we should place it in a common header file if possible

ovidiusm:
Related: https://github.com/ai-dynamo/nixl/pull/869

meson.build
```diff
@@ -235,11 +235,8 @@ plugins_inc_dirs = include_directories('src/plugins')
 utils_inc_dirs = include_directories('src/utils')
 
 subdir('src')
-
-if get_option('buildtype') != 'release'
-  subdir('test')
-  subdir('examples')
-endif
+subdir('test')
+subdir('examples')
```
rakhmets: We need options that allow us to build examples and test applications.
The default value of these options should be false.

Tests and examples should be enabled for CI.

examples/cpp/nixl_etcd_example.cpp
```diff
@@ -132,6 +141,7 @@ int main() {
     ret1 = A1.getAvailPlugins(plugins);
     assert (ret1 == NIXL_SUCCESS);
 
+    CHECK_NIXL_ERROR(ret1, "Failed to get available plugins", AGENT1_NAME);
```
rakhmets: can we remove asserts corresponding to the added error checkers?

meson.build
aranadive: I can add the option but I think they should be enabled by default. This would simplify build for customers to run/try NIXL tests. It would also avoid CI changes. The build time added should be minimal.

examples/cpp/nixl_etcd_example.cpp
aranadive: Done

examples/cpp/nixl_etcd_example.cpp
aranadive: Done

meson.build
aranadive: Added

src/utils/common/util.h
```diff
@@ -21,4 +21,21 @@
...
+    } while (0)
+
+#define CHECK_NIXL_ERROR(result, message)                                                       \
+    do {                                                                                        \
+        if (0 != result) {                                                                      \
+            std::cerr << "NIXL: " << message << " (Error code: " << result << ")" << std::endl; \
+            exit(EXIT_FAILURE);                                                                 \
+        }                                                                                       \
+    } while (0)
+
```
rakhmets: I think there is no need for a macro.
```C++
void
nixl_exit_on_failure(bool success, std::string_view message, std::string_view agent = {}) {
    if (success) return;

    NIXL_ERROR << message
               << (agent.empty() ? "" : " for agent " + std::string{agent})
               << ": " << nixlEnumStrings::statusStr(status) << " [" << status << "]";
    exit(EXIT_FAILURE);
}

void
nixl_exit_on_failure(nixl_status_t status, std::string_view message, std::string_view agent = {}) {
    nixl_exit_on_failure(status == NIXL_SUCCESS, message, agent);
}
```

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -275,15 +267,14 @@ void *releaseValidationPtr(nixl_mem_t mem_type, void *addr)
...
-        assert(0);
+        CHECK_NIXL_ERROR(1, "Unsupported memory type!");
     }
     return NULL;
 }
 
 void
 allocateWrongGPUTest(nixlUcxEngine *ucx, int dev_id) {
-    nixlBlobDesc desc;
+    nixlBlobDesc desc = {0};
```
rakhmets, line 279: Why is this change necessary? Why is it not enough to create using the default constructor (as it was before)?

examples/cpp/nixl_example.cpp
```diff
@@ -30,6 +39,10 @@ void check_buf(void* buf, size_t len) {
     // Do some checks on the data.
     for(size_t i = 0; i<len; i++){
         assert (((uint8_t*) buf)[i] == 0xbb);
+        if (((uint8_t *)buf)[i] != 0xbb) {
```
rakhmets: The assert on the previous line is no longer needed.

meson_options.txt
```diff
@@ -30,4 +30,6 @@ option('log_level', type: 'combo', choices: ['trace', 'debug', 'info', 'warning'
 option('rust', type: 'boolean', value: false, description: 'Build Rust bindings')
 
 # Tests
+option('build_tests', type: 'boolean', value: true, description: 'Build all tests')
+option('build_examples', type: 'boolean', value: true, description: 'Build all examples')
```
rakhmets, line 34: >> We need options that allow us to build examples and test applications.
>> The default value of these options should be false.
>>
>> Tests and examples should be enabled for CI.

> I can add the option but I think they should be enabled by default. This would simplify build for customers to run/try NIXL tests. It would also avoid CI changes. The build time added should be minimal.

The default build usually provides the minimum set of binary files required to use the library. Examples are optional part that helps users write applications using the API provided by the library. I think it would be better to provide an alias/option to build a developer version that includes tests and examples.

aranadive: Im not sure about that. Both tests and examples are a way for a NIXL customer to use NIXL OOB. IMO, we should do a release build with examples and tests enabled. If a developer wants to control the type of build or disable tests/examples we have separate options to do that.
If we want other umbrella options that cover several other options, lets do that separately.

test/unit/plugins/ucx/ucx_backend_test.cpp
aranadive: Was necessary since release build flagged the desc.len below as unitialized. BasicDesc's default constructor does not initialize values.

examples/cpp/nixl_example.cpp
aranadive: Removed

src/utils/common/util.h
aranadive: Thanks. I added this as a dependency that tests/examples can use and also to avoid a circular dependency with nixl library that defines nixlEnumStrings.

ovidiusm: I think the comment was about refactoring them into functions instead of macros, since they call exit()

meson_options.txt
rakhmets: There are several use cases related to the library usage/distribution.
- If the user needs to install (or build and install) the NIXL library as a dependency to run the application, then we should not install examples along with it, much less tests.
- If the user needs to install (or build and install) the NIXL library to develop an application using the NIXL API, then we should install NIXL library along with headers. We may also install examples, but as source code, not as compiled executables.
- If the user wants to develop the NIXL library, then we should install tests and examples.

Building tests and examples is only necessary for library developers, not for general use.

ovidiusm: For what it's worth, both points have merit, but I am slightly in favor of disabling building tests by default, since that is more common. About examples, not sure.

aranadive: In general, I dont disagree with you @rakhmets. However, with NIXL today the main way customers (frameworks) consume NIXL is through the python wheel where this isnt relevant. But other folks like AI practitioners just want to try NIXL (but arent really developers), will build from scratch and by building the test and examples by default, gives them best experience to run things. Developers can customize their builds as necessary. 
The reason we want tests to be included by default is that it also gives a baseline performance for several plugins - POSIX, GDS(MT), etc. which folks want to know. We dont compile or distribute nixlbench by default.

aranadive: We can reorg more if we want in a follow up PR.

PR #881, opened 2025-10-08
DEVICE/API: Update NIXL device API return status

michal-shalev:
## What?
Following https://github.com/openucx/ucx/pull/10928
Update NIXL device API documentation to reflect that post operations return `NIXL_IN_PROG` (not `NIXL_SUCCESS`). Optimize status conversion by removing dead `UCS_OK` branch from hot path.

## Why?
Post operations now return `UCS_INPROGRESS` from underlying UCX API, which maps to `NIXL_IN_PROG`. The documentation incorrectly listed `NIXL_SUCCESS` as a possible return value. The status converter also had an unnecessary `UCS_OK` check that added a branch to every post operation.

## How?
- Updated docs for all post functions to remove `NIXL_SUCCESS` and clarify completion checking
- Simplified `nixlGpuConvertUcsStatus` to only handle `UCS_INPROGRESS` and errors
- Inlined full status handling in `nixlGpuGetXferStatus` where `UCS_OK` is actually used

### DIFF
--- a/src/api/gpu/ucx/nixl_device.cuh
+++ b/src/api/gpu/ucx/nixl_device.cuh
@@ -63,7 +63,11 @@ struct nixlGpuXferReqParams {
  */
 __device__ inline nixl_status_t
 nixlGpuConvertUcsStatus(ucs_status_t status) {
-    return status == UCS_OK ? NIXL_SUCCESS : NIXL_ERR_BACKEND;
+    if (status == UCS_INPROGRESS) {
+        return NIXL_IN_PROG;
+    }
+    printf("UCX returned error: %d\n", status);
+    return NIXL_ERR_BACKEND;
 }
 
 /**
@@ -76,9 +80,11 @@ nixlGpuConvertUcsStatus(ucs_status_t status) {
  * @param size          [in]  Size in bytes of the memory to be transferred.
  * @param channel_id    [in]  Channel ID to use for the transfer.
  * @param is_no_delay   [in]  Whether to use no-delay mode.
- * @param xfer_status   [out] Status of the transfer. If null, the status is not reported.
+ * @param xfer_status   [out] Status of the transfer. If not null, use @ref
+ *                            nixlGpuGetXferStatus to check for completion.
  *
- * @return nixl_status_t      Error code if call was not successful
+ * @return NIXL_IN_PROG       Operation successfully posted.
+ * @return NIXL_ERR_BACKEND   An error occurred in UCX backend.
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
@@ -107,9 +113,11 @@ nixlGpuPostSingleWriteXferReq(nixlGpuXferReqH req_hndl,
  * @param signal_offset      [in]  Offset of the signal to be sent.
  * @param channel_id         [in]  Channel ID to use for the transfer.
  * @param is_no_delay        [in]  Whether to use no-delay mode.
- * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ * @param xfer_status        [out] Status of the transfer. If not null, use @ref
+ *                                 nixlGpuGetXferStatus to check for completion.
  *
- * @return nixl_status_t    Error code if call was not successful
+ * @return NIXL_IN_PROG            Operation successfully posted.
+ * @return NIXL_ERR_BACKEND        An error occurred in UCX backend.
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
@@ -143,9 +151,11 @@ nixlGpuPostSignalXferReq(nixlGpuXferReqH req_hndl,
  * @param signal_offset      [in]  Offset of the signal to be sent.
  * @param channel_id         [in]  Channel ID to use for the transfer.
  * @param is_no_delay        [in]  Whether to use no-delay mode.
- * @param xfer_status        [out] Status of the transfer. If null, the status is not reported.
+ * @param xfer_status        [out] Status of the transfer. If not null, use @ref
+ *                                 nixlGpuGetXferStatus to check for completion.
  *
- * @return nixl_status_t    Error code if call was not successful
+ * @return NIXL_IN_PROG            Operation successfully posted.
+ * @return NIXL_ERR_BACKEND        An error occurred in UCX backend.
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
 __device__ nixl_status_t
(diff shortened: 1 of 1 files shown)


src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -63,7 +63,11 @@ struct nixlGpuXferReqParams {
  */
 __device__ inline nixl_status_t
 nixlGpuConvertUcsStatus(ucs_status_t status) {
-    return status == UCS_OK ? NIXL_SUCCESS : NIXL_ERR_BACKEND;
+    if (status == UCS_INPROGRESS) {
+        return NIXL_IN_PROG;
+    }
+    printf("UCX returned error: %d\n", status);
```
rakhmets, line 69:
```suggestion
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -234,6 +246,7 @@ nixlGpuGetXferStatus(nixlGpuXferStatusH &xfer_status) {
     case UCS_INPROGRESS:
         return NIXL_IN_PROG;
     default:
+        printf("UCX returned error: %d\n", status);
```
rakhmets:
```suggestion
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -63,7 +63,11 @@ struct nixlGpuXferReqParams {
  */
 __device__ inline nixl_status_t
 nixlGpuConvertUcsStatus(ucs_status_t status) {
-    return status == UCS_OK ? NIXL_SUCCESS : NIXL_ERR_BACKEND;
+    if (status == UCS_INPROGRESS) {
```
tvegas1: need to add NIXL_SUCCESS too

michal-shalev: No, UCX doesn't return UCS_OK, so no need for an extra branch

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: I don't agree with you, we need an error log here. It doesn't add another branch, we already know that it's an error and NIXL_ERR_BACKEND is too generic.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: same

src/api/gpu/ucx/nixl_device.cuh
tvegas1: but i see `uct_cuda_ipc_ep_put_single` returns UCS_OK? anyhow it would seem safer to handle UCS_OK?

src/api/gpu/ucx/nixl_device.cuh
rakhmets: This logging method is not suitable in this context.
I think we don't need such logging in the device code.

src/api/gpu/ucx/nixl_device.cuh
michal-shalev: Seems like `cuda_ipc` and `rc_gda` are not aligned, okay, good catch, I pushed a fix

src/api/gpu/ucx/nixl_device.cuh
rakhmets: Moreover, using naked `printf` as a logging method is generally not convenient, since the logging level and output are not controlled.

michal-shalev: Why is it not suitable?

rakhmets: Printing the UCX error code does not provide useful information to the user, as it requires access to the UCX source code. The error information should be logged by UCX.

michal-shalev: I understand the concern about device API logging. However, I'm keeping this printf for now for the following reasons:
1. UCX doesn't consistently log errors in device code paths, leaving us with no visibility when things fail.
2. `NIXL_ERR_BACKEND` alone provides zero actionable information - at minimum we need the UCX error code to debug.
3. We're actively debugging integration issues with this code right now, and this printf has been essential for troubleshooting
4. Proper logging infrastructure is tracked separately - I've opened a ticket to implement a proper logging macro for device code

This is a pragmatic debugging aid during urgent integration work. I'll address the logging infrastructure properly in the follow-up ticket, but for now this printf is necessary and helpful.

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -63,7 +63,15 @@ struct nixlGpuXferReqParams {
  */
 __device__ inline nixl_status_t
 nixlGpuConvertUcsStatus(ucs_status_t status) {
```
ofirfarjun7, line 65: The goal of this function was to use minimal branches for translating from ucx status to NIXL in data path.
So IMO it should be:
`return status == UCS_INPROGRESS ? NIXL_IN_PROG: NIXL_ERR_BACKEND;`
And we should use it only in post functions.

You should keep `nixlGpuGetXferStatus`  unchanged.

ofirfarjun7: I noticed Thomas comment. 
I don't like it but because UCX returns both UCS_OK & INPROGRESS and NIXL API is broken let's continue with this PR and improve later.

michal-shalev: We decided to use `UCS_STATUS_IS_ERR` to eliminate the additional branch and prevent performance degradation.

ofirfarjun7: Adding to Michal.
I don't like either approaches but we will have separate discussion about it, for now main goal of this PR is fixing the broken API with minimal changes and performance impact.

michal-shalev: I pushed a fix

src/api/gpu/ucx/nixl_device.cuh
ofirfarjun7: Let's push the device debugging macro soon, for now leave this print and change it when macro is merged.

rakhmets: > I understand the concern about device API logging. However, I'm keeping this printf for now for the following reasons:
> 
> 1. UCX doesn't consistently log errors in device code paths, leaving us with no visibility when things fail.
> 2. `NIXL_ERR_BACKEND` alone provides zero actionable information - at minimum we need the UCX error code to debug.
> 3. We're actively debugging integration issues with this code right now, and this printf has been essential for troubleshooting
> 4. Proper logging infrastructure is tracked separately - I've opened a ticket to implement a proper logging macro for device code
> 
> This is a pragmatic debugging aid during urgent integration work. I'll address the logging infrastructure properly in the follow-up ticket, but for now this printf is necessary and helpful.

1. This issue should be addressed on UCX level, and not as w/a in NIXL.
2. As I previously mentioned, UCX error code without explanation only makes sense in conjunction with the UCX source code, and is almost useless in terms of more general NIXL usage.
3. (4.) I cannot agree that it is acceptable to make such changes to the main branch. Such code can only be used in personal branches during development.

PR #885, opened 2025-10-09
POSIX backend: LINUXAIO API introduced

anton-nayshtut:
This PR introduces a Linux AIO plugin for the POSIX backend.

Linux AIO, although only available on Linux, is known for much better performance than POSIX AIO.

This patch implements a Linux AIO backend and integrates it into the NIXL build system so it is only built when the platform supports it.

The PR also introduces the `LINUXAIO` API parameter to `nixlbench`.

Here are the `nixlbench` results (with `--storage_enable_direct` on top of NVME drive - [SAMSUNG MZPLJ1T6HBJR-00007](https://semiconductor.samsung.com/ssd/enterprise-ssd/pm1735/mzplj1t6hbjr-00007/)):

**AIO**

```

4096                1              0.165290       24.8           35.0           35.0           1.3            2.0            23.5           47.0
8192                1              0.289647       28.3           51.0           51.0           1.4            2.0            26.9           48.0
16384               1              0.449769       36.4           65.0           65.0           1.7            4.0            34.6           166.0
32768               1              0.847057       38.7           60.0           60.0           1.5            3.0            37.1           49.0
65536               1              1.393177       47.0           59.0           59.0           1.5            2.0            45.4           55.0
131072              1              1.976728       66.3           59.0           59.0           1.5            3.0            64.7           189.0
262144              1              1.984314       132.1          69.0           69.0           1.7            6.0            130.3          255.0
524288              1              2.681113       195.5          61.0           61.0           1.9            7.0            193.6          528.0
1048576             1              2.706190       387.5          61.0           61.0           1.8            5.0            385.6          835.0
2097152             1              2.683741       781.4          61.0           61.0           2.3            7.0            778.0          1609.0
4194304             1              2.621102       1600.2         59.0           59.0           2.1            6.0            1597.0         2184.0
8388608             1              2.647123       3169.0         21.0           21.0           2.0            6.0            3166.6         5939.0
16777216            1              2.680175       6259.7         60.0           60.0           2.1            8.0            6256.5         10677.0
33554432            1              2.617827       12817.7        61.0           61.0           2.2            13.0           12814.3        17233.0
67108864            1              2.586386       25947.0        60.0           60.0           3.8            25.0           25942.0        31822.0
```

**URING**
```

4096                1              0.208734       19.6           117.0          117.0          3.3            19.0           16.2           160.0
8192                1              0.418782       19.6           83.0           83.0           1.9            5.0            17.6           168.0
16384               1              0.774555       21.2           72.0           72.0           2.1            6.0            19.0           169.0
32768               1              1.312282       25.0           58.0           58.0           2.5            6.0            22.4           184.0
65536               1              1.736099       37.7           63.0           63.0           3.1            8.0            34.5           314.0
131072              1              2.148162       61.0           57.0           57.0           4.4            9.0            56.5           334.0
262144              1              2.804096       93.5           64.0           64.0           7.3            15.0           86.1           333.0
524288              1              2.636124       198.9          68.0           68.0           12.8           28.0           186.0          526.0
1048576             1              2.656291       394.8          69.0           69.0           21.0           56.0           373.6          1053.0
2097152             1              2.867947       731.2          59.0           59.0           23.9           81.0           706.1          1458.0
4194304             1              2.642940       1587.0         166.0          166.0          2.7            6.0            1581.3         2471.0
8388608             1              2.629579       3190.1         48.0           48.0           2.5            5.0            3186.5         7507.0
16777216            1              2.634350       6368.6         50.0           50.0           2.8            17.0           6364.7         7024.0
33554432            1              2.595498       12927.9        81.0           81.0           2.6            5.0            12923.7        23889.0
67108864            1              2.589345       25917.3        82.0           82.0           3.4            12.0           25911.9        34322.0
```

**LINUXAIO**
```

4096                1              0.103491       39.6           123.0          123.0          0.2            1.0            0.2            1.0
8192                1              0.202123       40.5           131.0          131.0          0.2            1.0            3.7            1.0
16384               1              0.407086       40.2           174.0          174.0          0.2            1.0            2.4            1.0
32768               1              0.839628       39.0           436.0          436.0          0.7            2.0            0.6            2.0
65536               1              2.037954       32.2           110.0          110.0          0.2            1.0            0.2            1.0
131072              1              4.011799       32.7           103.0          103.0          0.2            1.0            0.2            1.0
262144              1              5.590394       46.9           122.0          122.0          0.2            1.0            0.2            1.0
524288              1              13.309550      39.4           101.0          101.0          0.2            1.0            0.2            1.0
1048576             1              26.005428      40.3           103.0          103.0          0.2            1.0            0.3            1.0
2097152             1              4.383999       478.4          121.0          121.0          2.0            109.0          4.8            291.0
4194304             1              6.484923       646.8          112.0          112.0          2.9            166.0          8.7            539.0
8388608             1              13.066368      642.0          116.0          116.0          4.6            275.0          16.8           1048.0
16777216            1              26.129505      642.1          102.0          102.0          9.2            565.0          34.5           2161.0
33554432            1              52.217702      642.6          102.0          102.0          17.7           1095.0         83.2           5233.0
67108864            1              84.701161      792.3          275.0          275.0          41.5           2603.0         252.2          15875.0

```

### DIFF
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -91,13 +91,14 @@ class nixlAgentData {
                            std::hash<std::string>, strEqual>     remoteSections;
 
         // State/methods for listener thread
-        nixlMDStreamListener               *listener;
-        std::map<nixl_socket_peer_t, int>  remoteSockets;
-        std::thread                        commThread;
-        std::vector<nixl_comm_req_t>       commQueue;
-        std::mutex                         commLock;
-        bool                               commThreadStop;
-        bool                               useEtcd;
+        nixlMDStreamListener *listener;
+        std::map<nixl_socket_peer_t, int> remoteSockets;
+        std::thread commThread;
+        std::vector<nixl_comm_req_t> commQueue;
+        std::mutex commLock;
+        std::atomic<bool> commThreadStop;
+        std::atomic<bool> agentShutdown;
+        bool useEtcd;
         std::unique_ptr<nixlTelemetry> telemetry_;
         std::exception_ptr commThreadException_;
 
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -194,12 +194,18 @@ nixlAgent::nixlAgent(const std::string &name, const nixlAgentConfig &cfg) :
 
     if (data->useEtcd || cfg.useListenThread) {
         data->commThreadStop = false;
+        data->agentShutdown = false;
         data->commThread = std::thread(&nixlAgentData::commWorker, data.get(), std::ref(*this));
     }
 }
 
 nixlAgent::~nixlAgent() {
     if (data && (data->useEtcd || data->config.useListenThread)) {
+        data->agentShutdown = true;
+        while (!data->commQueue.empty()) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        }
+
         data->commThreadStop = true;
         if(data->commThread.joinable()) data->commThread.join();
 
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -676,6 +676,10 @@ nixlAgentData::commWorkerInternal(nixlAgent *myAgent) {
 }
 
 void nixlAgentData::enqueueCommWork(nixl_comm_req_t request){
+    if (agentShutdown) {
+        NIXL_WARN << "Agent shutting down, unable to accept new requests";
+        return;
+    }
     std::lock_guard<std::mutex> lock(commLock);
     commQueue.push_back(std::move(request));
 }


src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+nixl_status_t
+linuxAioQueue::submit(const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
+    if (!num_ios_to_submit) {
+        return NIXL_IN_PROG;
+    }
+
+    int ret = io_submit(io_ctx, num_ios_to_submit, ios_to_submit.data());
+    if (ret != num_entries) {
```
ovidiusm: Why compare with num_entires and not num_ios_to_submit?

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return NIXL_ERR_BACKEND;
+    }
+
+    // Check buffer and length
+    if (!buf || len == 0) {
+        NIXL_ERROR << "Invalid buffer or length provided to prepareIO";
+        return NIXL_ERR_BACKEND;
+    }
+
+    int idx = num_ios_to_submit;
```
ovidiusm, line 129: After a previous submit() num_ios_to_submit is 0, right? But what if the previous transfer has not completed?

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,150 @@
...
+            NIXL_ERROR << absl::StrFormat("linux_aio submit failed: %s", nixl_strerror(-ret));
+        } else {
+            NIXL_ERROR << absl::StrFormat(
+                "linux_aio submit failed. Partial submission: %d/%d", num_entries, ret);
+        }
+        return NIXL_ERR_BACKEND;
+    }
+
+    num_completed = 0;
+    num_ios_to_submit = 0;
```
ovidiusm, line 73: Why is this reset to 0?

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,150 @@
...
+    return NIXL_IN_PROG;
+}
+
+nixl_status_t
+linuxAioQueue::checkCompleted() {
+    if (num_completed == num_entries) {
```
ovidiusm, line 79: Why are we not comparing num_submitted? Maybe I am mistunderstanding how the API should be used, but there are several counters and it is not clear

src/plugins/posix/posix_backend.cpp
```diff
@@ -104,6 +106,23 @@ namespace {
...
+                const auto &value = custom_params->at("use_linux_aio");
+                if (value == "true" || value == "1") {
+#ifndef HAVE_LINUXAIO
+                    NIXL_ERROR << "linux_aio backend requested but not available - not built with "
+                                  "libaio support";
+                    return queue_t::UNSUPPORTED;
+#endif
+                    if (!QueueFactory::isLinuxAioAvailable()) {
+                        NIXL_ERROR << "linux_aio backend requested but not available at runtime";
+                        return queue_t::LINUXAIO;
```
ovidiusm: Is this return value correct?

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: I just used the uring as a base and it came from there,

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: I'm a newbie here, but as far as I understand [this API](https://github.com/ai-dynamo/nixl/blob/main/docs/nixl.md#transfer), [current NIXL code](https://github.com/ai-dynamo/nixl/blob/3c6a0f3fd60f9e3c47e0ead3a84de1c5442c57d6/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp#L901) and both existing AIO plugins  -  the xfer is not designed to be reused.

So, basically, there can’t be a `prepIO` call while the previous transfer hasn’t completed.

Please correct me if I'm wrong.

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Good question. I just followed the URING case above.

src/plugins/posix/linux_aio_queue.cpp
ovidiusm: Yes you are right, the queue is not reused for other transfers. Thanks for clarifying

src/plugins/posix/linux_aio_queue.cpp
ovidiusm: It does not matter since the queue will not be used for another prep/submit again

src/plugins/posix/linux_aio_queue.h
```diff
@@ -0,0 +1,59 @@
...
+class linuxAioQueue : public nixlPosixQueue {
+private:
+    io_context_t io_ctx; // I/O context
+    std::vector<struct iocb> ios; // Array of I/Os
+    int num_entries; // Total number of entries expected
+    std::vector<struct iocb *> ios_to_submit; // Array of I/Os to submit
+    int num_ios_to_submit; // Total number of entries to submit
+    std::vector<bool> completed; // Track completed I/Os
+    int num_completed; // Number of completed operations
+    int num_submitted; // Track number of submitted I/Os
```
ovidiusm: I do not see `num_submitted` assigned to anywhere, do we need it?

src/plugins/posix/linux_aio_queue.cpp
ovidiusm: Clear now

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                 ") failed with " + std::to_string(res));
+    }
+
+    ios_to_submit.assign(num_entries, nullptr);
+}
+
+linuxAioQueue::~linuxAioQueue() {
+    // There should not be any in-flight I/Os at destruction time
+    if (num_submitted > num_completed) {
+        NIXL_ERROR << "Programming error: Destroying linuxAioQueue with "
```
ovidiusm: I do not see how this check would trigger, given that num_submitted is always 0

src/plugins/posix/posix_backend.cpp
ovidiusm: Could we please find out? It might be wrong there as well

src/plugins/posix/linux_aio_queue.h
anton-nayshtut: We don't. Sorry. It was a legacy that came from AIO plugin. Removed. Thanks!

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: Removed! Thanks!

src/plugins/posix/posix_backend.cpp
anton-nayshtut: There was a bug there. I fixed it in a separate commit (it’s now the first in the series, so it can be merged independently) and adjusted the LINUXAIO patch accordingly. Thanks!

Unfortunately, there's another bug - the `nixlbench` ignores this issue, continues to the execution and then crashes:
```
WARNING: Adjusting num_iter to 1024 to allow equal distribution to 4 threads
WARNING: Adjusting warmup_iter to 128 to allow equal distribution to 4 threads
Connecting to ETCD at http://localhost:2379
ETCD Runtime: Registered as rank 0 item 1 of 1
POSIX backend with API type: URING
E1009 16:13:23.755869 2162473 posix_backend.cpp:99] io_uring backend requested but not available at runtime
E1009 16:13:23.756003 2162473 posix_backend.cpp:219] Failed to initialize POSIX backend - requested queue type not available: UNSUPPORTED
E1009 16:13:23.756051 2162473 nixl_agent.cpp:334] createBackend: backend initialization error for 'POSIX'
Waiting for all processes to start... (expecting 1 total: 1 initiators and 1 targets)
All processes are ready to proceed
Creating file: /mnt/nvme/nixlbench_posix_test_file_initiator_0
Creating file: /mnt/nvme/nixlbench_posix_test_file_initiator_1
Creating file: /mnt/nvme/nixlbench_posix_test_file_initiator_2
Creating file: /mnt/nvme/nixlbench_posix_test_file_initiator_3
[spdk-perf-01:2162473:0:2162473] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace (tid:2162473) ====
...
=================================
Segmentation fault (core dumped)
```

So I just wonder whether we should also fix this bug as a part of this patch set or as a separate PR.

Please advise.

ovidiusm: Let's fix it separately so that we do not block this PR.

anton-nayshtut:
> /build

@ovidiusm , I see that it needs a rebase. Could you please advise on the correct procedure? Should I rebase and push manually?

meson.build
```diff
@@ -31,11 +31,15 @@ thread_dep = dependency('threads')
 
 # Check for libaio (for POSIX plugin and test)
```
benlwalker: In a different PR we should fix this mistake where it assumes libaio is required for POSIX AIO. It's not. We don't need any symbol test or extra library for POSIX AIO, it's baked into glibc. But your Linux AIO additions here are correct.

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,143 @@
...
+linuxAioQueue::submit(const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
+    if (!num_ios_to_submit) {
+        return NIXL_IN_PROG;
+    }
+
+    int ret = io_submit(io_ctx, num_ios_to_submit, ios_to_submit.data());
+    if (ret != num_ios_to_submit) {
+        if (ret < 0) {
+            NIXL_ERROR << absl::StrFormat("linux_aio submit failed: %s", nixl_strerror(-ret));
+        } else {
```
benlwalker, line 65: Partial submission isn't a failure. We should handle it by trying again later.

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,143 @@
...
+        return NIXL_ERR_BACKEND;
+    }
+
+    for (int i = 0; i < rc; i++) {
+        struct iocb *io = events[i].obj;
+        size_t idx = (size_t)io->data;
+
+        ios_to_submit[idx] = nullptr; // Mark as completed
+
+        if (events[i].res < 0) {
```
benlwalker, line 99: There's no way to report that a single operation failed? The whole backend simply fails? This isn't a comment so much on your patch but on this entire POSIX backend I guess.

ovidiusm:
> > /build
> 
> @ovidiusm , I see that it needs a rebase. Could you please advise on the correct procedure? Should I rebase and push manually?

There are no conflicts, either a normal merge or a rebase should just work. You can also click on the update branch button and pull, it does a merge.

meson.build
anton-nayshtut: I know. I just wanted to make this change a separate PR as you mentioned.

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: I completely agree. However, it seems that the current API is built on the premise that it’s all or nothing, so for now I decided to follow it rather than break it by implementing a retry mechanism in functions that aren’t intended for this but are repeatedly called until success - like  `checkCompleted`.

Generally speaking, I think we need to make the API aware of retries and partial submissions.

Please advise.

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: Yeah, it seems so, unfortunately.

anton-nayshtut:
> > > /build
> > 
> > 
> > @ovidiusm , I see that it needs a rebase. Could you please advise on the correct procedure? Should I rebase and push manually?
> 
> There are no conflicts, either a normal merge or a rebase should just work. You can also click on the update branch button and pull, it does a merge.

Done! Thanks!

src/plugins/posix/linux_aio_queue.cpp
ovidiusm: How large would it have to be in practice for full submission to be rejected?

ovidiusm: Asking because we have code freeze on Monday, and I am wondering if you think you can merge this new API by then as is and fix the partial submission issue in a separate PR since a fix would be touching code that is out of scope of this PR, or you think it is best to postpone everything and merge in a future version.

Right now we are switching to major / bugfix release cadence, so 0.7.0 would be a good candidate for this feature and 0.7.1 would be a good candidate for the fix. Unless you think that the code will be broken and prefer not to ship it now. IIUC currently requests are rejected if the client app is trying to write/read too many descriptors at once, which may be acceptable if the limit is high and users have not reported such issues until now in the current implementation.

anton-nayshtut: It behaves the same way as, for example, `UringQueue` which submits IOs using [io_uring_submit](https://man7.org/linux/man-pages/man3/io_uring_submit.3.html) API. This API also returns the number of submitted submission queue entries.

That’s why I implemented the Linux AIO plugin in the same way. It adds functionality and doesn’t make things worse.

So, I think we can add it.

That said, I believe we must adjust the API to be aware of possible partial submissions ASAP.

benchmark/kvbench/commands/args.py
```diff
@@ -170,7 +170,7 @@ def nixl_bench_args(func):
     func = click.option(
         "--posix_api_type",
         type=str,
-        help="API type for POSIX operations [AIO, URING] (only used with POSIX backend",
+        help="API type for POSIX operations [AIO, URING, LINUXAIO] (only used with POSIX backend",
```
vvenkates27: Can Linux AIO be made default, is it available with manylinux pip builds? @aranadive ?

src/plugins/posix/linux_aio_queue.cpp
vvenkates27: I don't think we need change of API for partial submission, we can entirely handle this inside the plugin implementation. Having said that, we can address this together for all POSIX IOs in a separate PR.

src/plugins/posix/linux_aio_queue.cpp
vvenkates27: In NIXL if entire transfer is not complete its considered a failure. If user wants to get more visibility, they can split their transfers to more requests. Everything is async

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -0,0 +1,143 @@
...
+    return NIXL_IN_PROG;
+}
+
+nixl_status_t
+linuxAioQueue::checkCompleted() {
+    if (num_completed == num_entries) {
+        return NIXL_SUCCESS;
+    }
+
+    struct io_event events[32];
```
vvenkates27, line 83: Why specifically 32 events?

benchmark/kvbench/commands/args.py
anton-nayshtut: > Can Linux AIO be made default, is it available with manylinux pip builds? @aranadive ?

It can. As @benlwalker rightly mentioned in one of his comments, the current POSIX AIO dependency is incorrect. NIXL currently assumes that `libaio` is required for POSIX AIO, while in reality it is part of GLIBC.

We’re going to fix this dependency in a separate PR, but the fact that POSIX AIO is available in the manylinux build means that `libaio` is available there. This means we can make Linux AIO the default option.

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: Up to. No specific reason. We can pick any other number.

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: > I don't think we need change of API for partial submission, we can entirely handle this inside the plugin implementation. Having said that, we can address this together for all POSIX IOs in a separate PR.

Could you please elaborate on your suggestion? As far as I understand, the `submit()` API is only called once, and then `checkCompleted()` is called repeatedly in a loop until the transfer is finished. Is that correct? If so, how would you recommend implementing partial submission handling? Should the plugin keep sending the rest of the data in `checkCompleted()` until it’s all done, or do you have something else in mind?

meson.build
anton-nayshtut: I added a patch that fixes the dependency issue.

benchmark/kvbench/commands/args.py
vvenkates27: Ok if there is no objection for release builds/pip wheels from @aranadive  @ovidiusm , I am ok to make linux aio the default

src/plugins/posix/linux_aio_queue.cpp
vvenkates27: We should not need to call checkCompleted to ensure progress. checkCompletion should be needed to ensure buffers can be freed/reused. IO submission should happen in a non-blocking fashion. The checkCompletion called repeatedly is an artifact of the test rather than a requirement

src/plugins/posix/linux_aio_queue.cpp
vvenkates27: Would this need to be a tunable? can it be determined based on batch size?

anton-nayshtut: I considered different options here but chose the simplest one for the following reasons:
1. static arrays are faster.
2. limiting the number of IOs handled in a single `checkCompleted` call has a throttling effect. If the caller executes other logic between consecutive `checkCompleted` calls, this approach gives that logic room to breathe and helps avoid stalls.
3. ultimately, it doesn’t seem to affect overall performance - at least if we consider function call overhead relatively  negligible. In. In the end, we have to iterate over all the IOs anyway - whether with more `checkCompleted` calls handling fewer IOs each, or fewer calls handling more IOs each.

benchmark/kvbench/commands/args.py
aranadive: @vvenkates27 should we just drop posix aio support and promote linuxaio as aio? Im not sure it serves any purpose other than getting more perf issues...

vvenkates27: Yes lets do it

aranadive: @anton-nayshtut Instead of calling this LINUXAIO, lets call this AIO (default) and the previous AIO as POSIXAIO. We can remove POSIXAIO in some future version.

anton-nayshtut: @aranadive done as a separate patch.

vvenkates27:
> > > > /build
> > > 
> > > 
> > > @ovidiusm , I see that it needs a rebase. Could you please advise on the correct procedure? Should I rebase and push manually?
> > 
> > 
> > There are no conflicts, either a normal merge or a rebase should just work. You can also click on the update branch button and pull, it does a merge.
> 
> Done! Thanks!

@anton-nayshtut the clang is failing currently can you fix that?

vvenkates27:
Anything else missing @aranadive  for this PR, can we merge once CI passes?

anton-nayshtut:
> > > > > /build
> > > > 
> > > > 
> > > > @ovidiusm , I see that it needs a rebase. Could you please advise on the correct procedure? Should I rebase and push manually?
> > > 
> > > 
> > > There are no conflicts, either a normal merge or a rebase should just work. You can also click on the update branch button and pull, it does a merge.
> > 
> > 
> > Done! Thanks!
> 
> @anton-nayshtut the clang is failing currently can you fix that?

Done!

aranadive (approved):

PR #894, opened 2025-10-12
DEVICE/API: Simplify createGpuXferReq to single-step API

michal-shalev:
## What?
Replace two-step GPU transfer request creation with a single `createGpuXferReq` API that takes descriptor lists directly and returns both transfer request and GPU request handles.

## Why?
Simplify the API by reducing GPU transfer request creation from two separate function calls to a single call.

## How?
Updated `createGpuXferReq` to substitute calls to `createXferReq` + `createGpuXferReq`. Removed the old `createGpuXferReq` method and updated tests accordingly.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -862,7 +862,7 @@ xferBenchNixlWorker::allocateMemory(int num_threads) {
         iov_lists.push_back(iov_list);
 
         /* Workaround for a GUSLI registration bug which resets memory to 0 */
-        if (seg_type == DRAM_SEG) {
+        if (XFERBENCH_BACKEND_GUSLI == xferBenchConfig::backend && seg_type == DRAM_SEG) {
             for (auto &iov : iov_list) {
                 if (isInitiator()) {
                     memset((void *)iov.addr, XFERBENCH_INITIATOR_BUFFER_ELEMENT, buffer_size);
--- a/src/infra/nixl_descriptors.cpp
+++ b/src/infra/nixl_descriptors.cpp
@@ -112,19 +112,19 @@ nixlBlobDesc::nixlBlobDesc(const nixlBasicDesc &desc,
 }
 
 nixlBlobDesc::nixlBlobDesc(const nixl_blob_t &blob) {
-    size_t meta_size = blob.size() - sizeof(nixlBasicDesc);
-    if (meta_size > 0) {
+    if (blob.size() > sizeof(nixlBasicDesc)) {
+        const size_t meta_size = blob.size() - sizeof(nixlBasicDesc);
         metaInfo.resize(meta_size);
         blob.copy(reinterpret_cast<char*>(this), sizeof(nixlBasicDesc));
         blob.copy(reinterpret_cast<char*>(&metaInfo[0]),
                  meta_size, sizeof(nixlBasicDesc));
-    } else if (meta_size == 0) {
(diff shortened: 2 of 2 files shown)


src/api/cpp/nixl.h
```diff
@@ -322,12 +322,22 @@ class nixlAgent {
...
+         * @param  remote_agent   [in]  Remote agent name for accessing the remote data
+         * @param  gpu_req_hndl   [out] GPU transfer request handle
+         * @param  req_hndl       [out] Transfer request handle
+         * @param  extra_params   [in]  Optional extra parameters
+         * @return nixl_status_t  Error code if call was not successful
          */
         nixl_status_t
-        createGpuXferReq(const nixlXferReqH &req_hndl, nixlGpuXferReqH &gpu_req_hndl) const;
+        createGpuXferReq(const nixl_xfer_dlist_t &local_descs,
+                         const nixl_xfer_dlist_t &remote_descs,
```
itayalroy, line 336: I think it's confusing to have local_descs and remote_descs with different lengths, which implies that the last `len(remote_descs) - len(local_descs)` remote descs are inline only.

Instead, I think we should enforce `len(local_descs) == len(remote_descs)` and introduce a third parameter- `inline_descs` which will describe remote memory that has no corresponding src buffer (and therefore can only be written into with inline data)

src/core/nixl_agent.cpp
```diff
@@ -1231,25 +1221,165 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
...
+                            const nixl_xfer_dlist_t &remote_descs,
+                            const nixlBasicDesc &signal_desc,
+                            const std::string &remote_agent,
+                            nixlGpuXferReqH &gpu_req_hndl,
+                            nixlXferReqH *&req_hndl,
+                            const nixl_opt_args_t *extra_params) const {
+    nixl_status_t ret1, ret2;
+    nixl_opt_b_args_t opt_args;
+
+    std::unique_ptr<backend_set_t> backend_set = std::make_unique<backend_set_t>();
```
aranadive, line 1244: If most of the code here is common with CreateXferReq can do refactor of both instead of duplication.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -39,14 +39,14 @@
 #endif
 
 namespace {
-    void moveNotifList(notif_list_t &src, notif_list_t &tgt)
```
aranadive, line 42: Can we skip the files that are only clang formatted but otherwise weren't modified for the PR issue?

michal-shalev: removed the unrelated changes

michal-shalev:
/ok to test [aa36008](https://github.com/ai-dynamo/nixl/pull/894/commits/aa36008b02ed54d52be5c8fa8d43b24405771dce)

michal-shalev:
/ok to test [aa36008](https://github.com/ai-dynamo/nixl/pull/894/commits/aa36008b02ed54d52be5c8fa8d43b24405771dce)

michal-shalev:
/ok to test [aa36008](https://github.com/ai-dynamo/nixl/pull/894/commits/aa36008b02ed54d52be5c8fa8d43b24405771dce)

michal-shalev:
/ok to test [a124461](https://github.com/ai-dynamo/nixl/pull/894/commits/a12446185e565f37f742f0431bee072d85e56983)

michal-shalev:
@brminich @yosefe @ofirfarjun7 @itayalroy 
Closing this PR following the latest API discussions.

PR #900, opened 2025-10-14
add support for custom backend parameters in the rust bindings

cheese-head:
## What?
This PR adds support for custom backend parameters in the NIXL Rust bindings, enabling users to pass arbitrary key-value configuration parameters to NIXL backends.

## Why?
This PR will allow from Rust to
- Configure backend-specific options (e.g., UCX transport settings, GDS options)
- Passing custom parameters for plugin initialization

## How?
It exposes to APIs outside of the crate. 

- try_clone
- from_iter


roiedanino:
Hi @cheese-head,
Can you please resolve conflicts and clang-format issues?

Thanks!

src/bindings/rust/src/lib.rs
```diff
@@ -41,42 +41,42 @@ mod bindings {
...
 };
 
 // Re-export status codes
 pub use bindings::{
     nixl_capi_status_t_NIXL_CAPI_ERROR_BACKEND as NIXL_CAPI_ERROR_BACKEND,
     nixl_capi_status_t_NIXL_CAPI_ERROR_INVALID_PARAM as NIXL_CAPI_ERROR_INVALID_PARAM,
+    nixl_capi_status_t_NIXL_CAPI_ERROR_NO_TELEMETRY as NIXL_CAPI_ERROR_NO_TELEMETRY,
     nixl_capi_status_t_NIXL_CAPI_IN_PROG as NIXL_CAPI_IN_PROG,
     nixl_capi_status_t_NIXL_CAPI_SUCCESS as NIXL_CAPI_SUCCESS,
-    nixl_capi_status_t_NIXL_CAPI_ERROR_NO_TELEMETRY as NIXL_CAPI_ERROR_NO_TELEMETRY
```
roiedanino: Why this change it needed?

src/bindings/rust/wrapper.cpp
```diff
@@ -44,82 +44,81 @@ nixl_capi_thread_sync_to_nixl(nixl_capi_thread_sync_t sync) {
...
 };
 
 struct nixl_capi_backend_s {
-  nixlBackendH* backend;
+    nixlBackendH *backend;
 };
 
 struct nixl_capi_opt_args_s {
-  nixl_opt_args_t args;
+    nixl_opt_args_t args;
```
roiedanino: Please revert all changes that are irrelevant for this PR

This probably happened while fixing clang-format errors. The CI shows the exact diff required for the fix (which contains only clang-format issues in your PR code, and not in the entire project). You can apply the patch suggested in CI using git.

roiedanino (changes requested):
Please make sure your PR only contains changes relevant to backend parameters, without formatting other parts of the code

cheese-head:
@roiedanino @aranadive clang format changes have been reverted and fixed.

src/bindings/rust/src/utils/params.rs
```diff
@@ -82,6 +82,86 @@ impl Params {
...
+    ///
+    /// let map = HashMap::from([
+    ///     ("access_key", "*********"),
+    ///     ("secret_key", "*********"),
+    ///     ("bucket", "my-bucket"),
+    /// ]);
+    ///
+    /// let params = Params::try_from_iter(map.iter().map(|(k, v)| (*k, *v)))?;
+    /// ```
+    pub fn try_from_iter<I, K, V>(iter: I) -> Result<Self, NixlError>
```
roiedanino: Maybe it would be better to call this function `from` just like in HashMap, following language convention

src/bindings/rust/src/utils/params.rs
```diff
@@ -82,6 +82,86 @@ impl Params {
...
+
+    /// Creates a new Params object by copying from another Params
+    ///
+    /// # Example
+    /// ```ignore
+    /// let original_params = agent.get_plugin_params("OBJ")?.1;
+    /// let mut modified_params = original_params.try_clone()?;
+    /// modified_params.set("bucket", "my-custom-bucket")?;
+    /// ```
+    pub fn try_clone(&self) -> Result<Self, NixlError> {
```
roiedanino: Maybe we can name it `clone`?

src/bindings/rust/tests/tests.rs
```diff
@@ -220,6 +220,64 @@ fn test_params_iteration() {
...
+    use std::collections::HashMap;
+
+    let map = HashMap::from([
+        ("key1", "value1"),
+        ("key2", "value2"),
+        ("key3", "value3"),
+    ]);
+
+    let params = Params::try_from_iter(map.iter().map(|(k, v)| (*k, *v)))
+        .expect("Failed to create params from iterator");
```
roiedanino:
```suggestion
    let params = Params::from(&map)
        .expect("Failed to create params from iterator");
```

should work as well

src/bindings/rust/tests/tests.rs
```diff
@@ -220,6 +220,64 @@ fn test_params_iteration() {
...
+    }
+
+    assert_eq!(found_keys.len(), 3, "Should have 3 key-value pairs");
+    assert_eq!(found_keys.get("key1"), Some(&"value1".to_string()));
+    assert_eq!(found_keys.get("key2"), Some(&"value2".to_string()));
+    assert_eq!(found_keys.get("key3"), Some(&"value3".to_string()));
+}
+
+#[test]
+fn test_params_try_clone() {
```
roiedanino:
```suggestion
fn test_params_clone() {
```

src/bindings/rust/tests/tests.rs
```diff
@@ -220,6 +220,64 @@ fn test_params_iteration() {
...
+    for param in original_params.iter().unwrap() {
+        let param = param.unwrap();
+        original_map.insert(param.key.to_string(), param.value.to_string());
+    }
+
+    let mut copied_map = std::collections::HashMap::new();
+    for param in copied_params.iter().unwrap() {
+        let param = param.unwrap();
+        copied_map.insert(param.key.to_string(), param.value.to_string());
+    }
```
roiedanino: Maybe we can do `HashMap::from(copied_params)` if Params iterator provides a key and a value?

src/bindings/rust/tests/tests.rs
```diff
@@ -220,6 +220,64 @@ fn test_params_iteration() {
     }
 }
 
+#[test]
+fn test_params_try_from_iter() {
```
roiedanino:
```suggestion
fn test_params_from_iter() {
```

src/bindings/rust/src/utils/params.rs
```diff
@@ -52,9 +52,9 @@ impl<'a> Iterator for ParamIterator<'a> {
...
 
         match status {
-            0 if !has_next => None,
+            0 if key_ptr.is_null() => None,
             0 => {
-                // SAFETY: If status is 0, both pointers are valid null-terminated strings
+                // SAFETY: If status is 0 and key_ptr is not null, both pointers are valid null-terminated strings
                 let result = unsafe {
                     let key = CStr::from_ptr(key_ptr).to_str().unwrap();
                     let value = CStr::from_ptr(value_ptr).to_str().unwrap();
```
roiedanino, line 74: I'm refering to line 61 (`Ok(ParamPair { key, value })`) it could be nicer to just return `Ok((key, value))` which will enable iterating for key, value:

```rust
    for (key, value) in params.iter() {
       ...
    }
```

It will also allow us to remove the `ParamPair` object, and use `HashMap::from(params.iter())`

src/bindings/rust/src/utils/params.rs
cheese-head: Since it can produce an error, would it be better to implement TryFrom?

src/bindings/rust/src/utils/params.rs
cheese-head: Do you want the entire ParamPair object removed?

src/bindings/rust/src/utils/params.rs
roiedanino: Most APIs can produce an error, that's what the `Result<>` return type is for; `Result` object forces the user code to handle errors.

src/bindings/rust/src/utils/params.rs
roiedanino: Why would we want that?

cheese-head:
@roiedanino @aranadive Can I get a review on this? I believe I have addressed all comments

contrib/Dockerfile
```diff
@@ -224,4 +224,4 @@ RUN export PATH=$VIRTUAL_ENV/bin:$PATH && \
     --output-dir /workspace/nixl/dist
 
 RUN cp build/src/bindings/python/nixl-meta/nixl-*.whl dist/
-RUN uv pip install dist/nixl*cp${DEFAULT_PYTHON_VERSION//./}*.whl dist/nixl-*-none-any.whl
+RUN uv pip install dist/nixl*cp${DEFAULT_PYTHON_VERSION//./}*.whl dist/nixl-*-none-any.whl
```
roiedanino, line 17: Missing an empty line at the end of the file

src/bindings/rust/src/utils/params.rs
```diff
@@ -130,16 +139,16 @@ impl Params {
...
+    pub fn clone(&self) -> Result<Self, NixlError> {
         let mut params = Self::create()?;
 
         if let Ok(iter) = self.iter() {
-            for pair in iter {
-                let pair = pair?;
-                params.set(pair.key, pair.value)?;
+            for result in iter {
+                let (key, value) = result?;
+                params.set(key, value)?;
```
roiedanino: wouldn't `Params::from(&self)` work here?

src/bindings/rust/src/utils/params.rs
```diff
@@ -77,11 +78,99 @@ impl Drop for ParamIterator<'_> {
...
+    /// ```
+    pub fn clone(&self) -> Result<Self, NixlError> {
+        let mut params = Self::create()?;
+
+        if let Ok(iter) = self.iter() {
+            for result in iter {
+                let (key, value) = result?;
+                params.set(key, value)?;
+            }
+        }
```
roiedanino: Perhaps we can use `Params::from(&self)` instead?

PR #908, opened 2025-10-14
libfabric: fix asymmetrical rails for heterogeneous node types.

fengjica:
This change:
 - passes the number of rail endpoints to others via serialization,
 - added all remote endpoints for each local rails into addresses,
 - round-robin pick local rails and remote endpoints for data transfer.

## What?
Fix the assumption in the code today that rails configuration is the same across nodes.

## Why?
Users intend to use AWS p5en and p6 nodes together.

## How?
See changes described above.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -1030,7 +1030,14 @@ xferBenchNixlWorker::exchangeIOV(const std::vector<std::vector<xferBenchIOV>> &l
                     if (basic_desc) {
                         remote_iov_list.push_back(basic_desc.value());
                     }
-                } else if (XFERBENCH_BACKEND_POSIX == xferBenchConfig::backend) {
+                } else if (XFERBENCH_BACKEND_GUSLI == xferBenchConfig::backend) {
+                    xferBenchIOV iov_remote(iov);
+                    iov_remote.addr = xferBenchConfig::gusli_bdev_byte_offset + file_offset;
+                    iov_remote.len = block_size;
+                    iov_remote.devId = 11;
+                    remote_iov_list.push_back(iov_remote);
+                    file_offset += block_size;
+                } else {
                     xferBenchIOV iov_remote(iov);
                     iov_remote.addr = file_offset;
                     iov_remote.len = block_size;
@@ -1041,13 +1048,6 @@ xferBenchNixlWorker::exchangeIOV(const std::vector<std::vector<xferBenchIOV>> &l
                         file_offset += block_size;
                         fd_idx = 0;
                     }
-                } else if (XFERBENCH_BACKEND_GUSLI == xferBenchConfig::backend) {
-                    xferBenchIOV iov_remote(iov);
-                    iov_remote.addr = xferBenchConfig::gusli_bdev_byte_offset + file_offset;
-                    iov_remote.len = block_size;
-                    iov_remote.devId = 11;
-                    remote_iov_list.push_back(iov_remote);
-                    file_offset += block_size;
                 }
             }
             res.push_back(remote_iov_list);


src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -794,13 +795,17 @@ nixlLibfabricRailManager::deserializeConnectionInfo(
     return NIXL_SUCCESS;
 }
 
+const std::string NUM_RAILS_TAG {"num_rails"};
```
yexiangd: keep it in libfabric_common

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -889,6 +893,21 @@ nixlLibfabricEngine::unloadMD(nixlBackendMD *input) {
...
+ *****************************************/
+
+void
+nixlLibfabricPublicMetadata::derive_remote_selected_endpoints() {
+    remote_selected_endpoints_.clear();
+
+    for (size_t i = 0; i < rail_remote_key_list_.size(); ++i) {
+        if (rail_remote_key_list_[i] != 0) {
+            remote_selected_endpoints_.push_back(i);
+        }
```
yexiangd, line 906: nit: else through some warning message

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -136,29 +136,34 @@ nixlLibfabricRailManager::shouldUseStriping(size_t transfer_size) const {
...
     bool use_striping = shouldUseStriping(transfer_size) && selected_rails.size() > 1;
-
+    NIXL_DEBUG << "use_striping: " << use_striping;
     if (!use_striping) {
         // Round-robin: use one rail for entire transfer
-        size_t rail_idx = round_robin_counter.fetch_add(1) % selected_rails.size();
-        size_t rail_id = selected_rails[rail_idx];
+        const auto counter_value = round_robin_counter.fetch_add(1);
+        const size_t rail_id = selected_rails[counter_value % selected_rails.size()];
+        const size_t remote_ep_idx =
```
yexiangd: use either `remote_ep_idx` or `remote_ep_id` across the board. in striping case you are using `remote_ep_id`

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -794,13 +797,17 @@ nixlLibfabricRailManager::deserializeConnectionInfo(
     return NIXL_SUCCESS;
 }
 
+const std::string NUM_RAILS_TAG{"num_rails"};
```
yexiangd: do you want to keep constants in a header file like libfabric_common.h or in rail_manager.h?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -816,11 +823,21 @@ nixl_status_t
...
     const std::string &key_prefix,
-    size_t expected_count,
     std::vector<std::array<char, LF_EP_NAME_MAX_LEN>> &endpoints_out) const {
-    endpoints_out.resize(expected_count);
 
-    for (size_t rail_id = 0; rail_id < expected_count; ++rail_id) {
+    size_t num_rails;
+    try {
+        const std::string str = ser_des.getStr(NUM_RAILS_TAG);
+        num_rails = std::stoull(str);
```
yexiangd: size_t range might be different from unsigned longlong. You can add a range check
```
    unsigned long long temp = std::stoull(ser_des.getStr(NUM_RAILS_TAG));
    
    if (temp > std::numeric_limits<size_t>::max()) {
        NIXL_ERROR << "Number too large for size_t: " << temp;
        return NIXL_ERR_BACKEND;
    }
```

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -579,15 +575,19 @@ nixlLibfabricEngine::createAgentConnection(
     }
 
     // Process all control rails in one operation
+    // Use only one local control rail.
+    std::unordered_map<size_t, std::vector<fi_addr_t>> control_rail_endpoints_map;
+    control_rail_endpoints_map[0].reserve(rail_manager.getNumControlRails());
```
yexiangd: wondering why do you assume Use only one local control rail but reserves space that's able to support X remote control rail?

also, is it duplicated with allocation in insertAllAddresses()? 
```
fi_addrs_out[rail_id].reserve(endpoints.size());
```

src/utils/libfabric/libfabric_rail_manager.h
```diff
@@ -260,12 +263,14 @@ class nixlLibfabricRailManager {
     serializeMemoryKeys(const std::vector<uint64_t> &keys, void *buffer, std::string &str) const;
     /** Deserialize memory keys and remote address
      * @param serialized_data Serialized memory information
+     * @param max_num_keys Max number of keys
```
yexiangd: should we just call it num_keys?  I feel the "max" part is a little confusing, wdy think?

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Please see reply below.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Done.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Done in the new update.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: I'd like to keep it in the cpp file, as it doesn't need to be exposed outside.

I moved it to the top of the file.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Yes. Please see the modified version in the update.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: In the recent update, I changed to to the same for control rails as data rails. I believe you are right, I should not make any assumption here even if only the 0th control rail is used later, which is the decision of that code.

src/utils/libfabric/libfabric_rail_manager.h
fengjica: Done.

PR #915, opened 2025-10-16
Bindings: package separate wheels for CUDA 12 and 13

ovidiusm:
## What?
Add support for building separate nixl wheels for CUDA 12 and CUDA 13.

## Why?

NIXL binaries are linked against libcudart.so.12 or libcudart.so.13. The two major versions are incompatible. Therefore we need separate wheels.

## How?
* Base image (CUDA 12 or CUDA 13 manylinux image) chosen by pipeline orchestration.
* Rename package in pyproject.toml to generate nixl-cu12 or nixl-cu13 wheels.
* Adjustments to the uv scripts so that we can control which pytorch version we install.
* Removed all instances of uv run that were not needed.
* GPUNETIO plugin build had to be disabled since a dependency from DOCA 3.1 is not compatible with CUDA 13

`uv run script.py` is not desirable, because behind the scenes it:
* Searches for a .venv or creates a temporary one automatically if it cannot find it, which in most cases we do not need
* Installs/upgrades/downgrades packages automatically based on heuristics on what script.py imports, but since pytorch is installed from a different index, it does not know about that and tries to pull it from the default 12.x index, which installs CUDA 12 binaries, which we do not want
* Without the above, it is better to activate the venv and run the scripts directly with python since we have full control of the package installation

## Tests

Tested:
* Build works with one or several python versions
* Build works on x86_64 and aarch64
* Package installs, loads and can run a transfer on nvidia/cuda:12.9.1-runtime-ubuntu24.04 and nvidia/cuda:13.0.1-runtime-ubuntu24.04 images
* Python client code does not change, `import nixl` and all other imports still work exactly as before
* Pypi test index:

```
$ uv pip install --index-url https://test.pypi.org/simple/ "nixl[cu12]"
Using Python 3.11.13 environment at: /home/ubuntu/.venv
Resolved 28 packages in 2.19s
Prepared 2 packages in 4.08s
Installed 2 packages in 40ms
 + nixl==0.6.1
 + nixl-cu12==0.6.1
 
$ uv pip uninstall nixl
Using Python 3.11.13 environment at: /home/ubuntu/.venv
Uninstalled 1 package in 53ms
 - nixl==0.6.1
 
$ uv pip install --index-url https://test.pypi.org/simple/ "nixl[cu13]"
Using Python 3.11.13 environment at: /home/ubuntu/.venv
Resolved 28 packages in 2.15s
Prepared 1 package in 4.18s
Installed 2 packages in 172ms
 + nixl==0.6.1
 + nixl-cu13==0.6.1
```

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -214,6 +214,7 @@ RUN cd src/bindings/rust && cargo build --release --locked
 
 # Build wheel using the build-wheel.sh script for better UCX plugin bundling and library management
 RUN export PATH=$VIRTUAL_ENV/bin:$PATH && \
+    mkdir -p dist && \
     ./contrib/build-wheel.sh \
     --python-version $DEFAULT_PYTHON_VERSION \
     --platform manylinux_2_39_$ARCH \
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -269,6 +269,7 @@ ARG WHL_PYTHON_VERSIONS="3.9,3.10,3.11,3.12"
 ARG WHL_PLATFORM="manylinux_2_28_$ARCH"
 RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
     export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d .)" && \
+    mkdir -p dist && \
     for PYTHON_VERSION in "${PYTHON_VERSIONS[@]}"; do \
         export PATH=$VIRTUAL_ENV/bin:$PATH && \
         ./contrib/build-wheel.sh \


contrib/Dockerfile
```diff
@@ -167,23 +167,41 @@ RUN wget --tries=3 --waitretry=5 --timeout=30 --read-timeout=60 \
...
-    uv pip install --upgrade "meson>=0.64.0" pybind11 patchelf
-
 # Install pybind11 via apt
 RUN apt-get update && apt-get install -y --no-install-recommends pybind11-dev
 
 ENV NIXL_PREFIX=$NIXL_PREFIX
 RUN rm -rf build && \
     mkdir build && \
-    uv run meson setup -Dlibfabric_path=$LIBFABRIC_INSTALL_PATH build/ --prefix=$NIXL_PREFIX && \
+    meson setup -Dlibfabric_path=$LIBFABRIC_INSTALL_PATH build/ --prefix=$NIXL_PREFIX && \
```
aranadive, line 204: Did we get an updated version of meson? AFAIK the cuda 12 base container had meson deb pkg older than our requirement.

ovidiusm: We are picking up meson from the virtual environment, not from the base image. The venv is  activated above with `ENV PATH="$VIRTUAL_ENV/bin:$PATH"`

aranadive: oh right

meson.build
```diff
@@ -101,10 +101,10 @@ if cuda_dep.found()
     nvcc_flags_link += ['-gencode=arch=compute_90,code=sm_90']
     add_project_link_arguments(nvcc_flags_link, language: 'cuda')
     message('nvcc version: ' + nvcc.version())
-    if nvcc.version().version_compare('>=12.8')
+    if nvcc.version().version_compare('>=12.8') and nvcc.version().version_compare('<13.0')
         doca_gpunetio_dep = dependency('doca-gpunetio', required : false)
     else
-        warning('CUDA version is less than 12.8, GPUNETIO plugin will be disabled')
+        warning('GPUNETIO plugin not supported in CUDA version: ' + nvcc.version())
```
aranadive, line 107: This works on CUDA 13 but will be disabled?

aranadive (approved):

meson.build
ovidiusm: We need to upgrade DOCA to 3.2 since one of their libs is linking to libcudart.so.12 instead of libcuda.so.1, but it has not been released yet. Probably in November.

Until then I agreed with the plugin owner @e-ago to disable the build on CUDA 13

src/bindings/python/nixl-meta/meson.build
```diff
@@ -0,0 +1,31 @@
...
+
+source_dir = meson.current_source_dir()
+build_dir = meson.current_build_dir()
+
+uv = find_program('uv', required: false)
+
+if uv.found()
+  meta_wheel = custom_target(
+    'build_nixl_meta',
+    output: ['nixl-0.6.1-py3-none-any.whl'],
```
ovidiusm: Should try to find a better way than hardcoding the version here and in related toml file

roiedanino:
@ryanolson Any modifications needed on Dynamo side?

src/bindings/python/nixl-meta/meson.build
roiedanino: I think we can take the version from a variable or `meson.project_version()`

contrib/build-wheel.sh
```diff
@@ -73,23 +73,23 @@ done
...
+    echo "Invalid CUDA_MAJOR: '$CUDA_MAJOR'"
+    exit 1
+fi
+PKG_NAME="nixl-cu${CUDA_MAJOR}"
+./contrib/tomlutil.py --set-name $PKG_NAME pyproject.toml
 uv build --wheel --out-dir $TMP_DIR --python $PYTHON_VERSION
 
 # Bundle libraries
-uv pip install auditwheel patchelf
+auditwheel repair --exclude 'libcuda*' --exclude 'libcufile*' --exclude 'libssl*' --exclude 'libcrypto*' --exclude 'libefa*' --exclude 'libhwloc*' --exclude 'libfabric*' $TMP_DIR/nixl-*.whl --plat $WHL_PLATFORM --wheel-dir $OUTPUT_DIR
```
roiedanino: Gitlab CI:
16.44 Successfully built /tmp/tmp.2vxafjn15j/nixl_cu12-0.6.1-cp312-cp312-linux_x86_64.whl
16.45 + auditwheel repair --exclude 'libcuda*' --exclude 'libcufile*' --exclude 'libssl*' --exclude 'libcrypto*' --exclude 'libefa*' --exclude 'libhwloc*' --exclude 'libfabric*' '/tmp/tmp.2vxafjn15j/nixl-*.whl' --plat manylinux_2_39_x86_64 --wheel-dir /workspace/nixl/dist
16.59 usage: auditwheel [-h] [-V] [-v] command ...
16.59 auditwheel: error: cannot access /tmp/tmp.2vxafjn15j/nixl-*.whl. No such file

ovidiusm: Fixed

src/bindings/python/nixl-meta/meson.build
ovidiusm: Fixed

src/bindings/python/nixl-meta/README.md
```diff
@@ -0,0 +1,9 @@
+# nixl
+
+This is a *meta package* that declares optional dependencies on CUDA variants.
+
+Install one of:
+```bash
+pip install "nixl[cu12]"   # for CUDA 12
```
aranadive, line 7: What does pip install nixl now do? Does it install both these wheels?

mkhazraee (approved):

src/bindings/python/nixl-meta/README.md
ovidiusm: It does not install either. I will try to add a message in a separate PR.

PR #917, opened 2025-10-16
nixlbench: Provide remote iovs for other storage backends (#913)

ovidiusm:
Cherry pick of #913

Fix the if else when creating remote iovs for storage backends


aranadive:
Closing as dup of #916

PR #919, opened 2025-10-17
Flush out commQueue before stopping listener (device-api)

2dm:
## What?
Verify that all enqueued work has completed. 

## Why?
There is a race between the agent destructor and jobs on `commQueue` since the background thread might finish without checking for existing work in the queue. In particular, the missed job is probably `invalidateLocalMD` that is called right before, which will keep the MD intact and prevent the re-addition of that rank later. 

## How?
1. Stop enqueueing
2. Check queue is empty (the thread is processing or completed the last work that was enqueued)
3. Only then set `commThreadStop`

### DIFF
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -90,8 +90,8 @@ class nixlAgentData {
         std::thread commThread;
         std::vector<nixl_comm_req_t> commQueue;
         std::mutex commLock;
-        bool commThreadStop;
-        bool agentShutdown;
+        std::atomic<bool> commThreadStop;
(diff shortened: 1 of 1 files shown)


2dm:
@michal-shalev can you confirm if base of this branch (a241f80) is not expected to pass gtest at this time?

michal-shalev:
@2dm right. Currently, the Device API gtest is broken in NIXL due to a wireup issue that needs to be fixed in UCX.

2dm:
this change will target main branch (PR #830). Closing this one.

PR #924, opened 2025-10-19
CI: Switch from PyTorch to cuda-dl-base images for unification

Alexey-Rivkin:
## What?
Switch base images from PyTorch to cuda-dl-base for both Ubuntu versions:
- Ubuntu 24.04: `pytorch:25.02-py3` → `cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04`
- Ubuntu 22.04: `pytorch:24.10-py3` → `cuda-dl-base:24.10-cuda12.6-devel-ubuntu22.04`

## Why?
Unify base image usage across the project. Production Dockerfiles already use cuda-dl-base; this aligns CI/test images with that standard while maintaining the same CUDA versions (12.9 for Ubuntu 24, 12.6 for Ubuntu 22).

## How?
- Updated base image references in CI Dockerfiles, build matrix, test matrix, and AWS EFA configs
- Added `python3` and `python3-pip` installation (not included in cuda-dl-base by default)

### DIFF
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -44,6 +44,8 @@ env:
   INSTALL_DIR: ${CONTAINER_WORKSPACE}/nixl_install
   # Manual timeout - ci-demo doesn't handle docker exec
   TEST_TIMEOUT: 30
+  # NPROC for bare-metal: containers see all host CPUs, need to limit parallelism
+  NPROC: 16
 
 steps:
   - name: Get Environment Info


.ci/dockerfiles/Dockerfile.gpu_test
```diff
@@ -13,7 +13,7 @@
 #   docker run --gpus all --privileged -it nixl-gpu-test
 #
 # Build arguments:
-#   BASE_IMAGE: Base NVIDIA PyTorch image (default: nvcr.io/nvidia/pytorch:25.02-py3)
+#   BASE_IMAGE: Base NVIDIA cuda-dl-base image (default: nvcr.io/nvidia/cuda-dl-base:25.02-cuda12.8-devel-ubuntu24.04)
```
dpressle: wrong tag, should be 25.06-cuda12.9-devel-ubuntu24.04

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -6,7 +6,7 @@
 # Key Components:
 # - Job Configuration: Defines timeout, failure behavior, and Kubernetes resources
 # - Docker Images: Specifies the container images used for different build stages
-#   - PyTorch images (24.10 and 25.02) for building and testing
+#   - cuda-dl-base images (25.02 for Ubuntu 24.04, 24.10 for Ubuntu 22.04) for building and testing
```
dpressle: 25.06 not 25.02

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -46,18 +46,14 @@ matrix:
 
 env:
   NIXL_INSTALL_DIR: /opt/nixl
-  TEST_TIMEOUT: 30
+  TEST_TIMEOUT: 60
```
dpressle: why change? no test is running more then 10 minutes so old 30 minutes timeout is still valid.

.gitlab/build.sh
```diff
@@ -53,11 +56,16 @@ fi
 ARCH=$(uname -m)
 [ "$ARCH" = "arm64" ] && ARCH="aarch64"
 
+# Make /workspace writable (if it exists)
+[ -d /workspace ] && $SUDO chmod 777 /workspace
```
dpressle: This is very CI implementation specific it does not make sense to do this change here if its not relevant to 99% of usages of this build script.
lets change the workspace folder name in test-matrix or chmod it in gpu dockerfile

.gitlab/build.sh
```diff
@@ -210,11 +229,11 @@ export UCX_TLS=^cuda_ipc
 
 # shellcheck disable=SC2086
 meson setup nixl_build --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Dbuild_docs=true -Drust=false ${EXTRA_BUILD_ARGS} -Dlibfabric_path="${LIBFABRIC_INSTALL_DIR}"
-ninja -C nixl_build && ninja -C nixl_build install
+ninja -j${NPROC:-$(nproc)} -C nixl_build && ninja -j${NPROC:-$(nproc)} -C nixl_build install
```
dpressle: instead of duplicating this so many times just have this definition in common script and use parameter here

.gitlab/build.sh
```diff
@@ -210,11 +229,11 @@ export UCX_TLS=^cuda_ipc
...
 
 # TODO(kapila): Copy the nixl.pc file to the install directory if needed.
 # cp ${BUILD_DIR}/nixl.pc ${INSTALL_DIR}/lib/pkgconfig/nixl.pc
 
 cd benchmark/nixlbench
 meson setup nixlbench_build -Dnixl_path=${INSTALL_DIR} -Dprefix=${INSTALL_DIR}
-ninja -C nixlbench_build && ninja -C nixlbench_build install
+ninja -j${NPROC:-$(nproc)} -C nixlbench_build && ninja -j${NPROC:-$(nproc)} -C nixlbench_build install
```
dpressle: same as above

Alexey-Rivkin:
> what about contrib/Dockerfile ? should we change default base image there as well? what about build-container job defaults ?

Changed all to unify the build base
09a6dd9

.gitlab/build.sh
Alexey-Rivkin: 83f5ed5

.gitlab/build.sh
Alexey-Rivkin: 83f5ed5

.gitlab/build.sh
Alexey-Rivkin: 7a095ad

.ci/jenkins/lib/build-matrix.yaml
Alexey-Rivkin: dd06177

.ci/jenkins/lib/build-matrix.yaml
Alexey-Rivkin: 09a6dd9

.ci/dockerfiles/Dockerfile.gpu_test
Alexey-Rivkin: 09a6dd9

.gitlab/build.sh
```diff
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Ensure CUDA_HOME is set (cuda-dl-base images don't set it by default)
+export CUDA_HOME=${CUDA_HOME:-/usr/local/cuda}
```
dpressle: this seems like it should be in common script its something we might need not only on build.
And i would check if it exists before exporting it.

.ci/scripts/common.sh
```diff
@@ -89,3 +89,6 @@ fi
 
 # Default to false, unless TEST_LIBFABRIC is set. AWS EFA tests must set it to true.
 export TEST_LIBFABRIC=${TEST_LIBFABRIC:-false}
+
+# Set default parallelism for make/ninja (can be overridden by NPROC env var)
+export NPROC=${NPROC:-$(nproc)}
```
dpressle: if we want to do it right for containers the calculation should be different:
```bash
if [[ -f /.dockerenv  ||  -f /run/.containerenv  ||  -n "${KUBERNETES_SERVICE_HOST}" ]]; then
    if [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
        limit=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
    elif [ -f /sys/fs/cgroup/memory.max/memory.limit_in_bytes ]; then
        limit=$(cat /sys/fs/cgroup/memory.max/memory.limit_in_bytes)
    else
        limit=$((4 * 1024 * 1024 * 1024))
    fi
    # use 1 process for 1GB of memory
    nproc=$((${limit} / (1024 * 1024 * 1024)))
    # max 16 processes per container
    nproc=$(( $nproc > 16 ? 16 : $nproc ))
else
    nproc=$(nproc --all)
fi
```

Alexey-Rivkin: f606793

.gitlab/build.sh
Alexey-Rivkin: f606793

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -42,7 +42,6 @@ taskName: "${name}/${arch}/ucx-${ucx_version}/${axis_index}"
 env:
   CONTAINER_WORKSPACE: /workspace
   INSTALL_DIR: ${CONTAINER_WORKSPACE}/nixl_install
-  NPROC: "16"
```
dpressle, line 45: you cant remove this on bare-metal hosts because its not k8s and you will get the whole cous on each container (up to 4 at same time) which is wrong, so either keep the 16 NPROC definition or limit the container's cpu/mem in the docker run command and then your function will work

Alexey-Rivkin: fac210e

.ci/scripts/common.sh
Alexey-Rivkin: f606793

ovidiusm (approved):

PR #928, opened 2025-10-20
NIXL meta-package

ovidiusm:
## What?
Create a meta package for Python that depends on nixl-cu12 and nixl-cu13

## Why?
CUDA 13 support requires additional wheel packages. This meta package ensures that `pip install nixl` does not pull the old CUDA 12 package. Instead the user should run `pip install nixl[cu12]` or `pip install nixl[cu13]`.

### DIFF
--- a/.ci/dockerfiles/Dockerfile.gpu_test
+++ b/.ci/dockerfiles/Dockerfile.gpu_test
@@ -32,7 +32,7 @@ ARG _GID=30
 ARG _LOGIN=svc-nixl
 ARG _GROUP=hardware
 ARG _HOME=/home/$_LOGIN
-ARG WORKSPACE
+ARG WORKSPACE=/workspace
 
 # Labels for documentation
 LABEL maintainer="NVIDIA NIXL Team"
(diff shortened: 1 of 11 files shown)


src/bindings/python/nixl-meta/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.6.1"
+description = "Meta package for nixl CUDA variants"
+readme = "README.md"
+requires-python = ">=3.9"
+
+# No direct dependencies — only optional ones
+dependencies = []
+
+[project.optional-dependencies]
+cu12 = ["nixl-cu12>=0.6.1"]
```
brminich, line 31: should it be `0.7.1`?

ovidiusm: Yes it should. main generates 0.6.1 at the moment, that's why, but I will change it

ovidiusm:
Easier to merge it into https://github.com/ai-dynamo/nixl/pull/915 since it seems I have to solve the same scripting issues

PR #944, opened 2025-10-22
nixlbench: Fix timer for PostXferReq

aranadive:
The timer for measuring PostXferReq should be moved to place after callong PostXferReq


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1178,7 +1181,7 @@ execTransferIterations(nixlAgent *agent,
         // Standard path: Single request for all iterations
         for (int i = 0; i < num_iter; ++i) {
             thread_stats.post_duration.add(timer.lap());
```
ovidiusm: If we measure post duration inside the function, shouldn't we remove it from here?

aranadive: Done

PR #946, opened 2025-10-23
PLUGINS/UCX: Removed CUDA context management.

rakhmets:
## What?
Removed CUDA context management from UCX backend.

## Why?
The CUDA context setting requirements have been relaxed since version 1.19.0 of UCX.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -87,17 +87,18 @@ public:
     void
     reserve(size_t size) {
         requests_.reserve(size);
+        NIXL_ASSERT(connections_.empty());
     }
 
     nixl_status_t
     append(nixl_status_t status, nixlUcxReq req, ucx_connection_ptr_t conn) {
-        connections_.insert(conn);
         switch (status) {
         case NIXL_IN_PROG:
             requests_.push_back(req);
+            connections_.insert(conn);
             break;
         case NIXL_SUCCESS:
-            // Nothing to do
+            connections_.insert(conn);
             break;
         default:
             // Error. Release all previously initiated ops and exit:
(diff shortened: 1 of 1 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -442,6 +442,9 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
     unsigned ucp_version = UCP_VERSION(major_version, minor_version);
     if (ucp_version >= UCP_VERSION(1, 19)) {
         config.modify ("MAX_COMPONENT_MDS", "32");
+    } else {
+        NIXL_INFO
+            << "UCX version is less than 1.19, multi-GPU within a single process is not supported";
```
brminich: it is not only that, probably cuda may not work properly in general, because older UCX expects some context to be set

rakhmets: Updated the message.

brminich:
can we introduce some tests before merging it?
i barely remember there was an issue when some thread tried to perform some operations without any active context set.

rakhmets:
> can we introduce some tests before merging it? i barely remember there was an issue when some thread tried to perform some operations without any active context set.

There is a test. The test from this PR https://github.com/ai-dynamo/nixl/pull/336 had previously failed without changes related to manipulating the CUDA context in NIXL and without relaxing the requirements for setting the CUDA context in UCX.

brminich:
@ovidiusm is there any docker files related implications in removing cuda dep?

rakhmets:
> @ovidiusm is there any docker files related implications in removing cuda dep?

This PR doesn't completely eliminate NIXL's dependence on CUDA libraries. It only removes the UCX plugin dependency on CUDA. There are still plugins that depend on CUDA libraries. Thus, I don't expect any changes to the docker files.

PR #947, opened 2025-10-23
DEVICE/API: Wait for wireup completion in createGpuXferReq

michal-shalev:
## What?
Add blocking wait for endpoint wireup completion in `createGpuXferReq()`.

## Why?
Previously, users had to implement workarounds in their applications to wait for wireup completion before calling `createGpuXferReq()` (as shown in UCX tests). 
This PR moves the wireup handling into the library, simplifying the API and removing the burden from application code.

## How?
- Retry `ucp_device_mem_list_create()` in a loop while it returns `UCS_ERR_NOT_CONNECTED`
- Document the blocking behavior in the public API

### DIFF
--- a/.ci/jenkins/lib/build-container-matrix.yaml
+++ b/.ci/jenkins/lib/build-container-matrix.yaml
@@ -16,7 +16,7 @@ kubernetes:
   requests: "{memory: 8Gi, cpu: 4000m}"
 
 runs_on_dockers:
-  - { name: "podman-v5.0.2", url: "quay.io/podman/stable:v5.0.2", privileged: true }
+  - { name: "podman-v5.6.2", url: "quay.io/podman/stable:v5.6.2", privileged: true }
 
 # Build matrix
 matrix:
@@ -57,8 +57,6 @@ steps:
   - name: Prepare
     run: |
       # Setup podman and dependencies
-      rm -f /etc/containers/storage.conf
-      podman system reset -f || true
       ln -sfT $(type -p podman) /usr/bin/docker
       yum install -y git gettext
 
--- a/.ci/jenkins/lib/build-matrix.yaml
+++ b/.ci/jenkins/lib/build-matrix.yaml
@@ -37,7 +37,7 @@ runs_on_dockers:
   - { name: "ubuntu24.04-cuda12-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04" }
   - { name: "ubuntu24.04-cuda13-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:25.10-cuda13.0-devel-ubuntu24.04" }
   - { name: "ubuntu22.04-cuda-dl-base", url: "nvidia/cuda:13.0.1-devel-ubuntu22.04" }
-  - { name: "podman-v5.0.2", url: "quay.io/podman/stable:v5.0.2", category: 'tool', privileged: true }
+  - { name: "podman-v5.6.2", url: "quay.io/podman/stable:v5.6.2", category: 'tool', privileged: true }
 
 matrix:
   axes:
@@ -84,8 +84,6 @@ steps:
     parallel: false
     containerSelector: "{ name: 'podman.*' }"
     run: |
-      # change storage driver to improve build performance
-      rm -f /etc/containers/storage.conf ; podman system reset -f || true
       # symlink podman to docker - scripts works with docker commands
       ln -sfT $(type -p podman) /usr/bin/docker
       # install git for building container image
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -144,7 +144,7 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/${UCX_VERSION}" | tar xz
 ( \
   cd openucx-ucx* && \
   ./autogen.sh && \
-  ./configure \
+  ./contrib/configure-release-mt \
           --prefix="${UCX_INSTALL_DIR}" \
           --enable-shared \
           --disable-static \
@@ -154,8 +154,7 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/${UCX_VERSION}" | tar xz
           --enable-devel-headers \
           --with-verbs \
           --with-dm \
-          ${UCX_CUDA_BUILD_ARGS} \
-          --enable-mt && \
+          ${UCX_CUDA_BUILD_ARGS} && \
         make -j && \
         make -j install-strip && \
         $SUDO ldconfig \
--- a/.gitlab/test_cpp.sh
+++ b/.gitlab/test_cpp.sh
@@ -41,6 +41,10 @@ export PATH=${INSTALL_DIR}/bin:$PATH
 export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
 export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
 
+# Set UCX GDA max system latency to allow GDA on SYS topology
+# TODO: Remove this once CI setups have better GPU-NIC locality
+# export UCX_IB_GDA_MAX_SYS_LATENCY=1us
+
 echo "==== Show system info ===="
 env
 nvidia-smi topo -m || true
@@ -48,11 +52,35 @@ ibv_devinfo || true
 uname -a || true
 cat /sys/devices/virtual/dmi/id/product_name || true
 
+echo "==== NVIDIA Peermem check ===="
+if ! lsmod | grep -q nvidia_peermem; then
+    echo "nvidia_peermem module not loaded"
+fi
+
+if [ -f /sys/kernel/mm/memory_peers/nv_mem/version ]; then
+    cat /sys/kernel/mm/memory_peers/nv_mem/version
+else
+    echo "/sys/kernel/mm/memory_peers/nv_mem/version not found "
+fi
+
+if [ -f /sys/module/nvidia_peermem/version ]; then
+    cat /sys/module/nvidia_peermem/version
+else
+    echo "/sys/module/nvidia_peermem/version not found"
+fi
+
+if [ -f /sys/module/nv_peer_mem/version ]; then
+    cat /sys/module/nv_peer_mem/version
+else
+    echo "/sys/module/nv_peer_mem/version not found"
+fi
+
 echo "==== Running ETCD server ===="
 etcd_port=$(get_next_tcp_port)
 etcd_peer_port=$(get_next_tcp_port)
 export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:${etcd_port}"
 export NIXL_ETCD_PEER_URLS="http://127.0.0.1:${etcd_peer_port}"
+export NIXL_ETCD_NAMESPACE="/nixl/cpp_ci/${etcd_port}"
 etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} \
      --listen-peer-urls ${NIXL_ETCD_PEER_URLS} --initial-advertise-peer-urls ${NIXL_ETCD_PEER_URLS} \
      --initial-cluster default=${NIXL_ETCD_PEER_URLS} &
@@ -88,7 +116,8 @@ kill -s INT $telePID
 # fi
 
 # shellcheck disable=SC2154
-gtest-parallel --workers=1 --serialize_test_cases ./bin/gtest -- --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port"
+# TODO: enable PrepGpuSignal and ucxDeviceApi tests once the problem in UCX is fixed
+gtest-parallel --workers=1 --serialize_test_cases ./bin/gtest -- --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port" --gtest_filter=-*PrepGpuSignal*:*ucxDeviceApi*
 ./bin/test_plugin
 
 # Run NIXL client-server test
--- a/.gitlab/test_nixlbench.sh
+++ b/.gitlab/test_nixlbench.sh
@@ -50,6 +50,7 @@ etcd_port=$(get_next_tcp_port)
 etcd_peer_port=$(get_next_tcp_port)
 export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:${etcd_port}"
 export NIXL_ETCD_PEER_URLS="http://127.0.0.1:${etcd_peer_port}"
+export NIXL_ETCD_NAMESPACE="/nixl/nixlbench_ci/${etcd_port}"
 etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} \
      --listen-peer-urls ${NIXL_ETCD_PEER_URLS} --initial-advertise-peer-urls ${NIXL_ETCD_PEER_URLS} \
      --initial-cluster default=${NIXL_ETCD_PEER_URLS} &
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -63,6 +63,7 @@ etcd_port=$(get_next_tcp_port)
 etcd_peer_port=$(get_next_tcp_port)
 export NIXL_ETCD_ENDPOINTS="http://127.0.0.1:${etcd_port}"
 export NIXL_ETCD_PEER_URLS="http://127.0.0.1:${etcd_peer_port}"
+export NIXL_ETCD_NAMESPACE="/nixl/python_ci/${etcd_port}"
 etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} \
      --listen-peer-urls ${NIXL_ETCD_PEER_URLS} --initial-advertise-peer-urls ${NIXL_ETCD_PEER_URLS} \
      --initial-cluster default=${NIXL_ETCD_PEER_URLS} &
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ $ git clone https://github.com/openucx/ucx.git
 $ cd ucx
 $ git checkout v1.20.x
 $ ./autogen.sh
-$ ./configure                          \
+$ ./contrib/configure-release-mt       \
     --enable-shared                    \
     --disable-static                   \
     --disable-doxygen-doc              \
@@ -65,8 +65,7 @@ $ ./configure                          \
     --with-cuda=<cuda install>         \
     --with-verbs                       \
     --with-dm                          \
-    --with-gdrcopy=<gdrcopy install>   \
-    --enable-mt
+    --with-gdrcopy=<gdrcopy install>
 $ make -j
 $ make -j install-strip
 $ ldconfig
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -263,7 +263,8 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
         exit(EXIT_FAILURE);
     }
 
-    agent->createBackend(backend_name, backend_params, backend_engine);
+    CHECK_NIXL_ERROR(agent->createBackend(backend_name, backend_params, backend_engine),
+                     "createBackend failed!");
 }
 
 xferBenchNixlWorker::~xferBenchNixlWorker() {
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -123,7 +123,8 @@ RUN cd /usr/local/src && \
      git clone https://github.com/openucx/ucx.git && \
      cd ucx && 			     \
      git checkout $UCX_REF &&        \
-     ./autogen.sh && ./configure     \
+     ./autogen.sh && \
+     ./contrib/configure-release-mt  \
          --prefix=$UCX_PREFIX        \
          --enable-shared             \
          --disable-static            \
@@ -135,8 +136,7 @@ RUN cd /usr/local/src && \
          --with-verbs                \
          --with-dm                   \
          --with-gdrcopy=/usr/local   \
-         --with-efa                  \
-         --enable-mt &&              \
+         --with-efa  &&              \
      make -j${NPROC:-$(nproc)} &&                \
      make -j${NPROC:-$(nproc)} install-strip &&  \
      ldconfig
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -188,7 +188,8 @@ RUN cd /usr/local/src && \
      git clone https://github.com/openucx/ucx.git && \
      cd ucx && 			     \
      git checkout $UCX_REF &&	     \
-     ./autogen.sh && ./configure     \
+     ./autogen.sh && \
+     ./contrib/configure-release-mt  \
          --enable-shared             \
          --disable-static            \
          --disable-doxygen-doc       \
@@ -199,8 +200,7 @@ RUN cd /usr/local/src && \
          --with-verbs                \
          --with-dm                   \
          --with-gdrcopy=/usr/local   \
-         --with-efa                  \
-         --enable-mt &&              \
+         --with-efa  &&              \
      make -j &&                      \
      make -j install-strip &&        \
      ldconfig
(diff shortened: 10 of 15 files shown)


src/api/cpp/nixl.h
```diff
@@ -325,6 +325,9 @@ class nixlAgent {
          * @param  req_hndl     [in]  Transfer request obtained from makeXferReq/createXferReq
          * @param  gpu_req_hndl [out] GPU transfer request handle
          * @return nixl_status_t Error code if call was not successful
+         *
+         * @note   This call may block if the endpoint wireup is not yet complete.
+         *         It will progress the worker until the connection is established.
```
rakhmets: `the endpoint wireup` is terminology related to the UCX library. I'm not sure it's appropriate to mention this here without extensive clarification. I was thinking of advising you to use only `connection establishment`. However, it also may be confusing since there is `nixlAgent::makeConnection`, and the user may think that this will make this call non-blocking...
It is also not entirely clear what kind of `worker` we are talking about here, without diving into the implementation of the method in a specific backend.

IMO, this note raises more questions than it explains.

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -33,6 +33,7 @@ namespace nixl::ucx {
 
 nixlGpuXferReqH
 createGpuXferReq(const nixlUcxEp &ep,
+                 nixlUcxWorker &worker,
```
brminich, line 65: can't we get it from nixlUcxEp?

src/api/cpp/nixl.h
michal-shalev: I've revised the documentation to avoid backend-specific terminology.

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: There's no worker in `nixlUcxEp` class currently. The `nixlUcxEp` is created by `nixlUcxWorker::connect()`, which passes the raw `ucp_worker_h` handle to the endpoint constructor, but it's not saved.
Getting the worker from the ep would require changing `nixlUcxEp`.

src/api/cpp/nixl.h
```diff
@@ -325,6 +325,8 @@ class nixlAgent {
          * @param  req_hndl     [in]  Transfer request obtained from makeXferReq/createXferReq
          * @param  gpu_req_hndl [out] GPU transfer request handle
          * @return nixl_status_t Error code if call was not successful
+         *
+         * @note   This call may block briefly until the associated connection is established.
```
ofirfarjun7: I will remove the "briefly".

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: Added `ucp_worker_h` and a getter to `nixlUcxEp`

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +74,15 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    ucp_worker_h worker = ep.getWorker();
     ucp_device_mem_list_handle_h ucx_handle;
-    ucs_status_t ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle);
+    ucs_status_t ucs_status;
+    // Workaround: loop until wireup is completed
+    while ((ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle)) ==
```
ofirfarjun7, line 113: Maybe we should add a timeout?

src/utils/ucx/ucx_utils.h
```diff
@@ -93,6 +93,7 @@ class nixlUcxEp {
     };
 private:
     ucp_ep_h            eph{nullptr};
+    ucp_worker_h worker{nullptr};
```
rakhmets: This is contrary to the C++ Core Guidelines.
`ucp_worker_h` is a raw pointer. `nixlUcxWorker` owns the pointer using `unique_ptr`. Other classes should not own this raw pointer.

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: added

src/api/cpp/nixl.h
michal-shalev: removed

src/utils/ucx/ucx_utils.h
brminich: what is your suggestion to get worker for the corresponding ep?

rakhmets: Whoever creates `nixlUcxWorker` and `nixlUcxEp` is responsible for ensuring consistency between objects.
In this case (`createGpuXferReq`), it is better to pass worker as a parameter.

brminich: i see your point
what would be the best way for ep to refer to the worker as it seems to be logically correct to have such a reference?

rakhmets: Do you mean `nixlUcxEp`?
`nixlUcxEp` is a class that proxies `ucp_ep` structure. It owns the pointer to the structure (should be `unique_ptr`). And proxies UCP API involving `ucp_ep`, e.g. `nixlUcxEp::estimateCost(...)` -> `ucp_ep_evaluate_perf(m_ep, ...)`.

`nixlUcxWorker` is a parameter for `nixlUcxEp` creation. The reason why `nixlUcxWorker` seems to be more than just a parameter, and the `nixlUcxEp` must contain a reference to it, lies in the knowledge of the implementation of `ucp_ep` structure, IMO.

If we really want to refer to `nixlUcxWorker` from `nixlUcxEp`, then:
1. `nixlUcxEngine` should create and own `nixlUcxWorker` via a shared pointer.
2. `nixlUcxEngine` must share the ownership of `nixlUcxWorker` with `nixlUcxEp` during the creation of `nixlUcxEp`.

michal-shalev: I pushed a new commit, please check it

rakhmets: It's the same issue.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i11-never-transfer-ownership-by-a-raw-pointer-t-or-reference-t
Now `nixlUcxEp` owns `nixlUcxWorker` via reference.

rakhmets: > If we really want to refer to `nixlUcxWorker` from `nixlUcxEp`, then:
> 
> 1. `nixlUcxEngine` should create and own `nixlUcxWorker` via a shared pointer.
> 2. `nixlUcxEngine` must share the ownership of `nixlUcxWorker` with `nixlUcxEp` during the creation of `nixlUcxEp`.

But I don't recommend it, since it overcomplicates things. I'd recommend just to pass reference to `nixlUcxWorker` to `createGpuXferReq`.

michal-shalev: Okay, @rakhmets @brminich can we have a quick call to decide? I'm getting questions from architects about this PR.

brminich: ok, let's get back to the original approach @michal-shalev

michal-shalev: done @brminich @rakhmets

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: decided to return to passing the worker

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,16 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    auto start = std::chrono::steady_clock::now();
+    constexpr auto timeout = std::chrono::milliseconds(500);
```
brminich: how is it used?

michal-shalev: added back the timeout

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,20 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    auto start = std::chrono::steady_clock::now();
```
rakhmets:
```suggestion
    const auto start = std::chrono::steady_clock::now();
```

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: resolved? @brminich

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,20 @@ createGpuXferReq(const nixlUcxEp &ep,
...
     ucp_device_mem_list_handle_h ucx_handle;
-    ucs_status_t ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle);
+    ucs_status_t ucs_status;
+    // Workaround: loop until wireup is completed
+    while ((ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle)) ==
+           UCS_ERR_NOT_CONNECTED) {
+        worker.progress();
+
+        if (std::chrono::steady_clock::now() - start > timeout) {
+            throw std::runtime_error("Timeout waiting for endpoint wireup completion");
```
rakhmets: `... has been exceeded`?

michal-shalev: please do not approve PRs if you have comments, I can't merge without resolving all the comments

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: please do not approve PRs if you have comments, I can't merge without resolving all the comments

rakhmets: Ok.
Please add `const`.

michal-shalev: thanks :)
added

rakhmets (dismissed):

brminich (dismissed):

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,23 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    const auto start = std::chrono::steady_clock::now();
+    constexpr auto timeout = std::chrono::seconds(5);
```
rakhmets: What do you think about making it configurable via environment variable?

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,23 @@ createGpuXferReq(const nixlUcxEp &ep,
...
+    while ((ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle)) ==
+           UCS_ERR_NOT_CONNECTED) {
+        for (const auto &w : all_workers) {
+            w->progress();
+        }
+
+        if (std::chrono::steady_clock::now() - start > timeout) {
+            throw std::runtime_error(
+                "Timeout waiting for endpoint wireup completion has been exceeded");
+        }
```
rakhmets, line 119: I think it makes sense to swap the time check and the execution of the progress on the workers. Otherwise, we may throw the exception even when the wireup is completed on this iteration.

Optional. I'd prefer to do a time loop. E.g.:
```C++
for (const auto start = std::chrono::steady_clock::now();
     std::chrono::steady_clock::now() - start <= timeout;)
{
    status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle);
    if (status != UCS_ERR_NOT_CONNECTED) {
        break;
    }

    for (const auto &w : workers) {
        w->progress();
    }
}

if (status == UCS_ERR_NOT_CONNECTED) {
    throw std::runtime_error("Timeout waiting for endpoint wireup completion has been exceeded");
} else if (status != UCS_OK) {
    throw std::runtime_error(std::string("Failed to create device memory list: ") +
                             ucs_status_string(ucs_status));
}
```

src/utils/ucx/gpu_xfer_req_h.h
```diff
@@ -19,17 +19,20 @@
...
 class nixlUcxEp;
 class nixlUcxMem;
+class nixlUcxWorker;
 
 namespace nixl::ucx {
 class rkey;
 
 nixlGpuXferReqH
 createGpuXferReq(const nixlUcxEp &ep,
+                 const std::vector<std::unique_ptr<nixlUcxWorker>> &all_workers,
```
rakhmets: It looks like `all_` is redundant for this parameter. `workers` would be enough.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1103,7 +1103,8 @@ nixlUcxEngine::create(const nixlBackendInitParams &init_params) {
 
 nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     : nixlBackendEngine(&init_params),
-      sharedWorkerIndex_(1) {
+      sharedWorkerIndex_(1),
+      progressThreadEnabled_(init_params.enableProgTh) {
```
rakhmets, line 1081: Does this also work if `num_threads` is set to a non-zero value using custom parameters?

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: removed all_workers

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: removed progress

src/plugins/ucx/ucx_backend.cpp
rakhmets: If so, then:
1. `nixlUcxEngine::createGpuXferReq` should return `NIXL_ERR_NOT_SUPPORTED`.
2. `nixlUcxThreadEngine::createGpuXferReq`, and `nixlUcxThreadPoolEngine::createGpuXferReq` should call the same method (e.g. protected `nixlUcxEngine` method).
3. There is no need for the extra class field.

src/utils/ucx/gpu_xfer_req_h.h
```diff
@@ -19,11 +19,13 @@
 #define NIXL_SRC_UTILS_UCX_GPU_XFER_REQ_H_H
 
 #include <vector>
+#include <memory>
 
 #include "nixl_types.h"
 
 class nixlUcxEp;
 class nixlUcxMem;
+class nixlUcxWorker;
```
rakhmets: Please revert.

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: added `NIXL_UCX_GPU_XFER_TIMEOUT_MS`

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: removed

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,31 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    const auto start = std::chrono::steady_clock::now();
```
rakhmets: `start` should be set right before the loop.

Please consider the following to avoid the possibility of adding extra code between initialization and usage of this variable:
```C++
for (const auto start = std::chrono::steady_clock::now();
     std::chrono::steady_clock::now() - start <= timeout;)
{
    status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle);
    if (status != UCS_ERR_NOT_CONNECTED) {
        break;
    }
}
```

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,31 @@ createGpuXferReq(const nixlUcxEp &ep,
...
+    const char *timeout_env = getenv("NIXL_UCX_GPU_XFER_TIMEOUT_MS");
+    if (timeout_env) {
+        timeout_ms = atoi(timeout_env);
+        if (timeout_ms <= 0) {
+            NIXL_WARN << "Invalid NIXL_UCX_GPU_XFER_TIMEOUT_MS value: " << timeout_env
+                      << ", using default " << default_timeout_ms << " ms";
+            timeout_ms = default_timeout_ms;
+        }
+    }
+    const auto timeout = std::chrono::milliseconds(timeout_ms);
```
rakhmets: the code can be put into a separate function.

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,8 +77,31 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
+    const auto start = std::chrono::steady_clock::now();
+    constexpr int default_timeout_ms = 5000;
+    int timeout_ms = default_timeout_ms;
+    const char *timeout_env = getenv("NIXL_UCX_GPU_XFER_TIMEOUT_MS");
```
rakhmets: Please use a string_view for NIXL_UCX_GPU_XFER_TIMEOUT_MS.

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -21,6 +21,9 @@
 #include "rkey.h"
 #include "config.h"
 
+#include <chrono>
+#include <thread>
```
rakhmets, line 27: Can be removed.

src/utils/ucx/gpu_xfer_req_h.h
```diff
@@ -19,6 +19,7 @@
 #define NIXL_SRC_UTILS_UCX_GPU_XFER_REQ_H_H
 
 #include <vector>
+#include <memory>
```
rakhmets: Please remove.

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: why? both are used

rakhmets: Sorry, missed usage of `<thread>`.

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: moved start

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: done

src/utils/ucx/gpu_xfer_req_h.cpp
michal-shalev: done

src/utils/ucx/gpu_xfer_req_h.h
michal-shalev: removed

src/plugins/ucx/ucx_backend.cpp
michal-shalev: implemented only for `nixlUcxThreadEngine`

src/plugins/ucx/ucx_backend.h
```diff
@@ -308,6 +312,9 @@ class nixlUcxEngine : public nixlBackendEngine {
     // Context to use when current context is missing
     nixlUcxCudaDevicePrimaryCtxPtr m_cudaPrimaryCtx;
 
+    /* Progress thread */
+    bool progressThreadEnabled_;
```
rakhmets: Please remove.

src/plugins/ucx/ucx_backend.h
```diff
@@ -69,6 +69,16 @@ class nixlUcxPrivateMetadata : public nixlBackendMD {
...
+        [[nodiscard]] nixlUcxMem &
+        getMem() noexcept {
+            return mem;
+        }
+
+        [[nodiscard]] const nixlUcxMem &
+        getMem() const noexcept {
+            return mem;
+        }
+
```
rakhmets: it looks like something added by mistake

michal-shalev: removed

src/plugins/ucx/ucx_backend.h
michal-shalev: removed

rakhmets (approved):

src/utils/ucx/gpu_xfer_req_h.cpp
yosefe: Maybe better to keep progress to not fully rely on progress thread here?
@rakhmets

yosefe: also do we still need to move createGpuXferReq to nixlUcxThreadEngine if not doing progress?

src/utils/ucx/gpu_xfer_req_h.cpp
```diff
@@ -74,15 +104,31 @@ createGpuXferReq(const nixlUcxEp &ep,
...
+    while ((ucs_status = ucp_device_mem_list_create(ep.getEp(), &params, &ucx_handle)) ==
+           UCS_ERR_NOT_CONNECTED) {
+        if (!timeout_warned && std::chrono::steady_clock::now() - start > timeout) {
+            NIXL_WARN << "Timeout on creating device memory list has been exceeded (timeout="
+                      << timeout.count() << " ms)";
+            timeout_warned = true;
+        }
+
+        worker.progress();
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
yosefe: 1. IMO better progress after the sleep to increase probably we get a message
2. maybe we should do the busy loop in nixlUcxEngine::createGpuXferReq since we have all the workers there already and we can progress
3. i'd suggest to sleep only if progress returned 0

michal-shalev: 1. moved
2. I thought we said that we'll progress just the related worker, why all workers? also notice that there's a try catch
3. done

src/plugins/ucx/ucx_backend.h
```diff
@@ -308,6 +308,9 @@ class nixlUcxEngine : public nixlBackendEngine {
     // Context to use when current context is missing
     nixlUcxCudaDevicePrimaryCtxPtr m_cudaPrimaryCtx;
 
+    /* Progress thread */
+    bool progressThreadEnabled_;
```
rakhmets:
```suggestion
    const bool progressThreadEnabled_;
```
The field is constant.
The comment is redundant.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1630,6 +1631,10 @@ nixlUcxEngine::createGpuXferReq(const nixlBackendReqH &req_hndl,
         return NIXL_ERR_INVALID_PARAM;
     }
 
+    if (!progressThreadEnabled_) {
+        NIXL_WARN << "Progress thread must be enabled for GPU transfer requests";
```
rakhmets, line 1657: `must` -> `have to`?

brminich: AI says must is better)

src/plugins/ucx/ucx_backend.h
brminich: agree about the comment

src/plugins/ucx/ucx_backend.cpp
rakhmets: Since this is a warning, this is not a strict requirement. The method can work even when there is no progress thread. That is why I recommended `have to`.

yosefe (approved):

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Is this still relevant or we can resolve? @rakhmets @brminich

rakhmets: Yes, it is, at least for me.
It doesn't look quite right to me to issue a warning, write in it that `Progress thread must be enabled`, continue working.

ofirfarjun7: Hi.
I don't think that `have to be enabled` is less strict.
So, I suggest to move forward with it.

If we really want to make it less strict we should neither use `must` or `have to`.

PR #951, opened 2025-10-25
BUILD: Add meson option for selective plugin building

michal-shalev:
## What?
Add new `enable_plugins` and `disable_plugins` meson options to selectively build specific plugins or exclude plugins from the default set.

## Why?
Building all plugins increases build time and requires all plugin dependencies to be present. When developing or deploying with only specific plugins (e.g., just UCX), building unnecessary plugins wastes time and resources. These options allow users to:
- Reduce build time significantly
- Avoid unnecessary dependencies
- Speed up development iteration when working on a single plugin
- Easily exclude problematic plugins while keeping the rest enabled

## How?
- Added `enable_plugins` meson option that accepts a comma-separated list of plugin names to build (e.g., `UCX,GPUNETIO`)
- Added `disable_plugins` meson option that accepts a comma-separated list of plugin names to exclude (e.g., `GDS,MOONCAKE`)
- Options are mutually exclusive - error if both are set
- Default behavior (neither option set) builds all available plugins (backward compatible)
- Created a global `enabled_plugins` dictionary that is computed once in the main `meson.build` and used throughout
- Applied plugin filtering to:
  - `src/plugins/meson.build` - only builds enabled plugin libraries
  - `test/unit/plugins/meson.build` - only builds unit tests for enabled plugins
  - `test/gtest/plugins/meson.build` - skips if required plugins aren't enabled

Usage examples:
```bash
# Build only GDS plugin
meson setup build -Denable_plugins=GDS

# Build GDS and HF3FS plugins
meson setup build -Denable_plugins=GDS,HF3FS

# Build all plugins EXCEPT GDS
meson setup build -Ddisable_plugins=GDS

# Build all plugins except GDS and MOONCAKE
meson setup build -Ddisable_plugins=GDS,MOONCAKE

# Build all plugins (default)
meson setup build
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "bindgen"
 version = "0.71.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -177,6 +186,7 @@ checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 name = "nixl-sys"
 version = "0.8.0"
 dependencies = [
+ "bincode",
  "bindgen",
  "cc",
  "libc",
--- a/benchmark/kvbench/pyproject.toml
+++ b/benchmark/kvbench/pyproject.toml
@@ -18,7 +18,7 @@ name = "NIXLKVBench"
 version = "0.2.0"
 description = "Benchmarking utility for testing KVCache transfers in large language models (LLMs) using the NIXL runtime."
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "click==8.1.7",
     "pytest==7.4.4",
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -189,6 +189,8 @@ RUN uv pip install --upgrade meson meson-python pybind11 patchelf pyYAML click t
 RUN CUDA_SHORT_VERSION=cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d .) && \
     FLAGS="--index-url https://download.pytorch.org/whl/$CUDA_SHORT_VERSION" && \
     uv pip install $FLAGS torch torchvision torchaudio
+# Upgrade setuptools to latest version for compatibility with PEP 639 (license format)
+RUN uv pip install --upgrade 'setuptools>=80.9.0'
 
 COPY --from=nixl . /workspace/nixl
 COPY --from=nixlbench . /workspace/nixlbench
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -191,6 +191,8 @@ RUN uv pip install --upgrade meson meson-python pybind11 patchelf pyYAML click t
 # Install PyTorch
 RUN export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d .)" && \
     uv pip install torch torchvision torchaudio
+# Upgrade setuptools to latest version for compatibility with PEP 639 (license format)
+RUN uv pip install --upgrade 'setuptools>=80.9.0'
 
 WORKDIR /workspace/nixl
 COPY . /workspace/nixl
--- a/contrib/Dockerfile.manylinux
+++ b/contrib/Dockerfile.manylinux
@@ -18,7 +18,7 @@ ARG BASE_IMAGE
 ARG BASE_IMAGE_TAG
 FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG}
 
-ARG DEFAULT_PYTHON_VERSION="3.12"
+ARG DEFAULT_PYTHON_VERSION="3.14"
 ARG ARCH="x86_64"
 ARG UCX_REF="v1.20.x"
 ARG LIBFABRIC_VERSION="v1.21.0"
@@ -244,6 +244,8 @@ RUN uv pip install --upgrade meson meson-python pybind11 patchelf pyYAML click s
 # Install PyTorch
 RUN export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d .)" && \
     uv pip install torch torchvision torchaudio
+# Upgrade setuptools to latest version for compatibility with PEP 639 (license format)
+RUN uv pip install --upgrade 'setuptools>=80.9.0'
 
 COPY . /workspace/nixl
 WORKDIR /workspace/nixl
@@ -267,10 +269,11 @@ RUN echo "/usr/local/nixl/lib/$ARCH-linux-gnu" > /etc/ld.so.conf.d/nixl.conf && 
 
 # Create the wheel
 # No need to specifically add path to libcuda.so here, meson finds the stubs and links them
-ARG WHL_PYTHON_VERSIONS="3.9,3.10,3.11,3.12,3.13"
+ARG WHL_PYTHON_VERSIONS="3.10,3.11,3.12,3.13,3.14"
 ARG WHL_PLATFORM="manylinux_2_28_$ARCH"
 RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
     export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | cut -d. -f1,2 | tr -d .)" && \
+    export UV_INDEX_STRATEGY=unsafe-best-match && \
     mkdir -p dist && \
     for PYTHON_VERSION in "${PYTHON_VERSIONS[@]}"; do \
         export PATH=$VIRTUAL_ENV/bin:$PATH && \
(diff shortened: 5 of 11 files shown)


iyastreb:
Seems to be useful
Maybe it can be just a regex implementing both allow and deny logic, similar to `GTEST_FILTER`?
E.g.
-Dbuild_plugins=GDS // allow, only GDS is used
-Dbuild_plugins=^GDS //deny, all except GDS are used

meson_options.txt
```diff
@@ -25,6 +25,7 @@ option('cudapath_inc', type: 'string', value: '', description: 'Include path for
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
 option('static_plugins', type: 'string', value: '', description: 'Plugins to be built-in, comma-separated')
+option('build_plugins', type: 'string', value: '', description: 'List of plugins to build, comma-separated. Default (empty) builds all available plugins')
```
ovidiusm: Could we name this plugins?

meson.build
```diff
@@ -21,6 +21,20 @@ project('nixl', 'CPP', version: '0.6.1',
     meson_version: '>= 0.64.0'
 )
 
+build_plugins_opt = get_option('build_plugins')
+enabled_plugins = build_plugins_opt == '' ? [] : build_plugins_opt.split(',')
+
+should_build_plugin = {}
```
ovidiusm: `should_build_plugin` should be called `enabled_plugins` IMO

meson.build
```diff
@@ -21,6 +21,20 @@ project('nixl', 'CPP', version: '0.6.1',
...
+foreach plugin : ['UCX', 'UCX_MO', 'LIBFABRIC', 'POSIX', 'OBJ', 'GDS', 'GDS_MT', 'MOONCAKE', 'Mooncake', 'HF3FS', 'GUSLI', 'GPUNETIO']
```
ovidiusm: Let's make a list `all_plugins` and do case insensitive comparison to avoid duplicates

aranadive:
@michal-shalev thanks for adding this. I think we can take a little bit better approach:
We should auto-generate a list of plugins during build by parsing the src/plugins directory for all the plugins. Then, create a map from plugin to plugin_directory. Then just iterate over this list in src/plugins/meson.build, ignoring ones in the user-provided mask. Just make sure all the dependencies for a plugin should be moved to the plugins meson.build and not handled in the src/plugins/meson.build.

meson.build
brminich: note `UCX_MO` is removed

src/plugins/meson.build
```diff
@@ -13,48 +13,55 @@
...
 
-ucx_backend_inc_dirs = include_directories('./ucx')
-
-if ucx_dep.found()
+if ucx_dep.found() and should_build_plugin.get('UCX')
     subdir('ucx')
+endif
+
+if ucx_dep.found() and should_build_plugin.get('UCX_MO')
+    ucx_backend_inc_dirs = include_directories('./ucx')
```
brminich: if some pluging is requested, but can't be built, we should print an error or/and fail the build.
wdyt?

meson_options.txt
michal-shalev: done

meson.build

meson.build

michal-shalev: removed

src/plugins/meson.build
michal-shalev: If a user explicitly requests a plugin, the build explicitly fails with an error message, just like it would fail regardless of this PR, are you suggesting to add an additional error?

michal-shalev:
/ok to test [71dbc9e](https://github.com/ai-dynamo/nixl/pull/951/commits/71dbc9e0d65e240f5b7d0074a42cc0cfaabe0dbb)

meson.build
```diff
@@ -21,6 +21,69 @@ project('nixl', 'CPP', version: '0.8.0',
...
+if enabled_plugins_list.length() > 0
+    foreach p : enabled_plugins_list
+        if p.startswith('^')
+            is_exclusion_mode = true
+            break
+        endif
+    endforeach
+endif
+
+if is_exclusion_mode
```
brminich: i'd not support mixed list of plugins. Imo it should be either enable or disable list. I.e. if it starts with `^` then all plugins should be disabled

meson.build
```diff
@@ -21,6 +21,69 @@ project('nixl', 'CPP', version: '0.8.0',
...
+endif
+
+all_plugins = ['UCX', 'LIBFABRIC', 'POSIX', 'OBJ', 'GDS', 'GDS_MT', 'MOONCAKE', 'HF3FS', 'GUSLI', 'GPUNETIO']
+
+enabled_plugins = {}
+
+is_exclusion_mode = false
+if enabled_plugins_list.length() > 0
+    foreach p : enabled_plugins_list
+        if p.startswith('^')
```
ovidiusm: Could we enforce that the first one must start with ^?

meson.build
```diff
@@ -21,6 +21,69 @@ project('nixl', 'CPP', version: '0.8.0',
...
+            plugin_name = p.substring(1)
+            if plugin_name in all_plugins
+                enabled_plugins += {plugin_name: false}
+            else
+                warning('Unknown plugin excluded: ' + plugin_name)
+            endif
+        else
+            if p not in all_plugins
+                 warning('Unknown plugin requested: ' + p)
```
ovidiusm: This should be an error

meson.build
```diff
@@ -21,6 +21,69 @@ project('nixl', 'CPP', version: '0.8.0',
...
+        enabled_plugins += {plugin: true}
+    endforeach
+
+    foreach p : enabled_plugins_list
+        if p.startswith('^')
+            plugin_name = p.substring(1)
+            if plugin_name in all_plugins
+                enabled_plugins += {plugin_name: false}
+            else
+                warning('Unknown plugin excluded: ' + plugin_name)
```
ovidiusm: This should be an error

README.md
```diff
@@ -144,6 +144,7 @@ Common build options:
 - `disable_gds_backend`: Disable GDS backend (default: false)
 - `cudapath_inc`, `cudapath_lib`: Custom CUDA paths
 - `static_plugins`: Comma-separated list of plugins to build statically
+- `plugins`: Comma-separated list of plugins to build. Default builds all available. Prefix with `^` to exclude.
```
brminich: can we better introduce 2 options? Like `enable_plugins` and `disable_plugins`
If enable_plugins is non-empty, use it as the whitelist (ignore disable_plugins)
If only disable_plugins is set, build all except those listed
Or: error if both are set (safest, forces user to be explicit)

ovidiusm: I like this option

ovidiusm (dismissed):

meson.build
```diff
@@ -21,67 +21,53 @@ project('nixl', 'CPP', version: '0.8.0',
...
-if enabled_plugins_list.length() > 0
-    foreach p : enabled_plugins_list
-        if p.startswith('^')
-            is_exclusion_mode = true
-            break
+if enable_plugins_opt != ''
+    enabled_plugins_list = []
+    foreach p : enable_plugins_opt.split(',')
+        p_stripped = p.strip()
+        if p_stripped not in all_plugins
```
brminich, line 40: pls include `UCCL` to all_plugins, it has been recently merged

meson.build
```diff
@@ -21,6 +21,55 @@ project('nixl', 'CPP', version: '0.8.0',
...
+enabled_plugins = {}
+
+if enable_plugins_opt != ''
+    enabled_plugins_list = []
+    foreach p : enable_plugins_opt.split(',')
+        p_stripped = p.strip()
+        if p_stripped not in all_plugins
+            error('Requested plugin "' + p_stripped + '" is not available. Available plugins are: ' + ', '.join(all_plugins))
+        endif
+        enabled_plugins_list += p_stripped
```
brminich: should it be `[p_stripped]`?

src/plugins/meson.build
```diff
@@ -13,45 +13,50 @@
...
 # limitations under the License.
 
-if ucx_dep.found()
+if ucx_dep.found() and enabled_plugins.get('UCX')
     subdir('ucx')
 endif
 
-subdir('posix')  # Always try to build POSIX backend, it will handle its own dependencies
-subdir('obj')  # Always try to build Obj backend, it will handle its own dependencies
+if enabled_plugins.get('POSIX')
```
brminich, line 24: can we fail a build if some pluging is requested but can't be built?

meson.build
michal-shalev: added

meson.build
michal-shalev: AFAIK `array += element` works in meson but I changed it to be consistent with the rest of the file

michal-shalev:
/ok to test [9fb3d11](https://github.com/ai-dynamo/nixl/pull/951/commits/9fb3d11d2821f7343b2ca532a6614e27217e7766)

src/plugins/meson.build
michal-shalev: added an error

src/plugins/meson.build
```diff
@@ -13,50 +13,109 @@
...
-    subdir('libfabric')
+if enabled_plugins.get('LIBFABRIC')
+    if not libfabric_dep.found() and is_explicit_enable
+        error('LIBFABRIC plugin requested but libfabric dependency not found')
+    elif libfabric_dep.found()
+        subdir('libfabric')
+    endif
 endif
 
 disable_gds_backend = get_option('disable_gds_backend')
```
aranadive, line 40: Can we remove this option and the disable mooncake option since we can now use the mask.

brminich: is it safe to remove?
maybe just mention it is deprecated?

ovidiusm: We should remove separately, and consider deprecation first

src/plugins/meson.build
ovidiusm: Marking this as resolved as it was addressed in the last version

PR #953, opened 2025-10-27
nixlbench: Allow compilation when etcd is not available

benlwalker:
## What?
The code already handled running without etcd, but now allow it to compile without etcd.

## Why?
My VM doesn't have etcd and I'm running storage workloads anyway, so I don't need etcd.

## How?
The code already runs without etcd, I just had to fix the compilation to not include the etcd stuff if etcd isn't available.


aranadive:
CI Gitlab error is due to network. Passed previously. Merging.

PR #954, opened 2025-10-27
Bindings: Add Python wrapper in meta-package

ovidiusm:
## What?
Ensures that nixl-cu12 and nixl-cu13 do not both overwrite nixl dir, and adds __init__.py wrapper to nixl metapackage.

Installing nixl metapackage installs nixl-cu12 by default. If nixl-cu13 is installed, it takes precedence and is used instead.

## Why?
* Support both CUDA 12 and CUDA 13.
* Removes file conflicts between nixl-cu12 and nixl-cu13 packages.
* Downstream projects like dynamo or vllm can continue to depend on nixl, so there is no need to do `pip install dynamo[nixl-cu12]` or `pip install dynamo[nixl-cu13]`.
* `import nixl` works transparently:

```
Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import nixl
>>> dir(nixl)
['DEFAULT_COMM_PORT', 'LOGGING_CONFIG', 'Optional', 'Union', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '_api', '_pkg', '_utils', 'attr', 'ba_buf', 'candidates', 'free_passthru', 'get_logger', 'importlib', 'logger', 'logging', 'malloc_passthru', 'module', 'nixlBind', 'nixl_agent', 'nixl_agent_config', 'nixl_backend_handle', 'nixl_prepped_dlist_handle', 'nixl_xfer_handle', 'np', 'os', 'pickle', 'pkg', 'set_log_level_by_env', 'setup_logging', 'sub_name', 'submodules', 'sys', 'torch', 'verify_transfer']
>>>
```

Legacy imports work as well without changes:

```
>>> import nixl._utils as nixl_utils
>>> from nixl._api import nixl_agent, nixl_agent_config
>>> from nixl.logging import get_logger
>>> nixl_agent_config(backends=["UCX"])
<nixl_cu12._api.nixl_agent_config object at 0x7f00946aaae0>
>>>
```

PR #955, opened 2025-10-27
Bindings: Make nixl depend on nixl-cu12

ovidiusm:
## What?
Make `nixl-cu12` a required dependency of `nixl`.

## Why?
To improve user experience, `pip install nixl` will automatically pull `nixl-cu12` dependency, so requirements.txt / pyproject.toml do not have to change for existing users of nixl / CUDA 12.

To use nixl with CUDA 13, users still have to run `pip install nixl[cu13]` or `pip install nixl-cu13`. The CUDA 13 package has higher priority when imported when both are installed.


ovidiusm:
Will add this to https://github.com/ai-dynamo/nixl/pull/954

PR #957, opened 2025-10-28
CI consolidation: Use multi-stage Dockerfile

Alexey-Rivkin:
## What?
Consolidates NIXL build procedures by using contrib/Dockerfile multi-stage build for both CI and customer workflows.

## Why?
- Alignment: CI now uses the same Docker image build process as customers
- Consolidation: Single source of truth for NIXL builds across all environments

#### Multi-stage Dockerfile
1. Stage 1 (nixl-base): Dependencies (libfabric, etcd, aws-sdk, DOCA, rust)
2. Stage 2 (nixl-build): UCX and NIXL compilation (default target)

#### CI Changes
- Jenkins uses ci-demo auto-build with Harbor registry caching
- build.sh simplified to NIXL-only compilation

#### Backward Compatibility
- build-container.sh unchanged, defaults to full build

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -56,7 +56,7 @@ ARCH=$(uname -m)
 # Skip dependency installation if running in pre-built nixl-base image
 if [ -n "${NIXL_BASE_IMAGE_ENV}" ]; then
     # Use pre-installed libfabric from base image
-    LIBFABRIC_INSTALL_DIR=${LIBFABRIC_INSTALL_DIR:-/usr/local}
+    LIBFABRIC_INSTALL_DIR=/usr/local
 else
 
 # Some docker images are with broken installations:


.gitlab/build.sh
```diff
@@ -53,6 +53,13 @@ fi
 ARCH=$(uname -m)
 [ "$ARCH" = "arm64" ] && ARCH="aarch64"
 
+# Skip dependency installation if running in pre-built base image
+if [ "${NIXL_DEPS_PREINSTALLED}" = "true" ]; then
```
dpressle: name is not clear i would call it something like:
NIXL_BASE_IMAGE_ENV

And just check if it exists no need to have special value for it
if [[ -z $NIXL_BASE_IMAGE_ENV ]]

.gitlab/build.sh
```diff
@@ -53,6 +53,13 @@ fi
...
+    # Set paths for pre-installed dependencies
+    UCX_INSTALL_DIR=${UCX_INSTALL_DIR:-/usr}
+    LIBFABRIC_INSTALL_DIR=${LIBFABRIC_INSTALL_DIR:-/usr/local}
```
dpressle: Why you need it here? we already have it in lines 26 and 34

contrib/Dockerfile
```diff
@@ -17,19 +17,17 @@ ARG BASE_IMAGE="nvcr.io/nvidia/cuda-dl-base"
 ARG BASE_IMAGE_TAG="25.06-cuda12.9-devel-ubuntu24.04"
 ARG OS
 
-FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG}
+FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} AS nixl-deps-base
```
dpressle: can we keep it simple and call it nixl_base i dont see why we need the deps in the name it does not describe the true nature/purpose of the target

contrib/Dockerfile
```diff
@@ -58,17 +56,41 @@ RUN apt-get update -y && \
...
     clang \
     hwloc \
     libhwloc-dev \
-    libcurl4-openssl-dev libssl-dev uuid-dev zlib1g-dev # aws-sdk-cpp dependencies
+    libcurl4-openssl-dev libssl-dev uuid-dev zlib1g-dev \
+    python3-dev \
+    python3-pip \
+    curl \
+    wget \
+    libnuma-dev \
```
dpressle: but its already installed in line 81 (in original file) please check for all duplications

contrib/Dockerfile
dpressle, line 201: why this is in base image? i dont think it relevant for our CI

.gitlab/build.sh
Alexey-Rivkin: Because in our base image we already have UCX and libfabric installed into different locations

contrib/Dockerfile
Alexey-Rivkin: Renamed to 'nixl-base'

contrib/Dockerfile
Alexey-Rivkin: 'libnuma-dev' was the only dup

contrib/Dockerfile
Alexey-Rivkin: Moved, thanks!

.gitlab/build.sh
Alexey-Rivkin: Done

contrib/Dockerfile
```diff
@@ -190,14 +215,20 @@ RUN uv pip install --upgrade meson meson-python pybind11 patchelf pyYAML click t
...
     uv pip install torch torchvision torchaudio
 
+# Re-declare ARGs needed in this stage
+ARG NIXL_PREFIX="/usr/local/nixl"
+ARG NIXL_PLUGIN_DIR="$NIXL_PREFIX/lib/x86_64-linux-gnu/plugins"
+ARG ARCH="x86_64"
+ARG DEFAULT_PYTHON_VERSION="3.12"
+ARG UCX_PREFIX="/usr"
+ARG UCX_PLUGIN_DIR="$UCX_PREFIX/lib/ucx"
+ARG LIBFABRIC_INSTALL_PATH="/usr/local"
```
dpressle: Usually we put ARGS under FROM not inside the file

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -34,9 +38,21 @@ kubernetes:
   requests: "{memory: 8Gi, cpu: 8000m}"
 
 runs_on_dockers:
-  - { name: "ubuntu24.04-cuda-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04" }
-  - { name: "ubuntu22.04-cuda-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:24.10-cuda12.6-devel-ubuntu22.04" }
-  - { name: "podman-v5.0.2", url: "quay.io/podman/stable:v5.0.2", category: 'tool', privileged: true }
+  - {
+      file: "contrib/Dockerfile",
+      name: "ubuntu24.04-nixl-deps",
```
dpressle: lets call the image nixl-base

.gitlab/build.sh
```diff
@@ -53,6 +53,12 @@ fi
 ARCH=$(uname -m)
 [ "$ARCH" = "arm64" ] && ARCH="aarch64"
 
+# Skip dependency installation if running in pre-built base image
+if [ -n "${NIXL_BASE_IMAGE_ENV}" ]; then
```
dpressle, line 57: this is strange because it conflicts with the script logic of setting these variables in lines at the top.
one is an actual script command line argument which a user will not expect to get replaced by internal script logic.

dpressle: ?

.gitlab/build.sh
```diff
@@ -215,6 +189,40 @@ rm "libfabric-${LIBFABRIC_VERSION#v}.tar.bz2"
...
+          --with-dm \
+          ${UCX_CUDA_BUILD_ARGS} \
+          --enable-mt && \
+        make -j && \
+        make -j install-strip && \
+        $SUDO ldconfig \
+)
+
+# Ubuntu 22.04 specific setup
+if grep -q "Ubuntu 22.04" /etc/os-release 2>/dev/null; then
```
dpressle, line 217: Is this for gitlab tests? because in our CI we will use base image and its part of it already right?

contrib/Dockerfile
```diff
@@ -115,35 +144,17 @@ RUN wget --tries=3 --waitretry=5 \
...
-         --with-verbs                \
-         --with-dm                   \
-         --with-gdrcopy=/usr/local   \
-         --with-efa                  \
-         --enable-mt &&              \
-     make -j${NPROC:-$(nproc)} &&                \
-     make -j${NPROC:-$(nproc)} install-strip &&  \
-     ldconfig
+# Clean up old UCX installations to avoid conflicts
+RUN rm -rf /usr/lib/ucx /opt/hpcx/ucx /usr/local/ucx && \
```
dpressle, line 148: can you explain this? usually we clean before we re-install maybe we should have default ucx and just overrun it in CI

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -34,8 +38,20 @@ kubernetes:
   requests: "{memory: 8Gi, cpu: 8000m}"
 
 runs_on_dockers:
-  - { name: "ubuntu24.04-cuda-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04" }
-  - { name: "ubuntu22.04-cuda-dl-base", url: "nvcr.io/nvidia/cuda-dl-base:24.10-cuda12.6-devel-ubuntu22.04" }
+  - {
+      file: "contrib/Dockerfile",
+      name: "ubuntu24.04-nixl-base",
+      uri: "$arch/$name",
```
dpressle, line 44: lets think about the naming convention of these images why os is part of the name but arch not?
$arch/$os/$name:<tag>
x86_64/ubuntu24.04/nixl-base:xxx
or
$arch/$name-$os:<tag>
x86_64/nixl-base-ubuntu24.04:xxx

Alexey-Rivkin: `$arch/$os/$name:` looks better, thanks!

.gitlab/build.sh
Alexey-Rivkin: `LIBFABRIC_INSTALL_DIR` var has two different purposes in this script:
1. Where to build libfabric (in backward compatibility mode)
2. Where to find libfabric (in base image mode)

When nixl-base image is used, we have to overwrite the path so it points to pre-installed dependencies. Without the override meson install will fail.

.gitlab/build.sh
Alexey-Rivkin: We decided not to bundle Python dependencies into the base image, so those upgrades are needed for every use case.

contrib/Dockerfile
Alexey-Rivkin: Main already cleaned UCX:
```
-RUN rm -rf /usr/lib/ucx
-RUN rm -rf /opt/hpcx/ucx
```
Adding one more path that was conflicting. 
Since UCX install moved out of the base image, libfabric installs first and drops files under /usr/local/ucx.

brminich:
@Alexey-Rivkin is it still relevant?

brminich:
@Alexey-Rivkin  can we close it?

PR #961, opened 2025-10-29
Libfabric: update required minimum version to v1.21.0

yexiang-aws:
## What?
1. Update required minimum libfabric version to v1.18.3
2. Use v1.18.3 during build
3. Clean up logs for better debugging


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1207,10 +1207,10 @@ nixlLibfabricRail::postRead(void *local_buffer,
...
     }
 
-    NIXL_TRACE << "Posting RDMA read on endpoint: " << std::hex << endpoint
-               << " local_buffer: " << local_buffer << " length: " << length
-               << " dest_addr: " << dest_addr << " remote_addr: " << (void *)remote_addr
-               << " remote_key: " << remote_key << " context: " << &req->ctx;
+    NIXL_TRACE << "Posting RDMA read on endpoint=" << std::hex << endpoint
+               << " local_buffer=" << local_buffer << " length=" << length
+               << " dest_addr=" << dest_addr << " remote_addr=" << (void *)remote_addr
+               << " remote_key=" << remote_key << " context=" << &req->ctx;
```
akkart-aws, line 1213: Why is it required to replace all : to = ?

yexiangd: just to unify the syntax for easier log search.  We used both : and = randomly everywhere

PR #965, opened 2025-10-29
nixlbench: fix cudaFree problem

e-ago:
GPUNetIO plugin has a persistent kernel running on a dedicated stream.
NIXL Bench uses the sync memory op cudaFree to free VRAM buffers but the cudaFree requires to synchronize the whole GPU device before proceeding.

Due to the gpunetIo persistent kernel, nixlbench executed with gpunetio plugin hangs on the cudaFree.

This PR resolves this issue replacing cudaFree with cudaFreeAsync

### DIFF
--- a/src/api/python/__init__.py
+++ b/src/api/python/__init__.py
@@ -12,17 +12,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-import os
-
-if "NIXL_PLUGIN_DIR" not in os.environ:
-    # name for local installation
-    package_root = os.path.dirname(__file__)
-    plugin_dir = os.path.join(package_root, ".nixl.mesonpy.libs/plugins/")
-
-    # fallback for PyPI installation
-    if not os.path.isdir(plugin_dir):
-        plugin_dir = os.path.join(package_root, ".nixl_pybind.mesonpy.libs/plugins/")
-
-    if os.path.isdir(plugin_dir):
-        os.environ["NIXL_PLUGIN_DIR"] = plugin_dir
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -836,11 +836,11 @@ nixlLibfabricEngine::loadMetadataHelper(const std::vector<uint64_t> &rail_keys,
     pub_md->remote_buf_addr_ = reinterpret_cast<uint64_t>(buffer);
     pub_md->conn_ = conn;
 
-    output = pub_md.release();
     NIXL_DEBUG << "Metadata loaded with"
                << " Remote addr: " << (void *)pub_md->remote_buf_addr_ << " Remote keys for "
                << pub_md->rail_remote_key_list_.size() << " rails"
                << " Remote fi_addr: " << pub_md->conn_->rail_remote_addr_list_[0][0];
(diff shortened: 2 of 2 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -633,7 +633,17 @@ xferBenchNixlWorker::cleanupBasicDescVram(xferBenchIOV &iov) {
...
+        /*
+         * CUDA streams allow for concurrent execution of kernels and memory operations. However,
+         * memory management functions like cudaFree are implicitly synchronized with all streams to
+         * guarantee safety. This means cudaFree will wait for all kernels (in any stream) that
+         * might use the memory to finish before actually freeing it.
+         * If the application hangs on cudaFree due to kernels running in other streams, switching
+         * to cudaFreeAsync can allow the host to proceed without waiting for the entire device
+         * synchronization.
+         */
+        CHECK_CUDA_ERROR(cudaFreeAsync((void *)iov.addr, 0), "Failed to deallocate CUDA buffer");
```
aranadive, line 645: Will this cause problems for other CUDA-based plugins that are not using streams? Or should we restrict this for GPUNetIO plugin?

e-ago: this may cause an issue on CUDA kernel based plugins. Atm we have GPUNetIO but also UCX using GPUNetIO (thus CUDA kernels used for communications) may incur in this issue.
To be clear, this is not a GPUNetIO issue, it's a CUDA driver issue.

rakhmets, line 645: Does this mean that an application using NIXL cannot use synchronous memory allocation and deallocation (e.g. cudaMalloc, cudaFree)?
It looks like a significant limitation on the use of the framework.

Wouldn't it better to resolve the issue on backend side? E.g. create and destroy stream and launch kernels in frames of `nixlDocaEngine::connect`, and `nixlDocaEngine::disconnect`.

rakhmets: Changing the approach to connect/disconnect way requires significant development, and has issues with scalability at the moment.

Please update the documentation adding restriction on use synchronous CUDA memory management API in conjunction with GPUNetIO plugin.

aranadive: Seems like the cudaFreeAsync will just act like cudaFree for synchronous memory deallocation here, no?

rakhmets: No, it won't. `nixlDocaEngine::wait_stream` is created with `cudaStreamNonBlocking`. This stream doesn't synchronize with the default stream. Thus, `cudaFreeAsync(..., 0)` doesn't sync with the kernel launched on `nixlDocaEngine::wait_stream`.

e-ago: Correct. @rakhmets  @aranadive  can we merge? I will follow up with the discussed changes in another PR

rakhmets (approved):

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
rakhmets: Do we need to change `cudaMalloc` to `cudaMallocAsync`?

PR #977, opened 2025-10-31
[build] make `libfabric` truly optional

wseaton:
On certain systems that already have libfabric installed, it may be undesirable to build nixl with explicit support (eg. libfabric version on the host system is too old). This allows a build-time opt out (basically bypass the current behavior of attempting to link with auto-detection), and should be backwards compatible with existing buids/CI.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -188,7 +188,9 @@ else
 endif
 
 libfabric_path = get_option('libfabric_path')
-if libfabric_path != ''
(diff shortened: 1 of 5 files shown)


aranadive:
Thanks for creating the PR. We will most likely with a generic way to enable/disable plugins like #951.

brminich:
@wseaton can we close it given that we have options `enable_plugins` and `disable_plugins` now?

PR #982, opened 2025-11-03
NIXL/UCX: Optimize request handling

iyastreb:
## What?
sglang workflow (and other LLMs workflows as well) is using huge batches (around 64k elements of very small size <1KB). This leads to large amount of UCP request being stored and tracked by NIXL. The problem is that NIXL checking so many requests delays calling progress on workers, which leads to higher latency and lower TTFT.

This PR aims to significantly reduce the request handling overhead by:
- Remove _init and _fini callbacks for UCP requests
- Reduce request memory footprint (removed `amBuffer`, `completed` fields)
- Store connection shared_ptr in handle, not in request
- When checking batch status, check status of LAST request. If it's incomplete - return IN_PROG immediately without checking others
- Improve AM message handling, so that buffer is deallocated in callback, and detach it from request

## Perf boost:
With this change I see performance improvement around **50% in nixlbench** (512 msg size, 65000 batch size, 8 progress threads), and **20% TTFT reduction in sglang**.

## Next steps
Potential improvements in the next PR:
- Aggressive optimization: drop most of the batch requests immediately after creating. This gives a huge perf boost. But we need to make sure it's consistent and error handling works well. Need to take into account multiple EPs use case
- Can we postpone request put step from `status` to `release`, so that workers are not blocked by this bookkeeping stage


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -358,8 +328,11 @@ class nixlUcxBackendH : public nixlBackendReqH {
     }
 
     void
-    append(nixlUcxIntReq *req) {
-        requests_.push_back(req);
+    append(nixlUcxReq req, ucx_connection_ptr_t conn) {
+        nixlUcxIntReq *req_int = reinterpret_cast<nixlUcxIntReq *>(req);
```
rakhmets: Please use `auto`, since the type is explicitly mentioned on the right side.

src/plugins/ucx/ucx_backend.h
```diff
@@ -288,7 +288,7 @@ class nixlUcxEngine : public nixlBackendEngine {
     nixl_status_t
     notifSendPriv(const std::string &remote_agent,
                   const std::string &msg,
-                  nixlUcxReq &req,
+                  nixlUcxReq *req,
```
rakhmets: I think it makes sense to make it `nixlUcxReq *req = nullptr`, so you can use the method without adding `nullptr`.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -236,11 +236,25 @@ nixl_status_t nixlUcxEp::disconnect_nb()
...
-nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
-                                void* hdr, size_t hdr_len,
-                                void* buffer, size_t len,
-                                uint32_t flags, nixlUcxReq &req)
-{
+using nixl_ucx_am_cb_ctx_t = std::pair<void *, nixlUcxEp::am_deleter_t>;
+using nixl_ucx_am_cb_ctx_ptr_t = std::unique_ptr<nixl_ucx_am_cb_ctx_t>;
+
+void
+nixlUcxEp::sendAmCallback(void *request, ucs_status_t status, void *user_data) {
```
rakhmets, line 243: This method can be put into unnamed namespace in .cpp file to avoid adding it to the class interface.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -236,11 +236,25 @@ nixl_status_t nixlUcxEp::disconnect_nb()
...
+    nixl_ucx_am_cb_ctx_t *ctx = (nixl_ucx_am_cb_ctx_t *)user_data;
```
rakhmets: const_cast

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1768,12 +1736,29 @@ nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
...
+                     (void *)buffer->data(),
+                     buffer->size(),
+                     UCP_AM_SEND_FLAG_EAGER,
+                     tmp_req,
+                     deleter);
+    if (ret != NIXL_SUCCESS && ret != NIXL_IN_PROG) {
+        /* Failed to send active message, delete buffer */
+        delete buffer;
+    } else if (req != nullptr) {
+        *req = tmp_req;
```
rakhmets: What is the reason for adding temporary variable? Why can't we use `req`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1768,12 +1736,29 @@ nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
...
     // TODO: replace with mpool for performance
 
-    auto buffer = std::make_unique<std::string>(ser_des.exportStr());
-    ret = ep->sendAm(
-        NOTIF_STR, NULL, 0, (void *)buffer->data(), buffer->size(), UCP_AM_SEND_FLAG_EAGER, req);
-    if (ret == NIXL_IN_PROG) {
-        nixlUcxIntReq* nReq = (nixlUcxIntReq*)req;
-        nReq->amBuffer = std::move(buffer);
+    std::string *buffer = new std::string(ser_des.exportStr());
+    auto deleter = [buffer, req](void *request, void *ptr) {
```
rakhmets: `req` and `request` look similar. Maybe `request` and `completed_request`?

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: done

src/plugins/ucx/ucx_backend.cpp
iyastreb: req can be nullptr now, if caller is not interested to track it

src/plugins/ucx/ucx_backend.h
iyastreb: ok

src/utils/ucx/ucx_utils.cpp
iyastreb: I thought about it, but still prefer to keep it in the class to reduce the visibility scope and keep related things together

src/utils/ucx/ucx_utils.cpp
iyastreb: static_cast

src/plugins/ucx/ucx_backend.cpp
rakhmets: Wouldn't it be better to pass `req` to `sendAm`, and add there:
```C++
if (req != nullptr) {
    req = static_cast<nixlUcxReq>(request);
}
```
?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1759,21 +1727,38 @@ int nixlUcxEngine::progress() {
...
+    nixlUcxReq tmp_req;
+    ret = ep->sendAm(NOTIF_STR,
+                     NULL,
+                     0,
+                     (void *)buffer->data(),
+                     buffer->size(),
+                     UCP_AM_SEND_FLAG_EAGER,
+                     tmp_req,
+                     deleter);
+    if (ret != NIXL_SUCCESS && ret != NIXL_IN_PROG) {
```
rakhmets: It seems we can delete buffer if `ret` is `NIXL_SUCCESS`.
Maybe use `unique_ptr`, and release it if `ret == NIXL_IN_PROG`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1759,21 +1727,38 @@ int nixlUcxEngine::progress() {
...
+        delete buffer;
+        if (nullptr == req) {
+            /* Caller is not interested in the request, free it */
+            ucp_request_free(completed_request);
+        }
+    };
+
+    nixlUcxReq tmp_req;
+    ret = ep->sendAm(NOTIF_STR,
+                     NULL,
```
rakhmets: nullptr

src/utils/ucx/ucx_utils.cpp
rakhmets: ah... Now I see your answer. It's `ucx_utils` file containing dozens of class declarations and definitions...

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: I disabled immediate completion with UCP_OP_ATTR_FLAG_NO_IMM_CMPL, so normally callback is always called. But I like your idea, let me check if it works

src/plugins/ucx/ucx_backend.cpp
iyastreb: No, this is not an option, because we need to return request from the function immediately, so that caller can track it. If we do that in callback, then req remains unassigned until request completion

src/plugins/ucx/ucx_backend.cpp
iyastreb: I removed UCP_OP_ATTR_FLAG_NO_IMM_CMPL flag
Now I call deleter from sendAm if it completes immediately, and caller code is cleaner

src/plugins/ucx/ucx_backend.cpp
rakhmets: I think I wasn't clear in my last comment.
```C++
    nixl_status_t
    notifSendPriv(const std::string &remote_agent,
                  const std::string &msg,
                  const std::unique_ptr<nixlUcxEp> &ep,
                  nixlUcxReq *req = nullptr) const;
...
nixl_status_t
nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
                             const std::string &msg,
                             const std::unique_ptr<nixlUcxEp> &ep,
                             nixlUcxReq *req) const {
...
    ret = ep->sendAm(NOTIF_STR,
                     nullptr,
                     0,
                     static_cast<void *>(buffer->data()),
                     buffer->size(),
                     UCP_AM_SEND_FLAG_EAGER,
                     req,
                     deleter);
...
    nixl_status_t
    sendAm(unsigned msg_id,
           void *hdr,
           size_t hdr_len,
           void *buffer,
           size_t len,
           uint32_t flags,
           nixlUcxReq *req = nullptr,
           am_deleter_t deleter = nullptr);
...
nixl_status_t
nixlUcxEp::sendAm(unsigned msg_id,
                  void *hdr,
                  size_t hdr_len,
                  void *buffer,
                  size_t len,
                  uint32_t flags,
                  nixlUcxReq *req,
                  am_deleter_t deleter) {
...
    if (UCS_PTR_IS_PTR(request)) {
        ctx.release();
        if (req != nullptr) {
            *req = static_cast<nixlUcxReq>(request);
        }
        return NIXL_IN_PROG;
    }
    if (deleter) {
        deleter(nullptr, buffer);
    }
...
```

iyastreb: I see, makes sense, done

src/utils/ucx/ucx_utils.cpp
```diff
@@ -251,10 +265,25 @@ nixl_status_t nixlUcxEp::sendAm(unsigned msg_id,
...
     ucs_status_ptr_t request = ucp_am_send_nbx(eph, msg_id, hdr, hdr_len, buffer, len, &param);
     if (UCS_PTR_IS_PTR(request)) {
-        req = (void*)request;
+        ctx.release();
+        if (req != nullptr) {
+            *req = static_cast<nixlUcxReq>(request);
+        }
         return NIXL_IN_PROG;
+    } else {
+        if (deleter) {
```
brminich: maybe
```suggestion
    } else if (deleter) {
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -295,51 +288,28 @@ class nixlUcxIntReq {
...
-}
-
 /****************************************
  * Backend request management
 *****************************************/
 
 class nixlUcxBackendH : public nixlBackendReqH {
 private:
     std::vector<nixlUcxIntReq *> requests_;
+    std::set<ucx_connection_ptr_t> connections_;
```
rakhmets, line 300: Shouldn't it be declared before `requests_`, since `connections_` contains shared pointers to objects, and `requests_`
contains raw pointers to the same objects?

iyastreb: I can put it before, but it does not matter at all, because nixlUcxIntReq is not constructed/destructed as C++ object, its lifetime is managed by UCP memory pool

src/utils/ucx/ucx_utils.cpp
iyastreb: ok

rakhmets (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -394,37 +368,37 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-                _internalRequestReset(req);
-                worker->reqRelease((nixlUcxReq)req);
-            } else {
+            nixl_status_t ret = ucx_status_to_nixl(ucp_request_check_status(req));
+            if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
+                worker->reqRelease(req);
+            } else if (ret == NIXL_IN_PROG) {
+                if (out_ret == NIXL_SUCCESS) {
+                    out_ret = NIXL_IN_PROG;
+                }
```
brminich, line 393: i'd return once the first IN_PROG request found
can be addressed in the following optimization PR

iyastreb: I don't think we need to optimize this, because this is not a regular workflow:
- Since LAST request completed successfully, all the previous request must be completed as well
- If some of them is IN_PROG - this means our check is not consistent, so I thought of adding a warning maybe
- But more importantly we want to check all the request to detect whether some of them are in ERROR state. If we return early we might never reach failing request

PR #988, opened 2025-11-04
chore: Update the UCX ver to 1.20.0

Alexey-Rivkin:
## What?
Update the UCX ver to 1.20.0

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -659,6 +659,7 @@ void xferBenchUtils::checkConsistency(std::vector<std::vector<xferBenchIOV>> &io
                                           xferBenchConfig::num_initiator_dev);
         gusli_devmap_init = true;
     }
+    bool pass_check_consistency = true;
     for (const auto &iov_list: iov_lists) {
         for(const auto &iov: iov_list) {
             void *addr = NULL;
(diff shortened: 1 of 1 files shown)


Alexey-Rivkin:
Depends on the UCX v1.20 release

.ci/jenkins/pipeline/proj-jjb.yaml
```diff
@@ -272,8 +272,8 @@
           description: "NIXL version to use (tag like 0.7.1, branch name, or commit hash)"
       - string:
           name: "UCX_VERSION"
-          default: "v1.19.0"
-          description: "UCX version to use (tag like v1.19.0, branch name, or commit hash)"
+          default: "v1.20.0"
```
brminich: there is no such tag yet - it should be a branch v1.20.x

brminich:
@dpressle can we merge this?
@ovidiusm for vis

dpressle:
> @dpressle can we merge this? @ovidiusm for vis

Did we released ucx 1.20 in GH? do we have tarball for that version? i think we have some processes that downloads these packages and not cloning

ovidiusm:
No, UCX 1.20 has not been released yet, 1.20.x is a release branch. That's why I am wondering if it might be better to point to the latest hash on the branch instead of the branch itself

Alexey-Rivkin:
> No, UCX 1.20 has not been released yet, 1.20.x is a release branch. That's why I am wondering if it might be better to point to the latest hash on the branch instead of the branch itself

@brminich WDYT?

brminich:
> > No, UCX 1.20 has not been released yet, 1.20.x is a release branch. That's why I am wondering if it might be better to point to the latest hash on the branch instead of the branch itself
> 
> @brminich WDYT?

IMO, pointing to the branch is ok, because it is release branch and we expect that only critical fixes will be merged there.
1.20 is not released yet, which is fine imo. AFAIR, we did the same with 1.19

ovidiusm:
Then I am also OK with using 1.20.x

PR #992, opened 2025-11-04
make nixlbench exit with failure when iov consistency check fails.

fengjica:
## What?
Today nixlbench does not exit when consistency check fails, but only prints in stderr. This PR tries to change it to exit with EXIT_FAILURE.

## Why?
Without exiting with an error code, one needs to go to the log file to check this result. It's easy to ignore that in dev and CI.

## How?
 -


fengjica:
Hi! Can someone help look at this PR?

iyastreb (approved):

PR #998, opened 2025-11-05
Add NIXL_PREFIX to the lib search path

milesial:
This allows us to export `NIXL_PREFIX` to the path where the python wheel was installed, and build the rust lib with the linker finding those `.so` in `...python3.12/site-packages/.nixl.mesonpy.libs`


aranadive (approved):

milesial:
@aranadive can this one be merged?

PR #1000, opened 2025-11-05
Update readme with latest instructions for python / CUDA 13

ovidiusm:
## What?
Add detailed instructions for installing the python wheel, and building from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,23 @@ NVIDIA Inference Xfer Library (NIXL) is targeted for accelerating point to point
 ## Pre-build Distributions
 ### PyPI Wheel
 
-The nixl python API and libraries, including UCX, are available directly through PyPI:
+The nixl python API and libraries, including UCX, are available directly through PyPI.
+
+It can be installed for CUDA 12 with:
 
 ```
-pip install nixl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,12 +13,24 @@ NVIDIA Inference Xfer Library (NIXL) is targeted for accelerating point to point
...
+For CUDA 13 with:
 
 ```
-pip install nixl
+pip install nixl[cu13]
 ```
 
+For backwards compatibility, `pip install nixl` installs automatically `nixl[cu12]`, continuing to work seamlessly for CUDA 12 users without requiring changes to downstream project dependencies.
+
+If both `nixl-cu12` and `nixl-cu13` are installed at the same time in an environment, `nixl-cu13` takes precedence.
```
aranadive, line 32: Can you add that GPUNetIO plugin only works with cu12?

ovidiusm: I don't think this is the right place to mention it

ovidiusm:
Could we merge this?

README.md
aranadive: Ok

aranadive (approved):

PR #1005, opened 2025-11-07
PLUGINS/UCX: Moved src/utils/ucx/* to src/plugins/ucx.

rakhmets:
## What?
Moved `src/utils/ucx/*` to `src/plugins/ucx`.

## Why?
All code related to the `UCX` plugin can be moved to `src/plugins/ucx`, as the `UCX_MO` plugin has been removed: https://github.com/ai-dynamo/nixl/pull/898.

### DIFF
--- a/test/unit/plugins/ucx/ucx_am_test.cpp
+++ b/test/unit/plugins/ucx/ucx_am_test.cpp
@@ -28,28 +28,29 @@ struct sample_header {
     uint32_t test;
 };
 
-ucs_status_t check_buffer (void *arg, const void *header,
-   		                   size_t header_length, void *data,
-				           size_t length,
-				           const ucp_am_recv_param_t *param)
-{
-    struct sample_header* hdr = (struct sample_header*) header;
-    //TODO: is data 8 byte aligned?
-    uint64_t recv_data = *((uint64_t*) data);
+ucs_status_t
+check_buffer(void *arg,
+             const void *header,
+             size_t header_length,
+             void *data,
+             size_t length,
+             const ucp_am_recv_param_t *param) {
+    struct sample_header *hdr = (struct sample_header *)header;
+    // TODO: is data 8 byte aligned?
+    uint64_t recv_data = *((uint64_t *)data);
 
-    if (hdr->test != 0xcee)
-	    return UCS_ERR_INVALID_PARAM;
+    if (hdr->test != 0xcee) return UCS_ERR_INVALID_PARAM;
 
-    assert (length == 8);
-    assert (recv_data == 0xdeaddeaddeadbeef);
+    assert(length == 8);
+    assert(recv_data == 0xdeaddeaddeadbeef);
 
     std::cout << "check_buffer passed\n";
 
     return UCS_OK;
 }
 
-int main()
-{
+int
+main() {
     vector<string> devs;
     devs.push_back("mlx5_0");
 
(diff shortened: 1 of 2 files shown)


src/plugins/ucx/config.cpp
```diff
@@ -24,21 +24,21 @@
 
 namespace nixl::ucx {
 void
-config::modify (std::string_view key, std::string_view value) const {
-    const char *env_val = std::getenv (absl::StrFormat ("UCX_%s", key.data()).c_str());
+config::modify(std::string_view key, std::string_view value) const {
```
rakhmets, line 27: Unfortunately, it is impossible to avoid formatting changes for the moved files.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -192,19 +188,16 @@ nixlUcxEp::closeImpl(ucp_ep_close_flags_t flags)
...
-                                UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
-    ep_params.err_mode        = err_handling_mode;
-    ep_params.err_handler.cb  = err_cb_wrapper;
-    ep_params.err_handler.arg = reinterpret_cast<void*>(this);
-    ep_params.address         = reinterpret_cast<ucp_address_t*>(addr);
+    ep_params.field_mask = UCP_EP_PARAM_FIELD_REMOTE_ADDRESS | UCP_EP_PARAM_FIELD_ERR_HANDLER |
+        UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
+    ep_params.err_mode = err_handling_mode;
+    ep_params.err_handler.cb = err_cb_wrapper;
+    ep_params.err_handler.arg = reinterpret_cast<void *>(this);
```
ColinNV: `static_cast` should be enough here.

rakhmets: Fixed.

rakhmets:
> Move and reformat? Ok.

Right. I cannot just move the files, as clang formatter treats this as modification. Thus, reformatting is also applied.

src/plugins/ucx/meson.build
```diff
@@ -13,21 +13,41 @@
...
+ucx_backend_dependencies = [asio_dep,
+                            cuda_dep,
+                            nixl_common_dep,
+                            nixl_infra,
+                            serdes_interface,
+                            thread_dep,
+                            ucx_dep]
+
+ucx_backend_include_directories = [nixl_inc_dirs,
+                                   utils_inc_dirs]
```
brminich, line 37: is utils needed? what does it point to?

test/unit/utils/ucx/meson.build
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
brminich: we have to also move /test/utils/ucx content to /test/plugin/ucx

src/plugins/ucx/meson.build
rakhmets: Yes, it is. It is used on lines 43 and 50,

rakhmets: https://github.com/ai-dynamo/nixl/blob/d82d09e0fb3c8bb2a574401c37ffa6b8cd71297c/meson.build#L338

test/unit/utils/ucx/meson.build
rakhmets: Done.

src/plugins/ucx/meson.build
brminich: what do we need from utils for ucx backend?

rakhmets: We need `src/utils/common/nixl_log.h`.

PR #1010, opened 2025-11-10
ETCD: initial heartbeat thread

tstamler:
## What?
Add a heartbeat mechanism to NIXL ETCD metadata to invalidate stale metadata if an agent dies. 

When an agent uploads metadata to ETCD, it will only lease that data and not store permanently. Then, a heartbeat thread will be spawned to continue renewing the lease on the metadata as long as the agent exists. If the lease is not renewed, the metadata will be removed, which will automatically tell any NIXL agents watching to remove that agent. 

## Why?

If a NIXL agent has uploaded metadata to ETCD and is unexpectedly killed (or user exits application without invalidating), the metadata it had uploaded will remain in ETCD indefinitely and can confuse applications that rely on ETCD for agent coordination.


src/core/nixl_listener.cpp
```diff
@@ -232,6 +240,23 @@ class nixlEtcdClient {
...
 
+    ~nixlEtcdClient() {
+        heartbeat_thread_stop = true;
+        if (heartbeat_thread.joinable()) {
+            heartbeat_thread.join();
+        }
+    }
+
+    void
+    startHeartbeatThread(uint64_t lease_id) {
```
aranadive, line 263: I think this needs to be a queue of lease ids. If an agent can store multiple metadata within etcd we will have multiple lease ids. We should add all those to a queue (like the commqueue) and the heartbeat thread should just loop over all of them.

src/core/nixl_listener.cpp
```diff
@@ -588,7 +633,7 @@ nixlAgentData::commWorkerInternal(nixlAgent *myAgent) {
                     }
                     NIXL_DEBUG << "Successfully loaded metadata for agent: " << remote_agent;
 
-                    etcdClient->setupAgentWatcher(remote_agent);
+                    etcdClient->setupAgentWatcher(remote_agent, metadata_label);
```
aranadive: If a remote agent sends multiple metadata, this watch will only be installed for the first MD since we only allow 1 watch per agent. Would we have a case where we need to only invalidate a certain MD in this flow?

src/core/nixl_listener.cpp
aranadive: Another way would be add some special known key to deem an agent active and create a lease_id only for that. The agent keeps this key alive, and the other agents watcher only checks this key. If this key is deleted that means the agent is gone, so delete all MD for it.

aranadive (dismissed):
LGTM.

2dm:
@tstamler I added the fixes we talked about.

src/core/nixl_listener.cpp
```diff
@@ -224,14 +232,43 @@ class nixlEtcdClient {
 
         NIXL_DEBUG << "Using etcd namespace for agents: " << namespace_prefix;
 
+        etcd::Response response = etcd->leasegrant((heartbeat.count()) * 2);
+        lease_id = response.value().lease();
```
aranadive: is this valid if response is incorrect?

tstamler: probably wouldn't break anything but there's no need to get the lease_id here anyway, I can just check response.is_ok() first, so I've done that

aranadive (approved):

src/core/nixl_listener.cpp
```diff
@@ -323,7 +360,7 @@ class nixlEtcdClient {
 
             // Get current index to watch from
             etcd::Response response = etcd->get(metadata_key);
-            int64_t watch_index = response.index();
+            int64_t watch_index = response.index() + 1;
```
ovidiusm, line 364: Could you please explain this change?

src/core/nixl_listener.cpp
```diff
@@ -224,14 +232,43 @@ class nixlEtcdClient {
...
+        heartbeat_thread_stop = true;
+        if (heartbeat_thread.joinable()) {
+            heartbeat_thread.join();
+        }
+    }
+
+    void
+    startHeartbeatThread(uint64_t lease_id) {
+        while (!heartbeat_thread_stop) {
+            // keep alive for twice the heartbeat interval
```
ovidiusm, line 265: Could you please explain why 2x?

src/core/nixl_listener.cpp
```diff
@@ -224,14 +232,43 @@ class nixlEtcdClient {
 
         NIXL_DEBUG << "Using etcd namespace for agents: " << namespace_prefix;
 
+        etcd::Response response = etcd->leasegrant((heartbeat.count()) * 2);
+
+        if (response.is_ok()) {
+
```
ovidiusm, line 238: Empty line?

src/core/nixl_listener.cpp
```diff
@@ -224,14 +232,43 @@ class nixlEtcdClient {
 
         NIXL_DEBUG << "Using etcd namespace for agents: " << namespace_prefix;
 
+        etcd::Response response = etcd->leasegrant((heartbeat.count()) * 2);
```
ovidiusm, line 235: Why 2x?

examples/python/heartbeat_example.py
```diff
@@ -0,0 +1,97 @@
...
+logger = get_logger(__name__)
+
+
+def run_target():
+    """
+    Target mode function that runs in a subprocess.
+    This posts metadata to etcd and then is killed.
+    """
+    logger.info("Target subprocess started")
+
```
ovidiusm, line 33: Could we please reduce the empty lines in this file?

src/api/cpp/nixl_params.h
```diff
@@ -61,6 +61,12 @@ class nixlAgentConfig {
          */
         std::chrono::microseconds etcdWatchTimeout;
 
+        /**
+         * @var Heartbeat interval in seconds
+         *      Interval for heartbeat that keeps remote metadata valid.
+         */
+        std::chrono::seconds heartbeatInterval;
```
ovidiusm, line 68: I think, to be able to review this PR, there needs to be a detailed description that explains the why & how as per template.

src/core/nixl_listener.cpp
2dm: Setting a watcher for the current index can "replay" events from that index. 
For example, an agent is removed and re-added, as can happen in the an elastic use case. 
The created watcher gets the delete event and closes.  
Setting +1 for the index it is created on, where it failed to fetch, will notify it of changes that happens from this point on.

brminich (commented):
can we reuse `commWorker` thread instead of creating a new one?
Or, if it  is not possible, use etcd builtin `etcd::KeepAlive keepalive` which is sipposed to do that internally

brminich:
@tstamler, can you pls apply/review comments if still relevant?

tstamler:
> @tstamler, can you pls apply/review comments if still relevant?

Sorry for the delay! I wasn't aware that etcd::keepAlive was implemented like that. I had tried to keep it separate from the commWorker in case the commWorker had a lot of work that would delay a heartbeat too long and expire the lease. If it really does all this for us, I will try a patch with etcd::keepAlive. I will want to test with deepEP fault tolerance first. Otherwise, this should also make redundant most of the comments.

PR #1012, opened 2025-11-11
UTILS/UCX: Set UCX_IB_PCI_RELAXED_OREDERING=try.

rakhmets:
## What?
Set `UCX_IB_PCI_RELAXED_OREDERING=try`.


src/utils/ucx/ucx_utils.cpp
```diff
@@ -464,6 +464,8 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
         config.modify ("MAX_RMA_RAILS", "2");
     }
 
+    config.modify("IB_PCI_RELAXED_ORDERING", "try");
```
brminich: minor - can we add to line 455? To have all 3 `config.modify` one after another?

PR #1020, opened 2025-11-13
plugins/posix: AIO resubmission fixed

anton-nayshtut:
The Linux AIO plugin wasn’t capable of re-submission; it required IO to be re-prepared prior to re-submission.

It also used to return `NIXL_IN_PROG` for the next `submit()`, and `NIXL_SUCCESS` for `checkCompleted()`, making the upper layer think that the re-submission succeeded. This, in turn, caused incorrect throughput-related calculations.

This patch fixes the bug.


src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -58,6 +57,11 @@ linuxAioQueue::submit(const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
         return NIXL_IN_PROG;
     }
 
+    if (num_ios_to_complete) {
+        NIXL_ERROR << "previous submit is not completed";
```
aranadive:
```suggestion
        NIXL_ERROR << "previously submitted IO is not yet complete";
```

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -91,20 +94,15 @@ linuxAioQueue::checkCompleted() {
     }
 
     for (int i = 0; i < rc; i++) {
-        struct iocb *io = events[i].obj;
-        size_t idx = (size_t)io->data;
-
-        ios_to_submit[idx] = nullptr; // Mark as completed
```
aranadive, line 97: Do you still need to set this then in prepIO?

benlwalker: Yes, he still needs that. This is gathering all of the prepIO calls for a single request into this array and then submitting them in post. This class is called a "queue" but it's not a queue at all - it's just a batching mechanism for a single request.

aranadive: Sorry I meant setting the io->data in prepIO. The idx value seems unused now.

benlwalker: Oh, yes he doesn't need to set the io->data anymore.

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/linux_aio_queue.cpp
anton-nayshtut: Removed, thanks!

barneuman (approved):
LGTM

Issue #1022, opened 2025-11-12
[Documentation] nixlbench ETCD barrier synchronization timing guidance needed


dmvevents:
## Problem
nixlbench fails with ETCD barrier synchronization errors when nodes don't start within close proximity:

```
Error in barrier wait... rank 0 completed 1/2 ranks)
Failed to synchronize at start barrier
```

## Environment
- Platform: AWS SageMaker HyperPod EKS  
- Setup: 2-node H100 configuration with ETCD coordination
- Network: EFA-enabled

## Details
When using `kubectl exec` to start nixlbench on multiple nodes, if there's >3 seconds delay between node startups (or AWS credentials expire between calls), ETCD barrier synchronization fails.

## Suggestion
Add documentation guidance:
1. Nodes should start within 3 seconds of each other
2. Recommend using Kubernetes Jobs for simultaneous startup
3. Document ETCD barrier timeout settings

## Workaround
Start both nodes within 3 seconds:
```bash
kubectl exec node1 -- nixlbench ... &
sleep 1
kubectl exec node2 -- nixlbench ...
```

Or use K8s Jobs for proper simultaneous startup.

Successfully implemented with K8s StatefulSet + coordination script.

Reference: https://github.com/dmvevents/dynamo-workshop/blob/main/NIXLBENCH_TESTING.md

rmccorm4:
Hi @dmvevents, thanks for raising this. Moved this issue to the nixl repository. CC @mkhazraee @kahalon

aranadive:
Thanks. By default this timeout is 60s. Added PR #1026.

PR #1026, opened 2025-11-14
nixlbench: Update README.md

aranadive:
Add info on etcd barrier timeout and some cleanup. Addresses #1022.

### DIFF
--- a/benchmark/nixlbench/README.md
+++ b/benchmark/nixlbench/README.md
@@ -557,12 +557,11 @@ The workers automatically coordinate ranks through ETCD as they connect. Note, t
 ```bash
 # Basic UCX benchmark
 ./nixlbench --etcd_endpoints http://etcd-server:2379 --backend UCX
-sleep 2
-./nixlbench --etcd_endpoints http://etcd-server:2379 --backend UCX
+sleep 2 && ./nixlbench --etcd_endpoints http://etcd-server:2379 --backend UCX
 
 # UCX with specific devices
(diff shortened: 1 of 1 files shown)


benchmark/nixlbench/README.md
```diff
@@ -562,9 +557,12 @@ The workers automatically coordinate ranks through ETCD as they connect.
 ```bash
 # Basic UCX benchmark
 ./nixlbench --etcd_endpoints http://etcd-server:2379 --backend UCX
+sleep 2
```
brminich: why do we need it here?
also it is not added below when running with specific devices

benchmark/nixlbench/README.md
```diff
@@ -814,6 +798,12 @@ export UCX_LOG_LEVEL=DEBUG # Verbose UCX logging
 export UCX_PROTO_INFO=y # See transport used by UCX
 ```
 
+#### Etcd Cleanup
```
brminich:
```suggestion
#### ETCD Cleanup
```

for consistency (it is `ETCD` thru the file)

itay:
Any chance we can get this closed out?

benchmark/nixlbench/README.md
aranadive: Just to avoid any failure if there are delays in reaching etcd server from first nixlbench instance. Added for the specific device case.

benchmark/nixlbench/README.md
brminich: why is it resolved without a fix?
@aranadive

PR #1027, opened 2025-11-14
log: replace std::cout with NIXL_INFO in nixlDescList::print

matumu:
## What?
replace std::cout with NIXL_INFO in nixlDescList::print

### DIFF
--- a/src/infra/nixl_descriptors.cpp
+++ b/src/infra/nixl_descriptors.cpp
@@ -313,7 +313,7 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
 template <class T>
 void nixlDescList<T>::print() const {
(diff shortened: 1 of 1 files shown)


src/infra/nixl_descriptors.cpp
```diff
@@ -313,7 +313,7 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
 template <class T>
 void nixlDescList<T>::print() const {
-    std::cout << "DescList of mem type " << type << std::endl;
+    NIXL_INFO << "DescList of mem type " << type;
```
brminich, line 316: std::count here is on purpose - because we expect that print() method would always print without the need to set `NIXL_LOG_LEVEL=info`

matumu: Thanks for explaining the design decision

PR #1030, opened 2025-11-14
fix: use release build for ucx

nv-anants:
## What?
Switch to configure-release for ucx build, as recommended in docs - https://github.com/openucx/ucx?tab=readme-ov-file#release-builds
 
## Why?
Optimize libucx packaged in nixl by using release build

## How?
It adds some extra flags like disable-debug/assertions.... in the configure command

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -266,39 +266,13 @@ void nixlUcxEngine::vramFiniCtx()
 }
 
 /****************************************
- * UCX request management
-*****************************************/
-
-
-class nixlUcxIntReq {
-public:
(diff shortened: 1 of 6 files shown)


.gitlab/build.sh
```diff
@@ -144,7 +144,7 @@ curl -fSsL "https://github.com/openucx/ucx/tarball/${UCX_VERSION}" | tar xz
 ( \
   cd openucx-ucx* && \
   ./autogen.sh && \
-  ./configure \
+  ./contrib/configure-release \
```
rakhmets: You can use `./contrib/configure-release-mt` and remove `--enable-mt`.

nv-anants: updated

brminich (approved):

PR #1046, opened 2025-11-21
TEST/UCX: Remove obsolete UCX test.

ColinNV:
## What?
Remove `allocateWrongGPUTest` from `ucx_backend_test`.

## Why?
This test checks that registering memory on a different GPU fails in UCX; it needs to be removed because recent UCX versions can now handle this case. (It also feels more like a UCX test than a NIXL one.)

### DIFF
--- a/docs/BackendGuide.md
+++ b/docs/BackendGuide.md
@@ -108,8 +108,8 @@ Note that getNotif does not know which agent it should look for to receive the n
 
 A key underlying abstraction for NIXL library is a descriptor list, that is made of a memory space (host/GPU/block/File/Obj-Store) and a list of descriptors. There are 2 types of descriptors used for the SB API.
 
-*For transfers: (addr, len, devID, metadata), where metadata is a pointer to an nixlBackendMD object relevant to the registered memory that this descriptor falls within.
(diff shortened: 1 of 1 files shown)


test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -274,24 +274,6 @@ void *releaseValidationPtr(nixl_mem_t mem_type, void *addr)
...
-    allocateBuffer(VRAM_SEG, dev_id, desc.len, buf);
-
-    desc.devId = dev_id;
-    desc.addr = (uint64_t) buf;
-
-    int ret = ucx->registerMem(desc, VRAM_SEG, md);
-
-    nixl_exit_on_failure((ret == NIXL_ERR_NOT_SUPPORTED), "Failed to register memory", "test");
-
-    releaseBuffer(VRAM_SEG, dev_id, buf);
```
brminich, line 292: @rakhmets should we reconsider this whole test with removal of CUDA dep in #946 ?

rakhmets: The changes in the test doesn't directly relate to the #946.
The test can be re-written regardless of changes by aforementioned PR.

yosefe (approved):

brminich (approved):

PR #1050, opened 2025-11-24
BUILD: Re-enable device API tests

michal-shalev:
## What?
Re-enable device API tests (`PrepGpuSignal` and `ucxDeviceApi`) in the CI and add GPU availability checks to prevent test failures on non-GPU systems. Additionally, add UCX version checking to ensure the Device API is only enabled with UCX >= 1.21.

## Why?
These tests were previously disabled due to a problem in UCX that has been fixed (https://github.com/openucx/ucx/pull/11011). Additionally, these tests would fail on systems without GPUs, so proper runtime checks are needed to skip gracefully when no GPU is available. The UCX Device API requires version 1.21 or later, so version checking is needed to avoid using the API with older UCX versions.

## How?
- Removed the `--gtest_filter=-*PrepGpuSignal*:*ucxDeviceApi*` exclusion from the `gtest-parallel` command in `.gitlab/test_cpp.sh`
- Removed commented-out UCX GDA configuration that is no longer needed
- Added `hasCudaGpu()` utility function to both `single_write_test.cu` and `test_transfer.cpp` to check GPU availability at runtime
- Added configure-time UCX version check (>= 1.21) in `meson.build` using `UCS_STATIC_ASSERT` to verify version at compile-time
- Added runtime version check in `ucx_utils.cpp` to conditionally enable `UCP_FEATURE_DEVICE` only with UCX >= 1.21
- Added DOCA GPUNETIO dependency check to Device API enabling condition

### DIFF
--- a/src/api/python/__init__.py
+++ b/src/api/python/__init__.py
@@ -12,3 +12,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from ._api import (
+    DEFAULT_COMM_PORT,
+    nixl_agent,
+    nixl_agent_config,
+    nixl_backend_handle,
+    nixl_prepped_dlist_handle,
+    nixl_xfer_handle,
+)
+
+__all__ = [
+    # Constants
+    "DEFAULT_COMM_PORT",
+    # Main classes
+    "nixl_agent",
+    "nixl_agent_config",
+    "nixl_backend_handle",
+    "nixl_prepped_dlist_handle",
+    "nixl_xfer_handle",
+]


test/gtest/test_transfer.cpp
```diff
@@ -38,6 +38,20 @@
...
+inline bool
+hasCudaGpu() {
+#ifdef HAVE_CUDA
+    int count = 0;
+    auto err = cudaGetDeviceCount(&count);
+    return (err == cudaSuccess && count > 0);
+#else
+    return false;
+#endif
+}
```
yosefe: seems like duplication of the code added in test/gtest/device_api/single_write_test.cu

meson.build
```diff
@@ -160,11 +160,11 @@ if ucx_path != ''
   # Check if path is absolute
   if ucx_inc_path.startswith('/')
     ucx_dep = declare_dependency(
-      link_args : ['-L' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
+      link_args : ['-L' + ucx_lib_path, '-Wl,-rpath,' + ucx_lib_path, '-lucp', '-lucs', '-luct'],
```
yosefe: maybe better to add separate meson option for ucx_rpath. anyway, why was it needed in this pr?

meson.build
```diff
@@ -193,6 +193,17 @@ endif
 nvcc_prog = find_program('nvcc', required: false)
 ucx_gpu_device_api_available = false
 if ucx_dep.found() and cuda_dep.found() and nvcc_prog.found()
+    ucx_version_result = cpp.run('''
+        #include <ucp/api/ucp.h>
+        int main() {
+            unsigned major, minor, release;
+            ucp_get_version(&major, &minor, &release);
+            unsigned version = UCP_VERSION(major, minor);
+            return (version >= UCP_VERSION(1, 21)) ? 0 : 1;
```
yosefe: 1. maybe check the version only in CI scripts (and set gtest_filter accordingly)? 
2. if checking here, we can add static assert that version >=2.21 to the check in lines 214-217 (have_host_side = cpp.compiles .. ) since static assert can be checked in compile time

meson.build
michal-shalev: This PR adds a runtime version check that executes a test program calling `ucp_get_version()`. Without rpath, the test program loads system UCX (1.16.0) instead of the custom path UCX (1.21.0), causing the version check to fail incorrectly.
Regarding "ucx_rpath" suggestion, IMO we should keep it automatic. When users specify `ucx_path`, they want to use that UCX, rpath ensures this happens. Adding an option adds complexity and potential for misconfiguration.

meson.build
michal-shalev: 1. IMO we shouldn't build code and tests that aren't supported.
2. The runtime check makes it immediately clear why the device API is disabled (version too old), the preprocessor check would just show a compile failure without indicating it's a version issue. This helps users quickly understand whether they need to upgrade UCX or if there's a different problem.

yosefe: 2. using compiler check can also be quite indicative with UCS_STATIC_ASSERT(...) and will not need to change anything with rpath (which changes existing behavior)

michal-shalev: Used `UCS_STATIC_ASSERT` and kept `ucx_version_ok` for clarity

test/gtest/test_transfer.cpp
michal-shalev: I'm planning to remove this test from `test/gtest/test_transfer.cpp` in the next PR.
I can also move this func to `test/gtest/common.h` if you prefer.

meson.build
michal-shalev: removed this change

meson.build
```diff
@@ -193,6 +193,15 @@ endif
 nvcc_prog = find_program('nvcc', required: false)
 ucx_gpu_device_api_available = false
 if ucx_dep.found() and cuda_dep.found() and nvcc_prog.found()
+    ucx_version_ok = cpp.compiles('''
+            #include <ucp/api/ucp_version.h>
+            #include <ucs/sys/compiler_def.h>
+            int main() {
+                UCS_STATIC_ASSERT(UCP_API_MAJOR > 1 || (UCP_API_MAJOR == 1 && UCP_API_MINOR >= 21));
```
yosefe: 1. unite it with line 212-215
2. `UCS_STATIC_ASSERT(UCP_VERSION(UCP_API_MAJOR, UCP_API_MINOR) >= UCP_VERSION(1, 2));`

test/gtest/test_transfer.cpp
yosefe: yes IMO better move to common

brminich (commented):
👍 besides code duplication comment from @yosefe

test/gtest/test_transfer.cpp
michal-shalev: moved

meson.build

meson.build
```diff
@@ -201,11 +201,16 @@ if ucx_dep.found() and cuda_dep.found() and nvcc_prog.found()
...
-
-    if have_gpu_side and have_host_side
+            #include <ucs/sys/compiler_def.h>
+            int main() {
+                UCS_STATIC_ASSERT(UCP_VERSION(UCP_API_MAJOR, UCP_API_MINOR) >= UCP_VERSION(1, 21));
+                return 0;
+            }
+        ''', dependencies: ucx_dep, name: 'UCX Device API (host-side, version >= 1.21)')
+
+    if have_gpu_side and have_host_side and doca_gpunetio_dep.found()
```
ofirfarjun7: Why we need to add doca check here? This check is not covered by `have_gpu_side` ?

michal-shalev: Okay so I just to conclude we're going to let users use device without forcing them to use pkg-config, I removed the check.

PR #1051, opened 2025-11-24
plugins/posix: per-engine IO queue

anton-nayshtut:
## What?
Introduce an IO queue abstraction and a per-engine IO queue and route all IOs for a given engine through that shared queue.

## Why?
This avoids the expensive per-IO AIO ring creation/setup and improves performance by reusing the same submission/completion infrastructure across requests.

## How?
The IO queue abstraction now only wraps the lowest-level IO-related API. It works with a set of IOs (writes or reads) while leaving the Xfer level logic to the upper layer (`posix_backend`).

Queue access is now `sync_mode`-aware: the engine uses the `nixlLock` primitive to protect shared queue state when running in STRICT or RW mode, while avoiding unnecessary locking otherwise.

## Performance

* 1 thread
* Batch size fixed at 256
* Direct I/O (O_DIRECT)
* Block sizes from 4 KiB → 256 KiB
* Same 8 GiB test file on /mnt/sdb1/test
* Same Nixlbench configuration otherwise

| Block Size | AIO (old) | AIO (new) | Δ AIO | URING (old) | URING (new) | Δ URING | POSIXAIO (old) | POSIXAIO (new) | Δ POSIXAIO
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
4 KiB | 0.392 | 0.419 | +6.9% | 0.400 | 0.418 | +4.5% | 0.107 | 0.104 | ≈ –3%
8 KiB | 0.450 | 0.478 | +6.2% | 0.458 | 0.476 | +4.0% | 0.172 | 0.169 | ≈ –2%
16 KiB | 0.496 | 0.503 | +1.4% | 0.498 | 0.501 | +0.7% | 0.257 | 0.254 | ≈ –1%
32 KiB | 0.517 | 0.522 | +1.0% | 0.518 | 0.522 | +0.7% | 0.337 | 0.338 | ≈ +0.3%
64 KiB | 0.521 | 0.525 | +0.8% | 0.523 | 0.525 | +0.5% | 0.362 | 0.361 | ≈ –0.3%
128 KiB | 0.526 | 0.529 | +0.6% | 0.527 | 0.529 | +0.4% | 0.378 | 0.368 | –2.6%
256 KiB | 0.526 | 0.529 | +0.5% | 0.527 | 0.529 | +0.4% | 0.429 | 0.428 | ≈ –0.2%

**Observation (READ)**
* AIO and URING clearly benefit from the new pool & queue settings - improvement is most visible at small/medium blocks (4–32 KiB), reaching **+4–7%**.
* At 128–256 KiB the gain becomes small (~0.4–0.6%).

| Block Size | AIO (old) | AIO (new) | Δ AIO | URING (old) | URING (new) | Δ URING | POSIXAIO (old) | POSIXAIO (new) | Δ POSIXAIO
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
4 KiB | 0.462 | 0.503 | +8.9% | 0.468 | 0.502 | +7.3% | 0.120 | 0.141 | +17.5%
8 KiB | 0.508 | 0.512 | +0.8% | 0.511 | 0.519 | +1.6% | 0.199 | 0.222 | +11.6%
16 KiB | 0.520 | 0.522 | +0.4% | 0.522 | 0.521 | ≈ 0% | 0.286 | 0.309 | +8.0%
32 KiB | 0.526 | 0.527 | +0.2% | 0.525 | 0.526 | +0.2% | 0.366 | 0.384 | +4.9%
64 KiB | 0.507 | 0.507 | ≈ 0% | 0.509 | 0.508 | ≈ 0% | 0.421 | 0.425 | +1.0%
128 KiB | 0.513 | 0.514 | +0.2% | 0.510 | 0.514 | +0.7% | 0.464 | 0.463 | ≈ 0%
256 KiB | 0.515 | 0.515 | ≈ 0% | 0.513 | 0.515 | +0.4% | 0.480 | 0.480 | ≈ 0%

**Observation (WRITE)**
* Biggest relative gains at small blocks - especially 4 KiB (**+7–9%** for AIO/URING, **+17%** for POSIXAIO).
* POSIXAIO write benefits noticeably more than in the read case (up to **+17%** at 4 KiB, still **+5–9%** up to 32 KiB).
* At ≥ 64 KiB almost all differences disappear - bandwidth saturates close to ~0.51–0.53 GB/s regardless of API.

**NOTE:** Performance in the multi-threaded tests remained unchanged. Results were within +/-1.5%, which is within the margin of error across all runs.


src/plugins/posix/file_api.h
```diff
@@ -0,0 +1,89 @@
...
+    typedef std::unique_ptr<nixlPosixFileApi> (*nixlPosixFileApiCreateFn)(void);
+
+    nixlPosixFileApi() {}
+
+    virtual ~nixlPosixFileApi() {}
+
+    virtual nixl_status_t
+    init(uint32_t max_chunks) = 0;
+    virtual nixl_status_t
+    add(int fd,
```
benlwalker: add/post/pump are eventually intended to be called on the thread pool threads, I believe? So my guess at the meaning of these is as follows:
1) add() adds an entry to the underlying submission queue, but doesn't "ring the doorbell"
2) post() "rings the doorbell", which is calling io_submit or io_uring_enter() to submit everything queued
3) pump() is checking for completions

I also think there must be exactly one of these nixlPosixFileApi per thread, right? And I think the underlying submission queue is allocated in init()?

If I'm close to correct here, I'd recommend the following renames:

1) Rename nixlPosixFileApi to nixlPosixIOQueue
2) Split add() into two - enqueue_read() and enqueue_write()
3) Rename "pump" to "poll" or "checkCompletions"

anton-nayshtut: You're 100% right on all your assumptions. 😄 I'll do the rename as you proposed. 👍

anton-nayshtut: Done! Thanks! For now, add() has been renamed to enqueue() as discussed offline.

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,89 @@
...
+            bool read,
+            nixlPosixIOQueueDoneCb clb,
+            void *ctx) = 0;
+    virtual nixl_status_t
+    post(void) = 0;
+    virtual nixl_status_t
+    poll(void) = 0;
+
+    static std::unique_ptr<nixlPosixIOQueue>
+    getApi(const std::string &api_name) {
```
iyastreb: `std::string_view` is superior alternative, because when you call this function with string literal then no intermediate std::string is created.
`std::string_view` covers both `const char *` and `std::string` cases

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,89 @@
...
+#include <map>
+#include <string>
+#include <memory>
+#include "backend_aux.h"
+
+typedef void (*nixlPosixIOQueueDoneCb)(void *ctx, uint32_t data_size, int error);
+
+class nixlPosixIOQueue {
+public:
+    typedef std::unique_ptr<nixlPosixIOQueue> (*nixlPosixIOQueueCreateFn)(void);
```
iyastreb: why not use std::function instead?
With std::function you can register lambdas with captures

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,89 @@
...
+
+    static uint32_t
+    normalizedMaxIOS(uint32_t max_ios) {
+        uint32_t m = std::max(MIN_IOS, max_ios);
+        m = std::min(m, MAX_IOS);
+        return m;
+    }
+
+protected:
+    static std::map<std::string, nixlPosixIOQueueCreateFn> apis;
```
iyastreb: This class is implicit singleton, which is often considered as anti-pattern
Maybe these fields should not be static?

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,89 @@
...
+    static const uint32_t MIN_IOS;
+    static const uint32_t MAX_IOS;
+};
+
+#define NIXL_POSIX_IO_QUEUE_REGISTER(io_queue_name, class_name)               \
```
iyastreb: Since it's C++, better use template, not macro

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,89 @@
...
+class nixlPosixIOQueue {
+public:
+    typedef std::unique_ptr<nixlPosixIOQueue> (*nixlPosixIOQueueCreateFn)(void);
+
+    nixlPosixIOQueue() {}
+
+    virtual ~nixlPosixIOQueue() {}
+
+    virtual nixl_status_t
+    init(uint32_t max_ios) = 0;
```
iyastreb: Normally ctor initializes the object, init seems redundant

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,220 @@
...
+    nixlPosixIoUringIO *
+    getBufInfo(struct iocb *io);
+    nixl_status_t
+    doSubmit(void);
+    nixl_status_t
+    doCheckCompleted(void);
+
+private:
+    uint32_t num_ios_;
+    bool initialized_;
```
iyastreb: Having `initialized_` field is bad practice.
When you call a ctor, object is either properly initialized or fails with exception

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,227 @@
...
+    nixlPosixLinuxAioIO *
+    getBufInfo(struct iocb *io);
+    nixl_status_t
+    doSubmit(void);
+    nixl_status_t
+    doCheckCompleted(void);
+
+private:
+    uint32_t num_ios_;
+    bool initialized_;
```
iyastreb: I propose to remove this

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.h
anton-nayshtut: For now, I replaced it by a static array. Please don't hesitate to suggest an alternative.

src/plugins/posix/io_queue.h
anton-nayshtut: The macro has changed. Please re-examine.

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.cpp
```diff
@@ -0,0 +1,58 @@
...
+#ifdef HAVE_LIBAIO
+    {"POSIXAIO", nixlPosixAioIOQueueCreate},
+#endif
+#ifdef HAVE_LIBURING
+    {"URING", nixlPosixIoUringIOQueueCreate},
+#endif
+#ifdef HAVE_LINUXAIO
+    {"AIO", nixlPosixLinuxAioIOQueueCreate},
+#endif
+    {nullptr, nullptr}};
```
iyastreb: empty entry is redundant

src/plugins/posix/io_queue.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+std::unique_ptr<nixlPosixIOQueue>
+nixlPosixIOQueue::instantiate(std::string_view io_queue_type, uint32_t max_ios) {
+    int i = 0;
+    while (factories[i].name != nullptr) {
+        if (io_queue_type.compare(factories[i].name) == 0) {
+            return factories[i].createFn(max_ios);
+        }
+        i++;
+    }
```
iyastreb:
```
    for (const auto &factory : factories) {
        if (io_queue_type == factory.name) {
            return f.createFn(max_ios);
```

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,221 @@
...
+    io->clb_ = clb;
+    io->ctx_ = ctx;
+    io->io_.data = io;
+    ios_to_submit_.push_back(io);
+
+    return NIXL_SUCCESS;
+}
+
+nixlPosixLinuxAioIOQueue::~nixlPosixLinuxAioIOQueue() {
+    if (initialized_) {
```
iyastreb: initialized to be removed

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+    using nixlPosixIOQueueCreateFn =
+        std::function<std::unique_ptr<nixlPosixIOQueue>(uint32_t max_ios)>;
+
+    nixlPosixIOQueue(uint32_t max_ios) : max_ios_(normalizedMaxIOS(max_ios)) {}
+
+    virtual ~nixlPosixIOQueue() {}
+
+    virtual nixl_status_t
+    enqueue(int fd,
```
iyastreb, line 43: This enqueue encapsulates all possible params, even if some of them are not used by an actual impl, right?
IMO this is not very clean from design perspective

At least we can define a structure for these params, need to think more on that
The class itself may be a template, so that we can encapsulate the common structures (free_list, etc)
This way you can reuse much more than just an interface

Maybe it should be something like:
```
template <typename Entry>
class nixlQ {
    protected:
        std::list<Entry> free_list_;  // reuse common logic in all impl
        ..
}

File: src/plugins/posix/linux_aio_io_queue.cpp [hunk 13]
```diff
@@ -0,0 +1,221 @@
...
+    doSubmit(void);
+    nixl_status_t
+    doCheckCompleted(void);
+
+private:
+    bool initialized_;
+    io_context_t io_ctx_; // I/O context
+    std::vector<nixlPosixLinuxAioIO> ios_; // Array of ios
+    std::list<nixlPosixLinuxAioIO *> free_ios_; // List of free ios
+    std::list<nixlPosixLinuxAioIO *> ios_to_submit_; // List of ios to submit
```
iyastreb: This can be reused from the base class template

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: Missed it. Sorry. Fixed!

src/plugins/posix/io_queue.h
anton-nayshtut: > This enqueue encapsulates all possible params, even if some of them are not used by an actual impl, right?
No. All params are used by each impl but they're treated differently.

I'll think about templates though. Thanks!

anton-nayshtut: On the other hand, the whole idea of this design is to properly use inheritance - have generic `nixlPosixIOQueue` pointer and instantiate it as a specific IO Queue type. As far as I understand, templates are incompatible with this idea. Am I right?

src/plugins/posix/io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.h
anton-nayshtut: Oh! I get it. I can have an additional level of inheritance, which would be a template. Will do. Thanks!

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.h
anton-nayshtut: It has been removed.

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+#include "io_queue.h"
+#include "common/nixl_log.h"
+#include <liburing.h>
+#include <absl/strings/str_format.h>
+
+#define MAX_IO_SUBMIT_BATCH_SIZE 64
+#define MAX_IO_CHECK_COMPLETED_BATCH_SIZE 64
+
```
benlwalker, line 25: I feel like the capitalization on IO should be consistent everywhere. Not Io vs IO.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,197 @@
...
+nixlPosixIoUringIOQueue::nixlPosixIoUringIOQueue(uint32_t max_ios) : nixlPosixIOQueueImpl<nixlPosixIoUringIO>(max_ios) {
+    io_uring_params params = {};
+    if (io_uring_queue_init_params(max_ios_, &uring, &params) < 0) {
+        throw std::runtime_error(
+            absl::StrFormat("Failed to initialize io_uring instance: %s", nixl_strerror(errno)));
+    }
+}
+
+nixl_status_t
+nixlPosixIoUringIOQueue::doSubmit(void) {
```
benlwalker: Do we need doSubmit and doCheckCompleted to be different functions from post/poll? Or can we inline them.

anton-nayshtut:
> can you please describe the rationale of the changes? if any code reorganization, can it be split in multiple PRs?

Sure, I can try! 😄  The idea is to minimize the API abstraction to the absolute minimum, have it operate on batches of IOs (which is what real file APIs care about), and concentrate all the Xfer-level logic within the POSIX plugin itself.

For now, we’re still creating an API queue per Xfer. However, in the upcoming patches, I’ll introduce a threading model to the POSIX plugin. Then, we’ll only create an API queue per thread, and it will handle all the Xfers assigned to that thread.

We believe this approach will let us add multithreading support to the POSIX plugin, while maximizing performance and keeping the common logic in one place.

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: I agree. Please point out any inconsistencies you’ve found - I’d be happy to fix them.

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: We can inline them, but I thought that this way it’s much more obvious what each function does.

Anyway, I’ll inline them if you think it’s worth it.

Please advise.

brminich:
@vvenkates27 can you pls review?
i think bettter merge it before #1130

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,149 @@
...
+
+#include <stdint.h>
+#include <list>
+#include <memory>
+#include <vector>
+#include <functional>
+#include <cassert>
+#include "backend_aux.h"
+
+typedef void (*nixlPosixIOQueueDoneCb)(void *ctx, uint32_t data_size, int error);
```
vvenkates27: Consider` using std::function<void(void*, uint32_t, int)> instead of raw function pointer for type safety?
sth like?
 using nixlPosixIOQueueCreateFn =
        std::function<std::unique_ptr<nixlPosixIOQueue>(uint32_t max_ios)>;

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,149 @@
...
+    virtual nixl_status_t
+    poll(void) = 0;
+
+    static std::unique_ptr<nixlPosixIOQueue>
+    instantiate(std::string_view io_queue_type, uint32_t max_ios);
+
+protected:
+    static uint32_t
+    normalizedMaxIOS(uint32_t max_ios) {
+        uint32_t m = std::max(MIN_IOS, max_ios);
```
vvenkates27: Can we use std::clamp?
sth like return std::clamp(max_ios, MIN_IOS, MAX_IOS);

src/plugins/posix/posix_backend.cpp
```diff
@@ -23,118 +23,110 @@
...
-    static nixlPosixQueue::queue_t getQueueType(const nixl_b_params_t* custom_params) {
-        using queue_t = nixlPosixQueue::queue_t;
-
-        // Check for explicit backend request
-        if (custom_params) {
-            // First check if AIO is explicitly requested
-            if (custom_params->count("use_aio") > 0) {
-                const auto& value = custom_params->at("use_aio");
-                if (value == "true" || value == "1") {
-                    if (!QueueFactory::isLinuxAioAvailable()) {
```
vvenkates27: This Check is removed why?
Previously, we checked QueueFactory::isLinuxAioAvailable() before returning the queue type. Now we just return the string without checking if the backend is actually available. This could lead to confusing errors at queue creation time instead of early detection.?

src/plugins/posix/posix_backend.cpp
```diff
@@ -143,202 +135,230 @@ namespace {
...
+                                           const nixl_b_params_t *params)
+    : operation(op),
+      local(loc),
+      remote(rem),
+      opt_args(args),
+      custom_params_(params),
+      queue_depth_(loc.descCount()) {
+
+    std::string io_queue_type = params->at("io_queue_type");
+    if (io_queue_type.empty()) {
```
vvenkates27: We should probably check 
it == params->end() ? 
asn params->at("io_queue_type") can throw std::out_of_range?

src/plugins/posix/posix_backend.cpp
```diff
@@ -143,202 +135,230 @@ namespace {
...
+        NIXL_ERROR << absl::StrFormat("Failed to initialize io queue: %s", e.what());
         return NIXL_ERR_BACKEND;
     }
 }
 
-nixl_status_t nixlPosixBackendReqH::prepXfer() {
+nixl_status_t
+nixlPosixBackendReqH::prepXfer() {
+    return NIXL_SUCCESS;
+}
```
vvenkates27, line 188: Is prepXfer a No op and all pre is moved to PostXfer/checkXfer?

src/plugins/posix/posix_backend.h
```diff
@@ -23,107 +23,131 @@
...
-    std::unique_ptr<nixlPosixQueue> queue;           // Async I/O queue instance
-    const nixlPosixQueue::queue_t   queue_type_;     // Type of queue used
+    const nixl_xfer_op_t &operation; // The transfer operation (read/write)
+    const nixl_meta_dlist_t &local; // Local memory descriptor list
+    const nixl_meta_dlist_t &remote; // Remote memory descriptor list
+    const nixl_opt_b_args_t *opt_args; // Optional backend-specific arguments
+    const nixl_b_params_t *custom_params_; // Custom backend parameters
+    const int queue_depth_; // Queue depth for async I/O
+    std::unique_ptr<nixlPosixIOQueue> io_queue_; // Async I/O queue instance
+    int num_confirmed_ios_; // Number of confirmed IOs
```
vvenkates27, line 36: num_confirmed_ios_ is not initialized in the constructor. While it's set to 0 in postXfer(), it could be undefined if checkXfer() is called before postXfer(). Should initialize in constructor or use int num_confirmed_ios_ = 0;.

vvenkates27:
> @vvenkates27 can you pls review? i think bettter merge it before #1130

Done.

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_queue.h
anton-nayshtut: That’s cool! I didn’t know it existed. Thanks! All done!

src/plugins/posix/posix_backend.h
anton-nayshtut: Great catch! Thanks!

I fixed it a bit differently. I set `num_confirmed_ios_ = queue_depth_`, so if `checkXfer` is called before `postXfer`, it will return `NIXL_SUCCESS` right away. I also added a comment explaining this logic.

WDYT?

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Sorry, Autoformat and bad rebase. I think we must merge [PR#1130](https://github.com/ai-dynamo/nixl/pull/1130) ASAP to create a baseline and avoid such issues.

anton-nayshtut: @vvenkates27 , please ignore my previous commit.

**Briefly:** We don’t have `QueueFactory` anymore - this patch removes it.

_Details:_

Instead, we now have IO queues which inherit `class nixlPosixIOQueue`. When an IO queue is compiled, it exports a constructor (`nixlPosixIOQueue::nixlPosixIOQueueCreateFn`). These constructors are stored in the `factories` array (see `io_queue.cpp`), so they can be instantiated by name later using `nixlPosixIOQueue::instantiate`.

This approach is much more generic and allows us to handle all IO queue types the same way, using their base class API (`class nixlPosixIOQueue`).

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Yes. It's been moved to `PostXfer()` or, more specifically, to `io_queue_->enqueue()`

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Thanks! Fixed!

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: The comment seems to be outdated.

src/plugins/posix/posix_backend.h
vvenkates27: Nice!
There's however a  similar issue in io_queue.h - the template class nixlPosixIOQueueImpl has num_ios_outstanding_ which is also not initialized:
nixlPosixIOQueueImpl(uint32_t max_ios)     : nixlPosixIOQueue(max_ios), ios_(max_ios_), num_ios_outstanding_(0) {
Also minor typo: "iitialize" → "initialize" in the comment.

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -0,0 +1,163 @@
...
+        submitted_ios++;
+    }
+
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlPosixAioIOQueue::checkCompleted(uint32_t &completed_ios) {
+    assert(!ios_in_flight_.empty());
```
vvenkates27: We should handle this more gracefully instead of an assert, sth like?

completed_ios = 0;
if (ios_in_flight_.empty()) {
    return NIXL_SUCCESS;
}

src/plugins/posix/posix_backend.cpp
vvenkates27: Got it, might be easier to control/rate limit submission based on API.

src/plugins/posix/posix_backend.cpp
vvenkates27: The new factory approach looks good, but getIoQueueType() always returns "AIO" as default regardless of what's compiled. If only libaio is available this might fail, Should we cascade through available options like in the old approach?

src/plugins/posix/posix_backend.cpp
```diff
@@ -73,87 +72,68 @@ castPosixHandle(nixlBackendReqH *handle) {
...
         // First check if AIO is explicitly requested
         if (custom_params->count("use_aio") > 0) {
             const auto &value = custom_params->at("use_aio");
             if (value == "true" || value == "1") {
-                if (!QueueFactory::isLinuxAioAvailable()) {
-                    NIXL_ERROR << "linux_aio backend requested but not available at runtime";
-                    return queue_t::UNSUPPORTED;
-                }
-                return queue_t::AIO;
+                return "AIO";
```
vvenkates27, line 83: The new code always returns "AIO" as default, even if only io_uring is compiled in. Should cascade between different APIs like the old approach?
or have priority? Something like this?

#ifdef HAVE_LINUXAIO
    return "AIO";           // First priority
#elif defined(HAVE_LIBURING)
    return "URING";         // Second priority (only if AIO not compiled)
#elif defined(HAVE_LIBAIO)
    return "POSIXAIO";      // Third priority
#endif

src/plugins/posix/posix_aio_io_queue.cpp
anton-nayshtut: I'm not entirely sure. My line of thoughts was the following:

The `checkCompleted` is called from `nixlPosixIOQueueImpl::complete` which, win turn, is called from `nixlPosixIOQueueImpl::poll`. 

`ios_in_flight_` can only be empty if `submitBatch` had either not been called or returned `submitted_ios == 0`. In both cases `num_ios_outstanding_`  would be 0. Which means that `nixlPosixIOQueueImpl::poll` allowed an illegal call.

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Nice catch! Thanks! I prefer to keep ifdefs in one place, so I solved it using the `nixlPosixIOQueue::getDefaultIoQueueType` API. Hope it's OK!

src/plugins/posix/posix_backend.h
anton-nayshtut: Thanks! Fixed!

It’s just professional habit - when you allocate with `calloc`, you don’t have to worry about zeroing afterwards. I’ll pay attention.

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Fixed! Thanks!

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+protected:
+    nixl_status_t
+    submit(void) {
+        if (ios_to_submit_.empty()) {
+            return NIXL_SUCCESS; // No blocks to submit
+        }
+
+        assert(num_ios_outstanding_ <= MAX_OUTSTANDING_IOS);
```
aranadive: These are removed in release builds, so please fail or log accordingly. Generally, we should avoid asserts since in practice the condition will not be handled.

anton-nayshtut: I completely agree that we should avoid asserts in user-facing parts of our code. Users may misuse (or even abuse) the library APIs, so in those cases I try to add proper error handling.

That said, some parts of our code are purely internal. In those areas, API misuse usually points to a logical issue or a bug on our side. In those cases, I think asserts can be appropriate because they help us avoid silently hiding underlying problems.

In this specific case, the **internal** limit for `num_ios_outstanding_` is `MAX_OUTSTANDING_IOS`. So the only reason this assertion would fire is if we have a bug in our code, which is why I added an assert here._

Please advise.

brminich: maybe better `NIXL_ASSERT`?

src/plugins/posix/io_queue.h
```diff
@@ -0,0 +1,152 @@
...
+
+    nixl_status_t
+    complete(void) {
+        uint32_t completed_ios = 0;
+        nixl_status_t status = checkCompleted(completed_ios);
+        if (status != NIXL_SUCCESS) {
+            return status;
+        }
+
+        assert(completed_ios <= num_ios_outstanding_);
```
brminich: NIXL_ASSERT

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,173 @@
...
+        throw std::runtime_error(
+            absl::StrFormat("Failed to initialize io_uring instance: %s", nixl_strerror(errno)));
+    }
+}
+
+nixl_status_t
+nixlPosixIoUringIOQueue::submitBatch(uint32_t to_submit, uint32_t &submitted_ios) {
+    assert(!ios_to_submit_.empty());
+    assert(ios_to_submit_.size() >= to_submit);
+
```
brminich: NIXL_ASSERT

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+nixl_status_t
+nixlPosixLinuxAioIOQueue::submitBatch(uint32_t to_submit, uint32_t &submitted_ios) {
+    struct iocb *ios[MAX_OUTSTANDING_IOS];
+    nixlPosixLinuxAioIO *ios_to_submit[MAX_OUTSTANDING_IOS];
+
+    assert(!ios_to_submit_.empty());
+    assert(ios_to_submit_.size() >= to_submit);
+
```
brminich: NIXL_ASSERT

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    ios_to_submit_.push_back(io);
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlPosixAioIOQueue::submitBatch(uint32_t to_submit, uint32_t &submitted_ios) {
+    assert(!ios_to_submit_.empty());
+    assert(ios_to_submit_.size() >= to_submit);
```
brminich: NIXL_ASSERT

src/plugins/posix/posix_aio_io_queue.cpp
brminich: NIXL_ASSERT then?

src/plugins/posix/io_queue.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
iyastreb: minor: for new files I think only 2026 is needed

src/plugins/posix/io_queue.h
anton-nayshtut: Sure! Thanks! Done!

src/plugins/posix/io_queue.h
anton-nayshtut: Done! Thanks!

src/plugins/posix/posix_aio_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: Done! Thanks!

brminich (approved):

brminich:
@aranadive @vvenkates27 can we merge?

src/plugins/posix/io_uring_io_queue.cpp
benlwalker: Let's make them all capitals - so this one is nixlPosixIOUringIO instead.

benlwalker: I think it's all the places you wrote IoUring just become IOUring

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,197 @@
...
+    io->ctx_ = ctx;
+
+    ios_to_submit_.push_back(io);
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlPosixIoUringIOQueue::post(void) {
+    nixl_status_t status = doSubmit();
```
benlwalker: This is the only place doSubmit() is called so please put the implementation directly into this function.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+nixl_status_t
+nixlPosixIoUringIOQueue::poll(void) {
+    nixl_status_t status = post();
+    if (status < 0) {
+        return status;
+    }
+
+    return doCheckCompleted();
```
benlwalker, line 178: This is the only place doCheckCompleted() is called so please put the implementation directly into this function.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include "io_queue.h"
+#include "common/nixl_log.h"
+#include <liburing.h>
+#include <absl/strings/str_format.h>
+
+#define MAX_IO_SUBMIT_BATCH_SIZE 64
+#define MAX_IO_CHECK_COMPLETED_BATCH_SIZE 64
+
+struct nixlPosixIoUringIO {
+public:
```
benlwalker, line 27: Remove this public - that's the default visibility for struct.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (status < 0) {
+        return status;
+    }
+
+    return NIXL_IN_PROG;
+}
+
+nixl_status_t
+nixlPosixLinuxAioIOQueue::poll(void) {
+    nixl_status_t status = doSubmit();
```
benlwalker: Can you make this call post() like the io_uring implementation?

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (free_ios_.size() == max_ios_) {
+        return NIXL_SUCCESS; // All ios are free now
+    }
+
+    return NIXL_IN_PROG; // Some blocks are in flight, need to check again
+}
+
+nixl_status_t
+nixlPosixLinuxAioIOQueue::post(void) {
+    nixl_status_t status = doSubmit();
```
benlwalker: Inline the doSubmit() logic here.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+nixl_status_t
+nixlPosixLinuxAioIOQueue::poll(void) {
+    nixl_status_t status = doSubmit();
+    if (status < 0) {
+        return status;
+    }
+
+    return doCheckCompleted();
```
benlwalker, line 193: inline the doCheckCompleted() logic here.

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -0,0 +1,190 @@
...
+    nixl_status_t status = doSubmit();
+    if (status < 0) {
+        return status;
+    }
+    return NIXL_IN_PROG;
+}
+
+nixl_status_t
+nixlPosixAioIOQueue::poll(void) {
+    nixl_status_t status = doSubmit();
```
benlwalker: Make this call post()

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -0,0 +1,190 @@
...
+            break;
+        }
+    }
+
+    return ios_in_flight_.empty() ? NIXL_SUCCESS : NIXL_IN_PROG;
+}
+
+nixl_status_t
+nixlPosixAioIOQueue::post(void) {
+    nixl_status_t status = doSubmit();
```
benlwalker: Inline doSubmit logic here

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+nixl_status_t
+nixlPosixAioIOQueue::poll(void) {
+    nixl_status_t status = doSubmit();
+    if (status < 0) {
+        return status;
+    }
+
+    return doCheckCompleted();
```
benlwalker, line 175: Inline doCheckCompleted logic here

benlwalker (changes requested):
Just some code organization comments but otherwise looks good to me.

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: Done! I also normalized the queue implementation names. They're now called `nixlPosixIOQueue<IMPL>`:
* `nixlPosixIOQueueLinuxAIO`
* `nixlPosixIOQueueUring`
* `nixlPosixIOQueueAIO` - I wasn’t sure if this should be `nixlPosixIOQueuePosixAIO`, but the double “Posix” felt awkward, so I left it as-is for now.

WDYT?

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: This code gets removed in the next commit (batched submission). Should I spend time improving code that’s going away anyway?

BTW, I think we can just squash these commits. I left them separate mainly to show that I both refactored the old code and implemented batched submission for all queue types.

WDYT? @benlwalker

src/plugins/posix/meson.build
```diff
@@ -53,7 +52,7 @@ endif
 
 if has_posix_aio
     compile_defs += ['-DHAVE_LIBAIO']
```
benlwalker: I think we really need to change the name of this HAVE_LIBAIO define. There is a library called libaio, but it's for Linux AIO. POSIX AIO does not require a library - it's in glibc. I think it should be named HAVE_POSIXAIO, or we just have nothing defined at all because we assume we have POSIX AIO available.

src/plugins/posix/posix_backend.cpp
benlwalker, line 196: You moved the queue from the each request to the engine. That means multiple requests share the same queue now, so this check for completion isn't correct. We aren't waiting for the entire queue to drain, we're just waiting for all of the tasks in the queue associated with this particular request to finish.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -146,7 +146,7 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
     int rc;
     struct timespec timeout = {0, 0};
 
-    if (free_ios_.size() == max_ios_) {
+    if (free_ios_.size() == ios_pool_size_) {
```
benlwalker, line 149: This isn't tracking completions correctly. Multiple requests have sent ios on this queue, so those requests need to check if their individual ios are done, not if the queue is entirely drained.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -133,7 +134,7 @@ nixlPosixIOQueueUring::doCheckCompleted(void) {
     // Mark all seen
     io_uring_cq_advance(&uring, count);
 
-    if (free_ios_.size() == max_ios_) {
+    if (free_ios_.size() == ios_pool_size_) {
```
benlwalker, line 136: This isn't tracking completions correctly. Multiple requests have sent ios on this queue, so those requests need to check if their individual ios are done, not if the queue is entirely drained. But right now it's only completing requests if the entire queue drains to empty.

src/plugins/posix/meson.build
anton-nayshtut: I agree. But it's not a part of this PR. Will create a separate one.

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: I believe the tracking is correct.

There are two different kinds of tracking going on here:

* `nixlPosixBackendReqH` tracks progress by incrementing `num_confirmed_ios_` and comparing it against `queue_depth_`. This ensures that `checkXfer()` for a specific `nixlPosixBackendReqH` returns `NIXL_IN_PROG` while some chunks are still unconfirmed, and `NIXL_SUCCESS` once the entire request completes.
* Each queue tracks all IOs submitted on it - this is what this line is about. It effectively means: “if there are no more in-flight IOs, return `NIXL_SUCCESS`”. The intent was for the queue to signal to the upper layer when it has nothing left to do. However, as you can see, that signal is currently ignored by `nixlPosixBackendReqH::checkXfer()` - it only checks for errors.

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: See my commend above.

src/plugins/posix/meson.build
anton-nayshtut: See https://github.com/ai-dynamo/nixl/pull/1296

src/plugins/posix/linux_aio_io_queue.cpp
benlwalker: Oh I see - NIXL_SUCCESS AND NIXL_IN_PROG are both >= 0.

brminich:
@anton-nayshtut can you pls resolve conflicts?

src/plugins/posix/io_uring_io_queue.cpp
anton-nayshtut: Done! Thanks!D

src/plugins/posix/linux_aio_io_queue.cpp
anton-nayshtut: Done! Thanks!

brminich (approved):

benlwalker:
+1 vote from me

PR #1052, opened 2025-11-25
CI Podman build: use native overlay and skip git install

Alexey-Rivkin:
### What?
Remove redundant `git` installation and reset Podman storage configuration in the CI pipeline.

### Why?
1. **Stability:** `yum install git` fails frequently due to network/mirror issues. The base image already has `git`.
2. **Performance:** Deleting `storage.conf` removes the forced `fuse-overlayfs` driver, enabling the much faster native `overlay` driver to prevent build hangs and reduce duration.

### DIFF
--- a/.ci/jenkins/lib/build-container-matrix.yaml
+++ b/.ci/jenkins/lib/build-container-matrix.yaml
@@ -60,7 +60,7 @@ steps:
       rm -f /etc/containers/storage.conf
       podman system reset -f || true
       ln -sfT $(type -p podman) /usr/bin/docker
-      yum install -y git gettext
+      yum install -y gettext
 
   - name: Build NIXLBench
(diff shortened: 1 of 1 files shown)


.ci/jenkins/lib/build-container-matrix.yaml
```diff
@@ -57,6 +57,8 @@ steps:
   - name: Prepare
     run: |
       # Setup podman and dependencies
+      rm -f /etc/containers/storage.conf
+      podman system reset -f || true
       ln -sfT $(type -p podman) /usr/bin/docker
       yum install -y git gettext
```
dpressle: you dont need git any more right?

Alexey-Rivkin: The image already includes git v2.51.1, which is recent enough. No extra installation needed.

PR #1057, opened 2025-11-25
Enable libfabric backend plugin with CXI provider for Slingshot network

joshishr:
Enable CXI provider support in libfabric backend plugin for HPE's Slingshot network.

### DIFF
--- a/src/utils/libfabric/libfabric_common.cpp
+++ b/src/utils/libfabric/libfabric_common.cpp
@@ -50,6 +50,9 @@ getAvailableNetworkDevices() {
     hints->mode = FI_CONTEXT;
     hints->ep_attr->type = FI_EP_RDM;
 
+    // Add CXI-compatible memory registration mode
+    hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_ENDPOINT | FI_MR_ALLOCATED | FI_MR_PROV_KEY;
+
     int ret = fi_getinfo(FI_VERSION(1, 18), NULL, NULL, 0, hints, &info);
     if (ret) {
         NIXL_ERROR << "fi_getinfo failed " << fi_strerror(-ret);
@@ -85,7 +88,9 @@ getAvailableNetworkDevices() {
         }
     }
 
-    if (provider_device_map.find("efa") != provider_device_map.end()) {
+    if (provider_device_map.find("cxi") != provider_device_map.end()) {
+        return {"cxi", provider_device_map["cxi"]};
+    } else if (provider_device_map.find("efa") != provider_device_map.end()) {
         return {"efa", provider_device_map["efa"]};
     } else if (provider_device_map.find("sockets") != provider_device_map.end()) {
         return {"sockets", {provider_device_map["sockets"][0]}};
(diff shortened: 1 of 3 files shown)


ryanhankins:
The most recent force commit seems to fix the formatting changes in the checks, so I'm not sure why the checks continue to fail.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -179,7 +179,8 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
 
         // For TCP providers, use offset 0 instead of virtual address
         // TCP providers don't support FI_MR_VIRT_ADDR and expect offset-based addressing
-        if (data_rails_[rail_id]->provider_name == "tcp" ||
+        if (data_rails_[rail_id]->provider_name == "cxi" ||
+            data_rails_[rail_id]->provider_name == "tcp" ||
```
sielicki, line 183: @akkart-aws was there a reason you couldn't just check for FI_MR_VIRT_ADDR directly here?

brminich:
pls note clang format check errors

ryanhankins:
> pls note clang format check errors

Thank you for looking at this.  I believe the force commit fixed the clang format errors, but it doesn't seem like the Clang Format Check picked up that they were fixed.

ryanhankins:
> pls note clang format check errors

Far as I can tell, this commit resolved the errors.

image

src/utils/libfabric/libfabric_rail_manager.cpp
akkart-aws: No, we initially started with support for EFA and TCP since we could test these providers with AWS Infrastructure.
We can consolidate the check to just FI_MR_VIRT_ADDR.

src/utils/libfabric/libfabric_common.cpp
```diff
@@ -50,6 +50,9 @@ getAvailableNetworkDevices() {
     hints->mode = FI_CONTEXT;
     hints->ep_attr->type = FI_EP_RDM;
 
+    // Add CXI-compatible memory registration mode
+    hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_ENDPOINT | FI_MR_ALLOCATED | FI_MR_PROV_KEY;
```
akkart-aws, line 54: Setting `hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_ENDPOINT | FI_MR_ALLOCATED | FI_MR_PROV_KEY` breaks EFA provider discovery because EFA doesn't support `FI_MR_ENDPOINT` (it uses `FI_MR_VIRT_ADDR` instead).

Recommend removing the mr_mode hint entirely and letting each provider report its own capabilities. This allows both CXI and EFA to work correctly.

src/utils/libfabric/libfabric_rail_manager.cpp
sielicki: Sure, makes sense. I think it would be simpler and more portable to other providers to just check the bit here instead, if it works. It's a lot quicker to check the one bit instead of these implicit string conversions and comparisons.

Thank you for taking a look.

ovidiusm:
Should this be included in the upcoming release (0.9.0)? FYI: Code freeze is on December 12th

ryanhankins:
This can be closed.  It was merged as #1242.

joshishr:
Closing the pull request.

PR #1058, opened 2025-11-25
Object: Multi-agent multi-part uploads

cheese-head:
## What?
Adds multipart upload support to the nixl object storage backend, allowing large objects to be uploaded in parts rather than as a single operation.
## Why?
multiple agents often need to collaborate on writing a single large object to storage. With multipart uploads, each agent can independently upload its assigned portion without coordinating with others or maintaining a centralized buffer. 
## How?
The implementation extends the existing transfer API to support multipart operations:
Metadata format: Extended the metaInfo field to accept multipart parameters using the format `object_key;upload_id=X;part_number=Y`. 
Regular uploads still work with just the object key.

Upload flow: 
When postXfer detects multipart parameters, it calls uploadPartAsync instead of putObjectAsync. Each part upload returns an ETag that's stored in a new uploadIdToETags_ map.

ETag retrieval: The queryMem method can now query ETags by upload_id, allowing clients to retrieve all ETags needed to complete the multipart upload.

Better diagnostics: Added detailed logging throughout the upload/download paths to make debugging easier, including object error codes, HTTP status, and request IDs.

### DIFF
--- a/src/plugins/obj/obj_s3_client.cpp
+++ b/src/plugins/obj/obj_s3_client.cpp
@@ -208,11 +208,11 @@ awsS3Client::putObjectAsync(std::string_view key,
             const Aws::S3::Model::PutObjectOutcome &outcome,
             const std::shared_ptr<const Aws::Client::AsyncCallerContext> &context) {
             if (outcome.IsSuccess()) {
-                NIXL_DEBUG << absl::StrFormat("S3 PUT SUCCESS: key=%s, size=%zu bytes", key, len);
+                NIXL_DEBUG << absl::StrFormat("PUT SUCCESS: key=%s, size=%zu bytes", key, len);
                 callback(true);
             } else {
                 const auto &error = outcome.GetError();
-                std::string error_msg = absl::StrFormat("S3 PUT FAILED: key=%s, size=%zu bytes\n"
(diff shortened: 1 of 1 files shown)


yosefe:
Is it possible to add example/CI test/gtest to demonstrate how multipart can be used on NIXL NB API level (python/C++)?

cheese-head:
@yosefe Would a python example suffice? I have an example of how to initiate the multipart upload with boto3, execute the transfers with nixl, and then complete the multipart upload with boto3. 
It's currently in `nixl/examples/python/obj_multipart_upload_example.py` in my local branch

aranadive:
> @yosefe Would a python example suffice? I have an example of how to initiate the multipart upload with boto3, execute the transfers with nixl, and then complete the multipart upload with boto3. It's currently in `nixl/examples/python/obj_multipart_upload_example.py` in my local branch

Yes. Can you include that in this PR?

ovidiusm:
Are we planning to include this in NIXL 0.9.0? Code freeze is on Friday December 12th

cheese-head:
This not required for the next release. More scoping needs to be done

brminich:
@cheese-head is it targeted for 0.10.0 release? Can you pls resolve conflicts if it is?

PR #1060, opened 2025-11-27
nixlbench: allocate page aligned memory by default

ovidiusm:
## What?
Nixlbench should allocate page aligned memory for best performance in tests involving DRAM.

## Why?
* Transfers from unaligned memory may be slower. Applications should use aligned memory wherever possible for best performance.
* Easier comparison with ucx_perftest and similar


brminich:
should `xferBenchUtils::checkConsistency` be also updated?

ovidiusm:
> should `xferBenchUtils::checkConsistency` be also updated?

I do not think so, it is outside the time measurements

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
rakhmets: Looks like an overcomplication.
Can be just: `AllocationType allocation_type = AllocationType::POSIX_MEMALIGN`.
And the line `AllocationType type = allocation_type.value_or(AllocationType::POSIX_MEMALIGN);` can be removed.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -365,10 +365,7 @@ std::optional<xferBenchIOV>
 xferBenchNixlWorker::initBasicDescDram(size_t buffer_size, int mem_dev_id) {
     void *addr;
 
-    AllocationType type = AllocationType::CALLOC;
-    if (xferBenchConfig::storage_enable_direct) {
-        type = AllocationType::POSIX_MEMALIGN;
-    }
+    AllocationType type = AllocationType::POSIX_MEMALIGN;
 
     if (!allocateXferMemory(buffer_size, &addr, type)) {
```
rakhmets:
```suggestion
    if (!allocateXferMemory(buffer_size, &addr)) {
```

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -577,11 +574,7 @@ xferBenchNixlWorker::initBasicDescFile(size_t buffer_size, xferFileState &fstate
...
     // Fill up with data
     void *buf;
-    AllocationType type = AllocationType::MALLOC;
-
-    if (xferBenchConfig::storage_enable_direct) {
-        type = AllocationType::POSIX_MEMALIGN;
-    }
+    AllocationType type = AllocationType::POSIX_MEMALIGN;
 
     if (!allocateXferMemory(buffer_size, &buf, type) || !buf) {
```
rakhmets:
```suggestion
    if (!allocateXferMemory(buffer_size, &addr)) {
```

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
ovidiusm: I agree

ovidiusm:
At this point we can remove the function completely and use posix_memalign?

rakhmets:
> At this point we can remove the function completely and use posix_memalign?

I think there's still the common code that will be in this function.
```C++
namespace {
void *
allocateXferMemory(size_t buffer_size) {
    if (xferBenchConfig::page_size == 0) {
        std::cerr << "Error: Invalid page size returned by sysconf" << std::endl;
        return nullptr;
    }

    void *memptr;
    if (posix_memalign(&memptr, xferBenchConfig::page_size, buffer_size) != 0) {
        std::cerr << "Failed to allocate " << buffer_size
                  << " bytes of page-aligned DRAM memory" << std::endl;
        return nullptr;
    }

    memset(memptr, 0, buffer_size);
    return memptr;
}
}
```

brminich:
> > should `xferBenchUtils::checkConsistency` be also updated?
> 
> I do not think so, it is outside the time measurements

let me rephrase -  why do we need different allocation methods in that method?

PR #1061, opened 2025-11-27
CI: Change k8s namespace to nbu-swx-nixl

Alexey-Rivkin:
## What?
Update the Kubernetes namespace in the CI build matrices from `swx-media` to `nbu-swx-nixl`.

## Why?
The `swx-media` namespace is currently overutilized by other projects, causing resource constraints. 
The `nbu-swx-nixl` namespace is unused and available for exclusive use.


Alexey-Rivkin:
K8s namespaces are tightly coupled to their Jenkins instances, so spanning jobs into another namespace is blocked by design.

Issue #972, opened 2025-10-30
Logging issues while using nixl agents


vshawrh:
Hi Team,
Due to https://github.com/ai-dynamo/nixl/blob/main/src/api/python/_api.py#L26
We are unable to use loggers in our local code.
As the the nixl logger supersedes while using nixl agent
The logger.info doesn't work

i.e.
def test_nixl_api(self):
        """Test that nixl API can be imported."""
        test_name = "test_nixl_api"
        try:
            from nixl._api import nixl_agent, nixl_agent_config
            logger.info("nixl._api imported successfully")
            log_test_result(test_name, 'passed', f"NIXL API imported successfully on {DEVICE}")
            assert True
        except Exception as e:
            log_test_result(test_name, 'failed', f"NIXL API not available: {e}")
            pytest.fail(f"NIXL API not available: {e}")

gleon99:
Hi @vshawrh 
Any chance you imported NIXL **after** defining your logger? In such case it would obviously overwrite it..

gleon99:
Hi @vshawrh  can reply please?

PR #1062, opened 2025-11-27
API/PYTHON: controlled export of python APIs using __all__

roiedanino:
## What?
Added a controlled export of API classes in `_api/__init__.py`

## Why?
Currently, users are importing everything being created under the `python` folder, including the logger object for example. This way, NIXL developers have full control of whether a certain class/object / constant is exposed to the user.

### DIFF
--- a/src/bindings/rust/src/descriptors/reg.rs
+++ b/src/bindings/rust/src/descriptors/reg.rs
@@ -15,6 +15,7 @@
 
 use super::*;
 use super::sync_manager::{BackendSyncable, SyncManager};
+use std::ops::{Index, IndexMut};
 
 /// Public registration descriptor used for indexing and comparisons
 #[derive(Debug, Clone, PartialEq)]
@@ -108,7 +109,7 @@ impl<'a> RegDescList<'a> {
     pub fn get_type(&self) -> Result<MemType, NixlError> { Ok(self.mem_type) }
 
     /// Adds a descriptor to the list
-    pub fn add_desc(&mut self, addr: usize, len: usize, dev_id: u64) -> Result<(), NixlError> {
+    pub fn add_desc(&mut self, addr: usize, len: usize, dev_id: u64) {
         self.add_desc_with_meta(addr, len, dev_id, &[])
     }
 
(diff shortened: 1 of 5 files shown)


gleon99:
@roiedanino can manually test it fixes https://github.com/ai-dynamo/nixl/issues/972 ?

src/api/python/__init__.py
```diff
@@ -12,3 +12,23 @@
...
+    nixl_agent,
+    nixl_agent_config,
+    nixl_backend_handle,
+    nixl_prepped_dlist_handle,
+    nixl_xfer_handle,
+)
+
+__all__ = [
+    # Main classes
+    "nixl_agent",
```
gleon99, line 29: Minor: consistent order w above

roiedanino:
> @roiedanino can manually test it fixes #972 ?

Checked and it worked:

```python3

import logging

logging.basicConfig(
    level=logging.INFO,  # Set the minimum severity level to log
    format='[MY LOGGER FORMAT] %(asctime)s - %(levelname)s - %(message)s', # Define the log message format

)

logger = logging.getLogger(__name__)

DEVICE='SOME DEVICE'

def log_test_result(a, b ,c):
    print('[MY PRINT]', a, b, c)

def test_nixl_api():
    """Test that nixl API can be imported."""
    test_name = "test_nixl_api"
    try:
        from nixl_cu12._api import nixl_agent, nixl_agent_config
        logger.info("nixl._api imported successfully")
        log_test_result(test_name, 'passed', f"NIXL API imported successfully on {DEVICE}")
        assert True
    except Exception as e:
        log_test_result(test_name, 'failed', f"NIXL API not available: {e}")
        print(f"NIXL API not available: {e}")

test_nixl_api()
```

Output:
```bash
[MY LOGGER FORMAT] 2025-11-27 15:16:46,513 - INFO - nixl._api imported successfully
[MY PRINT] test_nixl_api passed NIXL API imported successfully on SOME DEVICE
```

ovidiusm:
This works for _api but note that there is also `src/bindings/python/nixl-meta/nixl/__init__.py` which becomes the `nixl` module. It imports symbols from "_api", "_bindings", "_utils", "logging"

It allows users to write `from nixl import ...` or simply `import nixl` which gets translated to an import from the nixl_cuXY module.

The right way to import from nixl is via `nixl`

However we want to support existing users that already have `_api` imports written before we introduced the `nixl` module.

Do we want to use the same approach?

PR #1069, opened 2025-12-02
CI: Reduce GPU test to CUDA 13 and UCX master

Alexey-Rivkin:
## What?
Limit the GPU test matrix to CUDA 13 with UCX master.

## Why?
Hardware constraints combined with the NIXL tests’ fixed sleep timings create bottlenecks that significantly extend CI running time.

### DIFF
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -31,12 +31,11 @@ matrix:
   axes:
     image:
       - nvcr.io/nvidia/cuda-dl-base:25.10-cuda13.0-devel-ubuntu24.04
-      # - nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04
     arch:
       - x86_64
     ucx_version:
       - master
-      # - v1.20.x
+      - v1.20.x
 
 taskName: "${name}/${arch}/ucx-${ucx_version}/${axis_index}"
 


.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -31,12 +31,12 @@ matrix:
   axes:
     image:
       - nvcr.io/nvidia/cuda-dl-base:25.10-cuda13.0-devel-ubuntu24.04
-      - nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04
+      # - nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04
```
yosefe: pls remove this (don't comment out)
we can remove cuda12.9

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -31,12 +31,12 @@ matrix:
...
     arch:
       - x86_64
     ucx_version:
       - master
-      - v1.20.x
+      # - v1.20.x
```
yosefe: pls restore testing with ucx 1.20.x

.ci/jenkins/lib/test-matrix.yaml
Alexey-Rivkin: f866583a

.ci/jenkins/lib/test-matrix.yaml
Alexey-Rivkin: f866583a

PR #1071, opened 2025-12-02
Add support for Python 3.14; remove 3.9

ovidiusm:
## What?
Add support for Python 3.14 (GIL enabled).
Remove support for Python 3.9.
Upgrades license format to https://peps.python.org/pep-0639/

## Why?
PyTorch [v2.9.0](https://github.com/pytorch/pytorch/releases/tag/v2.9.0) added support for Python 3.14 and removed support for Python 3.9.

Setuptools deprecated the old license format and it will soon not possible to upload wheels to PyPi anymore unless we migrate to the new format.


pyproject.toml
```diff
@@ -14,16 +14,17 @@
 # limitations under the License.
 
 [build-system]
-requires = ["meson-python", "pybind11", "patchelf", "pyyaml", "types-PyYAML", "pytest", "build", "setuptools"]
+requires = ["meson-python", "pybind11", "patchelf", "pyyaml", "types-PyYAML", "pytest", "build", "setuptools>=80.9.0"]
 build-backend = "mesonpy"
 
 [project]
-name = 'nixl-cu12'
+name = "nixl-cu12"
```
roiedanino, line 21: Either all strings use double quotes or all strings should use single quotes
Anyhow, it should be fixed in a separate PR, so maybe revert this change

ovidiusm: I agree, switched to double quotes everywhere

PR #1081, opened 2025-12-04
CI: Increase test sleep

Alexey-Rivkin:
## What?
Increase hardcoded sleep durations across CI scripts, unit tests, and examples.

## Why?
Prevent random test failures during high-concurrency runs where system performance is slower, causing fixed waits to expire prematurely.

## How?
Scaled up sleep intervals (typically by 3x–5x, with larger increases for tight polling loops) in:
- GitLab CI shell scripts (test_python.sh, test_cpp.sh, etc.)
- C++ GTests (metadata_exchange, telemetry_test, device_api/utils.cu, etc.)
- Plugin unit tests (nixl_hf3fs_test.cpp, nixl_hf3fs_mt_test.cpp)
- Benchmark and example code (etcd_rt.cpp, nixl_p2p_storage_example.py)

### DIFF
--- a/test/gtest/metadata_exchange.cpp
+++ b/test/gtest/metadata_exchange.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");


Alexey-Rivkin:
> The safest and most reliable way to start an etcd member and wait until it is ready to serve is by repeatedly polling its health check endpoint or using the etcdctl command until it returns a healthy status.
> 
> This approach replaces an unreliable sleep command with a precise check for the service's operational readiness. use AI to generate proper check

10edcd9e

.gitlab/test_python.sh
```diff
@@ -67,7 +67,7 @@ export NIXL_ETCD_NAMESPACE="/nixl/python_ci/${etcd_port}"
 etcd --listen-client-urls ${NIXL_ETCD_ENDPOINTS} --advertise-client-urls ${NIXL_ETCD_ENDPOINTS} \
      --listen-peer-urls ${NIXL_ETCD_PEER_URLS} --initial-advertise-peer-urls ${NIXL_ETCD_PEER_URLS} \
      --initial-cluster default=${NIXL_ETCD_PEER_URLS} &
-sleep 15
+sleep 5
```
brminich: 👍

test/gtest/telemetry_test.cpp
```diff
@@ -155,7 +155,7 @@ TEST_F(telemetryTest, TransferBytesTracking) {
     EXPECT_NO_THROW(telemetry.updateMemoryDeregistered(1024));
     EXPECT_NO_THROW(telemetry.addXferTime(std::chrono::microseconds(100), true, 2000));
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(5));
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
brminich, line 155: better to define some constants instead of just numbers

brminich (dismissed):

ovidiusm (dismissed):

.ci/scripts/common.sh
```diff
@@ -116,3 +116,17 @@ if [ -z "$NPROC" ]; then
     fi
     export NPROC=$nproc
 fi
+
+wait_for_etcd() {
```
dpressle, line 120: this test should be port aware as we will have multiple etcd running on same host

dpressle (dismissed):

test/gtest/telemetry_test.cpp
dpressle: I dont think its in the scope of this PR, we only update sleep values.

dpressle:
Tested successfully with 8 executers see builds 180-183

brminich (dismissed):

PR #1084, opened 2025-12-05
feat(libfabric): disable unsolicited write recv for EFA RDM to reduce…

akkart-aws:
__Disable unsolicited write recv for EFA RDM to reduce CQ overflow__

Add support for FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV option to disable unsolicited write receive operations on EFA RDM endpoints. This helps reduce the likelihood of completion queue overflow under high load.

Changes:

- Add fi_setopt() call to disable unsolicited write recv before endpoint enable in libfabric_rail.cpp

### What?

Adds `fi_setopt()` call to disable unsolicited write receive on EFA RDM endpoints using the `FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV` option with backward compatibility for older libfabric versions.

### Why?

Unsolicited write receive increases completion queue pressure, which can cause CQ overflow under high load with GPU-to-GPU transfers. Disabling this feature improves stability at the cost of some performance.

### How?

Calls `fi_setopt()` with `FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV=false` in the rail constructor before `fi_enable()`. The code is wrapped in `#ifdef HAVE_FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV` to ensure backward compatibility with older libfabric versions.

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -533,6 +533,27 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
             throw std::runtime_error("fi_ep_bind av failed for rail " + std::to_string(rail_id));
         }
 
+#ifdef HAVE_FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV
+        if (provider_name == "efa") {
+            // Disable unsolicited write recv for EFA RDM to reduce CQ overflow likelihood
+            const bool use_unsolicited_write_recv = false; // Set to false to disable the feature
+            ret = fi_setopt(&endpoint->fid,
+                            FI_OPT_ENDPOINT,
+                            FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV,
+                            &use_unsolicited_write_recv,
+                            sizeof(use_unsolicited_write_recv));
+            if (ret && ret != -FI_ENOSYS && ret != -FI_ENOPROTOOPT) {
+                NIXL_WARN << "fi_setopt FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV failed for rail "
+                          << rail_id << ": " << fi_strerror(-ret) << " - continuing anyway";
+            } else if (ret == 0) {
+                NIXL_INFO << "Successfully disabled unsolicited write recv for rail " << rail_id;
+            } else if (ret == -FI_ENOSYS || ret == -FI_ENOPROTOOPT) {
+                NIXL_DEBUG << "FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV not supported for rail "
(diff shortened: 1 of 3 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -533,6 +533,25 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
             throw std::runtime_error("fi_ep_bind av failed for rail " + std::to_string(rail_id));
         }
 
+#ifdef FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV
+        // Disable unsolicited write recv for EFA RDM to reduce CQ overflow likelihood
+        bool use_unsolicited_write_recv = false; // Set to false to disable the feature
```
fengjica: const?

akkart-aws: Done

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -533,6 +533,27 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
             throw std::runtime_error("fi_ep_bind av failed for rail " + std::to_string(rail_id));
         }
 
+#ifdef HAVE_FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV
```
amitrad-aws, line 536: I don't think we need this ifdef.
If it is an older version, it will just return unsupported opt

akkart-aws: The issue is that FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV is an enum value [libfabric PR #11675](https://github.com/ai-dynamo/nixl/pull/1084), not a preprocessor macro. In libfabric versions < 2.3.x, this enum value doesn't exist in the headers at all. When we try to compile against older libfabric, we get a compile-time error: 'FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV' was not declared in this scope. We never reach the runtime check where fi_setopt() would return an error code. The preprocessor guard is necessary to conditionally compile this code only when building against libfabric >= 2.3.x."

bwbarrett: Then you need a build system configuration-time check that sets the macro based on "is that enum value defined", and not version numbers.

bwbarrett (changes requested):
Please squash your commits

src/utils/libfabric/libfabric_rail.cpp
akkart-aws: Thanks for the confirmation. I was looking into the aws-ofi-nccl code and it uses autoconf macro AC_CHECK_DECLS . I will handle it similarly for meson.

PR #1085, opened 2025-12-07
UTILS/COMMON: log NIXL version when loaded

roiedanino:
## What?
Logging NIXL version in log level INFO, when nixl is loaded

## Why?
Currently, there are no user-friendly ways to figure out which NIXL version is loaded when not using Python/Rust bindings.

## Output example:
```bash
root@lego02:/rdanino/nixl/install/bin# NIXL_LOG_LEVEL=info ./nixl_example                                                                                                                                                                                                                  
I1207 15:00:46.246258    7271 nixl_log.cpp:79] NIXL version: 0.8.0
I1207 15:00:46.327740    7271 nixl_plugin_manager.cpp:461] Discovered and loaded backend plugin: UCX                                         
I1207 15:00:46.333553    7271 nixl_plugin_manager.cpp:461] Discovered and loaded backend plugin: POSIX                                       
I1207 15:00:46.343864    7271 nixl_plugin_manager.cpp:246] Failed to load plugin from /rdanino/nixl/install/lib/aarch64-linux-gnu/plugins/libplugin_OBJ.so: libaws-cpp-sdk-s3.so: cannot open shared object file: No such file or directory                                                
I1207 15:00:46.343882    7271 nixl_plugin_manager.cpp:399] Failed to load plugin 'OBJ' from any directory                                    
```

### DIFF
--- a/src/utils/common/nixl_log.cpp
+++ b/src/utils/common/nixl_log.cpp
@@ -80,7 +80,7 @@ void InitializeNixlLogging()
 #ifdef NIXL_GIT_HASH
               << " (git: " << NIXL_GIT_HASH << ")"
 #endif
-	;
+        ;
 #endif
 
     if (invalid_env_var) {


src/utils/common/nixl_log.cpp
```diff
@@ -75,6 +75,10 @@ void InitializeNixlLogging()
     absl::SetStderrThreshold(settings.min_severity);
     absl::InitializeLog();
 
+#ifdef NIXL_VERSION
+    NIXL_INFO << "NIXL version: " << NIXL_VERSION;
```
aranadive: Also, add the top short commit hash. If someone just used a tarball, then commit should displayed as zeros.

aranadive: Maybe this should be WARN, so its always displayed?

roiedanino: > Maybe this should be WARN, so it's always displayed?

I don't think it's a good practice to use error/warning logs for non-error-related logs

> Also, add the top short commit hash. If someone just used a tarball, then commit should displayed as zeros.

Good idea, I will try

aranadive: > I don't think it's a good practice to use error/warning logs for non-error-related logs

The problem is the version will not get displayed by default and user will have to modify log level to show this and re-run workload. Just use the std::cout << absl_string .. to output the log instead of the NIXL_LOG macro?

roiedanino: `std::cout` won't have the logger formatting or logger destination (in case its being written to a file), and user won't have to re-run workload, he can just execute `nixl_example` or some smaller workload in case he wants just to check the version, however, if debugging is needed (and that's usually the case where the version/commit is important), user will have to provide debug logs anyway

gleon99: \+ @brminich  we need some way to log unconditionally. I agree w Roie that WARN is not a good idea.
Maybe enable INFO by default?

roiedanino: We can also do something hacky like:
```cpp
setLogLevel(INFO)
NIXL_INFO << Version...;
setLogLevel(log_level) # back to the original log level
```

The issue with this approach is that the user might think that the log level was INFO in that run, and that there were no additional INFO messages

gleon99: Exactly, that's a problematic approach as well.

brminich: IMO,  we can't print unconditionally, only real issues should be printed by default

ovidiusm (dismissed):

brminich:
note clang format failure

PR #1086, opened 2025-12-07
meson.build: POSIX plugin deps fixed

anton-nayshtut:
## What?
The patch fixes the POSIX plugin dependencies.

## Why?
Without this fix, NIXL fails to compile if `libaio` is missing, while some customers like TRTLLM use NIXL for `kvcache` transfer and aren't even using the POSIX plugin at all.

## How?
* File APIs discovery moved to the main `meson.build`.
* Each API now has its own `has_<API>` boolean variable set to true if the corresponding API is available on the platform and false otherwise.
* A `has_posix_plugin` boolean variable introduced and set to true if at least one of the file APIs is available on the platform.
* The POSIX plugin's `meson.build` is now solely responsible for the logic related to the compilation of the relevant API backends: source files, preprocessor defines, dependencies, etc.

### DIFF
--- a/src/plugins/posix/posix_backend.cpp
+++ b/src/plugins/posix/posix_backend.cpp
@@ -111,6 +111,10 @@ namespace {
             if (custom_params->count("use_posix_aio") > 0) {
                 const auto &value = custom_params->at("use_posix_aio");
                 if (value == "true" || value == "1") {
+                    if (!QueueFactory::isPosixAioAvailable()) {
+                        NIXL_ERROR << "posix_aio backend requested but not available at runtime";
+                        return queue_t::UNSUPPORTED;
+                    }
                     return queue_t::POSIXAIO;
                 }
             }
@@ -122,7 +126,13 @@ namespace {
         if (QueueFactory::isUringAvailable()) {
             return queue_t::URING;
         }
-        return queue_t::POSIXAIO;
+        if (QueueFactory::isPosixAioAvailable()) {
+            return queue_t::POSIXAIO;
+        }
+
+        // Should never reach here. At least one of the queues should be available.
+        NIXL_ASSERT(false);
+        return queue_t::UNSUPPORTED;
     }
 }
 
--- a/src/plugins/posix/queue_factory_impl.cpp
+++ b/src/plugins/posix/queue_factory_impl.cpp
@@ -19,7 +19,10 @@
 #include "queue_factory_impl.h"
 #include "posix_queue.h"
 #include "posix_backend.h"
+
+#ifdef HAVE_LIBAIO
 #include "aio_queue.h"
+#endif
 
 #ifdef HAVE_LIBURING
 #include "uring_queue.h"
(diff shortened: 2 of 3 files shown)


src/plugins/posix/meson.build
```diff
@@ -13,52 +13,51 @@
...
+plugin_link_args = []
+
+if has_linux_aio
+    compile_defs += ['-DHAVE_LINUXAIO']
+    posix_sources += ['linux_aio_queue.cpp']
+    plugin_deps += [ linux_aio_dep ]
+    message('Linux AIO found')
+    has_api = true
 else
-    subdir_done()
```
brminich, line 41: is it safe to remove `subdir_done`?
maybe
```
if not has_posix_plugin
    subdir_done()
endif

Comment (anton-nayshtut) [reply src/plugins/posix/meson.build:41]: I just moved it to the beginning of the file. See line 16:
```
if not has_posix_plugin
    message('POSIX plugin will be disabled: no supported file APIs found')
    subdir_done()
endif
```

**Comment by anton-nayshtut** [comment]
> would it be an issue that `queue_factory_impl.cpp` unconditionally includes `aio_queue.h`?
> 
> also `createPosixAioQueue` uses `aioQueue`

It's not really an issue as we always compile on desktop and server distros which use GLIBC, so POSIX AIO is always available. However, I agree it looks ugly, so I added a separate commit that re-arranges and unifies the queue creation. 

Thanks!

File: src/plugins/posix/posix_backend.cpp [hunk 2]
```diff
@@ -122,7 +126,12 @@ namespace {
         if (QueueFactory::isUringAvailable()) {
             return queue_t::URING;
         }
-        return queue_t::POSIXAIO;
+        if (QueueFactory::isPosixAioAvailable()) {
+            return queue_t::POSIXAIO;
+        }
+
+        assert(false); // Should never reach here. At least one of the queues should be available.
```
brminich: minor: NIXL_ASSERT ?

brminich:
pls check clang format failures

src/plugins/posix/posix_backend.cpp
anton-nayshtut: Done! Thanks!

src/plugins/posix/meson.build
```diff
@@ -13,52 +13,51 @@
...
+    message('Linux AIO not found')
```
ovidiusm: This should be a warning, since we know that perf is lower with the other ones

anton-nayshtut: I'm not sure we can say that. It depends on specific xfer characteristics (block size, batch size, etc.), but now that we fixed the [AIO re-submission bug](https://github.com/ai-dynamo/nixl/pull/1020), overall URING seems to be showing similar numbers. What is true is that POSIX AIO significantly underperforms both.

Please advise.

anton-nayshtut: As we agreed offline, the `message` has been changed to `warning` for both Linux AIO and URING.

brminich (approved):

PR #1088, opened 2025-12-08
Cyclic buffer telemetry exporter moved to use plug-in infrastructure

alexanderbilk:
Introducing shared memory cyclic buffer telemetry exporter to work as static built-in plugin.
This telemetry exporter is a default choice in case no other exporter is specified

### DIFF
--- a/src/core/telemetry/buffer_exporter.cpp
+++ b/src/core/telemetry/buffer_exporter.cpp
@@ -17,16 +17,11 @@
 #include "buffer_exporter.h"
 #include "common/nixl_log.h"
 
-constexpr const char telemetryDirVar[] = "NIXL_TELEMETRY_DIR";
-
 namespace {
 static std::filesystem::path
 getFilePath(const nixlTelemetryExporterInitParams &init_params) {
+    // if we reach here, we ensured env var is set
     auto telemetry_dir = std::getenv(telemetryDirVar);
-    if (!telemetry_dir) {
-        throw std::invalid_argument(std::string(telemetryDirVar) + " is not set");
-    }
-
     return std::filesystem::path(telemetry_dir) / init_params.agentName.data();
 }
 } // namespace
--- a/src/core/telemetry/buffer_exporter.h
+++ b/src/core/telemetry/buffer_exporter.h
@@ -24,6 +24,8 @@
 
 #include <filesystem>
 
+constexpr const char telemetryDirVar[] = "NIXL_TELEMETRY_DIR";
+
 /**
  * @class nixlTelemetryBufferExporter
  * @brief Shared memory buffer based telemetry exporter implementation
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -28,6 +28,7 @@
 #include "telemetry_event.h"
 #include "util.h"
 #include "plugin_manager.h"
+#include "buffer_exporter.h"
 
 using namespace std::chrono_literals;
 namespace fs = std::filesystem;
@@ -81,6 +82,11 @@ nixlTelemetry::initializeTelemetry() {
         NIXL_INFO << "No telemetry exporter was specified, using default: "
                   << defaultTelemetryPlugin;
         exporter_name = defaultTelemetryPlugin;
+        if (!std::getenv(telemetryDirVar)) {
+            NIXL_DEBUG << telemetryDirVar
+                       << " is not set, NIXL telemetry is enabled without any exporter";
+            return;
+        }
     }
     auto &plugin_manager = nixlPluginManager::getInstance();
     std::shared_ptr<const nixlTelemetryPluginHandle> plugin_handle =


src/core/telemetry/buffer_exporter.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+            NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
+                      << " with size: " << getMaxEventsBuffered();
+        } else {
+            throw std::invalid_argument("TELEMETRY_DIR_VAR is not set");
+        }
+    }
+    catch (const std::exception &e) {
+        NIXL_ERROR << "Failed to create telemetry buffer exporter: " << e.what();
+        throw e;
```
rakhmets: There is no need for `try/throw/catch` here, as this exception is being handled by `nixlTelemetryPluginCreator::createExporter`. You can just throw an exception here.

src/core/telemetry/buffer_exporter.cpp
```diff
@@ -0,0 +1,58 @@
...
+        auto telemetry_dir = std::getenv(telemetryDirVar);
+        if (telemetry_dir != nullptr) {
+            auto full_file_path = std::string(telemetry_dir) + "/" + init_params.agentName.data();
+            buffer_ = std::make_unique<sharedRingBuffer<nixlTelemetryEvent>>(
+                full_file_path, true, TELEMETRY_VERSION, getMaxEventsBuffered());
+
+            NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
+                      << " with size: " << getMaxEventsBuffered();
+        } else {
+            throw std::invalid_argument("TELEMETRY_DIR_VAR is not set");
```
rakhmets: Please use `telemetryDirVar`.

src/core/telemetry/buffer_exporter.cpp
```diff
@@ -0,0 +1,58 @@
...
+ */
+#include "buffer_exporter.h"
+#include "common/nixl_log.h"
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <iomanip>
+#include <thread>
+#include <chrono>
```
rakhmets: It seems that these headers are redundant. Please remove.

src/core/telemetry/buffer_exporter.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef _TELEMETRY_BUFFER_EXPORTER_H
+#define _TELEMETRY_BUFFER_EXPORTER_H
+
+#include "common/cyclic_buffer.h"
+#include "telemetry/telemetry_exporter.h"
+#include "telemetry_event.h"
+#include "nixl_types.h"
+
+#include <string>
+#include <fstream>
```
rakhmets: Please remove.

src/core/telemetry/buffer_plugin.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include "buffer_exporter.h"
+#include "telemetry/telemetry_plugin.h"
+#include "telemetry/telemetry_exporter.h"
+
+// Plugin type alias for convenience
+using buffer_exporter_plugin_t = nixlTelemetryPluginCreator<nixlTelemetryBufferExporter>;
+
+nixlTelemetryPlugin *
+createStaticBUFFERPlugin() {
```
rakhmets, line 26:
```suggestion
createStaticBufferPlugin() {
```

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +72,35 @@ nixlTelemetry::initializeTelemetry() {
...
+        .agentName = agentName_,
+        .maxEventsBuffered = buffer_size,
+    };
+
+    try {
+        exporter_ = plugin_handle->createExporter(init_params);
+    }
+    catch (const std::exception &e) {
+        throw e;
+    }
```
rakhmets:
```suggestion
    exporter_ = plugin_handle->createExporter(init_params);
```

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +72,35 @@ nixlTelemetry::initializeTelemetry() {
...
+            throw std::runtime_error("Failed to load default telemetry plugin");
+        }
+    }
 
-    buffer_ = std::make_unique<sharedRingBuffer<nixlTelemetryEvent>>(
-        full_file_path, true, TELEMETRY_VERSION, buffer_size);
+    struct nixlTelemetryExporterInitParams init_params = {
+        .agentName = agentName_,
+        .maxEventsBuffered = buffer_size,
+    };
```
rakhmets:
```suggestion
    const nixlTelemetryExporterInitParams init_params{agentName_, buffer_size};
```

src/core/telemetry/buffer_exporter.cpp
brminich: or maybe `NIXL_TELEMETRY_DIR`?

src/core/telemetry/telemetry.cpp
```diff
@@ -27,20 +27,23 @@
...
+#include "plugin_manager.h"
 
 using namespace std::chrono_literals;
 namespace fs = std::filesystem;
 
 constexpr std::chrono::milliseconds DEFAULT_TELEMETRY_RUN_INTERVAL = 100ms;
 constexpr size_t DEFAULT_TELEMETRY_BUFFER_SIZE = 4096;
 
-nixlTelemetry::nixlTelemetry(const std::string &file_path, backend_map_t &backend_map)
+// constexpr std::string_view defaultTelemetryExporter = "buffer";
```
brminich: remove?

src/core/telemetry/buffer_exporter.cpp
alexanderbilk: removed

src/core/telemetry/buffer_exporter.cpp
alexanderbilk: Changed

src/core/telemetry/buffer_exporter.cpp
alexanderbilk: Removed

src/core/telemetry/buffer_exporter.h
alexanderbilk: Removed

src/core/telemetry/telemetry.cpp
alexanderbilk: Applied

src/core/telemetry/telemetry.cpp
alexanderbilk: Applied

src/core/telemetry/telemetry.cpp
alexanderbilk: Removed

src/core/telemetry/buffer_plugin.cpp
alexanderbilk: I've made it upper case to match style of static backend plugin creators. Can we keep it as is?

src/core/nixl_agent.cpp
```diff
@@ -126,19 +125,12 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
...
-    const char *telemetry_env_dir = std::getenv(TELEMETRY_DIR_VAR);
 
     if (telemetry_env_val != nullptr) {
         if (!strcasecmp(telemetry_env_val, "y") || !strcasecmp(telemetry_env_val, "1") ||
             !strcasecmp(telemetry_env_val, "yes") || !strcasecmp(telemetry_env_val, "on")) {
             telemetryEnabled = true;
-            if (telemetry_env_dir != nullptr) {
-                std::string telemetry_file = std::string(telemetry_env_dir) + "/" + name;
-                telemetry_ = std::make_unique<nixlTelemetry>(telemetry_file, backendEngines);
-                NIXL_DEBUG << "NIXL telemetry is enabled with output file: " << telemetry_file;
```
ovidiusm, line 138: Minor comment: I see that these debugs are removed, some below remain, and there is a new info for the plugin constructor

Would it make sense to be consistent and also have a debug here?

src/core/telemetry/buffer_plugin.cpp
rakhmets: ok.

src/api/cpp/telemetry/telemetry_plugin.h
```diff
@@ -29,6 +29,8 @@ enum class nixl_telemetry_plugin_api_version : unsigned int { V1 = 1 };
 inline constexpr nixl_telemetry_plugin_api_version nixlTelemetryPluginApiVersionV1 =
     nixl_telemetry_plugin_api_version::V1;
 
+constexpr std::string_view defaultTelemetryPlugin = "BUFFER";
```
rakhmets: Can be moved to `telemetry.cpp`.

rakhmets:
Please add a description to the PR, mentioning that the telemetry exporter based on shared memory buffer was added as a static plugin, and is also used as the default telemetry plugin.

src/core/telemetry/telemetry.cpp
```diff
@@ -27,20 +27,21 @@
...
+nixlTelemetry::nixlTelemetry(const std::string &agent_name, backend_map_t &backend_map)
     : pool_(1),
       writeTask_(pool_.get_executor(), DEFAULT_TELEMETRY_RUN_INTERVAL, false),
-      file_(file_path),
+      agentName_(agent_name),
       backendMap_(backend_map) {
-    if (file_path.empty()) {
-        throw std::invalid_argument("Telemetry file path cannot be empty");
+    if (agent_name.empty()) {
+        throw std::invalid_argument("Agent name cannot be empty");
```
ColinNV: Perhaps clarify who is throwing, like "Expected non-empty agent name in nixl telemetry create".

src/core/telemetry/buffer_exporter.cpp
```diff
@@ -0,0 +1,45 @@
...
+    : nixlTelemetryExporter(init_params) {
+    auto telemetry_dir = std::getenv(telemetryDirVar);
+    if (telemetry_dir != nullptr) {
+        auto full_file_path = std::string(telemetry_dir) + "/" + init_params.agentName.data();
+        buffer_ = std::make_unique<sharedRingBuffer<nixlTelemetryEvent>>(
+            full_file_path, true, TELEMETRY_VERSION, getMaxEventsBuffered());
+
+        NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
+                  << " with size: " << getMaxEventsBuffered();
+    } else {
```
ColinNV: I would suggest inverting the logic of the `if`, which puts the `throw` in the line following it, and makes the `else` keyword superfluous.

src/core/telemetry/buffer_exporter.h
```diff
@@ -0,0 +1,47 @@
...
+     * @brief Constructor using init params (plugin-compatible)
+     * @param init_params Initialization parameters
+     */
+    explicit nixlTelemetryBufferExporter(const nixlTelemetryExporterInitParams &init_params);
+
+    nixl_status_t
+    exportEvent(const nixlTelemetryEvent &event) override;
+
+private:
+    std::unique_ptr<sharedRingBuffer<nixlTelemetryEvent>> buffer_;
```
ColinNV: If calculating `full_file_path` is moved to a utility function (in an anonymous namespace) then `buffer_` can directly be of type `sharedRingBuffer<nixlTelemetryEvent>`, no need for a pointer. This could also be done later.

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +70,26 @@ nixlTelemetry::initializeTelemetry() {
...
         throw std::invalid_argument("Telemetry buffer size cannot be 0");
     }
 
-    NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
-              << " with size: " << buffer_size;
+    auto exporter_name =
+        std::getenv(telemetryExporterVar) ? std::string(std::getenv(telemetryExporterVar)) : "";
+
+    auto &plugin_manager = nixlPluginManager::getInstance();
+    auto plugin_handle = plugin_manager.loadTelemetryPlugin(exporter_name.data());
```
ColinNV: Sometimes writing out the types makes code more readable.

src/api/cpp/telemetry/telemetry_plugin.h
alexanderbilk: Moved

src/core/telemetry/telemetry.cpp
alexanderbilk: Applied

src/core/telemetry/buffer_exporter.cpp
alexanderbilk: Applied

src/core/nixl_agent.cpp
alexanderbilk: Added corresponding debug output to initializeTelemetry()

src/core/telemetry/telemetry.cpp
alexanderbilk: Explicitly set type for plugin_handle

src/core/telemetry/buffer_exporter.h
alexanderbilk: Not sure I got you here. Why we don't want this buffer to be a unique pointer?

ColinNV: If you _can_ directly keep it as member by value then keeping it indirectly by pointer is just unnecessary overhead.

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,31 @@ nixlTelemetry::initializeTelemetry() {
...
-              << " with size: " << buffer_size;
+    auto exporter_name =
+        std::getenv(telemetryExporterVar) ? std::string(std::getenv(telemetryExporterVar)) : "";
+
+    auto &plugin_manager = nixlPluginManager::getInstance();
+    std::shared_ptr<const nixlTelemetryPluginHandle> plugin_handle =
+        plugin_manager.loadTelemetryPlugin(exporter_name.data());
+
+    if (plugin_handle == nullptr) {
+        NIXL_WARN << "Failed to load telemetry plugin: " << exporter_name
```
ovidiusm: If there is a telemetry exporter configured and it is failing to load, this should be an error; also, I do not think we should fall back to another exporter

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,31 @@ nixlTelemetry::initializeTelemetry() {
...
-    auto full_file_path = fs::path(file_);
-
     if (buffer_size == 0) {
         throw std::invalid_argument("Telemetry buffer size cannot be 0");
     }
 
-    NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
-              << " with size: " << buffer_size;
+    auto exporter_name =
+        std::getenv(telemetryExporterVar) ? std::string(std::getenv(telemetryExporterVar)) : "";
```
ovidiusm: Could we avoid double getenv?

src/core/telemetry/telemetry.cpp
ovidiusm: What I mean is: if exporter_name is "": load default; otherwise: load configured
Either way, emit error log when load failed

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,31 @@ nixlTelemetry::initializeTelemetry() {
...
+    std::shared_ptr<const nixlTelemetryPluginHandle> plugin_handle =
+        plugin_manager.loadTelemetryPlugin(exporter_name.data());
+
+    if (plugin_handle == nullptr) {
+        NIXL_WARN << "Failed to load telemetry plugin: " << exporter_name
+                  << ", using default plugin: " << defaultTelemetryPlugin;
+        exporter_name = defaultTelemetryPlugin.data();
+        plugin_handle = plugin_manager.loadTelemetryPlugin(exporter_name.data());
+        if (plugin_handle == nullptr) {
+            throw std::runtime_error("Failed to load default telemetry plugin");
```
ovidiusm: To me it seems unexpected that we throw an exception only in this case; if we want to throw an exception, we should throw it when the configured exporter fails to load, not just the default

src/core/telemetry/telemetry.cpp
alexanderbilk: Changed

src/core/telemetry/telemetry.cpp
alexanderbilk: Removed

src/core/telemetry/telemetry.cpp
alexanderbilk: No fall back routine anymore, throwing exception on plugin load failure

ovidiusm (dismissed):

src/core/telemetry/buffer_exporter.h
alexanderbilk: Applied

brminich (approved):

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,34 @@ nixlTelemetry::initializeTelemetry() {
...
+        NIXL_INFO << "No telemetry exporter was specified, using default: "
+                  << defaultTelemetryPlugin;
+        exporter_name = defaultTelemetryPlugin;
+    }
+    auto &plugin_manager = nixlPluginManager::getInstance();
+    std::shared_ptr<const nixlTelemetryPluginHandle> plugin_handle =
+        plugin_manager.loadTelemetryPlugin(exporter_name);
+
+    if (plugin_handle == nullptr) {
+        NIXL_ERROR << "Failed to load telemetry plugin: " << exporter_name;
```
rakhmets:
```suggestion
```
There is no need to print an error message here. It should be printed using exception handling code.

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,34 @@ nixlTelemetry::initializeTelemetry() {
...
+        throw std::runtime_error("Failed to load telemetry plugin: " + std::string(exporter_name));
+    }
+
+    const nixlTelemetryExporterInitParams init_params{agentName_, buffer_size};
+    exporter_ = plugin_handle->createExporter(init_params);
+    if (!exporter_) {
+        NIXL_ERROR << "Failed to create telemetry exporter: " << exporter_name;
+        return;
```
rakhmets, line 103:
```suggestion
        throw std::runtime_error("Failed to create telemetry exporter: " + std::string(exporter_name));
```

rakhmets (dismissed):

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +71,34 @@ nixlTelemetry::initializeTelemetry() {
...
     }
 
-    NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
-              << " with size: " << buffer_size;
+    const char *exporter_name = std::getenv(telemetryExporterVar);
+
+    if (!exporter_name) {
+        NIXL_INFO << "No telemetry exporter was specified, using default: "
+                  << defaultTelemetryPlugin;
+        exporter_name = defaultTelemetryPlugin;
```
ovidiusm, line 84: If NIXL_TELEMETRY_DIR is not set, we should skip attempting to create the default plugin. This is to maintain backwards compatibility with the previous behavior.

By the way, we need to document the new behavior in docs/telemetry.md

brminich: probably we can fix the rest of comments in the next PR for prometheus

src/core/telemetry/telemetry.cpp
brminich: probably we can fix the rest of comments in the next PR for prometheus

src/core/telemetry/telemetry.cpp
alexanderbilk: This code checks exporter name which is something new and an old API is not aware of. So it is expected to for NIXL_TELEMETRY_EXPORTER not to be set. Check for NIX_TELEMETRY_DIR is inside buffer_exporter implementation. And if it is not set as well, we won't create an exporter

alexanderbilk: I've intended to add documentation to prometheus PR encompassing  all telemetry changes

src/core/telemetry/telemetry.cpp
alexanderbilk: @ovidiusm should I push this fix now? before next CI run?

src/core/telemetry/telemetry.cpp
ovidiusm: But why create the buffer exporter if we know we will fail and throw an exception?

src/core/telemetry/telemetry.cpp
alexanderbilk: There is no handler for this exception, besides we want to keep existing functionality which allows telemetry to be enabled without actual exporter. i.e NIXL_TELEMETRY_DIR is not set

src/core/telemetry/telemetry.cpp
alexanderbilk: How do we know it at this point of execution? 
The idea is what NIXL_TELEMETRY_DIR  is something belonging to buffer exporter only. Why to bring it out to a broader scope?

ovidiusm: I do not like the idea of throwing exceptions to handle this case. It's not an exceptional case, it is how the config was handled before

src/core/telemetry/telemetry.cpp
ovidiusm: Now please

src/core/telemetry/telemetry.cpp
alexanderbilk: But it is handled gracefully, and this exception equals to simple NIXL_ERROR to the final user. imho it's the best way to mitigate it given changed infra

brminich (approved):

src/core/telemetry/telemetry.cpp
ovidiusm: If you want to do it that way, that's a backwards incompatible change, and we cannot make it in this release. The old config needs to be documented as deprecated.

ovidiusm: The idea is the following.

Telemetry, until now, as per https://github.com/ai-dynamo/nixl/blob/main/docs/telemetry.md, could work in 3 ways:

1. Disabled
Nothing is recorded.

2. NIXL_TELEMETRY_ENABLE=1 NIXL_TELEMETRY_DIR unset
In this case, events can be read through the API only.

3. NIXL_TELEMETRY_ENABLE=1 NIXL_TELEMETRY_DIR=/path
In this case, events can be read through the API; and they are also exported to the directory.

The problem with the current code is:
If NIXL_TELEMETRY_ENABLE=1 NIXL_TELEMETRY_DIR unset:
NIXL_TELEMETRY_EXPORTER unset => exporter_name is NULL => exporter_name = defaultTelemetryPlugin => loadTelemetryPlugin(exporter_name); => plugin_handle->createExporter(init_params); which internally throws an exception then caught nixlTelemetryPluginCreator::createExporter which logs:

NIXL_ERROR << "Failed to create exporter: " << e.what();

returns NULL and then we log another error:

NIXL_ERROR << "Failed to create telemetry exporter: " << exporter_name;

But there should be no error logged. This will be surely flagged in verification.

alexanderbilk: Moved NIXL_TELEMETRY_DIR before attempt to load default telemetry plugin

src/core/telemetry/telemetry.cpp
alexanderbilk: fixed

src/core/telemetry/telemetry.cpp
alexanderbilk: Failure to initialize exporter should not be considered a critical error. It could occur for various reasons. e.g proemtheus exporter could not be initialized because desired port is occupied. We don't want in that case to break the entire logic

PR #1089, opened 2025-12-08
DEVICE/API/GTEST: Fix SingleWriteTest to set channel_id

michal-shalev:
## What?
Updates `SingleWriteTest` to set `channel_id` parameter in `nixlGpuPostSingleWriteXferReq`.

## Why?
Recent UCX changes (commit `1c60acc`) introduced a `channel_id` parameter. The test call site was missing this argument, causing the `is_no_delay` flag (true) to be interpreted as `channel_id` (1). This resulted in an illegal memory access on the GPU since only channel 0 is configured.

```
[==========] Running 9 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 9 tests from ucxDeviceApi/SingleWriteTest
[ RUN      ] ucxDeviceApi/SingleWriteTest.BasicSingleWriteTest/UCX_BLOCK
Failed to synchronize: an illegal memory access was encountered
Failed to launch kernel: an illegal memory access was encountered
../test/gtest/device_api/single_write_test.cu:441: Failure
  status
    Which is: -3
  NIXL_SUCCESS
    Which is: 0
Kernel launch failed with status: -3
```

## How?
Explicitly passes `0` as the `channel_id` in the `nixlGpuPostSingleWriteXferReq` call.

### DIFF
--- a/test/gtest/device_api/single_write_test.cu
+++ b/test/gtest/device_api/single_write_test.cu
@@ -48,7 +48,7 @@ TestSingleWriteKernel(nixlGpuXferReqH req_hdnl,
 
     for (size_t i = 0; i < num_iters; ++i) {
         status = nixlGpuPostSingleWriteXferReq<level>(
-            req_hdnl, index, src_offset, remote_offset, size, is_no_delay, xfer_status_ptr);
+            req_hdnl, index, src_offset, remote_offset, size, 0, is_no_delay, xfer_status_ptr);
         if (status != NIXL_IN_PROG) {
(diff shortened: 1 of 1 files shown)


michal-shalev:
/ok to test [3e25a6b](https://github.com/ai-dynamo/nixl/pull/1089/commits/3e25a6b01881b4948880577b4e8022258e3e5306)

michal-shalev:
/ok to test [0cc0d83](https://github.com/ai-dynamo/nixl/pull/1089/commits/0cc0d837c8ab7f0d3925d9fd2896b1b06e5a2403)

PR #1091, opened 2025-12-08
Prometheus telemetry exporter

alexanderbilk:
Introducing Prometheus compatible telemetry exporter


src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include "common/nixl_log.h"
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <iomanip>
+#include <thread>
+#include <chrono>
+
+inline const uint16_t prometheusExporterDefaultPort = 9090;
```
ColinNV: This is a `.cpp` file, these variables should be just `const`, not `inline`, and be placed in an anonymous namespace. (The `uint16_t` could also be `constexpr`, but it doesn't really matter here.)

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+inline const std::string prometheusExporterPublicAddress = "0.0.0.0";
+
+inline constexpr char prometheusPortVar[] = "NIXL_TELEMETRY_PROMETHEUS_PORT";
+inline constexpr char prometheusLocalVar[] = "NIXL_TELEMETRY_PROMETHEUS_LOCAL";
+
+nixlTelemetryPrometheusExporter::nixlTelemetryPrometheusExporter(
+    const nixlTelemetryExporterInitParams &init_params)
+    : nixlTelemetryExporter(init_params),
+      registry_(std::make_shared<prometheus::Registry>()) {
+    auto port_str = std::getenv(prometheusPortVar);
```
ColinNV, line 42: The logic to determine the port should be moved to a separate function, then `port_` can be initialized in the member initializer list and made `const`.

Same goes for `local_`, might need rearranging the order of the member variables.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+nixl_status_t
+nixlTelemetryPrometheusExporter::exportEvent(const nixlTelemetryEvent &event) {
+    try {
+        const std::string event_name(event.eventName_);
+
+        switch (event.category_) {
+        case nixl_telemetry_category_t::NIXL_TELEMETRY_TRANSFER: {
+            auto it = counters_.find(event_name);
```
ColinNV: `const`

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+            if (it != counters_.end()) {
+                it->second->Increment(event.value_);
+            }
+            break;
+        }
+        case nixl_telemetry_category_t::NIXL_TELEMETRY_PERFORMANCE:
+        case nixl_telemetry_category_t::NIXL_TELEMETRY_MEMORY: {
+            auto it = gauges_.find(event_name);
```
ColinNV, line 175: We ignore when it's not found?

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+                                    .Name("agent_tx_requests_num")
+                                    .Help("Number of requests sent by the agent")
+                                    .Register(*registry_);
+
+    auto &rx_requests_counter = prometheus::BuildCounter()
+                                    .Name("agent_rx_requests_num")
+                                    .Help("Number of requests received by the agent")
+                                    .Register(*registry_);
+
+    registerCounter(
```
ColinNV, line 106: It looks like the build counter/register counter, and the build gauge/register gauge, pairs of functions should be put into helper functions, that would also remove the duplication of the names.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    registerGauge("agent_xfer_post_time",
+                  xfer_post_time_gauge,
+                  {{"category", prometheusExporterPerformanceCategory}});
+    registerGauge("agent_memory_registered",
+                  memory_registered_gauge,
+                  {{"category", prometheusExporterMemoryCategory}});
+    registerGauge("agent_memory_deregistered",
+                  memory_deregistered_gauge,
+                  {{"category", prometheusExporterMemoryCategory}});
+}
```
ColinNV, line 145: We need to remember to find a better mid-term solution that at least doesn't repeat the names as strings. Ideally not every counter and gauge needs to even be known here in advance, it's error prone to have to keep in sync.

src/core/telemetry/telemetry.cpp
```diff
@@ -69,17 +70,31 @@ nixlTelemetry::initializeTelemetry() {
...
         throw std::invalid_argument("Telemetry buffer size cannot be 0");
     }
 
-    NIXL_INFO << "Telemetry enabled, using buffer path: " << full_file_path
-              << " with size: " << buffer_size;
+    auto exporter_name =
+        std::getenv(telemetryExporterVar) ? std::string(std::getenv(telemetryExporterVar)) : "";
+
+    auto &plugin_manager = nixlPluginManager::getInstance();
+    auto exporter_plugin_dir = std::getenv(telemetryExporterPluginDirVar);
```
ovidiusm: Since we agreed that we would remove this and rely on NIXL_PLUGIN_DIR. Please remember to remove.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Changed to just const

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Applied

src/core/telemetry/telemetry.cpp
alexanderbilk: Yes, removed

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Added helper functions

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Creation and registration moved to helper function

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Yes, we don't expect any unregistered events if they come not from backend engine

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Yes, that's the known issue but it is also critical for performance and prometheus DB cardinality

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+const char prometheusLocalVar[] = "NIXL_TELEMETRY_PROMETHEUS_LOCAL";
+
+const std::string prometheusExporterTransferCategory = "NIXL_TELEMETRY_TRANSFER";
+const std::string prometheusExporterPerformanceCategory = "NIXL_TELEMETRY_PERFORMANCE";
+const std::string prometheusExporterMemoryCategory = "NIXL_TELEMETRY_MEMORY";
+const std::string prometheusExporterBackendCategory = "NIXL_TELEMETRY_BACKEND";
+const std::string prometheusExporterLocalAddress = "127.0.0.1";
+const std::string prometheusExporterPublicAddress = "0.0.0.0";
+
+static uint16_t
```
rakhmets:
```suggestion
uint16_t
```

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+        return port;
+    }
+    catch (const std::exception &e) {
+        NIXL_WARN << "Invalid port " << port_str
+                  << "', expected numeric port. Using default: " << prometheusExporterDefaultPort;
+        return prometheusExporterDefaultPort;
+    }
+}
+
+static bool
```
rakhmets:
```suggestion
bool
```

docs/telemetry.md
```diff
@@ -2,15 +2,18 @@
...
 ## Architecture
 
 ### Telemetry Components
 
 1. **Telemetry Collection**: Built into the NIXL core library, collects events and metrics
-2. **Shared Memory Buffer**: Cyclic buffer implementation for efficient event storage
-3. **Telemetry Readers**: C++ and Python applications to read and display telemetry data
+2. **Shared Memory Buffer**: Static built in implemenation of telemetry exporter. Uses shared memory cyclic buffer for efficient event storage
+3. **Telemetry Readers**: C++ and Python applications to read and display telemetry data from cyclic buffer
+4. **Prometheus exporter**: EXPERIMENTAL prometheus compitable telemetry exporter, see [src/plugins/telemetry/prometheus/README.md](../src/plugins/telemetry/prometheus/README.md).
```
ovidiusm:
```suggestion
4. **Prometheus exporter**: EXPERIMENTAL prometheus compatible telemetry exporter, see [src/plugins/telemetry/prometheus/README.md](../src/plugins/telemetry/prometheus/README.md).
```

docs/telemetry.md
```diff
@@ -35,6 +38,33 @@ The telemetry system supports the following event categories:
...
+| `agent_tx_bytes` | `NIXL_TELEMETRY_TRANSFER` | No | bytes | Total bytes transmitted by the agent |
+| `agent_rx_bytes` | `NIXL_TELEMETRY_TRANSFER` | No | bytes | Total bytes received by the agent |
+| `agent_tx_requests_num` | `NIXL_TELEMETRY_TRANSFER` | No | count | Number of transmit requests sent by the agent |
+| `agent_rx_requests_num` | `NIXL_TELEMETRY_TRANSFER` | No | count | Number of receive requests processed by the agent |
+| `agent_xfer_time` | `NIXL_TELEMETRY_PERFORMANCE` | Yes | microseconds | Transfer time from start to complete (per request) |
+| `agent_xfer_post_time` | `NIXL_TELEMETRY_PERFORMANCE` | Yes | microseconds | Time from start to posting to backend (per request) |
+| Backend-specific events | `NIXL_TELEMETRY_BACKEND` | - | varies | Dynamic events generated by backend implementations |
+| Error status strings | `NIXL_TELEMETRY_ERROR` | No | count | Error occurrences by status type |
+
+**Statefull**
```
ovidiusm:
```suggestion
**Stateful**
```

src/plugins/telemetry/prometheus/meson.build
```diff
@@ -0,0 +1,59 @@
...
+if prometheus_found
+    # Prometheus Exporter Plugin
+    prometheus_exporter_plugin = shared_library(
+        'libtelemetry_exporter_prometheus',
+        'prometheus_plugin.cpp',
+        'prometheus_exporter.cpp',
+        include_directories: [nixl_inc_dirs, utils_inc_dirs],
+        dependencies: [nixl_infra, absl_log_dep, prometheus_core_dep, prometheus_pull_dep],
+        install: true,
+        install_dir: get_option('libdir') / 'nixl' / 'telemetry_exporters',
```
ovidiusm:
```suggestion
        install_dir: plugin_install_dir,
```

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+                      << prometheusExporterDefaultPort;
+            port = prometheusExporterDefaultPort;
+        }
+        return port;
+    }
+    catch (const std::exception &e) {
+        NIXL_WARN << "Invalid port " << port_str
+                  << "', expected numeric port. Using default: " << prometheusExporterDefaultPort;
+        return prometheusExporterDefaultPort;
+    }
```
rakhmets, line 59: Optional change. You can skip it if you think the current version is better.
```suggestion
    try {
        int port = std::stoi(port_str);
        if (port < 1 || port > 65535) {
            throw std::out_of_range;
        }
        return port;
    }
    catch (const std::exception &e) {
        NIXL_WARN << "Invalid port '" << port_str
                  << "', expected numeric port between 1-65535. Using default: "
                  << prometheusExporterDefaultPort;
        return prometheusExporterDefaultPort;
    }
```

src/plugins/telemetry/prometheus/README.md
```diff
@@ -0,0 +1,77 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# NIXL Prometheus Telemetry exporter plug-in
+
+This telemetry exporter plug-in exports NIXL telemetry events in Prometheus format, by_exposing an HTTP endpoint that can be scraped by Prometheus servers.
```
ovidiusm:
```suggestion

This telemetry exporter plug-in exports NIXL telemetry events in Prometheus format, by exposing an HTTP endpoint that can be scraped by Prometheus servers.
```

src/plugins/telemetry/prometheus/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Dependencies
+
+The Prometheus exporter requires the prometheus-cpp library, which is included as a subproject:
```
ovidiusm:
```suggestion
The Prometheus exporter requires the prometheus-cpp library, which is included as a subproject.
```

src/plugins/telemetry/prometheus/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+libcurl is not handled by prometheus-cpp libaray. need to install the libcurl package:
```
ovidiusm:
```suggestion
libcurl is not downloaded automatically. To build, you need to install the libcurl package:
```

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+            bind_address_ = prometheusExporterPublicAddress + ":" + std::to_string(port_);
+        }
+
+        exposer_ = std::make_unique<prometheus::Exposer>(bind_address_);
+        exposer_->RegisterCollectable(registry_);
+
+        initializeMetrics();
+        NIXL_INFO << "Prometheus exporter initialized on " << bind_address_;
+    }
+    catch (const std::exception &e) {
```
rakhmets: try/catch is not needed here.
`nixlTelemetryPluginCreator<nixlTelemetryPrometheusExporter>::createExporter` handles exceptions.

src/plugins/telemetry/prometheus/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+## Configuration
+
+To enable the Prometheus plug-in, set the following environment variables:
+
+```bash
+export NIXL_TELEMETRY_ENABLE="y" # Enable NIXL telemetry
+export NIXL_TELEMETRY_EXPORTER="prometheus" # Sets which plug-in to select in format libtelemetry_exporter_${NIXL_TELEMETRY_EXPORTER}.so
+export NIXL_PLUGIN_DIR="path/to/dir/with/.so/files" # Sets where to find plug-in libs (NOTE: the same var is used for backend plug-ins search)
```
ovidiusm: This is optional. I suggest taking it out of the code block and mentioning that it can be used optionally.

Normally we do not want users to use it. The default path where the plugins are located is built in.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+static bool
+getLocal() {
+    auto local_str = std::getenv(prometheusLocalVar);
+    if (local_str &&
+        (!strcasecmp(local_str, "y") || !strcasecmp(local_str, "1") ||
+         !strcasecmp(local_str, "yes"))) {
+        return true;
+    }
+    return false;
```
rakhmets:
```suggestion
    return local_str &&
        (!strcasecmp(local_str, "y") || !strcasecmp(local_str, "1") ||
         !strcasecmp(local_str, "yes"));
```

src/plugins/telemetry/prometheus/README.md
```diff
@@ -0,0 +1,77 @@
...
+### Optional Configuration
+
+You can configure the exposed prometheus port:
+
+```bash
+# Default port is 9090
+export NIXL_TELEMETRY_PROMETHEUS_PORT="<port_num>"
+```
+
+Default addres is public, but you configure to expose prometheus endpoint only locally:
```
ovidiusm:
```suggestion
Default addres is public, but you configure to expose prometheus endpoint only on localhost:
```

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// static std::string
+
```
rakhmets:
```suggestion
```

rakhmets:
Please update the PR description.

src/core/telemetry/telemetry.cpp
```diff
@@ -88,6 +88,7 @@ nixlTelemetry::initializeTelemetry() {
             return;
         }
     }
+
```
brminich: better avoid this change as it is single change in the file

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: removed

src/plugins/telemetry/prometheus/meson.build
alexanderbilk: Applied

src/plugins/telemetry/prometheus/README.md
alexanderbilk: Changed

docs/telemetry.md
alexanderbilk: Addressed by AlexN in his commit

docs/telemetry.md
alexanderbilk: Addressed by AlexN in his commit

src/plugins/telemetry/prometheus/README.md
alexanderbilk: Applied

src/plugins/telemetry/prometheus/README.md
alexanderbilk: Applied

src/plugins/telemetry/prometheus/README.md
alexanderbilk: Moved to optional section

src/plugins/telemetry/prometheus/README.md
alexanderbilk: Applied

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Removed

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Removed

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Applied your version

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Removed

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
alexanderbilk: Applied

ovidiusm (approved):

ovidiusm:
Let's please try to merge it as it is, and any minor changes in a follow up

src/plugins/telemetry/prometheus/prometheus_exporter.h
```diff
@@ -0,0 +1,89 @@
...
+
+private:
+    // Prometheus components
+    const bool local_ = false;
+    const uint16_t port_;
+    const std::string agent_name_;
+    const std::string hostname_;
+    std::shared_ptr<prometheus::Registry> registry_;
+    std::unique_ptr<prometheus::Exposer> exposer_;
+    std::string bind_address_;
```
rakhmets, line 61: `local_`, `port_`, and `bind_address_` can be removed from the class interface.
A local variable (`const auto bind_address = getBindAddress();`) in the class constructor is sufficient.

PR #1095, opened 2025-12-08
Nixlbench: Fix consistency check

aranadive:
The memory should be initialized only when the check flag is used. Updated the comment as well.


akkart-aws:
Should the PR-CI always run with the check enabled, so that any backend will not break the check-consistency in the future. ? https://github.com/ai-dynamo/nixl/blob/main/.gitlab/test_nixlbench.sh#L62-L65

aranadive:
> Should the PR-CI always run with the check enabled, so that any backend will not break the check-consistency in the future. ? https://github.com/ai-dynamo/nixl/blob/main/.gitlab/test_nixlbench.sh#L62-L65

Yes but maybe with smaller options to not affect test time. Will do in a separate PR to avoid delaying the fix.

PR #1096, opened 2025-12-09
feature: rust - add metadata support to descriptor reg

cheese-head:
## What?
Enable NixlDescriptor to pass metadata through add_storage_desc() during memory registration.
- Changed NixlDescriptor::metadata() return type from &[u8] to Vec<u8>
- Updated add_storage_desc() to call desc.metadata() and pass it to add_desc_with_meta()

### DIFF
--- a/src/bindings/rust/src/descriptors/reg.rs
+++ b/src/bindings/rust/src/descriptors/reg.rs
@@ -221,9 +221,10 @@ impl<'a> RegDescList<'a> {
         let addr = unsafe { desc.as_ptr() } as usize;
         let len = desc.size();
         let dev_id = desc.device_id();
+        let metadata = desc.metadata();
 
-        // Add to list
-        self.add_desc(addr, len, dev_id);
+        // Add to list with metadata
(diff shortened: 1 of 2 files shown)


src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -221,9 +221,10 @@ impl<'a> RegDescList<'a> {
         let addr = unsafe { desc.as_ptr() } as usize;
         let len = desc.size();
         let dev_id = desc.device_id();
+        let metadata = desc.metadata();
 
-        // Add to list
-        self.add_desc(addr, len, dev_id);
+        // Add to list with metadata
+        self.add_desc_with_meta(addr, len, dev_id, &metadata);
```
roiedanino, line 227: How about adding `add_storage_desc_with_metadata`, similar to what happens in `add_desc and add_desc_with_metadata`?

Otherwise, I don't see the benefit of this change

roiedanino:
Hi @cheese-head 
Is this still relevant?

PR #1098, opened 2025-12-09
API: Add getMappedPtrs

michal-shalev:
## What?
Added `getMappedPtrs` API to expose UCX's direct access via `ucp_rkey_ptr`.

## Why?
This API wraps `ucp_rkey_ptr` to expose direct pointer access when supported by the underlying UCX lane.

## How?
- Added `nixlAgent::getMappedPtrs(nixlDlistH*, vector<void*>&)`
- Implemented `nixl::ucx::rkey::getPtr()` wrapper for `ucp_rkey_ptr`
- Returns `nullptr` for lanes without direct access capability

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1111,11 +1111,6 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     const auto engine_config =
         (engine_config_it != custom_params->end()) ? engine_config_it->second : "";
 
-#ifdef HAVE_CUDA
-    // Enable cuda_ipc for same-process transfers (needed for getMappedPtrs)
-    setenv("UCX_CUDA_IPC_ENABLE_SAME_PROCESS", "y", 0);
-#endif
-
     uc = std::make_unique<nixlUcxContext>(
         devs, init_params.enableProgTh, num_workers, init_params.syncMode, engine_config);
 
--- a/test/gtest/test_transfer.cpp
+++ b/test/gtest/test_transfer.cpp
@@ -728,6 +728,9 @@ TEST_P(TestTransfer, getMappedPtrs) {
         GTEST_SKIP() << "No CUDA-capable GPU is available, skipping test.";
     }
 
+    // Enable cuda_ipc for same-process transfers (needed for getMappedPtrs)
+    setenv("UCX_CUDA_IPC_ENABLE_SAME_PROCESS", "y", 0);
+
     std::vector<MemBuffer> buffers_local;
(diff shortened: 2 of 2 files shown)


src/utils/ucx/rkey.cpp
```diff
@@ -34,4 +34,21 @@ rkey::unpackUcpRkey(const nixlUcxEp &ep, const void *rkey_buffer) {
...
+    void *ptr = nullptr;
+    const ucs_status_t status = ucp_rkey_ptr(rkey_.get(), raddr, &ptr);
+
+    if (status == UCS_OK) {
+        return ptr;
+    }
+
+    if (status == UCS_ERR_UNREACHABLE) {
+        return nullptr;
+    }
```
rakhmets, line 49: Why don't we throw an exception in this case?
`getMappedPtrs` returns `NIXL_SUCCESS`, when every `rkey::getPtr` returns `nullptr`.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1748,6 +1753,32 @@ nixlUcxEngine::prepGpuSignal(const nixlBackendMD &meta,
...
+        const nixlMetaDesc &desc = meta_dlist[i];
+        const auto *ucx_meta = static_cast<const nixlUcxPublicMetadata *>(desc.metadataP);
+
+        try {
+            const auto &rkey_obj = ucx_meta->getRkey(worker_id);
+            ptrs[i] = rkey_obj.getPtr(desc.addr);
+        }
+        catch (const std::exception &e) {
+            NIXL_ERROR << "getMappedPtrs failed for descriptor " << i << ": " << e.what();
+            return NIXL_ERR_BACKEND;
```
rakhmets, line 1770: I think we don't want to modify `std::vector<void *> &ptrs` in this case.

src/utils/ucx/rkey.cpp
```diff
@@ -34,4 +34,21 @@ rkey::unpackUcpRkey(const nixlUcxEp &ep, const void *rkey_buffer) {
...
+rkey::getPtr(uint64_t raddr) const {
+    void *ptr = nullptr;
+    const ucs_status_t status = ucp_rkey_ptr(rkey_.get(), raddr, &ptr);
+
+    if (status == UCS_OK) {
+        return ptr;
+    }
+
+    if (status == UCS_ERR_UNREACHABLE) {
+        return nullptr;
```
rakhmets, line 48: I think we should just return ptr as `ucp_rkey_ptr` shouldn't modify it in case of error.

michal-shalev:
@brminich @yosefe @ofirfarjun7 @itayalroy 
Closing this PR following the latest API discussions.

PR #1101, opened 2025-12-09
Optimize nixlDlistH internals to reduce allocations and improve performance

benlwalker:
## What?
The first commit is a simple refactor to add accessors to nixlDlistH, without changing the implementation.
The second commit changes the std::unordered_map to a std::vector. Since the number of backends is small, this will improve performance and reduce memory use.
The third commit changes the std::vector to a std::array, which removes another allocation and a pointer chase. However, the third commit limits the number of backends in a nixlDlistH to 16.
The fourth commit replaces the std::array with a C-style zero length array and an overloaded new operator. This removes the limit of 16 backends again, at the cost of making new() a little more complex.

## Why?
nixlAgent::prepXferDlist does a lot of memory allocations and a lot of work, and for many LLM frameworks it is unfortunately called in the hot path. I am trying to optimize.

## How?
Replacing complex data structures with arrays.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -617,7 +617,7 @@ nixlAgent::prepXferDlist(const std::string &agent_name,
 
     // TODO [Perf]: Avoid heap allocation on the datapath, maybe use a mem pool
 
-    nixlDlistH *handle = new nixlDlistH;
+    nixlDlistH *handle = new (backend_set->size()) nixlDlistH(backend_set->size());
     if (init_side) {
         handle->isLocal = true;
         handle->remoteAgent = "";
--- a/src/core/transfer_request.h
+++ b/src/core/transfer_request.h
@@ -71,72 +71,110 @@ public:
 
 class nixlDlistH {
 private:
-    static constexpr size_t MAX_BACKENDS = 16;
-    std::array<std::pair<nixlBackendEngine *, std::shared_ptr<nixl_meta_dlist_t>>, MAX_BACKENDS>
-        descs;
+    size_t capacity_;
     size_t size_;
 
     std::string remoteAgent;
     bool isLocal;
 
-    // Helper method to find an entry by backend
-    inline auto
-    find(nixlBackendEngine *backend) {
-        return std::find_if(descs.begin(), descs.begin() + size_, [backend](const auto &p) {
-            return p.first == backend;
-        });
+    using DescPair = std::pair<nixlBackendEngine *, std::shared_ptr<nixl_meta_dlist_t>>;
+
+    // Helper to get typed pointer to the storage
+    inline DescPair *
+    getDescs() {
+        return reinterpret_cast<DescPair *>(descs_storage);
     }
 
-    inline auto
+    inline const DescPair *
+    getDescs() const {
+        return reinterpret_cast<const DescPair *>(descs_storage);
+    }
+
+    // Helper method to find an entry by backend
+    inline DescPair *
+    find(nixlBackendEngine *backend) {
+        auto descs = getDescs();
+        return std::find_if(
+            descs, descs + size_, [backend](const auto &p) { return p.first == backend; });
+    }
+
+    inline const DescPair *
     find(nixlBackendEngine *backend) const {
-        return std::find_if(descs.begin(), descs.begin() + size_, [backend](const auto &p) {
-            return p.first == backend;
-        });
+        auto descs = getDescs();
+        return std::find_if(
+            descs, descs + size_, [backend](const auto &p) { return p.first == backend; });
     }
 
 public:
-    inline nixlDlistH() : size_(0) {}
+    // Custom operator new to allocate space for flexible array member
+    static void *
+    operator new(size_t base_size, size_t num_backends) {
+        using DescPair = std::pair<nixlBackendEngine *, std::shared_ptr<nixl_meta_dlist_t>>;
+        size_t total = base_size + num_backends * sizeof(DescPair);
+        return ::operator new(total);
+    }
 
-    inline ~nixlDlistH() {
-        // shared_ptr handles cleanup automatically
+    static void
+    operator delete(void *p) noexcept {
+        ::operator delete(p);
+    }
+
+    // Placement delete for exception safety
+    static void
+    operator delete(void *p, size_t) noexcept {
+        ::operator delete(p);
+    }
+
+    explicit nixlDlistH(size_t capacity) : capacity_(capacity), size_(0) {}
+
+    ~nixlDlistH() {
+        // Manually destroy each pair in the flexible array member
+        auto descs = getDescs();
+        for (size_t i = 0; i < size_; ++i) {
+            descs[i].~DescPair();
+        }
     }
 
     // Accessor methods to encapsulate internal data structure
     inline size_t
     count(nixlBackendEngine *backend) const {
-        return find(backend) != (descs.begin() + size_) ? 1 : 0;
+        return find(backend) != (getDescs() + size_) ? 1 : 0;
     }
 
     inline std::shared_ptr<nixl_meta_dlist_t>
     at(nixlBackendEngine *backend) {
         auto it = find(backend);
-        if (it == descs.begin() + size_) throw std::out_of_range("Backend not found in descs");
+        auto descs = getDescs();
+        if (it == descs + size_) throw std::out_of_range("Backend not found in descs");
         return it->second;
     }
 
     inline std::shared_ptr<nixl_meta_dlist_t>
     at(nixlBackendEngine *backend) const {
         auto it = find(backend);
-        if (it == descs.begin() + size_) throw std::out_of_range("Backend not found in descs");
+        auto descs = getDescs();
+        if (it == descs + size_) throw std::out_of_range("Backend not found in descs");
         return it->second;
     }
 
-    inline std::shared_ptr<nixl_meta_dlist_t>
+    inline std::shared_ptr<nixl_meta_dlist_t> &
     operator[](nixlBackendEngine *backend) {
+        auto descs = getDescs();
         auto it = find(backend);
-        if (it != descs.begin() + size_) return it->second;
-        if (size_ >= MAX_BACKENDS)
-            throw std::out_of_range("nixlDlistH: Maximum number of backends exceeded");
-        descs[size_] = std::make_pair(backend, nullptr);
+        if (it != descs + size_) return it->second;
+        if (size_ >= capacity_) throw std::out_of_range("nixlDlistH: Maximum capacity exceeded");
+        // Use placement new to construct the pair in uninitialized memory
+        new (&descs[size_]) DescPair(backend, nullptr);
         return descs[size_++].second;
     }
 
     inline void
     erase(nixlBackendEngine *backend) {
+        auto descs = getDescs();
         auto it = find(backend);
-        if (it != descs.begin() + size_) {
+        if (it != descs + size_) {
             // Shift elements left to fill the gap
-            for (auto next = it + 1; next != descs.begin() + size_; ++next, ++it) {
+            for (auto next = it + 1; next != descs + size_; ++next, ++it) {
                 *it = *next;
             }
             --size_;
@@ -146,25 +184,29 @@ public:
     // Iterators for range-based for loops
     inline auto
     begin() {
-        return descs.begin();
+        return getDescs();
     }
 
     inline auto
     end() {
-        return descs.begin() + size_;
+        return getDescs() + size_;
     }
 
     inline auto
     begin() const {
-        return descs.begin();
+        return getDescs();
     }
 
     inline auto
     end() const {
-        return descs.begin() + size_;
+        return getDescs() + size_;
     }
 
     friend class nixlAgent;
+
+    // Flexible array member - must be last (using std::byte for storage)
+    alignas(std::pair<nixlBackendEngine *, std::shared_ptr<nixl_meta_dlist_t>>) std::byte
+        descs_storage[];
 };
 
 #endif


src/core/nixl_agent.cpp
```diff
@@ -627,7 +627,7 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
 
     // TODO [Perf]: Avoid heap allocation on the datapath, maybe use a mem pool
 
-    nixlDlistH *handle = new nixlDlistH;
+    nixlDlistH *handle = new(backend_set->size()) nixlDlistH(backend_set->size());
```
iyastreb: This is not exception safe
Please use a smart pointer, unique_ptr for example

src/core/transfer_request.h
```diff
@@ -70,20 +74,97 @@ class nixlXferReqH {
...
+        // Iterators for range-based for loops
+        inline auto begin() { return descs; }
+        inline auto end() { return descs + size_; }
+        inline auto begin() const { return descs; }
+        inline auto end() const { return descs + size_; }
+
     friend class nixlAgent;
+
+    // Flexible array member - must be last
+    std::pair<nixlBackendEngine*, nixl_meta_dlist_t*> descs[];
```
iyastreb: why not to store unique_ptrs, so that elements are automatically destroyed?

src/core/transfer_request.h
```diff
@@ -70,20 +74,97 @@ class nixlXferReqH {
 
 class nixlDlistH {
     private:
-        std::unordered_map<nixlBackendEngine*, nixl_meta_dlist_t*> descs;
+        size_t capacity_;
```
iyastreb: Could you please provide perf measurement before and after this change?
What exactly do we gain?

src/core/transfer_request.h
```diff
@@ -70,20 +74,97 @@ class nixlXferReqH {
...
+
+        inline auto find(nixlBackendEngine* backend) const {
+            return std::find_if(descs, descs + size_,
+                [backend](const auto& p) { return p.first == backend; });
+        }
+
     public:
-        inline nixlDlistH() { }
+        // Custom operator new to allocate space for flexible array member
+        static void* operator new(size_t base_size, size_t num_backends) {
```
iyastreb: Personally I don't like this complexity, unless it's really gives a huge perf increase. Maybe easier option is just to use a memory pool so that we don't contsruct heavy entries for each call, but reuse from the pool?

src/core/nixl_agent.cpp
```diff
@@ -627,7 +627,7 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
 
     // TODO [Perf]: Avoid heap allocation on the datapath, maybe use a mem pool
```
iyastreb, line 618: memory pool sounds like a superior alternative to these changes

src/core/nixl_agent.cpp
iyastreb, line 613: Another optimization would be to allocate tmp set on stack, and assign here a pointer to it. This way we get rid of heap allocation and make the code exception safe

src/core/nixl_agent.cpp
```diff
@@ -627,7 +627,7 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
...
 
-    nixlDlistH *handle = new nixlDlistH;
+    nixlDlistH *handle = new(backend_set->size()) nixlDlistH(backend_set->size());
     if (init_side) {
```
iyastreb, line 621: Maybe just create a proper ctor for nixlDlistH that takes all the args needed for initialization. So that:
- we encapsulate initialization in one place
- make it exception safe
- get rid of semi-valid (half-constructed) objects

src/core/nixl_agent.cpp
```diff
@@ -637,18 +637,18 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
     }
 
     for (auto & backend : *backend_set) {
-        handle->descs[backend] = new nixl_meta_dlist_t(descs.getType());
+        (*handle)[backend] = new nixl_meta_dlist_t(descs.getType());
```
iyastreb: maybe another optimization would be to create a vector of these objects at once, because we know the size in advance.

src/core/transfer_request.h
benlwalker: How do I use a memory pool if I don't know the number of backends up front? I can't pre-allocate these requests without putting a limit on the maximum number of backends. I'd also like to pool the nixl_meta_desc_t but similarly I don't know the sizes. I think placing limits (maybe user configurable) would go a long way toward allowing internal performance enhancements.

src/core/nixl_agent.cpp
benlwalker: See other comment. I don't know the sizes until this moment.

src/core/transfer_request.h
iyastreb: I think you can still use a memory pool even if we don't know the number of backends upfront, basically it would return already constructed object which you need to adjust with actual number of backends. IMO this is doable, but it is not necessarily needed. I'm not convinced that optimizing this code gives any visible improvement. What problem do we solve? First I would suggest to:
- Make perf measurement to define the problem you are solving (maybe it does not exist?)
- Provide numbers of perf boost with a fix
- Prefer clarity over complexity (this is a NIXL guideline). This new code is quite involved..

src/core/transfer_request.h
benlwalker: I've now made major updates to this PR. The first commit includes a new "null" plugin used for performance benchmarking. I've updated nixlbench to call makeXferReq inside the hot path as all software I've examined actually does. Then, I added a new optimization that avoids copying the nixl_meta_desc_t arrays inside makeXferReq if the user sets skipDescMerge. This is safe due to shared_ptr usage. I then ran nixlbench for several comparisons.

Comparing commit `ed95236` (baseline) vs `39f2de6` (skipDescMerge enabled with shared_ptr).

| Batch Size | Baseline (GB/s) | Optimized (GB/s) | Improvement | Status |
|:----------:|----------------:|-----------------:|------------:|:------:|
| 4 | 68.38 | 116.61 | **+70.5%** | ✅ |
| 8 | 148.07 | 212.64 | **+43.6%** | ✅ |
| 16 | 217.66 | 347.12 | **+59.5%** | ✅ |
| 32 | 300.90 | 497.43 | **+65.3%** | ✅ |
| 64 | 348.55 | 688.22 | **+97.4%** | ✅ |
| 128 | 443.94 | 769.77 | **+73.4%** | ✅ |

Comparing commit `39f2de6` (shared_ptr baseline) vs `c212f84` (flexible array member optimization).

| Batch Size | Baseline (GB/s) | Optimized (GB/s) | Improvement | Status |
|:----------:|----------------:|-----------------:|------------:|:------:|
| 4 | 118.13 | 119.59 | **+1.2%** | ✅ |
| 8 | 220.80* | 230.60* | **+4.4%** | ✅ |
| 16 | 267.17 | 312.37 | **+16.9%** | ✅ |
| 32 | 499.51 | 498.37 | -0.2% | ~ |
| 64 | 593.22 | 644.09 | **+8.6%** | ✅ |
| 128 | 794.26 | 800.07 | **+0.7%** | ✅ |

src/core/nixl_agent.cpp
benlwalker: I think I've seen an even better idea. In the skipMergeDesc case, we already have the exact object we need allocated. If I just track the lifetime with a shared_ptr I can skip allocating new lists entirely.

src/core/nixl_agent.cpp
benlwalker: Use a smart pointer for handle? It wasn't using a smart pointer before my patch.

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -47,7 +47,7 @@ DEFINE_string(worker_type, XFERBENCH_WORKER_NIXL, "Type of worker [nixl, nvshmem
 DEFINE_string(
     backend,
     XFERBENCH_BACKEND_UCX,
-    "Name of NIXL backend [UCX, GDS, GDS_MT, POSIX, GPUNETIO, Mooncake, HF3FS, OBJ, GUSLI] \
+    "Name of NIXL backend [UCX, GDS, GDS_MT, POSIX, GPUNETIO, Mooncake, HF3FS, OBJ, GUSLI, NULL] \
```
iyastreb, line 50: IMO we don't need a change in production code that is only needed to benchmark request creation.
I think this can be code just by calling request creation in a loop.

src/core/nixl_agent.cpp
iyastreb: Yes, there was a problem there even before this change.
Since you change this part, maybe this could be fixed as well
What happens with allocation if exception is thrown in the middle?

src/core/transfer_request.h
iyastreb: Personally I don't like the idea of NULL backend, because:
- a lot of production code change (200+LOC)
- the scope is unclear. Does it exist just to test request allocation in a loop? This can be done in gtest..
- It's not a real engine, because it does not transfer anything...

I find this benchmark a bit artificial, because we measure only request creation time.

src/core/nixl_agent.cpp
benlwalker: I'd rather fix existing, unrelated problems in a separate PR. This one has enough going on.

src/core/transfer_request.h
benlwalker: Having a NULL backend is very common in storage libraries. Linux has one in /dev/null. SPDK has one. They're really useful for building benchmarks of various software layers. You can use it to measure the software overhead of many different parts of NIXL.

I did change nixlbench here considerably. I could make this a separate PR that we work on first, and then come back to this one. The reason nixlbench needs to change is that it doesn't currently operate like all of the production software I've inspected. All of the applications I've seen at a minimum call makeXferReq in the hot path, immediately before they post the I/O. They must operate this way due to the way the NIXL API is designed - specifically because requests are tied to ranges when they are allocated and they can't later be adjusted. So nixlbench needs to reflect the actual usage pattern for this library.

Given the above, I've made the benchmark *less* artificial.

src/core/transfer_request.h
benlwalker: If I have to adjust the sizes on the fly, that's extra memory allocations. That's not the direction we want to go for performance.

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -47,14 +47,16 @@ DEFINE_string(worker_type, XFERBENCH_WORKER_NIXL, "Type of worker [nixl, nvshmem
...
               (only used with nixl worker)");
 DEFINE_string(initiator_seg_type, XFERBENCH_SEG_TYPE_DRAM, "Type of memory segment for initiator \
               [DRAM, VRAM]. Note: Storage backends always use DRAM locally.");
 DEFINE_string(target_seg_type, XFERBENCH_SEG_TYPE_DRAM, "Type of memory segment for target \
               [DRAM, VRAM]. Note: Storage backends determine remote type automatically.");
 DEFINE_string(scheme, XFERBENCH_SCHEME_PAIRWISE, "Scheme: pairwise, maytoone, onetomany, tp");
-DEFINE_string(mode, XFERBENCH_MODE_SG, "MODE: SG (Single GPU per proc), MG (Multi GPU per proc) [default: SG]");
+DEFINE_string(mode,
```
aranadive, line 57: Please see CONTRIBUTING.md on clang formatting only the diff.

src/core/transfer_request.h
aranadive: Sorry I dont understand the need for a NULL plugin. NIXL isnt a storage library. It also gives a wrong idea of what NIXL is to be used for. Its also a completely testing-based utility and should be kept on the testing side either through gtests or an isolated individual test.

Since nixlbench was written before any production app used NIXL, adding makeXferReq support to mimic real-world apps makes sense. But lets keep the previous path as well and do a separate PR with these changes.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1108,75 +1147,78 @@ execTransferIterations(nixlAgent *agent,
...
+
+        if (__builtin_expect(rc != NIXL_SUCCESS, 0)) {
+            std::cout << "NIXL Xfer failed with status: " << nixlEnumStrings::statusStr(rc)
+                      << std::endl;
+            agent->releaseXferReq(req);
+            agent->releasedDlistH(local_dlist_hndl);
+            agent->releasedDlistH(remote_dlist_hndl);
+            return -1;
         }
+        thread_stats.transfer_duration.add(timer.lap());
```
aranadive, line 1172: Lets keep this flow as well. And add the makeXferReq as additional way to create transfers.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1108,75 +1147,78 @@ execTransferIterations(nixlAgent *agent,
...
+    for (int i = 0; i < local_desc.descCount(); ++i) {
+        local_indices.push_back(i);
+    }
+    for (int i = 0; i < remote_desc.descCount(); ++i) {
+        remote_indices.push_back(i);
+    }
 
-            if (__builtin_expect(agent->releaseXferReq(req) != NIXL_SUCCESS, 0)) {
-                std::cout << "NIXL releaseXferReq failed" << std::endl;
-                return -1;
```
vvenkates27, line 1154: We already seems to have had createXferReq in data path for measurement. It would be good to support both paths. and PrepXferDlist must be outside the executeXfer

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1108,75 +1147,78 @@ execTransferIterations(nixlAgent *agent,
...
-            }
+    // Create/execute/release per iteration
+    for (int i = 0; i < num_iter; ++i) {
+        nixl_status_t make_rc = agent->makeXferReq(
```
vvenkates27, line 1185: Iterations is for removing measurement noise - .. this will affect network tests also, I don't like the idea of makeXfer in iterations.

src/core/transfer_request.h
iyastreb: You can just use an existing telemetry functionality to measure and report overhead, like the other parts of the code do. No need to create frameworks to prove that vector is populated faster than a map. Having this NULL backend code in production makes no sense, it just bloats the code and wastes compiler cycles.
I'm also sure that there are better ways to address the issue, e.g. by using a memory pool.
Can we discuss these potential improvements before implementing them?

src/core/nixl_agent.cpp
iyastreb: Since you are adding a new code, it must be done the right way.
No need to accumulate a technical debt

src/core/transfer_request.h
iyastreb: Also what does 800GBps mean in your measurement, if NULL backend does not transfer a single byte?
It is just wrong metric. How much does this optimization give with **real** workflow?
If it's around 1% - I would say it does not justify a huge complexity associated with it.
There are better ways to optimize it, I propose to discuss on a meeting

src/core/transfer_request.h
iyastreb: It's true, it may adjust the size on the fly for a very limited number of initial iterations, until it reaches a certain limit. After that it should remain at that size.
It's a trivial change, easy to implement and maintain, and it should give us much better performance I believe.
Can we make a mempool prototype and measure it's overhead?

src/core/transfer_request.h
benlwalker: * I can't use existing telemetry because the code paths become dominated by the backend plugins today, at least for the storage plugins. I don't have a set up where I can test UCX - I assume that's the fast path - but I really want to isolate the generic portions of the code and improve them. There are parallel efforts to improve the performance of the backends also ongoing and I do expect the performance of the backends to eventually improve to the point where these optimizations will be important. The storage backends are still 1 to 1.5 orders of magnitude off of the actual "speed of light" in terms of CPU efficiency.
* Having the NULL backend does not bloat anything. Plugins are dynamically loaded from shared libraries when the user requests to create the backend. If the user doesn't create the backend, it doesn't exist. Do you have specific lines of code in mind that you are concerned adding a NULL plugin bloats? 
* A memory pool does not address one of the big problems that this solution is fixing, namely that in makeXferReq it has to do a long chain of data dependent loads to find the correct nixl_meta_dlist_t for the backend. Avoiding all the pointer chasing is a big part of the improvement here. 
* A memory pool may improve overall performance here, given that the data path is doing a lot of allocations. I may *still* add a memory pool even with the change I'm proposing. But a memory pool has a lot of really tricky aspects and will be a much bigger lift to do correctly than the relatively simple thing I've done. For example:
  * Where do I put the memory pool? Is it global on the nixlAgent? That causes two issues - one is that nixlAgent doesn't have any threading requirements so now we need to take a lock to access the pool, while the other is that we don't keep the requests local to the current CPU (in the cache). Many memory allocators, including the one in glibc, automatically use thread-local caching, so they may actually beat a lock-based pool. And locks aren't free, even when single threaded and uncontended, because they do an atomic broadcast. So maybe I do my own per-thread caching? But NIXL doesn't have any per-thread context structure for me to piggy-back off of, so I guess I use a thread local cache. But it's entirely valid to allocate a request on one thread and release it on another, so I have to additionally handle that case with locks as a fallback to rebalance bad allocation patterns. I think it quickly becomes clear that any sort of memory pool is going to be extremely complicated and thousand(s) of lines of code.
  * What size objects do I make by default? If I allow them to be resizeable as needed, do I size them back down when they are released back to the pool? If I don't then a one time large request will permanently eat up memory. And if I make them dynamically sized using a std::vector or std::unordered_map, we're back to putting in all the pointer chasing through the data structures.
  * How many objects do I put in the pool? If I run out of entries in the pool, do I allocate more? Do I release them beyond some number on free, to keep the total memory use down? This behavior probably needs to be all user configurable.
* 800 Gbps is just a proxy for how many times per second it was able to call makeXferReq. Having a null disk/plugin/backend is an extremely common practice, they all work the same way like this, and no one gets confused by what it means.
  
I argue that what I've done in this patch is the *simple* solution. I'm trying to avoid adding "huge complexity" as you described it. The other approaches would require a lot more code.

src/core/nixl_agent.cpp
```diff
@@ -37,42 +37,54 @@ static const std::vector<std::vector<std::string>> illegal_plugin_combinations =
 };
 
 /*** nixlEnumStrings namespace implementation in API ***/
-std::string nixlEnumStrings::memTypeStr(const nixl_mem_t &mem) {
-    static std::array<std::string, FILE_SEG+1> nixl_mem_str = {
-           "DRAM_SEG", "VRAM_SEG", "BLK_SEG", "OBJ_SEG", "FILE_SEG"};
-    if (mem<DRAM_SEG || mem>FILE_SEG)
-        return "BAD_SEG";
+std::string
```
vvenkates27, line 40: Please let us move NIXL core changes to a separate PR and have more eyes on it with all stakeholders involved., as it would affect all NIXL Plugins.

benchmark/nixlbench/src/utils/utils.cpp
vvenkates27: +1 - There are too many formatting changes which are unrelated to the core PR, separating them can make reviewing this easier.

src/core/transfer_request.h
vvenkates27: A NULL backend typically helps isolate data path behavior where we want to separate storage performance from network performance in a networked storage solution. Given that this PR primarily measures lightweight core functions using this plugin, I tend to agree with @aranadive’s perspective.

Rather than introducing a separate plugin, could we consider:
- A testing‑only mode in `nixlbench` that writes to `/dev/null` and uses a stub for reads, or
- A parameterized NULL mode in the storage plugins (e.g., POSIX) gated for test environments, or
- Leverage the existing mocks in `test/gtest/mocks/` for core‑function performance tests.
These options should achieve the same goal while keeping maintenance overhead low.

iyastreb: 1) If you can't use telemetry - you can just test it from gtest by running in a loop. I don't think that 300+LOC backend is really needed for that.
2) With code bloat I meant the growing of the production code base that we need to maintain, and compile every time.
3) Let's discuss it on the meeting with NIXL team - if others are convinced that it adds a value - we can have it. But in a separate PR. Personally I don't see added value
4) Regarding optimization itself - I have doubts just because it complicates a lot the simple code. Therefore I propose to explore other options, maybe less efficient but more simple. Currently to me it looks like we are solving the problem that does not really exist yet.
5) For memory pool - yes, it might not be that simple, I do not have the exact proposal. But having a common memory pool infra would be useful in this and other places (e.g. for UCX backend handles), and seems to be a superior alternative.
6) To me this solution (both opt and NULL BE) looks unnecessary complex. It would be ok if it drastically improves performance of the hot-path, otherwise I would spend more time trying to make it simpler. Code clarity over complexity is NIXL design principle.

Just my opinion. We better have a meeting to discuss all of that

benlwalker:
I'm going to break this PR up into several different, smaller ones. The scope has become too large.

PR #1103, opened 2025-12-09
CI: Add consistency check for nixlbench

aranadive:
Lets add consistency check to catch any data validation errors. Also, no need for target_type with POSIX plugin since op_type determines direction.

### DIFF
--- a/.gitlab/test_nixlbench.sh
+++ b/.gitlab/test_nixlbench.sh
@@ -59,7 +59,7 @@ sleep 5
 echo "==== Running Nixlbench tests ===="
 cd ${INSTALL_DIR}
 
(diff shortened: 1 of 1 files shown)


.gitlab/test_nixlbench.sh
```diff
@@ -59,15 +59,22 @@ sleep 5
...
+run_nixlbench_noetcd() {
+    args="$@"
+    ./bin/nixlbench $DEFAULT_NB_PARAMS $args
 }
 
 run_nixlbench_one_worker() {
     benchmark_group=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
     args="$@"
-    run_nixlbench --benchmark_group $benchmark_group $args
+    run_nixlbench_noetcd $args
```
brminich, line 76: should it also include benchmark group?

aranadive: No, the benchmark group is only for when using etcd. I removed the benchmark_group generation here.

PR #1104, opened 2025-12-10
CI: Add BUILD_NIXL_EP param to container build job

Alexey-Rivkin:
## What?
Add `BUILD_NIXL_EP` boolean param to enable building NIXL containers with EP example support.

## Why?
NIXL v0.9.0 introduces EP example functionality that requires CI verification coverage.

## How?
Add boolean param to `proj-jjb.yaml` that passes `--build-nixl-ep` flag to `build-container.sh` when enabled.

### DIFF
--- a/.ci/jenkins/lib/build-container-matrix.yaml
+++ b/.ci/jenkins/lib/build-container-matrix.yaml
@@ -51,6 +51,7 @@ pipeline_start:
     echo "ENABLE_NIXL_BUILD: ${env.ENABLE_NIXL_BUILD}"
     echo "ENABLE_NIXLBENCH_BUILD: ${env.ENABLE_NIXLBENCH_BUILD}"
     echo "BUILD_TARGET: ${params.BUILD_TARGET}"
+    echo "BUILD_NIXL_EP: ${params.BUILD_NIXL_EP}"
 
 # Build pipeline
 steps:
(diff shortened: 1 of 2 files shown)


Alexey-Rivkin:
Test build (internal link):
https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/job/nixl-ci-build-container/32

.ci/jenkins/lib/build-container-matrix.yaml
```diff
@@ -206,6 +217,7 @@ pipeline_stop:
                 <p>Base Image: <b>${params.BASE_IMAGE}:${params.BASE_IMAGE_TAG}</b></p>
                 <p>Architectures: <b>x86_64, aarch64</b></p>
                 ${params.UPDATE_LATEST ? '<p>Latest tag updated: <b>Yes</b></p>' : ''}
+                ${params.BUILD_NIXL_EP ? "<p>Build NIXL EP: <b>Yes</b></p>" : ''}
```
brminich, line 220: should we print no otherwise?

Alexey-Rivkin: Should we also print "no" for UPDATE_LATEST?

ovidiusm: I would do it unless there is a reason not to

dpressle: In case its not selected we dont print anything which is fine for this job from our perspective

brminich: ok

brminich (approved):

PR #1107, opened 2025-12-10
Change log error to info for failed fetch and invalidate from etcd

2dm:
## What?
Fetching MD from etcd will log an error when the MD is missing, and then a watcher is set.
`invalidateRemoteMD` will also log an error if the MD in not found. 
These are not unexpected events, thus should be categorized as error or presented to the user as errors.

## Why?
**Fetch** - In async run, agent can try to fetch a peer's MD before it is set by the peer. This is a valid use case in which the agent will then wait for the MD to be added. Failing to fetch at this point is not an error. A watch is set for this key for whenever it is added, or return error on timeout.
**invalidate** - local invalidation can fail if the MD is not found. However, this should also not be an error because the MD might have been invalidated by running `invalidateLocalMD` by the remote agent. At the end of the `invalidateRemoteMD` there is no valid MD, so it should not be presented as an error.

## How?
1. Changed these log prints to info. 
2. Added test cases to view the log print (these are not evaluated during the test)


2dm:
@aranadive @ovidiusm can you please review?

aranadive (approved):

PR #1113, opened 2025-12-11
posix: Implement batched descriptor submission with 16-descriptor limit

benlwalker:
## What?
Limit POSIX queue submissions to 16 descriptors outstanding at a time. 

## Why?
Some network attached filesystems are exhibiting strange performance behavior when we send too many requests to a single file, rather than spreading the work across all the files more evenly. This solves the issue, but keeps enough requests outstanding to still efficiently reach line rate with a single file. On a local file in /tmp, I get greatly improved performance as well, which was unexpected:

Linux AIO with batch size 32 and I/O size 1MB - 3.47x faster than current main branch
URING with batch size 32 and I/O size 1MB - 22x faster than current main branch

## How?
It only submits up to 16 elements out of a batch at any given time. The batching is transparent to callers - they continue polling checkXfer() until the batch completes - but this does mean that users need to actually call checkXfer to make forward progress. We expect to later introduce a progress thread to the POSIX backend so that progress will be made regardless of whether the user calls checkXfer.

I ran data verification tests for all impacted use cases with nixlbench and they pass.

### DIFF
--- a/src/plugins/posix/aio_queue.cpp
+++ b/src/plugins/posix/aio_queue.cpp
@@ -22,13 +22,13 @@
 #include <string.h>
 #include <time.h>
 #include <stdexcept>
+#include <absl/strings/str_format.h>
 
 aioQueue::aioQueue(int num_entries, nixl_xfer_op_t operation)
     : aiocbs(num_entries),
       num_entries(num_entries),
       completed(num_entries),
       num_completed(0),
-      num_submitted(0),
       operation(operation) {
     if (num_entries <= 0) {
         throw std::runtime_error("Invalid number of entries for AIO queue");
@@ -40,8 +40,9 @@ aioQueue::aioQueue(int num_entries, nixl_xfer_op_t operation)
 
 aioQueue::~aioQueue() {
     // There should not be any in-flight I/Os at destruction time
-    if (num_submitted > num_completed) {
-        NIXL_ERROR << "Programming error: Destroying aioQueue with " << (num_submitted - num_completed) << " in-flight I/Os";
+    if (num_ios_outstanding > 0) {
+        NIXL_ERROR << "Programming error: Destroying aioQueue with " << num_ios_outstanding
+                   << " in-flight I/Os";
     }
 
     // Cancel any remaining I/Os
@@ -53,11 +54,13 @@ aioQueue::~aioQueue() {
 }
 
 nixl_status_t
-aioQueue::submit (const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
-    num_submitted = 0;
-    // Submit all I/Os at once
-    for (auto& aiocb : aiocbs) {
+aioQueue::submitBatch(int start_idx, int count, int &submitted_count) {
+    // Submit the batch
+    submitted_count = 0;
+    for (int i = start_idx; i < start_idx + count; i++) {
+        auto &aiocb = aiocbs[i];
         if (aiocb.aio_fildes == 0 || aiocb.aio_nbytes == 0) continue;
+
         // Check if file descriptor is valid
         if (aiocb.aio_fildes < 0) {
             NIXL_ERROR << "Invalid file descriptor in AIO request";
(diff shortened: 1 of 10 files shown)


anton-nayshtut:
+1

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -100,9 +87,18 @@ linuxAioQueue::checkCompleted() {
...
 
-    num_ios_to_complete -= rc;
+    num_ios_outstanding -= rc;
+    if (!num_ios_outstanding) {
+        // No outstanding IOs, check if we need to submit more
+        if (num_ios_submitted_total < num_ios_to_submit) {
+            // More IOs to submit, signal to caller to submit more
+            return NIXL_IN_PROG;
+        }
+        // All IOs submitted and completed
```
anton-nayshtut, line 97: `num_ios_submitted_total` needs to be zeroed here to enable the next sending attempt. Same applies to other queues.

anton-nayshtut (changes requested):

src/plugins/posix/aio_queue.cpp
```diff
@@ -73,35 +76,33 @@ aioQueue::submit (const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
...
+        submitted_count++;
     }
 
-    completed.assign(num_entries, false);
-    num_completed = 0;
-    return NIXL_IN_PROG;
+    return NIXL_SUCCESS;
 }
 
 nixl_status_t aioQueue::checkCompleted() {
```
vvenkates27, line 95: Completed is not reset. I think this may be a bug. Without this reset, subsequent transfers will skip checking IOs because completed[i] remains true from the previous transfer, causing an infinite loop in checkCompleted().
Probably need something like 
 num_completed = 0;
 completed.assign(num_entries, false);

src/plugins/posix/aio_queue.cpp
```diff
@@ -73,35 +76,33 @@ aioQueue::submit (const nixl_meta_dlist_t &, const nixl_meta_dlist_t &) {
...
-                // If we hit the kernel limit, cancel all submitted I/Os and return error
-                NIXL_ERROR << "AIO submit failed: kernel queue full";
-                for (auto& cb : aiocbs) {
-                    if (cb.aio_fildes != 0) {
-                        aio_cancel(cb.aio_fildes, &cb);
-                    }
-                }
-                return NIXL_ERR_BACKEND;
+                // If we hit the kernel limit, stop submitting and return partial success
+                NIXL_ERROR << absl::StrFormat(
```
vvenkates27, line 80: Consider using NIXL_INFO? Since EAGAIN is an expected condition that the batching logic handles gracefully (returning NIXL_SUCCESS with partial count), logging it as an error may be misleading?

src/plugins/posix/linux_aio_queue.cpp
vvenkates27: +1

src/plugins/posix/uring_queue.cpp
```diff
@@ -94,38 +94,43 @@ UringQueue::~UringQueue() {
...
+    // io_uring_submit can return partial submissions
+    submitted_count = ret;
+    if (ret != count) {
+        NIXL_ERROR << absl::StrFormat("io_uring submit partial: %d/%d", ret, count);
+    }
+
+    return NIXL_SUCCESS;
 }
 
 nixl_status_t UringQueue::checkCompleted() {
```
vvenkates27, line 129: Same issue as the other two APIs.

src/plugins/posix/linux_aio_queue.cpp
```diff
@@ -52,37 +50,26 @@ linuxAioQueue::~linuxAioQueue() {
...
-            NIXL_ERROR << absl::StrFormat("linux_aio submit failed: %s", nixl_strerror(-ret));
-        } else {
-            NIXL_ERROR << absl::StrFormat(
-                "linux_aio submit failed. Partial submission: %d/%d", num_ios_to_submit, ret);
-        }
-        return NIXL_ERR_BACKEND;
+    // io_submit can return partial submissions
+    submitted_count = ret;
+    if (ret != count) {
+        NIXL_ERROR << absl::StrFormat("linux_aio submit partial: %d/%d", ret, count);
```
vvenkates27, line 65: Same question here.. NIXL_ERROR here is misleading.

src/plugins/posix/uring_queue.cpp
```diff
@@ -94,38 +94,43 @@ UringQueue::~UringQueue() {
...
-    if (num_completed == num_entries) {
+    // Check if all IOs are submitted and completed
+    if (num_ios_submitted_total >= num_ios_to_submit && num_completed == num_ios_to_submit) {
```
vvenkates27, line 131: We use if (num_ios_submitted_total >= num_ios_to_submit && num_ios_outstanding == 0) in aio_queue.cpp, probably makes sense to keep them uniform as num_ios_outstanding is managed by the base class and is more reliable?

src/plugins/posix/posix_queue.h
```diff
@@ -24,19 +24,29 @@
...
-        virtual ~nixlPosixQueue() = default;
-        virtual nixl_status_t
-        submit (const nixl_meta_dlist_t &local, const nixl_meta_dlist_t &remote) = 0;
-        virtual nixl_status_t checkCompleted() = 0;
-        virtual nixl_status_t prepIO(int fd, void* buf, size_t len, off_t offset) = 0;
+protected:
+    int num_ios_to_submit = 0;
+    int num_ios_submitted_total = 0;
+    int num_ios_outstanding = 0;
+    static constexpr int MAX_IO_OUTSTANDING = 16;
```
vvenkates27, line 31: Add some comments why we need this, its probably a good idea to add a README for POSIX plugin as the  knobs and usage increases.

brminich:
@benlwalker can you pls address comments?

PR #1120, opened 2025-12-12
fix(rust-bindings): detect NIXL availability and fallback to stubs

ryanolson:
- Use built-in compiler types in wrapper.h instead of system headers (stdbool.h, stddef.h, stdint.h) to fix bindgen failures on systems where libclang cannot locate standard headers

- Add check_nixl_available() to detect if NIXL library is installed by checking pkg-config and common library paths for libnixl.so

- Fallback to stub API when NIXL is not found, with clear warning: "NIXL library not found, building with stub API"

This allows the Rust bindings to build successfully on systems without NIXL installed, enabling development and testing of Rust code that depends on nixl-sys without requiring the full NIXL installation.

### DIFF
--- a/src/bindings/rust/build.rs
+++ b/src/bindings/rust/build.rs
@@ -84,6 +84,34 @@ fn get_nixl_libs() -> Option<Vec<pkg_config::Library>> {
     }
 }
 
+/// Check if NIXL library is available on the system
+fn check_nixl_available() -> bool {
+    // First try pkg-config
+    if pkg_config::probe_library("nixl").is_ok() {
+        return true;
+    }
+
+    // Fall back to checking common library paths
+    let nixl_root_path =
+        env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
+    let arch = get_arch();
+
+    // Check various possible library locations
+    let possible_paths = [
+        format!("{}/lib/{}-linux-gnu/libnixl.so", nixl_root_path, arch),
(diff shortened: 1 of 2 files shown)


src/bindings/rust/build.rs
```diff
@@ -84,6 +84,34 @@ fn get_nixl_libs() -> Option<Vec<pkg_config::Library>> {
...
+        return true;
+    }
+
+    // Fall back to checking common library paths
+    let nixl_root_path =
+        env::var("NIXL_PREFIX").unwrap_or_else(|_| "/opt/nvidia/nvda_nixl".to_string());
+    let arch = get_arch();
+
+    // Check various possible library locations
+    let possible_paths = [
```
milesial, line 100: Can the check for availability use the same set of paths that would be sent to the linker? We're missing 2 of the 5 right now. And maybe use the same function for both so they are always aligned

```
    // Add all possible library paths
    println!("cargo:rustc-link-search=native={}", nixl_lib_path);
    println!("cargo:rustc-link-search=native={}", nixl_root_path);
    println!("cargo:rustc-link-search=native={}/lib", nixl_root_path);
    println!("cargo:rustc-link-search=native={}/lib64", nixl_root_path);
    println!("cargo:rustc-link-search=native={}/lib/x86_64-linux-gnu", nixl_root_path);
```

brminich:
@ryanolson can you pls anser the comment if this pr is still relevant?

ryanolson:
We can close. Stubs fixed in main

PR #1122, opened 2025-12-13
EXAMPLES: Modify nixl_ep to use UCX channels 

2dm:
## What?
This PR introduce updates to nixl-ep example:
1. Updates from DeepEP library 
2. Use of channel API

### DIFF
--- a/examples/device/ep/csrc/kernels/nixl_ep.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep.cu
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2025 DeepSeek
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * This file incorporates material from the DeepSeek project, licensed under the MIT License.
  * The modifications made by NVIDIA are licensed under the Apache License, Version 2.0.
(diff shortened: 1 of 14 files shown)


examples/device/ep/meson.build
```diff
@@ -78,6 +74,15 @@ nixl_ep_inc_dirs = [
     torch_inc_dirs,
 ]
 
+doca_build_deps = []
+if doca_gpunetio_dep.found()
+    doca_build_deps = [doca_gpunetio_dep]
+    doca_inc_dirs = doca_gpunetio_dep.get_pkgconfig_variable('includedir').split(':')
+    foreach inc : doca_inc_dirs
```
itayalroy: I believe this is redundant, meson dependencies should handle include paths and nixl ep compiles without this change

examples/device/ep/nixl_ep/__init__.py
```diff
@@ -18,6 +18,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import torch
```
itayalroy, line 21: Why is this needed?

examples/device/ep/csrc/kernels/nixl_ep.cu
```diff
@@ -301,7 +301,7 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
                                rank, local_expert_idx, src_rank, poll_count, current_value);
                     }
                 }
-                num_recv_tokens = current_value - 1;
+                current_value == 0 ? num_recv_tokens = 0 : num_recv_tokens = current_value - 1;
```
itayalroy, line 304: How did fault tolerance work without this fix until now?

2dm: If the kill falls on combine it will work fine, but dispatch will crash it.

examples/device/ep/meson.build
2dm: fixed

examples/device/ep/nixl_ep/__init__.py
2dm: import error due to topk_idx_t

itayalroy:
Do we have some perf comparison between multiple workers single channel and single worker multiple channels?
I think it's worth to run some sanity at least to verify that this does not cause major degradations

itayalroy (changes requested):
This PR does not reduce UCX backend's `num_workers` to 1

+ @eranrs this is why you probably did not see a huge benefit with it. The PR needs to be fixed & re-tested for control path perf

itayalroy:
Merged in #1175

PR #1127, opened 2025-12-14
Plugins/OBJ: Add support for S3 CRT Client

aranadive:
## What?
This PR introduces the **S3 CRT client** as an additional S3 client type for the object plugin to improve performance for large object transfers by automatically using multipart uploads and ranged GET requests, which increase throughput for large KV-to-object transfers. Both **S3Client** and **S3CRTClient** are supported and selected via a new backend parameter, **`crtMinLimit`**, so that when the object size exceeds this threshold the **S3CRTClient** is used. Shared logic between S3 and S3 CRT flows is refactored into common files, and the Dockerfiles are updated to compile and include the S3 CRT client in the build.

## Why?
The S3 CRT client is designed for high-throughput transfers and leverages multipart upload and byte-range fetches automatically, which yields better performance and reliability for large objects compared with the standard S3 client. Using it only above a configurable size threshold via `crtMinLimit` allows the system to gain these benefits for large transfers while keeping behavior unchanged for smaller objects and minimizing unnecessary resource usage.

## How?
- Add **`crtMinLimit`** as a backend configuration parameter to determine when to route operations to **S3CRTClient** instead of **S3Client** based on object size.
- Instantiate and use **S3CRTClient** for objects larger than `crtMinLimit`, taking advantage of automatic multipart uploads and ranged GETs for improved throughput.
- Refactor common S3 logic into shared modules used by both clients and update Dockerfiles so the build process compiles and packages the S3 CRT client and its runtime dependencies.


src/plugins/obj/s3_crt/client.cpp
```diff
@@ -0,0 +1,146 @@
...
+        "GetObjectStreamBuf", reinterpret_cast<unsigned char *>(data_ptr), data_len);
+    auto stream_factory = Aws::MakeShared<Aws::IOStreamFactory>(
+        "GetObjectStreamFactory", [preallocated_stream_buf]() -> Aws::IOStream * {
+            return new Aws::IOStream(preallocated_stream_buf.get());
+        });
+
+    Aws::S3Crt::Model::GetObjectRequest request;
+    request.WithBucket(bucketName_)
+        .WithKey(Aws::String(key))
+        .WithRange(absl::StrFormat("bytes=%d-%d", offset, offset + data_len - 1));
```
jgoldsch12, line 114: By providing a range, does this short circuit the range get optimizations in CRT?

https://github.com/awslabs/aws-c-s3/blob/main/docs/GetObject.md

aranadive: It shouldnt matter since the CRT client is invoked for larger data sizes and s3 crt should still be able to do the range get optimizations.

src/plugins/obj/s3_crt/client.cpp
```diff
@@ -0,0 +1,146 @@
...
+              auto *opts = new Aws::SDKOptions();
+              Aws::InitAPI(*opts);
+              return opts;
+          }(),
+          [](Aws::SDKOptions *opts) {
+              Aws::ShutdownAPI(*opts);
+              delete opts;
+          }) {
+
+    Aws::S3Crt::ClientConfiguration config;
```
jgoldsch12, line 35: This part I'm not fully understanding.  The min object size for CRT parameter is set, but then isn't used to configure the part size in the S3 CRT client config.  Instead, two SDKs are used  (one for larger objects and one for smaller), when CRT could handle both by configuring the part size and thresholds for multi-part upload.  I'm curious why this approach wasn't considered?

aranadive: Good question. I also optimizing for small object sizes where I believe CRT isnt good for latency and using the regular S3 SDK would be beneficial. The crtMinLimit provides a cutoff for stopping S3 usage and using/enabling CRT SDK. Also gives some flexibility for end user. To only use CRT, crtMinLimit can just be set to 1.

src/plugins/obj/s3_crt/client.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <aws/core/utils/memory/stl/AWSStringStream.h>
+#include <absl/strings/str_format.h>
+#include <iostream>
+
+awsS3CrtClient::awsS3CrtClient(nixl_b_params_t *custom_params,
+                               std::shared_ptr<Aws::Utils::Threading::Executor> executor)
+    : awsOptions_(
+          []() {
+              auto *opts = new Aws::SDKOptions();
+              Aws::InitAPI(*opts);
```
vvenkates27: Is this safe if both clients call it? do we need a std::call_once or something?

src/plugins/obj/s3_crt/client.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    s3CrtClient_->PutObjectAsync(
+        request,
+        [callback, preallocated_stream_buf, data_stream](
+            const Aws::S3Crt::S3CrtClient *,
+            const Aws::S3Crt::Model::PutObjectRequest &,
+            const Aws::S3Crt::Model::PutObjectOutcome &outcome,
+            const std::shared_ptr<const Aws::Client::AsyncCallerContext> &) {
+            if (!outcome.IsSuccess()) {
+                std::cerr << "putObjectAsync (CRT) error:" << outcome.GetError() << std::endl;
```
vvenkates27: use NIXL_ERROR?

src/plugins/obj/s3_crt/client.h
```diff
@@ -0,0 +1,75 @@
...
+// Forward declaration from s3/client.h
+using put_object_callback_t = std::function<void(bool success)>;
+using get_object_callback_t = std::function<void(bool success)>;
+
+/**
+ * Concrete implementation using AWS SDK S3CrtClient for high-performance transfers.
+ * The S3 CRT (Common Runtime) client uses AWS Common Runtime for improved performance with
+ * large objects, providing better throughput and lower CPU utilization.
+ */
+class awsS3CrtClient {
```
vvenkates27: Consider class awsS3CrtClient : public iS3Client to reduce code duplication in postXfer().?

src/plugins/obj/s3_crt/client.cpp
aranadive: AWS did add ref counting for [Init and Shutdown](https://github.com/aws/aws-sdk-cpp/pull/2710) so its protected against it. But may not be a bad idea to call it just once since we will have more S3 client types in future.

aranadive: Added.

src/plugins/obj/s3_crt/client.h
aranadive: Yes. Done.

src/plugins/obj/s3_crt/client.cpp
aranadive: Done

vvenkates27 (approved):
LGTM

PR #1129, opened 2025-12-15
pre-commit: Run 'git clang-format' in pre-commit

benlwalker:
## What?
This will detect code format issues in C/C++ that would have failed the CI.

## Why?
My PRs keep failing in CI because I forget to run git clang-format

## How?
I added a custom pre-commit hook that runs git clang-format. This will only format changed lines.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,6 +47,14 @@ repos:
         args: ["--toml", "pyproject.toml"]
         exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*\.md|.*\.rs$|src/utils/serdes/serdes.h)
 
+  - repo: local
+    hooks:
+      - id: staged-clang-format
+        name: clang-format
+        entry: git clang-format
+        language: system
+        files: \.(cpp|h|hpp|cc|c|cxx|cu|cuh)$
+
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:


.pre-commit-config.yaml
```diff
@@ -47,6 +47,14 @@ repos:
         args: ["--toml", "pyproject.toml"]
         exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*\.md|.*\.rs$|src/utils/serdes/serdes.h)
 
+  - repo: local
+    hooks:
+      - id: staged-clang-format
+        name: clang-format
+        entry: git clang-format
+        language: system
+        files: \.(cpp|h|hpp|cc|c|cxx|cu|cuh)$
```
brminich, line 56: is it really needed for git clang-format?
maybe pass ` pass_filenames: false` instead?

dpressle, line 56: Would this check only modified files? you dont want to get errors on files you didnt updated

yosefe (changes requested):
IMO such PR will add unneeded effort when working on "temporary" patches that are not upstreamed immediately, and not necessarily need to adhere to code formatting rules. If there is a way to make it "opt in" then ok, otherwise it's too restrictive

brminich:
> IMO such PR will add unneeded effort when working on "temporary" patches that are not upstreamed immediately, and not necessarily need to adhere to code formatting rules. If there is a way to make it "opt in" then ok, otherwise it's too restrictive

@yosefe, so shall we close that?

PR #1132, opened 2025-12-16
perf(nixl_ep): parallelize metadata fetch in _nixl_agents_connect

eranrs:
Fire all fetchRemoteMD requests upfront and poll all peers in a unified loop instead of sequentially. Reduces connect latency from O(N * avg_latency) to O(max_latency).

## What?

Parallelize metadata fetching in `_nixl_agents_connect`:
- Fire all `fetchRemoteMD()` requests upfront before polling
- Poll all peers in a unified loop instead of sequentially per-peer

## Why?

The previous implementation fetched and waited for each peer sequentially, causing O(N × avg_latency) connect time. With 8 ranks, this resulted in ~800ms connect latency.

## How?

Fire all fetch requests first, then poll all peers in a unified loop. This reduces latency to O(max_latency).

**Result:** `agents_connect` reduced from ~800ms to ~250ms (~3x faster) with 8 ranks.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -269,16 +269,31 @@ void Buffer::_nixl_agents_connect(const std::vector<int>& ranks) {
         nixl_agent_info->remote_agent_names[remote_rank] = std::to_string(remote_rank);
     }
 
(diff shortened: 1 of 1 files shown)


ebarilanM:
@itayalroy  please review

PR #1135, opened 2025-12-16
Install the missing headers required by external plugin builds

piotr-topnotch:
## What?
Installs the missing C++ headers required to build an external plugin.


piotr-topnotch:
I don't think the GitLab failure is related to my changes. It looks like a timeout issue.

aranadive:
Thanks @piotr-topnotch

PR #1136, opened 2025-12-16
Change nixlbench to optionally call createXferReq on every I/O

benlwalker:
## What?
Change nixlbench to call makeXferReq on every I/O. The prepXferDlist calls are outside of the main I/O loop.

## Why?
1. We observe that all real software using NIXL calls either makeXferReq or createXferReq in the I/O path, immediately before posting the request.
2. I cannot imagine a system that wouldn't call makeXferReq on demand. Once a request is created, it's tied to all of the ranges it will be transferirng. There's no fathomable system that could know all of the exact batches it will transfer over the life of the application at start up time.

Reason 1 is insufficient for the change here because all real software I can find also busy polls in a loop after calling postXferReq and that doesn't mean we should make NIXL synchronous. Instead, it's the combination of reason 1 and reason 2 that justifies the change.

## How?
For the GUSLI path there was already logic to call createXferReq in the I/O path. I first made that always the case, rather than just for GUSLI. Then I broke createXferReq into two prepXferDlist calls up front and makeXferReq in the I/O path because that's more efficient and still realistic.

### DIFF
--- a/benchmark/nixlbench/README.md
+++ b/benchmark/nixlbench/README.md
@@ -462,9 +462,10 @@ sudo systemctl start etcd && sudo systemctl enable etcd
 
 #### Storage Backend Options (GDS, GDS_MT, POSIX, HF3FS, OBJ)
 ```
---filepath PATH            # File path for storage operations
---num_files NUM            # Number of files used by benchmark (default: 1)
---storage_enable_direct    # Enable direct I/O for storage operations
+--filepath PATH                        # File path for storage operations
+--num_files NUM                        # Number of files used by benchmark (default: 1)
+--storage_enable_direct                # Enable direct I/O for storage operations
+--recreate_xfer_req_per_iteration      # Recreate transfer request on every iteration (default: false, auto-enabled for GUSLI)
 ```
 
 #### Backend-Specific Options
@@ -690,6 +691,7 @@ GUSLI provides direct user-space access to block storage devices, supporting loc
 **Notes**:
 - Number of devices in `--device_list` must match `--num_initiator_dev` and `--num_target_dev`
 - Direct I/O is automatically enabled for GUSLI (no need to specify `--storage_enable_direct`)
+- Transfer request recreation per iteration is automatically enabled for GUSLI (can be overridden with `--recreate_xfer_req_per_iteration=false`)
 
 ### Worker Types
 
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -238,6 +238,14 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
             xferBenchConfig::storage_enable_direct = true;
         }
 
+        // GUSLI backend requires per-iteration request recreation - enable it automatically
+        if (!xferBenchConfig::recreate_xfer_req_per_iteration) {
+            std::cout
+                << "GUSLI backend: Automatically enabling recreate_xfer_req_per_iteration"
+                << std::endl;
+            xferBenchConfig::recreate_xfer_req_per_iteration = true;
+        }
+
         // Parse and configure GUSLI devices from general device_list parameter
         int expected_num_devices =
             isInitiator() ? xferBenchConfig::num_initiator_dev : xferBenchConfig::num_target_dev;
@@ -1109,7 +1117,7 @@ prepareTransferDescriptors(nixl_xfer_dlist_t &local_desc,
 }
 
 // Execute transfers with configurable request lifecycle behavior
-// recreate_per_iteration: true for GUSLI (bug workaround), false for standard backends
+// recreate_per_iteration: if true, creates/releases transfer request on each iteration
 static int
 execTransferIterations(nixlAgent *agent,
                        const nixl_xfer_op_t op,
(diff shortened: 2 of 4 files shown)


aranadive:
Lets keep the previous createXferReq path as well. This denotes the absolute limit for the system config similar to many performance tests as well as to test nixl API in general. MakeXferReq can denote an alternate path for applications to perform flexible allocations.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1108,75 +1107,78 @@ execTransferIterations(nixlAgent *agent,
...
-                return -1;
-            }
-            total_prepare_duration += timer.lap();
+    nixl_status_t prep_rc =
+        agent->prepXferDlist(NIXL_INIT_AGENT, local_desc, local_dlist_hndl, &params);
+    if (NIXL_SUCCESS != prep_rc) {
+        std::cerr << "prepXferDlist (local) failed: " << nixlEnumStrings::statusStr(prep_rc)
+                  << std::endl;
+        return -1;
+    }
```
iyastreb: I would propose to use a scope guard to automatically release handles when goes out of scope. This way we don't need release calls in many places, and it's also exception safe. Just allocate it once right after prepXferDlist, same for remote_dlist_hndl
```
    auto local_guard = make_scope_guard ([&] {
        agent->releasedDlistH(local_dlist_hndl);
    });

**Comment by benlwalker** [comment]
> Lets keep the previous createXferReq path as well. This denotes the absolute limit for the system config similar to many performance tests as well as to test nixl API in general. MakeXferReq can denote an alternate path for applications to perform flexible allocations.

I strongly disagree here. It's not possible to create a system that doesn't call either createXferReq or makeXferReq on each I/O request, so it isn't worth measuring without it. If we need to pull together the team to have a meeting on this, let's do that after the holidays.

**Comment by brminich** [comment]
imo this change is important. Recently we saw that nixlbench does not really correspond to some KV cache serving benchmarks.
Maybe let's have this new flow as a non-default option?
@aranadive @iyastreb @ovidiusm

**Comment by iyastreb** [comment]
> imo this change is important. Recently we saw that nixlbench does not really correspond to some KV cache serving benchmarks. Maybe let's have this new flow as a non-default option? @aranadive @iyastreb @ovidiusm

Yes, I fully agree that in order to reproduce real world applications behavior it would be extremely useful to have an optional "recreate-per-tx" behavior. On my side I'm using a trivial 2 lines patch to test that:
```
diff --git a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
index cceb2044..2fcaa9eb 100644
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -1174,6 +1174,8 @@ execTransferIterations(nixlAgent *agent,
         // Standard path: Single request for all iterations
         for (int i = 0; i < num_iter; ++i) {
             nixl_status_t rc = execSingleTransfer(agent, req, timer, thread_stats);
+            agent->releaseXferReq(req);
+            agent->createXferReq(op, local_desc, remote_desc, target, req, &params);
 
             if (__builtin_expect(rc != NIXL_SUCCESS, 0)) {
                 std::cout << "NIXL Xfer failed with status: " << nixlEnumStrings::statusStr(rc)
```

Maybe if we add a config option for that - that would be enough?

**Comment by benlwalker** [comment]
I've reworked this to be a user parameter to control the behavior.

File: benchmark/nixlbench/README.md [hunk 2]
```diff
@@ -462,9 +462,10 @@ sudo systemctl start etcd && sudo systemctl enable etcd
 
 #### Storage Backend Options (GDS, GDS_MT, POSIX, HF3FS, OBJ)
 ```
---filepath PATH            # File path for storage operations
---num_files NUM            # Number of files used by benchmark (default: 1)
---storage_enable_direct    # Enable direct I/O for storage operations
+--filepath PATH                        # File path for storage operations
+--num_files NUM                        # Number of files used by benchmark (default: 1)
+--storage_enable_direct                # Enable direct I/O for storage operations
+--recreate_xfer_req_per_iteration      # Recreate transfer request on every iteration (default: false, auto-enabled for GUSLI)
```
iyastreb, line 468: Maybe we can choose shorter name?
`persistent_xfer_req` (true by default)
`make_req_per_iter` ("iter" is already used in other args like num_iter, warmup_iter)

brminich: 👍 both options suggested by Ilia sound good to me

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1139,7 +1147,7 @@ execTransferIterations(nixlAgent *agent,
     // Execute transfer iterations
     // Branch prediction hint: most backends don't recreate per iteration
     if (__builtin_expect(recreate_per_iteration, 0)) {
-        // GUSLI path: Create/execute/release per iteration
+        // Per-iteration recreation path: Create/execute/release per iteration
```
brminich, line 1150: i'd remove `__builtin_expect` on the line above now

brminich:
pls note clang format failures

benlwalker:
Anton made a similar change in PR #1051. I'm going to close this PR and we'll go with that one.

PR #1140, opened 2025-12-17
Add microbenchmarks

benlwalker:
## What?
Add a new gtest directory to hold microbenchmarks. Then add a new benchmark for nixlAgent::makeXferReq.

## Why?
These will allow us to test the performance of isolated pieces of NIXL easily.

## How?
This is a gtest based test. It runs like the other unit tests. I could have used Google Benchmark but I didn't want to pull in another dependency.

## Summary by CodeRabbit

* **Tests**
  * Added new benchmark test suite for transfer request operations with configurable test scenarios.
  * Updated test build configuration for consistent test execution behavior.


test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+        ASSERT_EQ(status, NIXL_SUCCESS);
+        ASSERT_NE(backend_handle, nullptr);
+        backend_handle_ptr = backend_handle;
+    }
+
+
+    /**
+     * @brief Create and register memory buffers
+     */
+    void
```
iyastreb: More C++ idiomatic way would be to return std::vector from the function, no extra copy made

test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+    createRegisteredMem(size_t size, size_t count, std::vector<void *> &buffers) {
+        buffers.clear();
+        nixlDescList<nixlBlobDesc> desc_list(DRAM_SEG);
+
+        for (size_t i = 0; i < count; i++) {
```
iyastreb, line 197: since we know count in advance, better reserve the memory before the loop to avoid vector reallocations:
`buffers.reserve(count)`

test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+        nixl_status_t status = agent->registerMem(desc_list);
+        ASSERT_EQ(status, NIXL_SUCCESS);
+    }
+
+    /**
+     * @brief Deregister memory buffers
+     */
+    void
+    deregisterMem(const std::vector<void *> &buffers, size_t size) {
```
iyastreb: I think better option is to have a vector of bytes: `std::vector<std::vector<uint8_t>>`.
- benefit from RAII, elements are destroyed with vector
- exception safe
- size is stored inside the elements, so no need to pass it from outside

test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+        // Cleanup
+        agent->releasedDlistH(local_dlist);
+        agent->releasedDlistH(remote_dlist);
+        deregisterMem(src_buffers, desc_size);
+        deregisterMem(dst_buffers, desc_size);
+    }
+
+    std::string
+    getAgentName() {
```
iyastreb, line 439:
```suggestion
    std::string_view
    getAgentName() const {
```

test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+        return "benchmark_agent";
+    }
+
+    gtest::ScopedEnv env;
```
iyastreb: I think according to NIXL formatting rules, member variables should have _ suffix?

test/gtest/benchmark/make_xfer_req.cpp
benlwalker: I know, but to be able to use the gtest ASSERT_* macros the function needs to return void. And I do want to use these ASSERT_ macros because it's a fatal error if any of the conditions in this function hit.

test/gtest/benchmark/make_xfer_req.cpp
benlwalker: The problem here is that prepXferDlist and the nixlAgent constructor both take std::string, so if I make this change I end up having to just convert it to a string anyway.

test/gtest/benchmark/make_xfer_req.cpp
benlwalker: With the change to use std::vector of bytes I was able to make this change.

test/gtest/benchmark/make_xfer_req.cpp
iyastreb: ok, but at least function can be const

test/gtest/benchmark/make_xfer_req.cpp
```diff
@@ -0,0 +1,473 @@
...
+ *
+ * This test fixture creates a minimal environment with a single agent and mock backend
+ * to measure the performance of makeXferReq in isolation from any real backend implementation.
+ * Uses loopback transfers (agent to itself) to avoid multi-agent overhead.
+ */
+class BenchmarkMakeXferReq : public ::testing::Test {
+protected:
+    void
+    SetUp() override {
+        env_.addVar("NIXL_TELEMETRY_ENABLE", "n");
```
brminich, line 71: why is it needed? We may have other vars affecting perf, e.g. NIXL_LOG_LEVEL
(also it is disabled by default)

test/gtest/benchmark/meson.build
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
iyastreb, line 1: pls update all copyrights to 2026

iyastreb (dismissed):

brminich (dismissed):

brminich:
@benlwalker can you pls update copyright in all files touched by this PR, otherwise it will break CI once it it merged (because of merge commit)

brminich:
@benlwalker can you pls update copyrights?

PR #1142, opened 2025-12-18
feat(libfabric): Implement notification fragmentation for large messages

akkart-aws:
---

## What?

Implemented notification message fragmentation in the NIXL libfabric backend to support unlimited notification sizes by splitting messages into 256-byte chunks.

## Why?

TensorRT-LLM's disaggregated serving architecture requires exchanging connection metadata between nodes. On P5 instances with 32 EFA devices, this connection data is approximately 1,792 bytes (32 devices × 56 bytes per endpoint), which exceeds the previous 1,024-byte notification limit, causing connection establishment to fail.

This change enables TensorRT-LLM to work on P5 instances and scales to support 256+ rails per node (~14KB connection data).

## How?

__Sender-Side:__

- Fragmentation happens once in `prepXfer()`, splitting messages into 256-byte chunks
- Pre-fragmented notifications stored in backend handle
- `notifSendPriv()` sends all fragments with sequence metadata

__Receiver-Side:__

- Fragments stored in pending map indexed by xfer_id
- Dual-level tracking: waits for both fragment completion AND write completion
- `checkPendingNotifications()` reassembles complete messages

__Additional Improvements:__

- Added recv pool pre-posting (128 requests per rail) for improved throughput
- Added `__attribute__((packed))` to BinaryNotification for cross-platform compatibility
- Extracted `fragmentNotificationMessage()` helper function for code reusability

__Modified Files:__

- `src/utils/libfabric/libfabric_common.h`
- `src/plugins/libfabric/libfabric_backend.h`
- `src/plugins/libfabric/libfabric_backend.cpp`
- `src/utils/libfabric/libfabric_rail.cpp`

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/plugins/libfabric/libfabric_backend.h
+++ b/src/plugins/libfabric/libfabric_backend.h
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_common.h
+++ b/src/utils/libfabric/libfabric_common.h
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_rail_manager.h
+++ b/src/utils/libfabric/libfabric_rail_manager.h
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_common.cpp
+++ b/src/utils/libfabric/libfabric_common.cpp
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_rail.h
+++ b/src/utils/libfabric/libfabric_rail.h
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/utils/libfabric/libfabric_rail_manager.cpp
+++ b/src/utils/libfabric/libfabric_rail_manager.cpp
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");


src/plugins/libfabric/libfabric_backend.h
```diff
@@ -158,13 +159,17 @@ class nixlLibfabricBackendH : public nixlBackendReqH {
...
 
+    /** Atomically increment submitted request count */
+    void
+    increment_submitted_requests();
+
     /** Get current count of completed requests */
     size_t
     get_completed_requests_count() const;
 
     /** Get total number of requests used for this transfer */
```
fengjica: Do we want to update this comment? I feel it is a little generic.

src/plugins/libfabric/libfabric_backend.h
```diff
@@ -224,28 +229,38 @@ class nixlLibfabricEngine : public nixlBackendEngine {
...
-        PendingNotification(const std::string &agent,
-                            const std::string &msg,
-                            uint16_t xfer_id,
-                            uint32_t expected_cnt = 0)
-            : remote_agent(agent),
-              message(msg),
-              post_xfer_id(xfer_id),
-              expected_completions(expected_cnt),
-              received_completions(0) {}
+        PendingNotification()
```
fengjica: This constructor and the one below can be having one call the other, to eliminated duplicated code.

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,16 +100,42 @@
...
 
 /**
- * @brief Binary notification format with counter-based matching
+ * @brief Binary notification header containing all metadata fields
+ *
+ * This structure contains all notification metadata except the message payload.
+ * Used to calculate the optimal fragment size automatically.
+ */
+struct BinaryNotificationHeader {
+    char agent_name[1024]; // Fixed-size agent name (null-terminated)
```
fengjica: If the space efficiency is critical here, is it better to make this a field for agent_name size, and move the actual agent_name to the start of the message? So nothing is wasted if the agent_name is not as long as 1024 bytes.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1166,10 +1179,55 @@ nixlLibfabricEngine::releaseReqH(nixlBackendReqH *handle) const {
     return NIXL_SUCCESS;
 }
 
-// notifSendPriv that accept control request
+/****************************************
+ * Notification Functions
+ *****************************************/
+
+std::vector<BinaryNotification>
+nixlLibfabricEngine::fragmentNotificationMessage(const std::string &message,
```
fengjica: Is it better to move fragments to the function parameter pointer, to avoid extra copies?

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,16 +100,42 @@
...
+ * @note The __attribute__((packed)) ensures consistent byte layout across platforms,
+ *       preventing padding-related data corruption during network serialization.
  */
 struct BinaryNotification {
-    char agent_name[256]; // Fixed-size agent name (null-terminated)
-    char message[1024]; // Fixed-size message (binary data, not null-terminated)
-    uint32_t message_length; // Actual length of message data
-    uint16_t xfer_id; // 16-bit postXfer ID (unique per postXfer call)
-    uint32_t expected_completions; // Total write requests for this xfer_id
+    BinaryNotificationHeader header; // All metadata fields
```
fengjica: Same info (e.g. agent_name and others) is repeated in every fragment. Can we transfer it only once in the first fragment?

src/utils/libfabric/libfabric_rail_manager.h
```diff
@@ -182,8 +182,9 @@ class nixlLibfabricRailManager {
                              const std::vector<size_t> &remote_selected_endpoints,
                              const std::unordered_map<size_t, std::vector<fi_addr_t>> &dest_addrs,
                              uint16_t agent_idx,
+                             uint16_t xfer_id,
                              std::function<void()> completion_callback,
-                             BinaryNotification *binary_notif);
+                             std::function<void()> submission_callback);
```
fengjica: If this is not called in an asynchronous place, it can be just called on the caller site after this function?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1371,53 +1465,64 @@ nixlLibfabricEngine::processNotification(const std::string &serialized_notif) {
...
-            }
+               << " notif_xfer_id=" << notif_xfer_id << " notif_seq_id=" << notif_seq_id << "/"
+               << notif_seq_len << " expected_completions: " << expected_completions;
+
+    {
+        std::lock_guard<std::mutex> lock(receiver_tracking_mutex_);
+        auto it = pending_notifications_.find(notif_xfer_id);
+        if (it == pending_notifications_.end()) {
+            PendingNotification pending_notif(notif_xfer_id);
+            pending_notifications_[notif_xfer_id] = pending_notif;
```
fengjica: Can we use emplace() to eliminate an extra copy?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1371,53 +1465,64 @@ nixlLibfabricEngine::processNotification(const std::string &serialized_notif) {
...
-        // Check if any notifications can now be completed (after releasing the lock)
-        checkPendingNotifications();
-    } else {
-        // Regular notification without expected completions - process immediately
-        NIXL_TRACE << "Regular notification (expected_completions=0), processing immediately";
-        std::lock_guard<std::mutex> lock(notif_mutex_);
-        notifMainList_.push_back({remote_name, msg});
-        NIXL_TRACE << "Regular notification processed immediately: " << msg;
+        // Initialize fragment vector on first fragment
+        if (!pending_notifications_[notif_xfer_id].buffer_resized) {
```
fengjica: can we remove 'buffer_resized' by just checking the size of message_fragments, which will be 0 before resizing at the first time?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1519,27 +1625,22 @@ nixlLibfabricEngine::addReceivedXferId(uint16_t xfer_id) {
...
-            // Case 2: Write arrived before notification - create placeholder with INT_MAX
-            PendingNotification placeholder;
-            placeholder.remote_agent = ""; // Empty until notification arrives
-            placeholder.message = ""; // Empty until notification arrives
-            placeholder.post_xfer_id = xfer_id;
-            placeholder.expected_completions = INT_MAX; // Sentinel value
-            placeholder.received_completions = 1; // Start with this completion
-
+        if (it == pending_notifications_.end()) {
+            PendingNotification placeholder(xfer_id);
```
fengjica: Similarly to the above, can the extra copy be eliminated?

src/utils/libfabric/libfabric_common.h
```diff
@@ -50,7 +50,7 @@
 #define NIXL_LIBFABRIC_RECV_POOL_SIZE 128 // Number of recv requests to pre-post per rail
 
 // Notification fragmentation constants
-#define NIXL_LIBFABRIC_NOTIFICATION_FRAGMENT_SIZE 1024 // Size of BinaryNotification.message field
+#define NIXL_LIBFABRIC_NOTIFICATION_FRAGMENT_SIZE 256 // Size of BinaryNotification.message field
```
amitrad-aws: Why did you reduce this?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1243,40 +1238,41 @@ nixlLibfabricEngine::notifSendPriv(const std::string &remote_agent,
...
 
         // Set fragmentation fields for this notification
-        binary_notification.notif_xfer_id = notif_xfer_id;
-        binary_notification.expected_completions = expected_completions;
-        binary_notification.notif_seq_id = static_cast<uint16_t>(seq_id);
-        binary_notification.notif_seq_len = static_cast<uint16_t>(binary_notifications.size());
-        binary_notification.total_message_length = total_message_length;
+        binary_notification.header.notif_xfer_id = notif_xfer_id;
+        binary_notification.header.expected_completions = expected_completions;
+        binary_notification.header.notif_seq_id = static_cast<uint16_t>(seq_id);
```
amitrad-aws: Just define seq_id as uint16_t

src/utils/libfabric/libfabric_common.h
amitrad-aws: I think the most robust and space saving way would be have that as part of the dynamic data size.
Have the message start with the agent_name followed by the data (only in first fregment). We just need to agent_name_size to the header instead of the agent_name

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Yes. Done.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Yes. Thanks, good catch. Done

src/utils/libfabric/libfabric_rail_manager.h
akkart-aws: Done.

src/utils/libfabric/libfabric_common.h
akkart-aws: Done.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Thanks. Good catch. Done.

src/utils/libfabric/libfabric_common.h
akkart-aws: Done. I have split the fragmentation header such that we have a common header for all fragments and then a special header just for the first fragment which has additional metadata which are common across all fragments
like total_payload_length, expected_completions, agen_name_len

src/plugins/libfabric/libfabric_backend.h
akkart-aws: If we are using emplace, we can remove the default constructor here since emplace allows you to create the map entry directly.

src/plugins/libfabric/libfabric_backend.h
akkart-aws: Done.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Yes. Done.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: done

src/utils/libfabric/libfabric_common.h
akkart-aws: Sorry thats a transient mistake in the commit. There was no need to reduce the fragmentation size from 1024 to 256. I only did this for a local testing to ensure the fragments are joined correctly on the receiver side. 
Eventually the fragment size will be set to the total buffer size. We only need to fragment the notification message if its greater than 8192 bytes.
 
I squashed the above commits as you suggested into one commit
`add notification fragmentation...", "add attribute((packed))...", "auto-calculate fragment size..." into one commit.`

But however nixl PR does do a squash of all the commits in the PR.
The reason why I split into multiple commits was so that we can review the commits individually. or understand how we are changing the notification flow step by step from fixed size notification to variable size notification.
Actually we can merge this now into one commit .

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1189,35 +1189,75 @@ nixlLibfabricEngine::fragmentNotificationMessage(const std::string &message,
...
     total_message_length = static_cast<uint32_t>(message.length());
-    size_t num_fragments = (total_message_length + fragment_size - 1) / fragment_size;
-
-    // Resize vector to hold all fragments
+    const size_t max_control_msg_size = BinaryNotification::getFragmentSize();
+    
+    // Calculate fragment 0 capacity (has extra headers + agent_name)
+    size_t frag0_overhead = sizeof(BinaryNotificationHeader)
+                          + sizeof(BinaryNotificationFirstFragmentHeader)
+                          + agent_name.length();
```
amitrad-aws: The agent_name sohuld be part of the total length, don't assume it fits into the first segment

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1189,35 +1189,75 @@ nixlLibfabricEngine::fragmentNotificationMessage(const std::string &message,
...
-                   << ", length=" << fragment_len;
+        // Set header fields
+        BinaryNotificationHeader header;
+        header.notif_xfer_id = 0;  // Will be set later in notifSendPriv
+        header.notif_seq_id = static_cast<uint16_t>(frag_idx);
+        header.notif_seq_len = static_cast<uint16_t>(num_fragments);
+        
+        if (frag_idx == 0) {
+            // Fragment 0: Pack first_frag_header + agent_name + message
+            size_t msg_len = std::min(frag0_capacity, static_cast<size_t>(total_message_length));
```
amitrad-aws: Will this work if msg_len will come out 0 (valid corner case)?

src/utils/libfabric/libfabric_common.h
```diff
@@ -100,79 +100,194 @@
...
+                          uint32_t expected_completions,
+                          uint16_t agent_name_length) {
+        first_frag_header_.total_payload_length = total_payload_length;
+        first_frag_header_.expected_completions = expected_completions;
+        first_frag_header_.agent_name_length = agent_name_length;
+    }
+
+    /** @brief Set agent name (only for fragment 0) */
     void
     setAgentName(const std::string &name) {
```
amitrad-aws: There is a assumption here that this is called before setMessage.
If there is no return value to check, if there is already something in the buffer, prepend the agent name

src/utils/libfabric/libfabric_common.h
```diff
@@ -100,79 +100,194 @@
...
+            // Fragment 1+: write message only
+            memcpy(ptr + offset, buffer_.data(), buffer_.size());
+            offset += buffer_.size();
+        }
+
+        return offset;
+    }
+
+    /** @brief Deserialize from buffer */
+    static BinaryNotification
```
amitrad-aws: Should return a reference, there is a lot of data that might be copied here

src/utils/libfabric/libfabric_common.h
```diff
@@ -100,79 +100,194 @@
...
+        memcpy(&notif.header_, ptr + offset, sizeof(notif.header_));
+        offset += sizeof(notif.header_);
+
+        if (notif.header_.notif_seq_id == 0) {
+            // Fragment 0: read first_frag_header + agent_name + message
+            memcpy(&notif.first_frag_header_, ptr + offset, sizeof(notif.first_frag_header_));
+            offset += sizeof(notif.first_frag_header_);
+
+            // Read agent_name + message
+            size_t remaining = size - offset;
```
amitrad-aws: Duplicated with else, this can be outside the if statement

amitrad-aws (changes requested):

src/utils/libfabric/libfabric_common.h
akkart-aws: Done.

src/utils/libfabric/libfabric_common.h
akkart-aws: Done

src/utils/libfabric/libfabric_common.h
akkart-aws: I combined the two setMessage and setAgentName to setCombinedPayload. which will take a concatenated string of agent_name + notification_message

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: i think so, but I believe utlitimately if the agent name length + first fragment header + common header > fragment length, this would have failed.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Right. I have fixed it now.

src/plugins/libfabric/libfabric_backend.h
```diff
@@ -224,28 +225,29 @@ class nixlLibfabricEngine : public nixlBackendEngine {
...
-              post_xfer_id(xfer_id),
-              expected_completions(expected_cnt),
-              received_completions(0) {}
+                                       // notif_xfer_id
+        uint16_t expected_msg_fragments; // Total fragments expected (from notif_seq_len)
+        uint16_t received_msg_fragments; // Fragments received so far
+        uint32_t total_message_length; // Total length of complete message (all fragments)
+        uint16_t agent_name_length; // Length of agent_name in combined payload
+
+        // Only using parameterized constructor to be used with try_emplace() for map operations
```
fengjica: Is it better to remove this comment line, as it gives details of the implementation?

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,51 +101,158 @@
...
  */
 struct BinaryNotification {
-    char agent_name[256]; // Fixed-size agent name (null-terminated)
-    char message[1024]; // Fixed-size message (binary data, not null-terminated)
-    uint32_t message_length; // Actual length of message data
-    uint16_t xfer_id; // 16-bit postXfer ID (unique per postXfer call)
-    uint32_t expected_completions; // Total write requests for this xfer_id
+    BinaryNotificationHeader header_;
+    BinaryNotificationMetadata metadata_; // Only valid for seq_id=0
+    std::vector<char> buffer_; // Chunk of (agent_name + message) combined payload
```
fengjica: std::string?

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,51 +101,158 @@
...
+
+    /** @brief Constructor */
+    BinaryNotification() {
+        memset(&header_, 0, sizeof(header_));
+        memset(&metadata_, 0, sizeof(metadata_));
+    }
+
+    /** @brief Get the maximum control message size */
+    static size_t
+    getFragmentSize() {
```
fengjica: Can this be a static member variable of the struct, if there is no future plan to make it anything complex? Also, better to name it something like max_control_msg_size or max_fragment_size to show the purpose currently in the comment.

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,51 +101,158 @@
...
+ * Fragment 1+ layout: [Header:10B] [combined_payload_chunk:variable]
+ *
+ * After reassembly, use metadata.agent_name_length to split the combined payload:
+ *   - agent_name = combined_payload.substr(0, agent_name_length)
+ *   - message = combined_payload.substr(agent_name_length)
+ *
+ * @note The __attribute__((packed)) ensures consistent byte layout across platforms,
+ *       preventing padding-related data corruption during network serialization.
  */
 struct BinaryNotification {
```
fengjica: I believe this can be a class, rather than a struct, as the member vars are selectively used in (de)serialization.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1179,46 +1283,74 @@ nixlLibfabricEngine::notifSendPriv(const std::string &remote_agent,
...
-        control_request,
-        connection->control_rail_remote_addr_list_[control_rail_id][0],
-        connection->agent_index_);
+        // Update first fragment header with expected_completions (only for fragment 0)
+        // Note: agent_name_length was already set during fragmentation
+        if (seq_id == 0) {
+            const BinaryNotificationMetadata &metadata = binary_notification.getMetadata();
+            binary_notification.setMetadata(
+                total_message_length, expected_completions, metadata.agent_name_length);
+        }
```
fengjica, line 1303: can we move fragmentation logic to a later place in the code where notif_xfer_id and expected_completions are available, so that we don't have to update them here?

src/utils/libfabric/libfabric_common.h
akkart-aws: Done.

src/utils/libfabric/libfabric_common.h
akkart-aws: Done

src/utils/libfabric/libfabric_common.h
akkart-aws: Done.

src/plugins/libfabric/libfabric_backend.h
akkart-aws: Done.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: The current design intentionally separates concerns between prepXfer() and postXfer() for performance optimization (https://github.com/ai-dynamo/nixl/blob/main/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp#L1103-L1181):

Current Flow:
prepXfer() (called once per createXferReq()) - SLOW PATH

Performs expensive fragmentation: string concatenation, memory allocation, vector operations
Creates BinaryNotification fragments with placeholder values for notif_xfer_id and expected_completions
Stores fragments in the handle for reuse
postXfer() (called multiple times per postXferReq()) - FAST PATH

Generates new notif_xfer_id for each iteration
Calculates expected_completions based on actual submitted requests
Only updates these 2 fields in the pre-created fragments (minimal overhead)
Why this design is optimal:
Reusability: The same notification message is created once in prepXfer() but reused across multiple postXfer() calls with different xfer_id values. Moving fragmentation to postXfer() would repeat expensive operations on every iteration.

Performance: The current overhead in the fast path is just updating 2 uint32_t fields. Moving fragmentation would add significant overhead (string operations, memory allocation) to every postXfer() call.

src/utils/libfabric/libfabric_common.h
```diff
@@ -99,51 +101,161 @@
...
+     */
     void
-    setMessage(const std::string &msg) {
-        message_length = std::min(msg.length(), sizeof(message));
-        memcpy(message, msg.data(), message_length);
-        // Zero out remaining space for consistency
-        if (message_length < sizeof(message)) {
-            memset(message + message_length, 0, sizeof(message) - message_length);
-        }
+    setPayload(std::string payload) {
```
fengjica, line 189: IIUC, this still needs one copy, unless the caller can also do std::move(), which may not be the case. "const std::string &" not worse in that case, right? Or is string_view considered?

fengjica (approved):

src/utils/libfabric/libfabric_common.h
fengjica: Discussed offline.

akkart-aws: Yes. There were two ways I could think of doing this.
1. Use of const std:string & and do the memcpy in the later stage.
2. Use the std::string argument and use the std::move to move the source string.
 I choose option 2 because we call setPayload when we fragment the string and we use a substr operation and pass the rvalue to the setPayload method. Using std::string and later doing a move helps for compiler optimization. Its not a lot and I dont have a strong preference for this to be done either way. 
 If required we can revert this .

PR #1144, opened 2025-12-18
nixl_ep: Metadata exchange via torch TCPStore

itayalroy:
Most inference engines already have their own communication groups,
making it unnecessary to launch a separate ETCD server solely for
nixl_ep connection establishment.

This commit adds support for metadata exchange using an external
torch.distributed.TCPStore instance (as used by vLLM).

This is likely an intermediate solution. Further discussion is needed
for the broader NIXL approach, but the nixl_ep API changes are expected
to remain unchanged.


yosefe:
@tstamler  @brminich  can you pls review?

PR #1145, opened 2025-12-18
CI: Add CUDA file support to clang-format

michal-shalev:
## What?
Fix CI to check formatting on CUDA (.cu/.cuh) files.

## Why?
`clang-format-diff` was silently ignoring CUDA files by default, allowing unformatted code to pass CI.

## How?
Added `-iregex` flag to explicitly include .cu/.cuh extensions. Deduplicated file pattern regex.


michal-shalev:
/ok to test [07e7550](https://github.com/ai-dynamo/nixl/pull/1145/commits/07e75507f4939cd6e7e16032656f2f7c9f9aefc6)

PR #1149, opened 2025-12-21
fix(libfabric): Use PCI bus ID for GPU-to-EFA mapping

akkart-aws:
## What?
Fix incorrect EFA device selection when using `CUDA_VISIBLE_DEVICES` by switching from GPU ID-based to PCI bus ID-based topology mapping.

## Why?
The previous implementation used GPU IDs (0, 1, 2...) to map GPUs to EFA devices. This broke when `CUDA_VISIBLE_DEVICES` was set because:
- `CUDA_VISIBLE_DEVICES=1` makes physical GPU 1 appear as device 0 to the application
- Topology mapping still used physical GPU IDs, causing mismatch
- Result: Wrong EFA devices selected for GPU memory, breaking GPU Direct RDMA

Example failure case:
```
CUDA_VISIBLE_DEVICES=1  # Physical GPU 1 → appears as device 0
App registers memory on "device 0" (actually physical GPU 1)
Old code: Looks up GPU 0 in topology → selects EFA devices for physical GPU 0
```

## How?

**Changed topology mapping from GPU ID to GPU-PCI bus ID**:
- Topology layer now uses GPU-PCI bus ID as key instead of GPU ID
- `gpu_to_efa_devices` → `pci_to_efa_devices`
- `getEfaDevicesForGpu(int)` → `getEfaDevicesForGPUPci(string)`

**Backend queries PCI bus ID from memory address**:
- Extended `cudaQueryAddr()` to return GPU'S PCI bus ID via `cuDeviceGetPCIBusId()`
- Query happens after setting CUDA context in `registerMem()`
- PCI bus ID passed to rail_manager for topology lookup

**Rail manager uses PCI bus ID for rail selection**:
- Updated `registerMemory()` and `selectRailsForMemory()` to accept PCI bus ID
- Calls `topology->getEfaDevicesForGPUPci(pci_bus_id)` instead of GPU ID lookup

**Benefits**:
- Works correctly with `CUDA_VISIBLE_DEVICES` - GPU-PCI bus ID is stable regardless of device renumbering
- More robust - uses physical hardware identifier instead of application-visible device ID

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -56,7 +56,7 @@
 
 #ifdef HAVE_CUDA
 static int
-cudaQueryAddr(void *address, bool &is_dev, CUdevice &dev, CUcontext &ctx) {
+cudaQueryAddr(void *address, bool &is_dev, CUdevice &dev, CUcontext &ctx, std::string &pci_bus_id) {
     CUmemorytype mem_type = CU_MEMORYTYPE_HOST;
     uint32_t is_managed = 0;
     CUpointer_attribute attr_type[4];
@@ -75,6 +75,19 @@ cudaQueryAddr(void *address, bool &is_dev, CUdevice &dev, CUcontext &ctx) {
     result = cuPointerGetAttributes(4, attr_type, attr_data, (CUdeviceptr)address);
     is_dev = (mem_type == CU_MEMORYTYPE_DEVICE);
 
+    // Get PCI bus ID if device memory
+    if (result == CUDA_SUCCESS && is_dev) {
+        char pci_buf[32];
+        CUresult pci_result = cuDeviceGetPCIBusId(pci_buf, sizeof(pci_buf), dev);
+        if (pci_result == CUDA_SUCCESS) {
+            pci_bus_id = std::string(pci_buf);
+        } else {
+            pci_bus_id = "";
+        }
+    } else {
+        pci_bus_id = "";
+    }
+
     return (CUDA_SUCCESS != result);
 }
 
@@ -89,6 +102,7 @@ nixlLibfabricCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &wa
     bool is_dev;
     CUdevice dev;
     CUcontext ctx;
+    std::string pci_bus_id; // Not used here, but required by cudaQueryAddr
     int ret;
 
     was_updated = false;
@@ -96,7 +110,7 @@ nixlLibfabricCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &wa
     if (expected_dev == -1) return -1;
     if (myDevId_ != -1 && expected_dev != myDevId_) return -1;
 
-    ret = cudaQueryAddr(address, is_dev, dev, ctx);
+    ret = cudaQueryAddr(address, is_dev, dev, ctx, pci_bus_id);
     if (ret) return ret;
     if (!is_dev) return 0;
     if (dev != expected_dev) return -1;
@@ -734,6 +748,7 @@ nixlLibfabricEngine::registerMem(const nixlBlobDesc &mem,
     priv->length_ = mem.len;
     priv->gpu_device_id_ = mem.devId; // Store GPU device ID
 
+    std::string pci_bus_id = "";
 #ifdef HAVE_CUDA
     // Handle CUDA memory registration with GPU Direct RDMA support
     if (nixl_mem == VRAM_SEG) {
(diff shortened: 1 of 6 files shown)


src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -135,18 +135,31 @@ nixlLibfabricTopology::discoverEfaDevices() {
...
+                 device,
+                 function);
+        std::string normalized_id(normalized_pci);
+
+        auto it = pci_to_efa_devices.find(normalized_id);
+        if (it != pci_to_efa_devices.end()) {
+            NIXL_DEBUG << "Found EFA devices for PCI " << pci_bus_id << " (normalized to "
+                       << normalized_id << ")";
+            return it->second;
+        }
```
fengjica, line 158: Could we add logging for the case that normalized_id is not found?

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -543,11 +560,9 @@ nixlLibfabricTopology::buildTopologyAwareGrouping() {
 nixl_status_t
 nixlLibfabricTopology::buildFallbackMapping() {
```
fengjica, line 567: This private function is used only in one place. Maybe delete it?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -321,16 +323,25 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
 std::vector<size_t>
 nixlLibfabricRailManager::selectRailsForMemory(void *mem_addr,
                                                nixl_mem_t mem_type,
-                                               int gpu_id) const {
+                                               int gpu_id,
```
fengjica, line 326: This is only used when being compared < 0. If that's unnecessary we probably can drop this argument completely?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -429,6 +444,8 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
 
         struct fid_mr *mr;
         uint64_t key;
+        // Pass gpu_id parameter to individual rail's registerMemory calls
+        // For VRAM, this is not used since we already selected rails based on PCI bus ID
```
fengjica: Is this comment line "For VRAM, this is not used since we already selected rails based on PCI bus ID" accurate? gpu_id is used internally in registerMemory(), right?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -734,9 +748,11 @@ nixlLibfabricEngine::registerMem(const nixlBlobDesc &mem,
     priv->length_ = mem.len;
     priv->gpu_device_id_ = mem.devId; // Store GPU device ID
 
+    std::string pci_bus_id = "";
 #ifdef HAVE_CUDA
     // Handle CUDA memory registration with GPU Direct RDMA support
     if (nixl_mem == VRAM_SEG) {
+
```
fengjica: extra line?

src/utils/libfabric/libfabric_topology.cpp
akkart-aws: NIXL_WARN << "No EFA devices found for PCI bus ID " << pci_bus_id << ", returning all devices";

We have a warn on L162

src/utils/libfabric/libfabric_topology.cpp
akkart-aws: I was thinking we can keep the interface for now ?
The purpose of buildFallbackMapping was to take in a topology xml and then generate a grouping of EFA Cards and the GPUs in case the toplogy map cannot be generated by hwloc.
Although we are not using this currently, I think we would use this sometime in the future. (like for Neuron or instance types like GB200 where we may actuall have to use the topology file to generate the grouping)

src/utils/libfabric/libfabric_rail_manager.cpp
akkart-aws: Yeah. I wanted to drop this argument for gpu_id, but then for NEURON we might still use gpu_id, so can we delete it in a subsequent commit.
Additionally not all device types might support apis like `cuDeviceGetPCIBusId` which would get us the gpu_id or pci_bus_id from registerd memory. so I kept this argument for such cases.
Also this is in sync with the fact that nixl agent's memomory regsitration api uses gpu_id.

src/utils/libfabric/libfabric_rail_manager.cpp
akkart-aws: No. you are right. this commnent is not accurate. gpu_id is passed to mr_attr.device.cuda.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Done

amitrad-aws (dismissed):

src/utils/libfabric/libfabric_topology.cpp
fengjica: Yeah, I see that there is a warning line on 162. I was concerned is that it won't distinguish the two different else cases, if a customer ticket attached a log file with only that line.

fengjica (dismissed):

src/utils/libfabric/libfabric_topology.cpp
akkart-aws: I have added two distinct log messages for the two cases.

fengjica (approved):

test/unit/utils/libfabric/libfabric_topology_test.cpp
```diff
@@ -44,14 +44,31 @@ main() {
...
+
+                char pci_bus_id[32];
+                snprintf(pci_bus_id,
+                         sizeof(pci_bus_id),
+                         "%04x:%02x:%02x.0",
+                         prop.pciDomainID,
+                         prop.pciBusID,
+                         prop.pciDeviceID);
+
+                auto gpu_devices = topology.getEfaDevicesForGPUPci(pci_bus_id);
```
fengjica, line 60: Is it better to call this "efa_devices"?
And can we assert here it's not empty?

fengjica (approved):

test/unit/utils/libfabric/libfabric_topology_test.cpp
akkart-aws: Yes. Agreed . I will resolve this conversation and address this in a separate PR.

PR #1155, opened 2025-12-24
nixl_ep: Migrate elastic.py to TCPStore

itayalroy:
This commit migrates elastic.py to TCPStore-based metadata exchange
instead of ETCD.

Note that this TCPStore instance can be re-used for rank management
(instead of the custom TCP based implementation we currently have)
but distributed rank management performs worse in scale (tested
with 64 concurrent get_rank() calls) so rank server will stay
for now.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2025 DeepSeek
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * This file incorporates material from the DeepSeek project, licensed under the MIT License.
  * The modifications made by NVIDIA are licensed under the Apache License, Version 2.0.
--- a/examples/device/ep/nixl_ep/buffer.py
+++ b/examples/device/ep/nixl_ep/buffer.py
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 DeepSeek
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # This file incorporates material from the DeepSeek project, licensed under the MIT License.
 # The modifications made by NVIDIA are licensed under the Apache License, Version 2.0.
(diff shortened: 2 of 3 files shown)


itayalroy:
@2dm @RoeyAzran1992 @ebarilanM to review

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -482,7 +482,7 @@ def _fetch_remote_metadata_from_tcp_store(self, remote_ranks: List[int]):
             f"NIXL_EP/{rank}" for rank in remote_ranks if rank != self.rank
         ]
         if remote_md_keys:
-            self.tcp_store_group.wait(remote_md_keys, timedelta(seconds=30))
+            self.tcp_store_group.wait(remote_md_keys, timedelta(seconds=300))
 
         remote_mds = []
         for rank in remote_ranks:
```
RoeyAzran1992: 2 things that are not P0 IMO:
1) Parallelize the get MD operations. Might be useful when scaling in a large number of ranks.
2) Adding group_id to the key, so one TCPStore server may serve many groups.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -233,7 +233,7 @@ void Buffer::destroy() {
     cudaFree(counters_buffer_ptr);
     cudaFree(rdma_buffer_ptr);
 
-    if (nixl_agent_info and nixl_agent_info->agent != nullptr) {
+    if (nixl_agent_info and nixl_agent_info->agent != nullptr and getenv("NIXL_ETCD_ENDPOINTS")) {
```
RoeyAzran1992, line 236: nothing too critical, but I think it would be cleaner to have a runtime field similar to "boot sideband_md_exchange" (initialized and connect_ranks) to rely on for identifing usage of etcd.

RoeyAzran1992 (approved):
Overall looks great, I suggested a few changes that should not gate the PR.

examples/device/ep/nixl_ep/buffer.py
itayalroy: 1. Actually TCPStore has a `multi_get` operation which fetches several keys in a single round trip, I think this is the best we can do here (since we still have to get all keys before we call `nixl_ep_cpp`'s `connect_ranks`) so following this comment I updated the PR accordingly.
2. We don't really have a group id and not sure I want to introduce it, since I can't really think of any scenario in which a user should have multiple instances of `nixl_ep.Buffer` for the same group. We are using a `NIXL_EP/` prefix in TCPStore to not interfere with other applications

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Currently the only NIXL based MD exchange method we support in nixl ep is ETCD so I believe we can rely on this env var, once we support additional NIXL native MD exchange methods (NIXL TCP for example) we'll probably have to introduce this `sideband_md_exchange` or similar

brminich (approved):

PR #1156, opened 2025-12-26
examples: async isend/irecv semantics over NIXL

itayalroy:
Add example demonstrating MPI-like isend/irecv semantics over NIXL.

Uses sender-initiated WRITE transfers with notification-based coordination for matching send/recv pairs by tag and sequence number.

Can serve as a reference for applications using isend/irecv semantics, such as weight transfer in vLLM.


brminich:
@itayalroy CI error is relevant

aranadive:
@itayalroy @eranrs Please fix error: 
examples/python/isend_irecv.py:223: error: Incompatible types in assignment (expression has type "bytes", variable has type "tuple[Any, ...] | None")  [assignment]
Found 1 error in 1 file (checked 1 source file)

examples/python/isend_irecv.py
```diff
@@ -0,0 +1,367 @@
...
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    logger.info("=" * 60)
+    logger.info("NIXL isend/irecv Example")
+    logger.info("=" * 60)
+
+    listen_port = args.port if args.mode == "receiver" else 0
+    config = nixl_agent_config(True, True, listen_port)
```
aranadive, line 307: Modify to take backend name as well.

PR #1170, opened 2026-01-05
Clang: Use CommentsPragmas

aranadive:
## What?
Use commentPragmas in clang

## Why?
Adding the new copyright year in the header increases the line length beyond the 100 limit. Clang formatter will wrap the line which shouldnt be done for the copyright header - it should be on one line.

## How?
Add CommentPragmas to ignore the SPDX lines.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -38,7 +38,7 @@ BreakConstructorInitializers: BeforeColon
 BreakStringLiterals: true
 ColumnLimit: 100
 # Ignore formatting for lines starting with 'SPDX'
-CommentPragmas: '^ SPDX'
+CommentPragmas: '^ *SPDX'
 CompactNamespaces: false
 ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4


.clang-format
```diff
@@ -36,7 +36,7 @@ BreakBeforeBinaryOperators: None
 BreakBeforeTernaryOperators: false
 BreakConstructorInitializers: BeforeColon
 BreakStringLiterals: true
-ColumnLimit: 100
+ColumnLimit: 120
```
ovidiusm: I suggest trying first CommentPragmas:

```suggestion
ColumnLimit: 100
# Ignore formatting for lines starting with 'SPDX'
CommentPragmas: '^ SPDX'
```

https://clang.llvm.org/docs/ClangFormatStyleOptions.html

aranadive: Sure.

.clang-format
```diff
@@ -37,6 +37,8 @@ BreakBeforeTernaryOperators: false
 BreakConstructorInitializers: BeforeColon
 BreakStringLiterals: true
 ColumnLimit: 100
+# Ignore formatting for lines starting with 'SPDX'
+CommentPragmas: '^ SPDX'
```
ovidiusm:
```suggestion
CommentPragmas: '^ *SPDX'
```

This matches correctly *.cpp files

PR #1171, opened 2026-01-05
Plugins/UCX: Fix copyright years

aranadive:
Update copyright years for UCX plugin.


ovidiusm (approved):

ovidiusm (approved):

brminich:
why is it needed, given that there is no change in these files yet?

brminich (changes requested):
why needed?

aranadive:
> why needed?

These were merged as part of PR 946 after I updated the CI copyright check.

PR #1173, opened 2026-01-06
[WIP][DNM] NIXL_EP: single handle for rdma write and amo

jeynmann:
## What?
Combined handle for rdma write and amo.
Based on this [MR#4](https://gitlab-master.nvidia.com/adv-dev-team/ai/elastic-ep/nixl_ep/-/merge_requests/4/commits)

## Why?
Take use of ucx device channel api.
Reduce memory consumption.

## How?
Add offset info instead of multiple handlers.
Select channel based on local expert id.


itayalroy:
please rebase it over latest main branch

PR #1175, opened 2026-01-06
nixl_ep: Move to UCX multi-channel API

itayalroy:
Move to UCX multi-channel API instead of creating multiple workers.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -112,10 +112,10 @@ static ino_t ipc_namespace_inode_get() {
     return st.st_ino;
 }
 
-void Buffer::update_memory_buffers(int num_ranks, int64_t num_rdma_bytes)
+void Buffer::update_memory_buffers(int num_ranks, int num_experts_per_rank, int64_t num_rdma_bytes)
 {
     if (!available) {
-        init(num_ranks, num_rdma_bytes);
+        init(num_ranks, num_experts_per_rank, num_rdma_bytes);
         available = true;
     } else {
         throw std::runtime_error("Multiple calls to update_memory_buffers are not supported");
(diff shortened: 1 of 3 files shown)


examples/device/ep/nixl_ep/buffer.py
```diff
@@ -460,6 +460,7 @@ def update_memory_buffers(
         self.group_size = num_ranks
         self.num_rdma_bytes = num_rdma_bytes
         os.environ["NIXL_EP_NUM_CHANNELS"] = str(num_experts_per_rank)
+        os.environ["UCX_RC_GDA_NUM_CHANNELS"] = str(num_experts_per_rank)
```
ebarilanM: lets allow user provide num_channels and set the default according num expert. so we can test it without touching the code

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -460,6 +460,7 @@ def update_memory_buffers(
         self.group_size = num_ranks
         self.num_rdma_bytes = num_rdma_bytes
         os.environ["NIXL_EP_NUM_CHANNELS"] = str(num_experts_per_rank)
```
ebarilanM: why need both? it is still in used?

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -733,6 +739,8 @@ void Buffer::_nixl_ep_gpu_ctx_update() {
     nixl_ctx->gpu[1].local_barrier_buffer = sync_buffer_ptr;
     nixl_ctx->gpu[0].num_ranks = max_num_ranks;
     nixl_ctx->gpu[1].num_ranks = max_num_ranks;
+    nixl_ctx->gpu[0].num_channels = env_num_channels;
```
ebarilanM: not env_ prefix i want that someday UCX will support it per ep

ebarilanM (changes requested):
lets not overwrite and allow channel to be controlled easily by user

examples/device/ep/nixl_ep/buffer.py
itayalroy: This env var was abused in order to pass max_experts_per_rank to cpp, I removed it and added an actual `num_experts_per_rank` param to the cpp api

PR #1189, opened 2026-01-09
nixl_ep: Fix barrier with rank connect/disconnect

itayalroy:
The previous barrier implementation assumed all
ranks participate in the same number of barriers,
using a single global counter to verify remote ranks
arrived at the barrier.

With dynamic rank addition/removal, ranks can have
different barrier counts. Therefore- switch to per-pair
barrier counters.


RoeyAzran1992 (approved):
looks good to me.

brminich (approved):

2dm:
Looks good. I think we should consider fixing the previous single-value barrier usage that was broken, as it makes it easier to identify and debug barrier issues. Not a big deal and other than that, this is a good solution.

2dm (approved):

PR #1190, opened 2026-01-09
releasing GIL for 'postXferReq' so it doesn't block other copies

ishaan-shivhare:
## What?
Releasing GIL for postXferReq so that it doesn't block other copies

### DIFF
--- a/src/bindings/python/nixl_bindings.cpp
+++ b/src/bindings/python/nixl_bindings.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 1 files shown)


src/bindings/python/nixl_bindings.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
brminich:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

ovidiusm (dismissed):

brminich:
@ishaan-shivhare can you pls rebase with main to pick up CI fixes from https://github.com/ai-dynamo/nixl/pull/1197?

ishaan-shivhare:
looks like this patch has been added to main, thanks for your help! Will build from trunk now

PR #1192, opened 2026-01-12
benchmark/nixlbench: Add --filenames param

omer-b1:
Add an optional parameter of a comma-separated list of filenames for backend storage.


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aranadive: Should be 2025-2026

PR #1200, opened 2026-01-13
nixl_ep: eliminate channels modulo

jeynmann:
## What?
- Eliminate the #channels modulo from nixl_ep.

## Why?
UCX already do modulo operation for channel id.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -672,8 +672,6 @@ std::string Buffer::get_local_metadata() const {
 }
 
 void Buffer::_nixl_ep_gpu_ctx_update() {
-    int num_ucx_channels = max_experts_per_rank;
-
     /* Initialize local counter arrays */
     nixl_ctx->gpu[0].local_counters = counters_buffer_ptr;
     nixl_ctx->gpu[1].local_counters = counters_buffer_ptr + max_num_ranks * max_experts_per_rank;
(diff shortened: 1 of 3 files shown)


examples/device/ep/csrc/nixl_ep.cpp
itayalroy: is this necesarry?

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -787,6 +786,7 @@ void Buffer::_nixl_agent_init() {
     // Set UCX-specific parameters
     init_params["ucx_error_handling_mode"] = "none";
     init_params["num_workers"] = std::to_string(1);
+    init_params["engine_config"] = "RC_GDA_NUM_CHANNELS=4";
```
itayalroy: should be configurable, not hard coded

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -459,7 +459,6 @@ def update_memory_buffers(
         """
         self.group_size = num_ranks
         self.num_rdma_bytes = num_rdma_bytes
-        os.environ.setdefault("UCX_RC_GDA_NUM_CHANNELS", str(num_experts_per_rank))
```
itayalroy, line 462: this should stay. If the user did not configure num channels we can discuss the default value, but it for sure should not be a hardcoded value

PR #1201, opened 2026-01-13
UCX: fix config parsing

jeynmann:
## What?
Fix ucx config parsing.

## Why?
`string_view` doesn't include end char '\0', which result in config modify not work.

## How?
Use `string` instead of `string_view`.

### DIFF
--- a/.ci/jenkins/lib/build-matrix.yaml
+++ b/.ci/jenkins/lib/build-matrix.yaml
@@ -42,7 +42,7 @@ env:
   TEST_TIMEOUT: 30
   UCX_TLS: "^shm"
   STORAGE_DRIVER: 'overlay'
-  CI_IMAGE_TAG: "20251218-1"
+  CI_IMAGE_TAG: "20260129-1"
 
 
 runs_on_dockers:
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -30,7 +30,7 @@ env:
   NIXL_INSTALL_DIR: /opt/nixl
   TEST_TIMEOUT: 30
   NPROC: 32
-  CI_IMAGE_TAG: "20251218-1"
+  CI_IMAGE_TAG: "20260129-1"
 
 docker_opt: "-e NPROC -e EXECUTOR_NUMBER --security-opt seccomp=unconfined --security-opt apparmor=unconfined --shm-size=1G --ulimit nofile=65535:65535 --ulimit stack=67108864 --ulimit memlock=-1:-1 --network=host --ipc=host --cap-add=SYS_PTRACE --cap-add=SYS_ADMIN --privileged --gpus all --device=/dev/infiniband --device=/dev/gdrdrv --entrypoint=''"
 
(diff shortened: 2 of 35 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -461,11 +461,10 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
     std::stringstream stream(engine_config);
 
     while (std::getline(stream, elem, ',')) {
-        std::string_view elem_view = elem;
-        size_t pos = elem_view.find('=');
+        size_t pos = elem.find('=');
```
brminich: imo it is better to modify `config::modifyAlways`, so that such issue would not occur in another place

src/utils/ucx/config.cpp
```diff
@@ -35,12 +35,14 @@ config::modify (std::string_view key, std::string_view value) const {
 
 void
 config::modifyAlways (std::string_view key, std::string_view value) const {
-    const auto status = ucp_config_modify (config_.get(), key.data(), value.data());
+    std::string key_str = std::string(key);
```
ColinNV: The function arguments might as well be `const std::string&`?

src/utils/ucx/config.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
brminich:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

src/utils/ucx/config.cpp
```diff
@@ -35,12 +35,14 @@ config::modify (std::string_view key, std::string_view value) const {
 
 void
 config::modifyAlways (std::string_view key, std::string_view value) const {
-    const auto status = ucp_config_modify (config_.get(), key.data(), value.data());
+    const std::string& key_str = std::string(key);
+    const std::string& value_str = std::string(value);
```
brminich: seems this would be clearer with the same effect:
```suggestion
    std::string key_str(key);
    std::string value_str(value);
```

src/utils/ucx/config.cpp
```diff
@@ -35,12 +35,14 @@ config::modify (std::string_view key, std::string_view value) const {
 
 void
 config::modifyAlways (std::string_view key, std::string_view value) const {
```
iyastreb: Then better option is to change params list to be:
```
config::modifyAlways (const std::string &key, const std::string &value) const {

Comment (jeynmann) [reply src/utils/ucx/config.cpp]: The constructor doesn't allow implicit conversion.
```c++
template< class StringViewLike >
explicit basic_string( const StringViewLike& t,
                       const Allocator& alloc = Allocator() );
```

brminich:
@jeynmann can you pls resolve conflict

brminich (approved):

PR #1206, opened 2026-01-14
UCX: add rc gda num channels config

jeynmann:
## What?
Add rc gda num channels for ucx banckend.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -282,17 +282,21 @@ if ucx_dep.found() and cuda_dep.found() and nvcc_prog.found()
     endif
 
     have_host_side_v2 = cpp.has_function('ucp_device_remote_mem_list_create', dependencies: ucx_dep)
-    if have_host_side_v2
-        add_project_arguments('-DHAVE_UCX_GPU_DEVICE_API_V2', language: ['cpp'])
+    if have_gpu_side and have_host_side_v2
+        ucx_gpu_device_api_v2_available = true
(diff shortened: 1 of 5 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -409,6 +409,7 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
                                bool prog_thread,
                                unsigned long num_workers,
                                nixl_thread_sync_t sync_mode,
+                               size_t num_channels,
```
yosefe: maybe the backend parameter should be "num_device_channels" because it's only for device API

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -719,6 +719,10 @@ void Buffer::_nixl_agent_init() {
     }
 
     // Set UCX-specific parameters
+    const char* num_channels_env = std::getenv("NIXL_EP_NUM_CHANNELS");
```
itayalroy, line 722: can you explicitly set the default to 4 in nixl ep? I don't want nixl ep to count on the global default which can be changed silently in the future, this is application specific

itayalroy:
nixl ep part looks good

itayalroy:
@yosefe @brminich why are we not merging it?

brminich:
@jeynmann pls resolve merge conflict

itayalroy:
@jeynmann please fix CI issues, they are relevant

PR #1209, opened 2026-01-14
Refactor plugin manager to improve memory safety and string handling

ovidiusm:
## What?

- Fixes potential use-after-free in `nixlBackendStaticPluginInfo` where `const char*` was pointing to temporary string data.
- Ensures null-termination for file paths passed to logging and `dlsym`/`dlopen`.
- Simplifies `std::filesystem::path` construction by removing unsafe `.data()` calls.
- Resolves SonarQube warnings regarding memory safety and buffer overreads.

## Why?
SonarQube warnings and subsequent code review identified the above issues.


src/core/nixl_plugin_manager.cpp
```diff
@@ -168,12 +167,12 @@ nixlTelemetryPluginHandle::getVersion() const {
...
 // Telemetry plugin loader
 std::shared_ptr<const nixlPluginHandle>
-telemetryLoader(void *handle, const std::string_view &plugin_path) {
+telemetryLoader(void *handle, const std::string &plugin_path) {
     // Get the initialization function
     typedef nixlTelemetryPlugin *(*init_func_t)();
     init_func_t init = (init_func_t)dlsym(handle, "nixl_telemetry_plugin_init");
     if (!init) {
-        NIXL_ERROR << "Failed to find nixl_telemetry_plugin_init in " << plugin_path.data() << ": "
+        NIXL_ERROR << "Failed to find nixl_telemetry_plugin_init in " << plugin_path << ": "
```
rakhmets, line 175: Does removing `.data()` fix the issue? We don't need `std::string` in this method.

src/core/nixl_plugin_manager.cpp
```diff
@@ -476,8 +475,8 @@ nixlPluginManager::discoverTelemetryPlugin(const std::string &filename) {
 }
 
 void
-nixlPluginManager::discoverPluginsFromDir(const std::string_view &dirpath) {
-    std::filesystem::path dir_path(dirpath.data());
+nixlPluginManager::discoverPluginsFromDir(const std::string &dirpath) {
```
ColinNV: Or `const std::filesystem::path& dirpath`?

src/core/nixl_plugin_manager.cpp
ovidiusm: It would, but the caller is just sending a path as a const string reference in a one time initialization on startup, I don't think there is any value to use string_view here

ovidiusm: Would you be ok with the latest version of the code?

PR #1216, opened 2026-01-16
fix(libfabric): Apply CUDA context inside progress thread loop

dmvevents:
## Summary

Fixes GPU memory operation failures in the libfabric backend by applying the CUDA context inside the progress thread loop instead of outside it.

## Problem

The libfabric backend's progress thread currently doesn't properly handle CUDA context changes during operation. This causes failures when:

1. Progress thread starts in constructor BEFORE `registerMem()` is called
2. `pthrCudaCtx_` is NULL until first GPU memory registration
3. Unlike the UCX backend, libfabric doesn't restart the thread on context change

## Solution

Call `vramApplyCtx()` at the start of each progress loop iteration when `cuda_addr_wa_` is set.

```cpp
while (!progress_thread_stop_.load()) {
#ifdef HAVE_CUDA
    if (cuda_addr_wa_) {
        vramApplyCtx();
    }
#endif
    // ... rest of progress loop
}
```

This is safe because:
- `vramApplyCtx()` checks `cuda_addr_wa_` internally before doing anything
- Performance overhead is ~2us per iteration, acceptable for KV cache latency
- The UCX backend already handles this by restarting the progress thread

## Testing

- Tested on AWS p5.48xlarge (8x H100, 32 EFA)
- Verified with TRT-LLM disaggregated inference
- Verified with vLLM disaggregated inference
- No GPU memory errors or CUDA context issues after fix

## Test Plan

- [x] TRT-LLM disaggregated inference with GPU Direct RDMA
- [x] vLLM disaggregated inference with GPU Direct RDMA
- [x] Multi-rail (32 EFA) configuration
- [x] No performance regression observed

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1452,6 +1452,19 @@ nixlLibfabricEngine::progressThread() {
     NIXL_DEBUG << "PT: Thread started successfully for data rails only";
     // Main progress loop - continuously process completions only on data rails
     while (!progress_thread_stop_.load()) {
+#ifdef HAVE_CUDA
+        // v133b: Apply CUDA context INSIDE the loop before any GPU memory operations
+        // CRITICAL: Must be inside loop because:
+        //   1. Progress thread starts in constructor BEFORE registerMem() is called
+        //   2. pthrCudaCtx_ is NULL until first GPU memory registration
+        //   3. UCX backend restarts thread on context change; libfabric doesn't
+        // Safe to call repeatedly - vramApplyCtx() checks cuda_addr_wa_ first
+        // Performance: ~2us overhead per iteration, acceptable for KV cache latency
+        if (cuda_addr_wa_) {
+            vramApplyCtx();
+        }
+#endif
+
         // Process completions only on data rails (non-blocking)
         bool any_completions = false;
         nixl_status_t status = rail_manager.progressActiveDataRails();


src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1452,6 +1452,19 @@ nixlLibfabricEngine::progressThread() {
     NIXL_DEBUG << "PT: Thread started successfully for data rails only";
     // Main progress loop - continuously process completions only on data rails
     while (!progress_thread_stop_.load()) {
+#ifdef HAVE_CUDA
+        // v133b: Apply CUDA context INSIDE the loop before any GPU memory operations
+        // CRITICAL: Must be inside loop because:
+        //   1. Progress thread starts in constructor BEFORE registerMem() is called
+        //   2. pthrCudaCtx_ is NULL until first GPU memory registration
+        //   3. UCX backend restarts thread on context change; libfabric doesn't
```
brminich, line 1460: cuda dependency is removed from UCX backend and is handled inside UCX now, probably this comment is not relevant anymore

dmvevents:
## Closing Per Maintainer Clarification

@amitrad-aws - Thank you for clarifying the progress thread design.

You're correct that `fi_cq_read()` doesn't access GPU memory directly. Our concern was about thread-local CUDA context state, but after your explanation of the architecture, we understand the current design doesn't require it.

**Closing this PR.**

If we encounter CUDA context issues in production, we'll:
1. Capture complete stack traces with debug symbols
2. Verify the crash is in the progress thread, not the data posting path
3. Open a new issue with diagnostic data

Thank you for your time reviewing this.

PR #1217, opened 2026-01-16
fix(libfabric): Add EFA domain validation to detect RDM init failures

dmvevents:
## Summary

Adds validation after `fi_domain()` to detect when the EFA RDM layer fails to initialize properly. This prevents cryptic segfaults on certain glibc versions by catching the issue early with clear error messages.

## Problem

On Ubuntu 24.04.3 (glibc 2.39-0ubuntu8.6), EFA's `fi_domain()` can return success even when the internal RDM layer initialization fails with `efa_domain_init_rdm err:28`. The domain appears valid but is corrupted internally. Subsequent operations (like `fi_cq_open`) crash with:

```
segfault at 18 ip 00007f... sp 00007f... error 4 in libfabric.so
```

This is extremely difficult to debug because the segfault occurs deep in libfabric with no useful stack trace.

## Solution

1. **Domain Validation**: After `fi_domain()` succeeds, create a minimal test CQ to verify the domain is actually functional
2. **Clear Error Messages**: If validation fails, log detailed troubleshooting information:
   - Current glibc version
   - Specific error codes
   - Possible causes and workarounds
3. **Environment Variables**: Add escape hatches for advanced users:
   - `NIXL_SKIP_RDM_VALIDATION=1`: Skip validation (may crash later)
   - `NIXL_FORCE_DISABLE_GPU_RDMA=1`: Force disable GPU Direct RDMA at startup

### libfabric_rail.cpp
```cpp
// After fi_domain() succeeds, validate with test CQ
struct fi_cq_attr test_cq_attr = {};
test_cq_attr.format = FI_CQ_FORMAT_DATA;
test_cq_attr.wait_obj = FI_WAIT_NONE;
test_cq_attr.size = 1;

int validate_ret = fi_cq_open(domain, &test_cq_attr, &test_cq, NULL);
if (validate_ret != 0) {
    // Log detailed error and cleanup
    // Throw with actionable error message
}
```

### libfabric_backend.cpp
```cpp
// Check NIXL_FORCE_DISABLE_GPU_RDMA at engine startup
const char* force_disable_rdma = std::getenv("NIXL_FORCE_DISABLE_GPU_RDMA");
if (force_disable_rdma && std::string(force_disable_rdma) == "1") {
    setenv("FI_EFA_USE_DEVICE_RDMA", "0", 1);
}
```

## Testing

| Environment | glibc | Result |
|-------------|-------|--------|
| Ubuntu 24.04.3 | 2.39-0ubuntu8.6 | Validation catches issue, clear error |
| Ubuntu 22.04 | 2.35 | Validation passes, works normally |
| p5.48xlarge | 32 EFA | Multi-rail validated |

## Test Plan

- [x] Validation correctly detects corrupted domains
- [x] Clear error messages with glibc version logged
- [x] NIXL_SKIP_RDM_VALIDATION=1 bypasses check
- [x] NIXL_FORCE_DISABLE_GPU_RDMA=1 disables GPU RDMA
- [x] No performance impact on healthy systems
- [x] Works with TRT-LLM and vLLM

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+#include <gnu/libc-version.h>  // For glibc version detection
 #include "libfabric_backend.h"
 #include "serdes/serdes.h"
 #include "common/nixl_log.h"
@@ -294,7 +295,18 @@ nixlLibfabricEngine::nixlLibfabricEngine(const nixlBackendInitParams *init_param
       rail_manager(NIXL_LIBFABRIC_DEFAULT_STRIPING_THRESHOLD),
       runtime_(FI_HMEM_SYSTEM) {
 
-    NIXL_DEBUG << "Initializing Libfabric Backend";
+    // NIXL_FORCE_DISABLE_GPU_RDMA: Environment variable to force disable GPU Direct RDMA
+    // This is a workaround for glibc incompatibility issues with EFA RDM layer
+    const char* force_disable_rdma = std::getenv("NIXL_FORCE_DISABLE_GPU_RDMA");
+    if (force_disable_rdma && std::string(force_disable_rdma) == "1") {
+        NIXL_WARN << "NIXL_FORCE_DISABLE_GPU_RDMA=1 detected, setting FI_EFA_USE_DEVICE_RDMA=0";
+        setenv("FI_EFA_USE_DEVICE_RDMA", "0", 1);
+    }
+
+    // Log glibc version at engine startup for debugging
+    NIXL_INFO << "NIXL LibfabricEngine starting, glibc version: " << gnu_get_libc_version();
+
+    NIXL_DEBUG << "Initializing Libfabric Backend with GPU Support";
 
     // Query system runtime type from rail manager (determined once at topology discovery)
     runtime_ = rail_manager.getRuntime();
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-#include <gnu/libc-version.h>  // For glibc version detection
+#include <gnu/libc-version.h> // For glibc version detection
 #include "libfabric_backend.h"
 #include "serdes/serdes.h"
 #include "common/nixl_log.h"
@@ -297,7 +297,7 @@ nixlLibfabricEngine::nixlLibfabricEngine(const nixlBackendInitParams *init_param
 
     // NIXL_FORCE_DISABLE_GPU_RDMA: Environment variable to force disable GPU Direct RDMA
     // This is a workaround for glibc incompatibility issues with EFA RDM layer
-    const char* force_disable_rdma = std::getenv("NIXL_FORCE_DISABLE_GPU_RDMA");
+    const char *force_disable_rdma = std::getenv("NIXL_FORCE_DISABLE_GPU_RDMA");
     if (force_disable_rdma && std::string(force_disable_rdma) == "1") {
         NIXL_WARN << "NIXL_FORCE_DISABLE_GPU_RDMA=1 detected, setting FI_EFA_USE_DEVICE_RDMA=0";
         setenv("FI_EFA_USE_DEVICE_RDMA", "0", 1);
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -20,7 +20,7 @@
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
 #include "libfabric_common.h"
-#include <gnu/libc-version.h>  // For glibc version detection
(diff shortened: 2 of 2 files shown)


dmvevents:
## Additional Context: Production Bug on Ubuntu 24.04

@fengjica @amitrad-aws - We'd like to provide more context on why this validation is needed, following up on the earlier feedback that validation already exists.

### The Specific Bug This Catches

**Environment:** Ubuntu 24.04.3 LTS with glibc 2.39-0ubuntu8.6

**What Happens:**
1. `fi_domain()` returns `0` (success) ✓
2. Domain handle is non-null ✓  
3. First `fi_cq_open()` triggers segfault at offset 0x18

**Error Log from Production:**
```
[libfabric_backend.cpp:471] fi_domain() returned: 0 (SUCCESS)
[libfabric_backend.cpp:480] Creating completion queue...
Segmentation fault (core dumped)
efa:core:efa_domain_open():420<error> Invalid RDM domain initialization (errno 28)
```

### Why Current Validation Doesn't Catch This

**Current code (libfabric_backend.cpp:471):**
```cpp
int ret = fi_domain(...);
if (ret != 0) {
    return NIXL_ERR_BACKEND;
}
// Assumes domain is usable
```

**The issue:** `fi_domain()` succeeds (ret == 0) but the EFA RDM layer initialization failed internally. The corrupted domain passes the return code check but crashes on first use.

### What This PR Does

Adds a **functional test** of the domain immediately after creation:

```cpp
// Validate domain is actually usable with a test CQ
struct fid_cq *test_cq = nullptr;
struct fi_cq_attr test_attr = {.size = 1, .format = FI_CQ_FORMAT_DATA};
int test_ret = fi_cq_open(domain_, &test_attr, &test_cq, nullptr);

if (test_ret != 0) {
    NIXL_ERROR << "Domain validation failed: " << fi_strerror(-test_ret);
    NIXL_ERROR << "This may indicate glibc compatibility issues";
    // Log glibc version, suggest workarounds
    return NIXL_ERR_BACKEND;
}
fi_close(&test_cq->fid);  // Clean up test CQ
```

### Value Proposition

**Benefit:** Fail fast with clear error during initialization, not 60 seconds later during first transfer

**Diagnostics:** Log glibc version and suggest specific environment variable workarounds

**Safety:** Validates assumption that successful `fi_domain()` → usable domain

### Root Cause (Suspected)

From libfabric EFA provider logs, the RDM layer initialization can fail due to:
- glibc 2.39 threading changes affecting libfabric internal state
- Incompatibility with specific EFA firmware versions
- Race condition in EFA domain setup on certain kernel versions

The key point: **libfabric doesn't propagate this error** to `fi_domain()` return code.

### Alternative Approaches

If you prefer not to add test CQ creation, we could:

**Option A:** Document this as a known glibc 2.39 issue
- Add note to README about Ubuntu 24.04.3 compatibility
- Document workaround: `export FI_EFA_USE_DEVICE_RDMA=0`

**Option B:** Add runtime glibc version detection
```cpp
if (glibc_version >= "2.39") {
    NIXL_WARN << "glibc 2.39 detected, may encounter EFA RDM issues";
}
```

**Option C:** Keep this PR but make validation opt-in
- Add `NIXL_VALIDATE_DOMAIN=1` environment variable
- Only run test CQ if explicitly enabled

### Question for Maintainers

Which approach would you prefer:
1. Test CQ validation (this PR)
2. Documentation only (Option A)
3. Runtime warning (Option B)  
4. Opt-in validation (Option C)
5. Close this PR and track as known limitation

We're flexible and want to align with the project's philosophy on handling platform-specific edge cases.

Thank you for considering this feedback.

dmvevents:
Hi team — checking in on this PR. The EFA domain validation catches a real failure mode we hit on Ubuntu 24.04 where `fi_domain` initialization can fail silently, leading to crashes later in rail creation. The fix adds a test CQ creation after `fi_domain()` to verify the RDM layer initialized properly.

This appears to still be unique — I don't see it covered by recent merges. Is there still interest, or has the initialization path changed? Happy to rebase against main if needed.

dmvevents:
Hi @fengjica @amitrad-aws — seeking guidance on this PR.

Since this PR was filed, [#1272 "libfabric: CQ batch reads and threading model"](https://github.com/ai-dynamo/nixl/pull/1272) merged (Feb 3) and covers a large portion of what this PR proposed. The branch is also currently in conflict with `main`.

Three paths forward from our side; happy to take whichever you prefer:

1. **Close** — the core improvement is upstream via #1272 and the deltas aren't worth carrying separately.
2. **Rebase and slim down** — keep only the specific sub-elements from this PR that #1272 didn't address, pushed as a narrower, single-concern PR.
3. **Keep open** if there's a specific piece of the original approach you'd like us to rebase onto the post-#1272 codebase.

Please let us know. Default to close if we don't hear back in the next few days — we don't want to leave stale PRs on your queue.

dmvevents:
Closing per our default-to-close commitment in the May 5 guidance ask — the core optimization is covered by [#1272](https://github.com/ai-dynamo/nixl/pull/1272) and we haven't heard objections from maintainers. Happy to reopen if any piece is still wanted as a standalone PR; otherwise this reduces the open-PR queue.

PR #1219, opened 2026-01-16
fix(libfabric): Increase control buffer size for large model metadata

dmvevents:
## Summary

Increases control pool buffer size from 8KB to 512KB to support metadata transfers for large models.

## Problem

Large models with many KV cache segments generate metadata that exceeds the current 8KB buffer limit, causing transfer failures during disaggregated inference.

## Solution

Increase `SEND_RECV_BUFFER_SIZE` to 525,312 bytes (512KB + 1KB header) to accommodate large model metadata.

## Testing

- Tested with Qwen3-32B (TP8) - metadata fits in new buffer
- Tested with smaller models - no regression
- Memory overhead acceptable for inference workloads

### DIFF
--- a/src/utils/libfabric/libfabric_common.h
+++ b/src/utils/libfabric/libfabric_common.h
@@ -47,7 +47,7 @@
 // Request pool configuration constants
 #define NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL 4096 // SEND/RECV operations (1:1 with buffers)
 #define NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL 1024 // WRITE/read operations (no buffers)
-#define NIXL_LIBFABRIC_SEND_RECV_BUFFER_SIZE 525312  // v95: 512KB for large models
+#define NIXL_LIBFABRIC_SEND_RECV_BUFFER_SIZE 525312 // v95: 512KB for large models
 #define NIXL_LIBFABRIC_RECV_POOL_SIZE 1024 // Number of recv requests to pre-post per rail
 
 // Retry configuration constants


src/utils/libfabric/libfabric_common.h
```diff
@@ -47,7 +47,7 @@
 // Request pool configuration constants
 #define NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL 4096 // SEND/RECV operations (1:1 with buffers)
 #define NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL 1024 // WRITE/read operations (no buffers)
-#define NIXL_LIBFABRIC_SEND_RECV_BUFFER_SIZE 8192
+#define NIXL_LIBFABRIC_SEND_RECV_BUFFER_SIZE 525312 // v95: 512KB for large models
```
akkart-aws: Its not feasible to use a send_recv_buffer_size of 8192 because the actual control request pool size is 
NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL*NIXL_LIBFABRIC_SEND_RECV_BUFFER_SIZE
Ref: https://github.com/ai-dynamo/nixl/blob/19856045bbcc3c903dd716ec60b86b69b5f96c65/src/utils/libfabric/libfabric_rail.cpp#L161

Currently we are using 8K buffer size and then 4K control request per rail which is 4K * 8K = 32K

Increasing each buffer size to 512K will make the pool size grow to 2GB per process.

dmvevents:
## Closing - Duplicate Fix

@amitrad-aws - Thank you for pointing out this was already fixed in PR #1142.

We should have checked merged PRs more carefully before submitting. Closing this to reduce noise.

Apologies for the duplicate effort.

PR #1220, opened 2026-01-16
fix(core): Improve metadata handling robustness for timing races

dmvevents:
## Summary

Adds robustness fixes for metadata handling during disaggregated inference startup to handle timing races when agents publish metadata before KV cache is fully allocated.

## Problem

During disaggregated inference initialization, timing races can occur:

1. **Empty Descriptors**: Agents publish metadata before KV cache allocation completes
2. **Missing Markers**: MemSection marker not found due to partial serialization
3. **Corrupted Metadata**: importStr() fails but error not detected
4. **Permanent Failures**: Failed metadata loads are cached, preventing retry

## Fixes

| Fix | File | Change |
|-----|------|--------|
| v81 | nixl_memory_section.cpp | Continue on empty descriptor lists |
| v82 | nixl_agent.cpp | Handle missing MemSection gracefully |
| v86 | nixl_memory_section.cpp | Break on empty backend during load |
| v87 | libfabric_rail_manager.cpp | Validate importStr() return value |
| v88 | nixl_agent.cpp | Don't cache failed metadata loads |
| v89 | nixl_agent.cpp | Allow retry on loadRemoteSections failure |

### Before
```
Empty descriptors → NIXL_ERR_NOT_FOUND → Connection fails
Missing MemSection → NIXL_ERR_MISMATCH → Cached, never retries
```

### After
```
Empty descriptors → continue → Retry on next access
Missing MemSection → NIXL_SUCCESS → Retry on next access
```

## Testing

- [x] TRT-LLM disaggregated inference startup
- [x] vLLM disaggregated inference startup  
- [x] Multi-node initialization races
- [x] No regression on normal operation

## Summary by CodeRabbit

* **New Features**
  * Added GPU-transfer signaling support for enhanced GPU data-transfer workflows.

* **Bug Fixes**
  * Remote metadata loading now tolerates missing data and retries gracefully.
  * Improved handling of partial/empty descriptor updates to avoid stale state and preserve cached valid metadata after failed loads.

* **Tests**
  * Added tests validating retry convergence, empty-descriptor behavior, and cache preservation on failed loads.

### DIFF
--- a/src/utils/libfabric/libfabric_rail_manager.cpp
+++ b/src/utils/libfabric/libfabric_rail_manager.cpp
@@ -993,11 +993,7 @@ nixlLibfabricRailManager::deserializeMemoryKeys(const std::string &serialized_da
                                                 std::vector<uint64_t> &keys_out,
                                                 uint64_t &remote_addr_out) const {
     nixlSerDes ser_des;
-    // v87: Check importStr return for corrupted metadata
-    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
-        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
-        return NIXL_ERR_MISMATCH;
-    }
+    ser_des.importStr(serialized_data);
     // Load all rail keys instead of just one
     keys_out.clear();
     keys_out.reserve(num_keys);
@@ -1044,11 +1040,7 @@ nixlLibfabricRailManager::deserializeConnectionInfo(
     std::vector<std::array<char, LF_EP_NAME_MAX_LEN>> &data_endpoints_out) const {
 
     nixlSerDes ser_des;
-    // v87: Check importStr return for corrupted metadata
-    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
-        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
-        return NIXL_ERR_MISMATCH;
-    }
+    ser_des.importStr(serialized_data);
 
     // Use user prefix with standard suffixes
     std::string data_prefix = user_prefix + "_data_ep_";
--- a/src/utils/libfabric/libfabric_rail_manager.cpp
+++ b/src/utils/libfabric/libfabric_rail_manager.cpp
@@ -993,7 +993,11 @@ nixlLibfabricRailManager::deserializeMemoryKeys(const std::string &serialized_da
                                                 std::vector<uint64_t> &keys_out,
                                                 uint64_t &remote_addr_out) const {
     nixlSerDes ser_des;
-    ser_des.importStr(serialized_data);
+    // v87: Check importStr return for corrupted metadata
+    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
+        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
+        return NIXL_ERR_MISMATCH;
+    }
     // Load all rail keys instead of just one
     keys_out.clear();
     keys_out.reserve(num_keys);
@@ -1040,7 +1044,11 @@ nixlLibfabricRailManager::deserializeConnectionInfo(
     std::vector<std::array<char, LF_EP_NAME_MAX_LEN>> &data_endpoints_out) const {
 
     nixlSerDes ser_des;
-    ser_des.importStr(serialized_data);
+    // v87: Check importStr return for corrupted metadata
+    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
+        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
+        return NIXL_ERR_MISMATCH;
+    }
 
     // Use user prefix with standard suffixes
     std::string data_prefix = user_prefix + "_data_ep_";
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1528,9 +1528,12 @@ nixlAgent::loadRemoteMD (const nixl_blob_t &remote_metadata,
         return NIXL_ERR_BACKEND;
     }
 
-    if (sd.getStr("") != "MemSection") {
-        NIXL_ERROR_FUNC << "failed to deserialize remote metadata";
-        return NIXL_ERR_MISMATCH;
+    // v82+v88+v89: Handle missing MemSection without permanent caching
+    std::string marker = sd.getStr("");
+    if (marker != "MemSection") {
+        NIXL_WARN << "v82: MemSection not found (got '" << marker << "'), will retry";
+        agent_name = remote_agent;
+        return NIXL_SUCCESS;  // Allow retry
     }
 
     ret = data->loadRemoteSections(remote_agent, sd);
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -26,10 +26,8 @@
 #include "transfer_request.h"
 #include "agent_data.h"
 #include "plugin_manager.h"
-#include "common/configuration.h"
 #include "common/nixl_log.h"
 #include "common/operators.h"
-#include "common/hw_info.h"
 #include "telemetry.h"
 #include "telemetry_event.h"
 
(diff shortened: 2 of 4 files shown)


ColinNV:
Could you please update the PR so the it does not reformat code that was not changed?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -743,7 +743,11 @@ nixlLibfabricRailManager::deserializeMemoryKeys(const std::string &serialized_da
                                                 std::vector<uint64_t> &keys_out,
                                                 uint64_t &remote_addr_out) const {
     nixlSerDes ser_des;
-    ser_des.importStr(serialized_data);
+    // v87: Check importStr return for corrupted metadata
```
fengjica: I believe 'v87' is your own version control tag, is it? Should we leave them out? I even suggest the whole line comment can be omitted as the code following is self-explanatory.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -743,7 +743,11 @@ nixlLibfabricRailManager::deserializeMemoryKeys(const std::string &serialized_da
...
+    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
+        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
```
fengjica: "v87: " is unnecessary, right?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -793,7 +797,11 @@ nixlLibfabricRailManager::deserializeConnectionInfo(
     std::vector<std::array<char, LF_EP_NAME_MAX_LEN>> &control_endpoints_out) const {
 
     nixlSerDes ser_des;
-    ser_des.importStr(serialized_data);
+    // v87: Check importStr return for corrupted metadata
+    if (ser_des.importStr(serialized_data) != NIXL_SUCCESS) {
+        NIXL_ERROR << "v87: importStr failed, metadata may be corrupted";
```
fengjica: Similarly here.

fengjica (changes requested):
I only reviewed the change in libfabric_rail_manager.cpp. Mostly minor. Please see comments below. 

Thanks for the PR! I'm also wondering if you can separate the changes in this PR on libfabric/ to its own PR, because they are all on validating the serialization return status, while this PR is primarily working the race condition, which can be complicated.

dmvevents:
## PR Split Complete Per Request

@fengjica @ColinNV - Thank you for the feedback. We've split this PR as requested.

### The Split

**PR #1220 (this one) - Core Metadata Race Fixes:**
- Focus: `nixl_memory_section.cpp` and `nixl_agent.cpp` changes
- Scope: Empty descriptor handling, missing MemSection retry logic
- Key fixes:
  - v81: Allow empty descriptor lists (continue instead of error)
  - v86: Break on empty backend string (retry instead of hard error)
  - v82: Handle missing MemSection marker without permanent caching

**New PR #1270 - Libfabric Return Value Validation:**
- Focus: `libfabric_rail_manager.cpp` validation
- Scope: Check `importStr()` return values in deserialization
- Link: https://github.com/ai-dynamo/nixl/pull/1270

### Formatting Question

@ColinNV - You mentioned removing unrelated formatting. This PR currently includes:
- ~350 lines of formatting changes (spacing, line breaks)
- ~100 lines of actual logic changes (v81, v82, v86 fixes)

**Would you like us to:**
1. Create a third PR with just the formatting cleanup?
2. Revert all formatting and keep only logic changes?
3. Leave it as-is (formatting + logic together)?

We want to align with the project's preferences on formatting cleanup. Let us know what works best for review.

### Updated Commits

- `6c114e9` - Removed libfabric validation (now in #1270)
- `6e32b0d` - Updated copyright year

**This PR now focuses solely on core metadata race handling as requested.**

Thank you for the guidance on splitting this properly.

ColinNV:
Regarding the formatting, any PR should format all changed and new lines according to the coding style as per the .clang-format file, but leave the formatting of unchanged lines unchanged.

dmvevents:
Hi @ColinNV @fengjica — apologies for the delay. I'll rebase against main, apply `.clang-format` to all changed lines per the coding style (leaving unchanged lines as-is), and address the January review feedback. Will push an updated revision this week.

dmvevents:
@ColinNV — unit tests landed on the PR branch ([`e8f9db3`](https://github.com/ai-dynamo/nixl/pull/1220/commits/e8f9db3)). Three `TEST_F(MetadataExchangeTestFixture, ...)` cases, one per invariant you asked for:

1. **`RetryConvergenceAfterPartialLoad`** — first `loadRemoteMD` observes a partial view (producer still registering); `checkRemoteMD` returns `NIXL_ERR_NOT_FOUND` for the unregistered buffers. A second `loadRemoteMD` with the full view succeeds and resolves those same buffers. Retry path converges.

2. **`EmptyDescriptorSkipNoStaleState`** — an empty-descriptor `getLocalPartialMD` push is tolerated (either `NIXL_SUCCESS` with a connection-only view or an explicit failure), but the subsequent full `getLocalMD` + `loadRemoteMD` always lands cleanly and exposes every buffer. The empty push cannot poison the cache.

3. **`FailedLoadPreservesCachedMetadata`** — after a successful load, a corrupted blob fed to `loadRemoteMD` fails, and the previously cached view remains resolvable via `checkRemoteMD`. Failures are not cached and do not evict valid entries.

They reuse the existing `MetadataExchangeTestFixture` and the same `LogIgnoreGuard` pattern as the surrounding tests, so they pick up in the current meson build without extra wiring (`test/gtest/meson.build:68` already lists the file).

This closes the unit-test deliverable we committed to on Apr 13. Please ping me if you'd like any of the three invariants tightened or additional scenarios covered.

dmvevents:
### Status on the 3 critical CodeRabbit findings

Did a 3-model consensus review (GPT-5.4 + Gemini 3 Pro + Bedrock Claude) before touching anything, and all three agree this branch is not fixable in-place with a small, surgical commit. Summary of why, and what I recommend next:

**Root cause: branch is significantly behind upstream refactors.**
Upstream has merged `#1375 CORE/AGENT: Refactor` and `#1390 CORE: Refactor transfer_request.h` since this PR was opened. As a result, locally:
- `src/core/agent_data.h` and `src/core/transfer_request.h` reflect the NEW API (underscored members, `localSection_` by value, `const` members on `nixlXferReqH`, 5-arg constructor, raw-pointer `updateRequestStats` signature).
- `src/core/nixl_agent.cpp` still uses the OLD patterns (`data->name`, `data->memorySection`, zero-arg `make_unique<nixlXferReqH>()`, `handle->remoteAgent = ...` on const members, `std::unique_ptr<nixlTelemetry>&` telemetry signature).
- `ninja src/core/libnixl.so` fails with ~100 errors.

**Per finding:**
- **Line 93** (header/impl sync): this IS the compile break above. Fixing it requires adapting `nixl_agent.cpp` to the new `agent_data.h` / `transfer_request.h` API — 500+ lines of surgical refactor across `prepXferDlist`, `makeXferReq`, `createXferReq`, the listener and registration paths. That's well beyond a CodeRabbit nit and belongs to a proper rebase commit, not this PR's metadata-robustness scope.
- **Line 513** (self-remote mirror eviction): CodeRabbit's suggested diff calls `data->remoteSections[data->name]->remLocalData(descs, backend)`, but `nixlRemoteSection::remLocalData` does not exist in `src/infra/mem_section.h` (only `loadLocalData` / `loadRemoteData` are public). The legitimate concern (self-remote descriptors outliving their `metadataP` in `deregisterMem`) is real, but fixing it requires adding a new `remLocalData` method to `nixlRemoteSection` plus the 2-line call-site change. Post-refactor upstream may already address this with the new `remoteSections_` value-type layout; worth confirming against upstream before duplicating work.
- **Line 1225** (lock guard around `gpuReqToEngine`): the upstream refactor replaces `gpuReqToEngine` with `mvhToEngine`, so the specific field CodeRabbit flagged no longer exists post-rebase. The underlying "shared lock protecting a mutating `.emplace()`" concern does apply to whatever replaces it, but that's a post-rebase question.

**Recommendation (maintainers' call):**
1. Rebase this branch onto current `main` (which already has the `#1375` / `#1390` refactors). This is a large enough surgery that it probably deserves its own commit rather than being folded into this PR's history.
2. After rebase, re-run CodeRabbit. Findings #1 and #3 should dissolve; #2 can be addressed with a clean 2-line change (plus the small `remLocalData` helper if it's still needed post-refactor).
3. If preferred, I can open a companion PR that does the rebase + minimal CodeRabbit fixes as separate commits — just let me know.

Not pushing anything speculative to this branch until there's a decision on the rebase, since force-pushing a partial fix would obscure the real structural mismatch. Happy to drive whichever path the maintainers prefer.

dmvevents:
Closing this as superseded by #1625, #1690, and #1745.

This branch is too far behind the current core/agent refactors to review or land cleanly in its current form, and the intended metadata-race fixes have been carried forward in the follow-up work.

Thanks for splitting the original scope and for adding the metadata exchange tests.

PR #1221, opened 2026-01-16
fix(libfabric): Use explicit non-blocking CQ polling

dmvevents:
## Summary

Changes completion queue polling to use explicit non-blocking mode for more predictable behavior on EFA.

## Changes

- Set `cq_attr.wait_obj = FI_WAIT_NONE` (was `FI_WAIT_UNSPEC`)
- Set `blocking_cq_sread_supported = false`
- Add includes for memory barriers (preparation for future optimization)

## Rationale

`FI_WAIT_UNSPEC` allows the provider to choose the wait mechanism, which can lead to inconsistent behavior. Explicit `FI_WAIT_NONE` ensures predictable non-blocking polling across EFA driver versions.

## Testing

- Tested on AWS p5.48xlarge with EFA
- Verified with TRT-LLM and vLLM

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -16,8 +16,8 @@
  * limitations under the License.
  */
 
-#include <atomic>     // v120: for memory barriers
-#include <sched.h>    // v120: for sched_yield()
+#include <atomic> // v120: for memory barriers
+#include <sched.h> // v120: for sched_yield()
 #include "libfabric_rail.h"
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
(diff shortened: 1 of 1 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -388,7 +390,7 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
     : rail_id(id),
       device_name(device),
       provider_name(provider),
-      blocking_cq_sread_supported(true),
+      blocking_cq_sread_supported(false),  // v120: Force non-blocking polling
```
amitrad-aws: This is required to support versions of libfabric that don't support the block API

dmvevents:
Closing this PR based on maintainer feedback. The change would break compatibility with certain libfabric versions, and the team is working on removing this thread entirely. Thank you for the review.

PR #960, opened 2025-10-28
libfabric: Add CUDA memory registration support with fi_mr_regattr

yexiang-aws:
This enables proper GPU memory registration with libfabric for CUDA device memory transfers.

PR #1224, opened 2026-01-16
perf(libfabric): cache memory registrations to avoid repeated fi_mr_reg calls

dmvevents:
## Summary

This PR implements a reference-counted cache for memory registrations (`fi_mr_reg`) to avoid repeated expensive registration calls for the same GPU buffers.

### Changes

**Files Modified:**
- `src/utils/libfabric/libfabric_rail.h` - Add `MRCacheEntry` struct and cache members
- `src/utils/libfabric/libfabric_rail.cpp` - Implement cache logic in `registerMemory()`/`deregisterMemory()`

### Research Basis

Based on [He et al. "An efficient design for fast memory registration in RDMA" (JNCA 2009)](https://cis.temple.edu/~he/publications/Journals/JNCA-2009.pdf):

> "The cost of memory registration is so huge that it is much higher than the latency of RDMA operation itself. MRRC improves total cost by up to 70%."

### Implementation Details

```cpp
// MRCacheEntry structure
struct MRCacheEntry {
    struct fid_mr *mr;
    uint64_t key;
    std::atomic<uint32_t> ref_count;
    size_t length;
    nixl_mem_t mem_type;
    int gpu_id;
};

// Cache lookup in registerMemory()
uintptr_t buf_addr = reinterpret_cast<uintptr_t>(buffer);
auto it = mr_cache_.find(buf_addr);
if (it != mr_cache_.end() && entry_matches(it->second, len, mem_type, gpu_id)) {
    entry.ref_count.fetch_add(1);
    *mr = entry.mr;
    mr_cache_hits_++;
    return NIXL_SUCCESS;  // Cache hit - skip fi_mr_reg
}
// Cache miss - perform registration and add to cache
```

### Why This Might Help

- Memory registration involves kernel calls and page pinning
- KV cache buffers may be reused across requests
- Cache hit avoids expensive registration overhead

### Cache Statistics API

Included instrumentation to measure cache effectiveness:

```cpp
uint64_t getMRCacheHits() const;
uint64_t getMRCacheMisses() const;
size_t getMRCacheSize() const;
```

### Reviewer Concern Addressed

> "I don't think it'll be a perf improvement because I don't think MR happens in the hot path."

**Response:** This PR includes instrumentation to measure cache hits/misses. If cache hit rate is low, it confirms the reviewer's hypothesis and this optimization can be dropped. If hit rate is high, it validates that MR is in the hot path.

## Test Plan

- [ ] Unit tests pass
- [ ] Integration tests with EFA provider
- [ ] Check cache hit rate in logs: `grep "MRRC cache" logs`
- [ ] Benchmark comparison with baseline (results pending)

## Summary by CodeRabbit

* **New Features**
  * Added a memory-registration cache with bounded capacity, reference-counted entries, safe concurrent handling, and eviction of unused entries.
  * Added cache hit, miss, and size metrics for improved visibility into registration performance.
  * Registrations now reuse cached entries when available and validate required keys.

* **Bug Fixes**
  * Improved buffer-assignment bounds validation and diagnostic logging during pool expansion.


dmvevents:
## Proactive Bug Fix: Double-Free Vulnerability

During internal code review, we discovered a resource management bug in `deregisterMemory()` that needed fixing before merge.

### The Bug

**Location:** `libfabric_rail.cpp`, `deregisterMemory()` function (lines 1454-1485)

**Problem:**
```cpp
// Old code:
for (auto it = mr_cache_.begin(); it != mr_cache_.end(); ++it) {
    if (it->second.mr == mr) {
        // Handle ref count, may return early or break
        break;
    }
}
// Falls through even if mr NOT in cache
int ret = fi_close(&mr->fid);  // ❌ Closes even if not our MR!
```

If an MR pointer not in our cache was passed, we still called `fi_close()`, causing:
1. Potential double-free if MR was already closed elsewhere
2. Closing MRs that belong to other rails
3. Undefined behavior depending on libfabric provider

### The Fix

```cpp
// New code:
bool found_in_cache = false;
bool should_close = false;
for (auto it = mr_cache_.begin(); it != mr_cache_.end(); ++it) {
    if (it->second.mr == mr) {
        found_in_cache = true;
        uint32_t prev_count = it->second.ref_count.fetch_sub(1, ...);
        if (prev_count == 1) {
            should_close = true;
            // ...
        }
        break;
    }
}

if (!found_in_cache) {
    NIXL_ERROR << "MRRC: Attempted to deregister uncached MR";
    return NIXL_ERR_NOT_FOUND;
}

if (should_close) {
    int ret = fi_close(&mr->fid);  // ✅ Only close if we own it
    // ...
}
```

**Result:** Only close MRs that were in our cache and had their last reference decremented. Return error if trying to deregister an uncached MR.

### Testing

We verified this fix handles the case where:
1. Caller passes an MR pointer from another rail or already-closed MR
2. MR not found in cache (loop completes without finding it)
3. New behavior: Returns `NIXL_ERR_NOT_FOUND`, no close attempted
4. Old behavior: Calls `fi_close()` anyway, potential double-free

**Commit:** d3fc536

Thank you for your patience while we ensure this is production-ready.

dmvevents:
Hi team — this PR proposes a reference-counted cache for `fi_mr_reg` calls to avoid repeated registration of the same memory regions. In our disaggregated inference testing on EFA, we observed `fi_mr_reg` as a hot path (~50-100μs per call due to GDR memory pinning).

This optimization remains unique — #960 (merged Nov 2025) added CUDA memory registration support via `fi_mr_regattr`, which is the foundation this caching layer builds on. The cache itself (avoiding re-registration of already-pinned regions) is not covered by any merged PR.

Is there interest in this approach? I can rebase against current main and update the implementation to work with the threading model changes from #1272 and #1457 if so.

dmvevents:
Thanks to CodeRabbit for the careful pass — all three real findings addressed in [`9dcae64`](https://github.com/ai-dynamo/nixl/pull/1224/commits/9dcae64):

1. **Parameter-mismatch silent MR leak** — `try_emplace` is a no-op when the key already exists, so the freshly-registered MR was returned to the caller but never cached. New behavior: on mismatch with `ref_count == 0`, evict the stale entry (and only erase the slot if `fi_close` succeeds) so the later `try_emplace` actually inserts. On mismatch with an active refcount, fail with `NIXL_ERR_INVALID_PARAM` — replacing an actively-mapped MR with a different length is a caller bug we shouldn't paper over.

2. **`std::stoul` exception escape** — the env-var parse helper is now `noexcept` with `try/catch`; on `invalid_argument`, `out_of_range`, or a value of `0` we log a `WARN` and fall back to the 1024-entry default. No more chance of an exception propagating out of the magic-static initializer into `registerMemory`.

3. **Eviction erases on `fi_close` failure** — loop now continues past failed closes instead of erasing the slot, preserving our only handle to the libfabric resource. If nothing can be evicted, the freshly-allocated `mr` is closed and we return `NIXL_ERR_BACKEND`.

Also applied the `const std::lock_guard` nit in `getMRCacheSize`, `registerMemory`, and `deregisterMemory`.

Build passes: full `ninja -C builddir` rebuild of `libplugin_LIBFABRIC.so` clean.

dmvevents:
Thanks for the careful review. Pushed `f3d131f` addressing the three open threads.

**`libfabric_rail.cpp:1530` — try_emplace late-race leak (Major)**
Confirmed. Captured `inserted` from `try_emplace`, and on the late-hit path:
- `fi_close()` the redundant locally-registered MR (so the libfabric handle and any pinned GDR memory don't leak),
- bump the cached entry's `ref_count` so the existing entry stays alive for the caller's later `deregisterMemory()`,
- return the cached `mr` / `key` to the caller via `*mr_out` / `*key_out`,
- correct the metrics: this call ends up as a hit, so increment `mr_cache_hits_` and decrement `mr_cache_misses_` (we already incremented misses on the optimistic miss path before re-acquiring the lock).

**`libfabric_rail.cpp:1578` — log-level / O(n) deregister scan (Major)**
Two parts:
- *Log level:* agreed, demoted to `NIXL_DEBUG`. The rail manager (`libfabric_rail_manager.cpp:826-832`) already tolerates per-rail `NIXL_ERR_NOT_FOUND` for cross-rail / cleanup paths, so the ERROR was noise on legitimate flows.
- *O(1) secondary index (`unordered_map<fid_mr*, uintptr_t> mr_to_addr_`):* deferring to a follow-up. It's a structural change you flagged "Heavy lift" — it touches every `mr_cache_` insert/erase site (registerMemory hit-validate path, eviction, deregister, late-race hit) and needs its own correctness pass. I'd rather land the leak fix and the log-noise fix on this PR and open a separate change for the index so it gets a clean diff and its own test pass.

**`libfabric_rail.h:326` — Doxygen `@return` tags (Trivial)**
Done. `getMRCacheHits()` / `getMRCacheMisses()` / `getMRCacheSize()` now use `@brief` + `@return` blocks per the coding guideline.

| Thread | Action | Commit |
|---|---|---|
| `libfabric_rail.cpp:1530` | Fixed (late-race close + return cached entry + metric correction) | `f3d131f` |
| `libfabric_rail.cpp:1578` | Log demoted to DEBUG; secondary index deferred to follow-up | `f3d131f` |
| `libfabric_rail.h:326` | Doxygen `@return` tags added | `f3d131f` |

dmvevents:
clang-format follow-up in [`5c0c005`](https://github.com/dmvevents/nixl/commit/5c0c005) — 4 long-line wraps + a `} catch` block reflow per the project's clang-format-19 rule. Diff stays in scope (only lines we already touched in `f3d131f`). CI should clear now.

dmvevents:
Friendly ping @yexiangd @fengjica — this one is green and ready for a maintainer look whenever you have a moment:

- All CI checks pass (clang-format, pre-commit, copyright, python-checks x4).
- CodeRabbit review completed; the three findings were addressed in 9dcae64 / f3d131f / 5c0c005.
- Scope is tight: caches libfabric MR registrations to avoid repeated `fi_mr_regattr` on the hot path (perf-only, no API change).

No rush — just flagging that it's unblocked on our side and waiting on a review. Happy to rebase if you'd prefer it fresher against main.

dmvevents:
Rebased onto current `main` (merge commit 42e6563, conflict was the new `pollForCompletions()` declaration vs the MRRC members in `libfabric_rail.h` — both kept). Full-file compile check against the merged tree passes. The branch is conflict-free again; could a maintainer re-trigger CI (`/ok to test`)?

src/utils/libfabric/libfabric_rail.h
```diff
@@ -601,6 +659,13 @@ class nixlLibfabricRail {
     // Provider capability flags
     bool provider_supports_hmem_;
 
+    // Memory Registration Resource Cache (MRRC)
+    // Based on He et al. "An efficient design for fast memory registration in RDMA" (JNCA 2009)
+    mutable std::unordered_map<uintptr_t, MRCacheEntry> mr_cache_;
+    mutable std::mutex mr_cache_mutex_;
+    mutable std::atomic<uint64_t> mr_cache_hits_{0};
+    mutable std::atomic<uint64_t> mr_cache_misses_{0};
+
```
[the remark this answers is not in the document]
dmvevents: Fixed in 2042ecf3. `cleanup()` now closes and clears all remaining cache entries right after the request pools are cleaned and before the domain is closed (STEP 3b), matching your proposed shape: entries with a live `ref_count` are logged at WARN before closing, and `fi_close` failures are logged rather than aborting teardown so the domain close still proceeds. The reverse index added in the same commit is cleared alongside the map.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1607,10 +1767,50 @@ nixlLibfabricRail::deregisterMemory(struct fid_mr *mr) const {
...
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    // Actually close the MR (last reference from cache)
+    if (should_close) {
+        int ret = fi_close(&mr->fid);
+        if (ret) {
+            NIXL_ERROR << "fi_close failed on rail " << rail_id << ": " << fi_strerror(-ret);
+            return NIXL_ERR_BACKEND;
+        }
```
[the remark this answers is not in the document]
dmvevents: Agreed, and both sites changed together in 2042ecf3 exactly as you outlined: `registerMemory()` now returns the freshly registered MR uncached with `NIXL_SUCCESS` when the cache is full and nothing is evictable (the registration itself succeeded — a full cache is a cache limitation, not a registration failure), and `deregisterMemory()` closes the MR with `fi_close()` on the not-found path and returns `NIXL_SUCCESS`, which also restores the pre-cache contract for any MR the cache doesn't own. Close failures on that path still surface as `NIXL_ERR_BACKEND`.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1407,10 +1496,46 @@ nixlLibfabricRail::deregisterMemory(struct fid_mr *mr) const {
...
+                }
+            }
+        }
+    }
+
+    if (!found_in_cache) {
+        // MR not in cache - caller may be trying to deregister an MR from another rail
+        NIXL_ERROR << "MRRC: Attempted to deregister uncached MR on rail " << rail_id;
+        return NIXL_ERR_NOT_FOUND;
+    }
```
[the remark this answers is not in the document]
dmvevents: Both points addressed in 2042ecf3. The O(n) scan is gone: a `fid_mr* -> buffer address` reverse index (`mr_cache_by_mr_`, maintained in lockstep with `mr_cache_` under the same mutex) makes `deregisterMemory()` a pair of O(1) lookups. On the log level: the not-found path is now a supported case (uncached MRs are closed directly and return `NIXL_SUCCESS`, per the ownership thread), so it logs at DEBUG rather than ERROR.

src/utils/libfabric/libfabric_rail.h
```diff
@@ -439,6 +469,34 @@ class nixlLibfabricRail {
...
+    getMRCacheMisses() const {
+        return mr_cache_misses_.load(std::memory_order_relaxed);
+    }
+
+    /**
+     * @brief Get the current number of entries in the MR cache.
+     * @return Number of cache entries currently held by this rail.
+     */
+    size_t
+    getMRCacheSize() const;
```
rongbingzhou, line 498: Looks like these 3 functions are never used. Do we really need them?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1541,6 +1574,56 @@ nixlLibfabricRail::registerMemory(void *buffer,
...
+                           << " ref_count=" << entry.ref_count.load() << " key=" << entry.key;
+                return NIXL_SUCCESS;
+            }
+            // Parameters don't match the cached entry. If no one is holding a
+            // reference, evict the stale entry so the later try_emplace can
+            // actually insert the freshly-registered MR. Otherwise fail
+            // loudly: re-registering the same address with different length
+            // while an active reference exists means the caller has the old
+            // MR mapped but is asking us to pretend it has a new size.
+            if (entry.ref_count.load(std::memory_order_acquire) == 0) {
```
rongbingzhou, line 1602: Can we really reach this `ref_count == 0` condition? The only place that we sub `ref_count` is at line 1870 of `libfabric_rail.cpp`. Then at line 1871, when `prev_count == 1`, we deregister and remove the cache. So I don't think we can ever find a cache with `ref_count == 0`.

If that's the case, I worry that this behavior will introduce a regression. Previously we can successfully register the same address with a different length. Now with this logic, registering the same address with a different length will enter `else` logic and fail with `NIXL_ERR_INVALID_PARAM` error.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1638,16 +1721,125 @@ nixlLibfabricRail::registerMemory(void *buffer,
...
+        NIXL_TRACE << "MR key obtained: " << actual_key << " (" << std::hex << actual_key
+                   << std::dec << ")";
     }
+
+    // MRRC: Add to cache with ref_count=1
+    {
+        const std::lock_guard<std::mutex> lock(mr_cache_mutex_);
+
+        // Cache size limit and eviction logic
+        constexpr size_t NIXL_MR_CACHE_MAX_ENTRIES_DEFAULT = 1024;
```
rongbingzhou, line 1739: Could you add a comment to explain why you choose `1024` as default value?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1638,16 +1721,125 @@ nixlLibfabricRail::registerMemory(void *buffer,
...
+                    return NIXL_MR_CACHE_MAX_ENTRIES_DEFAULT;
+                }
+                return parsed;
+            }
+            catch (const std::exception &e) {
+                NIXL_WARN << "NIXL_MR_CACHE_MAX_ENTRIES='" << max_str << "' could not be parsed ("
+                          << e.what() << "); falling back to default";
+                return NIXL_MR_CACHE_MAX_ENTRIES_DEFAULT;
+            }
+        }();
```
rongbingzhou, line 1762: Suggestion: use `LibfabricUtils::getCustomIntParam()` mechanism to take care of env variables instead of implementing one by ourselves.

PR #1229, opened 2026-01-19
API: Device API V2

ofirfarjun7:
## What?
Device API V2.

## Why?
Gives more flexibility to the users while maintaining less GPU handles.
Users create GPU transfer request handle using `nixlAgent::createGpuXferReq` in V1. Also, before calling `nixlAgent::createGpuXferReq`, `nixlAgent::createGpuXferReq` must be called.
V2 also closes the missing functionality associated with the separate creation of the source and destination memory descriptors.

## How?
Related PRs:
https://github.com/ai-dynamo/nixl/pull/1230
https://github.com/ai-dynamo/nixl/pull/1245
https://github.com/ai-dynamo/nixl/pull/1255

PR #1230, opened 2026-01-19
CORE: Device API V2. Host-side implementation.

ofirfarjun7:
## What?
The implementation of host-side of Device API V2.

## Why?
Gives more flexibility to the users while maintaining less GPU handles.
Users create GPU transfer request handle using `nixlAgent::createGpuXferReq` in V1. Also, before calling `nixlAgent::createGpuXferReq`, `nixlAgent::createGpuXferReq` must be called.
V2 also closes the missing functionality associated with the separate creation of the source and destination memory descriptors.

## How?
Related PRs:
https://github.com/ai-dynamo/nixl/pull/1229
https://github.com/ai-dynamo/nixl/pull/1245
https://github.com/ai-dynamo/nixl/pull/1255

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -36,8 +36,6 @@ static const std::vector<std::vector<std::string>> illegal_plugin_combinations =
     {"GDS", "GDS_MT"},
 };
 const std::string nixl_invalid_agent = "INVALID_AGENT";
-constexpr std::string_view nixl_backend_not_found{
-    "No backend supports the required transfer registrations."};
 
 /*** nixlEnumStrings namespace implementation in API ***/
 std::string nixlEnumStrings::memTypeStr(const nixl_mem_t &mem) {
@@ -939,7 +937,8 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     }
 
     if (!handle->engine) {
-        NIXL_ERROR_FUNC << nixl_backend_not_found;
+        NIXL_ERROR_FUNC << "no specified or potential backend had the required "
+                           "registrations to be able to do the transfer";
         data->addErrorTelemetry(NIXL_ERR_NOT_FOUND);
         return NIXL_ERR_NOT_FOUND;
     }
@@ -1838,18 +1837,16 @@ backend_set_t
 nixlAgentData::getBackends(const nixl_opt_args_t *opt_args,
                            nixlMemSection *section,
                            nixl_mem_t mem_type) {
-    backend_set_t backends;
-    if (opt_args) {
+    if (opt_args && !opt_args->backends.empty()) {
+        backend_set_t backends;
         for (const auto &backend : opt_args->backends) {
             backends.insert(backend->engine);
         }
-    }
 
-    if (!backends.empty()) {
         return backends;
     }
 
-    backend_set_t *mem_type_backends{section->queryBackends(mem_type)};
+    const auto mem_type_backends = section->queryBackends(mem_type);
     return mem_type_backends ? *mem_type_backends : backend_set_t{};
 }
 
@@ -1857,12 +1854,12 @@ nixl_status_t
 nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
                           nixlMemoryViewH &mvh,
                           const nixl_opt_args_t *extra_params) const {
-    NIXL_SHARED_LOCK_GUARD(data->lock);
-
     const auto desc_count = static_cast<size_t>(dlist.descCount());
-    const nixl_mem_t mem_type{dlist.getType()};
+    const auto mem_type = dlist.getType();
     nixl_remote_meta_dlist_t remote_meta_dlist{mem_type};
     nixlBackendEngine *engine{nullptr};
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
     for (size_t i = 0; i < desc_count; ++i) {
         const auto &desc = dlist[i];
         if (desc.remoteAgent == nixl_invalid_agent) {
@@ -1877,8 +1874,8 @@ nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
         }
 
         if (engine) {
-            // Engine is already selected, populate the remote metadata
-            const auto status = it->second->populate(desc, engine, remote_meta_dlist);
+            // Engine has already been selected, add element to the remote metadata
+            const auto status = it->second->addElement(desc, engine, remote_meta_dlist);
             if (status != NIXL_SUCCESS) {
                 return status;
             }
@@ -1886,10 +1883,11 @@ nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
             continue;
         }
 
-        // Engine is not selected, try to find a backend that can populate the remote metadata
-        const backend_set_t backends{data->getBackends(extra_params, it->second, mem_type)};
+        // Engine has not been selected yet, try to find a backend that can add an element to the
+        // remote metadata
+        const auto backends = data->getBackends(extra_params, it->second, mem_type);
         for (const auto &backend : backends) {
-            const nixl_status_t status{it->second->populate(desc, backend, remote_meta_dlist)};
+            const auto status = it->second->addElement(desc, backend, remote_meta_dlist);
             if (status == NIXL_SUCCESS) {
                 NIXL_DEBUG << "Selected backend: " << backend->getType();
                 engine = backend;
@@ -1897,14 +1895,15 @@ nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
             }
         }
 
-        // If no backend can populate the remote metadata, return an error
+        // If no backend can add an element to the remote metadata, return an error
         if (!engine) {
             break;
         }
     }
 
     if (!engine) {
-        NIXL_ERROR_FUNC << nixl_backend_not_found;
+        NIXL_ERROR_FUNC
+            << "A backend capable of creating a list of remote memory descriptors was not found";
         return NIXL_ERR_NOT_FOUND;
     }
 
@@ -1913,7 +1912,7 @@ nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
         opt_args.customParam = extra_params->customParam;
     }
 
-    const nixl_status_t status{engine->prepMemoryView(remote_meta_dlist, mvh, &opt_args)};
+    const auto status = engine->prepMemoryView(remote_meta_dlist, mvh, &opt_args);
     if (status == NIXL_SUCCESS) {
         data->mvhToEngine.emplace(mvh, *engine);
     }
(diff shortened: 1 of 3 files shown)


src/core/nixl_agent.cpp
```diff
@@ -35,6 +35,10 @@ constexpr char TELEMETRY_ENABLED_VAR[] = "NIXL_TELEMETRY_ENABLE";
 static const std::vector<std::vector<std::string>> illegal_plugin_combinations = {
     {"GDS", "GDS_MT"},
 };
+const std::string nixl_invalid_agent = "INVALID_AGENT";
+constexpr std::string_view nixl_backend_not_found{
+    "No specified or potential backend has the required registrations to be able to do the "
+    "transfer"};
```
brminich: maybe just `No backend supports the required transfer registrations.`

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1834,145 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+    for (size_t i = 0; i < desc_count; ++i) {
+        const auto &desc = dlist[i];
+        if (desc.remoteAgent == nixl_invalid_agent) {
+            remote_meta_dlist.addDesc(nixlRemoteMetaDesc(nixl_invalid_agent));
+            continue;
+        }
+
+        const auto it = data->remoteSections.find(desc.remoteAgent);
+        if (it == data->remoteSections.end()) {
+            NIXL_ERROR << "Metadata for remote agent '" << desc.remoteAgent << "' not found";
```
brminich: why are you sometimes using `NIXL_ERROR_FUNC` and now `NIXL_ERROR`?

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1834,145 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+            const auto status = it->second->populate(desc, backend, remote_meta_dlist);
+            if (status == NIXL_SUCCESS) {
+                NIXL_DEBUG << "Selected backend: " << backend->getType();
+                engine = backend;
+                break;
+            }
+        }
+
+        if (!engine) {
+            break;
```
brminich, line 1900: why can't you return here? to support empty desc?

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1834,145 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+
+            continue;
+        }
+
+        if (backends.empty()) {
+            const auto mem_type_backends = it->second->queryBackends(mem_type);
+            if (mem_type_backends) {
+                backends.insert(mem_type_backends->begin(), mem_type_backends->end());
+            }
+        }
```
brminich: why can't you do that outside of the loop? Then you can unite backends population with another variant of `prepMemoryView`

src/core/nixl_agent.cpp
rakhmets: Replaced by `NIXL_ERROR_FUNC`.

src/core/nixl_agent.cpp
rakhmets: Updated.

src/core/nixl_agent.cpp
rakhmets: Moved the code outside the loop.

src/core/nixl_agent.cpp
rakhmets: Because we still need a backend to perform the memory view preparation.

rakhmets: I think I misunderstood your question.
Yes, to support the case when all descriptors are empty, or there are no descriptors. Otherwise, it will be the same check performed twice.

src/core/nixl_agent.cpp
rakhmets: Moved the common code to separate functions.

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,163 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+            const auto it = data->remoteSections.find(desc.remoteAgent);
+            if (it == data->remoteSections.end()) {
+                continue;
+            }
+
+            backends = getMemBackends(it->second, mem_type);
+            break;
+        }
+    }
+
```
brminich: seems this func can be simplified. E.g. can you return here if dlist is empty and/or no backends found?

rakhmets: Moved a little bit more code to a common function.

src/core/nixl_agent.cpp
```diff
@@ -936,8 +939,7 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     }
 
     if (!handle->engine) {
-        NIXL_ERROR_FUNC << "no specified or potential backend had the required "
-                           "registrations to be able to do the transfer";
+        NIXL_ERROR_FUNC << nixl_backend_not_found;
```
ColinNV: Would it be useful to include information on the "required transfer registrations"?

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
     return NIXL_ERR_NOT_FOUND;
 }
+
+backend_set_t
+nixlAgentData::getBackends(const nixl_opt_args_t *opt_args,
+                           nixlMemSection *section,
+                           nixl_mem_t mem_type) {
+    backend_set_t backends;
+    if (opt_args) {
+        for (const auto &backend : opt_args->backends) {
```
ColinNV, line 1842: Could `opt_args->backends` and `backends` be made the same type of container? Or what if `opt_args->backends` was already sorted?

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+            backends.insert(backend->engine);
+        }
+    }
+
+    if (!backends.empty()) {
```
ColinNV: If this clause is added to the `if (opt_args)` then the fallback below could be simplified.

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+}
+
+nixl_status_t
+nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
+                          nixlMemoryViewH &mvh,
+                          const nixl_opt_args_t *extra_params) const {
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    const auto desc_count = static_cast<size_t>(dlist.descCount());
+    const nixl_mem_t mem_type{dlist.getType()};
```
ColinNV: Given that uniform initialization didn't work out I would suggest using `= dlist.getType()` instead of curly braces wherever the braces aren't needed (for an initializer list).

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+
+    backend_set_t *mem_type_backends{section->queryBackends(mem_type)};
+    return mem_type_backends ? *mem_type_backends : backend_set_t{};
+}
+
+nixl_status_t
+nixlAgent::prepMemoryView(const nixl_remote_dlist_t &dlist,
+                          nixlMemoryViewH &mvh,
+                          const nixl_opt_args_t *extra_params) const {
+    NIXL_SHARED_LOCK_GUARD(data->lock);
```
ColinNV: Can the scope of the lock be safely reduced within the function?

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+nixlAgent::prepMemoryView(const nixl_xfer_dlist_t &dlist,
+                          nixlMemoryViewH &mvh,
+                          const nixl_opt_args_t *extra_params) const {
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    const nixl_mem_t mem_type{dlist.getType()};
+    const backend_set_t backends{data->getBackends(extra_params, data->memorySection, mem_type)};
+    nixl_meta_dlist_t meta_dlist{mem_type};
+    nixlBackendEngine *engine{nullptr};
+    for (const auto &backend : backends) {
```
ColinNV, line 1933: Looks like the same for-loop is also in `prepMemoryView()`, could be factored out.

src/core/nixl_agent.cpp
```diff
@@ -1848,3 +1833,141 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
...
+    return status;
+}
+
+void
+nixlAgent::releaseMemoryView(nixlMemoryViewH mvh) const {
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    const auto it = data->mvhToEngine.find(mvh);
+    if (it == data->mvhToEngine.end()) {
+        NIXL_WARN << "Invalid mvh[" << mvh << "] ";
```
ColinNV: There's something wrong with the spaces.

src/infra/nixl_memory_section.cpp
```diff
@@ -87,6 +87,26 @@ nixl_status_t nixlMemSection::populate (const nixl_xfer_dlist_t &query,
     return NIXL_SUCCESS;
 }
 
+nixl_status_t
+nixlMemSection::populate(const nixlRemoteDesc &query,
+                         nixlBackendEngine *backend,
+                         nixl_remote_meta_dlist_t &resp) const {
+    section_key_t sec_key = std::make_pair(VRAM_SEG, backend);
```
ColinNV: I believe `const section_key_t sec_key( VRAM_SEG, backend )` should work. Deduction guides in C++17 make the `make_pair` superfluous.

rakhmets: Fixed.

src/core/nixl_agent.cpp
rakhmets: Fixed.

src/core/nixl_agent.cpp
rakhmets: These loops call different methods. Updated the name of the method.

src/core/nixl_agent.cpp
rakhmets: Removed the changes unrelated to the Device API.

src/core/nixl_agent.cpp
rakhmets: Changed to `= dlist.getType()`.

src/core/nixl_agent.cpp
rakhmets: moved to the line with `if (opt_args)`.

src/core/nixl_agent.cpp
rakhmets: It could. However, it won't simplify the code. Since the type of `opt_args->backends` is `std::vector<nixlBackendH*>`, and it will create some extra code to handle this raw pointers...

src/core/nixl_agent.cpp
rakhmets: Moved the lock to couple lines below.

PR #1232, opened 2026-01-20
NIXL: Release GIL lock in time consuming functions

iyastreb:
## What?
Release GIL lock in time consuming functions

## Why?
To improve NIXL apps performance


src/bindings/python/nixl_bindings.cpp
```diff
@@ -644,7 +648,8 @@ PYBIND11_MODULE(_bindings, m) {
             py::arg("remote_descs"),
             py::arg("remote_agent"),
             py::arg("notif_msg") = std::string(""),
-            py::arg("backend") = std::vector<uintptr_t>({}))
+            py::arg("backend") = std::vector<uintptr_t>({}),
+            py::call_guard<py::gil_scoped_release>())
```
ovidiusm, line 655: This is not allowed. `init_indices_lambda ` is manipulating python objects

ovidiusm: Sorry I think I was looking at the wrong context. Let me check this again.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -715,12 +723,15 @@ PYBIND11_MODULE(_bindings, m) {
                 nixl_notifs_t new_notifs;
                 nixl_opt_args_t extra_params;
 
-                for (uintptr_t backend : backends)
-                    extra_params.backends.push_back((nixlBackendH *)backend);
+                {
+                    py::gil_scoped_release release;
```
brminich, line 730: why do you release here? if it is because of exception there are other APIs which do that as well

src/bindings/python/nixl_bindings.cpp
tvegas1, line 774: for my understanding, regardless of perf, we cannot release here because of `return py::bytes(ret_str)` that is python object?

src/bindings/python/nixl_bindings.cpp
iyastreb: This function consists of 2 parts: getting notifications from C++ module, and then converting result into a Python structure. For the second part GIL is needed, it crashes otherwise.
So we cannot hold the lock for the entire function duration -> therefore I limited the lock scope to the part that may call progress

src/bindings/python/nixl_bindings.cpp
iyastreb: Yes, correct, when accessing Python objects we must have GIL acquired.

iyastreb: We can have a scoped lock though, like I added in the `getNotifs`

tvegas1: yes maybe for completeness we can release in all calls in case it could end-up helping incidentally?

iyastreb: If I understood correctly releasing this lock in fast functions is not worth it.
So I added only to those which are proved to rake some time.
But in principle we can also release GIL in a few more functions: loadRemoteMD, sendLocalMD, createBackend

brminich:
@ovidiusm, @tvegas1 ok to merge?

tvegas1 (approved):

PR #1234, opened 2026-01-21
PLUGINS/UCX: Minor cleanup.

rakhmets:
## What?
Combined the function `getWorkerIdFromOptArgs` and checking the return value of this function into a single method.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1050,7 +1050,7 @@ nixl_status_t nixlUcxEngine::unloadMD (nixlBackendMD* input) {
 size_t
 nixlUcxEngine::getWorkerId(const nixl_opt_b_args_t *opt_args) const noexcept {
     if (opt_args) {
-        const auto worker_id = getWorkerId(*opt_args);
+        const auto worker_id = getWorkerIdFromOptArgs(*opt_args);
         if (worker_id) {
             return *worker_id;
         }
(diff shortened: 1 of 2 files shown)


tvegas1 (dismissed):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1101,8 +1104,7 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
     }
 
     /* TODO: try to get from a pool first */
```
iyastreb: this comment is related to a handle

src/plugins/ucx/ucx_backend.h
```diff
@@ -226,8 +221,8 @@ class nixlUcxEngine : public nixlBackendEngine {
         return uws[worker_id];
     }
 
-    size_t
-    getWorkerId() const;
+    [[nodiscard]] size_t
+    getWorkerId(const nixl_opt_b_args_t * = nullptr) const noexcept;
```
iyastreb:
```suggestion
    getWorkerId(const nixl_opt_b_args_t *opt_args = nullptr) const noexcept;
```

src/plugins/ucx/ucx_backend.h
```diff
@@ -289,6 +284,9 @@ class nixlUcxEngine : public nixlBackendEngine {
                        size_t start_idx,
                        size_t end_idx);
 
+    [[nodiscard]] std::optional<size_t>
+    getWorkerId(const nixl_opt_b_args_t &) const noexcept;
```
iyastreb:
```suggestion
    getWorkerIdImpl(const nixl_opt_b_args_t &opt)args) const noexcept;
```
IMO having 2 overloads that are different only by pointer/reference is confusing. Looking at the code I had to check several times which is what. IMO better use a different name

src/plugins/ucx/ucx_backend.cpp
rakhmets: Fixed.

src/plugins/ucx/ucx_backend.h
rakhmets: Fixed.

src/plugins/ucx/ucx_backend.h
rakhmets: Reused the old name.

PR #1237, opened 2026-01-22
fix:Fix role parameter validation logic error

chentianming521:
## What?
    Resolve compilation failure caused by missing string quotes for initiator and target literals in std::string::compare() calls (corrected to "initiator" and "target").

## Why?

## How?
architecture, approach, etc._

### DIFF
--- a/test/nixl/nixl_test.cpp
+++ b/test/nixl/nixl_test.cpp
@@ -273,7 +273,7 @@ int main(int argc, char *argv[]) {
 
     std::transform(role.begin(), role.end(), role.begin(), ::tolower);
 
-    if ((role.compare(initiator) != 0) && (role.compare(target) != 0)) {
+    if (role != initiator && role != target) {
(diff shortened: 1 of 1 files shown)


test/nixl/nixl_test.cpp
```diff
@@ -273,10 +273,10 @@ int main(int argc, char *argv[]) {
 
     std::transform(role.begin(), role.end(), role.begin(), ::tolower);
 
-    if (!role.compare(initiator) && !role.compare(target)) {
-            std::cerr << "Invalid role. Use 'initiator' or 'target'."
-                      << "Currently "<< role <<std::endl;
-            return 1;
+    if ((role.compare(initiator) != 0) && (role.compare(target) != 0)) {
```
brminich: maybe just
if (role != initiator && role != target) {

chentianming521: Ok

PR #1238, opened 2026-01-22
fix: tensor validation in get_xfer_descs/get_reg_descs

matumu:
## What?
Tensor validation in get_xfer_descs/get_reg_descs

## Why?
Passing a list containing non-tensor objects would cause an AttributeError when accessing get_xfer_descs/get_reg_descs


ovidiusm:
Relying on `isinstance` is not typically pythonic. What is the use case? Why did the app pass a list of objects where some were not tensors?

PR #1241, opened 2026-01-22
UTILS/UCX: Show warning when hardware is not supported by UCX

guy-ealey-morag:
## What?
Show a warning if a type of hardware is detected (IB devices or GPUs) but UCX don't support it (due to missing libraries, etc.)

## Why?
This will help users notice this issue faster, before they just experienced performance degradation with no explanations.

## How?
Detect hardware by scanning IPC devices, then check for UCX support by querying the supported memory types.
Print a warning in case of any mismatch.

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -639,8 +639,8 @@ nixlUcxContext::warnAboutHardwareSupportMismatch() const {
     };
     const auto status = ucp_context_query(ctx, &attr);
     if (status != UCS_OK) {
-        throw std::runtime_error("Failed to query UCX context: " +
-                                 std::string(ucs_status_string(status)));
+        NIXL_ERROR << "Failed to query UCX context: " << ucs_status_string(status);
+        return;
     }
 
     const nixl::hwInfo hw_info;
--- a/test/gtest/common.cpp
+++ b/test/gtest/common.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -27,6 +27,7 @@
 #include <sys/socket.h>
 #include <unistd.h>
 #include <random>
+#include "absl/log/log_sink_registry.h"
 
 namespace gtest {
 
@@ -131,6 +132,40 @@ PortAllocator::next_tcp_port() {
 
     throw std::runtime_error("No port available in range: " + std::to_string(instance._min_port) +
                              " - " + std::to_string(instance._max_port));
+}
+
+scopedTestLogSink::scopedTestLogSink() {
+    absl::AddLogSink(&sink_);
+}
+
+scopedTestLogSink::~scopedTestLogSink() {
+    absl::RemoveLogSink(&sink_);
+}
+
+void
+scopedTestLogSink::testLogSink::Send(const absl::LogEntry &entry) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (entry.log_severity() == absl::LogSeverity::kWarning) {
+        warnings_.emplace_back(entry.text_message());
+    }
+}
+
+size_t
+scopedTestLogSink::warningCount() const {
+    std::lock_guard<std::mutex> lock(sink_.mutex_);
+    return sink_.warnings_.size();
+}
+
+size_t
+scopedTestLogSink::countWarningsMatching(const std::string &substring) const {
+    std::lock_guard<std::mutex> lock(sink_.mutex_);
+    size_t count = 0;
+    for (const auto &w : sink_.warnings_) {
+        if (w.find(substring) != std::string::npos) {
+            ++count;
+        }
+    }
+    return count;
 }
 
 } // namespace gtest
--- a/test/gtest/common.h
+++ b/test/gtest/common.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -25,7 +25,11 @@
 #include <stack>
 #include <optional>
 #include <mutex>
+#include <vector>
+#include <string>
 #include "gtest/gtest.h"
+#include "absl/log/log_sink.h"
+#include "absl/log/log_entry.h"
 
 #ifdef HAVE_CUDA
 #include <cuda_runtime.h>
@@ -117,6 +121,47 @@ private:
     uint16_t _port = MIN_PORT;
     uint16_t _min_port = MIN_PORT;
     uint16_t _max_port = MAX_PORT;
+};
+
+/**
+ * @brief A scoped LogSink that captures log messages for testing assertions.
+ *
+ * This class registers itself with Abseil's logging system to intercept
+ * log messages on construction and unregisters on destruction. It can be
+ * used in tests to verify that expected warnings or errors are logged.
+ *
+ * Usage:
+ *   scopedTestLogSink sink;
+ *   // ... code that logs warnings ...
+ *   EXPECT_EQ(sink.warningCount(), 1);
+ *   EXPECT_EQ(sink.countWarnings("expected message"), 1);
+ */
+class scopedTestLogSink {
+public:
+    scopedTestLogSink();
+    ~scopedTestLogSink();
+
+    scopedTestLogSink(const scopedTestLogSink &) = delete;
+    scopedTestLogSink &
+    operator=(const scopedTestLogSink &) = delete;
+
+    [[nodiscard]] size_t
+    warningCount() const;
+
+    [[nodiscard]] size_t
+    countWarningsMatching(const std::string &substring) const;
+
+private:
+    class testLogSink : public absl::LogSink {
+    public:
+        void
+        Send(const absl::LogEntry &entry) override;
+
+        mutable std::mutex mutex_;
+        std::vector<std::string> warnings_;
+    };
+
+    testLogSink sink_;
 };
 
 struct nixlTestParam {
--- a/test/gtest/ucx_hw_warning_test.cpp
+++ b/test/gtest/ucx_hw_warning_test.cpp
@@ -1,0 +1,115 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+#include <string>
+#include <vector>
+
+#include "common.h"
+#include "ucx_utils.h"
+#include "common/hw_info.h"
+
+class UcxHardwareWarningTest : public ::testing::Test {
+protected:
+    gtest::ScopedEnv envHelper_;
+};
+
+/**
+ * Test that a warning is logged when NVIDIA GPUs are present but UCX
+ * CUDA support is not available.
+ */
+TEST_F(UcxHardwareWarningTest, WarnWhenGpuPresentButCudaNotSupported) {
+    const nixl::hwInfo hw_info;
+    if (hw_info.numNvidiaGpus == 0) {
+        GTEST_SKIP() << "No NVIDIA GPUs detected, skipping test";
+    }
+
+    // Disable CUDA transport in UCX
+    envHelper_.addVar("UCX_TLS", "^cuda");
+
+    std::vector<std::string> devs;
+    nixlUcxContext ctx(devs, false, 1, nixl_thread_sync_t::NIXL_THREAD_SYNC_NONE, 0);
+
+    gtest::scopedTestLogSink log_sink;
+    ctx.warnAboutHardwareSupportMismatch();
+
+    EXPECT_EQ(log_sink.warningCount(), 1);
+    EXPECT_EQ(log_sink.countWarningsMatching(
+                  "NVIDIA GPU(s) were detected, but UCX CUDA support was not found"),
+              1);
+
+    envHelper_.popVar();
+}
+
+/**
+ * Test that a warning is logged when IB devices are present but UCX
+ * RDMA support is not available.
+ *
+ * Note: This warning only triggers for UCX >= 1.21.
+ */
+TEST_F(UcxHardwareWarningTest, WarnWhenIbPresentButRdmaNotSupported) {
+    unsigned major, minor, release;
+    ucp_get_version(&major, &minor, &release);
+    if (UCP_VERSION(major, minor) < UCP_VERSION(1, 21)) {
+        GTEST_SKIP() << "UCX version " << major << "." << minor
+                     << " is less than 1.21, skipping test";
+    }
+
+    const nixl::hwInfo hw_info;
+    if (hw_info.numIbDevices == 0) {
+        GTEST_SKIP() << "No IB devices detected, skipping test";
+    }
+
+    // Disable IB transport in UCX
+    envHelper_.addVar("UCX_TLS", "^ib");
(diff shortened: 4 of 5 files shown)


src/utils/ucx/ucx_utils.cpp
```diff
@@ -470,10 +581,48 @@ nixlUcxContext::nixlUcxContext(std::vector<std::string> devs,
...
+        if (hw_info.numIbDevices > 0 && !UCS_BIT_GET(attr.memory_types, UCS_MEMORY_TYPE_RDMA)) {
+            NIXL_WARN << hw_info.numIbDevices
+                      << " IB device(s) were detected, but accelerated IB support was not found! "
+                         "Performance may be degraded.";
+        }
+    }
+    catch (const std::exception &e) {
+        ucp_cleanup(ctx);
+        throw;
     }
```
rakhmets: I know that the cause of having throw/catch in the same function is not because of your changes.
It's because `nixlUcxContext` has a raw pointer `ucp_context_h ctx`. Probably, it should be:
```
std::unique_ptr<ucp_context, void (*)(ucp_context_h)> ctx_;
```
, and initialized with `ucp_cleanup` as a deleter.

guy-ealey-morag: I moved the new logic to a function that is now called only after the object is constructed, so during an exception the deconstructor will be called to perform cleanup.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -647,6 +646,151 @@ nixlUcxContext::getGpuSignalSize() const {
...
+constexpr unsigned long kPciClassGpuDisplay = 0x0300;
+constexpr unsigned long kPciClassGpu3d = 0x0302;
+
+/**
+ * Query hardware information by scanning PCI devices
+ * @param info [out] Output structure to receive hardware info
+ * @return NIXL_SUCCESS on success, error code otherwise
+ */
+[[nodiscard]] nixl_status_t
+queryUcxHardware(nixlUcxHardwareInfo &info) {
```
rakhmets: It looks like this should be a ctor of `nixlUcxHardwareInfo`.
And it should throw an exception instead of return a status.
Then, you can remove the following code from `warnAboutHardwareSupportMismatch`:
```C++
    const auto hw_query_status = queryUcxHardware(hw_info);
    if (hw_query_status != NIXL_SUCCESS) {
        throw std::runtime_error("Failed to query hardware");
    }
```

guy-ealey-morag: Updated, please review

src/utils/ucx/ucx_utils.cpp
```diff
@@ -647,6 +646,151 @@ nixlUcxContext::getGpuSignalSize() const {
...
+    NIXL_TRACE << "UCX hardware query result: "
+               << "num_ib_devices=" << hw_info.numIbDevices << ", "
+               << "num_gpus=" << hw_info.numNvidiaGpus;
+
+    if (hw_info.numNvidiaGpus > 0 && !UCS_BIT_GET(attr.memory_types, UCS_MEMORY_TYPE_CUDA)) {
+        NIXL_WARN << hw_info.numNvidiaGpus
+                  << " NVIDIA GPU(s) were detected, but CUDA support was not found! "
+                  << "Performance may be degraded.";
+    }
+    if (hw_info.numIbDevices > 0 && !UCS_BIT_GET(attr.memory_types, UCS_MEMORY_TYPE_RDMA)) {
```
rakhmets: I have doubts about the reliability of this check. I think this memory type may be unsupported, and UCX may still have accelerated IB transports.

guy-ealey-morag: `UCS_MEMORY_TYPE_RDMA` is supported when `ib_mlx5dv_md` is available (and it also requires device memory and atomics support).

@brminich said it covers the relevant usecase, but we can open it for discussion again

rakhmets: Ah, I see. So, in fact, a stricter condition has been added than just having rc/dc/ud_x transports in UCX.

guy-ealey-morag: Yes, but it's only because of how `UCS_MEMORY_TYPE_RDMA` is already implemented in UCX.
We could change the logic to look for mlx_5 transports if you think it's necessary @rakhmets @brminich

rakhmets: I think it's ok to implement it in this way, since it covers our use cases.

src/utils/ucx/ucx_utils.cpp
```diff
@@ -642,6 +646,125 @@ nixlUcxContext::getGpuSignalSize() const {
...
+    const nixlUcxHardwareInfo hw_info;
+
+    NIXL_TRACE << "nixlUcxHardwareInfo { "
+               << "numNvidiaGpus=" << hw_info.numNvidiaGpus << ", "
+               << "numIbDevices=" << hw_info.numIbDevices << " }";
+
+    if (hw_info.numNvidiaGpus > 0 && !UCS_BIT_GET(attr.memory_types, UCS_MEMORY_TYPE_CUDA)) {
+        NIXL_WARN << hw_info.numNvidiaGpus
+                  << " NVIDIA GPU(s) were detected, but CUDA support was not found! "
+                  << "Performance may be degraded.";
```
```suggestion
                  << " NVIDIA GPU(s) were detected, but UCX CUDA support was not found! "
                  << "GPU memory is not supported.";
```

src/utils/ucx/ucx_utils.cpp
brminich: we can only do this check for UCX >= 1.21 (because you are fixing it in the current UCX master)

src/utils/ucx/ucx_utils.cpp
```diff
@@ -642,6 +646,125 @@ nixlUcxContext::getGpuSignalSize() const {
...
+            auto class_id = readSysfsUlong(device_path, "class", device_name);
+            if (!class_id) {
+                continue;
+            }
+            *class_id >>= 8;
+
+            // Check for InfiniBand device
+            if ((*vendor_id == kPciVendorMellanox) && (*class_id == kPciClassIb)) {
+                numIbDevices++;
+                NIXL_DEBUG << "Found IB device #" << numIbDevices << ": " << device_name
```
brminich: 👍

src/utils/ucx/ucx_utils.cpp
```diff
@@ -642,6 +646,125 @@ nixlUcxContext::getGpuSignalSize() const {
...
+    if (ctx_query_status != UCS_OK) {
+        throw std::runtime_error("Failed to query UCX context: " +
+                                 std::string(ucs_status_string(ctx_query_status)));
+    }
+
+    const nixlUcxHardwareInfo hw_info;
+
+    NIXL_TRACE << "nixlUcxHardwareInfo { "
+               << "numNvidiaGpus=" << hw_info.numNvidiaGpus << ", "
+               << "numIbDevices=" << hw_info.numIbDevices << " }";
```
```suggestion
    NIXL_DEBUG << "nixlUcxHardwareInfo { "
               << "numNvidiaGpus=" << hw_info.numNvidiaGpus << ", "
               << "numIbDevices=" << hw_info.numIbDevices << " }";
```

src/utils/ucx/ucx_utils.cpp
guy-ealey-morag: I added a check against the UCX version

src/utils/ucx/ucx_utils.cpp
```diff
@@ -647,6 +646,128 @@ nixlUcxContext::getGpuSignalSize() const {
 #endif
 }
 
+namespace {
+struct nixlUcxHardwareInfo {
```
rakhmets: What do you think about moving this into a separate file?
This `ucx_utils.cpp` already contains a lot of various functions belonging to different classes.
And this code looks quite self-sufficient.
(I would even consider moving it out of utils/ucx).

If you agree with moving it outside of this file, please note that we started to use namespaces for newly created files (e.g src/utils/ucx/[config|rkey].[h|cpp]).

guy-ealey-morag: Moved to `utils/common`

brminich:
@guy-ealey-morag pls note copyright and clang-format check failures

src/utils/ucx/ucx_utils.cpp
```diff
@@ -32,6 +32,8 @@ extern "C" {
 #endif
 }
 
+#include "nixl_types.h"
```
rakhmets: Why do we need this header?

src/utils/common/hw_info.cpp
```diff
@@ -0,0 +1,114 @@
...
+    constexpr unsigned long kPciClassGpuDisplay = 0x0300;
+    constexpr unsigned long kPciClassGpu3d = 0x0302;
+
+    std::optional<unsigned long>
+    readSysfsUlong(const std::filesystem::path &sysfs_path,
+                   const std::string &file_name,
+                   const std::string &device_name) {
+        std::ifstream file(sysfs_path / file_name);
+        if (!file.is_open()) {
+            NIXL_TRACE << "Failed to read " << file_name << " for device " << device_name;
```
rakhmets:
```suggestion
            NIXL_TRACE << "Failed to open " << file_name << " for device " << device_name;
```

src/utils/common/hw_info.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    constexpr const char *kPciDevicePath = "/sys/bus/pci/devices";
+    constexpr unsigned long kPciVendorMellanox = 0x15b3;
+    constexpr unsigned long kPciVendorNvidia = 0x10de;
+    constexpr unsigned long kPciClassIb = 0x0207;
+    constexpr unsigned long kPciClassGpuDisplay = 0x0300;
+    constexpr unsigned long kPciClassGpu3d = 0x0302;
+
+    std::optional<unsigned long>
+    readSysfsUlong(const std::filesystem::path &sysfs_path,
```
rakhmets, line 41: Can be made more strict by adding `[[nodiscard]]`, and `noexcept`.

src/utils/common/hw_info.cpp
```diff
@@ -0,0 +1,114 @@
...
+                   const std::string &file_name,
+                   const std::string &device_name) {
```
rakhmets: `string_view` since it's only used to print traces by this function.

src/utils/ucx/ucx_utils.h
```diff
@@ -196,6 +196,7 @@ class nixlUcxContext {
     /* Local UCX stuff */
     ucp_context_h ctx;
     nixl_ucx_mt_t mt_type;
+    unsigned ucp_version;
```
rakhmets: I'd recommend using the following format (from CodeStyle.md) for this new private class member despite the fact that the file contains a mixture of different formats.
```suggestion
    unsigned ucpVersion_;
```

guy-ealey-morag: This is not a new variable, I just made did a class member, should I still rename it?

rakhmets: Yes please.
There are different styles for local variables and private class members in NIXL.

src/utils/ucx/ucx_utils.cpp
guy-ealey-morag: Removed

src/utils/ucx/ucx_utils.cpp
```diff
@@ -647,6 +643,38 @@ nixlUcxContext::getGpuSignalSize() const {
 #endif
 }
 
+void
+nixlUcxContext::warnAboutHardwareSupportMismatch() const {
+    ucp_context_attr_t attr = {
+        .field_mask = UCP_ATTR_FIELD_MEMORY_TYPES,
+    };
+    const auto ctx_query_status = ucp_context_query(ctx, &attr);
```
rakhmets: Maybe just `status`, as there are no other statuses in this function.

rakhmets (dismissed):

brminich:
pls check if unit test is possible

rakhmets (dismissed):

guy-ealey-morag:
> pls check if unit test is possible

Added unit tests

brminich:
@guy-ealey-morag pls fix copyright and format check issues

test/gtest/ucx_hw_warning_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+        GTEST_SKIP() << "No NVIDIA GPUs detected, skipping test";
+    }
+
+    // Disable CUDA transport in UCX
+    envHelper_.addVar("UCX_TLS", "^cuda");
+
+    std::vector<std::string> devs;
+    nixlUcxContext ctx(devs, false, 1, nixl_thread_sync_t::NIXL_THREAD_SYNC_NONE, 0);
+
+    logSink_.clear();
```
rakhmets: Why do we need to clear logSink here?

test/gtest/common.cpp
```diff
@@ -133,4 +133,44 @@ PortAllocator::next_tcp_port() {
                              " - " + std::to_string(instance._max_port));
 }
 
+testLogSink::testLogSink() {
+    absl::AddLogSink(this);
```
rakhmets: To be honest, I don't really like it when a class registers itself, using `this`. If you think that, for example, moving registration to the level of a class that owns a logSink would be a better solution, then please consider this change.

rakhmets (approved):

test/gtest/ucx_hw_warning_test.cpp
guy-ealey-morag: To make sure we're looking at logs only from this point forwards

test/gtest/common.cpp
guy-ealey-morag: I wrapped the log sink logic with `scopedTestLogSink` to avoid using `this`

test/gtest/ucx_hw_warning_test.cpp
rakhmets: Maybe 
```suggestion
    gtest::scopedTestLogSink log_sink;
```
, as you always need a "clean" `logSink`. Then you can remove `logSink_` class member, and `clear` method.

test/gtest/common.h
```diff
@@ -25,7 +25,12 @@
 #include <stack>
 #include <optional>
 #include <mutex>
+#include <vector>
+#include <string>
 #include "gtest/gtest.h"
+#include "absl/log/log_sink.h"
+#include "absl/log/log_entry.h"
+#include "absl/log/log_sink_registry.h"
```
rakhmets: I guess these two can be moved to the `.cpp`.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -631,6 +632,39 @@ nixlUcxContext::getGpuSignalSize() const {
...
 }
 
+void
+nixlUcxContext::warnAboutHardwareSupportMismatch() const {
+    ucp_context_attr_t attr = {
+        .field_mask = UCP_ATTR_FIELD_MEMORY_TYPES,
+    };
+    const auto status = ucp_context_query(ctx, &attr);
+    if (status != UCS_OK) {
+        throw std::runtime_error("Failed to query UCX context: " +
```
ovidiusm: I'm not sure we should crash the application instead of log an error or warning.

guy-ealey-morag: Replaced with an error log

rakhmets: In this case, it should be `NIXL_WARN`, since the application continues to run, and the user/client code has no way to handle the error.

brminich:
CI failure seem to be related
```
[2026-02-03T13:10:15.964Z] Loading UCX plugin...
[2026-02-03T13:10:15.964Z] Segmentation fault (core dumped)
Test CPP failed with msg: Step Test CPP failed with exit code=139hudson.AbortException: parallel task failed with msg: hudson.AbortException: Test CPP failed with msg: Step Test CPP failed with exit code=139
Response Code :: 201
```

src/plugins/ucx/ucx_utils.cpp
rakhmets: After looking at this code again, I still think this is an exceptional situation. We are not expecting that `ucp_context_query` returns an error here. So, throwing exception looks like a valid reaction on it. And this exception is handled by `createEngine`.

src/utils/common/hw_info.cpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+} // namespace
+
+hwInfo::hwInfo() {
+    std::error_code ec;
+    std::filesystem::directory_iterator dir_iter(kPciDevicePath, ec);
+    if (ec) {
+        throw std::runtime_error("Failed to scan PCI devices directory: " + ec.message());
```
ovidiusm, line 72: We should not crash here either. In general let's please not throw exceptions for errors that are not fatal

rakhmets: But it will not crash. The exception will be caught by try/catch block, and the error will be printed.

src/plugins/ucx/ucx_utils.cpp
ovidiusm: At a high level view, this feature implements an API called `warnAboutHardwareSupportMismatch` so I find it unexpected that it would lead to failing backend creation. My expectation is that in the worst case, it generates a warning message, and in the normal case, it is quiet.

src/utils/common/hw_info.cpp
rakhmets: It will be caught by `nixlBackendPluginCreator<nixlUcxEngine>::createEngine`.
And the error will be printed. And the app will continue working.

src/plugins/ucx/ucx_utils.cpp
rakhmets: Yes, it generates a warning message, and in the normal case, it is quite.
However, this method works with a successfully created UCP context. And if `ucp_context_query` returns an error, it means that something exceptional has happened.

src/utils/common/hw_info.cpp
ovidiusm: Let's merge it as it is and I will open a PR on top of the change

PR #1243, opened 2026-01-23
NIXL: Dump descList on posting with debug log

iyastreb:
## What?
When debugging application it's extremely useful to know the message pattern being used.
This PR dumps descList content in a very compact form to debug log:
```
I0123 09:20:26.824393  214829 nixl_agent.cpp:1101] descList mem_type=1 count=64 content={65536:64,1024:8}
```

## Why?
When debugging application it's extremely useful to know the message pattern being used.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -35,7 +35,6 @@ constexpr char TELEMETRY_ENABLED_VAR[] = "NIXL_TELEMETRY_ENABLE";
 static const std::vector<std::vector<std::string>> illegal_plugin_combinations = {
     {"GDS", "GDS_MT"},
 };
-const std::string nixl_invalid_agent = "INVALID_AGENT";
 
 /*** nixlEnumStrings namespace implementation in API ***/
 std::string nixlEnumStrings::memTypeStr(const nixl_mem_t &mem) {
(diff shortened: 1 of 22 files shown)


src/api/cpp/nixl_descriptors.h
```diff
@@ -420,6 +420,12 @@ class nixlDescList {
      */
     void
     print() const;
+
+    /**
+     * @brief Dump the descriptor list into a string for debugging
+     */
+    std::string
+    dump() const;
```
rakhmets: Do we need both: `print` and `dump`? Can we replace `print` by `dump`, or modify `print` to suit your needs?

iyastreb: Well, I cannot change print, because it's already exposed as API, and being used from client applications including Rust binding. I cannot reuse it as well, because it prints data on the console.
What I can do however is to:
- implement both compact and "full" formats in dump
- call dump from print, to reuse the code

src/core/nixl_agent.cpp
```diff
@@ -1098,6 +1098,8 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
         return NIXL_ERR_BACKEND;
     }
 
+    NIXL_DEBUG << req_hndl->initiatorDescs->dump();
```
brminich: we typically use TRACE log  for data path. Do you think it can be a part of telemetry or at least be printed only when telemetry is enabled (to avoid extra branch)

iyastreb: With TRACE level we can't know the message pattern with release builds, right?
Yes, I like the idea of telemetry, let me try to check if it works out of the box

iyastreb: done

Issue #1146, opened 2025-12-18
nixlbench: failed to run multiple devices with backend GUSLI


omer-b1:
> nixlbench --backend=GUSLI \
           --device_list="0:K:/dev/nvme2n1,1:K:/dev/nvme2n2" \
           --num_initiator_dev=2 \
           --num_target_dev=2 \
           --op_type=READ
WARNING: Adjusting num_iter to 1008 to allow equal distribution to 1 threads
WARNING: Adjusting warmup_iter to 112 to allow equal distribution to 1 threads
Using null runtime for storage backend without ETCD
GUSLI backend: Automatically enabling storage_enable_direct for direct I/O
GUSLI backend initialized:
  Client name: NIXLBench
  Max simultaneous requests: 32
  Direct I/O: Enabled (required)
  Configured devices: 2
    Device 0 [K]: /dev/nvme2n1 (sec=0x3)
    Device 1 [K]: /dev/nvme2n2 (sec=0x3)
GUSLIc: initialized: 2 devices, log_fd=1 rv=0
Single instance storage backend - no synchronization needed
"****************************************************************************************************************************************************************"
NIXLBench Configuration
"****************************************************************************************************************************************************************"
Runtime (--runtime_type=[etcd])                             : ETCD
ETCD Endpoint                                               : disabled (storage backend)
Worker type (--worker_type=[nixl,nvshmem])                  : nixl
Backend (--backend=[UCX,GDS,GDS_MT,POSIX,Mooncake,HF3FS,OBJ]): GUSLI
Enable pt (--enable_pt=[0,1])                               : 0
Progress threads (--progress_threads=N)                     : 0
Device list (--device_list=dev1,dev2,...)                   : 0:K:/dev/nvme2n1,1:K:/dev/nvme2n2
Enable VMM (--enable_vmm=[0,1])                             : 0
filepath (--filepath=path)                                  :
Number of files (--num_files=N)                             : 1
Storage enable direct (--storage_enable_direct=[0,1])       : 1
Initiator seg type (--initiator_seg_type=[DRAM,VRAM])       : DRAM
Target seg type (--target_seg_type=[DRAM,VRAM])             : DRAM
Scheme (--scheme=[pairwise,manytoone,onetomany,tp])         : pairwise
Mode (--mode=[SG,MG])                                       : SG
Op type (--op_type=[READ,WRITE])                            : READ
Check consistency (--check_consistency=[0,1])               : 0
[...]

PR #1244, opened 2026-01-23
Plugins/GUSLI: Fix the incorrect offset

aranadive:
Previously, specifying multiple devices for GUSLI would not work since the offset specified during registration was different than the transfer descriptor.

This removes the incorrect offset during registration. Also, allow user to sepcify offset per file or default to 1MB.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -839,6 +839,12 @@ parseGusliDeviceList(const std::string &device_list,
                   << "). Using 'sec=0x3' for missing entries." << std::endl;
     }
 
+    if (!dev_offsets.empty() && dev_offsets.size() != devices.size()) {
+        std::cerr << "Warning: Number of device offsets (" << dev_offsets.size()
+                  << ") doesn't match number of devices (" << devices.size()
+                  << "). Using 'offset=1048576' for missing entries." << std::endl;
+    }
+
     if (num_devices > 0 && devices.size() != static_cast<size_t>(num_devices)) {
         std::cerr << "Error: Number of devices in device_list (" << devices.size()
                   << ") must match num_devices (" << num_devices << ")" << std::endl;
--- a/benchmark/nixlbench/src/utils/utils.h
+++ b/benchmark/nixlbench/src/utils/utils.h
@@ -217,7 +217,7 @@ struct GusliDeviceConfig {
     size_t dev_offset;
 };
 
-// Parser for GUSLI device list: "id:type:path,id:type:path:offset,..."
+// Parser for GUSLI device list: "id:type:path,id:type:path,..." and byte-based device offset list
 // security_list: comma-separated security flags; num_devices: expected device count (validation)
 std::vector<GusliDeviceConfig>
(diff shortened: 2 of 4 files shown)


benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -199,9 +199,10 @@ const std::vector<xferBenchParamInfo> xbench_params = {
         gusli_config_file,
         "",
         "Configuration file content for GUSLI backend (if empty, auto-generated from device_list)"),
-    NB_ARG_UINT64(gusli_bdev_byte_offset,
-                  1048576,
-                  "Byte offset in block device for GUSLI operations (default: 1MB)"),
+    NB_ARG_STRING(gusli_device_byte_offsets,
```
brminich, line 202: do you know if there are customers already using this parameter? If there are such users, we can introduce new parameter and keep the old one as deprecated (still accept that, but  do not show in the help messages)

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -777,6 +779,15 @@ parseGusliDeviceList(const std::string &device_list,
...
 
+    std::vector<size_t> dev_offsets;
+    if (!dev_offset_list.empty()) {
+        std::stringstream dev_ss(dev_offset_list);
+        std::string offset;
+        while (std::getline(dev_ss, offset, ',')) {
+            dev_offsets.push_back(std::stoull(offset));
+        }
+    }
+
```
brminich, line 790: should the error below be modified to mention offset in addition to `id:type:path`?

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -214,13 +214,15 @@ struct GusliDeviceConfig {
     char device_type; // 'F' for file, 'K' for kernel device, 'N' for networked server
     std::string device_path;
     std::string security_flags;
+    size_t dev_offset;
 };
 
-// Parser for GUSLI device list: "id:type:path,id:type:path,..."
+// Parser for GUSLI device list: "id:type:path,id:type:path:offset,..."
```
brminich: it is a bit confusing - is offset a part of devica_list string or is it taken from new `dev_offset_list` parameter?

benchmark/nixlbench/src/utils/utils.cpp
aranadive: Thats a good point. I know customers that are trying to use GUSLI and they brought up this issue of multiple files. They dont use the parameter explicitly as yet. I can sync with them that the parameter is changing.

benchmark/nixlbench/src/utils/utils.h
aranadive: No its a separate parameter. I removed it from the device list parser comment. Also updated README for the new parameter.

benchmark/nixlbench/src/utils/utils.cpp
aranadive: I added a warning like for security_flags. I removed specifying the offset from the device list itself.

aranadive:
Fixes #1146

PR #1245, opened 2026-01-23
PLUGINS/UCX: Device API V2. Backend implementation.

rakhmets:
## What?
The implementation of backend of Device API V2.

## Why?
Gives more flexibility to the users while maintaining less GPU handles.
Users create GPU transfer request handle using `nixlAgent::createGpuXferReq` in V1. Also, before calling `nixlAgent::createGpuXferReq`, `nixlAgent::createGpuXferReq` must be called.
V2 also closes the missing functionality associated with the separate creation of the source and destination memory descriptors.

## How?
Related PRs:
https://github.com/ai-dynamo/nixl/pull/1229
https://github.com/ai-dynamo/nixl/pull/1230
https://github.com/ai-dynamo/nixl/pull/1255

### DIFF
--- a/src/bindings/rust/tests/test_sync_manager.rs
+++ b/src/bindings/rust/tests/test_sync_manager.rs
@@ -1,0 +1,141 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use nixl_sys::{BackendSyncable, SyncManager};
+use std::cell::Cell;
+use std::rc::Rc;
+
+#[derive(Debug, PartialEq, Eq)]
+enum TestError {
+    SyncFailed,
+}
+
+#[derive(Debug, Clone)]
+struct DummyBackend {
+    sync_calls: Rc<Cell<u32>>,
+}
+
+#[derive(Debug, Clone)]
+struct DummyData {
+    value: Cell<i32>,
+    data_calls: Rc<Cell<u32>>,
+    should_fail: Cell<bool>,
+}
+
+impl DummyBackend {
+    fn default() -> Self {
+        Self {
+            sync_calls: Rc::new(Cell::new(0)),
+        }
+    }
+}
+
+impl DummyData {
+    fn default() -> Self {
+        Self {
+            value: Cell::new(0),
+            data_calls: Rc::new(Cell::new(0)),
+            should_fail: Cell::new(false),
+        }
+    }
+}
+
+impl BackendSyncable for DummyData {
+    type Backend = DummyBackend;
+    type Error = TestError;
+
+    fn sync_to_backend(&self, backend: &Self::Backend) -> Result<(), Self::Error> {
+        if self.should_fail.get() {
+            return Err(TestError::SyncFailed);
+        }
+        self.data_calls.set(self.data_calls.get() + 1);
+        backend
+            .sync_calls
+            .set(backend.sync_calls.get() + 1);
+        Ok(())
+    }
+}
+
+#[test]
+fn sync_manager_first_access_triggers_single_sync() {
+    let data = DummyData::default();
+    let backend = DummyBackend::default();
+    let data_calls = data.data_calls.clone();
+    let backend_calls = backend.sync_calls.clone();
+    let mgr = SyncManager::new(data, backend);
+
+    // First access -> sync happens
+    let _ = mgr.backend().unwrap().sync_calls.get();
+    assert_eq!(data_calls.get(), 1);
+    assert_eq!(backend_calls.get(), 1);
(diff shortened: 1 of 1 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1562,3 +1562,73 @@ nixlUcxEngine::genNotif(const std::string &remote_agent, const std::string &msg)
...
 }
+
+nixl_status_t
+nixlUcxEngine::prepMemoryView(const nixl_remote_meta_dlist_t &meta_dlist,
+                              nixlMemoryViewH &mvh,
+                              const nixl_opt_b_args_t *opt_args) const {
+    try {
+        const auto desc_count = static_cast<size_t>(meta_dlist.descCount());
+        std::vector<std::unique_ptr<nixl::ucx::remoteMem>> remote_mems;
+        const auto worker_id = getWorkerId(opt_args);
```
ColinNV: I would use the actual type here since it makes the code more readable.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1562,3 +1562,73 @@ nixlUcxEngine::genNotif(const std::string &remote_agent, const std::string &msg)
...
+        remote_mems.reserve(desc_count);
+        for (size_t i = 0; i < desc_count; ++i) {
+            if (meta_dlist[i].remoteAgent == nixl_invalid_agent) {
+                remote_mems.emplace_back(nullptr);
```
ColinNV: The `nullptr` argument is not needed.

src/utils/ucx/mem_list.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include "rkey.h"
+#include "ucx_utils.h"
+
+namespace {
+using remoteMem = nixl::ucx::remoteMem;
+
+ucp_device_mem_list_elem_t
+createElement(const std::unique_ptr<remoteMem> &mem) {
+    ucp_device_mem_list_elem_t element;
```
ColinNV, line 35: I'd normally suggest wrapping `ucp_device_mem_list_elem_t` in a C++ class that manages the bitfields depending on how it is constructed (or set later) which would simplify the code here and be less error prone...

src/plugins/ucx/ucx_backend.h
```diff
@@ -307,6 +320,8 @@ class nixlUcxEngine : public nixlBackendEngine {
     // Map of agent name to saved nixlUcxConnection info
     std::unordered_map<std::string, ucx_connection_ptr_t, std::hash<std::string>, strEqual>
         remoteConnMap;
+
+    mutable std::unordered_map<nixlMemoryViewH, nixl::ucx::memList> memViewMap;
```
ColinNV: Are you 120% sure that the `mutable` is ok, meaning that we never have multiple threads calling `const`-declared member functions which in this case are mutating but not thread-safe?

src/plugins/ucx/ucx_backend.cpp
rakhmets: Removed.

src/plugins/ucx/ucx_backend.cpp
rakhmets: Replaced by the type.

src/utils/ucx/mem_list.cpp
rakhmets: Yes, I think it would be a little better in this way. However, I'm afraid to apply the change in this PR. Because I'm not one hundred percent sure that passing the array of these C++ classes as `elements` to `ucp_device_mem_list_params_t`, and size of the class as `element_size` to `ucp_device_mem_list_params_t` will work and will not cause any crashes/asserts inside UCX.
I will try to apply this suggestion in the next PR.

src/plugins/ucx/ucx_backend.h
rakhmets: Yes, backend functions (`prepMemoryView`/`releaseMemoryView`, `createGpuXferReq`/`releaseGpuXferReq`) accessing the cache (`memViewMap`) are always secured by a lock by `nixlAgent`.

iyastreb: Do you mean `NIXL_SHARED_LOCK_GUARD(data->lock);`? By default it's NONE, but can be also RW, where multiple readers invoke prepXfer (same case with prepMemoryWindow?)

rakhmets: Yes, this lock.
The mt pattern is the same as for existing API.

rakhmets: I think your concerns are valid. I will remove caching in this implementation. And will remove the wrapping class.

rakhmets: Removed mutable field.

PR #1247, opened 2026-01-25
refactor(obj): Implement hierarchical object storage client architecture

aranadive:
## What?
Introduce a modular, inheritance‑based architecture for object storage clients to support multiple S3‑compatible backends and acceleration paths.

The previous flat architecture made it difficult to extend the object plugin with new vendor‑specific backends and acceleration methods. This refactoring provides:

- Clean separation of concerns with distinct client types  
- Easy addition of new vendor‑specific backends/engines
- Flexible runtime client selection via custom parameters  
- Polymorphic interface for simplified transfer logic  
- Better code organization matching the inheritance relationships  

Reorganized S3 clients into subdirectories reflecting inheritance:

- `s3/` – Standard S3 client (`awsS3Client`)  
- `s3_crt/` – S3 CRT client (`awsS3CrtClient`) inheriting from `awsS3Client`  
- `s3_accel/` – S3 Accelerated client using CRT implementation

## Why?
- Enable clean extensibility for vendor‑specific engines and clients.
- Keep the engine ABI stable via pImpl while allowing client‑specific overrides.
- Make runtime selection explicit (`accelerated/type/crtMinLimit`) and localized.

## How?
- Engine pImpls live next to their client families:  
  `DefaultObjEngineImpl`, `S3CrtObjEngineImpl`, `S3AccelObjEngineImpl`.
- Shared selection helpers moved to `src/utils/object/engine_utils.h`.
- CRT size‑threshold selection is implemented only in the CRT engine impl.

---

### Engine + Client Hierarchy (single tree)

```
+----------------------+
|   nixlObjEngine      |
+----------------------+
           |
           v
+------------------------------------+
| DefaultObjEngineImpl               |
| uses: awsS3Client                  |
+------------------------------------+
      |
      |--+------------------------------+
      |  | S3CrtObjEngineImpl           |
      |  | uses: awsS3CrtClient         |
      |  +------------------------------+
      |
      +-->+------------------------------+
          | S3AccelObjEngineImpl         |
          | uses: awsS3AccelClient       |
          +------------------------------+
                    |
                    v
         +------------------------------+
          | VendorObjEngineImpl*         |
          | uses: awsVendorClient        |
          +------------------------------+
```

---

### Engine Instantiation Conditions
- **VendorObjEngineImpl**  
  `accelerated=true` and `type=vendor`

- **S3AccelObjEngineImpl**  
  `accelerated=true`

- **S3CrtObjEngineImpl**  
  `crtMinLimit > 0` and not accelerated

- **DefaultObjEngineImpl**  
  fallback when none of the above match

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -223,6 +223,12 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
         }
 
         if (xferBenchConfig::obj_crt_min_limit > 0) {
+            // Warn if both CRT and accelerated options are set - CRT takes precedence
+            if (xferBenchConfig::obj_accelerated_enable) {
+                std::cerr << "Warning: Both obj_crt_min_limit and obj_accelerated_enable are set. "
+                          << "CRT client will be used (takes precedence over accelerated)."
+                          << std::endl;
+            }
             backend_params["crtMinLimit"] = std::to_string(xferBenchConfig::obj_crt_min_limit);
             std::cout << "OBJ backend with S3 CRT client enabled for objects >= "
                       << xferBenchConfig::obj_crt_min_limit << " bytes" << std::endl;
--- a/src/plugins/obj/obj_backend.cpp
+++ b/src/plugins/obj/obj_backend.cpp
@@ -28,6 +28,7 @@
 // Obj Engine Implementation
 // -----------------------------------------------------------------------------
 
+// TODO: Consider a registration pattern as more vendor engines are added.
 std::unique_ptr<nixlObjEngineImpl>
 createObjEngineImpl(const nixlBackendInitParams *init_params) {
     if (isAcceleratedRequested(init_params->customParams)) {
--- a/src/plugins/obj/s3/engine_impl.cpp
+++ b/src/plugins/obj/s3/engine_impl.cpp
@@ -57,15 +57,18 @@ public:
 
     nixl_status_t
     getOverallStatus() {
-        while (!statusFutures_.empty()) {
-            if (statusFutures_.back().wait_for(std::chrono::seconds(0)) ==
-                std::future_status::ready) {
-                auto current_status = statusFutures_.back().get();
+        // Iterate front-to-back to detect failures in earlier futures even if
+        // later futures are not yet ready. This ensures we return errors as
+        // soon as they occur rather than waiting for all futures to complete.
+        auto it = statusFutures_.begin();
+        while (it != statusFutures_.end()) {
+            if (it->wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+                auto current_status = it->get();
                 if (current_status != NIXL_SUCCESS) {
                     statusFutures_.clear();
                     return current_status;
                 }
-                statusFutures_.pop_back();
+                it = statusFutures_.erase(it);
             } else {
                 return NIXL_IN_PROG;
             }
@@ -109,6 +112,9 @@ DefaultObjEngineImpl::DefaultObjEngineImpl(const nixlBackendInitParams *init_par
     : executor_(std::make_shared<asioThreadPoolExecutor>(std::thread::hardware_concurrency())),
       s3Client_(s3_client),
       crtMinLimit_(getCrtMinLimit(init_params->customParams)) {
+    // DefaultObjEngineImpl only uses the standard S3 client, not the CRT client.
+    // The s3_client_crt parameter is accepted for API consistency with derived
+    // engine implementations (e.g., S3CrtObjEngineImpl) but is intentionally unused here.
     (void)s3_client_crt;
     if (s3Client_) s3Client_->setExecutor(executor_);
     NIXL_INFO << "Object storage backend initialized with injected S3 clients";


src/plugins/obj/obj_backend.cpp
```diff
@@ -16,222 +16,77 @@
...
-        NIXL_ERROR << absl::StrFormat("Error: Local memory type must be DRAM_SEG, got %d",
-                                      local.getType());
-        return false;
+std::unique_ptr<nixlObjEngineImpl>
+createObjEngineImpl(const nixlBackendInitParams *init_params,
+                    std::shared_ptr<iS3Client> s3_client,
+                    std::shared_ptr<iS3Client> s3_client_crt) {
+#if defined HAVE_CUOBJ_CLIENT
+    if (isAcceleratedRequested(init_params->customParams)) {
+        return std::make_unique<S3AccelObjEngineImpl>(init_params, s3_client, s3_client_crt);
```
ovidiusm: I cannot find the constructor that receives the s3_client_crt argument

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+DefaultObjEngineImpl::~DefaultObjEngineImpl() {
+    executor_->WaitUntilStopped();
+}
+
+nixl_status_t
+DefaultObjEngineImpl::registerMem(const nixlBlobDesc &mem,
+                                  const nixl_mem_t &nixl_mem,
+                                  nixlBackendMD *&out) {
+    nixl_mem_list_t supported_mems = {OBJ_SEG, DRAM_SEG};
```
ovidiusm, line 131: Which `supported_mems` has precedence, there is another one from:

```
#if defined HAVE_CUOBJ_CLIENT
static const nixl_mem_list_t supported_segments = {DRAM_SEG, OBJ_SEG, VRAM_SEG};
#else
static const nixl_mem_list_t supported_segments = {DRAM_SEG, OBJ_SEG};
#endif
```

jgoldsch12:
I think the implementation of getSupportedMems() in obj_backend.h needs to change so that it calls the engine implementation specific version.  The overided version I created does not get called, so any registration to VRAM gets rejected.

src/plugins/obj/obj_backend.cpp
aranadive: Fixed

src/plugins/obj/s3/engine_impl.cpp
aranadive: I have removed the VRAM_SEG from here and made it so that engines can re-implement getSupportedMems to expose VRAM_SEG as supported.

aranadive:
> I think the implementation of getSupportedMems() in obj_backend.h needs to change so that it calls the engine implementation specific version.  The overided version I created does not get called, so any registration to VRAM gets rejected.

Thanks. Hopefully this is fixed with the latest commit. The objengine should be able to expose VRAM_SEG in its getSupportedMems implementation.

ovidiusm (dismissed):
Approving nixlbench changes

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+class nixlObjBackendReqH : public nixlBackendReqH {
+public:
+    nixlObjBackendReqH() = default;
+    ~nixlObjBackendReqH() = default;
+
+    std::vector<std::future<nixl_status_t>> statusFutures_;
+
+    nixl_status_t
+    getOverallStatus() {
```
vvenkates27, line 59: Minor: The status futures iteration goes back-to-front (line 793-808 in patch), so if the last future isn't ready, NIXL_IN_PROG is returned even if earlier futures already failed. Consider iterating front-to-back in a future cleanup.

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -0,0 +1,265 @@
...
+DefaultObjEngineImpl::DefaultObjEngineImpl(const nixlBackendInitParams *init_params,
+                                           std::shared_ptr<iS3Client> s3_client,
+                                           std::shared_ptr<iS3Client> s3_client_crt)
+    : executor_(std::make_shared<asioThreadPoolExecutor>(std::thread::hardware_concurrency())),
+      s3Client_(s3_client),
+      crtMinLimit_(getCrtMinLimit(init_params->customParams)) {
+    (void)s3_client_crt;
+    if (s3Client_) s3Client_->setExecutor(executor_);
+    NIXL_INFO << "Object storage backend initialized with injected S3 clients";
+}
```
vvenkates27, line 121: Nit: (void)s3_client_crt; silently ignores the injected CRT client. A brief comment explaining why would help future readers.

src/plugins/obj/obj_backend.cpp
```diff
@@ -16,222 +16,82 @@
...
-            return std::stoull(it->second);
-        }
-        catch (const std::exception &e) {
-            NIXL_WARN << "Invalid crtMinLimit value: " << it->second
-                      << ", using default (CRT disabled)";
-            return 0;
-        }
+std::unique_ptr<nixlObjEngineImpl>
+createObjEngineImpl(const nixlBackendInitParams *init_params) {
+#if defined HAVE_CUOBJ_CLIENT
```
vvenkates27: Future consideration: As more vendor engines are added, this if-else chain will grow. A registration/factory pattern could make vendor additions cleaner without modifying core code.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -226,8 +226,16 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
             backend_params["crtMinLimit"] = std::to_string(xferBenchConfig::obj_crt_min_limit);
             std::cout << "OBJ backend with S3 CRT client enabled for objects >= "
                       << xferBenchConfig::obj_crt_min_limit << " bytes" << std::endl;
+        } else if (xferBenchConfig::obj_accelerated_enable) {
```
vvenkates27, line 235: Minor: If both obj_crt_min_limit > 0 and obj_accelerated_enable are set, CRT wins silently. Consider logging a warning about conflicting options.

vvenkates27 (dismissed):
Overall PR LGTM. Just some minor comments.

src/plugins/obj/s3/engine_impl.cpp
aranadive: Fixed.

src/plugins/obj/s3/engine_impl.cpp
aranadive: Added

src/plugins/obj/obj_backend.cpp
aranadive: I left a TODO here.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
aranadive: Good catch. Fixed.

PR #1251, opened 2026-01-26
libfabric: remove cm thread [Will squash before merge]

fengjica:
## What?
Remove CM thread.

## Why?
CM thread is not needed any more. See the commit description for details.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1274,6 +1274,13 @@ nixlLibfabricEngine::notifSendPriv(const std::string &remote_agent,
                        << " for fragment " << seq_id;
             return NIXL_ERR_BACKEND;
         }
+
+        // Progress the control rail to ensure the message is sent.
+        status = rail_manager.progressAllControlRails();
+        if (status != NIXL_SUCCESS && status != NIXL_IN_PROG) {
+            NIXL_ERROR << "Failed to progress control rails in notifSendPriv.";
+            return status;
+        }
     }
 
     NIXL_DEBUG << "Successfully sent all " << binary_notifications.size()


src/utils/libfabric/libfabric_rail.h
```diff
@@ -214,11 +214,9 @@ class DataRequestPool : public RequestPool {
 
 
 /** Connection state tracking for multi-rail connections */
-enum class ConnectionState {
+enum class ConnectionState: int {
     DISCONNECTED, ///< No connection attempt made, initial state
-    CONNECT_REQ_SENT, ///< Connection request sent, waiting for ACK
-    CONNECT_ACK_SENT, ///< Connection ACK sent (target side)
-    CONNECTED, ///< ACK received, ready for data transfers
+    CONNECTED = 3, ///< ACK received, ready for data transfers. Value set for backward compatibility.
```
akkart-aws: Any specific reason why  we really want to skip enum numbering here. ? 
I think its safe to assume that ConnectionState class is used only by libfabric_backend.cpp and libfabric_rail_manger.cpp and libfabric_rail.cpp, can we just verify if we are using the enum value of CONNECTED everywhere and just set this 0, 1 for better code readability.

fengjica: You are right. I removed it in the next revision. Also removed FAILED there, as it's not used anywhere in the code.

src/plugins/libfabric/libfabric_backend.cpp
yexiang-aws: double check if anything needs to be updated in this disconnect function.  We did not implement logic to explicitly remove AV. So this has been a no-op other than the ConnectionState flip.  Probably just remove the TODO comment

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -650,6 +614,10 @@ nixlLibfabricEngine::establishConnection(const std::string &remote_agent) const
 
     // Use single "Communicator" for CM
```
yexiangd, line 615: it would be great if you can update the log statement and remove below comment

src/plugins/libfabric/libfabric_backend.cpp
yexiangd, line 589: is `establishConnection` function still needed? after this PR, it's just state assignments. There are 2 call sites.
Call site 1: In connect() 
```cpp
// Connection exists but not established - trigger establishConnection()
nixl_status_t status = establishConnection(remote_agent);
```

Call site 2: In postXfer() 
```cpp
if (conn_it->second->overall_state_ == ConnectionState::DISCONNECTED) {
    nixl_status_t status = this->establishConnection(remote_agent);
}
```
which code just do
```
conn_it->second->overall_state_ = ConnectionState::CONNECTED;
```
In a separate PR, you can take this refactor.

yexiangd, line 591: I think we can remove all the mutex on connection management if plugin is just marking state and EFA is going to connect upon request

akkart-aws: We still need to consider the fact that postXfer is called by multiple threads, and establishConnection can potentially be called from each of these threads and     `mutable std::unordered_map<std::string, std::shared_ptr<nixlLibfabricConnection>> connections_` is shared by the entire libfabric_backend class.
I think it will be cleaner to remove the connections_ map and also the mutex associated with it. 

Is there any reason to explicitly maintain the connectionMap now that we libfabric handles the connection establishment internally.

PR #1252, opened 2026-01-27
add batch registration support for nixl

cheese-head:
## What?
Adds batch registration/deregistration methods to the `Agent` struct:
- `register_memories(&RegDescList, Option<&OptArgs>)`
- `deregister_memories(&RegDescList, Option<&OptArgs>)`

## Why?
The current `register_memory` API creates a new `RegDescList` for each descriptor and acquires the agent lock N times for N registrations. The batch API allows registering multiple descriptors with a single lock acquisition.

## How?
Two new methods that accept a pre-populated `RegDescList` and call the underlying `nixl_capi_register_mem` / `nixl_capi_deregister_mem` directly.

### DIFF
--- a/src/bindings/rust/src/agent.rs
+++ b/src/bindings/rust/src/agent.rs
@@ -312,6 +312,58 @@ impl Agent {
         })
     }
 
+    /// Registers multiple memory descriptors with the agent in a single batch.
+    pub fn register_memories(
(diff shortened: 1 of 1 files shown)


brminich:
@cheese-head is it still relevant? (it is still in draft)

PR #1256, opened 2026-01-27
nixl_ep: Migrate to memory view API

itayalroy:
Migrate nixl_ep to the new memory view device
API which greatly simplifies device & host init
code

### DIFF
--- a/examples/device/ep/csrc/kernels/nixl_ep.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep.cu
@@ -59,6 +59,10 @@ __forceinline__ __device__ bool is_rank_masked(int* mask_buffer_ptr, int rank) {
     }
 }
 
+__device__ __forceinline__ unsigned int doorbell_flag(int idx) {
+    return (idx + 1) % 4 == 0 ? UCP_DEVICE_FLAG_NODELAY : 0;
+}
+
 template <bool kUseFP8, bool kUseUE8M0, int kHidden>
 __global__ __launch_bounds__(1024, 1) void
 dispatch(void* packed_recv_x, void* packed_recv_x_scales,
@@ -189,7 +193,7 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
                         nixlMemDesc src_mdesc{nixl_ctx.local_mvh, 0, nixl_ctx.offset_get(src_ptr)};
                         nixlMemDesc dst_mdesc{nixl_ctx.remote_mvh, (size_t) dst_rank, nixl_ctx.offset_get(dst_ptr)};
                         EP_DEVICE_ASSERT(nixlPut<nixl_gpu_level_t::WARP>(src_mdesc, dst_mdesc, num_bytes_per_msg,
-                                dst_expert_local_idx, (slot_idx + 1) % 4 == 0 ? UCP_DEVICE_FLAG_NODELAY : 0) == NIXL_IN_PROG);
+                                dst_expert_local_idx, doorbell_flag(slot_idx)) == NIXL_IN_PROG);
                     } else {
                         // NOTES: only 2 load iterations for 7K hidden with 8 unrolls
                         const auto* src_int4_ptr = reinterpret_cast<const int4*>(src_ptr);
@@ -790,7 +794,7 @@ combine(void* combined_x,
                     nixlMemDesc src_mdesc{nixl_ctx.local_mvh, 0, nixl_ctx.offset_get(buf_ptr)};
                     nixlMemDesc dst_mdesc{nixl_ctx.remote_mvh, (size_t) dst_rank, nixl_ctx.offset_get(dst_ptr)};
                     EP_DEVICE_ASSERT(nixlPut<nixl_gpu_level_t::WARP>(src_mdesc, dst_mdesc, num_send_bytes,
-                            local_expert_idx, (token_idx - offset + 1) % 4 == 0 ? UCP_DEVICE_FLAG_NODELAY : 0) == NIXL_IN_PROG);
+                            local_expert_idx, doorbell_flag(token_idx - offset)) == NIXL_IN_PROG);
                 }
             }
         }


examples/device/ep/nixl_ep/buffer.py
```diff
@@ -52,10 +52,9 @@ class Buffer:
 
     def __init__(
         self,
-        nvlink_backend: Literal["nixl", "ipc", "none"] = "nixl",
+        nvlink_backend: Literal["nixl", "none"] = "nixl",
```
itayalroy: Ideally I would like to rename it to `enable_nvlink` now that it only has 2 options but this will require changes to vLLM/SGLang PRs, and also this flag will not be relevant for ht kernels (where nvlink is mandatory) so perhaps `enable_ll_nvlink` or similar  

 @yosefe @ebarilanM wdyt?

examples/device/ep/csrc/kernels/nixl_ep.cu
```diff
@@ -34,6 +34,19 @@ namespace nixl_ep {
 
 namespace ep_kernels {
 
+__device__ inline size_t gpu_nixl_ctx::offset_get(uint64_t ptr) {
```
ebarilanM: why not uint64_t ?

examples/device/ep/csrc/kernels/nixl_ep.cu
```diff
@@ -170,13 +183,13 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
...
                 if (not is_rank_masked<true>(mask_buffer_ptr, dst_rank)) {
                     if (dst_p2p_ptr == 0) {
-                        nixlGpuXferReqH batch_req = nixl_ctx.batch_get(dst_rank);
-                        size_t src_offset = nixl_ctx.batch_offset_get(src_ptr);
-                        size_t dst_offset = nixl_ctx.batch_offset_get(dst_ptr);
-                        EP_DEVICE_ASSERT(nixlGpuPostSingleWriteXferReq<nixl_gpu_level_t::WARP>(batch_req, 0, src_offset, dst_offset, num_bytes_per_msg, dst_expert_local_idx, (slot_idx + 1) % 4 == 0) == NIXL_IN_PROG);
+                        nixlMemDesc src_mdesc{nixl_ctx.local_mvh, 0, nixl_ctx.offset_get(src_ptr)};
+                        nixlMemDesc dst_mdesc{nixl_ctx.remote_mvh, (size_t) dst_rank, nixl_ctx.offset_get(dst_ptr)};
+                        EP_DEVICE_ASSERT(nixlPut<nixl_gpu_level_t::WARP>(src_mdesc, dst_mdesc, num_bytes_per_msg,
+                                dst_expert_local_idx, (slot_idx + 1) % 4 == 0 ? UCP_DEVICE_FLAG_NODELAY : 0) == NIXL_IN_PROG);
```
ebarilanM: is there any nameing also to 0? like UCP_DEVICE_FLAG_NODB?

examples/device/ep/csrc/kernels/nixl_ep.cu
```diff
@@ -1109,26 +1120,27 @@ void clean_mask_buffer(int* mask_buffer_ptr, int num_ranks, cudaStream_t stream)
...
-    if (thread_id < num_ranks && thread_id != rank) {
+    if (thread_id < nixl_ctx.max_num_ranks && thread_id != nixl_ctx.rank) {
         const auto dst_rank = thread_id;
         if (not is_rank_masked(mask_buffer_ptr, dst_rank)) {
-            int expected_cnt = atomicAdd(nixl_ctx.local_barrier_cnt + dst_rank, -1) - 1;
+            int expected_cnt = atomicAdd(nixl_ctx.sync_count_ptr + dst_rank, -1) - 1;
 
-            nixlGpuXferReqH barrier_req = nixl_ctx.remote_barrier_get(dst_rank);
-            nixlGpuPostSingleWriteXferReq<nixl_gpu_level_t::THREAD>(barrier_req, 0, dst_rank * sizeof(int), rank * sizeof(int), sizeof(int), 0);
+            nixlMemDesc src_mdesc{nixl_ctx.local_mvh, 1, dst_rank * sizeof(int)};
```
ebarilanM, line 1135: do you want give enum for 
0 - NIXL_EP_DATABUF
1- NIXL_EP_BAR_BUF

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -659,47 +603,47 @@ std::string Buffer::get_local_metadata() const {
...
-    if (nixl_ctx->gpu.batch_reqs == nullptr)
-        CUDA_CHECK(cudaMalloc(&nixl_ctx->gpu.batch_reqs, max_num_ranks * sizeof(nixlGpuXferReqH)));
+void Buffer::_nixl_ep_memory_views_create(void) {
+    nixl_remote_dlist_t remote_descs(VRAM_SEG);
+    nixl_remote_dlist_t barrier_descs(VRAM_SEG);
+    nixl_xfer_dlist_t local_descs(VRAM_SEG);
 
-    // Always copy the updated batch arrays to GPU (since new ranks may have been added)
-    CUDA_CHECK(cudaMemcpy(nixl_ctx->gpu.batch_reqs, nixl_ctx->gpu_batch_reqs.data(), max_num_ranks * sizeof(nixlGpuXferReqH), cudaMemcpyHostToDevice));
+    local_descs.addDesc(nixlBlobDesc((uintptr_t)(rdma_buffer_ptr), num_rdma_bytes, get_local_device_id(), ""));
```
ebarilanM, line 611: if last param is not usful , "") can we ask nixl make it optional?

examples/device/ep/nixl_ep/buffer.py
ebarilanM: i perfer disable_ll_nvlink so it is clear that default is set

ebarilanM (dismissed):

examples/device/ep/csrc/kernels/nixl_ep.cu
itayalroy: They don't have it, I think we can discuss it because it might be a bit confusing that the default flags value (0) will never ring db

@yosefe @rakhmets 

Anyways this shouldnt block merge, can change in a separate PR

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: @yosefe @rakhmets wdyt? why must we use nixlBlobDescs and not nixlBasicDescs for reg dlist?

examples/device/ep/csrc/kernels/nixl_ep.cu
itayalroy: fixed

examples/device/ep/nixl_ep/buffer.py
itayalroy: Renamed to disable_ll_nvlink

examples/device/ep/csrc/kernels/nixl_ep.cu
itayalroy: I don't like "databuf" since it is data + counters, we can introduce an enum in another PR if we find a good name

examples/device/ep/csrc/kernels/nixl_ep.cu
ofirfarjun7: `(slot_idx + 1) % 4 == 0 ? UCP_DEVICE_FLAG_NODELAY : 0`
This code repeats in more places. Maybe use macro/inline.

examples/device/ep/csrc/kernels/nixl_ep.cu
```diff
@@ -170,13 +183,13 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
                                      dst_expert_local_idx * num_ranks * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                                      rank * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                                      slot_idx * num_bytes_per_msg;
-                const auto dst_p2p_ptr = nixl_ctx.rdma_p2p_ptr_get(dst_ptr, dst_rank);
+                void* dst_p2p_ptr = nixl_ctx.p2p_ptr_get(dst_ptr, dst_rank);
```
ofirfarjun7, line 190: Maybe call it inside the `is_rank_masked` if?
In general it is called in more places before `if`, so maybe better call it only when it is needed.

examples/device/ep/csrc/kernels/nixl_ep.cu
itayalroy: fixed

PR #1261, opened 2026-01-28
PLUGINS/UCX: Removed extra copy from prepMemoryView.

rakhmets:
## What?
Removed extra copy from `prepMemoryView` by passing descriptor lists to functions creating memory view.
Also added classes creating `ucp_device_mem_list_elem_t` and `ucp_device_mem_list_params_t` (https://github.com/ai-dynamo/nixl/pull/1245#discussion_r2727695377), and a couple of additional exceptions.

## Why?
Creating an additional collection can significantly reduce performance.

### DIFF
--- a/src/plugins/ucx/mem_list.cpp
+++ b/src/plugins/ucx/mem_list.cpp
@@ -58,11 +58,9 @@ private:
 
 class memListParams {
 public:
-    explicit memListParams(const device_mem_vector_t &elements) noexcept;
-    memListParams(const device_mem_vector_t &&) = delete;
-
-    void
-    setWorker(const nixlUcxWorker &worker) noexcept;
+    explicit memListParams(const device_mem_vector_t &elements,
+                           const nixlUcxWorker *worker = nullptr) noexcept;
+    memListParams(const device_mem_vector_t &&, const nixlUcxWorker *worker = nullptr) = delete;
 
     [[nodiscard]] const ucp_device_mem_list_params_t *
     get() const noexcept {
@@ -113,19 +111,18 @@ memListElement::create(const nixlMetaDesc &desc) {
     return element;
 }
 
-memListParams::memListParams(const device_mem_vector_t &elements) noexcept {
+memListParams::memListParams(const device_mem_vector_t &elements,
+                             const nixlUcxWorker *worker) noexcept {
     params_.field_mask = UCP_DEVICE_MEM_LIST_PARAMS_FIELD_ELEMENTS |
         UCP_DEVICE_MEM_LIST_PARAMS_FIELD_ELEMENT_SIZE |
         UCP_DEVICE_MEM_LIST_PARAMS_FIELD_NUM_ELEMENTS;
     params_.elements = elements.data();
     params_.element_size = sizeof(ucp_device_mem_list_elem_t);
     params_.num_elements = elements.size();
-}
-
-void
-memListParams::setWorker(const nixlUcxWorker &worker) noexcept {
-    params_.field_mask |= UCP_DEVICE_MEM_LIST_PARAMS_FIELD_WORKER;
-    params_.worker = worker.get();
+    if (worker) {
+        params_.field_mask |= UCP_DEVICE_MEM_LIST_PARAMS_FIELD_WORKER;
+        params_.worker = worker->get();
+    }
 }
 
 template<typename T>
@@ -162,8 +159,7 @@ createMemList(const nixl_remote_meta_dlist_t &dlist, size_t worker_id, nixlUcxWo
 void *
 createMemList(const nixl_meta_dlist_t &dlist, const nixlUcxWorker &worker) {
     const device_mem_vector_t elements = createElements(dlist);
-    memListParams params{elements};
-    params.setWorker(worker);
+    const memListParams params{elements, &worker};
 
     ucp_device_local_mem_list_h handle{nullptr};
     const auto status = ucp_device_local_mem_list_create(params.get(), &handle);


src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,135 @@
...
+#ifdef HAVE_UCX_GPU_DEVICE_API_V2
+namespace nixl::ucx {
+class memListElement {
+public:
+    memListElement(const nixlRemoteMetaDesc &desc, size_t worker_id)
+        : element_(create(desc, worker_id)) {}
+
+    memListElement(const nixlMetaDesc &desc, size_t) : element_(create(desc)) {}
+
+    [[nodiscard]] ucp_device_mem_list_elem_t
```
ColinNV: Can this be returned by reference? It's larger than what we would normally pass "by value".

src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,135 @@
...
+    element.memh = md->getMem().getMemh();
+    element.local_addr = md->getMem().getBase();
     return element;
 }
 
+memListParams::memListParams(const std::vector<ucp_device_mem_list_elem_t> &elements) noexcept {
+    params_.field_mask = UCP_DEVICE_MEM_LIST_PARAMS_FIELD_ELEMENTS |
+        UCP_DEVICE_MEM_LIST_PARAMS_FIELD_ELEMENT_SIZE |
+        UCP_DEVICE_MEM_LIST_PARAMS_FIELD_NUM_ELEMENTS;
+    params_.elements = elements.data();
```
ColinNV, line 119: Given that this keeps a pointer to data owned by a function argument one could be paranoid and delete the overload for constant r-value references. Not that it completely solves the potential issue, but it might be better than nothing.

rakhmets: Deleted an overload for a constant r-value reference.

src/plugins/ucx/mem_list.cpp
rakhmets: I moved ctors to private, and make get a public static member, since the purpose of the class is only to create `ucp_device_mem_list_elem_t`.

src/plugins/ucx/mem_list.cpp
```diff
@@ -122,18 +182,24 @@ releaseMemList(void *mvh) noexcept {
...
-createMemList(const std::vector<nixlUcxMem> &mems, const nixlUcxWorker &worker) {
-    throw std::runtime_error("UCX GPU device API is not supported");
+createMemList(const nixl_meta_dlist_t &, const nixlUcxWorker &) {
+    throw std::runtime_error(error_message);
 }
 
 void
-releaseMemList(void *mvh) noexcept {}
+releaseMemList(void *) noexcept {
+    NIXL_WARN << error_message;
```
ColinNV: There must be something _really_ wrong if this is called? How about `std::terminate()`?

src/plugins/ucx/mem_list.cpp
```diff
@@ -100,14 +161,13 @@ createMemList(const std::vector<std::unique_ptr<remoteMem>> &mems, nixlUcxWorker
 }
 
 void *
-createMemList(const std::vector<nixlUcxMem> &mems, const nixlUcxWorker &worker) {
-    const auto elements = createElements(mems);
-    auto params = addElements(elements);
-    params.field_mask |= UCP_DEVICE_MEM_LIST_PARAMS_FIELD_WORKER;
-    params.worker = worker.get();
+createMemList(const nixl_meta_dlist_t &dlist, const nixlUcxWorker &worker) {
+    const auto elements = createElements(dlist);
```
ColinNV: I'd replace `auto` with the actual type here.

src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,137 @@
...
-std::vector<ucp_device_mem_list_elem_t>
-createElements(const std::vector<T> &mems) {
+[[nodiscard]] std::vector<ucp_device_mem_list_elem_t>
+createElements(const T &dlist, size_t worker_id = 0) {
+    const auto desc_count = static_cast<size_t>(dlist.descCount());
     std::vector<ucp_device_mem_list_elem_t> elements;
-    for (const auto &mem : mems) {
-        elements.emplace_back(createElement(mem));
+    elements.reserve(desc_count);
+    for (size_t i = 0; i < desc_count; ++i) {
```
ColinNV: Range-based for-loop?

src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,137 @@
...
-        element.rkey = mem->rkey_.get();
-        element.remote_addr = mem->addr_;
-        element.ep = mem->ep_.getEp();
-    } else {
+    if (desc.remoteAgent == nixl_invalid_agent) {
         element.field_mask = 0;
+        return element;
     }
+
+    auto md = static_cast<const nixlUcxPublicMetadata *>(desc.metadataP);
```
ColinNV: `const`

src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,137 @@
...
+    element.rkey = md->getRkey(worker_id).get();
+    element.remote_addr = static_cast<uint64_t>(desc.addr);
+    element.ep = md->conn->getEp(worker_id)->getEp();
     return element;
 }
 
 ucp_device_mem_list_elem_t
-createElement(const nixlUcxMem &mem) {
+memListElement::create(const nixlMetaDesc &desc) {
+    auto md = static_cast<const nixlUcxPrivateMetadata *>(desc.metadataP);
```
ColinNV: `const`

src/plugins/ucx/mem_list.cpp
rakhmets: I'd prefer to keep it just `auto`, as `const` is already mentioned as a part of the template argument.

src/plugins/ucx/mem_list.cpp
rakhmets: The same as above.

src/plugins/ucx/mem_list.cpp
rakhmets: Fixed.

src/plugins/ucx/mem_list.cpp
rakhmets: Done.

src/plugins/ucx/mem_list.cpp
rakhmets: Not, really. I just wanted to keep it in the same way as on a higher level: `create` throws an exception in case of an error (equivalent to: print `NIXL_ERROR` and return NIXL error status); `release` prints `NIXL_WARN` in case of an error.

ColinNV: I'd expect this function to only be called with the result of a `create()`, which can't exist because that always throws?

src/plugins/ucx/mem_list.cpp
ColinNV: The `const` in the cast refers to the pointee, the `const` here would apply to the pointer.

rakhmets: ah, right, got it.

src/plugins/ucx/mem_list.cpp
rakhmets: right... yes, agree with you, that it is better to have here something more, than just a warning.

rakhmets: Done.

src/plugins/ucx/mem_list.cpp
rakhmets: Fixed.

src/plugins/ucx/mem_list.cpp
```diff
@@ -17,76 +17,136 @@
...
+    const ucp_device_mem_list_elem_t element_;
+};
+
+class memListParams {
+public:
+    explicit memListParams(const device_mem_vector_t &elements) noexcept;
+    memListParams(const device_mem_vector_t &&) = delete;
+
+    void
+    setWorker(const nixlUcxWorker &worker) noexcept;
```
brminich: why is it better than having constructor with `worker` param?

rakhmets: Replaced by ctor.

brminich (approved):

PR #1263, opened 2026-01-28
CI: Add Azurite, Azure SDK and AZ CLI

aranadive:
Azure SDK is required to build the Azure Blob Plugin. AZ CLI and Azurite are their testing tools to connect to Azure instance and Azure Storage emulation respectively. Azurite needs npm to install the node.js package so install that as well.

Azurite emulates their storage blob and other services. We can utilize this to test the plugin in our CI. Atleast, until we get an Azure service.

I also added an az cli based test to validate the azurite install.

[Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-install-azurite?tabs=npm%2Cblob-storage)
[Azure CLI](https://learn.microsoft.com/en-us/cli/azure/what-is-azure-cli?view=azure-cli-latest)

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -213,7 +213,7 @@ else
       $SUDO make all BUILD_RELEASE=1 BUILD_FOR_UNITEST=0 VERBOSE=1 ALLOW_USE_URING=0 && \
       $SUDO ldconfig && \
       cd .. && \
-      rm -rf gusli
+      $SUDO rm -rf gusli
     )
 
     ( \


.gitlab/build.sh
```diff
@@ -217,6 +219,21 @@ else
       mkdir -p ${INSTALL_DIR}/bin &&
       cp ${TMPDIR}/gtest-parallel/* ${INSTALL_DIR}/bin/
     )
+
+    ( \
+      cd ${TMPDIR} && \
+      curl -sL https://aka.ms/InstallAzureCLIDeb | $SUDO bash && \
+      $SUDO npm install -g azurite && \
```
Alexey-Rivkin: Should we pin the azurite ver to the latest stable?

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -42,7 +42,7 @@ env:
   TEST_TIMEOUT: 30
   UCX_TLS: "^shm"
   STORAGE_DRIVER: 'overlay'
-  CI_IMAGE_TAG: "20251218-1"
+  CI_IMAGE_TAG: "20250128-1"
```
Alexey-Rivkin: Should be "20260128-1" for unification

.gitlab/test_azure.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+AZ_ACCOUNT_NAME="nixl-ci-dev"
+AZ_ACCOUNT_KEY="ZGV2c3RvcmVhY2NvdW50Mw=="   # "devstoreaccount3" base64
+export AZURITE_ACCOUNTS="${AZ_ACCOUNT_NAME}:${AZ_ACCOUNT_KEY}"
```
Alexey-Rivkin, line 28: Should we use secret vars instead?

.gitlab/test_azure.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
```
Alexey-Rivkin, line 1: Without `-e` flag jobs may pass even if the script fails

.gitlab/test_azure.sh
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+AZ_ACCOUNT_NAME="nixl-ci-dev"
+AZ_ACCOUNT_KEY="ZGV2c3RvcmVhY2NvdW50Mw=="   # "devstoreaccount3" base64
+export AZURITE_ACCOUNTS="${AZ_ACCOUNT_NAME}:${AZ_ACCOUNT_KEY}"
+AZURITE_WORKDIR="./azdata"
+mkdir -p "${AZURITE_WORKDIR}"
+
+blob_port=$(get_next_tcp_port)
```
Alexey-Rivkin, line 32: Undefined. Along with the missing `-e`, it means that the test only LOOKS green

.gitlab/test_azure.sh
aranadive: These values are just entirely made up for azurite. Adding secret vars I think just adds more complexity.

.gitlab/build.sh
aranadive: Sure.

.gitlab/test_azure.sh
aranadive: Done. Added common.sh.

.gitlab/build.sh
```diff
@@ -167,7 +178,9 @@ else
                   --enable-efa && \
       make -j"$NPROC" && \
       make install && \
-      $SUDO ldconfig \
+      $SUDO ldconfig && \
+      cd .. && \
+      rm -rf libfabric-*
```
isdrk, line 183: Those extra steps for removal are not needed.
We have: ```$SUDO rm -rf ${TMPDIR}``` at the end.
 
https://github.com/aranadive/nixl/blob/main/.gitlab/build.sh#L259

aranadive: I decided to leave them since we were running out of space when building. I would suggest keeping them until we switch to the ephemeral containers.

PR #1265, opened 2026-02-01
Add Neuron Support to nixlbench

akkart-aws:
### What?
Adds AWS Neuron (Trainium/Inferentia) accelerator support to nixlbench for VRAM segment benchmarking.

Key changes:
- New neuron.cpp/h providing Neuron runtime API wrappers
- Dynamic loading of libnrt.so.1 via dlopen()/dlsym()
- Neuron memory operations: neuronMalloc(), neuronFree(), neuronMemcpy(), neuronMemset()
- Runtime device detection via neuronCoreCount()

### Why?
nixlbench only supported CUDA for VRAM benchmarking. This adds Neuron support while avoiding hard dependencies on libnrt.

Key benefit: We use dlopen()/dlsym() to dynamically resolve Neuron runtime symbols, so nixlbench works on systems without libnrt installed. Single binary runs on both CUDA and Neuron systems.

### How?
- Dynamic Library Loading
dlopen_libnrt() loads libnrt.so.1 at runtime (tries /opt/aws/neuron/lib/ then system path)
Template-based _load_nrt_symbol() resolves NRT functions via dlsym()
Returns nullptr if library unavailable (graceful fallback)
- Device Detection
neuronCoreCount() returns visible Neuron cores or -1 if unavailable
Initialized early in worker to ensure FI_HMEM_NEURON backend creation
Used throughout: if (devId < neuronCoreCount()) to route to Neuron vs CUDA paths
- Conditional Execution
```
// Memory allocation
if (devid < neuronCoreCount()) {
    return getVramDescNeuron(...);  // Use NRT tensors
}
// Fall through to CUDA

// Memory operations
if (iov.devId < neuronCoreCount()) {
    neuronMemcpy(...);  // Use NRT API
} else {
    cudaMemcpy(...);    // Use CUDA API
}
```
- Build Changes
Added dependency('dl') for dlopen/dlsym
No changes to CUDA dependencies (maintains compatibility)

### Related
Builds on main NIXL Neuron support which added FI_HMEM_NEURON backend and topology detection.

### DIFF
--- a/benchmark/nixlbench/src/utils/neuron.cpp
+++ b/benchmark/nixlbench/src/utils/neuron.cpp
@@ -1,0 +1,215 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2026 Amazon.com, Inc. and affiliates.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "neuron.h"
+
+#include <dlfcn.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+
+namespace {
+
+void *
+dlopen_libnrt() {
+#define TRY_DLOPEN(path)                       \
+    do {                                       \
+        void *handle = dlopen(path, RTLD_NOW); \
+        if (handle) return handle;             \
+    } while (0)
+
+    static void *const libnrt_handle = []() -> void * {
+        TRY_DLOPEN("/opt/aws/neuron/lib/libnrt.so.1");
+        TRY_DLOPEN("libnrt.so.1");
+        return nullptr;
+    }();
+
+#undef TRY_DLOPEN
+
+    return libnrt_handle;
+}
+
+template<class Fn>
+Fn *
+_load_nrt_symbol(const char *fn_name, Fn *) {
+    void *libnrt_handle = dlopen_libnrt();
+    if (libnrt_handle) {
+        return reinterpret_cast<Fn *>(dlsym(libnrt_handle, fn_name));
+    }
+    return nullptr;
+}
+
+#define LOAD_NRT_SYMBOL(sym) _load_nrt_symbol(#sym, &sym)
+
+int
+nrt_init(int framework, const char *fw_version, const char *fal_version) {
+    static const auto fn = LOAD_NRT_SYMBOL(nrt_init);
+    if (!fn) return -1;
+    return fn(framework, fw_version, fal_version);
+}
+
+struct nrt_tensor;
+
+int
+nrt_tensor_allocate(int tensor_placement,
+                    int vnc,
+                    size_t size,
+                    const char *name,
+                    nrt_tensor **tensor) {
+    static const auto fn = LOAD_NRT_SYMBOL(nrt_tensor_allocate);
+    if (!fn) return -1;
+    return fn(tensor_placement, vnc, size, name, tensor);
+}
+
+void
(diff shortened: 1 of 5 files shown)


benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -0,0 +1,215 @@
...
+neuronFree(void *addr) {
+    if (!addr) return 0;
+
+    std::lock_guard lock{allocation_tracker_mutex};
+    return allocation_tracker.erase(addr) - 1;
+}
+
+int
+neuronMemcpy(void *dest, const void *src, size_t count, neuronMemcpyKind kind) {
+    nrt_tensor *tensor = getTensorFromVA(neuronMemcpyHostToDevice ? dest : src);
```
ovidiusm: Was it intentional to not test `kind` here?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -0,0 +1,215 @@
...
+                    const char *name,
+                    nrt_tensor **tensor) {
+    static const auto fn = LOAD_NRT_SYMBOL(nrt_tensor_allocate);
+    if (!fn) return -1;
+    return fn(tensor_placement, vnc, size, name, tensor);
+}
+
+void
+nrt_tensor_free(nrt_tensor **tensor) {
+    static const auto fn = LOAD_NRT_SYMBOL(nrt_tensor_free);
```
ovidiusm, line 85: Why not if (!fn) return?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -425,12 +431,26 @@ getVramDescCudaVmm(int devid, size_t buffer_size, uint8_t memset_value) {
...
+    return std::optional<xferBenchIOV>(std::in_place, (uintptr_t)addr, buffer_size, devid);
+}
+
 static std::optional<xferBenchIOV>
 getVramDesc(int devid, size_t buffer_size, bool isInit) {
-    CHECK_CUDA_ERROR(cudaSetDevice(devid), "Failed to set device");
     uint8_t memset_value =
         isInit ? XFERBENCH_INITIATOR_BUFFER_ELEMENT : XFERBENCH_TARGET_BUFFER_ELEMENT;
 
+    if (devid < neuronCoreCount()) {
```
ovidiusm: How does this check work? How do we know that CUDA dev ids do not fall in this range?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -600,8 +620,12 @@ xferBenchNixlWorker::cleanupBasicDescDram(xferBenchIOV &iov) {
 #if HAVE_CUDA
 void
 xferBenchNixlWorker::cleanupBasicDescVram(xferBenchIOV &iov) {
-    CHECK_CUDA_ERROR(cudaSetDevice(iov.devId), "Failed to set device");
+    if (iov.devId < neuronCoreCount()) {
```
ovidiusm: Same as above

benchmark/nixlbench/src/utils/neuron.cpp
mcuiaws: this is copy-paste error. thanks for catching.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
mcuiaws: there are no AWS instance types with both nvidia GPUs and Neuron

brminich: can you pls make some more explicit check? This one looks error prone

benchmark/nixlbench/src/utils/neuron.cpp
mcuiaws: It's not really possible for nrt_tensor_allocate symbol to exist without nrt_tensor_free. Also this function does not return an error so the memory would just be leaked. Crashing here is arguably better than leaking.  We can turn it into explicit assert.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
mcuiaws: We can change to:
```
if (neuronCoreCount() > 0)
```
with comment saying assume no CUDA cores exist if neuron cores are found.

ovidiusm: I think that would be better

benchmark/nixlbench/src/utils/neuron.cpp
akkart-aws: Addressed this comment.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
akkart-aws: Done

PR #1272, opened 2026-02-03
libfabric: CQ batch reads and threading model

yexiang-aws:
## What?
- Change threading model from FI_THREAD_SAFE to FI_THREAD_COMPLETION
- Implement batch CQ reads (64 entries) instead of single-entry reads
- Remove CQ lock for non-blocking reads to reduce contention

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -431,7 +431,7 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
         hints->domain_attr->mr_key_size = 2;
     }
     hints->domain_attr->name = strdup(device_name.c_str());
-    hints->domain_attr->threading = FI_THREAD_SAFE;
+    hints->domain_attr->threading = FI_THREAD_COMPLETION;
 
     try {
         // Get fabric info for this specific device - first try with FI_HMEM


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -24,6 +24,8 @@
 #include <cstring>
 #include <stdexcept>
 #include <stack>
+#include <thread>
+#include <chrono>
```
fengjica: Is the chrono header needed for any of the code change below?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -746,19 +748,18 @@ nixlLibfabricRail::progressCompletionQueue() const {
...
-        if (status != NIXL_SUCCESS) {
-            NIXL_ERROR << "Failed to process completion on rail " << rail_id;
-            return status;
+    if (ret > 0) {
+        for (int i = 0; i < ret; ++i) {
+            // Process completion using local data. Callbacks have their own thread safety
+            nixl_status_t status = processCompletionQueueEntry(&completions[i]);
+            if (status != NIXL_SUCCESS) {
+                NIXL_ERROR << "Failed to process completion " << i << " out of batch of "
+                           << NIXL_LIBFABRIC_CQ_BATCH_SIZE << " on rail " << rail_id;
```
fengjica: nit: is it better to log `ret`, instead of NIXL_LIBFABRIC_CQ_BATCH_SIZE, here?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -431,7 +433,8 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
         hints->domain_attr->mr_key_size = 2;
     }
     hints->domain_attr->name = strdup(device_name.c_str());
-    hints->domain_attr->threading = FI_THREAD_SAFE;
+    hints->domain_attr->threading = FI_THREAD_COMPLETION;
```
fengjica, line 434: IIUC, this will require our code to "serialize access to all objects that are associated through the use of having a shared completion structure. This includes endpoint, completion queue, counter, wait set, and poll set objects" (from the man page). Is this the status every where in the code today?
Or maybe just running through our full test cases to make sure it's good?

fengjica (dismissed):

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -704,13 +707,12 @@ nixlLibfabricRail::setXferIdCallback(std::function<void(uint32_t)> callback) {
 }
 
 // Per-Rail Completion Processing
-
+#define NIXL_LIBFABRIC_CQ_BATCH_SIZE 64
```
amitrad-aws: Let's set to 16

src/utils/libfabric/libfabric_rail.cpp
amitrad-aws: We have the cq_progress_mutex_ that protects this

amitrad-aws (changes requested):
I think this should be two seperate commits, they are unrelated

src/utils/libfabric/libfabric_rail.cpp
yexiang-aws: good catch, it's not needed

yexiangd: good catch, both are not needed

src/utils/libfabric/libfabric_rail.cpp
yexiangd: done

src/utils/libfabric/libfabric_rail.cpp
yexiangd: done, validate no difference between 8, 64 and 256 on both VRAM and DRAM tests with single thread.

PR #1274, opened 2026-02-03
API: Add getXferStatus overload with per-entry status codes

vvenkates27:
## What?
Add per-entry status tracking for batch transfers via a new getXferStatus overload that returns individual status codes for each transfer entry.
Changes include:
New getXferStatus(nixlXferReqH*, std::vector<nixl_status_t>&) API overload in core
New NIXL_IN_PROG_WITH_ERR status indicating partial completion (some entries failed while others still in progress)
S3/Object backend implementation using shared_future pattern for checkXferList
Rust bindings: get_xfer_status_list() method and XferStatus::InProgressWithError variant
Python bindings: getXferStatusList() method returning (overall_status, entry_status_list)

## Why?
The existing getXferStatus() API only returns an overall status for batch transfers. For large batch operations, users need finer-grained visibility into which specific entries have completed, failed, or are still in progress. This enables:
Early error detection without waiting for entire batch completion
Progress tracking for long-running batch transfers
Targeted retry logic for failed entries

## How?
Core API: Added an overload that accepts a std::vector<nixl_status_t>& output parameter populated with per-entry status codes
Status codes: NIXL_IN_PROG_WITH_ERR (value 2) indicates the batch is still in progress but at least one entry has failed
S3 Backend: Uses std::shared_future to track individual async operations and report per-entry completion
Bindings: C API wrapper (nixl_capi_get_xfer_status_list) bridges to language bindings with appropriate error mapping
Fallback: Backends that don't support per-entry status return NIXL_ERR_NOT_SUPPORTED, allowing callers to fall back to regular getXferStatus()
@mkhazraee, @yosefe 

## Summary by CodeRabbit

* **New Features**
  * Per-entry transfer event tracking: append-only events list, new in-progress-with-error status, and tracking flags for errors/successes.
  * APIs and bindings: new overloads and helpers to poll per-entry events; Python and Rust expose reusable event containers, track flag constants, and transfer-creation options to enable tracking.
* **Tests**
  * New unit and integration tests covering per-entry polling, event caching, shared-promise behavior, and partial-failure scenarios.

### DIFF
--- a/src/api/cpp/backend/backend_aux.h
+++ b/src/api/cpp/backend/backend_aux.h
@@ -33,6 +33,10 @@ struct nixlBackendOptionalArgs {
     nixl_blob_t notifMsg;
     bool        hasNotif = false;
     nixl_blob_t customParam;
+    /** Per-entry tracking mode for this transfer. Default 0 = no per-entry events.
+     *  Set NIXL_XFER_TRACK_ERRORS to append error events, NIXL_XFER_TRACK_SUCCESSES
+     *  to also append success events. Propagated from nixlAgentOptionalArgs::trackFlags
+     *  by the agent during prepXfer/postXfer. */
     nixl_xfer_track_flags_t trackFlags = 0;
 };
 
--- a/src/api/cpp/backend/backend_aux.h
+++ b/src/api/cpp/backend/backend_aux.h
@@ -33,6 +33,7 @@ struct nixlBackendOptionalArgs {
     nixl_blob_t notifMsg;
     bool        hasNotif = false;
     nixl_blob_t customParam;
+    nixl_xfer_track_flags_t trackFlags = 0;
 };
 
 using nixl_opt_b_args_t = nixlBackendOptionalArgs;
--- a/src/api/cpp/backend/backend_aux.h
+++ b/src/api/cpp/backend/backend_aux.h
@@ -27,11 +27,10 @@
 // level direction or so.
 typedef std::vector<std::pair<std::string, std::string>> notif_list_t;
 
-
 struct nixlBackendOptionalArgs {
     // During postXfer, user might ask for a notification if supported
     nixl_blob_t notifMsg;
-    bool        hasNotif = false;
+    bool hasNotif = false;
     nixl_blob_t customParam;
     /** Per-entry tracking mode for this transfer. Default 0 = no per-entry events.
      *  Set NIXL_XFER_TRACK_ERRORS to append error events, NIXL_XFER_TRACK_SUCCESSES
@@ -42,43 +41,42 @@ struct nixlBackendOptionalArgs {
 
 using nixl_opt_b_args_t = nixlBackendOptionalArgs;
 
-
 // A base class to point to backend initialization data
 // User doesn't know about fields such as local_agent but can access it
 // after the backend is initialized by agent. If we needed to make it private
 // from the user, we should make nixlBackendEngine/nixlAgent friend classes.
 class nixlBackendInitParams {
-    public:
-        std::string       localAgent;
+public:
+    std::string localAgent;
 
-        nixl_backend_t    type;
-        nixl_b_params_t*  customParams;
+    nixl_backend_t type;
+    nixl_b_params_t *customParams;
 
-        bool              enableProgTh;
-        nixlTime::us_t    pthrDelay;
-        nixl_thread_sync_t syncMode;
-        bool enableTelemetry_;
+    bool enableProgTh;
+    nixlTime::us_t pthrDelay;
+    nixl_thread_sync_t syncMode;
+    bool enableTelemetry_;
 };
 
 // Pure virtual class to have a common pointer type
 class nixlBackendReqH {
 public:
-    nixlBackendReqH() { }
-    virtual ~nixlBackendReqH() { }
+    nixlBackendReqH() {}
+
+    virtual ~nixlBackendReqH() {}
 };
 
 // Pure virtual class to have a common pointer type for different backendMD.
 class nixlBackendMD {
-    protected:
-        bool isPrivateMD;
+protected:
+    bool isPrivateMD;
 
-    public:
-        nixlBackendMD(bool isPrivate){
-            isPrivateMD = isPrivate;
-        }
+public:
+    nixlBackendMD(bool isPrivate) {
+        isPrivateMD = isPrivate;
+    }
 
-        virtual ~nixlBackendMD(){
-        }
+    virtual ~nixlBackendMD() {}
 };
 
 // Each backend can have different connection requirement
@@ -86,38 +84,40 @@ class nixlBackendMD {
 // a connection to a remote node. Note that local information
 // is passed during the constructor and through BackendInitParams
 class nixlBackendConnMD {
-  public:
+public:
     // And some other details
     std::string dstIpAddress;
-    uint16_t    dstPort;
+    uint16_t dstPort;
 };
 
 // A pointer required to a metadata object for backends next to each BasicDesc
 class nixlMetaDesc : public nixlBasicDesc {
-  public:
-        // To be able to point to any object
-        nixlBackendMD* metadataP;
+public:
+    // To be able to point to any object
+    nixlBackendMD *metadataP;
 
-        // Reuse parent constructor without the metadata pointer
-        using nixlBasicDesc::nixlBasicDesc;
+    // Reuse parent constructor without the metadata pointer
+    using nixlBasicDesc::nixlBasicDesc;
 
-        nixlMetaDesc() : nixlBasicDesc() { metadataP = nullptr; }
+    nixlMetaDesc() : nixlBasicDesc() {
+        metadataP = nullptr;
+    }
 
-        nixlMetaDesc(uintptr_t addr, size_t len, uint64_t dev_id, nixlBackendMD *metadata)
-            : nixlBasicDesc(addr, len, dev_id),
-              metadataP(metadata) {}
+    nixlMetaDesc(uintptr_t addr, size_t len, uint64_t dev_id, nixlBackendMD *metadata)
+        : nixlBasicDesc(addr, len, dev_id),
+          metadataP(metadata) {}
 
-        // No serializer or deserializer, using parent not to expose the metadata
+    // No serializer or deserializer, using parent not to expose the metadata
 
-        inline friend bool operator==(const nixlMetaDesc &lhs, const nixlMetaDesc &rhs) {
-            return (((nixlBasicDesc)lhs == (nixlBasicDesc)rhs) &&
-                          (lhs.metadataP == rhs.metadataP));
-        }
+    inline friend bool
+    operator==(const nixlMetaDesc &lhs, const nixlMetaDesc &rhs) {
+        return (((nixlBasicDesc)lhs == (nixlBasicDesc)rhs) && (lhs.metadataP == rhs.metadataP));
+    }
 
-        inline void print(const std::string &suffix) const {
-            nixlBasicDesc::print(", Backend ptr val: " +
-                                 std::to_string((uintptr_t)metadataP) + suffix);
-        }
+    inline void
+    print(const std::string &suffix) const {
+        nixlBasicDesc::print(", Backend ptr val: " + std::to_string((uintptr_t)metadataP) + suffix);
+    }
 };
 
 struct nixlRemoteMetaDesc : public nixlMetaDesc {
--- a/src/api/cpp/backend/backend_engine.h
+++ b/src/api/cpp/backend/backend_engine.h
@@ -149,6 +149,18 @@ class nixlBackendEngine {
         // Use a handle to progress backend engine and see if a transfer is completed or not
         virtual nixl_status_t checkXfer(nixlBackendReqH* handle) const = 0;
 
+        /**
+         * Optional: append new (index, status) events and return overall status.
+         * Default returns NIXL_ERR_NOT_SUPPORTED. Track mode is stored in handle from prep/post.
+         */
+        virtual nixl_status_t
+        checkXferEvents(nixlBackendReqH* handle,
+                        nixl_xfer_entry_events_t &events) const {
+            (void)handle;
+            (void)events;
+            return NIXL_ERR_NOT_SUPPORTED;
+        }
+
         //Backend aborts the transfer if necessary, and destructs the relevant objects
         virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
 
--- a/src/api/cpp/backend/backend_engine.h
+++ b/src/api/cpp/backend/backend_engine.h
@@ -154,9 +154,7 @@ class nixlBackendEngine {
          * Default returns NIXL_ERR_NOT_SUPPORTED. Track mode is stored in handle from prep/post.
          */
         virtual nixl_status_t
-        checkXferEvents(nixlBackendReqH *handle, nixl_xfer_entry_events_t &events) const {
-            (void)handle;
-            (void)events;
+        checkXferEvents(nixlBackendReqH *, nixl_xfer_entry_events_t &) const {
             return NIXL_ERR_NOT_SUPPORTED;
         }
 
--- a/src/api/cpp/backend/backend_engine.h
+++ b/src/api/cpp/backend/backend_engine.h
@@ -154,8 +154,7 @@ class nixlBackendEngine {
          * Default returns NIXL_ERR_NOT_SUPPORTED. Track mode is stored in handle from prep/post.
          */
         virtual nixl_status_t
-        checkXferEvents(nixlBackendReqH* handle,
-                        nixl_xfer_entry_events_t &events) const {
+        checkXferEvents(nixlBackendReqH *handle, nixl_xfer_entry_events_t &events) const {
             (void)handle;
             (void)events;
             return NIXL_ERR_NOT_SUPPORTED;
--- a/src/api/cpp/backend/backend_engine.h
+++ b/src/api/cpp/backend/backend_engine.h
@@ -31,228 +31,258 @@ constexpr size_t MAX_TELEMETRY_QUEUE_SIZE = 1000;
 
 // Base backend engine class for different backend implementations
 class nixlBackendEngine {
-    private:
-        // Members that cannot be modified by a child backend and parent bookkeep
-        nixl_backend_t  backendType;
-        nixl_b_params_t customParams;
-        std::vector<nixlTelemetryEvent> telemetryEvents_;
-        std::mutex telemetryEventsMutex_;
+private:
+    // Members that cannot be modified by a child backend and parent bookkeep
+    nixl_backend_t backendType;
+    nixl_b_params_t customParams;
+    std::vector<nixlTelemetryEvent> telemetryEvents_;
+    std::mutex telemetryEventsMutex_;
 
-    protected:
-        // Members that can be accessed by the child (localAgent cannot be modified)
-        bool              initErr = false;
-        const std::string localAgent;
-        const bool enableTelemetry_;
+protected:
+    // Members that can be accessed by the child (localAgent cannot be modified)
+    bool initErr = false;
+    const std::string localAgent;
+    const bool enableTelemetry_;
 
-        [[nodiscard]] nixl_status_t
-        setInitParam(const std::string &key, const std::string &value) {
-            if (customParams.emplace(key, value).second) {
-                return NIXL_SUCCESS;
-            }
-            return NIXL_ERR_NOT_ALLOWED;
+    [[nodiscard]] nixl_status_t
+    setInitParam(const std::string &key, const std::string &value) {
+        if (customParams.emplace(key, value).second) {
+            return NIXL_SUCCESS;
         }
+        return NIXL_ERR_NOT_ALLOWED;
+    }
 
-        [[nodiscard]] nixl_status_t getInitParam(const std::string &key, std::string &value) const {
-            const auto iter = customParams.find(key);
-            if (iter != customParams.end()) {
-                value = iter->second;
-                return NIXL_SUCCESS;
-            }
-            return NIXL_ERR_INVALID_PARAM;
+    [[nodiscard]] nixl_status_t
+    getInitParam(const std::string &key, std::string &value) const {
+        const auto iter = customParams.find(key);
+        if (iter != customParams.end()) {
+            value = iter->second;
+            return NIXL_SUCCESS;
         }
+        return NIXL_ERR_INVALID_PARAM;
+    }
 
-        void
-        addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) {
-            if (!enableTelemetry_) return;
-            if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) return;
-            std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
-            telemetryEvents_.emplace_back(
-                nixl_telemetry_category_t::NIXL_TELEMETRY_BACKEND, event_type, value);
+    void
+    addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) {
+        if (!enableTelemetry_) {
+            return;
         }
-
-    public:
-        explicit nixlBackendEngine(const nixlBackendInitParams *init_params)
-            : backendType(init_params->type),
-              customParams(*init_params->customParams),
-              localAgent(init_params->localAgent),
-              enableTelemetry_(init_params->enableTelemetry_) {}
-
-        nixlBackendEngine(nixlBackendEngine&&) = delete;
-        nixlBackendEngine(const nixlBackendEngine&) = delete;
-
-        void operator=(nixlBackendEngine&&) = delete;
-        void operator=(const nixlBackendEngine&) = delete;
-
-        virtual ~nixlBackendEngine() = default;
-
-        std::vector<nixlTelemetryEvent>
-        getTelemetryEvents() {
-            std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
-            return std::move(telemetryEvents_);
+        if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) {
+            return;
         }
+        std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
+        telemetryEvents_.emplace_back(
+            nixl_telemetry_category_t::NIXL_TELEMETRY_BACKEND, event_type, value);
+    }
 
-        bool getInitErr() const noexcept { return initErr; }
-        const nixl_backend_t& getType() const noexcept { return backendType; }
-        const nixl_b_params_t& getCustomParams() const noexcept { return customParams; }
+public:
+    explicit nixlBackendEngine(const nixlBackendInitParams *init_params)
+        : backendType(init_params->type),
+          customParams(*init_params->customParams),
+          localAgent(init_params->localAgent),
+          enableTelemetry_(init_params->enableTelemetry_) {}
 
-        // The support function determine which methods are necessary by the child backend, and
-        // if they're called by mistake, they will return error if not implemented by backend.
+    nixlBackendEngine(nixlBackendEngine &&) = delete;
+    nixlBackendEngine(const nixlBackendEngine &) = delete;
 
-        // Determines if a backend supports remote operations
-        virtual bool supportsRemote() const = 0;
+    void
+    operator=(nixlBackendEngine &&) = delete;
+    void
+    operator=(const nixlBackendEngine &) = delete;
 
-        // Determines if a backend supports local operations
-        virtual bool supportsLocal() const = 0;
+    virtual ~nixlBackendEngine() = default;
 
-        // Determines if a backend supports sending notifications. Related methods are not
-        // pure virtual, and return errors, as parent shouldn't call if supportsNotif is false.
-        virtual bool supportsNotif() const = 0;
+    std::vector<nixlTelemetryEvent>
+    getTelemetryEvents() {
+        std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
+        return std::move(telemetryEvents_);
+    }
 
-        virtual nixl_mem_list_t getSupportedMems() const = 0;  // TODO: Return by const-reference and mark noexcept?
+    bool
+    getInitErr() const noexcept {
+        return initErr;
+    }
 
+    const nixl_backend_t &
+    getType() const noexcept {
+        return backendType;
+    }
 
-        // *** Pure virtual methods that need to be implemented by any backend *** //
+    const nixl_b_params_t &
+    getCustomParams() const noexcept {
+        return customParams;
+    }
 
-        // Register and deregister local memory
-        virtual nixl_status_t registerMem (const nixlBlobDesc &mem,
-                                           const nixl_mem_t &nixl_mem,
-                                           nixlBackendMD* &out) = 0;
-        virtual nixl_status_t deregisterMem (nixlBackendMD* meta) = 0;
+    // The support function determine which methods are necessary by the child backend, and
+    // if they're called by mistake, they will return error if not implemented by backend.
 
-        // Make connection to a remote node identified by the name into loaded conn infos
-        // Child might just return 0, if making proactive connections are not necessary.
-        // An agent might need to connect to itself for local operations.
-        virtual nixl_status_t connect(const std::string &remote_agent) = 0;
-        virtual nixl_status_t disconnect(const std::string &remote_agent) = 0;
+    // Determines if a backend supports remote operations
+    virtual bool
+    supportsRemote() const = 0;
 
-        // Remove loaded local or remote metadata for target
-        virtual nixl_status_t unloadMD (nixlBackendMD* input) = 0;
+    // Determines if a backend supports local operations
+    virtual bool
+    supportsLocal() const = 0;
 
-        // Preparing a request, which populates the async handle as desired
-        virtual nixl_status_t prepXfer (const nixl_xfer_op_t &operation,
-                                        const nixl_meta_dlist_t &local,
-                                        const nixl_meta_dlist_t &remote,
-                                        const std::string &remote_agent,
-                                        nixlBackendReqH* &handle,
-                                        const nixl_opt_b_args_t* opt_args=nullptr
-                                       ) const = 0;
+    // Determines if a backend supports sending notifications. Related methods are not
+    // pure virtual, and return errors, as parent shouldn't call if supportsNotif is false.
+    virtual bool
+    supportsNotif() const = 0;
 
-        // Posting a request, which completes the async handle creation and posts it
-        virtual nixl_status_t postXfer (const nixl_xfer_op_t &operation,
-                                        const nixl_meta_dlist_t &local,
-                                        const nixl_meta_dlist_t &remote,
-                                        const std::string &remote_agent,
-                                        nixlBackendReqH* &handle,
-                                        const nixl_opt_b_args_t* opt_args=nullptr
-                                       ) const = 0;
-
-        // Use a handle to progress backend engine and see if a transfer is completed or not
-        virtual nixl_status_t checkXfer(nixlBackendReqH* handle) const = 0;
-
-        /**
-         * Optional: append new (index, status) events and return overall status.
-         * Default returns NIXL_ERR_NOT_SUPPORTED. Track mode is stored in handle from prep/post.
-         */
-        virtual nixl_status_t
-        checkXferEvents(nixlBackendReqH *, nixl_xfer_entry_events_t &) const {
-            return NIXL_ERR_NOT_SUPPORTED;
-        }
-
-        //Backend aborts the transfer if necessary, and destructs the relevant objects
-        virtual nixl_status_t releaseReqH(nixlBackendReqH* handle) const = 0;
-
-        // Prepare a memory view for remote buffers
-        virtual nixl_status_t
-        prepMemView(const nixl_remote_meta_dlist_t &,
-                    nixlMemViewH &,
-                    const nixl_opt_b_args_t * = nullptr) const {
-            return NIXL_ERR_NOT_SUPPORTED;
-        }
-
-        // Prepare a memory view for local buffers
-        virtual nixl_status_t
-        prepMemView(const nixl_meta_dlist_t &,
-                    nixlMemViewH &,
-                    const nixl_opt_b_args_t * = nullptr) const {
-            return NIXL_ERR_NOT_SUPPORTED;
-        }
-
-        // Release memory view handle
-        virtual void
-        releaseMemView(nixlMemViewH) const {}
-
-        // *** Needs to be implemented if supportsRemote() is true *** //
-
-        // Gets serialized form of public metadata
-        virtual nixl_status_t getPublicData (const nixlBackendMD* meta,
-                                             std::string &str) const {
-            return NIXL_ERR_BACKEND;
-        };
-
-        // Provide the required connection info for remote nodes, should be non-empty
-        virtual nixl_status_t getConnInfo(std::string &str) const {
-            return NIXL_ERR_BACKEND;
-        }
-
-        // Deserialize from string the connection info for a remote node, if supported
-        // The generated data should be deleted in nixlBackendEngine destructor
-        virtual nixl_status_t loadRemoteConnInfo (const std::string &remote_agent,
-                                                  const std::string &remote_conn_info) {
-            return NIXL_ERR_BACKEND;
-        }
-
-        // Load remote metadata, if supported.
-        virtual nixl_status_t loadRemoteMD (const nixlBlobDesc &input,
-                                            const nixl_mem_t &nixl_mem,
-                                            const std::string &remote_agent,
-                                            nixlBackendMD* &output) {
-            return NIXL_ERR_BACKEND;
-        }
+    virtual nixl_mem_list_t
+    getSupportedMems() const = 0; // TODO: Return by const-reference and mark noexcept?
 
 
-        // *** Needs to be implemented if supportsLocal() is true *** //
+    // *** Pure virtual methods that need to be implemented by any backend *** //
 
-        // Provide the target metadata necessary for local operations, if supported
-        virtual nixl_status_t loadLocalMD (nixlBackendMD* input,
-                                           nixlBackendMD* &output) {
-            return NIXL_ERR_BACKEND;
-        }
+    // Register and deregister local memory
+    virtual nixl_status_t
+    registerMem(const nixlBlobDesc &mem, const nixl_mem_t &nixl_mem, nixlBackendMD *&out) = 0;
+    virtual nixl_status_t
+    deregisterMem(nixlBackendMD *meta) = 0;
 
+    // Make connection to a remote node identified by the name into loaded conn infos
+    // Child might just return 0, if making proactive connections are not necessary.
+    // An agent might need to connect to itself for local operations.
+    virtual nixl_status_t
+    connect(const std::string &remote_agent) = 0;
+    virtual nixl_status_t
+    disconnect(const std::string &remote_agent) = 0;
 
-        // *** Needs to be implemented if supportsNotif() is true *** //
+    // Remove loaded local or remote metadata for target
+    virtual nixl_status_t
+    unloadMD(nixlBackendMD *input) = 0;
 
-        // Populate an empty received notif list. Elements are released within backend then.
-        virtual nixl_status_t getNotifs(notif_list_t &notif_list) { return NIXL_ERR_BACKEND; }
+    // Preparing a request, which populates the async handle as desired
+    virtual nixl_status_t
+    prepXfer(const nixl_xfer_op_t &operation,
+             const nixl_meta_dlist_t &local,
+             const nixl_meta_dlist_t &remote,
+             const std::string &remote_agent,
+             nixlBackendReqH *&handle,
+             const nixl_opt_b_args_t *opt_args = nullptr) const = 0;
 
-        // Generates a standalone notification, not bound to a transfer.
-        virtual nixl_status_t genNotif(const std::string &remote_agent, const std::string &msg) const {
-            return NIXL_ERR_BACKEND;
-        }
+    // Posting a request, which completes the async handle creation and posts it
+    virtual nixl_status_t
+    postXfer(const nixl_xfer_op_t &operation,
+             const nixl_meta_dlist_t &local,
+             const nixl_meta_dlist_t &remote,
+             const std::string &remote_agent,
+             nixlBackendReqH *&handle,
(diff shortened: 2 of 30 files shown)


src/api/cpp/nixl.h
```diff
@@ -287,6 +287,30 @@ class nixlAgent {
...
+         * @return nixl_status_t  Overall status:
+         *                        - NIXL_IN_PROG if any entry still in progress, no failures yet
+         *                        - NIXL_IN_PROG_WITH_ERR if any entry still in progress and at least one failed
+         *                        - NIXL_SUCCESS if all entries completed successfully
+         *                        - Error code if all entries completed and at least one failed
+         *                        - NIXL_ERR_NOT_SUPPORTED if backend doesn't support per-entry status
+         */
+        nixl_status_t
+        getXferStatus (nixlXferReqH* req_hndl,
+                       std::vector<nixl_status_t> &entry_status) const;
```
rakhmets: It seems that the following API would be enough for the purpose:
```suggestion
        std::vector<nixl_status_t>
        getXferStatus(nixlXferReqH& req_hndl) const;
```
So, you will not have method that returns two values: a vector of values, an a value summarizing the content of the vector.

vvenkates27: @rakhmets  Returning the status vector from getXferStatus will break backward compatibility, so we should have a separate API

mkhazraee: I expected std::vector<nixl_query_resp_t> &resp to be the output like queryMem, missed this, but I see since this is extension of xfer status we did status_t. Not sure it's better to extend the query_resp usage here to show success or a string of what went wrong, or keep it as status_t.

Regarding the output of single status vs vector, I say let's keep it as is, for three reasons. First, if the status happens to be success, we don't want to go over all entries. Second, user can also swap using their normal getXferStatus with this and keep the same functionality. Third, almost all of our APIs return single status, so good to keep that.

meson.build
```diff
@@ -258,6 +258,7 @@ endif
 # UCX GPU device API detection
 nvcc_prog = find_program('nvcc', required: false)
 ucx_gpu_device_api_available = false
+ucx_gpu_device_api_v2_available = false
```
brminich: is it unrelated fix? Can you pls submit in a separate PR then?

src/api/cpp/nixl.h
rakhmets: The main problem with this approach is that the same information is duplicated twice.
If the status is successful, there is no need to return a vector filled with `NIXL_SUCCESS`.
If I understood the idea correctly, then the user is only interested in transfers that ended with an error to then restart them, for example. Then it's better to return this information. So, the method will return a single status and information about failed transfers.
With this approach we don't need `NIXL_IN_PROG_WITH_ERR` this can be replaced by `NIXL_IN_PROG`, and information about failed transfers.

brminich: so it means that the polling should be done by checking the whole vector.
Imo returning a single status (like it is now) is better, because we can poll/wait until it returns something not equal to `IN_PROGRESS`

src/api/cpp/nixl_types.h
```diff
@@ -51,6 +51,7 @@ enum nixl_xfer_op_t {NIXL_READ, NIXL_WRITE};
  * @brief  An enumeration of status values and error codes for NIXL
  */
 enum nixl_status_t {
+    NIXL_IN_PROG_WITH_ERR = 2,  // In progress but at least one entry has failed
```
brminich: is it really needed? Can we just return `IN_PROG` until all statuses are known?
What is the use case to check statuses while none of them are completed yet?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -95,6 +95,8 @@ throw_nixl_exception(const nixl_status_t &status) {
     switch (status) {
     case NIXL_IN_PROG:
         return; // not an error
+    case NIXL_IN_PROG_WITH_ERR:
```
brminich, line 130: it is a bit confusing, because some error indeed occured

src/core/nixl_agent.cpp
```diff
@@ -1171,6 +1171,59 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+        if (data->telemetryEnabled) {
+            if (status == NIXL_SUCCESS) {
+                req_hndl->updateRequestStats(data->telemetry_, NIXL_TELEMETRY_FINISH);
+            } else if (status < 0) {
+                data->addErrorTelemetry(status);
+            }
+        }
+    }
+
+    return status;
```
brminich: minor:
```suggestion
    // Check if transfer is still in progress (either fully or with some errors)
    if ((status == NIXL_IN_PROG || status == NIXL_IN_PROG_WITH_ERR)) return status;

    // Update overall request status and handle errors/completion
    req_hndl->status = status;

    if (status < 0) {
        if (status == NIXL_ERR_REMOTE_DISCONNECT) {
            data->invalidateRemoteData(req_hndl->remoteAgent);
            return NIXL_ERR_REMOTE_DISCONNECT;
        } else {
            NIXL_ERROR_FUNC << "backend '" << req_hndl->engine->getType()
                            << "' returned error status " << req_hndl->status;
        }
    }

    // Update telemetry
    if (data->telemetryEnabled) {
       if (status == NIXL_SUCCESS) {
            req_hndl->updateRequestStats(data->telemetry_, NIXL_TELEMETRY_FINISH);
        } else if (status < 0) {
            data->addErrorTelemetry(status);
        }
    }
  
    return status;
```

src/api/cpp/nixl.h
rakhmets: Yes, I agree with returning a single status. In addition, the current proposed API returns a vector containing the statuses of all transfers associated with the request.
This means that in case of successful completion, the user will receive, in addition to `getXferStatus` method execution status, a vector of size N filled with `NIXL_SUCCESS`es.
And if you have 64k transfers in the batch, and only one failed, the user will still get a vector with (64k-1) `NIXL_SUCCESS`es.
I'm suggesting to return only information about failed transfers along with status returned by `getXferStatus`.
So, there will be the following cases:
1. `getXferStatus` returns `NIXL_SUCCESS`, the output about the failed transfers is empty.
2.  `getXferStatus` returns `NIXL_IN_PROG`,  the output about the failed transfers may contain the data.
3.  `getXferStatus` returns error code, the output about the failed transfers may contain the data.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -691,6 +694,30 @@ PYBIND11_MODULE(_bindings, m) {
...
             py::call_guard<py::gil_scoped_release>())
+        .def(
+            "getXferStatusList",
+            [](nixlAgent &agent,
+               uintptr_t reqh) -> std::pair<nixl_status_t, std::vector<nixl_status_t>> {
+                std::vector<nixl_status_t> entry_status;
+                nixl_status_t ret = agent.getXferStatus((nixlXferReqH *)reqh, entry_status);
+                // Don't throw for in-progress statuses
+                if (ret < 0 && ret != NIXL_ERR_NOT_SUPPORTED) {
+                    throw_nixl_exception(ret);
```
yanok: I think throwing an exception from a "get status" API (even the old one) is a questionable solution: we are already returning a `nixl_status_t`, why throw?

But the new API is simply useless with this throw: we will only ever get back `entry_status` if it's full of `NIXL_SUCCESS`, if there were some failures, overall status will be an error, so we will throw and `entry_status` is gone.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -691,6 +694,30 @@ PYBIND11_MODULE(_bindings, m) {
                 return ret;
             },
             py::call_guard<py::gil_scoped_release>())
+        .def(
+            "getXferStatusList",
```
yanok: For `getXferStatus` there is a higher-level counter-part `check_xfer_state` in `api/python/_api.py`.
1. If `NIXL_IN_PROG_WITH_ERR` is going to stay, it needs to be updated. Currently it will return "ERR" if `getXferStatus` returns `NIXL_IN_PROG_WITH_ERR`.
2. Do we need a higher-level counterpart for `getXferStatusList` as well?

src/bindings/python/nixl_bindings.cpp
```diff
@@ -691,6 +694,30 @@ PYBIND11_MODULE(_bindings, m) {
...
+            [](nixlAgent &agent,
+               uintptr_t reqh) -> std::pair<nixl_status_t, std::vector<nixl_status_t>> {
+                std::vector<nixl_status_t> entry_status;
```
yanok: That's going to alloc the vector of statuses on every call to `getXferStatusList`, so if there is a loop polling for status in Python code:
```
while ...:
   status, statuses = agent.getXferStatusList(handle)
   ...
```
It's going to allocate on _every_ iteration.

I'm going to work this around by polling with the normal `getXferStatus` and only asking for the fine-grained statuses if `getXferStatus` returns an error, but should we think about getting Python bindings for it?

src/bindings/python/nixl_bindings.cpp
roiedanino: Maybe `NIXL_ERR_WHILE_IN_PROG` ?

src/bindings/rust/src/agent.rs
```diff
@@ -981,6 +986,50 @@ impl Agent {
...
+    /// * `InvalidParam` - If the request handle is invalid
+    /// * `BackendError` - If there was a backend error
+    pub fn get_xfer_status_list(&self, req: &XferRequest) -> Result<(XferStatus, Vec<i32>), NixlError> {
+        // Pre-allocate a reasonably sized buffer
+        let mut entry_status: Vec<i32> = vec![0; 1024];
+        let mut entry_count: usize = entry_status.len();
+
+        let status = unsafe {
+            bindings::nixl_capi_get_xfer_status_list(
+                self.inner.write().unwrap().handle.as_ptr(),
```
roiedanino, line 999: should access the inner write lock outside the unsafe block

src/api/cpp/nixl.h
brminich: regarding N3 what would be your suggestion? to return a vector of failed status + indices?

src/bindings/python/nixl_bindings.cpp
yanok: Hm... the workaround won't 100% work though: `getXferStatus` can return the first error, so we still need to poll with `getXferStatusList` (doing allocation each time) until all the other transfers are done.

src/api/cpp/nixl.h
rakhmets: It depends on how we plan to use this information. If the status of the failed transfer affects the decision, then yes: index and status. If we just need a list of failed transfers, then an index is enough. In my opinion, the combination of index and status looks like a more general solution.

brminich: ok, it sounds reasonable to me
@vvenkates27 wdyt?

mkhazraee: I see pros and cons. If more than half of the entires need to show the status then this would be worse. Let's say one of the entries got to an error early on, while most of the entries are in progress, instead of NIXL_IN_PROG we get tot NIXL_ERR_IN_PROG.

Looking at the two usage scenarios, for networking backends, the odds of error is low, and if it happens, it's failure recovery, so maybe not that sensitive to how efficient is going over the list. For case of storage, I think usually the lists are not that long, Vish would know more.

The current per entry mode might be slightly more easy to follow, versus you get a list of only errors. If the results is success, don't check, if the result is nixl_IN_PROG, don't check, if results is NIXL_ERR_IN_PROG then check, but for the ones which haven't failed but are in progress they won't show up. But from API point of view, you gave me a descriptor list, you get a same size status list, one sentence.

All these said, I'm more inclined towards the current mode instead of index, but maybe I'm missing something.

meson.build
mkhazraee: I suggest not to have this at all, and deprecate V1 right away.

src/api/cpp/nixl_types.h
mkhazraee: This is useful for the single aggregate status. If one entry failed, the rest are in progress or done. Especially with storage, some might take much more time to arrive, so better to get current status instead of waiting until all are known.

src/bindings/python/nixl_bindings.cpp
mkhazraee: Yeah, we were discussing the two options, giving it a negative value with NIXL_ERR_ prefix, or a positive value like now, and said let's discuss in the PR with others. It's error, but doesn't mean fully stop, the rest is going on. So somewhere in between. WDYT are the pros and cons for each case?

src/api/cpp/nixl.h
vvenkates27: Thanks for the comments after reviewing and considering them, let me try to address all the comments here.

Simplicity: "You gave me N descriptors, you get N statuses" - this is easy to explain and use. The alternative requires tracking which indices you haven't heard about.

Completeness: With the current design, you always know the state of ALL entries. With "only failures", if an entry hasn't failed yet but is still in progress, it won't appear in the output - you'd need to track that separately.

Practical batch sizes: For storage backends (OBJ, POSIX, GDS), batch sizes are typically in the tens to hundreds or at max 1-2K, not 64k. The 64k case is more relevant to networking backends where per-entry failures are rare.

No re-allocation overhead: Example in OBJ backend implementation, the status vector is cached in cachedResults_ and reused across checkXfer calls - we're not allocating a new vector on every poll. The overhead is copying the cached results to the output parameter which is not much.

Consistency: Users can swap getXferStatus() with getXferStatus(entry_status) and get the same overall return value behavior.

Given these points, I believe the current approach is more suited than an index-based solution.

meson.build
vvenkates27: Sure I will remove this from this PR

src/api/cpp/nixl_types.h
vvenkates27: +1 to @mkhazraee

src/bindings/python/nixl_bindings.cpp
vvenkates27: Yes this is a case we have observed an error in one of the xfers. We can easily update to a negative value, if it characterizes it better.

src/core/nixl_agent.cpp
vvenkates27: I will add this to the update.

src/bindings/python/nixl_bindings.cpp
vvenkates27: I agree this needs a minor change in the binding. I will udpate the binding implementation.

src/bindings/python/nixl_bindings.cpp
vvenkates27: I have updated the binding, with this binding there is no need for allocating multiple vectors at the python level.

src/bindings/rust/src/agent.rs
vvenkates27: Done.

src/api/cpp/nixl.h
rakhmets: > Simplicity: "You gave me N descriptors, you get N statuses" - this is easy to explain and use. The alternative requires tracking which indices you haven't heard about.

This does not accurately describe the current API. The current API: "You gave me N descriptors, you get N+1 statuses, where (N+1)-th depends on the first N".

> Completeness: With the current design, you always know the state of ALL entries. With "only failures", if an entry hasn't failed yet but is still in progress, it won't appear in the output - you'd need to track that separately.

I'm not sure I understand this item. Why do you need track them separately? If it's in progress - you will wait until they finished using the return status of `getXferStatus`. IIUC the main point of that is to track individual failures.
Or is there a real practical request for tracking progress? Anyway, for the tracking progress you don't need to fill the vector with `NIXL_IN_PROG` values, only the total number of transfers in progress needed.

> Practical batch sizes: For storage backends (OBJ, POSIX, GDS), batch sizes are typically in the tens to hundreds or at max 1-2K, not 64k. The 64k case is more relevant to networking backends where per-entry failures are rare.

I don't think that there is a practical reason to return `(N-M-L)` `NIXL_ SUCCESS`es, and `L` `NIXL_IN_PROG`s (where `N` - batch size, `M` - number of failed transfers, and `L` - number of requests in progress).

> Consistency: Users can swap getXferStatus() with getXferStatus(entry_status) and get the same overall return value behavior.

Yes, I agree with that: https://github.com/ai-dynamo/nixl/pull/1274#discussion_r2769019912

src/api/cpp/nixl_types.h
brminich: is there a value in checking vector of statuses while some are still in progress? I'd expect that we want to wait for the completion of all of them and then check each one separately. Otherwise you would need to traverse the whole vector for each newly completed chunk, which seems to be intrusive.

src/bindings/python/nixl_bindings.cpp
brminich: imo, if we need it - it is an error still, so should be negative

src/bindings/python/nixl_bindings.cpp
```diff
@@ -696,26 +696,38 @@ PYBIND11_MODULE(_bindings, m) {
...
             "getXferStatusList",
-            [](nixlAgent &agent,
-               uintptr_t reqh) -> std::pair<nixl_status_t, std::vector<nixl_status_t>> {
-                std::vector<nixl_status_t> entry_status;
-                nixl_status_t ret = agent.getXferStatus((nixlXferReqH *)reqh, entry_status);
-                // Don't throw for in-progress statuses
-                if (ret < 0 && ret != NIXL_ERR_NOT_SUPPORTED) {
-                    throw_nixl_exception(ret);
+            [](nixlAgent &agent, uintptr_t reqh, py::list entry_status) -> nixl_status_t {
+                std::vector<nixl_status_t> cpp_status;
```
yanok: This still allocates a new vector at every call. It allocates on stack initially, I believe, so its rather cheap. But once backend's implementation of `getXferStatus` calls `resize`, it will allocate on heap.

You either want to allocate it once on the Python side and pass it to C++ as C array pointer or something. Or you can have it completely on the C++ side, in the agent, and give Python only a read-only view of it, but then the question is how long should it live and then it's safe to collect it.

yanok: or, we can have a separate `StatusVector` object in the python bindings. So, the user would have to first create it outside of the loop, and then inside the loop you just pass the vector you already own.

yanok:
```
sv = self.agent.new_status_vector()
while ...:
   status = self.agent.getXferStatusList(handle, sv)
```
something like this

vvenkates27: Like we discussed I don't think the use case really dictates a large number in the batches (not > 1K). So the impact is not going to be big. However its a really small change, so I have incorporated in the binding.

src/api/cpp/nixl_types.h
mkhazraee: I see two points. One challenge is we cannot guarantee all of them will finish, in case of network failure and storage, and further enforcing backends have to have a time out is hard to verify. A normal user would call the basic one without the list output, and in case of error go check the list, or in case of time out. 

The other point is stragglers, if user wants to do something early on. The point is we have that data in the backend anyways, current status of descriptors in the transfer. It's up to the user to ask for it or not, we have the single status value as before.

Making this method return invalid list unless we're in done or error can get confusing, and when in progress and some got errors, the user might want to do something about those elements. Since that info is available anyways, it's just exposing it to a savvy user and not gate-keeping it. Note that a backend might not support this mode of operation anyways.

mkhazraee:
@vvenkates27 we had a discussion with Mikhail and the team, couple points:
- For backward compatibility, some users checked if the status is less than zero go the slow path and check what's the error. NIXL_ERR_IN_PROG is practically an early error, where they should go to slow path and see what happened. So good to keep it as a negative value. If a backend doesn't support detailed status, it would be same as before, just returning success or error at the end, and IN_PROG in the middle. This might encourage backends to return early error with this new error code.

- Because the main point of this new API is to know the current state, and check for the updates, not wait until the end, then user probably calls it multiple times per xfer. If we return the full desc list, the user has to check each time what has changed compared to the last time. This can be inefficient. If we go with an output vector of (index, err) elements, then the user can check what new elements were added, as in last time there were 2 elements, if still 2 elements nothing new, if length of the vector has changed, now there is something new to consume. Making this part more efficient justifies slightly more complex output. Kind of events list, based on time when we got error for each descriptor.

- There might be a case that user is also interested in success values. So we can have an input flag, to include successes in the output event vector too or not. Default being not to use. That way if some elements are done and we can trigger another process, we can use this optimization. And this way we're covering error, success and in-prog as the 3 possible scenarios, so we're not missing anything from having per element status.

mkhazraee:
One more thing I thought of, if during checkXfer we have the flag to output just the errors or include successes too, internally we should potentially keep two vectors, can be optimized some but still. Another idea would be to have a flag during xfer creation, valid for createXferReq and makeXferReq, that says how we want to track this xfer. Maybe even we can make that parameter an enum of 3, no_per_entry, error_only, error_and_success. That way we can be fully backward compatible, error bad_param if they ask for per element but the xfer was created with no_per_entry flag. Also it doesn't make sense if a user sometimes wants errors sometimes doesn't for the same transfer, so this makes sense I think. Also it might enable on some optimization on how we track based on that flag.

vvenkates27:
@mkhazraee @brminich I have updated this PR based on all the conversations above. Please have a look. 
cc: @aranadive

src/bindings/python/nixl_bindings.cpp
```diff
@@ -30,6 +30,37 @@ namespace py = pybind11;
 
 typedef std::map<std::string, std::vector<py::bytes>> nixl_py_notifs_t;
 
+/**
+ * Wrapper to hold nixl_xfer_entry_events_t for reuse across getXferStatus polling loops.
+ * Create once, pass to getXferStatus each poll to avoid per-call allocation.
+ */
+class nixlXferEntryEvents {
```
ColinNV, line 37: This class wraps a `std::deque` without adding anything, it should be eliminated (just use `std::deque::at()` instead of `get()`).

src/bindings/python/nixl_bindings.cpp
ColinNV: Perhaps, instead of checking for positive or negative, there should be `constexpr` functions with meaningful names to classify/group error codes?

src/core/nixl_agent.cpp
```diff
@@ -1166,6 +1173,41 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    if (data->remoteSections.count(req_hndl->remoteAgent) == 0) {
+        NIXL_ERROR_FUNC << "remote agent '" << req_hndl->remoteAgent
+                        << "' was invalidated during transfer";
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    nixl_status_t status = req_hndl->engine->checkXferEvents(req_hndl->backendHandle, events_out);
```
ColinNV: `const`

src/core/nixl_agent.cpp
```diff
@@ -1166,6 +1173,41 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    if (status == NIXL_ERR_NOT_SUPPORTED) return status;
```
ColinNV: Better `switch( status )`?

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -53,27 +54,22 @@ class nixlObjBackendReqH : public nixlBackendReqH {
...
-            if (it->wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
-                auto current_status = it->get();
-                if (current_status != NIXL_SUCCESS) {
-                    statusFutures_.clear();
-                    return current_status;
-                }
-                it = statusFutures_.erase(it);
-            } else {
+        nixl_status_t first_error = NIXL_SUCCESS;
+        for (size_t i = 0; i < statusFutures_.size(); ++i) {
```
ColinNV: `for( const auto& future : statusFutures_ ) {`

src/api/cpp/nixl_types.h
```diff
@@ -64,9 +66,36 @@ enum nixl_status_t {
...
+ * @brief One (index, status) event from checkXferEvents. Events are appended over time;
+ *        user compares event list length to previous call to see new completions.
+ */
+struct nixl_xfer_entry_event_t {
+    size_t index; /**< Descriptor index in the transfer. */
+    nixl_status_t status; /**< NIXL_SUCCESS or an error code; never NIXL_IN_PROG (entries are only
+                             appended once complete). */
+};
+
+using nixl_xfer_entry_events_t = std::deque<nixl_xfer_entry_event_t>;
```
ColinNV, line 99: Why deque and not vector?

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -252,6 +251,31 @@ DefaultObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
     return req_h->getOverallStatus();
 }
 
+nixl_status_t
+DefaultObjEngineImpl::checkXferEvents(nixlBackendReqH *handle,
+                                      nixl_xfer_entry_events_t &events) const {
+    nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *>(handle);
+    nixl_xfer_track_flags_t flags = req_h->trackFlags;
+    if (flags == 0) return NIXL_ERR_NOT_SUPPORTED;
+
```
ColinNV, line 368: `events.reserve(req_h->statusFutures_.size()`

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -252,6 +251,31 @@ DefaultObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
     return req_h->getOverallStatus();
 }
 
+nixl_status_t
+DefaultObjEngineImpl::checkXferEvents(nixlBackendReqH *handle,
+                                      nixl_xfer_entry_events_t &events) const {
+    nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *>(handle);
+    nixl_xfer_track_flags_t flags = req_h->trackFlags;
```
ColinNV: `const`

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -252,6 +251,31 @@ DefaultObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
...
+    for (size_t i = 0; i < req_h->statusFutures_.size(); ++i) {
+        if (req_h->appended_[i]) continue;
+        if (req_h->statusFutures_[i].wait_for(std::chrono::seconds(0)) !=
+            std::future_status::ready) {
+            any_pending = true;
+            continue;
+        }
+        nixl_status_t s = req_h->statusFutures_[i].get();
+        if (s != NIXL_SUCCESS && req_h->firstError_ == NIXL_SUCCESS) req_h->firstError_ = s;
+        bool include = (s != NIXL_SUCCESS && (flags & NIXL_XFER_TRACK_ERRORS)) ||
```
ColinNV: `const` (and perhaps move the Boolean expression to a `constexpr` function).

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -205,6 +201,9 @@ DefaultObjEngineImpl::postXfer(const nixl_xfer_op_t &operation,
                                const nixl_opt_b_args_t *opt_args) const {
     nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *>(handle);
 
+    if (opt_args) req_h->trackFlags = opt_args->trackFlags;
+    req_h->appended_.resize(local.descCount(), false);
```
ColinNV: Should this be done conditionally?

src/api/cpp/backend/backend_engine.h
```diff
@@ -153,6 +153,17 @@ class nixlBackendEngine {
...
         virtual nixl_status_t checkXfer(nixlBackendReqH* handle) const = 0;
 
+        /**
+         * Optional: append new (index, status) events and return overall status.
+         * Default returns NIXL_ERR_NOT_SUPPORTED. Track mode is stored in handle from prep/post.
+         */
+        virtual nixl_status_t
+        checkXferEvents(nixlBackendReqH *handle, nixl_xfer_entry_events_t &events) const {
+            (void)handle;
+            (void)events;
```
rakhmets:
```suggestion
        checkXferEvents(nixlBackendReqH *, nixl_xfer_entry_events_t &) const {
```

src/api/cpp/nixl.h
```diff
@@ -287,6 +287,20 @@ class nixlAgent {
...
+        /**
+         * @brief  Get transfer status with per-entry (index, status) events.
+         *         Request must have been created with trackFlags != 0
+         *         (via extra_params in createXferReq/makeXferReq). Events are appended over time;
+         *         compare events_out.size() to the previous call to see new completions.
+         *
+         * @param  req_hndl      Transfer request handle after postXferReq
+         * @param  events_out [out] Append-only event list (index, status).
+         * @return nixl_status_t NIXL_ERR_INVALID_PARAM if xfer was created with trackFlags==0,
+         *         NIXL_IN_PROG_WITH_ERR (negative) when still in progress, NIXL_SUCCESS or error
```
rakhmets: As far as I understand, we do not want to impose restrictions on the use of this method. If this method can be used independently, without `getXferStatus nixlXferReqH*)`, then it should exactly repeat the logic of the returned values.

src/core/nixl_agent.cpp
```diff
@@ -1166,6 +1173,41 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    if (status == NIXL_IN_PROG) {
+        req_hndl->status = NIXL_IN_PROG;
+        return NIXL_IN_PROG_WITH_ERR; /* negative so (status < 0) → slow path */
```
rakhmets: Returning NIXL_IN_PROG_WITH_ERR in the absence of errors looks like an incorrect solution.

rakhmets (commented):
I would recommend separating API changes and implementation into separate PRs.

src/api/cpp/nixl_types.h
```diff
@@ -64,9 +66,36 @@ enum nixl_status_t {
...
+/**
+ * @brief Per-entry tracking flags (set of properties). Empty (0) = track only summarized status.
+ *        Chosen at createXferReq/makeXferReq. Extensible for future modes.
+ */
+enum nixl_xfer_track_flag_t : uint32_t {
+    NIXL_XFER_TRACK_ERRORS = 1u << 0,
+    NIXL_XFER_TRACK_SUCCESSES = 1u << 1,
+};
+
+using nixl_xfer_track_flags_t = uint32_t;
```
rakhmets, line 86: The combination of these types doesn't provide type safety.
It should be `enum class`, `struct`, and `operator |` to produce the struct from multiple flags.
Otherwise, additional checks will be required to verify the uint32_t function argument.

src/api/cpp/nixl_types.h
```diff
@@ -64,9 +66,36 @@ enum nixl_status_t {
...
+    NIXL_IN_PROG_WITH_ERR = -13
 };
 
+/**
+ * @brief Per-entry tracking flags (set of properties). Empty (0) = track only summarized status.
+ *        Chosen at createXferReq/makeXferReq. Extensible for future modes.
+ */
+enum nixl_xfer_track_flag_t : uint32_t {
+    NIXL_XFER_TRACK_ERRORS = 1u << 0,
+    NIXL_XFER_TRACK_SUCCESSES = 1u << 1,
```
roiedanino: Maybe we can define:

```cpp
#define NIXL_BIT(n) (1u << n)

```

And then:

```suggestion
    NIXL_XFER_TRACK_ERRORS    = NIXL_BIT(0),
    NIXL_XFER_TRACK_SUCCESSES = NIXL_BIT(1),
```

src/api/cpp/backend/backend_engine.h
vvenkates27: Fixed — removed the named parameters and the (void) casts from the default stub.

src/core/nixl_agent.cpp
vvenkates27: Fixed — NIXL_IN_PROG_WITH_ERR is now only returned when at least one entry in events_out actually carries an error status. Returns plain NIXL_IN_PROG otherwise.

src/api/cpp/nixl_types.h
vvenkates27: Done — added NIXL_BIT(n) macro and updated both flags to use it.

src/api/cpp/nixl_types.h
vvenkates27: Agreed it's more type-safe, but that's a broader API-breaking change that affects all callers. I'd prefer to keep uint32_t flags for now and track that refactor separately. The NIXL_BIT macro (done above) improves readability without breaking anything.

src/api/cpp/nixl.h
vvenkates27: The trackFlags != 0 guard is intentional: if the transfer was created without tracking, the backend never populated events, so returning an empty list would silently mislead the caller. Keeping it as NIXL_ERR_INVALID_PARAM makes the misconfiguration explicit.

src/api/cpp/backend/backend_aux.h
```diff
@@ -33,6 +33,11 @@ struct nixlBackendOptionalArgs {
     nixl_blob_t notifMsg;
     bool        hasNotif = false;
     nixl_blob_t customParam;
+    /** Per-entry tracking mode for this transfer. Default 0 = no per-entry events.
+     *  Set NIXL_XFER_TRACK_ERRORS to append error events, NIXL_XFER_TRACK_SUCCESSES
+     *  to also append success events. Propagated from nixlAgentOptionalArgs::trackFlags
+     *  by the agent during prepXfer/postXfer. */
+    nixl_xfer_track_flags_t trackFlags = 0;
```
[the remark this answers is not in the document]
vvenkates27: Already documented — the field has a block comment describing valid flag values and default 0 behavior.

src/bindings/python/nixl_bindings.cpp
vvenkates27: The wrapper exists to give Python users a named, allocatable type they can create once and reuse across polls — the intent is that nixlXferEntryEvents is the Python-facing name for this concept. That said, the wrapper itself is thin and we could instead bind  nixl_xfer_entry_events_t directly as a Python class with the same methods, which would eliminate the indirectio

src/bindings/python/nixl_bindings.cpp
vvenkates27: Good point. We already have isInProgress() used in the tests — we could promote that pattern to a set of constexpr helpers in nixl_types.h (e.g. nixl_status_is_in_prog, nixl_status_is_err) and use them consistently across the switch statements and callers. I'll keep that as a follow-up to avoid scope creep here

src/bindings/python/nixl_bindings.cpp
```diff
@@ -202,6 +242,34 @@ PYBIND11_MODULE(_bindings, m) {
...
+                 if (i >= v.size()) throw py::index_error("index out of range");
+                 const auto &e = v.get(i);
+                 return py::make_tuple(e.index, static_cast<int>(e.status));
+             })
+        .def(
+            "__iter__",
+            [](const nixlXferEntryEvents &v) {
+                return py::make_iterator(v.events().begin(), v.events().end());
+            },
+            py::keep_alive<0, 1>())
```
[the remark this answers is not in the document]
vvenkates27: Good catch — fixed. py::make_iterator is lazy and tries to py::cast each nixl_xfer_entry_event_t element, but that struct has no py::class_<> binding, so it would raise TypeError the moment any element is accessed. Changed __iter__ to materialise a list of (index, status) tuples, consistent with __getitem__ and to_list(). Also removed the now-unnecessary py::keep_alive<0, 1>() since the returned iterator keeps the list alive independently of the C++ object.

src/bindings/rust/src/agent.rs
```diff
@@ -979,6 +979,35 @@ impl Agent {
...
+                req.handle(),
+                events.as_ptr(),
+            )
+        };
+
+        match status {
+            NIXL_CAPI_SUCCESS => Ok(XferStatus::Success),
+            NIXL_CAPI_IN_PROG => Ok(XferStatus::InProgress),
+            NIXL_CAPI_ERROR_INVALID_PARAM => Err(NixlError::InvalidParam),
+            _ => Err(NixlError::BackendError),
```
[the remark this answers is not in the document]
vvenkates27: Fixed — NIXL_IN_PROG_WITH_ERR and NIXL_ERR_NOT_SUPPORTED are now passed through distinctly in wrapper.cpp/wrapper.h, and the Rust layer adds XferStatus::InProgressWithErr and NixlError::NotSupported to match.

src/bindings/rust/src/lib.rs
```diff
@@ -379,6 +387,30 @@ impl OptArgs {
         }
     }
 
+    /// Set per-entry tracking flags for create_xfer_req/make_xfer_req.
+    /// Use NIXL_CAPI_XFER_TRACK_ERRORS and/or NIXL_CAPI_XFER_TRACK_SUCCESSES.
+    pub fn set_track_flags(&mut self, track_flags: u32) -> Result<(), NixlError> {
```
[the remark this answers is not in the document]
vvenkates27: Fixed — doc comment now references [XFER_TRACK_ERRORS] and [XFER_TRACK_SUCCESSES] with rustdoc links.

src/bindings/rust/tests/tests.rs
```diff
@@ -2235,3 +2235,21 @@ fn test_desc_list_serialize_with_real_storage() {
...
+    assert!(events.iter().next().is_none());
+
+    let mut opt_args = OptArgs::new().expect("OptArgs::new");
+    opt_args
+        .set_track_flags(XFER_TRACK_ERRORS | XFER_TRACK_SUCCESSES)
+        .expect("set_track_flags");
+    assert_eq!(
+        opt_args.track_flags().unwrap(),
+        XFER_TRACK_ERRORS | XFER_TRACK_SUCCESSES
+    );
```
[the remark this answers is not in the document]
vvenkates27: Fixed — expanded the test to assert the default is 0, then round-trip XFER_TRACK_ERRORS alone, XFER_TRACK_SUCCESSES alone, and finally the combined mask.

src/core/nixl_agent.cpp
```diff
@@ -1157,6 +1187,41 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+        NIXL_ERROR_FUNC << "getXferStatus(req, events) requires xfer created with trackFlags != 0";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    NIXL_SHARED_LOCK_GUARD(data->lock);
+
+    if (data->remoteSections.count(req_hndl->remoteAgent) == 0) {
+        NIXL_ERROR_FUNC << "remote agent '" << req_hndl->remoteAgent
+                        << "' was invalidated during transfer";
+        return NIXL_ERR_NOT_FOUND;
```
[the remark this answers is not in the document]
vvenkates27: data->remoteSections was corrected to data->remoteSections_.

src/core/nixl_agent.cpp
vvenkates27: Fixed — status is now const.

src/core/nixl_agent.cpp
vvenkates27: Good call — refactored to switch (status).

src/core/nixl_agent.cpp
```diff
@@ -1157,6 +1187,41 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+
+    if (status == NIXL_IN_PROG) {
+        req_hndl->status = NIXL_IN_PROG;
+        return NIXL_IN_PROG_WITH_ERR; /* negative so (status < 0) → slow path */
+    }
+
+    req_hndl->status = status;
+    if (data->telemetryEnabled) {
+        if (status == NIXL_SUCCESS)
+            req_hndl->updateRequestStats(data->telemetry_, NIXL_TELEMETRY_FINISH);
```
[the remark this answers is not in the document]
vvenkates27: changed to data->telemetry_.get().

src/core/nixl_agent.cpp
```diff
@@ -1166,6 +1175,44 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
...
+        else if (status < 0)
+            data->addErrorTelemetry(status);
+    }
+    return status;
```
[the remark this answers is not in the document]
vvenkates27: remote-validity check is now guarded by req_hndl->status == NIXL_IN_PROG (so terminal requests aren't re-checked), and NIXL_ERR_REMOTE_DISCONNECT from checkXferEvents now calls invalidateRemoteData to match the legacy path.

src/plugins/obj/s3/engine_impl.cpp
vvenkates27: Done — switched to for (const auto &future : statusFutures_).

src/plugins/obj/s3/engine_impl.cpp
vvenkates27: Good catch — appended_ is only used in checkXferEvents when trackFlags != 0, so the resize is now conditional on req_h->trackFlags.

src/plugins/obj/s3/engine_impl.cpp
vvenkates27: Fixed — flags is now const.

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -336,6 +335,31 @@ DefaultObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
     return req_h->getOverallStatus();
 }
 
+nixl_status_t
+DefaultObjEngineImpl::checkXferEvents(nixlBackendReqH *handle,
+                                      nixl_xfer_entry_events_t &events) const {
+    nixlObjBackendReqH *req_h = static_cast<nixlObjBackendReqH *>(handle);
+    nixl_xfer_track_flags_t flags = req_h->trackFlags;
+    if (flags == 0) return NIXL_ERR_NOT_SUPPORTED;
```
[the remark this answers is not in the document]
vvenkates27: Fixed — added if (!handle) return NIXL_ERR_INVALID_PARAM; at the top of checkXferEvents, matching checkXfer.

src/plugins/obj/s3/engine_impl.cpp
vvenkates27: nixl_xfer_entry_events_t is a std::deque, which has no reserve() method — that's a std::vector-only API. If reducing allocations is the goal here, switching the underlying type to std::vector would be needed first, but that's a broader change. Skipping for now.

src/plugins/obj/s3/engine_impl.cpp
vvenkates27: Fixed const on include. On extracting to a constexpr function — the expression is two lines and the variable name already documents intent, so I'd leave it inline for now.

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -336,6 +335,31 @@ DefaultObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
...
+        }
+        nixl_status_t s = req_h->statusFutures_[i].get();
+        if (s != NIXL_SUCCESS && req_h->firstError_ == NIXL_SUCCESS) req_h->firstError_ = s;
+        bool include = (s != NIXL_SUCCESS && (flags & NIXL_XFER_TRACK_ERRORS)) ||
+            (s == NIXL_SUCCESS && (flags & NIXL_XFER_TRACK_SUCCESSES));
+        if (include) events.push_back({i, s});
+        req_h->appended_[i] = true;
+    }
+    return any_pending ? NIXL_IN_PROG : req_h->firstError_;
+}
```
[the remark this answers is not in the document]
vvenkates27: Fixed — added mutable std::mutex eventsMutex_ to nixlObjBackendReqH and a std::lock_guard at the start of checkXferEvents to serialize concurrent callers, preventing races on the packed vector<bool> appended_ bits and the non-atomic firstError_

src/plugins/meson.build
```diff
@@ -13,6 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Base deps for plugins that use += (e.g. OBJ when POSIX is disabled)
+plugin_deps = [nixl_infra, nixl_common_dep]
+compile_defs = []
+compile_flags = []
```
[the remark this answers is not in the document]
vvenkates27: Fixed. obj/meson.build now declares obj_plugin_deps as a local variable starting from the base deps ([nixl_infra, nixl_common_dep, ...]) rather than mutating the shared plugin_deps with +=. This prevents POSIX's deps (file_utils_interface, linux_aio_dep, etc.) from leaking into OBJ when both plugins are enabled. Also removed the now-unnecessary plugin_deps initialization from the parent meson.build (kept compile_defs/compile_flags since azure_blob and gusli still read those).

test/gtest/plugins/test_obj_checkXferList.cpp
```diff
@@ -0,0 +1,538 @@
...
+    } while (isInProgress(overall_status) && poll_count < max_polls);
+
+    ASSERT_EQ(overall_status, NIXL_SUCCESS);
+
+    // Call checkXferEvents again - should work with shared_future (no new events)
+    nixl_xfer_entry_events_t events2;
+    auto overall_status2 = localBackendEngine_->checkXferEvents(handle, events2);
+    ASSERT_EQ(overall_status2, NIXL_SUCCESS);
+    ASSERT_EQ(events1.size(), events2.size());
+
```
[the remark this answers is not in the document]
vvenkates27: Fixed in ResultCaching — reusing the same events container across subsequent calls and asserting events.size() == prev_size (no new events appended). SharedFutureMultipleCalls already had this pattern applied.

test/gtest/plugins/test_obj_checkXferList.cpp
```diff
@@ -0,0 +1,528 @@
...
+    nixl_status_t overall_status_events, overall_status_single;
+    int max_polls = 100;
+    int poll_count = 0;
+
+    do {
+        overall_status_events = localBackendEngine_->checkXferEvents(handle, events);
+        overall_status_single = localBackendEngine_->checkXfer(handle);
+
+        EXPECT_EQ(overall_status_events, overall_status_single);
+
```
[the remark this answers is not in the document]
vvenkates27: Already addressed — the per-iteration EXPECT_EQ(overall_status_events, overall_status_single) was removed from inside the polling loop. The two statuses are now compared once after the loop exits, with a comment explaining why in-loop comparison is race-prone.

PR #1275, opened 2026-02-04
nixl_ep: Disable fast fault detection for release builds

rtourgeman:
When running nixl_ep with vLLM, ranks were timing out during connection because ENABLE_FAST_DEBUG was always enabled, which sets a 10-second timeout. This timeout is too short for initializing large model like DeepSeek V3.

- Remove hardcoded -DENABLE_FAST_DEBUG=1 from cuda_args
- Enable ENABLE_FAST_DEBUG only when buildtype != release

### DIFF
--- a/examples/device/ep/meson.build
+++ b/examples/device/ep/meson.build
@@ -100,7 +100,6 @@ nixl_ep_cuda_args = [
     '-Xcompiler', '-Wno-sign-compare',
     '-Xcompiler', '-Wno-reorder',
     '-Xcompiler', '-Wno-attributes',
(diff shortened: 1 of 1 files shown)


examples/device/ep/meson.build
```diff
@@ -127,6 +126,12 @@ if get_option('buildtype') == 'debug'
     nixl_ep_cpp_args += ['-g']
 endif
 
+enable_fast_debug = get_option('buildtype') == 'debug'
```
itayalroy: Can we enable it as long as `buildtype != release`?
I would like `debugoptimized` for example to have fast debug enabled

PR #1277, opened 2026-02-04
UTILS/SERDES: Add more checks and small optimizations.

ColinNV:
## What?
Add some checks to the deserialization to detect and handle corrupt or unexpected data by printing an error messages instead of failing ... otherwise.

## Why?
The deserialization code does not properly validate the raw data it received and might behave strangely, also not all client code appears to call `getBufLen()` before `getBuf()`.

## How?
Explicitly validate the data in the deserialization calls (and some minor optimizations, partly to prevent redundancies with the new checking code).

### DIFF
--- a/.github/workflows/aws_efa_validation.yml
+++ b/.github/workflows/aws_efa_validation.yml
@@ -10,13 +10,14 @@ jobs:
   run_aws_tests:
     name: Run AWS Tests - ${{ matrix.test_name }}
     environment: SWX_AWS
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     env:
       AWS_DEFAULT_REGION: eu-central-1
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       # tmp workaround: pin UCX ver to 1.18 until EFA support is available in UCX 1.19.
       UCX_VERSION: v1.18.0
+      KUBECTL_VERSION: v1.35.0
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


src/utils/serdes/serdes.cpp
```diff
@@ -14,19 +14,18 @@
...
 
-nixlSerDes::nixlSerDes() {
-    workingStr = "nixlSerDes|";
-    des_offset = 11;
-
-    mode = SERIALIZE;
-}
+nixlSerDes::nixlSerDes()
+    : workingStr("nixlSerDes|"),
+      des_offset(workingStr.size()),
```
brminich, line 24: 👍

src/utils/serdes/serdes.cpp
```diff
@@ -48,28 +47,34 @@ nixl_status_t nixlSerDes::addStr(const std::string &tag, const std::string &str)
...
+        NIXL_ERROR << "Deserialization of tag " << tag << " failed for tag mismatch";
+        return "";
+    }
 
-    //get len
-    //_stringToBytes(&len, workingStr.data() + des_offset, sizeof(ssize_t));
-    _stringToBytes(&len, workingStr.substr(des_offset, sizeof(ssize_t)), sizeof(ssize_t));
-    des_offset += sizeof(ssize_t);
+    size_t len;
+    std::memcpy(&len, workingStr.data() + des_offset + tag.size(), sizeof(len));
```
brminich, line 62: maybe just `std::copy` to avoid scary `memcpy` name?

ColinNV: I can change it if you want to -- usually I prefer `memcpy` and `memcmp` for low-level stuff to make it more visible/grep-able.

brminich: ok, i do not have strong opinion

PR #1278, opened 2026-02-04
Dell ObjectScale S3 over RDMA engine support for OBJ plugin

jgoldsch12:
## What?
Introduces a OBJ plugin accelerated type to support Dell ObjectScale S3 over RDMA feature

## Why?
Introduce support for Dell ObjectScale S3 over RDMA feature for NIXL OBJ plugin

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -246,8 +246,8 @@ nixlUcxEp::sendAm(unsigned msg_id,
                   size_t len,
                   uint32_t flags,
                   nixlUcxReq *req,
-                  am_deleter_t deleter) {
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -100,8 +100,20 @@ has_posix_aio = cpp.has_function('aio_cancel', prefix: '#include <aio.h>', depen
 has_posix_plugin = has_linux_aio or has_io_uring or has_posix_aio
 
 # Check for CUObject Client library
-cuobj_dep = dependency('cuobjclient-13.1', required: false)
-if cuobj_dep.found()
+cc = meson.get_compiler('cpp')
```
jgoldsch12: I will revert this change, it can be ignored.  We are moving our environment to CTK 13.1.1

PR #1287, opened 2026-02-06
Warn when EFA hardware is present but LIBFABRIC backend is not used

guy-ealey-morag:
## What?
Warn when EFA hardware is present but LIBFABRIC backend is not used

## Why?
This is a similar warning to the already existing warnings about InfiniBand and CUDA preset but not used.
The only difference is that now instead of querying UCX, the check is done against the selected backend type.

## Summary by CodeRabbit

* **New Features**
  * Detects Amazon EFA hardware and warns when EFA devices are present but libfabric is not selected, recommending libfabric for best performance.
* **Tests**
  * Expanded hardware-warning tests to cover EFA scenarios, added version-based gating, and updated test fixture usage.
* **Refactor**
  * Hardware detection now uses a cached singleton for consistent device discovery; warning is triggered at relevant registration points.

### DIFF
--- a/test/gtest/hw_warning_test.cpp
+++ b/test/gtest/hw_warning_test.cpp
@@ -20,6 +20,7 @@
 #include <string>
 #include <vector>
 
+#include "absl/strings/str_join.h"
 #include "common.h"
 #include "nixl.h"
 #include "ucx_utils.h"
@@ -156,7 +157,7 @@ TEST_F(HardwareWarningTest, EfaHardwareMismatchWarning) {
 
 /**
  * Test that no warning is logged when EFA devices are present and the
- * LIBFABRIC backend is created (without UCX).
+ * LIBFABRIC backend is among the created backends.
  */
 TEST_F(HardwareWarningTest, EfaHardwareMismatchNoWarning) {
 #ifndef HAVE_LIBFABRIC
@@ -168,17 +169,28 @@ TEST_F(HardwareWarningTest, EfaHardwareMismatchNoWarning) {
         GTEST_SKIP() << "No EFA devices detected, skipping test";
     }
 
-    envHelper_.addVar("NIXL_PLUGIN_DIR", std::string(BUILD_DIR) + "/src/plugins/libfabric");
-    nixlAgent agent("EfaTestAgent", nixlAgentConfig(true));
+    const std::vector<std::vector<std::string>> test_cases = {
+        {"LIBFABRIC"},
+        {"UCX", "LIBFABRIC"},
+        {"LIBFABRIC", "UCX"},
+    };
 
-    nixlBackendH *backend;
-    EXPECT_EQ(agent.createBackend("LIBFABRIC", {}, backend), NIXL_SUCCESS);
+    for (size_t i = 0; i < test_cases.size(); ++i) {
+        const auto &backends = test_cases[i];
+        const auto backends_str = absl::StrJoin(backends, ", ");
 
-    const gtest::LogIgnoreGuard lig_reg_fail("registerMem: registration failed");
+        std::cout << "\n > Case " << i << ": backends=[" << backends_str << "]\n" << std::endl;
 
-    /* Call registerMem to trigger the warning check */
-    const nixl_reg_dlist_t descs(DRAM_SEG);
-    agent.registerMem(descs);
+        nixlAgent agent("EfaTestAgent", nixlAgentConfig(true));
 
-    envHelper_.popVar();
+        for (const auto &name : backends) {
+            nixlBackendH *backend;
+            EXPECT_EQ(agent.createBackend(name, {}, backend), NIXL_SUCCESS);
+        }
+
+        const gtest::LogIgnoreGuard lig_reg_fail("registerMem: registration failed");
+
+        const nixl_reg_dlist_t descs(DRAM_SEG);
+        agent.registerMem(descs);
+    }
 }


src/api/cpp/nixl.h
```diff
@@ -608,6 +608,13 @@ class nixlAgent {
         checkRemoteMD (const std::string remote_name,
                        const nixl_xfer_dlist_t &descs) const;
 
+        /**
+         * @brief Warn if hardware is detected that would benefit from a different backend.
+         *
+         * @param type The backend type being used
+         */
+        void
+        warnAboutHardwareSupportMismatch(const nixl_backend_t &type);
```
brminich: pls avoid any API changes

guy-ealey-morag: Actually this function can be private

test/gtest/hw_warning_test.cpp
```diff
@@ -113,3 +120,61 @@ TEST_F(UcxHardwareWarningTest, NoWarningWhenIbAndCudaSupported) {
...
+ * Test that a warning is logged when EFA devices are present but a
+ * non-LIBFABRIC backend is created.
+ */
+TEST_F(HardwareWarningTest, WarnWhenEfaPresentAndNonLibfabricBackend) {
+    const nixl::hwInfo hw_info;
+    if (hw_info.numEfaDevices == 0) {
+        GTEST_SKIP() << "No EFA devices detected, skipping test";
+    }
+
+    envHelper_.addVar("NIXL_PLUGIN_DIR", std::string(BUILD_DIR) + "/src/plugins/ucx");
```
tvegas1, line 134: maybe we can merge most of that test with next one, just passing different plugin name, and then using differents checks.

guy-ealey-morag: The expected result is different (one checks for the warning, the other checks that there was no warning),
so I think it's better to keep them separate.
Also for the no-warning test there is an additional skip check.

src/utils/common/hw_info.cpp
```diff
@@ -31,12 +32,23 @@ namespace nixl {
 namespace {
 
     constexpr const char *kPciDevicePath = "/sys/bus/pci/devices";
+
     constexpr unsigned long kPciVendorMellanox = 0x15b3;
     constexpr unsigned long kPciVendorNvidia = 0x10de;
+    constexpr unsigned long kPciVendorAmazon = 0x1d0f;
+
```
michal-shalev, line 39: Maybe remove the extra new lines?

test/gtest/meson.build
```diff
@@ -79,7 +83,7 @@ else
 endif
 
 if ucx_dep.found()
-    gtest_sources += 'ucx_hw_warning_test.cpp'
+    gtest_sources += 'hw_warning_test.cpp'
```
michal-shalev, line 87: Why did you change the file name?

test/gtest/hw_warning_test.cpp
```diff
@@ -113,3 +120,61 @@ TEST_F(UcxHardwareWarningTest, NoWarningWhenIbAndCudaSupported) {
...
+
+    gtest::scopedTestLogSink log_sink;
+
+    nixlBackendH *backend;
+    EXPECT_EQ(agent.createBackend("UCX", {}, backend), NIXL_SUCCESS);
+
+    unsigned expected_warnings = 1;
+
+    if (ucpVersion_ < UCP_VERSION(1, 19)) {
+        /* Ignore possible warning about UCX version */
```
michal-shalev: According to NIXL code style this comment should be `//`

test/gtest/meson.build
guy-ealey-morag: The new AWS hardware warning test is not related to UCX so I made the name more general

src/utils/common/hw_info.cpp
guy-ealey-morag: The extra lines group the constants (vendors, classes, etc)

test/gtest/hw_warning_test.cpp
tvegas1: not a big deal, seems you could bundle env, device count returned as true or false, backend creation in separate function.

src/core/nixl_agent.cpp
```diff
@@ -282,6 +283,23 @@ nixlAgent::getBackendParams (const nixlBackendH* backend,
...
+
+void
+warnAboutHardwareSupportMismatch(const nixl_backend_t &type) {
+    if (type != "LIBFABRIC") {
+        static const nixl::hwInfo hw_info;
+
+        if (hw_info.numEfaDevices > 0) {
+            NIXL_WARN << hw_info.numEfaDevices
+                      << " Amazon EFA(s) were detected, it is recommended to use "
+                         "the LIBFABRIC backend for best performance.";
```
brminich: let's also print the name of the instantiated backend.

src/core/nixl_agent.cpp
```diff
@@ -282,6 +283,23 @@ nixlAgent::getBackendParams (const nixlBackendH* backend,
     return NIXL_SUCCESS;
 }
 
+namespace {
+
+void
+warnAboutHardwareSupportMismatch(const nixl_backend_t &type) {
+    if (type != "LIBFABRIC") {
+        static const nixl::hwInfo hw_info;
```
brminich: maybe we can cache hwInfo data and take it from the system just once? Now we do it twice

tvegas1:
Some format related failures are still seen.

tvegas1 (dismissed):

brminich:
@guy-ealey-morag pls resolve test conflict

tvegas1 (approved):

src/core/nixl_agent.cpp
```diff
@@ -396,6 +414,8 @@ nixlAgent::createBackend(const nixl_backend_t &type,
 
         NIXL_DEBUG << "Created backend: " << type;
 
+        warnAboutHardwareSupportMismatch(type);
```
brminich: seems like in this case we will also warn if NIXL creates just some storage backend on AWS system.
Can we print a warning iff:
- we have EFA NICs
- UCX backend is created
- libfabric backend is not created

test/gtest/hw_warning_test.cpp
```diff
@@ -116,3 +119,63 @@ TEST_F(UcxHardwareWarningTest, NoWarningWhenIbAndCudaSupported) {
...
+    nixlAgent agent("EfaTestAgent", nixlAgentConfig(true));
+
+    nixlBackendH *backend;
+    EXPECT_EQ(agent.createBackend("LIBFABRIC", {}, backend), NIXL_SUCCESS);
+
+    /* Call registerMem to trigger the warning check */
+    const nixl_reg_dlist_t descs(DRAM_SEG);
+    agent.registerMem(descs);
+
+    envHelper_.popVar();
```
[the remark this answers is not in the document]
guy-ealey-morag: It will be caught by `LogProblemCounter`

src/core/nixl_agent.cpp
```diff
@@ -279,6 +280,24 @@ nixlAgent::getBackendParams (const nixlBackendH* backend,
     return NIXL_SUCCESS;
 }
 
+void
+nixlAgentData::warnAboutEfaHardwareMismatch() {
```
ColinNV, line 284: `warnOnEfaHardwareMismatch`?

`warnOnUcxForEfa`? (or `withEfa`?)

src/utils/common/hw_info.cpp
```diff
@@ -31,12 +32,23 @@ namespace nixl {
 namespace {
 
     constexpr const char *kPciDevicePath = "/sys/bus/pci/devices";
+
     constexpr unsigned long kPciVendorMellanox = 0x15b3;
     constexpr unsigned long kPciVendorNvidia = 0x10de;
+    constexpr unsigned long kPciVendorAmazon = 0x1d0f;
```
ColinNV, line 38: Aren't these `std::uint16_t`? It's ok, you're just using what you found, but we might want to fix this separately.

src/utils/common/hw_info.h
```diff
@@ -23,13 +23,19 @@ namespace nixl {
...
  */
-struct hwInfo {
+class hwInfo {
+public:
     unsigned numNvidiaGpus = 0;
     unsigned numIbDevices = 0;
+    unsigned numEfaDevices = 0;
 
+    /** Return a cached singleton instance of hwInfo */
+    static const hwInfo &
```
ColinNV: `[[nodiscard]]`

src/utils/common/hw_info.h
```diff
@@ -23,13 +23,19 @@ namespace nixl {
 /**
  * @brief Hardware information gathered by scanning PCI devices.
  *
- * Scans the sysfs PCI device directory to detect available
- * NVIDIA GPUs and InfiniBand devices on the system.
+ * Scans the sysfs PCI device directory to detect available hardware.
  */
-struct hwInfo {
+class hwInfo {
```
ColinNV, line 28: Did you want to make this non-copyable?

src/utils/common/hw_info.cpp
guy-ealey-morag: The values are read from the system as unsinged long so it's just simpler to use the same type.

src/core/nixl_agent.cpp
guy-ealey-morag: We already have `warnAboutHardwareSupportMismatch` so it uses the same pattern

src/utils/common/hw_info.cpp
ColinNV: That seems to be just an artifact of using `std::stoul()` for the string-to-integer conversion. But fixing all that is out-of-scope for this PR.

test/gtest/hw_warning_test.cpp
```diff
@@ -116,3 +119,66 @@ TEST_F(UcxHardwareWarningTest, NoWarningWhenIbAndCudaSupported) {
...
+    agent.registerMem(descs);
+
+    EXPECT_EQ(lig_efa_warn.getIgnoredCount(), 1);
+
+    envHelper_.popVar();
+}
+
+/**
+ * Test that no warning is logged when EFA devices are present and the
+ * LIBFABRIC backend is created (without UCX).
```
brminich: can we test that there is no warning when we have both ucx and efa, but efa plugin is instantiated first?

guy-ealey-morag: Good idea, I added more test cases for the no warning test.

src/core/agent_data.h
```diff
@@ -68,6 +68,7 @@ class nixlAgentData {
         nixlAgentConfig config;
         nixlLock        lock;
         bool telemetryEnabled = false;
+        bool efaWarningChecked = false;
```
iyastreb, line 71: Maybe we can avoid exposing this field in the class?
Maybe we can use local static variable, that is set to true only when warning is actually produced. Cause in principle we need a single warning per process, not per each agent, right?

guy-ealey-morag: Technically every agent can be initialized with different backends so I think checking this per agent is more robust.

iyastreb: But then you may produce multiple warnings within the same process. With local static we warn at most once. We can set isWarned flag after warning is fired

iyastreb: ok, it's hard to test with local static

PR #1288, opened 2026-02-07
TEST/GTEST: Fail on unexpected error or warning log.

ColinNV:
## What?
Fail the gtest when NIXL logs an unexpected error or warning message.

## Why?
To prevent testcases from passing when they should fail or from passing for the wrong reason.

## How?
Count the number of unexpected error or warning messages and return a non-zero value to the environment when the counter is greater than zero at the end of a test run. And provide a meanst for testcases to specify which such messages they do expect.

Requires #1334 to be merged first to correctly work in non-gpu ci tests.

### DIFF
--- a/test/gtest/metadata_exchange.cpp
+++ b/test/gtest/metadata_exchange.cpp
@@ -400,7 +400,8 @@ TEST_F(MetadataExchangeTestFixture, SocketSendLocalAndInvalidateLocal) {
 
         std::this_thread::sleep_for(std::chrono::seconds(3)); // Must exceed timeout to catch logs.
 
-        const size_t ignored = lig1.getIgnoredCount() + lig2.getIgnoredCount() + lig3.getIgnoredCount();
+        const size_t ignored =
+            lig1.getIgnoredCount() + lig2.getIgnoredCount() + lig3.getIgnoredCount();
         EXPECT_GE(ignored, 1);
     }
 


ColinNV:
This is a draft because the metadata exchange tests still need to be updated regarding which errors and warnings they expect.

test/gtest/common.cpp
```diff
@@ -134,38 +135,71 @@ PortAllocator::next_tcp_port() {
                              " - " + std::to_string(instance._max_port));
 }
 
-scopedTestLogSink::scopedTestLogSink() {
-    absl::AddLogSink(&sink_);
+namespace {
+    std::mutex log_problem_mutex;
+    size_t global_problem_count = 0;
+    std::list<log_ignore_entry_t> log_problem_ignore;
+    const std::regex non_gpu_regex("[0-9]+ NVIDIA GPU(s) were detected, but UCX CUDA support was "
```
guy-ealey-morag:
```suggestion
    const std::regex non_gpu_regex("[0-9]+ NVIDIA GPU\\(s\\) were detected, but UCX CUDA support was "
```

test/gtest/common.cpp
```diff
@@ -134,38 +135,71 @@ PortAllocator::next_tcp_port() {
...
-        warnings_.emplace_back(entry.text_message());
+size_t
+LogIgnoreGuard::getIgnoredCount() const noexcept {
+    const std::lock_guard lock(log_problem_mutex);
+    return iter_->second;
+}
+
+LogProblemCounter::LogProblemCounter() {
+    if (std::getenv("NIXL_CI_NON_GPU") != nullptr) {
+        log_problem_ignore.emplace_back(non_gpu_regex, 0);
```
guy-ealey-morag: Consider locking the mutex here for consistency and robustness

test/gtest/common.cpp
```diff
@@ -134,38 +135,71 @@ PortAllocator::next_tcp_port() {
                              " - " + std::to_string(instance._max_port));
 }
 
-scopedTestLogSink::scopedTestLogSink() {
-    absl::AddLogSink(&sink_);
+namespace {
```
guy-ealey-morag, line 138: There are a lot of "floating" global variables related to `LogProblemCounter`.
Is there a different pattern we can use to avoid that (such as singleton or other?)

test/gtest/common.cpp
```diff
@@ -134,38 +135,71 @@ PortAllocator::next_tcp_port() {
...
-    if (entry.log_severity() == absl::LogSeverity::kWarning) {
-        warnings_.emplace_back(entry.text_message());
+size_t
+LogIgnoreGuard::getIgnoredCount() const noexcept {
+    const std::lock_guard lock(log_problem_mutex);
+    return iter_->second;
+}
+
+LogProblemCounter::LogProblemCounter() {
+    if (std::getenv("NIXL_CI_NON_GPU") != nullptr) {
```
guy-ealey-morag: I think this logic can be implemented as a `LogIgnoreGuard` in `RunAllTests()`.
This way `LogProblemCounter` can stay clean from any test specific details.

test/gtest/common.cpp
```diff
@@ -134,38 +135,71 @@ PortAllocator::next_tcp_port() {
...
+        for (auto &[rx, count] : log_problem_ignore) {
+            if (std::regex_search(msg, rx)) {
+                ++count;
+                return;
+            }
         }
+        ++global_problem_count;
     }
-    return count;
+    std::cerr << "ATTENTION: Unexpected NIXL warning or error detected!" << std::endl;
```
guy-ealey-morag, line 197:
```suggestion

    std::cerr << "ATTENTION: Unexpected NIXL warning or error detected!" << std::endl;
```

test/gtest/main.cpp
```diff
@@ -65,11 +65,23 @@ void ParseArguments(int argc, char **argv) {
...
 int RunTests(int argc, char **argv) {
-  testing::InitGoogleTest(&argc, argv);
-  ParseArguments(argc, argv);
+    testing::InitGoogleTest(&argc, argv);
+    ParseArguments(argc, argv);
+    const int result = RunAllTests();
 
-  return RUN_ALL_TESTS();
+    if (const size_t problems = LogProblemCounter::getProblemCount(); problems > 0) {
+        std::cerr << "ATTENTION: Unexpected NIXL warnings and/or errors detected!" << std::endl;
```
guy-ealey-morag:
```suggestion
        std::cerr << "ATTENTION: " << problems << " unexpected NIXL warnings and/or errors detected!" << std::endl;
```

I think this can be one print

test/gtest/metadata_exchange.cpp
```diff
@@ -134,8 +131,11 @@ class MetadataExchangeTestFixture : public testing::Test {
 
     void TearDown() override
     {
-        for (auto &agent : agents_)
-            agent.agent->invalidateLocalMD(nullptr);
+        for (auto &agent : agents_) {
+            if (agent.agent) {
```
guy-ealey-morag, line 135: what issue does this solve?

test/gtest/metadata_exchange.cpp
```diff
@@ -348,9 +385,18 @@ TEST_F(MetadataExchangeTestFixture, SocketSendLocalAndInvalidateLocal) {
     // Send to invalid IP address, should not block the test
     send_args.ipAddr = "10.10.10.10";
     send_args.port = 1234;
-    ASSERT_EQ(src.agent->sendLocalMD(&send_args), NIXL_SUCCESS);
+    {
+        const LogIgnoreGuard lig1("poll timed out for ip_addr: 10.10.10.10 and port: 1234");
+        const LogIgnoreGuard lig2(
+            "Listener thread could not connect to IP 10.10.10.10 and port 1234");
```
guy-ealey-morag: I suggest using `send_args.ipAddr` and `send_args.port` to build the log strings (instead of repeating `10.10.10.10` and `1234`)

test/gtest/metadata_exchange.cpp
```diff
@@ -455,15 +501,33 @@ TEST_F(MetadataExchangeTestFixture, SocketSendLocalPartialWithErrors) {
...
+    {
+        const LogIgnoreGuard lig1("loadRemoteMD: no common backend found");
+        const LogIgnoreGuard lig2(std::regex("loadRemoteMD in listener thread failed for md from "
+                                             "peer 127.0.0.1:[0-9]+ with error NIXL_ERR_BACKEND"));
 
-    ASSERT_EQ(src.agent->sendLocalPartialMD({DRAM_SEG}, &send_args), NIXL_SUCCESS);
+        ASSERT_EQ(src.agent->sendLocalPartialMD({DRAM_SEG}, &send_args), NIXL_SUCCESS);
 
-    std::this_thread::sleep_for(sleep_time);
+        std::this_thread::sleep_for(sleep_time);
```
guy-ealey-morag:
```suggestion
        std::this_thread::sleep_for(std::chrono::seconds(1));
```

Like in the previous tests

test/gtest/metadata_exchange.cpp
```diff
@@ -475,6 +539,8 @@ TEST_F(MetadataExchangeTestFixture, LocalNonLocalMDExchange) {
     nixl_status_t status;
     std::string backend_name;
     for (const auto& name : std::set<std::string>{"GDS", "POSIX"}) {
+        const LogIgnoreGuard lig1("Error initializing GPU Direct Storage driver");
+        const LogIgnoreGuard lig2("createBackend: backend initialization error for 'GDS'");
```
guy-ealey-morag, line 549: Why are these warnings okay for this test?

test/gtest/metadata_exchange.cpp
```diff
@@ -506,8 +572,19 @@ TEST_F(MetadataExchangeTestFixture, EtcdSendLocalAndFetchRemote) {
     auto sleep_time = std::chrono::seconds(10);
     nixl_blob_t md;
 
-    ASSERT_EQ(dst.agent->fetchRemoteMD(src.name), NIXL_SUCCESS);
-    std::this_thread::sleep_for(sleep_time);
+    {
+        // Expected due to failure of checkRemoteMd() below?
```
guy-ealey-morag, line 582: `checkRemoteMD` is called outside of this scope, can you explain?

test/gtest/test_transfer.cpp
```diff
@@ -627,11 +634,16 @@ TEST_P(TestTransfer, GetXferTelemetryAPICfg) {
     // Disable telemetry from env var but through config, expecting a warning
     env.addVar("NIXL_TELEMETRY_ENABLE", "n");
 
+    const std::string rx = "ignoring the NIXL_TELEMETRY_ENABLE environment variable";
+    const LogIgnoreGuard lig(rx);
```
guy-ealey-morag:
```suggestion
    const LogIgnoreGuard lig("ignoring the NIXL_TELEMETRY_ENABLE environment variable");
```

I suggest preferring definition in one line when possible (applicable multiple times in this PR)

test/gtest/metadata_exchange.cpp
ColinNV: As I understand the test the intention is to try using the GDS backend and, if that fails or is unavailable, to use the POSIX backend. That's why GDS related errors are okay, and this was the least intrusive way to make that happen.

test/gtest/common.cpp
ColinNV: To what end? This is global state modeled by global variables.

test/gtest/main.cpp
ColinNV: Yes, I wanted it to grab more attention :-)

test/gtest/metadata_exchange.cpp
ColinNV: A few tests call `invalidateLocalMD()` as part of the test, calling it again here would lead to more warning/error logs. I changed those tests to get rid of the agent before finishing.

test/gtest/metadata_exchange.cpp
ColinNV: My understanding is that the expected failure of the `checkRemoteMD()` below is based on what `fetchRemoteMD()` does asynchronously, i.e. the failure of the check implies that some problem during the fetch is expected.

test/gtest/common.cpp
ColinNV: It's here because it's less/easier code than conditional instantiation of a `LogIgnoreGuard` but I have no strong opinion. I can change it.

test/gtest/metadata_exchange.cpp
ColinNV: Generally yes, but in this hard-coded test? Ok.

test/gtest/test_transfer.cpp
ColinNV: Ok.

test/gtest/common.cpp
ColinNV: Obsoleted by other change.

test/gtest/common.h
```diff
@@ -123,45 +126,49 @@ class PortAllocator {
...
- * Usage:
- *   scopedTestLogSink sink;
- *   // ... code that logs warnings ...
- *   EXPECT_EQ(sink.warningCount(), 1);
- *   EXPECT_EQ(sink.countWarnings("expected message"), 1);
- */
-class scopedTestLogSink {
+using log_ignore_entry_t = std::pair<std::regex, size_t>;
+
+class LogIgnoreGuard {
```
brminich, line 131: 👍

test/gtest/metadata_exchange.cpp
```diff
@@ -339,18 +376,33 @@ TEST_F(MetadataExchangeTestFixture, SocketSendLocalAndInvalidateLocal) {
...
+    const uint16_t port = 1234;
+    const std::string port_str = std::to_string(port);
+    send_args.ipAddr = ip_str;
+    send_args.port = port;
+    {
+        const LogIgnoreGuard lig1("poll timed out for ip_addr: " + ip_str +
+                                  " and port: " + port_str);
+        const LogIgnoreGuard lig2("Listener thread could not connect to IP " + ip_str +
+                                  " and port " + port_str);
+        const LogIgnoreGuard lig3("getsockopt gave error for ip_addr: " + ip_str +
```
brminich, line 396: if it can happen, does not it mean that lig2.count is 0?

test/gtest/test_transfer.cpp
```diff
@@ -627,11 +633,15 @@ TEST_P(TestTransfer, GetXferTelemetryAPICfg) {
     // Disable telemetry from env var but through config, expecting a warning
     env.addVar("NIXL_TELEMETRY_ENABLE", "n");
 
+    const LogIgnoreGuard lig("ignoring the NIXL_TELEMETRY_ENABLE environment variable");
```
brminich, line 636: IMO we will have to fix it. env vars should take preference
not realted to the PR though

test/gtest/metadata_exchange.cpp
ColinNV: Right, we should probably just check the sum.

test/gtest/main.cpp
```diff
@@ -76,10 +78,16 @@ namespace {
 int
 RunAllTests() {
     LogProblemCounter lpc;
-    std::optional<LogIgnoreGuard> lig;
+    std::list<LogIgnoreGuard> ligs;
+
+    if (std::getenv("AWS_BATCH_JOB_ID") != nullptr) {
+        ligs.emplace_back(aws_regex);
```
guy-ealey-morag, line 84: should we just put the string directly here? (same for `non_gpu_regex`)

ColinNV: I think in this case it really doesn't matter.

PR #1291, opened 2026-02-08
BACKEND/UCX: Minor utils refactoring.

ColinNV:
## What?
Improve some minor details in the UCX backend regarding code quality (and performance).

### DIFF
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -46,8 +46,7 @@ env:
   SLURM_GRES: gpu:2
   SLURM_MEM: 64G
   SLURM_MINCPUS: 24
-  SLURM_JOB_TIMEOUT: '01:30:00'
-  TEST_TIMEOUT: 50
+  SLURM_JOB_TIMEOUT: '02:10:00'
   STORAGE_DRIVER: overlay
   CI_IMAGE_TAG: "20260210-1"
 
@@ -121,7 +120,7 @@ steps:
 
   - name: Run CPP tests
     containerSelector: "{name: 'build_helper'}"
-    timeout: "${TEST_TIMEOUT}"
+    timeout: 50
     parallel: false
     run: |
       set -x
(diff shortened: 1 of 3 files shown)


src/plugins/ucx/ucx_utils.h
```diff
@@ -137,7 +137,7 @@ class nixlUcxEp {
            size_t len,
            uint32_t flags,
            nixlUcxReq *req = nullptr,
-           am_deleter_t deleter = nullptr);
+           const am_deleter_t &deleter = nullptr);
```
iyastreb, line 140: IMO this doesn't improve readability, I'm also curious why this complies, reference cannot be NULL. It seems it works because am_deteler is a pointer, not sure why to pass a pointer by reference then

ColinNV: It compiles because `am_deleter_t` is a `std::function` which can be implicitly constructed from `std::nullptr_t`. The change is not for readability, it's to pass the `std::function` by reference.

src/plugins/ucx/ucx_utils.h
```diff
@@ -199,18 +199,26 @@ class nixlUcxContext {
 private:
     /* Local UCX stuff */
     ucp_context_h ctx;
-    nixl_ucx_mt_t mt_type;
-    unsigned ucpVersion_;
+    const nixl_ucx_mt_t mt_type;
```
rakhmets: Maybe rename it to `mtType_` to align with our code style, as the corresponding code is changing in this PR.

ColinNV: Done.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -396,23 +397,36 @@ nixlUcxMtLevelIsSupported(const nixl_ucx_mt_t mt_type) noexcept {
...
+[[nodiscard]] unsigned
+makeUcpVersion() noexcept {
     unsigned major_version, minor_version, release_number;
     ucp_get_version(&major_version, &minor_version, &release_number);
-    ucpVersion_ = UCP_VERSION(major_version, minor_version);
+    return UCP_VERSION(major_version, minor_version);
+}
 
+[[nodiscard]] nixl_ucx_mt_t
+makeMtType(const bool prog_thread, const nixl_thread_sync_t sync_mode) {
```
rakhmets: noexcept?

PR #1296, opened 2026-02-09
posix: naming fixed

anton-nayshtut:
## What?
`HAVE_LIBAIO` renamed to `HAVE_POSIXAIO `

## Why?
`HAVE_LIBAIO` define was misleading. There is a library called [libaio](https://man7.org/linux/man-pages/man2/io_submit.2.html), but it's for Linux AIO. POSIX AIO only requires glibc.


benlwalker:
Looks good to me

aranadive (approved):

PR #1297, opened 2026-02-09
TEST/NIXL: Fail on unexpected error or warning in nixl_test.

ColinNV:
## What?
Catch more problems in nixl_test.

## How?
Similar to #1288.

### DIFF
--- a/test/nixl/nixl_test.cpp
+++ b/test/nixl/nixl_test.cpp
@@ -266,7 +266,8 @@ namespace {
 const std::string expected_for_target =
     "genNotif: no specified or potential backend could send the inter-agent notifications";
 const std::string expected_for_old_ucx =
-    "UCX version is less than 1.19, CUDA support is limited, including the lack of support for multi-GPU within a single process.";
+    "UCX version is less than 1.19, CUDA support is limited, including the lack of support for "
+    "multi-GPU within a single process.";
 
 class logProblemGuard : private absl::LogSink {
 public:


test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,44 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
 }
 
+namespace {
+const std::string expected =
+    "genNotif: no specified or potential backend could send the inter-agent notifications";
+
+class logProblemGuard : private absl::LogSink {
+public:
+    explicit logProblemGuard(const bool target) : target_(target) {
+        absl::AddLogSink(static_cast<absl::LogSink *>(this));
```
guy-ealey-morag, line 275: See this comment from previous PR: https://github.com/ai-dynamo/nixl/pull/1241#discussion_r2740839706
@rakhmets

ColinNV: This is only test code, it's the shorter solution, and it's a simple solution ... I don't see the problem?

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,44 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+    void
+    operator=(logProblemGuard &&) = delete;
+    void
+    operator=(const logProblemGuard &) = delete;
+
+    [[nodiscard]] static size_t
+    getProblemCount() noexcept;
+
+private:
+    const bool target_;
```
guy-ealey-morag:
```suggestion
    const bool isTarget_;
```

to not confuse with the `static const std::string target("target")` in the same file

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,44 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+    logProblemGuard(logProblemGuard &&) = delete;
+    logProblemGuard(const logProblemGuard &) = delete;
+
+    void
+    operator=(logProblemGuard &&) = delete;
+    void
+    operator=(const logProblemGuard &) = delete;
+
+    [[nodiscard]] static size_t
+    getProblemCount() noexcept;
```
guy-ealey-morag: This function is not implemented?

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,44 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+
+private:
+    const bool target_;
+
+    void
+    Send(const absl::LogEntry &entry) override {
+        if ((!target_) || (entry.text_message() != expected)) {
```
guy-ealey-morag: Shouldn't the `entry.log_severity()` be checked as well?

guy-ealey-morag: If I understand correctly, the `expected` variable is relevant only for the target role.
If it's the case I suggest renaming the variable to make it clear it's not always expected.

Also I think it would be more readable to negate the condition such that the exception throwing is skipped when it's true:
`if (target_ && (entry.text_message() == expected)) { return }`

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,44 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+            throw std::runtime_error("Unexpected NIXL warning or error detected!");
```
guy-ealey-morag: According to Cursor this can be problematic in multi-threaded code.
A possible solution is to just set a flag and check it after `runTarget/runInitiator` and fail if the flag is set.
wdyt?

test/nixl/nixl_test.cpp
ColinNV: Oops.

test/nixl/nixl_test.cpp
ColinNV: For now there don't seem to be any info messages, but I suppose it won't harm to exclude them.

ColinNV: Refactored to use a string variable instead of a flag.

test/nixl/nixl_test.cpp
ColinNV: Refactored, doesn't exist anymore.

test/nixl/nixl_test.cpp
ColinNV: The thread that throws will fall off a cliff and call `std::terminate()`, not sure see why we need to fail more gracefully here.

test/nixl/nixl_test.cpp
rakhmets: The purpose of placing `AddLogSink`/`RemoveLogSink` and overridden `Send()` to different classes is to have a single guard class and different heirs of `LogSink`. It makes sense to do it in this way in gtest, since it can be reused by other tests. But I'm not sure it's worth having here. Only if we move it to a common place, and reuse here and in gtest.

ColinNV: Yes, I get where you're coming from, it just didn't seem necessary here (yet).

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,54 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+    logProblemGuard(logProblemGuard &&) = delete;
+    logProblemGuard(const logProblemGuard &) = delete;
+
+    void
+    operator=(logProblemGuard &&) = delete;
+    void
+    operator=(const logProblemGuard &) = delete;
+
+private:
+    const std::string expected_;
```
rakhmets, line 291: I think this should be the last field, `Send` should be placed before.

test/nixl/nixl_test.cpp
```diff
@@ -258,6 +262,54 @@ static void runInitiator(const std::string &target_ip, int target_port, nixl_thr
...
+
+    void
+    Send(const absl::LogEntry &entry) override {
+        if (entry.log_severity() == absl::LogSeverity::kInfo) {
+            return;
+        }
+
+        if (entry.text_message() == expected_for_old_ucx) {
```
guy-ealey-morag, line 299: I think it would be cleaner and better long term to have a `expected_warnings` vector/set instead of having `expected_` and `expected_for_old_ucx`.
This avoids code repetition and will make it easily extendible if we want to allow more in the future.

(`expected_for_old_ucx` could be added to the vector by default, and `expected_` could be added if not empty)

ColinNV: Yes, I was optimistic, it already needs to be extended.

ColinNV:
At this point we will not clamp down on error and warning messages outside of gtest.

PR #1300, opened 2026-02-10
nixlbench: fix GUSLI READ consistency check failure and cleanup crash

aranadive:
## What?
Fix GUSLI READ consistency check failure and cleanup crash in nixlbench.

## Why?
Two bugs cause the reported failure:

1. File pre-population writes data at offset 0 but GUSLI reads at dev_offset (default 1MB). Additionally, if the file already exists with sufficient size, pre-population is skipped entirely, leaving stale data. This causes the consistency check to fail.

2. On consistency failure, exit() is called directly from checkConsistency(), bypassing GUSLI buffer deregistration. This triggers a ref_count assertion crash in the GUSLI library destructor (base_shm_element::~base_shm_element()).

## How?
Fix 1 - Correct file pre-population offset:
- Add ensureFileHasConsistencyData() helper that samples one page at the actual dev_offset to verify content, and only overwrites with the expected pattern (with a warning) when data differs.
- Replaces the misuse of initBasicDescFile() which was designed for file-backed storage backends, wrote at the wrong offset, and leaked file descriptors on the success path.

Fix 2 - Graceful shutdown on consistency failure:
- Change checkConsistency() from void to bool, returning the result instead of calling exit().
- Add validateTransfer() helper that centralizes the role/op-type descriptor selection logic into one place.
- Callers in processBatchSizes() now return EXIT_FAILURE on mismatch, allowing main()'s scope_guard to run deallocateMemory() and properly deregister GUSLI buffers before destruction.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -609,8 +609,9 @@ xferBenchNixlWorker::initBasicDescFile(size_t buffer_size, xferFileState &fstate
     memset(buf, XFERBENCH_TARGET_BUFFER_ELEMENT, buffer_size);
 
     size_t offset = start_offset;
+    char *write_ptr = static_cast<char *>(buf);
     while (buffer_size > 0) {
-        ssize_t rc = pwrite(fd, buf, buffer_size, offset);
+        ssize_t rc = pwrite(fd, write_ptr, buffer_size, offset);
         if (rc < 0) {
             std::cerr << "Failed to write to file: " << fd << " with error: " << strerror(errno)
                       << std::endl;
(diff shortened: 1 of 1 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -699,6 +699,75 @@ xferBenchNixlWorker::cleanupBasicDescBlk(xferBenchIOV &iov) {
...
+        while (remaining > 0) {
+            ssize_t rc = pwrite(fd, buf, remaining, offset);
+            if (rc < 0) {
+                std::cerr << "Failed to write to " << device.device_path << " at offset " << offset
+                          << ": " << strerror(errno) << std::endl;
+                free(buf);
+                close(fd);
+                return false;
+            }
+            remaining -= rc;
```
ovidiusm, line 759: This does not advance the buffer properly in case of partial writes, but not sure if this matters in practice

aranadive: Should be fixed now, here and another place.

PR #1301, opened 2026-02-10
CONFIG: Part 1 (env, api).

ColinNV:
## What?
Add new API for config access and implement for environment variables.

## Why?
Unify access to environment and prepare for config files.

## How?
This is the first PR with the API and environment access. It will be followed by a PR that implements config file reading, and one or more PRs that change all places in NIXL that call `getenv()` (and similar) to use the new API.

### DIFF
--- a/src/utils/common/configuration.h
+++ b/src/utils/common/configuration.h
@@ -15,8 +15,8 @@
  * limitations under the License.
  */
 
-#ifndef NIXL_SRC_UTILS_COMMON_CONFIG_H
-#define NIXL_SRC_UTILS_COMMON_CONFIG_H
+#ifndef NIXL_SRC_UTILS_COMMON_CONFIGURATION_H
+#define NIXL_SRC_UTILS_COMMON_CONFIGURATION_H
 
 #include <algorithm>
 #include <charconv>
@@ -31,9 +31,8 @@
 
 #include <absl/strings/str_join.h>
 
+#include "nixl_log.h"
 #include "nixl_types.h"
-
-#include "common/nixl_log.h"
 
 namespace nixl::config {
 
@@ -103,7 +102,7 @@ template<typename integer> struct integralTraits {
     convert(const std::string &value) {
         integer result;
         const auto status =
-            std::from_chars(begin(value), value.data() + value.size(), result, base(value));
+            std::from_chars(start(value), value.data() + value.size(), result, base(value));
         switch (status.ec) {
         case std::errc::invalid_argument:
             throw std::runtime_error("Invalid integer string '" + value + "' for type " +
@@ -133,7 +132,7 @@ private:
     }
 
     [[nodiscard]] static const char *
-    begin(const std::string &value) noexcept {
+    start(const std::string &value) noexcept {
         return value.data() + (isHex(value) ? 2 : 0);
     }
 };
--- a/test/gtest/configuration.cpp
+++ b/test/gtest/configuration.cpp
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-#include "common/config.h"
+#include "common/configuration.h"
 #include "gtest/gtest.h"
 
 #include <limits>


src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+            return true;
+        }
+
+        if (match(value, negative)) {
+            return false;
+        }
+
+        NIXL_ERROR << "Unknown value for bool '" << value << "' known are " << strJoin(positive)
+                   << " as positive and " << strJoin(negative) << " as negative (case insensitive)";
+        throw std::runtime_error("Conversion to bool failed");
```
rakhmets: Maybe it would be better to remove `NIXL_ERROR` from the method, move the explanation to `runtime_error`, and enable the exception handler to print a message. This way, the handler can decide whether it's really an error and further work is questionable, or whether it can be limited to a warning message.

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+struct convertTraits<integer, std::enable_if_t<std::is_signed_v<integer>>>
+    : integerTraits<integer, long long> {};
+
+template<typename integer>
+struct convertTraits<integer, std::enable_if_t<std::is_unsigned_v<integer>>>
+    : integerTraits<integer, unsigned long long> {};
+
+template<typename type, template<typename...> class traits = convertTraits>
+[[nodiscard]] nixl_status_t
+getNothrow(type &result, const std::string &env) {
```
rakhmets: noexcept?

ColinNV: This can still throw if the `std::string` for the environment variable's value can't allocate memory.

rakhmets: Then maybe it's better not to use `Nothrow` in the method name.

ColinNV: Right, though that's a very rare case in which nothing useful can happen anyhow.

(As you can probably guess my preferred solution would be to get rid of this method, but I'm not sure we can get away with that.)

src/utils/common/str_tools.h
```diff
@@ -62,4 +65,22 @@ class strEqual
...
+    std::string result = *iter;
+
+    for (++iter; iter != strings.end(); ++iter) {
+        result += delim;
+        result += *iter;
+    }
+    return result;
+}
+
 #endif
```
guy-ealey-morag:
```suggestion
#endif // NIXL_SRC_UTILS_COMMON_STR_TOOLS_H
```

src/utils/common/config.h
rakhmets: Maybe it can be implemented without this method. Since these are internal methods, we can use the exception approach wherever possible, until it comes to API methods that return status. So in the future, if there is a change in error handling, only the last part in interface methods will have to be changed.

src/utils/common/str_tools.h
```diff
@@ -62,4 +65,22 @@ class strEqual
           return true;
       }
 };
+
+template<typename container>
+std::string
+strJoin(const container &strings, const std::string &delim = ", ") {
```
guy-ealey-morag: I found that `absl::StrJoin` is already used in NIXL in several places, I think it does the same thing this function does.

src/utils/common/config.h
rakhmets: I also think that in most cases (if not all), incorrectly set environment variables should not lead to an error. But to issue a warning that the default value will be used.

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+
+#ifndef NIXL_SRC_UTILS_COMMON_CONFIG_H
+#define NIXL_SRC_UTILS_COMMON_CONFIG_H
+
+#include <algorithm>
+#include <cstdlib>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <strings.h> // strcasecmp
```
guy-ealey-morag:
```suggestion
#include <strings.h>
```

This is not usually done in NIXL

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+        NIXL_DEBUG << "Obtained environment variable " << name << "=" << value;
+        return std::string(value);
+    }
+    NIXL_DEBUG << "Missing environment variable " << name;
+    return std::nullopt;
+}
+
+[[nodiscard]] inline std::string
+getenvDefaulted(const std::string &name, const std::string &fallback) {
+    return getenvOptional(name).value_or(fallback);
```
guy-ealey-morag: Add a DEBUG log with the default value used

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+
+        NIXL_ERROR << "Unknown value for bool '" << value << "' known are " << strJoin(positive)
+                   << " as positive and " << strJoin(negative) << " as negative (case insensitive)";
+        throw std::runtime_error("Conversion to bool failed");
+    }
+
+private:
+    [[nodiscard]] static bool
+    match(const std::string &value, const std::vector<std::string> &haystack) noexcept {
+        static const auto pred = [&](const std::string &ref) {
```
guy-ealey-morag:
```suggestion
        const auto pred = [&](const std::string &ref) {
```

Making this `static` means that the lambda is created once, which means that `&value` will be captured only on the first call and then keep a stale reference.

guy-ealey-morag:
I suggest adding simple unit tests for the new env vars getters

src/utils/common/configuration.h
```diff
@@ -0,0 +1,166 @@
...
+            // TODO: Demangle? Manual name?
+            return NIXL_ERR_MISMATCH;
+        }
+    }
+    return NIXL_ERR_NOT_FOUND;
+}
+
+template<typename type, template<typename...> class traits = convertTraits>
+[[nodiscard]] type
+getValue(const std::string &env) {
```
guy-ealey-morag, line 163: I suggest making the naming more consistent: either `getValue` and `getValueOptional` or `get` and `getOptional`

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+getOptional(const std::string &env) {
+    if (const auto opt = getenvOptional(env)) {
+        return traits<type>::convert(*opt);
+    }
+    return std::nullopt;
+}
+
+template<typename type, template<typename...> class traits = convertTraits>
+[[nodiscard]] type
+getDefaulted(const std::string &env, const type &fallback) {
```
guy-ealey-morag: Is it possible to change `getDefaulted` to be an overload of `getValue`?

src/utils/common/config.h
ColinNV: > Maybe it can be implemented without this method. Since these are internal methods, we can use the exception approach wherever possible, until it comes to API methods that return status. So in the future, if there is a change in error handling, only the last part in interface methods will have to be changed.

Are you saying you agree to removing `getNothrow()`?

ColinNV: > I also think that in most cases (if not all), incorrectly set environment variables should not lead to an error. But to issue a warning that the default value will be used.

If an environment variable with the correct name was set the intention presumably was to use whatever value it configures; if that value can not be used/interpreted/converted then I don't think it's a good idea to continue with a potentially different value...

src/utils/common/config.h
ColinNV: Good catch.

src/utils/common/configuration.h
ColinNV: Hm, I find the current naming a better compromise between consistency, conciseness and expressiveness.

src/utils/common/config.h
ColinNV: It's probably not a good idea in the long run as in my experience it might very well collide with other overloads in the future.

ColinNV:
> I suggest adding simple unit tests for the new env vars getters

Of course, as soon as all details are nailed down.

src/utils/common/config.h
```diff
@@ -0,0 +1,166 @@
...
+
+template<typename integer, typename longLong> struct integerTraits {
+    static_assert(std::is_signed_v<integer> == std::is_signed_v<longLong>);
+    static_assert(std::is_unsigned_v<integer> == std::is_unsigned_v<longLong>);
+
+    [[nodiscard]] static integer
+    convert(const std::string &value) {
+        size_t pos = 0;
+        const longLong ll =
+            std::is_unsigned_v<integer> ? std::stoull(value, &pos) : std::stoll(value, &pos);
```
brminich: do we want to support octal and hex prefixes?

src/utils/common/config.h
brminich: i'd print a warining and continue with whatever value is set by default

src/utils/common/configuration.h
brminich: i tend to agree that `getValueOptional` looks more consistent

src/utils/common/str_tools.h
ColinNV: Then why does `str_tools.h` define a string split function which is also available in Abseil? I'll look into removing it.

ColinNV: #1309

src/utils/common/config.h
ColinNV: Good question. It's more flexible, but nowadays people often don't expect octal and might be surprised by the results when they write a supposedly decimal number with a leading 0... We could add hexadecimal support.

src/utils/common/config.h
ColinNV: After some offline discussions it was agreed that an environment variable that is set but has an illegal value should be hard error in order to not (relatively silently) hide mistakes.

src/utils/common/config.h
ColinNV: Again, an environment variable that is set but has an illegal value should be a hard error in order to not hide mistakes. But I'll move the message to the exception to remove redundant error reporting.

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,155 @@
...
+ */
+
+#include "common/config.h"
+#include "gtest/gtest.h"
+
+#include <limits>
+#include <stdlib.h>
+#include <string>
+
+namespace gtest {
```
rakhmets: Wouldn't it better to use the same namespace as the tested code?

src/utils/common/config.h
```diff
@@ -0,0 +1,173 @@
...
```
guy-ealey-morag: I found that `std::stoull` accepts negative strings as well and applies wrapping.
[click here](https://en.cppreference.com/w/cpp/string/basic_string/stoul.html#:~:text=If%20the%20minus%20sign%20was%20part%20of%20the%20input%20sequence%2C%20the%20numeric%20value%20calculated%20from%20the%20sequence%20of%20digits%20is%20negated%20as%20if%20by%20unary%20minus%20in%20the%20result%20type%2C%20which%20applies%20unsigned%20integer%20wraparound%20rules.)

We should probably first check that the string has only numeric characters in it before parsing.

src/utils/common/config.h
```diff
@@ -0,0 +1,173 @@
...
+
+template<typename integer>
+struct convertTraits<integer, std::enable_if_t<std::is_signed_v<integer>>>
+    : integerTraits<integer, long long> {};
+
+template<typename integer>
+struct convertTraits<integer, std::enable_if_t<std::is_unsigned_v<integer>>>
+    : integerTraits<integer, unsigned long long> {};
+
+// Keep this function, too, or just the regular three below?
```
guy-ealey-morag: Are you planning on using it yourself? If not then it might not be necessary
If you decide to keep it then it needs to be added to the unit tests

src/utils/common/config.h
```diff
@@ -0,0 +1,173 @@
...
+
+        if (match(value, positive)) {
+            return true;
+        }
+
+        if (match(value, negative)) {
+            return false;
+        }
+
+        const std::string msg = "Conversion to bool failed for string  '" + value + "' known are " +
```
guy-ealey-morag:
```suggestion
        const std::string msg = "Conversion to bool failed for string '" + value + "' known are " +
```

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,155 @@
...
+    testUnsigned() {
+        testSimpleSuccess("0", T(0));
+        testSimpleSuccess("1", T(1));
+        testSimpleSuccess("42", T(42));
+        const T maxValue = std::numeric_limits<T>::max();
+        const std::string maxString = std::to_string(maxValue);
+        testSimpleSuccess(maxString, maxValue);
+
+        testSimpleFailure<T>("");
+        testSimpleFailure<T>("-");
```
guy-ealey-morag, line 171:
```
testSimpleFailure<T>("-0");
testSimpleFailure<T>("-1");
testSimpleFailure<T>("-42");
```

Should these also fail?

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,155 @@
...
+namespace {
+    template<typename T>
+    void
+    testSigned() {
+        testSimpleSuccess("0", T(0));
+        testSimpleSuccess("1", T(1));
+        testSimpleSuccess("-1", T(-1));
+        testSimpleSuccess("42", T(42));
+        testSimpleSuccess("-42", T(-42));
+        const T minValue = std::numeric_limits<T>::min();
```
guy-ealey-morag:
```suggestion
        const T min_value = std::numeric_limits<T>::min();
```

local variables naming convention (according to the `CodeStyle.md`)

The same for `maxValue`, `maxString`, etc

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,155 @@
...
+    ASSERT_FALSE(nixl::config::getenvOptional(undefined).has_value());
+    ASSERT_EQ(nixl::config::getenvDefaulted(undefined, fallback), fallback);
+}
+
+namespace {
+
+    template<typename T>
+    void
+    testSimpleSuccess(const std::string &input, const T value) {
+        ASSERT_EQ(::setenv(variable.c_str(), input.c_str(), 1), 0);
```
guy-ealey-morag, line 68: I suggest starting by unsetting the env var, and testing that `getValueDefaulted` gives the default value

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,155 @@
...
+namespace gtest {
+
+namespace {
+
+    const std::string variable = "NIXL_CONFIG_TEST";
+    const std::string undefined = "ASDLFHASLK1298159816";
+
+} // namespace
+
+TEST(Config, EnvWrapper) {
```
guy-ealey-morag, line 34: What is tested here that is not done in the other tests?

src/utils/common/config.h
brminich: ok, let's add just hex support

test/gtest/configuration.cpp
ColinNV: All the other tests seem to be in `namespace gtest` so I assumed that was the preferred way...

src/utils/common/config.h
ColinNV: Switching to charconv.

src/utils/common/config.h
ColinNV: I don't want to implement or use it, but still need to confirm whether it can be removed. If not it will be tested.

src/utils/common/config.h
ColinNV: Ok. Also for signed integers?

test/gtest/configuration.cpp
ColinNV: The code is layered, the tests are layered...

test/gtest/configuration.cpp
ColinNV: Why shouldn't they fail for unsigned integers?

test/gtest/configuration.cpp
ColinNV: Right, that should be tested, but not here, it doesn't depend on `input`.

test/gtest/configuration.cpp
guy-ealey-morag: I meant that if these cases should fail then they should be added

test/gtest/configuration.cpp
rakhmets: I see... I don't know why it was done in this way in other tests. It doesn't add any benefit and doesn't solve any issues....

test/gtest/configuration.cpp
ColinNV: Gotcha, the last commit adds a failure test for negative integer.

test/gtest/configuration.cpp
ColinNV: I agree; for now I'll stick to the established style.

src/utils/common/config.h
```diff
@@ -0,0 +1,188 @@
...
+            }
+            break;
+        }
+        return result;
+    }
+
+private:
+    [[nodiscard]] static bool
+    isHex(const std::string &value) noexcept {
+        return std::is_unsigned_v<integer> && (value.size() > 2) && (value[1] == 'x');
```
brminich: maybe 
```
           value[0] == '0' &&
           (value[1] == 'x' || value[1] == 'X');
```

src/utils/common/config.h
brminich: pls keep it and remove comment above

ColinNV: Done, including tests.

test/gtest/configuration.cpp
rakhmets: But you already fixed namespaces in `srs` adding `nixl::config`. I think it makes sense to fix it in `test/gtest` too.
Because `gtest` a little bit confusing namespace. It doesn't belong to gtest framework, and it also forces the use of `nixl::config` here.

test/gtest/configuration.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "common/config.h"
+#include "gtest/gtest.h"
+
+#include <limits>
+#include <stdlib.h>
+#include <string>
+
+namespace gtest {
+
+namespace {
```
rakhmets, line 25: It should be outside of the current namespace.

test/gtest/configuration.cpp
ColinNV: Ok, then let's change it.

src/utils/common/configuration.h
guy-ealey-morag, line 1: I just noticed that there is another `config.h` that is generated by the build system and is used by multiple files in the project.
I suggest renaming this file to avoid confusion between the two.

src/utils/common/config.h
```diff
@@ -0,0 +1,188 @@
...
+            ((value[1] == 'x') || (value[1] == 'X'));
+    }
+
+    [[nodiscard]] static int
+    base(const std::string &value) noexcept {
+        return isHex(value) ? 16 : 10;
+    }
+
+    [[nodiscard]] static const char *
+    begin(const std::string &value) noexcept {
```
guy-ealey-morag: I suggest using a different name here to differ it from `std::begin` of iterators

src/utils/common/config.h
```diff
@@ -0,0 +1,188 @@
...
+#include <strings.h>
+#include <type_traits>
+#include <typeinfo>
+#include <vector>
+
+#include <absl/strings/str_join.h>
+
+#include "nixl_types.h"
+
+#include "common/nixl_log.h"
```
guy-ealey-morag:
```suggestion
#include "nixl_log.h"
```

My IDE couldn't find it

ColinNV: Ok

src/utils/common/configuration.h
ColinNV: There's also a `config.h` in the UCX plugin, but I suppose we can rename this one.

PR #1302, opened 2026-02-10
NUMA-aware rail selection policy for DRAM_SEG memory type (libfabric NIXL backend)

oa-aws:
## What?
Added a NUMA-aware rail selection policy for DRAM_SEG memory type, such that rails are restricted to total NUMA node bandwidth capacity with which the memory buffer is associated, while ensuring that the bandwidth capacity of each topmost PCIe switch (connected to hwloc package) is not exceeded. User can override bandwidth limitation settings with environment variable NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG, specified in Gbps (e.g. 100, 200, etc.). Corresponding custom libfabric backend parameter is max_bw_per_dram_seg.

## Why?
Normally, when issuing requests with DRAM_SEG from nixl agent, there is no justification to exceed the PCIe switch capacity, unless user explicitly wishes to do so.

## How?
The machine topology is investigated during backend/plugin startup phase, and PCIe switches connected to hwloc package are listed per NUMA node, so total capacity (with corresponding EFA devices connected to each switch) is recorded. Later, when a DRAM_SEG memory is registered, rails are selected according to computed limitation, or according to user override that may increase or decrease the limitation.

## Summary by CodeRabbit

* **New Features**
  * NUMA-aware, policy-driven DRAM rail selection for improved memory placement and bandwidth utilization
  * Environment-variable overrides and new config helpers for string/int parameters; rail ID reporting utility
  * Richer topology discovery with per‑NIC and per‑NUMA bandwidth metrics and PCIe speed/parent-switch metadata

* **Bug Fixes**
  * Earlier initialization failure reporting and stricter validation of rail/bandwidth limits

* **Tests**
  * Extensive topology unit tests and multiple static topology fixtures for broad scenario coverage

### DIFF
--- a/src/utils/libfabric/meson.build
+++ b/src/utils/libfabric/meson.build
@@ -1,5 +1,5 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/test/unit/utils/libfabric/meson.build
+++ b/test/unit/utils/libfabric/meson.build
@@ -1,5 +1,5 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-FileCopyrightText: Copyright (c) 2025 Amazon.com, Inc. and affiliates.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/test/unit/utils/libfabric/topo/g5.48xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/g5.48xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE topology SYSTEM "hwloc2.dtd">
 <topology version="2.0">
--- a/test/unit/utils/libfabric/topo/g6.48xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/g6.48xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE topology SYSTEM "hwloc2.dtd">
 <topology version="2.0">
--- a/test/unit/utils/libfabric/topo/p3dn.24xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/p3dn.24xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE topology SYSTEM "hwloc2.dtd">
 <topology version="2.0">
--- a/test/unit/utils/libfabric/topo/p4d.24xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/p4d.24xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE topology SYSTEM "hwloc2.dtd">
 <topology version="2.0">
--- a/test/unit/utils/libfabric/topo/p5.48xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/p5.48xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE topology SYSTEM "hwloc2.dtd">
 <topology version="2.0">
--- a/test/unit/utils/libfabric/topo/p5en.48xl-topo.xml
+++ b/test/unit/utils/libfabric/topo/p5en.48xl-topo.xml
@@ -1,3 +1,19 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 Amazon.com, Inc. and affiliates. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!--
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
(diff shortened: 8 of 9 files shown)


src/utils/libfabric/libfabric_rail_manager.h
```diff
@@ -34,6 +34,31 @@
...
+
+// rail selection policy parent interface type
+class nixlLibfabricRailSelectionPolicy {
+public:
+    virtual
+    ~nixlLibfabricRailSelectionPolicy() {}
+    nixlLibfabricRailSelectionPolicy(const nixlLibfabricRailSelectionPolicy&) = delete;
+    nixlLibfabricRailSelectionPolicy& operator=(const nixlLibfabricRailSelectionPolicy&) = delete;
+
+    /** Loads the policy from custom engine onfiguration. */
```
yexiangd: nit: configuration

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -943,3 +1243,417 @@ fi_hmem_iface
...
+                      << " (out of range) for device " << device << ", skipping";
+            continue;
+        }
+
+        // get parent switch bus id
+        uint16_t numa_node_id = UINT16_MAX;
+        size_t nic_link_speed = 0;
+        uint8_t bus_id = UINT8_MAX;
+        size_t switch_link_speed = 0;
+        ;
```
yexiangd: unnecessary semicolon

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -943,3 +1243,417 @@ fi_hmem_iface
...
+        if (!topology->getPcieDevData(
+                device, numa_node_id, nic_link_speed, bus_id, switch_link_speed)) {
+            NIXL_WARN << "Could not find additional data for device " << device << ", skipping";
+            continue;
+        }
+
+        if (numa_node_id >= numa_switch_map.size()) {
+            NIXL_WARN << "Invalid NUAM node id " << numa_node_id << " for device " << device
```
yexiangd: NUMA

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -656,6 +739,314 @@ nixlLibfabricTopology::isEfaDevice(hwloc_obj_t obj) const {
...
+        if (curr_speed != speed) {
+            NIXL_WARN << "Non-uniform NUMA node " << i << " PCIe capacity: " << curr_speed
+                      << " Gbps (expected " << speed << " Gbps)";
+            speed_uniform = false;
+        }
+        total_speed += curr_speed;
+    }
+    if (speed_uniform) {
+        avg_numa_speed = speed;
+        NIXL_DEBUG << "NUMA PCIe capacity is uniform accorss all nodes with link bandwidth: "
```
yexiangd: nit: accorss -> across in a few other places

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -90,6 +246,60 @@ nixlLibfabricRailManager::~nixlLibfabricRailManager() {
...
 
+nixl_status_t
+nixlLibfabricRailManager::init(const nixl_b_params_t &custom_params) {
+    // load from config or compute bandwidth limit per NUMA node
+    // then from that deduce rail count limit
+    // finally choose appropriate rail selection policy:
+    // 1 - limit is within a single NUMA node bound --> use numa-aware policy
+    // 2 - limit exceeds NUMA node capacity, but not machine capacity --> use numa-aware policy
+    // 3 - limit exceeds entire machine capcaity --> use default (all) policy
+    // in case no EFA device is found, we use defaul tpolicy
```
yexiangd: `default policy`

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -37,6 +40,159 @@ resetSeqId();
...
+
+// loads backend custom string configuration item with possible env var override
+static nixl_status_t
+getCustomStringParam(const nixl_b_params_t &custom_params,
+                     const std::string &key,
+                     std::string &value);
+
+// loads backend custom integer configuration item with possible env var override
+static nixl_status_t
+getCustomIntParam(const nixl_b_params_t &custom_params, const std::string &key, uint64_t &value);
```
yexiangd: do you mind keeping the helper functions in `libfabric_common.cpp` ? we have configuration constants defined in `libfabric_common.h`

src/utils/libfabric/libfabric_topology.h
```diff
@@ -79,14 +86,22 @@ class nixlLibfabricTopology {
...
 
+    // invalid NUMA node id
+    const uint16_t INVALID_NUMA_NODE_ID = UINT16_MAX;
+
     // Data structures for NIXL topology-aware grouping algorithm
     struct NicInfo {
         std::string libfabric_name;
         hwloc_obj_t hwloc_node;
+        size_t line_speed; // NOTE: multiple of 1000^3
+        size_t upstream_link_speed; // NOTE: multiple of 1024^3
```
yexiangd: this is GiB/s ?  it's conflicting with other occurrences that you commented: `// NOTE: speed is given in GB/s as float`

yexiang-aws (commented):
This is a XXL size change, it's easy to lost in the computation details and different scenarios you try to cover here. Can you see if you can add some unit tests to validate different topologies, selection_policy and user input different bandwidth?

src/utils/libfabric/libfabric_rail_manager.h
oa-aws: Fixed

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Fixed

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Fixed

src/utils/libfabric/libfabric_topology.cpp
oa-aws: Fixed

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Fixed

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Moved to libfabric_common.h/cpp

src/utils/libfabric/libfabric_topology.h
oa-aws: Fixed comments, also in places where used.

test/unit/utils/libfabric/libfabric_topology_test.cpp
```diff
@@ -18,14 +18,292 @@
...
+    {.enable = false,
+     .instance_type = "p6e-gb200",
+     .topo_file = "p6e-gb200-topo.xml",
+     .numa_node_count = 2,
+     .nic_count = 16,
+     .nic_line_speed = 400,
+     .nic_upstream_link_speed = 32,
+     .switch_count = 2,
+     .numa_capacity = 64,
+     .numa_rail_count = 4},
```
yexiangd: did you verify this config on gb200? where is the topo file p6e-gb200-topo.xml ?  

it seems you set .enable=false for the instance types to indicate that you don't have topo files yet, but wanted to cover?  If they are not validated, it's ok to drop them.

src/utils/libfabric/meson.build
```diff
@@ -34,10 +34,14 @@ libfabric_utils_headers = files(
 # Find hwloc dependency
 hwloc_dep = dependency('hwloc', required: true)
 
+# Find numa dependency
+numa_dep = dependency('numa', required: true)
```
yexiangd, line 38: good that `libnuma-dev` is already a dependency in the CI, so we shouldn't break the libfabric build in CI.  Please double check, and document the additional libnuma dependency (check where we specifically called out hwloc)

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -336,6 +547,83 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+        return false;
+    }
+
+    // now compute rail limit based on bandwidth limit
+    max_rails = 0;
+    max_bw = 0;
+
+    // get bandwidth from configuration or environment variable, and then deduce rail count
+    // NOTE: corresponding env var is NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG, and is specified in
+    // Gigabits per second also return value from getCustomIntParam() is not that important
```
yexiangd: > also return value from getCustomIntParam() is not that important

sorry, what do you mean here?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -336,6 +547,83 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+    nixl_status_t res =
+        LibfabricUtils::getCustomIntParam(custom_params, "max_bw_per_dram_seg", max_bw);
+    if (res != NIXL_SUCCESS) {
+        // bandwidth limit could not be obtained from user (either user did not specify, or
+        // configuration was malformed) so comput bandwidth limit from topology, and then deduce
```
yexiangd: nit: comput -> compute

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -336,6 +547,83 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+                     "rail selection for DRAM memory type aborted";
+        return false;
+    }
+
+    // now compute rail limit based on bandwidth limit
+    max_rails = 0;
+    max_bw = 0;
+
+    // get bandwidth from configuration or environment variable, and then deduce rail count
+    // NOTE: corresponding env var is NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG, and is specified in
```
yexiangd, line 554: Do you mind add user-facing documentation for `NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG` usage and its default behavior/value? it's getting buried in the code. Or you can split the change in separate commit/PR.  This PR is getting very large.

some reference: https://github.com/aws/aws-ofi-nccl/blob/master/include/nccl_ofi_param.h


Also do you mind updating NUMA-aware rail selection in the Libfabric README, and provide examples of when to override bandwidth limits ? Again, break down into standalone commits

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -943,3 +1234,445 @@ fi_hmem_iface
...
+nixlLibfabricNumaRailSelectionPolicy::selectSwitchRails(SwitchRailData &switch_rail_data,
+                                                        std::vector<size_t> &selected_rails) {
+    // verify input
+    if (selected_rails.size() > max_rails_) {
+        NIXL_ERROR << "Too many rails selected (internal error): selected " << selected_rails.size()
+                   << " while required " << max_rails_;
+        return false;
+    }
+    if (selected_rails.size() == max_rails_) {
+        NIXL_TRACE << "Reduandant call to selectSwitchRails() (iternal error)";
```
yexiangd: typo : internal ?
is it error? but your return is true

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -336,6 +547,83 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+
+    // now compute rail limit based on bandwidth limit
+    max_rails = 0;
+    max_bw = 0;
+
+    // get bandwidth from configuration or environment variable, and then deduce rail count
+    // NOTE: corresponding env var is NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG, and is specified in
+    // Gigabits per second also return value from getCustomIntParam() is not that important
+    nixl_status_t res =
+        LibfabricUtils::getCustomIntParam(custom_params, "max_bw_per_dram_seg", max_bw);
```
yexiangd, line 557: type mis-match,  `max_bw` is `size_t`, `value` is `uint64_t`
it happens to be on 64-bit systems, size_t is typically uint64_t
> getCustomIntParam(const nixl_b_params_t &custom_params, const std::string &key, uint64_t &value);

src/utils/libfabric/libfabric_topology.h
```diff
@@ -83,10 +90,24 @@ class nixlLibfabricTopology {
     struct NicInfo {
         std::string libfabric_name;
         hwloc_obj_t hwloc_node;
+        // NOTE: NIC line speed is in multiples of 1000^3, as reported by fi_info - this is compared
+        // against user env override, which is specified as 100, 200, etc, in Gbps (Gigabit per
+        // second) so we can deduce number of rails from user override
+        size_t line_speed;
+        // NOTE: upstream link speed is in multiples of 1024^3 (Gigabit per second), as reported by
```
yexiangd: > 1024^3 (Gigabit per second)

1024^3 is Gibibit

src/utils/libfabric/meson.build
oa-aws: Without this dependency, build of libplugin_LIBFABRIC.so fails with unresolved externals - so plugin requires this dependency - adding doc in plugin README. 

I am not sure I understand the CI dependency, and the possibility of breaking libfabric build in CI. Where can I see it?

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Fixing comment (there was a left over). Besides that it means that bandwidth limit for DRAM_SEG can be configured during backend/plugin creation via max_bw_per_dram_seg configuration key (in custom_params map), and user can override the value with env var NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG. From this bandwidth limit (whether computed or coming from user override), we can deduce the rail count limit (per NUMA node).

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Fixed

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Added details to nixlLibfabricRailManager::init() function documentation (I think this should be done in this commit).

README update as well in this PR.

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: This indicates an internal logic error, but should not be reported as error to the user (because we have the required number of rails). Removing the "internal error" from the trace message, and adding a comment explaining this situation.

src/utils/libfabric/libfabric_topology.h
oa-aws: There is difference in reporting by fi_getinfo and hwloc. fi_getinfo reports NIC speed as whole multiples of 1000^3, i.e. 100,000,000,000, which is 100 Gigabit per second. This value is converted to 100, 200, etc. 

hwloc reports PCIe link upstream speed as float type in multiples of 1024^3, originally as GB/s (Gigabyte per second), i.e. 31.5077. The value reported by hwloc is converted to Gigabit per second, and to size_t type (for deterministic integer computation).

Fixing comment a bit to be more clear.

test/unit/utils/libfabric/libfabric_topology_test.cpp
oa-aws: Removing p6e-gb200 from test. This is not an expected runtime topology.

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Correct. Fixed.

src/plugins/libfabric/README.md
```diff
@@ -47,6 +50,37 @@ $ cd <name_of_build_dir>
 $ ninja && ninja install
 ```
 
+## Runtime Configuration
+
+The following configuration controls the runtime behavior of the plugin:
+
+### max_bw_per_dram_seg
```
yexiangd, line 57: should be the env var name instead of the var name from code. Do you mind proof-reading this paragraph again, from user's perspective? some places don't make sense to me:
1. > User can override during plugin creation (in code), by specifying a value (string that can be parsed as integer) for "max_bw_per_dram_seg" in the custom parameter map of the plugin.

overkill? The end users are supposed to set the env var  `NIXL_LIBFABRIC_MAX_BW_PER_DRAM_SEG` instead of changing max_bw_per_dram_seg in code. Just keeping the next line is sufficient

2. > Specified as integer multiple of 1000^3 as common in NIC specification (e.g. 100, 200, 400, etc.)

do you mind just simply saying the unit, Gbps ? same as below `A multiple of 1000^3 as common in NIC specification `   Or just embed that in your en var name for clarity

src/utils/libfabric/libfabric_topology.h
yexiangd: I'm just calling out upstream link speed is in Gbps, which is 1000^3, not 1024^3, not the binary system

src/plugins/libfabric/README.md
oa-aws: Can the "user" also be code that is building nixlAgent object and providing initialization params for nixlAgent::createBackend()?

PR #1303, opened 2026-02-10
feature: add rust metadata support

cheese-head:
## What?

Add explicit metadata-aware variants for descriptor registration:
- `RegDescList::add_storage_desc_with_metadata()` — counterpart to `add_storage_desc()`
- `Agent::register_memory_with_metadata()` — counterpart to `register_memory()`

These call `NixlDescriptor::metadata()` and pass the result through to `add_desc_with_meta()` during registration. The existing `add_storage_desc()` and `register_memory()` methods remain unchanged (no metadata).

Also keeps the `NixlDescriptor::metadata()` trait method (default returns empty vec) from the original PR.

## Why?

KVBM in [dynamo](https://github.com/ai-dynamo/dynamo) needs to pass metadata (e.g., object storage keys) into NIXL during memory registration for the OBJ backend. The current `add_storage_desc` / `register_memory` path has no way to carry metadata through to `add_desc_with_meta`.

## How?

Follows the same pattern already established at the lower level:
- `add_desc(addr, len, dev_id)` — no metadata
- `add_desc_with_meta(addr, len, dev_id, metadata)` — with metadata

Now mirrored at the higher level:
- `add_storage_desc(desc)` / `register_memory(desc, opt)` — no metadata
- `add_storage_desc_with_metadata(desc)` / `register_memory_with_metadata(desc, opt)` — calls `desc.metadata()` and forwards it

Closes: https://github.com/ai-dynamo/nixl/pull/1096

### DIFF
--- a/src/bindings/rust/src/descriptors/reg.rs
+++ b/src/bindings/rust/src/descriptors/reg.rs
@@ -224,9 +224,10 @@ impl<'a> RegDescList<'a> {
         let addr = unsafe { desc.as_ptr() } as usize;
         let len = desc.size();
         let dev_id = desc.device_id();
+        let metadata = desc.metadata();
 
-        // Add to list
-        self.add_desc(addr, len, dev_id);
+        // Add to list with metadata
+        self.add_desc_with_meta(addr, len, dev_id, &metadata);
         Ok(())
     }
 
--- a/src/bindings/rust/src/descriptors/reg.rs
+++ b/src/bindings/rust/src/descriptors/reg.rs
@@ -224,6 +224,45 @@ impl<'a> RegDescList<'a> {
         let addr = unsafe { desc.as_ptr() } as usize;
         let len = desc.size();
         let dev_id = desc.device_id();
+
+        // Add to list without metadata
+        self.add_desc(addr, len, dev_id);
+        Ok(())
+    }
+
+    /// Add a descriptor from a type implementing NixlDescriptor, including metadata.
+    ///
+    /// This is the metadata-aware variant of [`add_storage_desc`]. It calls
+    /// [`NixlDescriptor::metadata()`] on the descriptor and passes the result
+    /// through to the backend via [`add_desc_with_meta`].
+    ///
+    /// Use this when the backend requires metadata (e.g., OBJ backend for
+    /// object storage keys).
+    ///
+    /// # Safety
+    /// The caller must ensure that:
+    /// - The descriptor remains valid for the lifetime of the list
+    /// - The memory region pointed to by the descriptor remains valid
+    pub fn add_storage_desc_with_metadata(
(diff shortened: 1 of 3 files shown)


src/bindings/rust/src/descriptors/reg.rs
```diff
@@ -225,11 +225,51 @@ impl<'a> RegDescList<'a> {
...
+    /// Use this when the backend requires metadata (e.g., OBJ backend for
+    /// object storage keys).
+    ///
+    /// # Safety
+    /// The caller must ensure that:
+    /// - The descriptor remains valid for the lifetime of the list
+    /// - The memory region pointed to by the descriptor remains valid
+    pub fn add_storage_desc_with_metadata(
+        &mut self,
+        desc: &'a dyn NixlDescriptor,
```
ryanolson, line 248: not sure if the NixlDescriptor should carry the metadata.  let me think on this some more. 

this feels right
```
add_desc_with_meta(addr, len, dev_id, metadata)
```

and this feels
```
add_descriptor_with_metdata(desc: &'a dyn NixlDescriptor, metadata: &'a NixlMetadata)
```

from the perspective of a BlockId in a Layout, it has no concept of it's NixlMetdata, only the ImmutableBlock does.

Now, if we have a NixlObject, then yes, this entity would carry with it knowledge of the metadata

src/bindings/rust/src/lib.rs
```diff
@@ -501,6 +501,15 @@ pub trait NixlDescriptor: MemoryRegion {
...
     /// Get the device ID for this memory region
     fn device_id(&self) -> u64;
+
+    /// Get optional metadata for this descriptor.
+    ///
+    /// Metadata is used by some backends (e.g., OBJ for object storage keys).
+    /// Returns owned bytes since they're copied during registration anyway.
+    /// Default implementation returns empty vec.
+    fn metadata(&self) -> Vec<u8> {
+        Vec::new()
```
ryanolson, line 511: Probably should be an `Option<Vec<u8>>`

there is cost with instantiating a Vec that should be avoided in the None path.

ryanolson (changes requested):
A lot of food for thought in this PR.

Might we consider a new trait, like NixlObjectMetdata that looks like

```
pub trait NixlObjectMetadata: NixlDescriptor {
   //
}
```

I think we need to work through the API a bit more and the implications.

src/bindings/rust/src/descriptors/reg.rs
cheese-head: So somewhat how nixlBlobDesc is a nixlBasicDesc with metaInfo

roiedanino:
Hi @cheese-head can you please set copy-right year to 2025-2026 in the modified files?

PR #1310, opened 2026-02-11
CI: Workaround for UCX bug.

ColinNV:
Workaround for the UCX bug that was fixed in UCX PR 11149, needed until that PR is "visible" here.

### DIFF
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -43,10 +43,10 @@ env:
   SLURM_NODES: 1
   SLURM_PARTITION: mizu
   SLURM_HEAD_NODE: scctl
-  SLURM_GRES: gpu:2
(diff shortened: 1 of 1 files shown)


.gitlab/test_nixlbench.sh
```diff
@@ -38,6 +38,7 @@ export CPATH=${INSTALL_DIR}/include:$CPATH
 export PATH=${INSTALL_DIR}/bin:$PATH
 export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
 export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
+export UCX_GDR_COPY_SHARED_MD=n
```
iyastreb, line 42: should we add TODO/FIXME to remove it once not needed?

PR #1316, opened 2026-02-12
meson.build: ucx_gpu_device_api_v2_available fixed

anton-nayshtut:
## What?
This PR fixes `ucx_gpu_device_api_v2_available` meson var assignment.

## Why?
Without this fix, I get the following error:

> ERROR: Unknown variable "ucx_gpu_device_api_v2_available"

meson 1.3.2


anton-nayshtut:
> Is this an urgent fix? This variable will be removed by #1342.

No, it’s not. That said, I don’t see a reason we can’t merge it now, and then #1342 can remove it once it's merged.

PR #1322, opened 2026-02-13
NIXL: Optimize nixlDescList iteration

iyastreb:
## What?
Optimize iteration over nixlDescList

## Why?
This change alone improves TTFT by 5% on large dimensions

## How?
inlining operator[] and removing virtual modifier

Perf testing:
```
      BEFORE    AFTER
prep: 3183194   1457517   // write access is much faster (inline non-virtual)
post:  816547    738207    // read access is faster (inline)
```

## Summary by CodeRabbit

* **Refactor**
  * Improved type safety and consistency in descriptor access mechanisms with updated method signatures.
  * Enhanced validation of descriptor list integrity at access time through runtime assertions.


src/api/cpp/nixl_descriptors.h
```diff
@@ -316,17 +316,18 @@ class nixlDescList {
      */
     inline bool
     isEmpty() const {
-        return (descs.size() == 0);
+        return (descs.empty());
```
brminich:
```suggestion
        return descs.empty();
```

src/infra/nixl_descriptors.cpp
```diff
@@ -206,22 +206,6 @@ nixlDescList<T>::nixlDescList(nixlSerDes* deserializer) {
...
-    // To be added only in debug mode
-    // if (index >= descs.size())
-    //     throw std::out_of_range("Index is out of range");
-    return descs[index];
-}
-
-// Setter
-template <class T>
-inline T& nixlDescList<T>::operator[](unsigned int index) {
-    assert(index < descs.size());
```
brminich, line 203: maybe we can use NIXL_ASSERT in new funcs?

src/api/cpp/nixl_descriptors.h
```diff
@@ -316,17 +316,18 @@ class nixlDescList {
...
-    const T &
-    operator[](unsigned int index) const;
-    virtual T &
-    operator[](unsigned int index);
+    inline const T &
+    operator[](unsigned int index) const {
+        return descs[index];
+    }
+
+    inline T &
```
brminich: is there a risk of breaking ABI?

src/infra/nixl_descriptors.cpp
iyastreb: I tried, but NIXL_ASSERT macro is not available in the api
We can either keep plain assert, or nothing

src/api/cpp/nixl_descriptors.h
iyastreb: Yes, unfortunately this breaks ABI compatibility

src/api/cpp/nixl_descriptors.h
iyastreb: done

src/api/cpp/nixl_descriptors.h
```diff
@@ -316,17 +316,18 @@ class nixlDescList {
...
 
-    /**
-     * @brief Operator [] overloading, get/set descriptor at [index].
-     *        Can throw std::out_of_range exception.
-     */
-    const T &
-    operator[](unsigned int index) const;
-    virtual T &
-    operator[](unsigned int index);
+    inline const T &
```
rakhmets: Why do we need `inline` here? Isn't this an implicitly inline function?

iyastreb: yes, redundant, will remove

src/api/cpp/nixl_descriptors.h
```diff
@@ -314,19 +314,20 @@ class nixlDescList {
...
     const T &
-    operator[](unsigned int index) const;
-    virtual T &
-    operator[](unsigned int index);
+    operator[](unsigned int index) const noexcept {
+        return descs[index];
+    }
+
+    T &
+    operator[](unsigned int index) noexcept {
```
rakhmets: I am afraid that it is not good to use `noexcept` here. The behavior when accessing an element beyond the size of the `std.::vector` is undefined. Thus, anything can happen, theoretically, even an exception. So, I suggest to use the same specifiers as `std.::vector::operator[]`.

rakhmets: Can we use `size_t` instead of `unsigned int`?

iyastreb: noexcept was proposed by @ColinNV in another PR: https://github.com/ai-dynamo/nixl/pull/1324#discussion_r2813223519
But I think you're right, since it's not noexcept in std::vector let's remove it also here

iyastreb: Yes, all of indexes must be size_t, I addressed it in https://github.com/ai-dynamo/nixl/pull/1324
Will also update here, only for these functions

rakhmets (approved):

brminich (approved):

src/infra/mem_section.h
```diff
@@ -71,7 +72,10 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
     verifySorted() const;
 
     nixlSectionDesc &
-    operator[](unsigned int index) override;
+    operator[](size_t index) {
+        assert(verifySorted());
+        return descs[index];
+    }
```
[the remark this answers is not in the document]
iyastreb: this is existing code, out of scope for this PR

src/api/cpp/nixl_descriptors.h
```diff
@@ -323,19 +323,20 @@ class nixlDescList {
...
-    virtual T &
-    operator[](unsigned int index);
+    operator[](size_t index) const {
+        return descs[index];
+    }
+
+    T &
+    operator[](size_t index) {
+        return descs[index];
+    }
```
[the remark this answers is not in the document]
iyastreb: we discussed it and decided to follow std::vector convention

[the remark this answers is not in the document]
ovidiusm: @coderabbitai You reviewed this code but you did not flag the ABI issue due to changing types. You should always flag ABI/API changes in src/api/* files from now on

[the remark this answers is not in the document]
ovidiusm: @coderabbitai The learning/skill should also say that if ABI changes are well justified, downstream projects such as Dynamo and TRT-LLM must be informed

PR #1324, opened 2026-02-13
NIXL: Zero-copy nixlDescList

iyastreb:
## What?
Enabled zero-copy for nixlDescList by allowing shallow copy.

## Why?
On large desc dimensions copying takes a lot of time. One of the copies is made inside Python binding, where we copy Python array into nixlDescList. This copy takes quite some time while holding the GIL lock. In certain cases we can eliminate this copy by making it shallow copy, when `initialize_xfer` is called sequentially after `get_xfer_descs`.
`initialize_xfer` makes its own copy (nixlMetaList), which can be also eliminated.
 
The proposed solution is to optionally enable zero-copy mode for nixlDescList. Appropriate Python binding modification will be done in a separate PR.

## How?
Added a common `view_` that points either to owned `descs` or to an external data. All the read accesses are done via this view, all the writes modify the view.

Python API is in separate PR: https://github.com/ai-dynamo/nixl/pull/1325

Perf testing:
```
before (us)     after (us)
calc: 1368233   1373060
desc: 453230    15355    (30x faster, copy eliminated)
prep: 3183194   1872914  (access is faster: inline non-virtual [])
post: 816547    781286   (access is faster: inline non-virtual [])
```

## Summary by CodeRabbit

* **New Features**
  * Added shallow-copy support for descriptor lists with view-based iteration.
  * Introduced explicit copy/move semantics and a new factory method for creating shallow copies.
  * Added iterator type aliases for improved API consistency.

* **Bug Fixes**
  * Added runtime protection to prevent modifications to shallow-copy descriptor lists.

* **API Changes**
  * Updated method signatures for parameter types and iteration interfaces across descriptor list operations.


src/api/cpp/nixl_descriptors.h
```diff
@@ -251,20 +253,63 @@ operator==(const nixlRemoteDesc &lhs, const nixlRemoteDesc &rhs);
...
+    const T *view_;
+    /** @var Size of the descriptor list */
+    size_t size_;
+
+    /**
+     * @brief Synchronize the view of the descriptor list with the data in the vector.
+     *        Can be called only in owner mode (not shallow copy), after vector is modified
+     */
+    inline void
+    syncView() {
```
ColinNV, line 282: The `inline` is superfluous, a `noexcept` could be added.

src/api/cpp/nixl_descriptors.h
```diff
@@ -280,15 +325,36 @@ class nixlDescList {
...
+
+    /**
+     * @brief Create a shallow copy of a nixlDescList.
+     *
+     * @param type         NIXL memory type of descriptor list
+     * @param view         View of the descriptor list
+     * @param size         Size of the descriptor list
+     * @return Shallow copy of the descriptor list
+     */
+    static const nixlDescList<T>
```
ColinNV, line 363: `const` does nothing when returning by value.

src/api/cpp/nixl_descriptors.h
```diff
@@ -251,20 +253,63 @@ operator==(const nixlRemoteDesc &lhs, const nixlRemoteDesc &rhs);
...
 template<class T>
 class nixlDescList {
-protected:
+private:
     /** @var NIXL memory type */
     nixl_mem_t type;
     /** @var Vector for storing nixlDescs */
     std::vector<T> descs;
+    /** @var Whether the descriptor list is a shallow copy */
+    bool isShallowCopy_;
```
ColinNV, line 271: I wonder whether "shallow copy" is the right term, how about something like non-owning view, or read-only view?

src/api/cpp/nixl_descriptors.h
```diff
@@ -308,46 +374,55 @@ class nixlDescList {
      */
     inline int
     descCount() const {
-        return descs.size();
+        return size_;
+    }
+
+    inline size_t
```
ColinNV: `inline` does nothing here, `noexcept` can be added.

src/api/cpp/nixl_descriptors.h
```diff
@@ -308,46 +374,55 @@ class nixlDescList {
...
+        return size_;
     }
 
     /**
      * @brief Check if nixlDescList is empty or not
      */
     inline bool
     isEmpty() const {
-        return (descs.size() == 0);
+        return (size_ == 0);
```
ColinNV, line 399: Brackets are superfluous (as is the `inline` while `noexcept` is missing).

src/api/cpp/nixl_descriptors.h
```diff
@@ -251,20 +253,63 @@ operator==(const nixlRemoteDesc &lhs, const nixlRemoteDesc &rhs);
...
+
+    /**
+     * @brief Private constructor to create a shallow copy of a nixlDescList.
+     *        Used internally by the makeShallowCopy to return only a const reference
+     *        to the shallow copy.
+     */
+    nixlDescList(const nixl_mem_t &type, const T *view, size_t size);
 
 public:
+    using iterator_t = typename std::vector<T>::iterator;
```
ColinNV, line 312: The inconsistency between the two iterator types looks a bit weird to me. ~~I'd use pointers for both cases and then remove the type aliases to make things clearer below.~~ I see the iterator is used for an `insert()` later.

src/api/cpp/nixl_descriptors.h
```diff
@@ -308,46 +374,55 @@ class nixlDescList {
...
 
-    /**
-     * @brief Operator [] overloading, get/set descriptor at [index].
-     *        Can throw std::out_of_range exception.
-     */
-    const T &
-    operator[](unsigned int index) const;
-    virtual T &
-    operator[](unsigned int index);
+    inline const T &
```
ColinNV, line 402: Same thing about `inline` and `noexcept` for this and the following functions.

src/api/cpp/nixl_descriptors.h
```diff
@@ -308,46 +374,55 @@ class nixlDescList {
...
+    operator[](size_t index) const {
+        return view_[index];
+    }
+
+    inline T &
+    operator[](size_t index) {
```
ColinNV, line 408: Maybe add an `at()` function to this class like `std::vector`'s that checks the index bounds?

src/api/cpp/nixl_descriptors.h
```diff
@@ -251,20 +253,63 @@ operator==(const nixlRemoteDesc &lhs, const nixlRemoteDesc &rhs);
...
-protected:
+private:
     /** @var NIXL memory type */
     nixl_mem_t type;
     /** @var Vector for storing nixlDescs */
     std::vector<T> descs;
+    /** @var Whether the descriptor list is a shallow copy */
+    bool isShallowCopy_;
+    /** @var Common view of the descriptor list, both for shallow copy and own data */
+    const T *view_;
```
ColinNV, line 273: Suggest naming this `data_` as a _view_ consists of data (a pointer) and size (a `size_t`).

src/infra/nixl_descriptors.cpp
```diff
@@ -149,18 +149,70 @@ void nixlBlobDesc::print(const std::string &suffix) const {
...
+nixlDescList<T>::swap(nixlDescList<T> &first, nixlDescList<T> &second) noexcept {
+    std::swap(first.type, second.type);
+    std::swap(first.descs, second.descs);
+    std::swap(first.isShallowCopy_, second.isShallowCopy_);
+    std::swap(first.view_, second.view_);
+    std::swap(first.size_, second.size_);
+}
+
+template<class T>
+nixlDescList<T>::nixlDescList(nixlDescList<T> &&d_list) noexcept
```
ColinNV, line 174: Do we call this, and the assignment below, on the hot path?

src/infra/nixl_descriptors.cpp
```diff
@@ -149,18 +149,70 @@ void nixlBlobDesc::print(const std::string &suffix) const {
...
+template<class T>
+nixlDescList<T>::nixlDescList(nixlSerDes *deserializer)
+    : isShallowCopy_(false),
+      view_(nullptr),
+      size_(0) {
     size_t n_desc;
     std::string str;
 
-    descs.clear();
+    clear();
```
ColinNV, line 197: If `view_` and `size_` are initialized as `syncView()` will then this call can be removed.

src/infra/nixl_descriptors.cpp
```diff
@@ -261,16 +314,16 @@ nixlDescList<nixlBasicDesc> nixlDescList<T>::trim() const {
 
 template <class T>
 int nixlDescList<T>::getIndex(const nixlBasicDesc &query) const {
-    auto itr = std::find(descs.begin(), descs.end(), query);
-    if (itr == descs.end()) return NIXL_ERR_NOT_FOUND; // not found
-    return itr - descs.begin();
+    auto itr = std::find(begin(), end(), query);
+    if (itr == end()) return NIXL_ERR_NOT_FOUND; // not found
```
ColinNV, line 300: The comment...

src/infra/nixl_descriptors.cpp
```diff
@@ -261,16 +314,16 @@ nixlDescList<nixlBasicDesc> nixlDescList<T>::trim() const {
 
 template <class T>
 int nixlDescList<T>::getIndex(const nixlBasicDesc &query) const {
-    auto itr = std::find(descs.begin(), descs.end(), query);
-    if (itr == descs.end()) return NIXL_ERR_NOT_FOUND; // not found
-    return itr - descs.begin();
+    auto itr = std::find(begin(), end(), query);
```
ColinNV, line 299: `const`

src/infra/nixl_descriptors.cpp
```diff
@@ -261,16 +314,16 @@ nixlDescList<nixlBasicDesc> nixlDescList<T>::trim() const {
...
+    if (itr == end()) return NIXL_ERR_NOT_FOUND; // not found
+    return itr - begin();
 }
 
 template <class T>
 nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
 
     nixl_status_t ret;
-    size_t n_desc = descs.size();
+    size_t n_desc = size_;
```
ColinNV, line 308: Use `size_` instead of `n_desc` in this function?

src/infra/nixl_descriptors.cpp
```diff
@@ -384,61 +437,62 @@ operator== <nixlRemoteMetaDesc>(const nixlDescList<nixlRemoteMetaDesc> &lhs,
...
 }
 
 bool
 nixlSecDescList::verifySorted() const {
-    const auto &vec = this->descs;
-    int size = (int)vec.size();
-    if (size <= 1) return (size == 1);
-    for (int i = 0; i < size - 1; ++i) {
-        if (vec[i + 1] < vec[i]) return false;
+    size_t count = size();
```
ColinNV, line 429: `const`

src/infra/nixl_descriptors.cpp
```diff
@@ -384,61 +437,62 @@ operator== <nixlRemoteMetaDesc>(const nixlDescList<nixlRemoteMetaDesc> &lhs,
...
+    if (count <= 1) {
+        return (count == 1);
```
ColinNV, line 431: Brackets are superfluous, and why is an empty list considered not sorted?

src/infra/nixl_descriptors.cpp
```diff
@@ -384,61 +437,62 @@ operator== <nixlRemoteMetaDesc>(const nixlDescList<nixlRemoteMetaDesc> &lhs,
...
+    }
+    for (size_t i = 0; i < count - 1; ++i) {
```
ColinNV, line 433: Could probably be implemented with `std::is_sorted()` or `std::is_sorted_until()`.

src/infra/nixl_descriptors.cpp
```diff
@@ -384,61 +437,62 @@ operator== <nixlRemoteMetaDesc>(const nixlDescList<nixlRemoteMetaDesc> &lhs,
...
+    return nixlDescList::operator[](index);
 }
 
 int
 nixlSecDescList::getIndex(const nixlBasicDesc &query) const {
-    auto itr = std::lower_bound(this->descs.begin(), this->descs.end(), query);
-    if (itr == this->descs.end()) return NIXL_ERR_NOT_FOUND;
-    if (static_cast<const nixlBasicDesc &>(*itr) == query)
-        return static_cast<int>(itr - this->descs.begin());
+    auto itr = std::lower_bound(begin(), end(), query);
```
ColinNV, line 443: `const`

src/infra/nixl_descriptors.cpp
```diff
@@ -384,61 +437,62 @@ operator== <nixlRemoteMetaDesc>(const nixlDescList<nixlRemoteMetaDesc> &lhs,
...
+    if (static_cast<const nixlBasicDesc &>(*itr) == query) return static_cast<int>(itr - begin());
     return NIXL_ERR_NOT_FOUND;
 }
 
 int
 nixlSecDescList::getCoveringIndex(const nixlBasicDesc &query) const {
-    auto itr = std::lower_bound(this->descs.begin(), this->descs.end(), query);
-    if (itr != this->descs.end() && itr->covers(query))
-        return static_cast<int>(itr - this->descs.begin());
+    auto itr = std::lower_bound(begin(), end(), query);
```
ColinNV, line 451: `const`

ColinNV (commented):
And as discussed I believe it would be cleaner and less error-prone to keep the two cases (owning, non-owning) somehow separate, however that's now possible at this point.

src/infra/nixl_descriptors.cpp
```diff
@@ -204,40 +256,41 @@ nixlDescList<T>::nixlDescList(nixlSerDes* deserializer) {
...
 }
 
-template <class T>
-void nixlDescList<T>::remDesc (const int &index){
-    if (((size_t) index >= descs.size()) || (index < 0))
+template<class T>
+void
+nixlDescList<T>::remDesc(size_t index) {
+    checkModifiable();
+    if ((index >= descs.size()) || (index < 0)) {
```
guy-ealey-morag, line 263:
```suggestion
    if (index >= descs.size()) {
```

`index < 0` is always `false` for `size_t`

src/api/cpp/nixl_descriptors.h
guy-ealey-morag: Could we also use `std::variant` to hold either the "view" type (data + size) or a `std::vector<T>`?
It might not be as simple but it would avoid any bugs related to using `descs` while `_isShallowCopy` is `true`

src/infra/nixl_descriptors.cpp
guy-ealey-morag, line 287: Iterating over descs will not work here when it's a shallow copy

src/infra/nixl_descriptors.cpp
```diff
@@ -300,9 +353,9 @@ nixl_status_t nixlDescList<T>::serialize(nixlSerDes* serializer) const {
...
     if (std::is_same<nixlBasicDesc, T>::value) {
-        ret = serializer->addStr("", std::string(
-                                 reinterpret_cast<const char*>(descs.data()),
-                                 n_desc * sizeof(nixlBasicDesc)));
+        ret = serializer->addStr(
+            "",
+            std::string(reinterpret_cast<const char *>(begin()), n_desc * sizeof(nixlBasicDesc)));
         if (ret) return ret;
     } else { // already checked it can be only nixlBlobDesc or nixlSectionDesc
         for (auto & elm : descs) {
```
guy-ealey-morag, line 343: Iterating over descs will not work here when it's a shallow copy

src/infra/nixl_descriptors.cpp
guy-ealey-morag, line 392: Accesing descs directly will not work here when it's a shallow copy.

iyastreb:
It's not relevant for 2 step API, which is now being used

PR #1325, opened 2026-02-13
NIXL: Zero-copy nixlDescList Python API

iyastreb:
## What?
Zero-copy Python API for the corresponding change in https://github.com/ai-dynamo/nixl/pull/1324

## Summary by CodeRabbit

* **New Features**
  * Added optional zero-copy mode for transfer descriptor creation. Users can now enable this feature via a new parameter (defaults to False) when working with transfer descriptors, allowing alternative creation paths where applicable while maintaining full backward compatibility.


src/api/python/_api.py
```diff
@@ -913,13 +913,15 @@ def check_remote_metadata(
...
     @param mem_type Optional memory type necessary for (a).
+    @param zero_copy Optional boolean flag to use zero copy for the transfer descriptor list.
     @return Transfer descriptor list, nixlXferDList.
     """
 
     def get_xfer_descs(
         self,
         descs,
         mem_type: Optional[str] = None,
+        zero_copy: bool = False,
```
roiedanino, line 926:
```suggestion
        zero_copy: Optional[bool] = False,
```

src/bindings/python/nixl_bindings.cpp
```diff
@@ -90,6 +90,15 @@ class nixlNoTelemetryError : public std::runtime_error {
     nixlNoTelemetryError(const char *what) : runtime_error(what) {}
 };
 
+class nixlXferDListWrapper : public nixl_xfer_dlist_t {
+    py::object ref_;
+
+public:
+    nixlXferDListWrapper(nixl_xfer_dlist_t &&base, py::object ref)
+        : nixl_xfer_dlist_t(std::move(base)),
+          ref_(ref) {}
```
roiedanino, line 99: Maybe I'm missing something, but why do we need the ref here? It seems you are always passing py::none(), if it's necessary for the python side, maybe we can do something like this:


```suggestion
    nixlXferDListWrapper(nixl_xfer_dlist_t &&base, py::object ref = py:none())
        : nixl_xfer_dlist_t(std::move(base)),
          ref_(ref) {}
```

iyastreb: This is needed to keep a reference to the original python array, to avoid its garbage collection.
We pass it when making a zero-copy:
```
                     auto new_list = nixl_xfer_dlist_t::makeShallowCopy(
                         mem, reinterpret_cast<const nixlBasicDesc *>(descs.data()), n);
                     return nixlXferDListWrapper(std::move(new_list), descs);

iyastreb:
It's not relevant for 2 step API, which is now being used

PR #1327, opened 2026-02-13
Dell ObjectScale S3 over RDMA engine support for OBJ plugin

jgoldsch12:
## What?
Introduces the Dell ObjectScale S3 over RDMA engine support for the OBJ Plugin.

## Why?
Enables the use of VRAM_SEG and GPU Direct for objects when utilizing a Dell ObjectScale as a backend

## Summary by CodeRabbit

* **New Features**
  * Dell ObjectScale accelerated backend with RDMA-based async put/get, VRAM support, and engine selection; engine constructors now accept injected S3 clients.

* **Bug Fixes / Behavior**
  * Standard S3 clients explicitly report RDMA operations as unsupported.
  * Improved client initialization/fallback behavior for mixed client configurations.

* **Documentation**
  * Added comprehensive Dell ObjectScale README with configuration and examples.

* **Tests**
  * Unit tests extended for Dell backend, VRAM paths, and RDMA flows.

### DIFF
--- a/test/gtest/hw_warning_test.cpp
+++ b/test/gtest/hw_warning_test.cpp
@@ -183,6 +183,12 @@ TEST_F(HardwareWarningTest, EfaHardwareMismatchNoWarning) {
 
         nixlAgent agent("EfaTestAgent", nixlAgentConfig(true));
 
+        const gtest::LogIgnoreGuard lig_no_efa(
+            "Could not find EFA devices, rail selection for DRAM memory type aborted");
+        const gtest::LogIgnoreGuard lig_rail_fallback(
+            "Using default \\(all\\) rail selection policy for DRAM memory type due to "
+            "previous errors");
+
         for (const auto &name : backends) {
             nixlBackendH *backend;
             EXPECT_EQ(agent.createBackend(name, {}, backend), NIXL_SUCCESS);


jgoldsch12:
@aranadive, I had to create a new PR for this change.

jgoldsch12:
@coderabbitai

jgoldsch12:
@coderabbitai review

src/plugins/obj/s3_accel/dell/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                       std::string_view rdma_desc,
+                                       put_object_callback_t callback) {
+    NIXL_DEBUG << absl::StrFormat(
+        "putObjectRdmaAsync: key=%s, data_ptr=%p, data_len=%zu, offset=%zu, rdma_desc=%s",
+        std::string(key).c_str(),
+        reinterpret_cast<void *>(data_ptr),
+        data_len,
+        offset,
+        rdma_desc.empty() ? "<empty>" : std::string(rdma_desc).c_str());
+
```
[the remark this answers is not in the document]
jgoldsch12: This is a fine thing to log.

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,552 @@
...
+                                          req.offset,
+                                          req.rdma_desc,
+                                          [status_promise](bool success) {
+                                              status_promise->set_value(success ? NIXL_SUCCESS :
+                                                                                  NIXL_ERR_BACKEND);
+                                          });
+        }
+    }
+
+    return NIXL_IN_PROG;
```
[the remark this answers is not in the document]
jgoldsch12: upper layer code in the nixl_agent already validates that the handle must not be NULL, otherwise an error is returned before this code is called.

src/plugins/obj/s3_accel/dell/README.md
```diff
@@ -0,0 +1,195 @@
...
+    {"access_key", "test"},
+    {"secret_key", "secret"},
+    {"bucket", "test-bucket"},
+    {"endpoint_override", "http://10.10.10.10:9000"},
+    {"scheme", "http"},
+    {"use_virtual_addressing", "false"},
+    {"req_checksum", "required"},
+    {"ca_bundle", "/root/ca-certs/cacert.pem"},
+    {"accelerated", "true"},
+    {"type", "dell"}
```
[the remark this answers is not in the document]
jgoldsch12: test and secret are just examples, but not actual credentials.

src/plugins/obj/s3_accel/dell/client.cpp
```diff
@@ -0,0 +1,131 @@
...
+        reinterpret_cast<void *>(data_ptr),
+        data_len,
+        offset,
+        rdma_desc.empty() ? "<empty>" : rdma_desc.c_str());
+
+    if (offset != 0) {
+        NIXL_ERROR << "putObjectRdmaAsync: offset is not 0, returning failure";
+        callback(false);
+        return;
+    }
```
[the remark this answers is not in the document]
jgoldsch12: 0 is a valid data length.  The result is creating an empty object.  That doesn't seem very useful, but it is allowed.

src/plugins/obj/obj_backend.h
```diff
@@ -79,6 +79,40 @@ class iS3Client {
...
+     * @param rdma_desc RDMA descriptor for the transfer
+     * @param callback Callback function to handle the result
+     */
+    virtual void
+    getObjectRdmaAsync(std::string_view key,
+                       uintptr_t data_ptr,
+                       size_t data_len,
+                       size_t offset,
+                       const std::string &rdma_desc,
+                       get_object_callback_t callback) = 0;
```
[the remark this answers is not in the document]
jgoldsch12: I believe this was fixed.

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+S3DellObsObjEngineImpl::deregisterMem(nixlBackendMD *meta) {
+    nixlObsObjMetadata *md = static_cast<nixlObsObjMetadata *>(meta);
+    if (md) {
+        if (md->nixlMem == OBJ_SEG) {
+            std::unique_ptr<nixlObsObjMetadata> obj_md_ptr =
+                std::unique_ptr<nixlObsObjMetadata>(md);
+            devIdToObjKey_.erase(obj_md_ptr->devId);
+        } else if ((md->nixlMem == DRAM_SEG) || (md->nixlMem == VRAM_SEG)) {
+            std::unique_ptr<nixlObsObjMetadata> mem_md_ptr =
+                std::unique_ptr<nixlObsObjMetadata>(md);
```
vvenkates27, line 403: For DRAM/VRAM we take ownership of `md` with `unique_ptr` (393–394) before calling `cuMemObjPutDescriptor`. On failure we return but the destructor still deletes `md`; the caller still holds that pointer → use-after-free / double-free. 
Maybe Call `cuMemObjPutDescriptor(md->localAddr)` first; only on success create the `unique_ptr` and adopt `md`?

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+/**
+ * Release a transfer request handle.
+ * Cleans up resources associated with the transfer request.
+ *
+ * @param handle Transfer request handle to release
+ * @return NIXL_SUCCESS on success
+ */
+nixl_status_t
+S3DellObsObjEngineImpl::releaseReqH(nixlBackendReqH *handle) const {
+    nixlObsObjBackendReqH *req_h = static_cast<nixlObsObjBackendReqH *>(handle);
```
vvenkates27, line 590: Add `if (handle == nullptr) return NIXL_ERR_INVALID_PARAM;` ?

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+ * @return NIXL_IN_PROG if operation started successfully
+ */
+nixl_status_t
+S3DellObsObjEngineImpl::postXfer(const nixl_xfer_op_t &operation,
+                                 const nixl_meta_dlist_t &local,
+                                 const nixl_meta_dlist_t &remote,
+                                 const std::string &remote_agent,
+                                 nixlBackendReqH *&handle,
+                                 const nixl_opt_b_args_t *opt_args) const {
+    nixlObsObjBackendReqH *req_h = static_cast<nixlObsObjBackendReqH *>(handle);
```
vvenkates27, line 516: if (handle == nullptr) return ?

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+ * Check the status of an ongoing transfer operation.
+ * Polls the futures associated with the transfer request to determine
+ * completion status.
+ *
+ * @param handle Transfer request handle to check
+ * @return NIXL_SUCCESS if completed, NIXL_IN_PROG if ongoing, error code on failure
+ */
+nixl_status_t
+S3DellObsObjEngineImpl::checkXfer(nixlBackendReqH *handle) const {
+    nixlObsObjBackendReqH *req_h = static_cast<nixlObsObjBackendReqH *>(handle);
```
vvenkates27, line 573: Add if (handle == nullptr) return?

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+          char *buf,
+          size_t size,
+          loff_t offset,
+          const cufileRDMAInfo_t *infop) {
+    if (infop == nullptr || infop->desc_str == nullptr) {
+        NIXL_ERROR << "objectGet: infop or infop->desc_str is null";
+        return -EINVAL;
+    }
+
+    void *ctx = cuObjClient::getCtx(handle);
```
vvenkates27, line 239: Check `getCtx(handle)` for null before using `ctx`; if null return `-EINVAL`

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+          const char *buf,
+          size_t size,
+          loff_t offset,
+          const cufileRDMAInfo_t *infop) {
+    if (infop == nullptr || infop->desc_str == nullptr) {
+        NIXL_ERROR << "objectPut: infop or infop->desc_str is null";
+        return -EINVAL;
+    }
+
+    void *ctx = cuObjClient::getCtx(handle);
```
vvenkates27, line 273: Same as objectGet: check `getCtx(handle)` for null before using `ctx`; if null return `-EINVAL`.

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,569 @@
...
+                                    nixlBackendMD *&out) {
+    if (!cuClient_->isConnected()) {
+        NIXL_ERROR << "CUObjClient is not connected.";
+        return NIXL_ERR_BACKEND;
+    }
+
+    auto supported_mems = {OBJ_SEG, DRAM_SEG, VRAM_SEG};
+    if (std::find(supported_mems.begin(), supported_mems.end(), nixl_mem) == supported_mems.end())
+        return NIXL_ERR_NOT_SUPPORTED;
+
```
vvenkates27, line 357: Reject duplicate devId. If `devIdToObjKey_.find(mem.devId) != devIdToObjKey_.end()` return `NIXL_ERR_INVALID_PARAM`.?

src/plugins/obj/s3_accel/dell/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    Aws::S3::Model::GetObjectRequest request;
+    request.WithBucket(bucketName_)
+        .WithKey(Aws::String(key))
+        .WithRange(absl::StrFormat("bytes=%zu-%zu", offset, offset + data_len - 1));
+    request.SetAdditionalCustomHeaderValue("x-rdma-info", std::string(rdma_desc));
+
```
vvenkates27, line 152: Guard against `offset + data_len` overflowing (e.g. `if ((offset + data_len) < offset)` then `callback(false)` and return) before building the range.?

src/plugins/obj/s3_accel/dell/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+ * @param callback Callback function invoked on completion with success/failure status
+ */
+void
+awsS3DellObsClient::getObjectRdmaAsync(std::string_view key,
+                                       uintptr_t data_ptr,
+                                       size_t data_len,
+                                       size_t offset,
+                                       std::string_view rdma_desc,
+                                       get_object_callback_t callback) {
+    if (data_len == 0) {
```
vvenkates27: Good call checking `data_len == 0` before building the range—avoids underflow in `offset + data_len - 1`.

jgoldsch12: I'm going back and forth on this one.
Doing a PutObject when a zero length buffer will create an empty object.  This is valid, maybe not valuable, but should we exclude it?
Doing a GetObject when the data range is zero or the data buffer is zero will return the object metadata.  Again, this is valid, but not useful when we have the queryMem API.  I agree that we do not want to underflow the data range.  The other S3 clients (vanilla, CRT, and Accel) all have the same issue and are not guarding against a zero length buffer.

I think we have two options:

1. Fail data_len == 0 for both Put and Get across all clients
2. Allow them to succeed and ensure we do not underflow the data range and only request zero bytes for GetObject

What are your thoughts @vvenkates27 ?

src/plugins/obj/s3_accel/dell/engine_impl.cpp
jgoldsch12: I think this is addressed in the latest code.  I added a mem_md_ptr.release() to transfer ownership back to the caller and allow them to retry.

src/plugins/obj/s3_accel/dell/engine_impl.cpp
jgoldsch12: I'm convinced this is worth adding.  I'll add it to post and checkXfer also

src/plugins/obj/s3_accel/dell/engine_impl.cpp
jgoldsch12: I'll handle with NIXL_ERR_INVALID_PARAM.  It seems the upper layer will fail the same way if a null handle is passed.  release and checkXfer do not seem to do this at the nixl_agent layer

src/plugins/obj/s3_accel/dell/engine_impl.cpp
jgoldsch12: same

src/plugins/obj/s3_accel/dell/engine_impl.cpp
jgoldsch12: I don't think this is an error.  If it is, the vanilla engine has the same issue.  Are we concerned about a possible memory leak?

src/plugins/obj/s3_accel/dell/README.md
```diff
@@ -0,0 +1,195 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aranadive, line 2: Can we keep only Dell S3 OBS related info in this README? Seems like a lot of this is a repeat from the generic OBJ plugin README. Maybe we can say refer to the main obj plugin README for the AWS S3 configuration.

jgoldsch12: I guess I was trying to keep the user from jumping around between guides and find the shortest path to getting this configured.  Is the concern about keeping multiple documents up to date?

src/plugins/obj/s3_accel/dell/client.cpp
jgoldsch12: This is fine, but please see my comment about the GetObject path and zero length objects.  Once we have a consensus on how this should be handled, I'll guard against the various underflowing issues.

src/plugins/obj/s3_accel/dell/README.md
aranadive: Yes potentially, we would have to keep all those in sync. Also, since this is based on the Object plugin it makes sense to just have dell-specific parts here. The other option, is to combine the dell configs into the main README as well.

src/plugins/obj/s3/client.h
```diff
@@ -63,6 +63,23 @@ class awsS3Client : public iS3Client {
     bool
     checkObjectExists(std::string_view key) override;
 
+    // RDMA methods - not supported by standard S3 client
```
aranadive: My intention with the s3 reorg effort was that we wouldnt need to add these to the main s3 client and only in the vendor-specific client. So, when the s3 client calls putObjectAsync, it would invoke the vendor-specific client's putObjectAsync which would all their own RDMA transfer functions. Is that something you looked into already?

jgoldsch12: I added these to enable unit testing with the mock client.  If there is a different way to achieve this, I'm happy to do it.

src/plugins/obj/s3_accel/dell/client.cpp
vvenkates27: My vote: Option 1 – Fail data_len == 0 for both Put and Get.
Zero-length transfers in a transfer library are almost always caller bugs, not intentional operations. Failing fast with a clear error is better than silently succeeding with no work done.
@aranadive  do you agree?

jgoldsch12: I've made this change in the latest code

src/plugins/obj/s3/client.h
aranadive: I see. For the mock client we can create a child RDMA client that can implement these functions rather than the base mock client. That should allow you to test these functions separately too.

src/plugins/obj/s3_accel/dell/README.md
jgoldsch12: Change has been made to reduce the duplicative info in the README

src/plugins/obj/s3/client.h
jgoldsch12: I made this change in the code.  Please take a look at the approach I took.

jgoldsch12: It can be found in this specific commit:

https://github.com/ai-dynamo/nixl/pull/1327/commits/b149a70094b491a3886cab02578d8136544ec9c5

src/plugins/obj/s3_accel/dell/engine_impl.h
```diff
@@ -0,0 +1,63 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef OBJ_PLUGIN_S3_DELL_ENGINE_IMPL_H
+#define OBJ_PLUGIN_S3_DELL_ENGINE_IMPL_H
```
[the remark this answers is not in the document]
jgoldsch12: fixed

src/plugins/obj/s3_accel/dell/engine_impl.cpp
```diff
@@ -0,0 +1,340 @@
...
+    : S3AccelObjEngineImpl(init_params) {
+    s3Client_ = std::make_shared<awsS3DellObsClient>(init_params->customParams, executor_);
+    NIXL_INFO << "Object storage backend initialized with S3 Dell ObjectScale client";
+
+    cuClient_ = std::make_shared<cuObjClient>(obs_ops, CUOBJ_PROTO_RDMA_DC_V1);
+    if (!cuClient_->isConnected()) {
+        NIXL_ERROR << "CUObjClient failed to connect.";
+        return;
+    }
+}
```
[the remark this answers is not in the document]
jgoldsch12: I tried this and it didn't work, so I abandoned this idea.

src/plugins/obj/s3_accel/dell/README.md
```diff
@@ -0,0 +1,163 @@
...
+## Dependencies
+
+This engine requires aws-sdk-cpp.  This dependency and steps to install it are provided in the OBJ backend documentation.
+
+This engine requires the CUDA Toolkit version 13.1.1 or later to be installed.
+
+[CUDA GDS Install and Setup](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html)
+
+The CUDA toolkit provides the CUObject Client library, which is required for this engine to function.  The library is included in the CUDA Toolkit installation.
+```
```
aranadive: Are the quotes required here?

src/plugins/obj/s3_accel/dell/rdma_interface.h
```diff
@@ -0,0 +1,57 @@
...
+#include <string_view>
+#include <cstdint>
+#include <functional>
+#include "nixl_types.h"
+
+/**
+ * Interface for S3 clients that support RDMA operations.
+ * Only Dell-specific clients implement this interface.
+ */
+class iS3RdmaClient {
```
aranadive: This should be iDellS3RdmaClient to indicate its specific for Dell vendor.

src/plugins/obj/s3/client.h
aranadive: I think the approach is good and the pattern should be reusable. Thanks for working on this.

aranadive (approved):
Apart from the few things pointed, generally looks good to me.

src/plugins/obj/s3_accel/dell/README.md
jgoldsch12: done

src/plugins/obj/s3_accel/dell/rdma_interface.h
jgoldsch12: done

jgoldsch12:
@vvenkates27 do you have any final comments?

PR #1334, opened 2026-02-16
Skip hardware warning tests in nixl-ci-non-gpu flow

guy-ealey-morag:
## What?
Skip hardware warning tests in nixl-ci-non-gpu flow

## Why?
The nixl-ci-non-gpu flow is emulating an environment without GPUs but due to the way containers work, the hardware of the host is still visible under `/sys/bus/pci/devices`.
This causes the tests to fail because the HW is detected but the relevant transports are not available.

### DIFF
--- a/.ci/jenkins/pipeline/Jenkinsfile
+++ b/.ci/jenkins/pipeline/Jenkinsfile
@@ -14,7 +14,7 @@
 // This library provides functionality for interacting with GitHub
 // It is used to update the commit status and upload logs
 @Library('blossom-github-lib@master')
-@Library('github.com/Mellanox/ci-demo@master')  // External library for matrix build functionality
+@Library('github.com/Mellanox/ci-demo@stable_nixl')  // External library for matrix build functionality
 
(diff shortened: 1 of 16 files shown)


ColinNV:
Can we use a more generic name instead of `NIXL_SKIP_HW_WARNING_TESTS`, one that is more descriptive of the circumstances like `NIXL_FAKE_GPUS_PRESENT` but a bit better. Reason: I'll need the same environment variable to ignore some warnings in other gtests. #1288

guy-ealey-morag:
> Can we use a more generic name instead of `NIXL_SKIP_HW_WARNING_TESTS`, one that is more descriptive of the circumstances like `NIXL_FAKE_GPUS_PRESENT` but a bit better. Reason: I'll need the same environment variable to ignore some warnings in other gtests. #1288

Renamed to `NIXL_CI_NON_GPU`, and also simplified the parsing to checking existence only.

ColinNV (approved):

PR #1335, opened 2026-02-16
libfabric: Remove delay in between post operation retries

yexiang-aws:
## What?
This PR removes delay in libfabric post operation retry logic for send, write, and read operations.

## Why?
The previous implementation used usleep() with exponential backoff (1ms base delay, 
capping at 100ms) when retrying operations that returned `-FI_EAGAIN`. This approach 
introduced unnecessary latency in the retry path. Since the completion queue is 
already being progressed before each retry to drain pending completions, actively 
sleeping adds no benefit and only delays recovery from transient resource exhaustion.

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -1057,17 +1057,12 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
                            << ", retrying (attempt " << attempt << ")";
             }
 
-            // Exponential backoff with cap to avoid overwhelming the system
-            int delay_us = std::min(NIXL_LIBFABRIC_BASE_RETRY_DELAY_US * (1 + attempt / 10),
-                                    NIXL_LIBFABRIC_MAX_RETRY_DELAY_US);
-
             // Progress completion queue to drain pending completions before retry
             nixl_status_t progress_status = progressCompletionQueue();
             if (progress_status == NIXL_SUCCESS) {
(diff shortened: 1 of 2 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1136,17 +1133,13 @@ nixlLibfabricRail::postWrite(const void *local_buffer,
...
-                                    NIXL_LIBFABRIC_MAX_RETRY_DELAY_US);
-
             // Progress completion queue to drain pending completions before retry
             nixl_status_t progress_status = progressCompletionQueue();
             if (progress_status == NIXL_SUCCESS) {
                 NIXL_TRACE << "Progressed completions on rail " << rail_id << " before retry";
             }
 
-            usleep(delay_us);
+            std::this_thread::yield();
```
akkart-aws: Is there a reason why we need the thread::yield here ? 
Can we not simply retry without yielding the thread ?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1213,17 +1206,13 @@ nixlLibfabricRail::postRead(void *local_buffer,
...
```
akkart-aws: Same as above

src/utils/libfabric/libfabric_rail.cpp
akkart-aws: NM. You put in the commit message that you wanted to time-slice between other threads.

fengjica (approved):

PR #1338, opened 2026-02-17
Optimization of createXferReq perf

iyastreb:
## What?
On large batches createXferReq takes quite a lot of time, mostly copying (70% of the time).
But apart from the copy it does validation and population, which take around 30% of the time.

## Why?
On large batches createXferReq takes quite a lot of time, mostly copying (70% of the time).
But apart from the copy it does validation and population, which take around 30% of the time.
This PR improves performance of the createXferReq by 13.5%.

## How?
inlining hot path functions and adding branch prediction

Perf testing:
```
prep step: 6280823 us => 5435771 us (13.5%)
```

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -87,8 +87,15 @@ int connectToIP(std::string ip_addr, int port) {
     // Check if connection was successful
     int error = 0;
     socklen_t len = sizeof(error);
-    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0 || error != 0) {
+    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0) {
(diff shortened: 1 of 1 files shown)


guy-ealey-morag (approved):

src/core/nixl_agent.cpp
```diff
@@ -853,13 +853,19 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
+    nixl_status_t ret1, ret2;
     nixl_opt_b_args_t opt_args;
-
-    std::unique_ptr<backend_set_t> backend_set = std::make_unique<backend_set_t>();
+    backend_set_t backend_set;
 
     req_hndl = nullptr;
 
+    // Check the correspondence between descriptor lists
+    if (local_descs.descCount() != remote_descs.descCount()) {
```
brminich, line 863: can we also add `__builtin_expect` for error flows making them slow path?

PR #1344, opened 2026-02-19
nixlbench: print P100 TX duration to detect single outliers

yehuday:
## What?
Add a TX P100 column to result table

## Why?
It often occurs that P99 is lower than Average due to a single large outlier. Add P100 to clarify these cases.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option to show extended statistics in benchmark reports. When enabled, reports include an extra P100 transfer-duration column alongside existing P99 metrics. The option is off by default and can be toggled via configuration file or command-line.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -71,6 +71,7 @@ NB_ARG_STRING(scheme, XFERBENCH_SCHEME_PAIRWISE, "Scheme: pairwise, manytoone, o
 NB_ARG_STRING(mode, XFERBENCH_MODE_SG, "MODE: SG (Single GPU per proc), MG (Multi GPU per proc)");
 NB_ARG_STRING(op_type, XFERBENCH_OP_WRITE, "Op type: READ, WRITE");
 NB_ARG_BOOL(check_consistency, false, "Enable Consistency Check");
+NB_ARG_BOOL(show_extended_stats, false, "Show extended statistics (P100)");
 NB_ARG_UINT64(total_buffer_size,
               8LL * 1024 * (1 << 20),
               "Total buffer size across device for each process");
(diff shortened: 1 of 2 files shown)


brminich:
impo p99 is enough, because output already containe quite many columns
@ovidiusm, @aranadive wdyt?

aranadive:
> impo p99 is enough, because output already containe quite many columns @ovidiusm, @aranadive wdyt?

We can track it but only print it if some option (show_enhanced_stats) is enabled?

brminich:
> > impo p99 is enough, because output already containe quite many columns @ovidiusm, @aranadive wdyt?
> 
> We can track it but only print it if some option (show_enhanced_stats) is enabled?

Ok, if you see a reason for that, but I'd avoid it as much as possible by default

ovidiusm:
We are filtering out outliers on purpose, I would also prefer to avoid it due to having so wide output already. If it is mainly due to confusion, maybe it would be better to adjust the documentation

yehuday:
anyone objects to adding it with --show-extended-stats ?

aranadive:
> anyone objects to adding it with --show-extended-stats ?

I don't but this shouldnt be enabled by default.

ovidiusm:
I would prefer not to add it, it complicates the output and not clear value

yehuday:
added --show-extended-stats which is off by default.
shouldn't harm the default run

ovidiusm (changes requested):
Please explain why showing this value is useful

yehuday:
@ovidiusm i mentioned it in the commit message 

It often occurs that P99 is lower than Average due to a single large outlier.
Add P100 to clarify these cases, but only show it when --show-extended-stats
flag is passed to avoid cluttering the default output.

brminich:
> @ovidiusm i mentioned it in the commit message
> 
> It often occurs that P99 is lower than Average due to a single large outlier. Add P100 to clarify these cases, but only show it when --show-extended-stats flag is passed to avoid cluttering the default output.

but it can be guessed from the fact it happened. Why is there a need to print this outlier? I do not think we need this change, unless @aranadive or @ovidiusm have strong opinion on that

brminich:
@yehuday, @ovidiusm, @aranadive let's close it?

PR #1345, opened 2026-02-20
CORE/LISTENER: Fix wrong error reporting.

ColinNV:
## What?
Fix wrong error reporting after `getsockopt()`.

## How?
Make sure `strerror()` is used for the correct `errno`-style value.

### DIFF
--- a/src/core/nixl_listener.cpp
+++ b/src/core/nixl_listener.cpp
@@ -87,8 +87,15 @@ int connectToIP(std::string ip_addr, int port) {
     // Check if connection was successful
     int error = 0;
     socklen_t len = sizeof(error);
-    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0 || error != 0) {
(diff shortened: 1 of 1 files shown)


src/core/nixl_listener.cpp
```diff
@@ -87,12 +87,19 @@ int connectToIP(std::string ip_addr, int port) {
     // Check if connection was successful
     int error = 0;
     socklen_t len = sizeof(error);
-    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0 || error != 0) {
+    if (getsockopt(ret_fd, SOL_SOCKET, SO_ERROR, &error, &len) < 0) {
```
brminich, line 90: can we leave the prev version, but add this line in the `if` block?
`if (error != 0) errno = error; // Sync errno if the error was inside the socket`

ColinNV: I separated it so that the wording in the error message makes it possible to distinguish whether `errno` or `error` was set, they represent errors on different levels...

brminich: ok

PR #1346, opened 2026-02-20
CONFIG: Part 2 (use).

ColinNV:
## What?
Consistently use the new config API from #1301.

## Why?
Unify access to environment and prepare for config files.

## How?
Replace all previous calls to `std::getenv()` with the new API functions. Remember that conversion failures (a value exists in the environment (and later configuration file) but can't be converted to its target type) intentionally do **not** fall back to any default behaviour in order to not ((relativley) silently) hide mistakes in the setup.

## Summary by CodeRabbit

* **Refactor**
  * Centralized configuration access across core services and plugins; several runtime identifiers made immutable at construction.

* **Configuration Updates**
  * New helpers for paths, durations, non-empty strings, and existence checks.
  * Many runtime settings (telemetry, exporter selection, etcd endpoints/namespace, plugin dir, storage bucket, RC mode, timeouts, Prometheus options) now read from centralized config.

* **Behavioral**
  * Telemetry, exporter loading, IP selection, and startup behavior follow config-driven logic.

* **Tests**
  * Minor telemetry test adjustment and cleanup fix.


src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -74,8 +75,7 @@ nixlMooncakeEngine::nixlMooncakeEngine(const nixlBackendInitParams *init_params)
     auto ips = findLocalIpAddresses();
     std::string segment_name = "127.0.0.1";
     if (!ips.empty()) segment_name = ips[0];
-    if (getenv("NIXL_MOONCAKE_IP_ADDR"))
-        segment_name = std::string(getenv("NIXL_MOONCAKE_IP_ADDR"));
+    (void)nixl::config::getValueWithStatus(segment_name, "NIXL_MOONCAKE_IP_ADDR");
```
rakhmets: Maybe it would be better to use `getValueDefaulted` with default value `127.0.0.1`, or `ips[0]` if `ips` is not empty.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -81,8 +52,8 @@ getHostname() {
 nixlTelemetryPrometheusExporter::nixlTelemetryPrometheusExporter(
     const nixlTelemetryExporterInitParams &init_params)
     : nixlTelemetryExporter(init_params),
-      local_(getLocal()),
-      port_(getPort()),
+      local_(nixl::config::getValueDefaulted(prometheusLocalVar, false)),
+      port_(nixl::config::getValueDefaulted(prometheusPortVar, prometheusExporterDefaultPort)),
```
rakhmets, line 56: It seems check for the value is missing here. Port should be in range the range `1..65535`

ColinNV: Not necessary, the default port is of type `uint16_t`, which is inferred for the return type and also used in the string-to-integer conversion. It should throw on overflow (and `test/gtest/configuration.cpp` verifies that it does). That leaves `0`, but I don't think that's worth checking for, in a way it's a valid port number.

rakhmets: Got it.

src/core/nixl_agent.cpp
```diff
@@ -125,24 +126,19 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
...
 
     memorySection = new nixlLocalSection();
-    const char *telemetry_env_val = std::getenv(TELEMETRY_ENABLED_VAR);
 
-    if (telemetry_env_val != nullptr) {
-        if (!strcasecmp(telemetry_env_val, "y") || !strcasecmp(telemetry_env_val, "1") ||
-            !strcasecmp(telemetry_env_val, "yes") || !strcasecmp(telemetry_env_val, "on")) {
+    const auto tmp = nixl::config::getValueOptional<bool>(TELEMETRY_ENABLED_VAR);
+
+    if (tmp) {
```
guy-ealey-morag: Why was `telemetry_env_val` renamed to `tmp`?

src/core/telemetry/telemetry.cpp
```diff
@@ -66,49 +67,59 @@ nixlTelemetry::~nixlTelemetry() {
...
 }
 
+namespace {
+[[nodiscard]] std::optional<std::string>
+getExporterName() {
+    if (const auto name = nixl::config::getValueOptional<std::string>(telemetryExporterVar)) {
+        return name;
+    }
+
+    NIXL_INFO << "No telemetry exporter was specified, using default: " << defaultTelemetryPlugin;
```
guy-ealey-morag: This should be logged after the existence check for `telemetryDirVar`, because if the existence check fails then the default value is not used (but still logged)

src/plugins/mooncake/mooncake_backend.cpp
guy-ealey-morag: Why is `getValueWithStatus` used here if we ignore the return value?

src/plugins/ucx/gpu_xfer_req_h.cpp
```diff
@@ -104,7 +81,8 @@ createGpuXferReq(const nixlUcxEp &ep,
     params.element_size = sizeof(ucp_device_mem_list_elem_t);
     params.num_elements = ucp_elements.size();
 
-    const auto timeout = get_gpu_xfer_timeout();
+    const auto timeout = nixl::config::getValueDefaulted("NIXL_UCX_GPU_XFER_TIMEOUT_MS",
```
guy-ealey-morag: This changed the log about using the default value from `WARN` to `DEBUG`, is this okay?

src/utils/object/s3/utils.cpp
```diff
@@ -69,11 +70,7 @@ getBucketName(nixl_b_params_t *custom_params) {
         }
     }
 
-    const char *env_bucket = std::getenv("AWS_DEFAULT_BUCKET");
-    if (env_bucket && env_bucket[0] != '\0') return std::string(env_bucket);
-
-    throw std::runtime_error("Bucket name not found. Please provide 'bucket' in custom_params or "
-                             "set AWS_DEFAULT_BUCKET environment variable");
+    return nixl::config::getNonEmptyString("AWS_DEFAULT_BUCKET");
```
guy-ealey-morag, line 73: The error text here now omits the part about the option to provide 'bucket' in custom params.
Maybe we should allow a custom error message for this case?

src/core/telemetry/telemetry.cpp
```diff
@@ -66,49 +67,59 @@ nixlTelemetry::~nixlTelemetry() {
...
+}
+
+} // namespace
+
 void
 nixlTelemetry::initializeTelemetry() {
-    auto buffer_size = std::getenv(TELEMETRY_BUFFER_SIZE_VAR) ?
-        std::stoul(std::getenv(TELEMETRY_BUFFER_SIZE_VAR)) :
-        DEFAULT_TELEMETRY_BUFFER_SIZE;
+    const size_t buffer_size = nixl::config::getValueDefaulted<size_t>(
```
rakhmets: auto?
as you usually use auto in such cases in this PR.

src/utils/common/configuration.h
guy-ealey-morag, line 66: This is a suggestion regarding the previous PR: 

These two vectors are heap allocated and could be changed to `array` of `string_view` like:
```C++
static constexpr std::array<std::string_view, 6> positive = {"y", "yes", "on", "1", "true", "enable"};
```

src/core/nixl_agent.cpp
brminich: why not `telemetry_enabled` or something mor meaningfull?

src/utils/common/configuration.h
ColinNV: It's two very small arrays that are allocated once and it's really not performance critical...

src/utils/object/s3/utils.cpp
ColinNV: I thought about it, but I'd expect that message to not be useful because only the developer of the software can do something about the params.

src/plugins/ucx/gpu_xfer_req_h.cpp
ColinNV: Probably but I'm not 100% sure. I do believe it should be consistent, one goal of this exercise is to make all places that need config values behave the same.

src/plugins/mooncake/mooncake_backend.cpp
ColinNV: Because it doesn't throw and provides the required semantics of changing the first argument only when everything went well. It's changed now.

src/utils/common/configuration.h
```diff
@@ -182,6 +198,22 @@ getValueDefaulted(const std::string &env, const type &fallback) {
...
 
+[[nodiscard]] inline std::string
+getNonEmptyString(const std::string &env) {
+    const std::string result = getValue<std::string>(env);
+
+    if (result.empty()) {
+        throw std::runtime_error("Environment variable '" + env + "' needs non-empty value");
+    }
+    return result;
+}
```
brminich, line 209: why is this method needed?

ColinNV: It felt like good utility. Implement the "if empty" check and error message only once. The caller now also doesn't have to use a variable (rather than a literal) for the environment variable name (or duplicate it).

src/core/telemetry/telemetry.cpp
ColinNV: You're right; I don't like gratuitous `auto`, but the type is repeated in the same line.

src/plugins/ucx/config.cpp
```diff
@@ -18,16 +18,22 @@
...
+const std::string prefix = "UCX_";
+
+} // namespace
+
 namespace nixl::ucx {
 void
 config::modify(std::string_view key, std::string_view value) const {
-    const char *env_val = std::getenv(absl::StrFormat("UCX_%s", key.data()).c_str());
+    // Temporary std::string from key needed until C++26.
+    const auto env_val = nixl::config::getValueOptional<std::string>(prefix + std::string(key));
```
brminich: why is having prefix var better?

src/plugins/ucx/gpu_xfer_req_h.cpp
brminich: @rakhmets

src/plugins/ucx/config.cpp
ColinNV: It's just to construct the `std::string` only once, and so with C++26 it simplifies to `prefix + key`. Since that's still far away I have no strong opinion on this.

brminich: me neither, but it looks less readable

ColinNV: Done.

src/plugins/ucx/gpu_xfer_req_h.cpp
rakhmets: IMO, it's ok to have a `DEBUG` log here.

src/core/telemetry/telemetry.cpp
```diff
@@ -66,49 +67,60 @@ nixlTelemetry::~nixlTelemetry() {
     }
 }
 
+namespace {
+[[nodiscard]] std::optional<std::string>
+getExporterName() {
+    if (const auto name = nixl::config::getValueOptional<std::string>(telemetryExporterVar)) {
+        return name;
```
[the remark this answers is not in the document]
ColinNV: The issue was carried over. The fix is wrong because `getNonEmptyString()` doesn't return an optional, but it's easy enough to fix properly.

brminich (dismissed):

brminich (approved):

PR #1349, opened 2026-02-21
libfabric: Progress control rail in PT for tcp provider

yexiang-aws:
## What?
Detect TCP provider by name and manually progress control rails in the progress thread.

## Why?

The TCP provider (libfabric's xnet provider) advertises FI_PROGRESS_AUTO for both control and data progress modes, but in practice requires  manual progress on the receiver side for control messages (fi_senddata). This causes control rail operations to hang when relying solely on the provider's auto-progress mechanism.

Observed behavior:
- **TCP provider**: Control messages require manual fi_cq_read() on remote despite FI_PROGRESS_AUTO
- **Sockets provider**: Control messages work without manual progress

## Summary by CodeRabbit

* **Improvements**
  * Progress system now detects provider type at runtime and conditionally handles control-rail processing within the background thread, reducing unnecessary work and improving efficiency.
  * Background loop can progress both data and control paths when needed, with clearer logging for dual-path operation.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1376,6 +1376,20 @@ nixlLibfabricEngine::genNotif(const std::string &remote_agent, const std::string
     return notifSendPriv(remote_agent, notifications, total_msg_len, 0, 0);
 }
 
+/**
+ * @brief Retrieve pending notifications from the notification queue.
+ *
+ * This function progresses completion queues (if needed) and retrieves any
+ * pending notifications. It avoids competing with the progress thread by
+ * conditionally progressing rails based on thread state:
+ *
+ * - Data rails: Progressed only if progress thread is disabled
+ * - Control rails: Progressed only if progress thread is disabled OR not handling them
+ *
+ * @param[out] notif_list List to append retrieved notifications to
+ * @return NIXL_SUCCESS if notifications retrieved, NIXL_IN_PROG if none available,
+ *         error code on failure
+ */
 nixl_status_t
 nixlLibfabricEngine::getNotifs(notif_list_t &notif_list) {
     // Progress data rails if progress thread is not enabled
@@ -1389,7 +1403,7 @@ nixlLibfabricEngine::getNotifs(notif_list_t &notif_list) {
 
     // Progress control rails only if progress thread is not handling them
     // Similar to data rails - avoid competing with progress thread
-    if (!progress_thread_enabled_ || !progress_thread_handles_control_rails_) {
+    if (!progress_thread_enabled_ || !progress_thread_handles_control_rails_.load()) {
         nixl_status_t progress_status = rail_manager.progressAllControlRails();
         if (progress_status != NIXL_SUCCESS && progress_status != NIXL_IN_PROG) {
             NIXL_ERROR << "Failed to progress control rails in getNotifs.";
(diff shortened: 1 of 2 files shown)


src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1417,22 +1423,54 @@ nixlLibfabricEngine::getNotifs(notif_list_t &notif_list) {
...
+            // Progress control rails for TCP provider (xnet)
+            if (provider_name == "tcp") {
+                progress_control_rails = true;
+                NIXL_DEBUG << "PT: TCP provider detected, will progress control rails";
+            }
+        }
+    }
+
+    // Store this decision so getNotifs knows not to compete
+    progress_thread_handles_control_rails_ = progress_control_rails;
```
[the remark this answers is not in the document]
fengjica: Data race is a valid concern.

It's probably better to move the whole logic of deciding progress_thread_handles_control_rails_ before the progress thread, that way we probably don't need any atomic.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1417,22 +1423,54 @@ nixlLibfabricEngine::getNotifs(notif_list_t &notif_list) {
...
+
+    // Main progress loop - continuously process completions
     while (!progress_thread_stop_.load()) {
-        // Process completions only on data rails (non-blocking)
+        // Process completions on data rails (non-blocking)
         bool any_completions = false;
         nixl_status_t status = rail_manager.progressActiveDataRails();
         if (status == NIXL_SUCCESS) {
             any_completions = true;
-            NIXL_DEBUG << "PT: Processed completions on data rails";
```
fengjica, line 1431: Keep this in the log will probably help us in he future to see that a user is using PT, if they can enable DEBUG. Otherwise, we will ask them to check their setup or upload the complete log, which is harder for them. 

Same argument to add some log for PT on control rail progress.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1417,22 +1423,54 @@ nixlLibfabricEngine::getNotifs(notif_list_t &notif_list) {
...
+    // TCP provider needs manual progress despite reporting FI_PROGRESS_AUTO
+    bool progress_control_rails = false;
+    if (rail_manager.getNumDataRails() > 0) {
+        fi_info *info = rail_manager.getDataRail(0).getRailInfo();
+        if (info && info->fabric_attr && info->fabric_attr->prov_name) {
+            std::string provider_name(info->fabric_attr->prov_name);
+            NIXL_DEBUG << "PT: Provider name: " << provider_name;
+
+            // Progress control rails for TCP provider (xnet)
+            if (provider_name == "tcp") {
```
bwbarrett, line 1466: I don't understand why testing for the tcp provider here is the right call.  Your progress behavior should be driven by the progress mode bits in the info object, not by names.

bwbarrett (changes requested):

src/plugins/libfabric/libfabric_backend.cpp
yexiangd: I wanted to check based on the progress mode bits. But tcp provider is FI_PROGRESS_AUTO as from - https://github.com/aws/libfabric/blob/2.4.0amzn1.0/prov/tcp/src/xnet_attr.c#L151-L152 . I don't understand why manual progress is still required. (both tcp and socket provider are FI_PROGRESS_AUTO, but socket provider doesn't need this change to manually progress, thus progress mode won't be able to differentiate tcp provider here.)


I specifically checked and printed data_progress and control_progress. Validated both are FI_PROGRESS_AUTO. So that I'm not mistakenly disabling the FI_PROGRESS_AUTO.

I'm asking the same question in ofiwg libfabric - https://github.com/ofiwg/libfabric/issues/11906

yexiangd:
Creating another PR for removing control rail. The changes in this PR will not be needed

PR #1350, opened 2026-02-23
Tests/Unit: Cleanup GDS Test Flags

aranadive:
Having both skip_write and skip_read flags is a bit confusing since you would always have to perform a write before read to verify test data. This was also incorrect since with skip_write, the read operation would read garbage values from newly created but empty files.

Fix this by removing the flags and simplifying test to do write then reads. Keep the measurement part for each operation.

## Summary by CodeRabbit

* **Tests**
  * All test phases now run by default — skip/read/write flags removed; selective disabling no longer supported.
  * Phase flow simplified and made unconditional; buffer-clearing, write, read and validation run with clearer progress reporting.
  * Validation now uses single-pass preallocation and reuses expected data for more efficient, consistent checks.

* **Chores**
  * Updated copyright year and messaging/formatting tweaks.
  * Added timestamped filename generation and streamlined validation buffer handling for efficiency.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/nixl/pull/1350" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
+++ b/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
@@ -83,7 +83,7 @@ void print_usage(const char* program_name) {
         << "  -D, --direct            Use O_DIRECT for file operations (bypass page cache)\n"
         << "  -h, --help              Show this help message\n"
         << "\nExample:\n"
-        << "  " << program_name << " -d -n 100 -s 2M -p 16 -b 256 -m 32M -t 5 -D /path/to/dir\n";
+        << "  " << program_name << " -d -n 100 -s 2M -p 16 -b 128 -m 16M -t 5 -D /path/to/dir\n";
 }
 
 void printProgress(float progress) {
--- a/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
+++ b/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
@@ -115,13 +115,6 @@ std::string generate_timestamped_filename(const std::string& base_name) {
     return base_name + std::string(timestamp);
 }
 
-void validateBuffer(void* expected, void* actual, size_t size, const char* operation) {
-    if (memcmp(expected, actual, size) != 0) {
-        std::cerr << "Data validation failed for " << operation << std::endl;
-        exit(-1);
-    }
-}
-
 // Helper function to fill buffer with repeating pattern
 void fill_test_pattern(void* buffer, size_t size) {
     char* buf = (char*)buffer;
--- a/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
+++ b/test/unit/plugins/cuda_gds/nixl_gds_test.cpp
@@ -197,7 +197,7 @@ int main(int argc, char *argv[])
     double                      total_data_gb = 0;
     bool                        use_direct = false;
     unsigned int                iterations = DEFAULT_ITERATIONS;
-    int                         parsed = 0;
+    int parsed = 0;
 
     // Parse command line options
     static struct option long_options[] = {{"dram", no_argument, 0, 'd'},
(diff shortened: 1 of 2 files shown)


test/unit/plugins/cuda_gds/nixl_gds_test.cpp
```diff
@@ -64,23 +64,26 @@ size_t parse_size(const char* size_str) {
...
+        << "  -p, --pool-size SIZE    Size of batch pool (default: 8, range: 1-32)\n"
+        << "  -b, --batch-limit SIZE  Maximum requests per batch  (default: 128, Max allowed: "
+           "1-128)\n"
+        << "  -m, --max-req-size SIZE Maximum size per request (default: 16M, Max allowed: 16M)\n"
+        << "  -t, --iterations N      Number of iterations for each transfer (default: "
+        << DEFAULT_ITERATIONS << ")\n"
+        << "  -D, --direct            Use O_DIRECT for file operations (bypass page cache)\n"
+        << "  -h, --help              Show this help message\n"
+        << "\nExample:\n"
+        << "  " << program_name << " -d -n 100 -s 2M -p 16 -b 256 -m 32M -t 5 -D /path/to/dir\n";
```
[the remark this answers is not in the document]
aranadive: Fixed

test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp
```diff
@@ -683,31 +615,47 @@ main (int argc, char *argv[]) {
         std::cout << "\n============================================================" << std::endl;
         std::cout << "PHASE 5: Validating read data" << std::endl;
         std::cout << "============================================================" << std::endl;
+
+        // Pre-allocate validation buffers once for all transfers
+        char *host_buffer = use_vram ? (char *)malloc(transfer_size) : NULL;
+        char *expected_buffer = (char *)malloc(transfer_size);
```
rakhmets: It's better to use `unique_ptr` here, e.g.:
```C++
auto buffer = std::make_unique<char[]>(transfer_size);
...
validate(..., buffer.get(), ...);
```

aranadive: Done

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
```diff
@@ -283,11 +253,12 @@ int main(int argc, char *argv[])
                 }
                 break;
             case 't':
-                iterations = atoi(optarg);
-                if (iterations <= 0) {
+                parsed = atoi(optarg);
```
rakhmets: The variable is not used outside this case.
Should be declared here as `const int`.

test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp
```diff
@@ -280,32 +251,29 @@ main (int argc, char *argv[]) {
...
             break;
         case 'G':
-            num_gpus = atoi (optarg);
-            if (num_gpus <= 0) {
+            parsed = atoi(optarg);
+            if (parsed <= 0) {
                 std::cerr << "Error: Number of GPUs must be positive\n";
                 return 1;
             }
+            num_gpus = parsed;
```
rakhmets: Please avoid code duplication. Smth like:
```C++
size_t parse_positive_size(const char* str, std::string_view name) {
    try {
        std::size_t pos = 0;
        size_t value = std::stoull(str, &pos);

        if ((str[pos] != '\0') || (value == 0))
            throw std::invalid_argument("");

        return value;
    }
    catch (...) {
        throw std::invalid_argument(
            "Number of " + std::string(name) + " must be a positive integer");
    }
}
```

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
aranadive: Not needed now.

test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp
aranadive: Added. Thanks.

PR #1355, opened 2026-02-24
CORE: Improve memory handling.

ColinNV:
## What?
Improve memory handling by simplifying the code incl. removing an uneeded indirection.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal memory and resource ownership for greater stability and fewer leaks.
  * Converted internal components to safer ownership and simplified cleanup logic.
  * Made agent internal data handle non-reassignable to strengthen runtime safety.
  * No visible behavior or API changes for end users.

### DIFF
--- a/src/api/cpp/nixl.h
+++ b/src/api/cpp/nixl.h
@@ -18,8 +18,8 @@
  * @file nixl.h (NVIDIA Inference Xfer Library)
  * @brief These are NIXL Core APIs for applications
  */
-#ifndef _NIXL_H
-#define _NIXL_H
+#ifndef NIXL_SRC_API_CPP_NIXL_H
+#define NIXL_SRC_API_CPP_NIXL_H
 
 #include "nixl_types.h"
 #include "nixl_params.h"
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -212,7 +212,7 @@ nixlAgent::~nixlAgent() {
             close(fd);
         }
 
-        data->listener.reset(); // TOOD: Needed?
+        data->listener.reset(); // TODO: Needed?
     }
 }
 


src/core/agent_data.h
```diff
@@ -87,16 +88,19 @@ class nixlAgentData {
...
-        nixlLocalSection*                                        memorySection;
+        nixlLocalSection memorySection;
 
         std::unordered_map<std::string,
                            std::unordered_map<nixl_backend_t, nixl_blob_t>,
                            std::hash<std::string>, strEqual>     remoteBackends;
-        std::unordered_map<std::string, nixlRemoteSection*,
-                           std::hash<std::string>, strEqual>     remoteSections;
+        std::unordered_map<std::string,
+                           std::unique_ptr<nixlRemoteSection>,
```
iyastreb, line 97: why not to store by value?

src/core/agent_data.h
```diff
@@ -122,7 +126,7 @@ class nixlAgentData {
         nixl_status_t
         invalidateRemoteData(const std::string &remote_name);
         [[nodiscard]] static backend_set_t
-        getBackends(const nixl_opt_args_t *opt_args, nixlMemSection *section, nixl_mem_t mem_type);
+        getBackends(const nixl_opt_args_t *opt_args, nixlMemSection &section, nixl_mem_t mem_type);
```
iyastreb, line 129: can we pass section by const?

iyastreb, line 129: It's very nice that you look into it, because I think resolving backends should be completely redesigned.
Since we have a fixed set of backends, there is no any need to use std::set which is dynamically constructed on the heap each time (on the hot path!).
We should probably just use an enum of backends with bit-per-backend.
Then instead of a set we can use just a bitmask.

src/core/agent_data.h
ColinNV: That's still on my todo list.

src/core/agent_data.h
iyastreb: maybe in a separate PR

ColinNV: Yes, this PR is still far from complete and it's getting too big, I'll close it and cut it differently.

ColinNV:
Closing and cutting differently to prevent it from becoming too large.

PR #1358, opened 2026-02-24
feat: dlopen-based nixl-sys stubs for runtime library resolution

nv-anants:
## What?
  - Build libnixl_capi.so shared library from wrapper.cpp (exports nixl_capi_* C symbols) so downstream consumers can load nixl at runtime without build-time linking
  - Rewrite stubs.cpp from abort-on-call to dlopen/dlsym lazy forwarding — stubs now dlopen("libnixl_capi.so") at runtime and forward all calls to the real implementation
  - Add libnixl.so existence check in build.rs before attempting to link, so the fallback to stubs works correctly when nixl-sys is used as a git dependency (headers available
   from source tree but libraries not installed)
  - Link -ldl in stub builds for dlopen/dlsym support

## Why?
  The dynamo project builds the ai-dynamo-runtime wheel without nixl present to avoid bundling nixl .so files in the wheel. Previously, the stubs compiled in this scenario would abort() when any nixl C API function was called, making nixl unusable at runtime even when installed.

  With this change, the stubs lazily load libnixl_capi.so via dlopen at runtime, allowing the wheel to be built without nixl while still using nixl features when the shared library is available in LD_LIBRARY_PATH. 

ref: OPS-3402
closes: OPS-3454

## Summary by CodeRabbit

* **New Features**
  * C API now performs runtime dynamic loading of the native library with clearer diagnostics.
  * Added a public function to detect when stub (fallback) implementations are in use.
  * Build now installs an explicit native C API library target.

* **Refactor**
  * API call paths resolve and forward to the real library at runtime instead of static stubs.
  * Build/link search and error reporting for the native library have been strengthened.

### DIFF
--- a/src/bindings/rust/build.rs
+++ b/src/bindings/rust/build.rs
@@ -141,9 +141,6 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
         ));
     }
 
-    // Register all candidate paths with the linker (deduplicated).
-    lib_search_paths.sort();
-    lib_search_paths.dedup();
     for path in &lib_search_paths {
         println!("cargo:rustc-link-search=native={}", path);
     }


src/bindings/rust/stubs.cpp
```diff
@@ -14,19 +14,57 @@
...
+        std::abort();
+    }
+    void* sym = dlsym(handle, name);
+    if (!sym) {
+        std::cerr << "nixl error: symbol '" << name
+                  << "' not found in libnixl_capi.so: " << dlerror() << "\n";
+        std::abort();
+    }
+    return sym;
+}
```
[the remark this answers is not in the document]
nv-anants: 76db8d832e455f1feeaf1642a7a52e2b785069a1

src/bindings/rust/stubs.cpp
```diff
@@ -14,19 +14,57 @@
...
-// NOTE: The original includes from nixl.h, nixl_types.h, cstring, exception, etc. are removed here.
-// The original blank lines around includes are also implicitly handled by this replacement.
+// Thread-safe lazy initialization of the nixl C API shared library handle.
+// C++11 guarantees thread-safe initialization of function-local statics.
+static void*
+get_nixl_handle()
+{
+    static void* handle = dlopen("libnixl_capi.so", RTLD_NOW | RTLD_GLOBAL);
+    return handle;
+}
```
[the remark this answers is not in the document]
nv-anants: 76db8d832e455f1feeaf1642a7a52e2b785069a1

src/bindings/meson.build
```diff
@@ -14,6 +14,19 @@
...
 
 subdir('python')
+
+# Shared library exposing the nixl C API (nixl_capi_* symbols).
+# Always built so LD_PRELOAD can override stub symbols at runtime when
+# downstream Rust binaries were compiled without nixl present.
+nixl_capi_inc = include_directories('rust/../../api/cpp', 'rust/../../infra', 'rust/../../core')
+shared_library('nixl_capi',
+    sources: ['rust/wrapper.cpp'],
+    include_directories: [nixl_capi_inc, include_directories('/usr/include')],
```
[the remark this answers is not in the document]
nv-anants: 76db8d832e455f1feeaf1642a7a52e2b785069a1

src/bindings/meson.build
```diff
@@ -14,6 +14,19 @@
 # limitations under the License.
 
 subdir('python')
+
+# Shared library exposing the nixl C API (nixl_capi_* symbols).
+# Always built so LD_PRELOAD can override stub symbols at runtime when
+# downstream Rust binaries were compiled without nixl present.
+nixl_capi_inc = include_directories('rust/../../api/cpp', 'rust/../../infra', 'rust/../../core')
```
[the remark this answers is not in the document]
nv-anants: 76db8d832e455f1feeaf1642a7a52e2b785069a1

src/bindings/rust/build.rs
```diff
@@ -106,6 +106,17 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
...
+    // Verify that nixl shared libraries actually exist before proceeding.
+    // Without this check, wrapper.cpp may compile (headers found in source tree)
+    // but linking will fail later when the .so files are missing.
+    let nixl_so = format!("{}/libnixl.so", nixl_lib_path);
+    if !std::path::Path::new(&nixl_so).exists() {
+        return Err(anyhow::anyhow!(
+            "libnixl.so not found at {}; nixl libraries are not installed",
+            nixl_so
+        ));
+    }
```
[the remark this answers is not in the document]
nv-anants: [5c40efb](https://github.com/ai-dynamo/nixl/pull/1358/commits/5c40efbdb98408cab83e9bea8760d038eb5d36a2)

src/bindings/rust/stubs.cpp
```diff
@@ -14,19 +14,75 @@
...
+            std::cerr << " dlopen error: " << h.error;
+        }
+        std::cerr << "\n";
+        std::abort();
+    }
+    void *sym = dlsym(h.handle, name);
+    if (!sym) {
+        std::cerr << "nixl error: symbol '" << name
+                  << "' not found in libnixl_capi.so: " << dlerror() << "\n";
+        std::abort();
```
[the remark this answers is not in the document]
nv-anants: [6892d12](https://github.com/ai-dynamo/nixl/pull/1358/commits/6892d1231609f370643bd1135742ce16ece458a3)

src/bindings/rust/build.rs
```diff
@@ -106,25 +106,46 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
...
+            lib_search_paths
+        ));
+    }
+
+    // Register all candidate paths with the linker.
+    for path in &lib_search_paths {
+        println!("cargo:rustc-link-search=native={}", path);
+    }
+    println!("cargo:rustc-link-search=native={}", nixl_root_path);
+
```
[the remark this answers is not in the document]
nv-anants: [dd35746](https://github.com/ai-dynamo/nixl/pull/1358/commits/dd35746ceb268f4796f8a9f594fba55c26579ab5)

src/bindings/rust/build.rs
```diff
@@ -106,31 +106,52 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
...
-    println!("cargo:rustc-link-search=native={}/lib64", nixl_root_path);
-    println!("cargo:rustc-link-search=native={}/lib/x86_64-linux-gnu", nixl_root_path);
+    // Collect all candidate library directories: NIXL_PREFIX-derived paths
+    // first, then any paths reported by pkg-config.
+    let mut lib_search_paths = vec![
+        nixl_lib_path.clone(),
+        format!("{}/lib", nixl_root_path),
+        format!("{}/lib64", nixl_root_path),
+        format!("{}/lib/{}-linux-gnu", nixl_root_path, arch),
+    ];
```
milesial, line 117: The diff is showing we're removing some paths we had before

nv-anants: updated

src/bindings/rust/build.rs
```diff
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
roiedanino, line 1:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

src/bindings/rust/stubs.cpp
```diff
@@ -39,39 +97,42 @@ struct nixl_capi_reg_dlist_s { /* empty */ };
...
 
 nixl_capi_status_t
-nixl_capi_create_agent(const char* name, nixl_capi_agent_t* agent)
-{
-  return nixl_capi_stub_abort();
+nixl_capi_create_agent(const char *name, nixl_capi_agent_t *agent) {
+    using fn_t = nixl_capi_status_t (*)(const char *, nixl_capi_agent_t *);
+    static fn_t real = (fn_t)resolve("nixl_capi_create_agent");
+    return real(name, agent);
 }
```
roiedanino, line 110: Wouldn't `__attribute__((weak))` do exactly that?

Can we please use `__attribute__((weak))` instead?

nv-anants: With `__attribute__((weak))`, each stub just aborts if called directly. The real library overrides these at link time. However, unlike dlopen, there's no   automatic runtime discovery — the real `libnixl_capi.so` must be linked at load time (via DT_NEEDED or LD_PRELOAD explicitly). The current dlopen approach finds the library at runtime from LD_LIBRARY_PATH without needing it linked.

ovidiusm:
Please update branch to avoid GPU test failure

src/bindings/rust/build.rs
ryanolson: adding arch is a great change!

src/bindings/rust/build.rs
ryanolson: I'm pretty sure we are supposed to keep the date range.  we should check with legal.  there seems to be a lot of confusion on this

nv-anants: +1 to 2025-2026, i thought we keep from file added - current year?

aranadive: Yes this should be the date range unless its a brand new file.

brminich: +1 to a range

src/bindings/rust/build.rs
```diff
@@ -106,31 +106,52 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
...
-            for path in lib.link_paths {
-                println!("cargo:rustc-link-search=native={}", path.display());
+        for lib in &libs {
+            for path in &lib.link_paths {
+                lib_search_paths.push(path.display().to_string());
             }
         }
     } else {
         println!("cargo:warning=pkg-config not available, using manual library paths");
     }
```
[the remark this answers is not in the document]
nv-anants: i think it is unrealted to this change?

src/bindings/rust/build.rs
```diff
@@ -106,31 +106,52 @@ fn build_nixl(cc_builder: &mut cc::Build) -> anyhow::Result<()> {
...
+    });
+    if !nixl_so_found {
+        return Err(anyhow::anyhow!(
+            "libnixl.so not found in any search path {:?}; nixl libraries are not installed",
+            lib_search_paths
+        ));
+    }
+
+    // Register all candidate paths with the linker (deduplicated).
+    lib_search_paths.sort();
```
ovidiusm: I suppose that you are sorting to dedup, but doesn't this break the precedence ordering? Shouldn't NIXL_PREFIX have priority?

nv-anants: good point yes, makes sense. I can either use like a hashset to preserve order and dedup but honestly dedup is not necessarily required  and i could just keep it simple as well, it could result in duplicate/inconsistent rustc-link-search registrations but not an issue as such. Which one do you prefer?

ovidiusm: I would skip the dedup

nv-anants: updated, thanks! https://github.com/ai-dynamo/nixl/pull/1358/changes/2ebfdf0f37d3e4e2e213952413225631d2e39d83

PR #1359, opened 2026-02-25
TEST/GTEST: Ignore IB warning.

ColinNV:
## What?
Don't fail the gtests when `IB devices were detected, but accelerated IB support was not found` is logged

## Why?
This message is occurring wrongly too often and needs to be ignored until the underlying issue is fixed.

## Summary by CodeRabbit

* **Tests**
  * Tests updated to always suppress a specific system message during test runs, reducing noisy or flaky output in CI and local runs while preserving other test behaviors.

### DIFF
--- a/.ci/jenkins/lib/build-matrix.yaml
+++ b/.ci/jenkins/lib/build-matrix.yaml
@@ -43,7 +43,7 @@ env:
   TEST_TIMEOUT: 30
   UCX_TLS: "^shm"
   STORAGE_DRIVER: 'overlay'
-  CI_IMAGE_TAG: "20260318-5"
+  CI_IMAGE_TAG: "20260323-1"
 
 
 runs_on_dockers:
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -49,7 +49,7 @@ env:
   SLURM_JOB_TIMEOUT: '02:20:00'
   SLURM_IMMEDIATE_TIMEOUT: "3600"
   STORAGE_DRIVER: overlay
-  CI_IMAGE_TAG: "20260318-5"
+  CI_IMAGE_TAG: "20260323-1"
 
 empty_volumes:
   - {mountPath: /var/lib/containers/storage, memory: false}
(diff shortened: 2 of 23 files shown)


guy-ealey-morag:
I think there is a potential issue with how we count how many times each log was ignored.

In some cases we might have duplicate entries in the `log_problem_ignore` list (for example one that is added by `RunAllTests()` and another added by the test itself).
If that log appears it is counted only towards the first one in the list.

I suggest changing the impl of `LogProblemCounter::Send()` to count all matches (without calling `return` on match, and then increase `global_problem_count` if any match was found)

ColinNV:
> I think there is a potential issue with how we count how many times each log was ignored.
> 
> In some cases we might have duplicate entries in the `log_problem_ignore` list (for example one that is added by `RunAllTests()` and another added by the test itself). If that log appears it is counted only towards the first one in the list.
> 
> I suggest changing the impl of `LogProblemCounter::Send()` to count all matches (without calling `return` on match, and then increase `global_problem_count` if any match was found)

I don't really want to make any other changes for a temporary work-around. Under normal circumstances we won't have two `logIgnoreGuards` for the same log message. Should it become permanent the other two instances of ignoring the same message would be removed.

guy-ealey-morag:
> > I think there is a potential issue with how we count how many times each log was ignored.
> > In some cases we might have duplicate entries in the `log_problem_ignore` list (for example one that is added by `RunAllTests()` and another added by the test itself). If that log appears it is counted only towards the first one in the list.
> > I suggest changing the impl of `LogProblemCounter::Send()` to count all matches (without calling `return` on match, and then increase `global_problem_count` if any match was found)
> 
> I don't really want to make any other changes for a temporary work-around. Under normal circumstances we won't have two `logIgnoreGuards` for the same log message. Should it become permanent the other two instances of ignoring the same message would be removed.

Of course it doesn't have to be part of this PR, it just made me notice the issue.
I think it's worth fixing separately because it will help prevent unexpected behavior in the future.

ColinNV:
@guy-ealey-morag @brminich @ovidiusm It feels like things have calmed down and we might not need this temporary work-around anymore?

brminich:
> @guy-ealey-morag @brminich @ovidiusm It feels like things have calmed down and we might not need this temporary work-around anymore?

lets keep it open for some time and see whether this issue pops up again

PR #1360, opened 2026-02-25
Swap cuda memory initialization

ovidiusm:
Allocation helper is also changing CUDA context (cudaSetDevice). Ensure that we end up on the context of agent 1 when we post the transfer from agent 1.

## Summary by CodeRabbit

* **Tests**
  * Updated unit tests to adjust memory allocation/registration and initialization ordering to improve test stability and context handling.
  * Refresh copyright year range in test headers.

### DIFF
--- a/test/unit/plugins/ucx/ucx_backend_test.cpp
+++ b/test/unit/plugins/ucx/ucx_backend_test.cpp
@@ -582,6 +582,8 @@ test_inter_agent_transfer(bool p_thread,
 
     void *addr1 = NULL, *addr2 = NULL;
     nixlBackendMD *lmd1, *lmd2;
+    // Always allocate the source memory last to ensure that we stay on the context
+    // of the source agent.
     allocateAndRegister(ucx2, dst_dev_id, dst_mem_type, addr2, len, lmd2);
     allocateAndRegister(ucx1, src_dev_id, src_mem_type, addr1, len, lmd1);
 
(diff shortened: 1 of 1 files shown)


test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -582,8 +582,8 @@ test_inter_agent_transfer(bool p_thread,
 
     void *addr1 = NULL, *addr2 = NULL;
     nixlBackendMD *lmd1, *lmd2;
-    allocateAndRegister(ucx1, src_dev_id, src_mem_type, addr1, len, lmd1);
     allocateAndRegister(ucx2, dst_dev_id, dst_mem_type, addr2, len, lmd2);
+    allocateAndRegister(ucx1, src_dev_id, src_mem_type, addr1, len, lmd1);
```
ColinNV, line 588: Perhaps add a comment on why the ordering is reversed here and below?

ovidiusm: Good idea, done

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -582,8 +582,10 @@ test_inter_agent_transfer(bool p_thread,
 
     void *addr1 = NULL, *addr2 = NULL;
     nixlBackendMD *lmd1, *lmd2;
-    allocateAndRegister(ucx1, src_dev_id, src_mem_type, addr1, len, lmd1);
+    // Always allocate the source memory last to ensure that we stay on the context
+    // of the source agent.
     allocateAndRegister(ucx2, dst_dev_id, dst_mem_type, addr2, len, lmd2);
+    allocateAndRegister(ucx1, src_dev_id, src_mem_type, addr1, len, lmd1);
```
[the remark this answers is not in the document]
ovidiusm: Why would we set the CUDA context when the source agent memory type is DRAM?

rakhmets (approved):

PR #1361, opened 2026-02-25
CORE/AGENT: Prepare memory handling refactoring.

ColinNV:
## What?
Some preparatory refactorings and cleanups.

## Why?
To prevent the memory handling refactoring from becoming too large.

## How?
- Fix include guards.
- Fix `nixlBackendH` constructor.
- Declare `nixlAgent`'s pimpl (`data`) as `const`.
- Stick two of `nixlAgentData`'s pointer data members in a `std::unique_ptr` (`memorySection`, which will remove the indirection in the next PR, and `listener`).
- Invert the logic of two tests in `nixlAgent::createBackend()` to make it more readable and more amenable to the next refactoring step.

## Summary by CodeRabbit

* **Refactor**
  * Switched internal resources to managed ownership and simplified lifecycle management for safer memory handling.
  * Unified plugin backend creation and registration into a clearer, single flow with more explicit validation.

* **Bug Fixes**
  * Strengthened initialization and error paths with improved cleanup on failures to reduce leaks and inconsistent backend states.

### DIFF
--- a/.ci/docs/setup_nvidia_gpu_with_rdma_support_on_ubuntu.md
+++ b/.ci/docs/setup_nvidia_gpu_with_rdma_support_on_ubuntu.md
@@ -34,7 +34,11 @@ sudo reboot
 If running on nvlink hosts like DGX we should also install fabric manager
 ```bash
 sudo apt install nvidia-fabricmanager-<version> # should be same as kernel version nvidia-fabricmanager-575
+sudo systemctl enable --now nvidia-fabricmanager
 ```
+
+**Important**: On NVSwitch-enabled systems (DGX A100, H100), `nvidia-fabricmanager` must be running before GPU initialization.
+Without it, CUDA will fail with "system not yet initialized" errors. Ensure the service is enabled at boot.
 
 Verify with `nvidia-smi`. Driver compatibility is critical for RDMA support[^1_1][^1_3].
 
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -23,53 +23,50 @@ set -x
 # Parse commandline arguments with first argument being the install directory.
 INSTALL_DIR=$1
 
-if [ -z "$INSTALL_DIR" ]; then
-    echo "Usage: $0 <install_dir>"
-    exit 1
-fi
-
-ARCH=$(uname -m)
-[ "$ARCH" = "arm64" ] && ARCH="aarch64"
-
-export LD_LIBRARY_PATH=${INSTALL_DIR}/lib:${INSTALL_DIR}/lib/$ARCH-linux-gnu:${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins:/usr/local/lib:$LD_LIBRARY_PATH
-export CPATH=${INSTALL_DIR}/include::$CPATH
-export PATH=${INSTALL_DIR}/bin:$PATH
-export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
-export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
-export NIXL_PREFIX=${INSTALL_DIR}
-# Raise exceptions for logging errors
-export NIXL_DEBUG_LOGGING=yes
-
 if [ -n "$VIRTUAL_ENV" ] && grep -q '^uv =' "$VIRTUAL_ENV/pyvenv.cfg" 2>/dev/null; then
     pip3="uv pip"
 else
     pip3="python3 -m pip"
 fi
 
-# Install build dependencies
-if [ -n "$VIRTUAL_ENV" ] ; then
-    # Install full build dependencies in venv
-    $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
-else
-    # Install minimal build dependencies in system python
-    $pip3 install --break-system-packages tomlkit
+if [ -n "$INSTALL_DIR" ]
+then
+    ARCH=$(uname -m)
+    [ "$ARCH" = "arm64" ] && ARCH="aarch64"
+
+    export LD_LIBRARY_PATH=${INSTALL_DIR}/lib:${INSTALL_DIR}/lib/$ARCH-linux-gnu:${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins:/usr/local/lib:$LD_LIBRARY_PATH
+    export CPATH=${INSTALL_DIR}/include::$CPATH
+    export PATH=${INSTALL_DIR}/bin:$PATH
+    export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
+    export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
+    export NIXL_PREFIX=${INSTALL_DIR}
+    # Raise exceptions for logging errors
+    export NIXL_DEBUG_LOGGING=yes
+
+    # Install build dependencies
+    if [ -n "$VIRTUAL_ENV" ] ; then
+        # Install full build dependencies in venv
+        $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
+    else
+        # Install minimal build dependencies in system python
+        $pip3 install --break-system-packages tomlkit
+    fi
+    # Set the correct wheel name based on the CUDA version
+    cuda_major=$(nvcc --version | grep -oP 'release \K[0-9]+')
+    case "$cuda_major" in
+        12|13) echo "CUDA $cuda_major detected" ;;
+        *) echo "Error: Unsupported CUDA version $cuda_major"; exit 1 ;;
+    esac
+    ./contrib/tomlutil.py --wheel-name "nixl-cu${cuda_major}" pyproject.toml
+    # Control ninja parallelism during pip build to prevent OOM (NPROC from common.sh)
+    $pip3 install --break-system-packages --config-settings=compile-args="-j${NPROC}" .
+    $pip3 install --break-system-packages dist/nixl-*none-any.whl
 fi
-# Set the correct wheel name based on the CUDA version
-cuda_major=$(nvcc --version | grep -oP 'release \K[0-9]+')
-case "$cuda_major" in
-    12|13) echo "CUDA $cuda_major detected" ;;
-    *) echo "Error: Unsupported CUDA version $cuda_major"; exit 1 ;;
-esac
-./contrib/tomlutil.py --wheel-name "nixl-cu${cuda_major}" pyproject.toml
-# Control ninja parallelism during pip build to prevent OOM (NPROC from common.sh)
-$pip3 install --break-system-packages --config-settings=compile-args="-j${NPROC}" .
-$pip3 install --break-system-packages dist/nixl-*none-any.whl
+
+# Install test dependencies
 $pip3 install --break-system-packages pytest
 $pip3 install --break-system-packages pytest-timeout
 $pip3 install --break-system-packages zmq
-
-# Add user pip packages to PATH
-export PATH="$HOME/.local/bin:$PATH"
 
 echo "==== Running ETCD server ===="
 etcd_port=$(get_next_tcp_port)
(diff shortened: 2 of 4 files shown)


src/core/agent_data.h
```diff
@@ -137,13 +138,15 @@ class nixlAgentData {
...
 // This class hides away the nixlBackendEngine from user of the Agent API
 class nixlBackendH {
     private:
         nixlBackendEngine* engine;
 
-        nixlBackendH(nixlBackendEngine* &engine) { this->engine = engine; }
-        ~nixlBackendH () {}
+        explicit nixlBackendH(nixlBackendEngine *engine) noexcept : engine(engine) {}
+
+        ~nixlBackendH() = default;
```
rakhmets, line 149: Can we remove this?

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,80 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
-        for (auto & elm : mems) {
-            backend_list = &data->memToBackend[elm];
-            // First time creating this backend handle, so unique
-            // The order of creation sets the preference order
-            backend_list->push_back(backend);
-        }
+    bknd_hndl = new nixlBackendH(backend);
+    if (!bknd_hndl) {
+        delete backend;
+        NIXL_ERROR_FUNC << "allocation of backend handle failed for '" << type << "'";
```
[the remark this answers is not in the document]
ColinNV: This change is on the list for the next PR.

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,80 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
+        return NIXL_ERR_BACKEND;
+    }
```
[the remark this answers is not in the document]
ColinNV: Correct, this, too, is on the list for the next PR. This one only reformats `createBackend()`.

src/core/nixl_agent.cpp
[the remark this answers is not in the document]
ColinNV: Not necessary.

src/core/nixl_agent.cpp
```diff
@@ -124,7 +124,7 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
     if (name.empty())
         throw std::invalid_argument("Agent needs a name");
 
-    memorySection = new nixlLocalSection();
+    memorySection.reset(new nixlLocalSection);
```
rakhmets, line 127: Can be placed to member initializer list.

src/core/nixl_agent.cpp
```diff
@@ -151,7 +151,7 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &cfg
 }
 
 nixlAgentData::~nixlAgentData() {
-    delete memorySection;
+    memorySection.reset();
```
rakhmets, line 154: It seems that we don't need this.

src/api/cpp/nixl.h
```diff
@@ -34,7 +34,7 @@
 class nixlAgent {
     private:
         /** @var  data  The members in agent class wrapped into single nixlAgentData member. */
-        std::unique_ptr<nixlAgentData> data;
+        const std::unique_ptr<nixlAgentData> data;
```
rakhmets, line 37: So, we can remove lines 56-57.

src/core/nixl_agent.cpp
```diff
@@ -1931,7 +1925,7 @@ nixlAgent::prepMemView(const nixl_xfer_dlist_t &dlist,
     nixlBackendEngine *engine{nullptr};
 
     NIXL_SHARED_LOCK_GUARD(data->lock);
-    const auto backends = data->getBackends(extra_params, data->memorySection, mem_type);
+    const auto backends = data->getBackends(extra_params, *data->memorySection.get(), mem_type);
```
```suggestion
    const auto backends = data->getBackends(extra_params, *data->memorySection, mem_type);
```

src/core/agent_data.h
ColinNV: If there's no compelling reason for the destructor to be `private`, yes, it would also simplify the next step.

src/core/nixl_agent.cpp
ColinNV: This will be made a data member without the pointer indirection in the next PR.

src/core/nixl_agent.cpp
ColinNV: Will be eliminated with the next PR. Destruction order is important here, this is to make it explicit until the order of the data members is fixed in the next PR.

src/api/cpp/nixl.h
ColinNV: We could, though it's nice to keep as "documentation"?

src/core/agent_data.h
rakhmets: I think the original intention was to allow only the nixlAgent (and nixlAgentData) to manage the lifecycle of the object. But I do not see any practical sense in how this private destructor helps.

ColinNV: Exactly, this class doesn't do anything itself... I'll simplify in the next PR.

src/api/cpp/nixl.h
rakhmets: In my opinion, this line already says that copying this class is prohibited. So, for me, lines 56-57 are a tautology.
However, I cannot say that I am very concerned about these two lines, and that I strongly recommend deleting them. We can keep them.

ColinNV: I hadn't really noticed them; can get rid of them in the next PR.

src/core/nixl_agent.cpp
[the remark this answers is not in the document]
rakhmets: @ColinNV, it looks like a valid suggestion.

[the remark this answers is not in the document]
ColinNV: > @ColinNV, it looks like a valid suggestion.

Yes, I just decided to keep reformatting of this function separate from changing it, to simplify reviewing. The next PR will make the changes.

src/core/nixl_agent.cpp
```diff
@@ -1837,7 +1831,7 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
 
 backend_set_t
 nixlAgentData::getBackends(const nixl_opt_args_t *opt_args,
-                           nixlMemSection *section,
+                           nixlMemSection &section,
```
iyastreb, line 1834: should we pass section by const ref?

src/core/nixl_agent.cpp
```diff
@@ -1848,7 +1842,7 @@ nixlAgentData::getBackends(const nixl_opt_args_t *opt_args,
         return backends;
     }
 
-    const auto mem_type_backends = section->queryBackends(mem_type);
+    const auto mem_type_backends = section.queryBackends(mem_type);
```
iyastreb, line 1845: Copying my comment from closed PR:
I think resolving backends should be completely redesigned.
Since we have a fixed set of backends, there is no any need to use std::set which is dynamically constructed on the heap each time (on the hot path!).
We should probably just use an enum of backends with bit-per-backend.
Then instead of a set we can use just a bitmask.

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,79 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
-    } else {
+    if (!plugin_handle) {
         NIXL_ERROR_FUNC << "unsupported backend '" << type << "'";
         return NIXL_ERR_NOT_FOUND;
     }
 
-    if (backend) {
-        if (backend->getInitErr()) {
+    // Plugin found, use it to create the backend
+    nixlBackendEngine *backend = plugin_handle->createEngine(&init_params);
```
iyastreb, line 331: why not to use smart pointer here, so that we don't need to explicitly delete a backend in case of an error/exception?

When you transfer ownership just call release()

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,79 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
-        data->backendHandles[type] = bknd_hndl;
-        mems = backend->getSupportedMems();
-        for (auto & elm : mems) {
-            backend_list = &data->memToBackend[elm];
-            // First time creating this backend handle, so unique
-            // The order of creation sets the preference order
-            backend_list->push_back(backend);
-        }
+    bknd_hndl = new nixlBackendH(backend);
+    if (!bknd_hndl) {
```
iyastreb, line 374: why this check? new can return nullptr? it may throw bad_alloc

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,79 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
-        data->backendEngines[type] = backend;
-        data->backendHandles[type] = bknd_hndl;
-        mems = backend->getSupportedMems();
-        for (auto & elm : mems) {
-            backend_list = &data->memToBackend[elm];
-            // First time creating this backend handle, so unique
-            // The order of creation sets the preference order
-            backend_list->push_back(backend);
-        }
+    bknd_hndl = new nixlBackendH(backend);
```
iyastreb, line 373: unique_ptr to be exception safe

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,79 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
         return NIXL_ERR_NOT_FOUND;
     }
 
-    if (backend) {
-        if (backend->getInitErr()) {
+    // Plugin found, use it to create the backend
+    nixlBackendEngine *backend = plugin_handle->createEngine(&init_params);
+
+    if (!backend) {
+        NIXL_ERROR_FUNC << "backend creation failed for '" << type << "'";
```
iyastreb, line 334: this error is already logged in createEngine

src/core/nixl_agent.cpp
```diff
@@ -327,80 +322,79 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
+        return NIXL_ERR_BACKEND;
+    }
+
+    if (backend->getInitErr()) {
```
iyastreb, line 338: init_error must disappear, we should not allow constructing semi-valid objects IMO. May it still happen that we create invalid objects? Need to throw an exception in createEngine and return nullptr

src/core/nixl_agent.cpp
ColinNV: Can't, it returns a mutable reference from a non-const method that is called on it.

src/core/nixl_agent.cpp
ColinNV: Sounds good, but not for this particular PR. Will look at this separately.

src/core/nixl_agent.cpp
ColinNV: The next PR with the actual memory handling refactoring will consistently use a smart pointer for this.

src/core/nixl_agent.cpp
ColinNV: Next PR.

src/core/nixl_agent.cpp
ColinNV: Next PR, this one just reformats this function.

src/core/nixl_agent.cpp
ColinNV: Yes, I have wanted to get rid of `init_error` since day 1, but as I understand it we can't fully do that for backwards compatibility. Let's discuss this "offline".

src/core/nixl_agent.cpp
ColinNV: Which one? It's not in `nixlBackendPluginHandle::createEngine()`.

src/core/nixl_agent.cpp
iyastreb: hmm, then we can maybe add a const overload of that method

src/core/nixl_agent.cpp
iyastreb: IMO better implement this change now, it costs nothing, but up to you

src/core/nixl_agent.cpp
iyastreb: I would double check it, but even if we still have it - createEngine can encapsulate this inside and return nullptr object

src/core/nixl_agent.cpp
ColinNV: You're right, the non-constness seems more of an oversight than a necessity, at least for this particular use case. I'll put it on my TODO list.

src/core/nixl_agent.cpp
iyastreb: right, I looked at another createEngine

src/core/nixl_agent.cpp
ColinNV: If you do it consistently, not just here, it's a bigger change, which is why I'm splitting it into this and the next PR.

iyastreb (approved):

PR #1363, opened 2026-02-25
[TESTS/NIXLBENCH] Reduce the number of CLI arg combinations

ovidiusm:
## What?
With increasing number of plugins covered, nixlbench tests are taking a long time (25 minutes) because of sweeping through multiple batch size (3 values) and message size (3 values).

There is little benefit in doing this, we can reduce the test to 1 value to make the CI significantly faster.

Before: 24-25 minutes
After: 3-4 minutes

## Summary by CodeRabbit

* **Tests**
  * Updated Nixlbench test configuration parameters for benchmarking. Adjusted initial block sizing and batch processing settings to modify test execution behavior in benchmark runs.


.gitlab/test_nixlbench.sh
```diff
@@ -63,7 +63,7 @@ wait_for_etcd
 echo "==== Running Nixlbench tests ===="
 cd ${INSTALL_DIR}
 
-DEFAULT_NB_PARAMS="--filepath /tmp --total_buffer_size 80000000 --start_block_size 4096 --max_block_size 16384 --start_batch_size 1 --max_batch_size 4"
```
aranadive, line 66: Can we do start_batch_size=1 and max_batch_size=2? This tests both single and multiple batch sizes.

ovidiusm: Is there value in doing that for all plugins/testcases? It will double test run time

aranadive: I think the overall test time should be about the same since the same amount of data is getting transferred with batch_size=1,2 and batch_size=4. Im okay with just testing 16K as block_size.

ovidiusm: The running time would double, could we avoid it? Right now PRs are queuing a lot and most are timing out

aranadive: Im not sure the running time would double but okay to skip for now.

PR #1365, opened 2026-02-25
TELEMETRY: Optimize nixlTelemetry::addXferTime.

ColinNV:
## What?
Optimize away five string constructions in `nixlTelemetry::addXferTime()`.

## Summary by CodeRabbit

* **Refactor**
  * Internal telemetry event handling modernized for leaner in-memory string usage and minor API modernization; telemetry behavior and event emissions remain unchanged.

* **Chores**
  * Updated license year and modernized header declarations for consistency.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -287,12 +287,11 @@ if ucx_dep.found() and cuda_dep.found() and nvcc_prog.found()
         add_project_arguments('-DHAVE_UCX_GPU_DEVICE_API', language: ['cpp', 'cuda'])
     endif
 
+    ucx_gpu_device_api_v2_available = false
     have_host_side_v2 = cpp.has_function('ucp_device_remote_mem_list_create', dependencies: ucx_dep)
     if have_gpu_side and have_host_side_v2
         ucx_gpu_device_api_v2_available = true
         add_project_arguments('-DHAVE_UCX_GPU_DEVICE_API_V2', language: ['cpp', 'cuda'])
-    else
-        ucx_gpu_device_api_v2_available = false
     endif
 
     summary({


src/core/telemetry/telemetry_event.h
```diff
@@ -14,19 +14,21 @@
...
+
 #include "nixl_types.h"
 
 constexpr char TELEMETRY_BUFFER_SIZE_VAR[] = "NIXL_TELEMETRY_BUFFER_SIZE";
 constexpr char TELEMETRY_RUN_INTERVAL_VAR[] = "NIXL_TELEMETRY_RUN_INTERVAL";
 
-constexpr int TELEMETRY_VERSION = 1;
-constexpr size_t MAX_EVENT_NAME_LEN = 32;
+constexpr inline int TELEMETRY_VERSION = 1;
+constexpr inline size_t MAX_EVENT_NAME_LEN = 32;
```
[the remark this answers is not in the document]
ColinNV: In C++17 `constexpr` implies `inline` for functions, but NOT for variables!

src/core/telemetry/telemetry.cpp
```diff
@@ -230,28 +230,22 @@ nixlTelemetry::updateMemoryDeregistered(uint64_t memory_deregistered) {
...
-                    .count();
-    std::lock_guard<std::mutex> lock(mutex_);
+    const char *bytes_name = is_write ? "agent_tx_bytes" : "agent_rx_bytes";
+    const char *requests_name = is_write ? "agent_tx_requests_num" : "agent_rx_requests_num";
+
+    const auto time = std::chrono::duration_cast<std::chrono::microseconds>(
+                          std::chrono::system_clock::now().time_since_epoch())
+                          .count();
+
+    const std::lock_guard lock(mutex_);
```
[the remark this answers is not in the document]
ColinNV: Everything should be `const` if possible, other code parts are irrelevant because this PR just does one thing, fix this function.

src/core/telemetry/telemetry_event.h
rakhmets, line 31: Maybe rename them since you are changing these lines.
At least, `MAX_EVENT_NAME_LEN`, since it is used only in this file.
```suggestion
constexpr inline int telemetry_version = 1;
constexpr inline size_t max_event_name_lane = 32;
```

ColinNV: You aren't wrong, but I'm trying to keep this focused. The missing `inline` is a mistake, which is why I'm fixing it anyhow.

src/core/telemetry/telemetry_event.h
```diff
@@ -57,18 +59,25 @@ struct nixlTelemetryEvent {
...
           category_(category),
           value_(value) {
-        strncpy(eventName_, event_name.c_str(), MAX_EVENT_NAME_LEN - 1);
+        strncpy(eventName_, event_name, MAX_EVENT_NAME_LEN - 1);
         eventName_[MAX_EVENT_NAME_LEN - 1] = '\0';
     }
+
+    nixlTelemetryEvent(uint64_t timestamp_us,
+                       nixl_telemetry_category_t category,
+                       const std::string &event_name,
```
iyastreb, line 78: maybe std::string_view event_name?

ColinNV: At the moment it's (nearly?) always being called with a `const std::string&` created from a C-string literal at the beginning of the call chain. I'd leave cleaning that up to a separate PR (and possibly *only* do `const char*`, it's not like this c'tor, at least as it stands now, uses the length.) Alternatively one could switch over to (only?) `std::string_view`.

PR #1370, opened 2026-02-26
INFRA: Refactor memory section details.

ColinNV:
## What?
Refactor memory section details to be/allow for more efficient, safer and cleaner code.

## How?
- Fix `mem_section.h`'s include guards.
- Use `std::unique_ptr` in `section_map_t`.
- Add a `const`-overload of `queryBackends()`.
- Bring some code in line with style guide.
- Declare some more local variables `const`.
- Change a function argument for an `enum` from by-reference to by-value.
- Use C++17 CTAD to simplify some variable initializations.
- Remove some misleading comments and add a clarification.
- Replace multiple related container operations with a single one.

## Summary by CodeRabbit

* **Refactor**
  * Switched to managed in-memory descriptor lists and value-based storage for safer lifetimes; constructor/destructor flows streamlined and serialization/load paths updated.

* **Bug Fixes**
  * Strengthened boundary checks and cleanup logic; backend-querying now has const-correct, noexcept variants for more predictable behavior.

* **Chores**
  * Header/include and guard updates and minor formatting adjustments.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -249,13 +249,12 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
 }
 
 namespace {
-nixl_status_t serializeSections(nixlSerDes* serializer,
-                                const section_map_t &sectionMap) {
-    size_t seg_count =
-        std::count_if(sectionMap.begin(), sectionMap.end(), [](const auto &pair) {
-            const section_key_t &sec_key = pair.first;
-            return sec_key.second->supportsRemote();
-        });
+nixl_status_t
+serializeSections(nixlSerDes *serializer, const section_map_t &sectionMap) {
+    size_t seg_count = std::count_if(sectionMap.begin(), sectionMap.end(), [](const auto &pair) {
+        const section_key_t &sec_key = pair.first;
+        return sec_key.second->supportsRemote();
+    });
 
     auto ret = serializer->addBuf("nixlSecElms", &seg_count, sizeof(seg_count));
     if (ret) {
@@ -341,7 +340,8 @@ nixlLocalSection::~nixlLocalSection() {
 
 /*** Class nixlRemoteSection implementation ***/
 
-nixlRemoteSection::nixlRemoteSection(std::string agent_name) noexcept : agentName(std::move(agent_name)) {}
+nixlRemoteSection::nixlRemoteSection(std::string agent_name) noexcept
+    : agentName(std::move(agent_name)) {}
 
 nixl_status_t nixlRemoteSection::addDescList (
                                  const nixl_reg_dlist_t& mem_elms,


src/infra/nixl_memory_section.cpp
```diff
@@ -97,7 +105,7 @@ nixlMemSection::addElement(const nixlRemoteDesc &query,
         return NIXL_ERR_NOT_FOUND;
     }
 
-    nixl_sec_dlist_t *base = it->second;
+    nixl_sec_dlist_t *base = it->second.get();
```
rakhmets: Maybe reference instead of pointer?

src/infra/nixl_memory_section.cpp
```diff
@@ -394,30 +406,30 @@ nixl_status_t nixlRemoteSection::loadLocalData (
...
-
-    if (sectionMap.count(sec_key) == 0) {
-        sectionMap[sec_key] = new nixl_sec_dlist_t(nixl_mem);
+    const nixl_mem_t nixl_mem = mem_elms.getType();
+    const section_key_t sec_key(nixl_mem, backend);
+    const auto [it, inserted] = sectionMap.try_emplace(sec_key);
+    if (inserted) {
+        it->second = std::make_unique<nixl_sec_dlist_t>(nixl_mem);
+        memToBackend[nixl_mem].insert(backend);
     }
```
rakhmets: The block is repeated three times. Can we get rid of duplicated code?

src/infra/mem_section.h
```diff
@@ -93,17 +95,22 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
 };
 
 using nixl_sec_dlist_t = nixlSecDescList;
-using section_map_t = std::map<section_key_t, nixl_sec_dlist_t*>;
+using section_map_t = std::map<section_key_t, std::unique_ptr<nixl_sec_dlist_t>>;
```
rakhmets: Maybe use `nixlSecDescList` instead of `nixl_sec_dlist_t` in modified code?
As the alias looks redundant, and it can be removed in the future with fewer changes.

src/infra/mem_section.h
```diff
@@ -93,17 +95,22 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
...
 
 class nixlMemSection {
     protected:
         std::array<backend_set_t, FILE_SEG+1>         memToBackend;
         section_map_t                                 sectionMap;
 
+        ~nixlMemSection() = default;
```
rakhmets, line 104: What is the reason to prohibit deletion via a pointer to `nixlMemSection`?

ColinNV: Base classes in C++ need to either have a `public virtual`, or a `protected` destructor. In this case the `public virtual` doesn't seem to be needed for polymorphic destruction, mostly/also because inheriting from `nixlMemSection` doesn't seem to model an "is-a" relationship. (So inheritance might not be the best tool here, but I didn't want to go that far.)

src/infra/nixl_memory_section.cpp
ColinNV: Or remove the intermediate when it's only used once...

src/infra/mem_section.h
rakhmets: Oh, right, this class doesn't even have virtual methods...
So, eventually, we should get rid of inheritance in this code.

src/infra/mem_section.h
```diff
@@ -93,17 +95,25 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
...
+        [[nodiscard]] nixlSecDescList &
+        emplaceBoth(nixl_mem_t nixl_mem, nixlBackendEngine *backend);
+
     public:
-        nixlMemSection () {};
+        nixlMemSection() = default;
 
-        backend_set_t* queryBackends (const nixl_mem_t &mem);
+        backend_set_t *
+        queryBackends(nixl_mem_t mem) noexcept;
```
rakhmets, line 113: Do we still need this overload?

src/infra/mem_section.h
ColinNV: (And even with virtual methods you can sometimes get away with a protected non-virtual destructor if you never delete polymorphically, for example if you only use `std::make_shared` for construction of the derived classes the `~shared_ptr` will directly call the derived class' destructor.)

src/infra/mem_section.h
ColinNV: It can't just be removed without changes to `nixl_agent.cpp`.

src/infra/mem_section.h
rakhmets: Right, that part I understand. I just didn't understand at first how this should improve usage for the `nixlMemSection` class.

src/infra/mem_section.h
```diff
@@ -93,17 +95,25 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
 };
 
 using nixl_sec_dlist_t = nixlSecDescList;
-using section_map_t = std::map<section_key_t, nixl_sec_dlist_t*>;
+using section_map_t = std::map<section_key_t, std::unique_ptr<nixlSecDescList>>;
```
iyastreb: Do you consider storing it by value?

src/infra/mem_section.h
```diff
@@ -93,17 +95,25 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
...
 
 class nixlMemSection {
     protected:
         std::array<backend_set_t, FILE_SEG+1>         memToBackend;
         section_map_t                                 sectionMap;
 
+        ~nixlMemSection() = default;
+
+        [[nodiscard]] nixlSecDescList &
+        emplaceBoth(nixl_mem_t nixl_mem, nixlBackendEngine *backend);
```
iyastreb: minor: maybe just `emplace`?

src/infra/nixl_memory_section.cpp
```diff
@@ -194,33 +210,38 @@ nixl_status_t nixlLocalSection::addDescList (const nixl_reg_dlist_t &mem_elms,
...
-        backend->deregisterMem((*target)[index].metadataP);
-        target->remDesc(index);
+        backend->deregisterMem(target[index].metadataP);
+        target.remDesc(index);
     }
 
-    if (target->descCount()==0) {
-        delete target;
-        sectionMap.erase(sec_key);
+    if (target.descCount() == 0) {
```
iyastreb: isEmpty

src/infra/nixl_memory_section.cpp
```diff
@@ -264,87 +285,80 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
...
                                                  const backend_set_t &backends,
                                                  const nixl_reg_dlist_t &mem_elms) const {
-    nixl_mem_t nixl_mem = mem_elms.getType();
+    const nixl_mem_t nixl_mem = mem_elms.getType();
     nixl_status_t ret = NIXL_SUCCESS;
     section_map_t mem_elms_to_serialize;
 
     // If there are no descriptors to serialize, just serialize empty list of sections
-    if (mem_elms.descCount() == 0)
+    if (mem_elms.descCount() == 0) {
```
iyastreb: isEmpty

src/infra/nixl_memory_section.cpp
```diff
@@ -264,87 +285,80 @@ nixl_status_t nixlLocalSection::serialize(nixlSerDes* serializer) const {
...
     }
-    // nixlMemSection destructor will clean up the rest
 }
 
 /*** Class nixlRemoteSection implementation ***/
 
-nixlRemoteSection::nixlRemoteSection (const std::string &agent_name) {
-    this->agentName = agent_name;
-}
+nixlRemoteSection::nixlRemoteSection(const std::string &agent_name) : agentName(agent_name) {}
```
iyastreb: can accept by value and std::move

src/infra/nixl_memory_section.cpp
```diff
@@ -390,34 +404,28 @@ nixl_status_t nixlRemoteSection::loadRemoteData (nixlSerDes* deserializer,
...
 }
 
-nixl_status_t nixlRemoteSection::loadLocalData (
-                                 const nixl_sec_dlist_t& mem_elms,
-                                 nixlBackendEngine* backend) {
+nixl_status_t
+nixlRemoteSection::loadLocalData(const nixlSecDescList &mem_elms, nixlBackendEngine *backend) {
 
-    if (mem_elms.descCount()==0) // Shouldn't happen
+    if (mem_elms.descCount() == 0) { // Shouldn't happen
```
iyastreb: isEmpty

src/infra/mem_section.h
ColinNV: I didn't, but you're right, this ain't Java.

PR #1375, opened 2026-02-27
CORE/AGENT: Refactor.

ColinNV:
## What?
Refactor agent, mostly making memory handling safer and simpler.

## Why?
Class `nixlAgentData` was doing a lot of error-prone manual memory management. And `nixlAgent` was not destroying backend engines in a consistent way.

## How?
- Declare two of `nixlAgentData` members `const`.
- Rename `nixlAgentData::memorySection` to `nixlAgentData::localSection` (in line with `nixlAgentData::remoteSections`).
- Replace all owning plain pointers in `nixlAgentData`'s members with `std::unique_ptr` or hold by-value.
- Re-order `nixlAgentData`'s data members so that the C++ destruction order matches the required destruction order.
- Now `nixlAgentData`'s destructor is superfluous and could be removed.
- Consistently destroy backend engines via their respective plugin.
- Optimize some container access patterns.
- A couple of other very minor refactorings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer lookups, stronger synchronization, and defensive checks to reduce races, leaks, and error paths.

* **New Features**
  * Telemetry now collects and publishes request/transfer statistics for better observability.
  * Backend lifecycle now integrates with plugins for more reliable engine teardown.

* **Chores / Refactor**
  * Ownership moved to smart-pointer semantics and local memory handling centralized for clearer lifecycle management and stability.

### DIFF
--- a/.ci/jenkins/pipeline/proj-jjb.yaml
+++ b/.ci/jenkins/pipeline/proj-jjb.yaml
@@ -301,7 +301,7 @@
       - string:
           name: "NIXL_VERSION"
           default: "{jjb_branch}"
-          description: "NIXL version to use (tag like 1.0.0, branch name, or commit hash)"
+          description: "NIXL version to use (tag like 1.0.1, branch name, or commit hash)"
       - string:
           name: "UCX_VERSION"
           default: "v1.20.x"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,7 +190,7 @@ checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
 
 [[package]]
 name = "nixl-sys"
-version = "1.0.0"
+version = "1.0.1"
 dependencies = [
  "anyhow",
  "bincode",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
 resolver = "3"
 
 [workspace.package]
-version = "1.0.0"
+version = "1.0.1"
 edition = "2021"
 description = "Low-level bindings to NIXL - NVIDIA Inference Xfer Library"
 authors = ["NIXL Developers <nixl-developers@nvidia.com>"]
--- a/benchmark/nixlbench/meson.build
+++ b/benchmark/nixlbench/meson.build
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project('nixlbench', 'CPP', version: '1.0.0',
+project('nixlbench', 'CPP', version: '1.0.1',
     default_options: ['buildtype=release',
                 'werror=true',
                 'cpp_std=c++17',
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -65,6 +65,7 @@ RUN apt-get update -y && \
     clang \
     hwloc \
     libhwloc-dev \
+    pkg-config \
     # aws-sdk-cpp dependencies
     libcurl4-openssl-dev libssl-dev uuid-dev zlib1g-dev \
     # azure-sdk-for-cpp dependencies
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -138,7 +138,7 @@ void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_rdma_byte
     _nixl_ep_init();
 }
 
-Buffer::~Buffer() noexcept(false) {
+Buffer::~Buffer() noexcept {
     if (not explicitly_destroy) {
         destroy();
     } else if (not destroyed) {
@@ -167,27 +167,58 @@ torch::Stream Buffer::get_comm_stream() const {
 }
 
 void Buffer::destroy() {
-    EP_HOST_ASSERT(not destroyed);
+    auto warn_cuda = [](cudaError_t status, const char* operation) noexcept {
+        if (status != cudaSuccess) {
+            std::cerr << "WARNING: destroy() failed to " << operation
+                      << ": " << cudaGetErrorString(status) << '\n';
+        }
+    };
+
+    auto warn_nixl = [](nixl_status_t status, const char* operation) noexcept {
+        if (status != NIXL_SUCCESS) {
+            std::cerr << "WARNING: destroy() failed to " << operation
+                      << ": " << nixlEnumStrings::statusStr(status) << '\n';
+        }
+    };
+
+    if (destroyed) {
+        return;
+    }
 
     // Synchronize
-    CUDA_CHECK(cudaDeviceSynchronize());
+    warn_cuda(cudaDeviceSynchronize(), "synchronize device");
 
     _nixl_ep_destroy();
 
-    cudaFree(rdma_buffer_ptr);
+    if (nixl_agent_info and nixl_agent_info->agent != nullptr) {
+        if (getenv("NIXL_ETCD_ENDPOINTS")) {
+            warn_nixl(nixl_agent_info->agent->invalidateLocalMD(),
+                      "invalidate local metadata");
+        }
 
-    if (nixl_agent_info and nixl_agent_info->agent != nullptr and getenv("NIXL_ETCD_ENDPOINTS")) {
-        nixl_agent_info->agent->invalidateLocalMD();
+        warn_nixl(nixl_agent_info->agent->deregisterMem(
+                      nixl_agent_info->rdma_reg_descs,
+                      &nixl_agent_info->extra_params),
+                  "deregister RDMA memory");
+        warn_nixl(nixl_agent_info->agent->deregisterMem(
+                      nixl_agent_info->sync_reg_descs,
+                      &nixl_agent_info->extra_params),
+                  "deregister sync memory");
+        warn_nixl(nixl_agent_info->agent->deregisterMem(
+                      nixl_agent_info->sync_count_reg_descs,
+                      &nixl_agent_info->extra_params),
+                  "deregister sync-count memory");
+
+        nixl_agent_info.reset();
     }
 
-    rdma_buffer_ptr = nullptr;
-
-    cudaFree(mask_buffer_ptr);
-    cudaFree(sync_buffer_ptr);
-    cudaFree(sync_count_ptr);
+    warn_cuda(cudaFree(rdma_buffer_ptr), "free RDMA buffer");
+    warn_cuda(cudaFree(mask_buffer_ptr), "free mask buffer");
+    warn_cuda(cudaFree(sync_buffer_ptr), "free sync buffer");
+    warn_cuda(cudaFree(sync_count_ptr), "free sync-count buffer");
 
     // Free workspace
-    CUDA_CHECK(cudaFree(workspace));
+    warn_cuda(cudaFree(workspace), "free workspace");
 
     destroyed = true;
     available = false;
@@ -318,8 +349,6 @@ void Buffer::disconnect_ranks(const std::vector<int>& remote_ranks_list) {
 
     _nixl_ep_memory_views_destroy();
 
-    _nixl_ep_memory_views_create();
-
     _nixl_agents_peer_info_cleanup(remote_ranks_list);
 
     _nixl_agents_disconnect(remote_ranks_list);
@@ -338,6 +367,8 @@ void Buffer::disconnect_ranks(const std::vector<int>& remote_ranks_list) {
                            *std::max_element(remote_ranks.begin(), remote_ranks.end()));
     }
     num_ranks = max_rank + 1;  // Sparse indexing maintained
+
+    _nixl_ep_memory_views_create();
 }
 
 std::tuple<torch::Tensor, std::optional<torch::Tensor>, torch::Tensor, torch::Tensor, torch::Tensor, std::optional<EventHandle>, std::optional<std::function<void()>>>
@@ -619,8 +650,10 @@ void Buffer::_nixl_ep_memory_views_create(void) {
     }
 
     EP_HOST_ASSERT(nixl_agent_info->agent->prepMemView(local_descs, gpu_ctx.local_mvh, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
-    EP_HOST_ASSERT(nixl_agent_info->agent->prepMemView(remote_descs, gpu_ctx.remote_mvh, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
-    EP_HOST_ASSERT(nixl_agent_info->agent->prepMemView(barrier_descs, gpu_ctx.barrier_mvh, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
+    if (!remote_ranks.empty()) {
+        EP_HOST_ASSERT(nixl_agent_info->agent->prepMemView(remote_descs, gpu_ctx.remote_mvh, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
+        EP_HOST_ASSERT(nixl_agent_info->agent->prepMemView(barrier_descs, gpu_ctx.barrier_mvh, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
+    }
 }
 
 void Buffer::_nixl_ep_memory_views_destroy(void) {
@@ -681,19 +714,21 @@ void Buffer::_nixl_agent_init() {
     nixl_agent_info->extra_params.backends.push_back(ucx_backend);
     nixl_agent_info->agent_name = agent_name;
 
-    /* Register RDMA buffer */
-    nixl_reg_dlist_t rdma_ptr_dlist(VRAM_SEG);
-    rdma_ptr_dlist.addDesc(nixlBlobDesc((uintptr_t)(rdma_buffer_ptr), num_rdma_bytes, get_local_device_id(), ""));
-    EP_HOST_ASSERT(agent->registerMem(rdma_ptr_dlist) == NIXL_SUCCESS);
+    nixl_agent_info->rdma_reg_descs.clear();
+    nixl_agent_info->rdma_reg_descs.addDesc(
+        nixlBlobDesc(reinterpret_cast<uintptr_t>(rdma_buffer_ptr), num_rdma_bytes, device_id, ""));
 
-    /* Register sync buffer */
-    nixl_reg_dlist_t sync_dlist(VRAM_SEG);
-    sync_dlist.addDesc(nixlBlobDesc((uintptr_t)(sync_buffer_ptr), max_num_ranks * sizeof(int), get_local_device_id(), ""));
-    EP_HOST_ASSERT(agent->registerMem(sync_dlist) == NIXL_SUCCESS);
+    nixl_agent_info->sync_reg_descs.clear();
+    nixl_agent_info->sync_reg_descs.addDesc(
+        nixlBlobDesc(reinterpret_cast<uintptr_t>(sync_buffer_ptr), max_num_ranks * sizeof(int), device_id, ""));
 
-    nixl_reg_dlist_t barrier_cnt_dlist(VRAM_SEG);
-    barrier_cnt_dlist.addDesc(nixlBlobDesc((uintptr_t)(sync_count_ptr), max_num_ranks * sizeof(int), get_local_device_id(), ""));
-    EP_HOST_ASSERT(agent->registerMem(barrier_cnt_dlist) == NIXL_SUCCESS);
+    nixl_agent_info->sync_count_reg_descs.clear();
+    nixl_agent_info->sync_count_reg_descs.addDesc(
+        nixlBlobDesc(reinterpret_cast<uintptr_t>(sync_count_ptr), max_num_ranks * sizeof(int), device_id, ""));
+
+    EP_HOST_ASSERT(agent->registerMem(nixl_agent_info->rdma_reg_descs, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
+    EP_HOST_ASSERT(agent->registerMem(nixl_agent_info->sync_reg_descs, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
+    EP_HOST_ASSERT(agent->registerMem(nixl_agent_info->sync_count_reg_descs, &nixl_agent_info->extra_params) == NIXL_SUCCESS);
 
     if (getenv("NIXL_ETCD_ENDPOINTS")) {
         status = nixl_agent_info->agent->sendLocalMD();
--- a/examples/device/ep/csrc/nixl_ep.hpp
+++ b/examples/device/ep/csrc/nixl_ep.hpp
@@ -68,6 +68,9 @@ struct NixlAgentInfo
     std::vector<std::string> remote_agent_names;
     nixl_opt_args_t extra_params;
     nixlBackendH* backend;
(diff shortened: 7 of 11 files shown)


src/core/agent_data.h
```diff
@@ -105,9 +96,21 @@ class nixlAgentData {
         std::atomic<bool> commThreadStop;
         std::atomic<bool> agentShutdown;
         bool useEtcd;
-        std::unique_ptr<nixlTelemetry> telemetry_;
         std::exception_ptr commThreadException_;
 
+        // The order of the following data members is crucial for destruction.
+        // Bookkeping for local connection metadata and user handles per backend
```
iyastreb:
```suggestion
        // Bookkeeping for local connection metadata and user handles per backend
```

src/core/agent_data.h
```diff
@@ -105,9 +96,21 @@ class nixlAgentData {
...
+        std::unordered_map<nixl_backend_t, std::unique_ptr<nixlBackendH>> backendHandles;
+        std::unordered_map<nixl_backend_t, nixl_blob_t> connMD;
+        backend_map_t backendEngines;
+        std::unordered_map<std::string,
+                           std::unique_ptr<nixlRemoteSection>,
```
iyastreb: pls consider storing by value

src/infra/mem_section.h
```diff
@@ -31,7 +31,14 @@
...
 using section_key_t = std::pair<nixl_mem_t, nixlBackendEngine*>;
 using backend_set_t = std::set<nixlBackendEngine*>;
-using backend_map_t = std::unordered_map<nixl_backend_t, nixlBackendEngine*>;
+
+struct nixlEngineDeleter {
+    void
+    operator()(nixlBackendEngine *) const noexcept;
+};
+
+using backend_ptr_t = std::unique_ptr<nixlBackendEngine, nixlEngineDeleter>;
```
iyastreb, line 40: just wondering why do we need a deleter?
dtor is not enough?

I mean, nixlBackendPlugin already keeps a pointer to destroy_engine, so you can just call it from ~nixlBackendPlugin

src/infra/nixl_memory_section.cpp
```diff
@@ -383,27 +383,36 @@ nixl_status_t nixlRemoteSection::addDescList (
     return NIXL_SUCCESS;
 }
 
-nixl_status_t nixlRemoteSection::loadRemoteData (nixlSerDes* deserializer,
-                                                 backend_map_t &backendToEngineMap) {
+nixl_status_t
+nixlRemoteSection::loadRemoteData(nixlSerDes *deserializer, backend_map_t &backendToEngineMap) {
```
iyastreb, line 387: if deserialiser is not supposed to be nullptr, we can pass it by ref
(maybe in a separate PR)

src/infra/mem_section.h
iyastreb: Maybe destroy_engine should accept engine by ref, in order to avoid any memory management
`void (*destroy_engine)(nixlBackendEngine &engine);`
Also it can be null, if plugin does not do anything useful in destroy_engine (=if it just deletes engine)

ColinNV: I would also think that the d'tor is enough and/or the engine should be handled with a `std::shared_ptr` so that the creator can set up any additional destruction without it being visible to the outside world. The deleter was implemented to keep strict compatibility with the `createEngine()` and `destroyEngine()` interface as in theory an out-of-tree plugin could do more than `delete` in `destroyEngine()`.

src/core/agent_data.h
ColinNV: I considered it but then wasn't sure about stability, an insert into the map can invalidate iterators, but not the pointers. Will check again whether it's a problem.

src/infra/nixl_memory_section.cpp
ColinNV: Might be a strict compatibility problem doing that.

src/core/agent_data.h
ColinNV: Done

src/core/nixl_agent.cpp
```diff
@@ -1187,7 +1163,7 @@ nixl_status_t
 nixlAgent::queryXferBackend(const nixlXferReqH* req_hndl,
                             nixlBackendH* &backend) const {
     NIXL_LOCK_GUARD(data->lock);
-    backend = data->backendHandles[req_hndl->engine->getType()];
+    backend = data->backendHandles[req_hndl->engine->getType()].get();
     return NIXL_SUCCESS;
```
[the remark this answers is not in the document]
ColinNV: This PR doesn't want to change the logic in any way, only the memory handling and other minor refactoring not changing the semantics (and I'm assuming this function will only be called for engine types that are already in the backend handle map).

src/core/agent_data.h
```diff
@@ -146,14 +152,29 @@ class nixlBackendH {
 
         explicit nixlBackendH(nixlBackendEngine *engine) noexcept : engine(engine) {}
 
+    public:
         ~nixlBackendH() = default;
 
-    public:
-        nixl_backend_t getType () const { return engine->getType(); }
+        // TODO? engine->getType() returns a const nixl_backend_t&
```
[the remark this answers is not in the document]
ColinNV: This change is out-of-scope for this PR, the TODO is for the future.

src/core/agent_data.h
```diff
@@ -146,14 +152,29 @@ class nixlBackendH {
...
-        bool supportsNotif  () const { return engine->supportsNotif (); }
+        bool
+        supportsLocal() const {
+            return engine->supportsLocal();
+        }
+
+        bool
+        supportsNotif() const {
+            return engine->supportsNotif();
+        }
```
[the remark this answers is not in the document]
ColinNV: This type needs a more substantial overhaul, until then I won't touch it more than necessary.

src/core/agent_data.h
```diff
@@ -102,9 +93,18 @@ class nixlAgentData {
...
 
+        // The order of the following data members is crucial for destruction.
+        // Bookkeeping for local connection metadata and user handles per backend
+        std::unordered_map<nixl_backend_t, std::unique_ptr<nixlBackendH>> backendHandles;
+        std::unordered_map<nixl_backend_t, nixl_blob_t> connMD;
+        backend_map_t backendEngines;
+        std::unordered_map<std::string, nixlRemoteSection, std::hash<std::string>, strEqual>
+            remoteSections;
+        std::unique_ptr<nixlTelemetry> telemetry_;
+        nixlLocalSection localSection;
```
rakhmets: The comment https://github.com/ai-dynamo/nixl/pull/1375#discussion_r2871727838 suggests to rename private fields.
I think, at least, `localSection` should be changed, since this is a new name, and will not introduce additional changed lines to the PR.

src/core/agent_data.h
```diff
@@ -102,9 +93,18 @@ class nixlAgentData {
...
         bool useEtcd;
-        std::unique_ptr<nixlTelemetry> telemetry_;
         std::exception_ptr commThreadException_;
 
+        // The order of the following data members is crucial for destruction.
+        // Bookkeeping for local connection metadata and user handles per backend
+        std::unordered_map<nixl_backend_t, std::unique_ptr<nixlBackendH>> backendHandles;
+        std::unordered_map<nixl_backend_t, nixl_blob_t> connMD;
+        backend_map_t backendEngines;
+        std::unordered_map<std::string, nixlRemoteSection, std::hash<std::string>, strEqual>
```
rakhmets: What do you think about using just `std::unordered_map<std::string, nixlRemoteSection>`?

src/core/agent_data.h
ColinNV: I don't like the inconsistent naming within a single class, but sure, can do.

rakhmets: Thanks,
I asked this, because we have already combined two types of names in this class. And I'd prefer to follow our code style for new fields in this case.

src/core/agent_data.h
ColinNV: Short answer, I'm all for it. Long answer, I didn't do it because I wanted to leave performance optimizations to a separate PR. What I mean is, if this data structure is performance critical it should be properly optimized, for which I have several ideas, but would need to know the usual key metrics (typical/average size, maximum size, insertion frequency, lookup frequency). If you think we can - permanently or temporarily - get away with what you're suggesting, I'm all for it. @iyastreb @brminich ?

src/core/agent_data.h
ColinNV: Fair enough, give me a minute :-)

src/core/agent_data.h
rakhmets: I think you're right that we can leave this type unchanged within this PR, and we can keep using custom comparison function.
And I think we can remove `strEqual`, since I don't think it gives any noticeable increase (perhaps in some cases it makes it worse) in performance. But it's probably better to change this in a separate request along with `remoteBackends`.

rakhmets (dismissed):

src/core/agent_data.h
iyastreb: I think it's fine to remove strEqual in a separate PR. I believe this string equals is not even used in 99% of the cases, because hash code for string is good enough.

I don't think it's very performance critical, because we do the lookup twice per entire operation (which may contain 1000x of requests).

src/core/nixl_agent.cpp
```diff
@@ -328,66 +312,63 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
+        data->connMD[type] = conn_info;
+    }
+
+    // TODO: Simplify this e.g. by making nixlBackendH's c'tor public?
+    std::unique_ptr<nixlBackendH> bknd_temp(new nixlBackendH(backend.get()));
+    const auto [it, inserted] = data->backendHandles.insert_or_assign(type, std::move(bknd_temp));
+    // NIXL_ASSERT(inserted);
+    bknd_hndl = it->second.get();
+
+    data->backendEngines.try_emplace(type, std::move(backend));
```
iyastreb: I'm not 100% sure about the logic here
What happens in a case if emplace fails? We already inserted raw pointer into several structures by this moment

ColinNV: The only reason it can fail is out-of-memory, and in that case there's not much we can do anyhow, which is presumably why the old version of this code also didn't bother to leave things consistent in that case.

iyastreb: Yes, I see, we assert in the beginning that there is no such key in the map yet.
So the only thing we can do is maybe insert it first into the map and in case of a success propagate raw pointer to other places?

ColinNV: If we need to be fully consistent in out-of-memory situations that should be a separate PR that generally fixes it?

src/core/agent_data.h
ColinNV: #1391

src/core/nixl_agent.cpp
iyastreb: ok

src/core/agent_data.h
```diff
@@ -63,8 +63,8 @@ using nixl_socket_map_t = std::map<nixl_socket_peer_t, int>;
 
 class nixlAgentData {
     private:
-        std::string     name;
-        nixlAgentConfig config;
+        const std::string name;
+        const nixlAgentConfig config;
```
michal-shalev: Minor: according to NIXL's code style, private class data members should use a trailing underscore suffix

src/core/agent_data.h
```diff
@@ -99,9 +90,17 @@ class nixlAgentData {
...
         bool useEtcd;
-        std::unique_ptr<nixlTelemetry> telemetry_;
         std::exception_ptr commThreadException_;
 
+        // The order of the following data members is crucial for destruction.
+        // Bookkeeping for local connection metadata and user handles per backend
+        std::unordered_map<nixl_backend_t, std::unique_ptr<nixlBackendH>> backendHandles;
+        std::unordered_map<nixl_backend_t, nixl_blob_t> connMD;
+        backend_map_t backendEngines;
+        std::unordered_map<std::string, nixlRemoteSection> remoteSections;
```
michal-shalev: Minor: according to NIXL's code style, private class data members should use a trailing underscore suffix

src/core/agent_data.h
ColinNV: I wanted to leave that to a separate PR that focuses on style; these two were only made `const`, not a far-reaching change.

michal-shalev:
Nice cleanup, the RAII migration is well done, especially removing the 5 manual delete calls on error paths in `createBackend`.

src/core/nixl_agent.cpp
```diff
@@ -1712,27 +1685,32 @@ nixl_status_t
...
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (descs.isEmpty()) {
+        return NIXL_SUCCESS;
+    }
+
+    nixl_meta_dlist_t dummy(descs.getType());
+    // We only add to data->remoteBackends if data->backendEngines[backend] exists
+    for (const auto &[backend, conn_info] : data->remoteBackends[remote_name]) {
```
michal-shalev: Maybe use find() instead of operator[] on two maps?

ColinNV: I agree, but didn't want to make this change in this PR since it can't be locally seen that it doesn't change semantics.

src/core/agent_data.h
ColinNV: Done.

src/core/nixl_agent.cpp
michal-shalev: okay

src/core/agent_data.h
michal-shalev: I don't really mind if it's fixe here or in a separate PR

src/core/agent_data.h
michal-shalev: I don't really mind if it's fixe here or in a separate PR

src/infra/nixl_memory_section.cpp
```diff
@@ -383,27 +383,36 @@ nixl_status_t nixlRemoteSection::addDescList (
...
             return NIXL_ERR_NOT_FOUND;
-        if (backendToEngineMap.count(nixl_backend) != 0) {
-            ret = addDescList(s_desc, backendToEngineMap[nixl_backend]);
-            if (ret) return ret;
+        }
+
+        const auto it = backendToEngineMap.find(nixl_backend);
+        if (it != backendToEngineMap.end()) {
+            ret = addDescList(s_desc, it->second.get());
+            if (ret) {
```
brminich: can we change it as we touch this code anyway?

```suggestion
            if (ret != NIXL_SUCCESS) {
```

src/infra/nixl_memory_section.cpp
```diff
@@ -383,27 +383,36 @@ nixl_status_t nixlRemoteSection::addDescList (
...
-                                                 backend_map_t &backendToEngineMap) {
+nixl_status_t
+nixlRemoteSection::loadRemoteData(nixlSerDes *deserializer, backend_map_t &backendToEngineMap) {
     nixl_status_t ret;
     size_t seg_count;
     nixl_backend_t nixl_backend;
 
     ret = deserializer->getBuf("nixlSecElms", &seg_count, sizeof(seg_count));
-    if (ret) return ret;
+    if (ret) {
```
brminich: can we change it as we touch this code anyway?

```suggestion
    if (ret != NIXL_SUCCESS) {
```

src/core/nixl_listener.cpp
```diff
@@ -744,22 +745,22 @@ nixl_status_t
...
+    if (backendEngines_.count(backend) == 0) {
+        NIXL_DEBUG << "Agent " << name_ << " does not support a remote backend: " << backend;
         return NIXL_ERR_NOT_SUPPORTED;
     }
 
     // No need to reload same conn info, error if it changed
-    if ((remoteBackends.count(remote_name) != 0) &&
-        (remoteBackends[remote_name].count(backend) != 0)) {
-        if (remoteBackends[remote_name][backend] != conn_info) {
+    if ((remoteBackends_.count(remote_name) != 0) &&
```
tvegas1: pretty minor: find() end() at(), maybe not easier to read though?

src/core/nixl_agent.cpp
```diff
@@ -324,66 +307,63 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
 
-    if (backend->supportsRemote()) data->notifEngines.push_back(backend);
+    if (backend->supportsRemote()) {
+        data->notifEngines.push_back(backend.get());
+        data->connMd_[type] = conn_info;
+    }
+
+    // TODO: Simplify this e.g. by making nixlBackendH's c'tor public?
+    std::unique_ptr<nixlBackendH> bknd_temp(new nixlBackendH(backend.get()));
+    const auto [it, inserted] = data->backendHandles_.insert_or_assign(type, std::move(bknd_temp));
```
tvegas1: would below work? maybe it's not improving that much..
```c++
// Create and insert or replace the backend handle
data->backendHandles_[type] = std::unique_ptr<nixlBackendH>(new nixlBackendH(backend.get()));
bknd_hndl = data->backendHandles_[type].get();

**Review by tvegas1 — APPROVED** [review]

File: src/core/nixl_agent.cpp (hunk 17 above)
Comment (ColinNV) [reply src/core/nixl_agent.cpp]: That's two container lookups, and while `std::unordered_map` has O(1) average case operations (if the hashing behaves) they are still slower than a `std::vector`.

Comment (tvegas1) [reply src/core/nixl_agent.cpp]: `bknd_hndl = (data->backendHandles_[type] = std::unique_ptr<nixlBackendH>(new nixlBackendH(backend.get()))).get();`? anyways does not matter.

Comment (ColinNV) [reply src/core/nixl_agent.cpp]: That would work, but is a bit harder to parse.

File: src/core/nixl_agent.cpp [hunk 18]
```diff
@@ -1712,27 +1684,32 @@ nixl_status_t
...
+        return NIXL_ERR_NOT_FOUND;
+    }
+
+    if (descs.isEmpty()) {
+        return NIXL_SUCCESS;
+    }
+
+    nixl_meta_dlist_t dummy(descs.getType());
+    // We only add to data->remoteBackends_ if data->backendEngines_[backend] exists
+    for (const auto &[backend, conn_info] : data->remoteBackends_[remote_name]) {
```
guy-ealey-morag, line 1707: accessing `data->remoteBackends_[remote_name]` creates an empty entry if `remote_name` does not exist in the map.
Should we first check if it exists before we iterate?

src/core/nixl_agent.cpp
```diff
@@ -324,66 +307,63 @@ nixlAgent::createBackend(const nixl_backend_t &type,
...
-    if (backend->supportsRemote()) data->notifEngines.push_back(backend);
+    if (backend->supportsRemote()) {
+        data->notifEngines.push_back(backend.get());
+        data->connMd_[type] = conn_info;
+    }
+
+    // TODO: Simplify this e.g. by making nixlBackendH's c'tor public?
+    std::unique_ptr<nixlBackendH> bknd_temp(new nixlBackendH(backend.get()));
+    const auto [it, inserted] = data->backendHandles_.insert_or_assign(type, std::move(bknd_temp));
+    // NIXL_ASSERT(inserted);
```
guy-ealey-morag: why is this in a comment?

src/core/nixl_agent.cpp
guy-ealey-morag: It is already promised that `type` key does not exist in the map, so I suggest using `emplace` or `try_emplace`

src/core/nixl_listener.cpp
```diff
@@ -574,7 +574,8 @@ nixlAgentData::commWorkerInternal(nixlAgent *myAgent) {
                     const std::string &metadata_label = req_ip;
 
                     // Use local storeMetadataInEtcd function
-                    nixl_status_t ret = etcdClient->storeMetadataInEtcd(name, metadata_label, my_MD);
+                    nixl_status_t ret =
```
guy-ealey-morag:
```suggestion
                    const nixl_status_t ret =
```

src/core/nixl_listener.cpp
```diff
@@ -619,7 +620,7 @@ nixlAgentData::commWorkerInternal(nixlAgent *myAgent) {
                         throw std::runtime_error("ETCD is not enabled");
                     }
 
-                    nixl_status_t ret = etcdClient->removeMetadataFromEtcd(name);
+                    nixl_status_t ret = etcdClient->removeMetadataFromEtcd(name_);
```
guy-ealey-morag:
```suggestion
                    const nixl_status_t ret = etcdClient->removeMetadataFromEtcd(name_);
```

src/infra/nixl_memory_section.cpp
```diff
@@ -383,27 +383,36 @@ nixl_status_t nixlRemoteSection::addDescList (
     return NIXL_SUCCESS;
 }
 
-nixl_status_t nixlRemoteSection::loadRemoteData (nixlSerDes* deserializer,
-                                                 backend_map_t &backendToEngineMap) {
+nixl_status_t
+nixlRemoteSection::loadRemoteData(nixlSerDes *deserializer, backend_map_t &backendToEngineMap) {
     nixl_status_t ret;
     size_t seg_count;
     nixl_backend_t nixl_backend;
```
guy-ealey-morag: Do we prefer moving declarations to the place of first use in C++?

src/core/nixl_agent.cpp
```diff
@@ -583,19 +560,19 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
...
-            backend_set = data->remoteSections[agent_name]->
-                                queryBackends(descs.getType());
-        else
-            backend_set = data->memorySection->
-                                queryBackends(descs.getType());
+        if (init_side) {
+            backend_set = data->localSection_.queryBackends(descs.getType());
+        } else {
+            backend_set = rem_sec_it->second.queryBackends(descs.getType());
+        }
```
guy-ealey-morag: Small refactoring suggestion:
```
const nixlMemSection &section = init_side ?
        static_cast<const nixlMemSection &>(data->localSection_) :
        static_cast<const nixlMemSection &>(rem_sec_it->second);
```

We can then use `section` to call `queryBackends`, and later `populate` without the `if (init_side)`.
The `static_cast<const nixlMemSection &>` is a bit ugly but it won't compile without it.

src/core/nixl_agent.cpp
ColinNV: As documentation.

src/core/nixl_agent.cpp
ColinNV: Presumably this check function can't (or shouldn't) be called  without a previous different function having created the entry. Ideally we could switch to `at()` separately.

src/infra/nixl_memory_section.cpp
ColinNV: Yes, in general it's better style in C++, in this PR I didn't change all places where it might be done to not make it explode in size.

src/core/nixl_agent.cpp
ColinNV: It's not explicitly checked, but over-arching logic says yes.

src/core/nixl_agent.cpp
ColinNV: While we're at it...

src/core/nixl_listener.cpp
ColinNV: ATM there are three lookups in `remoteBackends_`, that can be improved.

tvegas1 (approved):

PR #1376, opened 2026-02-27
API V1: cleanups

ovidiusm:
## What?

This PR introduces a V1 API cleanup with backward compatibility preserved.

- Clean up `nixl_opt_args_t` usage:
  - Deprecate `includeConnInfo` (kept functional for compatibility).
  - Deprecate `skipDescMerge` (kept functional for compatibility; no new API is introduced to encourage use).
  - Add `std::optional<nixl_blob_t> notif` as the preferred notification field. Keep legacy `notifMsg` + `hasNotif` supported as fallback for compatibility.
- Simplify `nixlAgentConfig` construction:
  - Make config fields public.
  - Enable explicit field-based initialization (instead of long positional constructor usage).
  - Keep legacy constructor overloads for compatibility.
- Improve ergonomics:
  - Add `prepXferDlist` overload that omits `agent_name` for local preparation.
- Add API versioning:
  - Introduce a typed API version scheme and set the current version to `V1`.

## Why?

These changes reduce API friction and ambiguity, improve readability at call sites, and prepare the API surface for V1 stabilization without breaking existing users.

Deprecation rationale:

- `includeConnInfo` was deprecated because it is effectively redundant:
  - connection info can be fetched by itself seperately by passing an empty descriptor list,
  - it’s rarely/never used in practice,
  - keeping it in a global optional-args struct adds API noise for little value.

- `skipDescMerge` was deprecated because it did not deliver practical benefit and can hurt runtime behavior:
  - no meaningful benefit after defragmentation,
  - it's enabled by default while the required computation adds overhead,
  - it complicates request/status handling paths,
  - better optimization decisions belong in backend internals (e.g., UCX) rather than exposed as a generic user-facing knob.

## Summary by CodeRabbit

* **New Features**
  * Public API version constant exposed; added overload to prepare transfer descriptor lists without specifying an agent.
  * Python bindings: module API version constant and no-argument agent config constructor available.

* **Refactor**
  * Agent configuration converted to a defaultable, field-configurable structure with sensible defaults and public fields.
  * Notification inputs consolidated from separate message/flag fields into a single unified notification field across APIs and bindings.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -37,17 +37,6 @@ static const std::vector<std::vector<std::string>> illegal_plugin_combinations =
     {"GDS", "GDS_MT"},
 };
 
-inline void
-setNotification(const nixl_opt_args_t &extra_params, nixl_opt_b_args_t &opt_args) {
-    if (extra_params.notif) {
-        opt_args.notifMsg = *extra_params.notif;
-        opt_args.hasNotif = true;
-    } else if (extra_params.hasNotif) {
-        opt_args.notifMsg = extra_params.notifMsg;
-        opt_args.hasNotif = true;
-    }
-}
-
 /*** nixlEnumStrings namespace implementation in API ***/
 std::string nixlEnumStrings::memTypeStr(const nixl_mem_t &mem) {
     static std::array<std::string, FILE_SEG+1> nixl_mem_str = {
@@ -765,7 +754,13 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
     }
 
     if (extra_params) {
-        setNotification(*extra_params, opt_args);
+        if (extra_params->notif) {
+            opt_args.notifMsg = *extra_params->notif;
+            opt_args.hasNotif = true;
+        } else if (extra_params->hasNotif) {
+            opt_args.notifMsg = extra_params->notifMsg;
+            opt_args.hasNotif = true;
+        }
     }
 
     if ((opt_args.hasNotif) && (!backend->supportsNotif())) {
@@ -951,7 +946,14 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
     }
 
     if (extra_params) {
-        setNotification(*extra_params, opt_args);
+        if (extra_params->notif) {
+            opt_args.notifMsg = *extra_params->notif;
+            opt_args.hasNotif = true;
+        } else if (extra_params->hasNotif) {
+            opt_args.notifMsg = extra_params->notifMsg;
+            opt_args.hasNotif = true;
+        }
+
         if (extra_params->customParam.length() > 0)
             opt_args.customParam = extra_params->customParam;
     }
@@ -1084,11 +1086,19 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
 
     // Updating the notification based on opt_args
     if (extra_params) {
-        opt_args.hasNotif = false;
-        setNotification(*extra_params, opt_args);
-        req_hndl->hasNotif = opt_args.hasNotif;
-        if (opt_args.hasNotif) {
-            req_hndl->notifMsg = opt_args.notifMsg;
+        if (extra_params->notif) {
+            req_hndl->notifMsg = *extra_params->notif;
+            opt_args.notifMsg = *extra_params->notif;
+            req_hndl->hasNotif = true;
+            opt_args.hasNotif = true;
+        } else if (extra_params->hasNotif) {
+            req_hndl->notifMsg = extra_params->notifMsg;
+            opt_args.notifMsg = extra_params->notifMsg;
+            req_hndl->hasNotif = true;
+            opt_args.hasNotif = true;
+        } else {
+            req_hndl->hasNotif = false;
+            opt_args.hasNotif = false;
         }
     }
 


src/api/cpp/nixl_params.h
```diff
@@ -27,39 +27,52 @@
  *        Other configs such as assigned IP/port or device access can be added.
  */
 class nixlAgentConfig {
-    private:
+    public:
```
rakhmets: Maybe just `struct nixlAgentConfig`.

src/api/cpp/nixl_params.h
```diff
@@ -22,90 +22,98 @@
...
+                    const bool capture_telemetry = kDefaultCaptureTelemetry,
+                    const std::chrono::microseconds &etcd_watch_timeout = kDefaultEtcdWatchTimeout)
+        : useProgThread(use_prog_thread),
+          useListenThread(use_listen_thread),
+          listenPort(port),
+          syncMode(sync_mode),
+          captureTelemetry(capture_telemetry),
+          pthrDelay(pthr_delay_us),
+          lthrDelay(lthr_delay_us),
+          etcdWatchTimeout(etcd_watch_timeout) {}
```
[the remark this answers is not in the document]
ovidiusm: This is a backward incompatible change, won't fix

src/bindings/python/nixl_bindings.cpp
```diff
@@ -570,8 +600,7 @@ PYBIND11_MODULE(_bindings, m) {
                     extra_params.backends.push_back((nixlBackendH *)backend);
 
                 if (notif_msg.size() > 0) {
-                    extra_params.notifMsg = notif_msg;
-                    extra_params.hasNotif = true;
+                    extra_params.notif = notif_msg;
                 }
```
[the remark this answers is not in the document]
ovidiusm: This would not be backward-compatible. Moreover, there is little benefit in sending an empty notification message.

src/api/cpp/nixl.h
```diff
@@ -184,6 +192,19 @@ class nixlAgent {
                        const nixl_xfer_dlist_t &descs,
                        nixlDlistH* &dlist_hndl,
                        const nixl_opt_args_t* extra_params = nullptr) const;
+        /**
+         * @brief  Prepare a local descriptor list for transfer requests.
+         *         Equivalent to prepXferDlist(NIXL_INIT_AGENT, ...).
```
mkhazraee: I say we don't need this line, and eventually we deprecate the other path.

src/api/cpp/nixl_types.h
```diff
@@ -132,6 +132,19 @@ using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
...
+ * @brief Public NIXL API version tags.
+ */
+enum class nixl_api_version : unsigned int {
+    V1 = 1,
+};
+
+/**
+ * @brief Alias for the supported public API version.
+ */
+inline constexpr nixl_api_version nixlApiVersionV1 = nixl_api_version::V1;
```
rakhmets: IMO, we don't really need this, `nixl_api_version::V1` is enough.

src/api/cpp/nixl_params.h
```diff
@@ -22,90 +22,98 @@
...
-         * @brief Default destructor for nixlAgentConfig
-         */
-        ~nixlAgentConfig () = default;
-
-    friend class nixlAgent;
-    friend class nixlAgentData;
+    /**
+     * @brief Default destructor for nixlAgentConfig
+     */
+    ~nixlAgentConfig() = default;
```
rakhmets: It seems like it's redundant. Can we remove this?

src/api/cpp/nixl_params.h
```diff
@@ -22,90 +22,98 @@
...
+     * @param use_listen_thread  Optional flag to determine use of listener thread
+     * @param port               Optional port for listener thread to listen on
+     * @param sync_mode          Optional Thread synchronization mode
+     * @param num_workers        Optional number of shared workers per backend
+     * @param pthr_delay_us      Optional delay for pthread in us
+     * @param lthr_delay_us      Optional delay for listener thread in us
+     * @param capture_telemetry  Optional flag to enable telemetry capture
+     * @param etcd_watch_timeout Optional timeout for etcd watch operations in microseconds
+     */
+    nixlAgentConfig(const bool use_prog_thread,
```
rakhmets: I think it would be better to add `explicit` and `noexcept`.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -109,7 +109,9 @@ xferBenchNixlWorker::xferBenchNixlWorker(int *argc, char ***argv, std::vector<st
 
     rank = rt->getRank();
 
-    nixlAgentConfig dev_meta(enable_pt, false, 0, sync_mode);
+    nixlAgentConfig dev_meta{};
```
rakhmets:
```suggestion
    nixlAgentConfig dev_meta;
```

examples/cpp/nixl_etcd_example.cpp
```diff
@@ -48,7 +48,8 @@ nixlAgent* createAgent(const std::string& name) {
         setenv("NIXL_ETCD_ENDPOINTS", ETCD_ENDPOINT.c_str(), 1);
     }
 
-    nixlAgentConfig cfg(true);
+    nixlAgentConfig cfg{};
+    cfg.useProgThread = true;
```
rakhmets: Why do we need this change?
Are we going to deprecate non-default constructor?

src/core/nixl_agent.cpp
```diff
@@ -656,6 +656,13 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
     }
 }
 
+nixl_status_t
+nixlAgent::prepXferDlist(const nixl_xfer_dlist_t &descs,
+                         nixlDlistH *&dlist_hndl,
+                         const nixl_opt_args_t *extra_params) const {
+    return prepXferDlist(NIXL_INIT_AGENT, descs, dlist_hndl, extra_params);
```
rakhmets, line 659: This one-liner can be placed in `.h`
And I'm also not entirely sure if this has sufficient reasons to be added to the API.

src/core/nixl_agent.cpp
```diff
@@ -750,9 +757,14 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
         total_bytes += (*local_descs)[local_indices[i]].len;
     }
 
-    if (extra_params && extra_params->hasNotif) {
-        opt_args.notifMsg = extra_params->notifMsg;
-        opt_args.hasNotif = true;
+    if (extra_params) {
+        if (extra_params->notif.has_value()) {
+            opt_args.notifMsg = extra_params->notif.value();
```
```suggestion
        if (extra_params->notif) {
            opt_args.notifMsg = *extra_params->notif;
```

src/core/nixl_agent.cpp
```diff
@@ -938,7 +950,10 @@ nixlAgent::createXferReq(const nixl_xfer_op_t &operation,
...
 
     if (extra_params) {
-        if (extra_params->hasNotif) {
+        if (extra_params->notif.has_value()) {
+            opt_args.notifMsg = extra_params->notif.value();
+            opt_args.hasNotif = true;
+        } else if (extra_params->hasNotif) {
             opt_args.notifMsg = extra_params->notifMsg;
             opt_args.hasNotif = true;
         }
```
rakhmets, line 955: Please add a function to avoid code duplication.

examples/cpp/nixl_etcd_example.cpp
ovidiusm: This is preferred usage.

src/api/cpp/nixl_types.h
ovidiusm: Removed

src/core/nixl_agent.cpp
ovidiusm: @mkhazraee This was an explicit request. Not sure about putting in the header.

src/core/nixl_agent.cpp
ovidiusm: fixed

src/core/nixl_agent.cpp
ovidiusm: This code is performance sensitive, it is written this way intentionally.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
ovidiusm: Replaced everywhere

src/api/cpp/nixl_params.h
ovidiusm: done

src/api/cpp/nixl_params.h
ovidiusm: done

src/core/nixl_agent.cpp
mkhazraee: In early days of nixl, using INITIATOR_AGENT got some confusion, specially with loopback, while python version with empty name was easier to understand. There were long discussions and the plan was to add this mode of operation when extra_params were removed and we added overloading, but recently we realized that would be ugly (removing extra_params, as 3 parameters alone would be 8 overloads, and we have more for some methods). Also realized the main benefit of that overloading was actually this very use case, so just adding this overload.

mkhazraee:
Overall LGTM, other than passing the tests, and one minor nitpicks: since we're deprecating includeConnInfo, let's remove its mentions from the nixl.h. Also inc_conn_info from _api.h (or mention to be deprecated), and also skip_desc_merge in _api.h (didn't see skipMerge in nixl.h anyways).

src/core/nixl_agent.cpp
rakhmets: If this code is performance-sensitive, then why do we add another conditional branch to this code?
And do you mean that adding the following function will hurt performance?
```C++
setNotification(const nixl_opt_args_t* extra_params) {
    if (extra_params) {
        if (extra_params->notif) {
            notifMsg = *extra_params->notif;
            hasNotif = true;
        } else if (extra_params->hasNotif) {
            notifMsg = extra_params->notifMsg;
            hasNotif = true;
        }
    }
}
```
I think the function (which is likely to be optimized/inlined by the compiler) brings zero performance penalty comparing to extra string copy (from extra_params to opt_args), and to extra `if` branch.

ovidiusm: But this is not about the extra branch, but adding another function for replacing 2 statements. It’s not guaranteed that it’s going to be inlined

rakhmets: I cannot agree that code duplication is a good code optimization pattern. In addition, code duplication contradicts several statements of the C++ core guidelines and rules of clean code.

ovidiusm: The end goal is to be able to deprecate the old parameters, if we add a wrapper for them it would be harder to do that because there will never be a need to do it

ovidiusm: Added helper function

src/api/cpp/nixl.h
```diff
@@ -37,6 +37,14 @@ class nixlAgent {
         const std::unique_ptr<nixlAgentData> data;
 
     public:
+        /**
+         * @brief Get the public NIXL API version tag.
+         */
+        static constexpr nixl_api_version
```
brminich: im not sure we need it. Maybe returning NIXL versions would be better? Like `major.minor`
Or maybe let's not introduce such API in this release

ovidiusm: Since we are not sure what will be needed, let's remove it

src/core/nixl_agent.cpp
ovidiusm: I reverted the helper function addition, because there was significant performance impact in the CI run for GPU tests that was reproducible (TestTransfer and TestErrorHandling were much slower). After https://github.com/ai-dynamo/nixl/pull/1376/changes/3e22ce676c5b6eefd15199b26f8c07ff08974220 it got back to normal

rakhmets: I think the problem was in `__builtin_expect`. If you add `__builtin_expect` to your current code, you will also see slow downs in tests that pass non-nullptr extra_params.

ovidiusm: It's possible, I am running some benchmarks now to be sure

ovidiusm: Avg Prep time doubles consistently with a helper function. This is nixlbench tests, not gtest. I tried with expect, no expect, with reference and with pointer, it is consistently worse. I will revert it.

mkhazraee:
We can remove the comments about skipMerge and connInfo in nixl.h and _api.py when we actually deprecate them, so good to go as is.

PR #1388, opened 2026-03-03
nixlbench/gusli: add --gusli_try_use_uring configurable option

omer-b1:
Add --gusli_try_use_uring flag to enable the io_uring engine in the GUSLI backend. Can be passed via CLI or --config_file TOML.

When set, gusli_backend will call io.params.set_try_use_uring(true) on each IO request. README updated with the new parameter.

## Summary by CodeRabbit

* **New Features**
  * Added gusli_try_use_uring option to attempt using io_uring for the GUSLI backend (default: disabled; configurable via CLI or config file).

* **Documentation**
  * Usage and parameter lists updated to document the new option.

* **Behavior**
  * The selected "try use uring" setting is logged during backend initialization and propagated through Gusli I/O paths.

* **Build**
  * Gusli build configuration updated to enable io_uring support by default.


aranadive:
@omer-b1 Please check dockerfiles and scripts to see where we build GUSLI and set ALLOW_USE_URING to 1. This will build GUSLI with uring support. Have your tried this already with this patch? What performance difference do you see with uring?

omer-b1:
> @omer-b1 Please check dockerfiles and scripts to see where we build GUSLI and set ALLOW_USE_URING to 1. This will build GUSLI with uring support. Have your tried this already with this patch? What performance difference do you see with uring?

yes i saw, indeed build must include ALLOW_USE_URING=1, but this just makes it compile with io uring but does not affect the io flow (still uses libaio):
`		#if defined(HAS_URING_LIB)
			if (!params.has_callback() && params.may_use_uring())	// Uring does not support async callback mode
				_exec = new uring_request_executor(bdev->b.dp->in_air, *this);
		#endif
`
according to following comment in the code i understand it is deliberately not enabled by default (even with compilation flag):
`uint16_t _try_using_uring_api : 1;		// If possible use uring api, more efficient for io's with large amount of ranges, but does not run in default containers of kubernetes due to security issues of lib-uring`

my new changes now allow to run it with io uring when the flag is used.
i did only a few tests so far and i was able to see slight improvement.

aranadive:
> > @omer-b1 Please check dockerfiles and scripts to see where we build GUSLI and set ALLOW_USE_URING to 1. This will build GUSLI with uring support. Have your tried this already with this patch? What performance difference do you see with uring?
> 
> yes i saw, indeed build must include ALLOW_USE_URING=1, but this just makes it compile with io uring but does not affect the io flow (still uses libaio): ` #if defined(HAS_URING_LIB) if (!params.has_callback() && params.may_use_uring()) // Uring does not support async callback mode _exec = new uring_request_executor(bdev->b.dp->in_air, *this); #endif ` according to following comment in the code i understand it is deliberately not enabled by default (even with compilation flag): `uint16_t _try_using_uring_api : 1; // If possible use uring api, more efficient for io's with large amount of ranges, but does not run in default containers of kubernetes due to security issues of lib-uring`
> 
> my new changes now allow to run it with io uring when the flag is used. i did only a few tests so far and i was able to see slight improvement.

@omer-b1 I agree on the nixl gusli backend changes but if GUSLI library itself is not compiled with IOURING support, the backend parameter will be ignored and it will fall back to libaio. Im asking to enable the GUSLI library build with the ALLOW_USE_URING flag set to 1 in the places where we build GUSLI library.

omer-b1:
> asking to enable the GUSLI library build with the ALLOW_USE_URING flag set to 1 in the places where we build GUSLI library.

ok so to clarify, as i understand it isnt supposed to be mandatory. the GUSLI library is externally built, not part of building the nixl/nixlbench so while building nixl it cant impact how GUSLI library is built.
this flag is called "try_use_uring" - meaning if indeed GUSLI is compiled with URING it will work.
I can suggest to print a warning in case the user uses this new parameter but the actuall GUSLI library is compiled without the flag. what do you think?

aranadive:
> > asking to enable the GUSLI library build with the ALLOW_USE_URING flag set to 1 in the places where we build GUSLI library.
> 
> ok so to clarify, as i understand it isnt supposed to be mandatory. the GUSLI library is externally built, not part of building the nixl/nixlbench so while building nixl it cant impact how GUSLI library is built. this flag is called "try_use_uring" - meaning if indeed GUSLI is compiled with URING it will work. I can suggest to print a warning in case the user uses this new parameter but the actuall GUSLI library is compiled without the flag. what do you think?

GUSLI is external but is also built and distributed via our Python wheel, using our various Dockerfiles (e.g., nixlbench) for compilation and testing. To enable URING support for nixl, it should first be configured in the Dockerfiles so the backend parameter consistency is maintained, i.e. works if set to true, disabled otherwise.

omer-b1:
> > > asking to enable the GUSLI library build with the ALLOW_USE_URING flag set to 1 in the places where we build GUSLI library.
> > 
> > 
> > ok so to clarify, as i understand it isnt supposed to be mandatory. the GUSLI library is externally built, not part of building the nixl/nixlbench so while building nixl it cant impact how GUSLI library is built. this flag is called "try_use_uring" - meaning if indeed GUSLI is compiled with URING it will work. I can suggest to print a warning in case the user uses this new parameter but the actuall GUSLI library is compiled without the flag. what do you think?
> 
> GUSLI is external but is also built and distributed via our Python wheel, using our various Dockerfiles (e.g., nixlbench) for compilation and testing. To enable URING support for nixl, it should first be configured in the Dockerfiles so the backend parameter consistency is maintained, i.e. works if set to true, disabled otherwise.

ok. done.
thanks

aranadive (dismissed):

aranadive (approved):

PR #1390, opened 2026-03-03
CORE: Refactor transfer_request.h.

ColinNV:
## What?
Refactor the two classes in transfer_request.h.

## Why?
Primarily to use smart instead of plain pointers, secondarily to clean up and adapt to new coding style.

## Summary by CodeRabbit

* **Refactor**
  * Transfer-request objects now require explicit parameters at construction and own their descriptor lists, clarifying ownership.
  * Descriptor lists use move semantics and are stored as owned containers to simplify lifecycle and error cleanup.
  * Telemetry update paths now accept non-owning telemetry pointers for stats reporting.
  * Transfer preparation and error paths updated to align with new ownership and telemetry semantics.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -1115,7 +1115,7 @@ xferBenchUtils::printStats(bool is_target,
 
     total_data_transferred = ((block_size * batch_size) * num_iter); // In Bytes
     avg_latency = (total_duration / (num_iter * batch_size)); // In microsec
-    if (IS_PAIRWISE_AND_MG()) {
+    if (IS_PAIRWISE_AND_MG() || (IS_PAIRWISE_AND_SG() && xferBenchConfig::num_initiator_dev > 1)) {
         total_data_transferred *= xferBenchConfig::num_initiator_dev; // In Bytes
         avg_latency /= xferBenchConfig::num_initiator_dev; // In microsec
     }


src/core/nixl_agent.cpp
```diff
@@ -78,7 +78,7 @@ nixlEnumStrings::statusStr(const nixl_status_t &status) {
 }
 
 inline void
-nixlXferReqH::updateRequestStats(std::unique_ptr<nixlTelemetry> &telemetry_pub,
+nixlXferReqH::updateRequestStats(const std::unique_ptr<nixlTelemetry> &telemetry_pub,
```
iyastreb: minor: maybe better pass it by pointer `const nixlTelemetry *`

src/core/transfer_request.h
```diff
@@ -34,54 +34,63 @@ enum nixl_telemetry_stat_status_t {
...
 
-        void
-        updateRequestStats(std::unique_ptr<nixlTelemetry> &telemetry,
-                           nixl_telemetry_stat_status_t stat_status);
+    void
+    updateRequestStats(const std::unique_ptr<nixlTelemetry> &telemetry,
+                       nixl_telemetry_stat_status_t stat_status);
 
-        friend class nixlAgent;
+    friend class nixlAgent;
```
iyastreb, line 54: Since we redesign this class, we can get rid of this friend. Friends are abused in many places for no good reason.
Ideally we should initialize all the fields in ctor/factory/setters:
```
nixlBackendEngine &engine; // ref, not ptr

nixlXferReqH(nixlBackendEngine &engine, std::unique_ptr<nixl_meta_dlist_t> initiatorDescs, std::unique_ptr<nixl_meta_dlist_t> targetDescs)

File: src/core/transfer_request.h [hunk 3]
```diff
@@ -34,54 +34,63 @@ enum nixl_telemetry_stat_status_t {
...
-        }
+    nixlDlistH(nixlDlistH &&) = delete;
+    nixlDlistH(const nixlDlistH &) = delete;
+
+    void
+    operator=(nixlDlistH &&) = delete;
+    void
+    operator=(const nixlDlistH &) = delete;
 
     friend class nixlAgent;
```
iyastreb: same here, can we get rid of friend and implement proper ctor and getters?

src/core/nixl_agent.cpp
ColinNV: Can do, `nixlTelemetry*` it is.

src/core/transfer_request.h
ColinNV: I'm not a fan of `friend` either and would be happy to get rid of it. If we open access whith getters (and, if needed, setters) can we just go `public`?

Independently I'll look into using the c'tor for initialization and making some fields `const`.

src/core/transfer_request.h
ColinNV: This one has fewer data members and is easier, the c'tor is trivial, only `descs` needs to be non-`const`, and again, if we have a getter to mutable `descs` we might as well make everything `public`.

src/core/transfer_request.h
iyastreb: With public access anyone can modify the objects, which I think is not desirable. I would propose to do it with least possible privilege.
Ideally we need an immutable object that is created once and then accessed (and one nice property is that they are thread safe by design).
What fields are actually mutable after initialization? I think only status, the rest are immutable

src/core/transfer_request.h
ColinNV: Widening the net might make the `public` unnecessary, let me check.

iyastreb: Why do you think descs are non-const? I think they are const
IIUC descs are populated here:
```
    for (auto & backend : *backend_set) {
        handle->descs[backend] = new nixl_meta_dlist_t(descs.getType());
        if (init_side)
            ret = data->memorySection->populate(
                       descs, backend, *(handle->descs[backend]));
        else
            ret = data->remoteSections[agent_name]->populate(
                       descs, backend, *(handle->descs[backend]));
        if (ret == NIXL_SUCCESS) {
            count++;
        } else {
            delete handle->descs[backend];
            handle->descs.erase(backend);
        }
    }

```
We just need change the construction approach: first populate descs object, and then properly construct nixlDList by moving descs - after that descs remains immutable. That's a classic dependency injection approach

ColinNV: It worked out for `nixlDlistH`.

src/core/transfer_request.h
ColinNV: It's a bit more involved here compared to `nixlDistH`, leaving it for a separate PR.

src/core/transfer_request.h
```diff
@@ -34,56 +35,65 @@ enum nixl_telemetry_stat_status_t {
...
+    updateRequestStats(nixlTelemetry *telemetry, nixl_telemetry_stat_status_t stat_status);
 
-        friend class nixlAgent;
+    friend class nixlAgent;
 };
 
-class nixlDlistH {
-    private:
-        std::unordered_map<nixlBackendEngine*, nixl_meta_dlist_t*> descs;
+struct nixlDlistH {
```
rakhmets, line 73: You can use a conventional class member declaration order as you changed all the lines in the class.
And you can do the same for `nixlXferReqH`.

src/core/nixl_agent.cpp
```diff
@@ -105,6 +115,11 @@ nixlXferReqH::updateRequestStats(std::unique_ptr<nixlTelemetry> &telemetry_pub,
                << duration.count() << "us.";
 }
 
+nixlDlistH::nixlDlistH(const bool is_local, const std::string &remote_agent, descs_t &&descs)
```
rakhmets: Maybe align the order of the parameters with the order of class members?

src/core/nixl_agent.cpp
```diff
@@ -610,46 +623,34 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
...
 
-    nixlDlistH *handle = new nixlDlistH;
-    if (init_side) {
-        handle->isLocal     = true;
-        handle->remoteAgent = "";
-    } else {
-        handle->isLocal     = false;
-        handle->remoteAgent = agent_name;
-    }
+    std::unordered_map<nixlBackendEngine *, std::unique_ptr<nixl_meta_dlist_t>> dlists;
```
rakhmets, line 647: `descs_t`?

src/core/nixl_agent.cpp
```diff
@@ -610,46 +623,34 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
...
-        else
-            ret = data->remoteSections[agent_name]->populate(
-                       descs, backend, *(handle->descs[backend]));
-        if (ret == NIXL_SUCCESS) {
-            count++;
-        } else {
-            delete handle->descs[backend];
-            handle->descs.erase(backend);
+    auto &mem_sec = init_side ? static_cast<nixlMemSection &>(*data->memorySection) :
+                                static_cast<nixlMemSection &>(*data->remoteSections[agent_name]);
```
rakhmets, line 650: `static_cast<nixlMemSection &>(init_side ? *data->memorySection : *data->remoteSections[agent_name])`?

src/core/nixl_agent.cpp
```diff
@@ -716,8 +717,8 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
         return NIXL_ERR_INVALID_PARAM;
     }
 
-    nixl_meta_dlist_t* local_descs  = local_side->descs.at(backend);
-    nixl_meta_dlist_t* remote_descs = remote_side->descs.at(backend);
+    nixl_meta_dlist_t &local_descs = *local_side->descs.at(backend);
+    nixl_meta_dlist_t &remote_descs = *remote_side->descs.at(backend);
```
rakhmets: Can they be cons?

src/core/nixl_agent.cpp
```diff
@@ -757,43 +758,42 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
         return NIXL_ERR_BACKEND;
     }
 
-    std::unique_ptr<nixlXferReqH> handle = std::make_unique<nixlXferReqH>();
-    handle->initiatorDescs = new nixl_meta_dlist_t(local_descs->getType(), desc_count);
-
-    handle->targetDescs = new nixl_meta_dlist_t(remote_descs->getType(), desc_count);
+    std::unique_ptr<nixlXferReqH> handle = std::make_unique<nixlXferReqH>(remote_side->remoteAgent,
```
rakhmets: I think you can use `auto` when `make_unique` is on the right side.

src/core/transfer_request.h
```diff
@@ -34,56 +35,65 @@ enum nixl_telemetry_stat_status_t {
...
-        nixl_xfer_telem_t telemetry;
+    nixlXferReqH(nixlXferReqH &&) = delete;
+    nixlXferReqH(const nixlXferReqH &) = delete;
 
-    public:
-        inline nixlXferReqH() { }
+    void
+    operator=(nixlXferReqH &&) = delete;
+    void
+    operator=(const nixlXferReqH &) = delete;
```
rakhmets: This is implied by the presence of constant unique pointers.

src/core/nixl_agent.cpp
ColinNV: Doesn't work, the two branches are different types (that both derive from `nixlMemSection`).

src/core/nixl_agent.cpp
ColinNV: I think it's useful to spell it out here. (It would be `nixlDlistH::descs_t`.)

src/core/nixl_agent.cpp
ColinNV: Ok.

src/core/transfer_request.h
```diff
@@ -34,56 +35,49 @@ enum nixl_telemetry_stat_status_t {
...
+    nixlDlistH(const std::string &remote_agent, const bool is_local, descs_t &&descs);
 
-        inline ~nixlDlistH() {
-            for (auto & elm : descs)
-                delete elm.second;
-        }
-
-    friend class nixlAgent;
+    const std::string remoteAgent;
+    const bool isLocal;
```
rakhmets: The presence of these two fields at the same time is redundant.
First, I wanted to recommend using std::optional.
However, I think it can be done even easier. As:
```C++
std::string remoteAgent = init_side ? std::string() : agent_name;
bool isLocal = init_side;
```
, where
```C++
bool init_side = agent_name == NIXL_INIT_AGENT;
```
, where
```C++
#define NIXL_INIT_AGENT ""
```
So, I think we can just store `agent_name` in `remoteAgent`.

ColinNV: I wasn't quite sure about this, sometimes writing things out explicitly is good even if they're technically redundant. But I suppose it's in the DNA of NIXL that an agent name is never empty? I'll make the change.

src/core/nixl_agent.cpp
```diff
@@ -105,6 +115,10 @@ nixlXferReqH::updateRequestStats(std::unique_ptr<nixlTelemetry> &telemetry_pub,
                << duration.count() << "us.";
 }
 
+nixlDlistH::nixlDlistH(const std::string &remote_agent, descs_t &&descs)
```
iyastreb, line 119: minor: I think classic way is to pass by value and then move

iyastreb (approved):

src/core/nixl_agent.cpp
ColinNV: It's only called once, and not with an r-value reference.

src/core/transfer_request.h
```diff
@@ -34,56 +35,48 @@ enum nixl_telemetry_stat_status_t {
...
 
-        nixl_xfer_telem_t telemetry;
+private:
+    nixlBackendEngine *engine = nullptr;
+    nixlBackendReqH *backendHandle = nullptr;
 
-    public:
-        inline nixlXferReqH() { }
+    const std::unique_ptr<nixl_meta_dlist_t> initiatorDescs;
+    const std::unique_ptr<nixl_meta_dlist_t> targetDescs;
```
michal-shalev, line 61: Minor: according to NIXL's code style, private class data members should use a trailing underscore suffix

ColinNV: It's not really a class in that sense.

brminich (approved):

PR #1410, opened 2026-03-06
PLUGINS/UCX: Timeout warning for device memory list creation.

rakhmets:
## What?
Added a warning message in case the time limit for creating the device memory list is exceeded.
Applied logic from the Device API V1 https://github.com/ai-dynamo/nixl/pull/947.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for device memory setup with a configurable warning threshold (default 5s); a single warning is emitted once when exceeded and retries continue with backoff.
  * Reduced CPU usage during memory list creation by yielding when progress stalls.
  * Standardized and clarified failure messages for device memory list creation across local and remote paths.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -1115,7 +1115,7 @@ xferBenchUtils::printStats(bool is_target,
 
     total_data_transferred = ((block_size * batch_size) * num_iter); // In Bytes
     avg_latency = (total_duration / (num_iter * batch_size)); // In microsec
-    if (IS_PAIRWISE_AND_MG()) {
+    if (IS_PAIRWISE_AND_MG() || (IS_PAIRWISE_AND_SG() && xferBenchConfig::num_initiator_dev > 1)) {
         total_data_transferred *= xferBenchConfig::num_initiator_dev; // In Bytes
         avg_latency /= xferBenchConfig::num_initiator_dev; // In microsec
     }
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -78,8 +78,18 @@ nixlEnumStrings::statusStr(const nixl_status_t &status) {
     }
 }
 
-inline void
-nixlXferReqH::updateRequestStats(std::unique_ptr<nixlTelemetry> &telemetry_pub,
+nixlXferReqH::nixlXferReqH(const std::string &remote_agent,
+                           const nixl_xfer_op_t backend_op,
+                           const nixl_mem_t local_type,
+                           const nixl_mem_t remote_type,
+                           const size_t desc_count)
+    : initiatorDescs(std::make_unique<nixl_meta_dlist_t>(local_type, desc_count)),
+      targetDescs(std::make_unique<nixl_meta_dlist_t>(remote_type, desc_count)),
+      remoteAgent(remote_agent),
+      backendOp(backend_op) {}
+
+void
+nixlXferReqH::updateRequestStats(nixlTelemetry *telemetry_pub,
                                  nixl_telemetry_stat_status_t stat_status) {
 
     static const std::array<std::string, 3> nixl_post_status_str = {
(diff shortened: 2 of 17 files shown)


src/plugins/ucx/mem_list.cpp
```diff
@@ -18,21 +18,29 @@
...
 #include <exception>
+#endif
 #include <string>
 #include <string_view>
-#endif
 #include <stdexcept>
 
 #ifdef HAVE_UCX_GPU_DEVICE_API
+namespace {
+constexpr std::string_view error_message{"Failed to create device memory list"};
```
ColinNV: I'd make ths a `std::string` here since that's what it's used as below.

src/plugins/ucx/mem_list.cpp
```diff
@@ -141,19 +149,33 @@ createElements(const T &dlist, size_t worker_id = 0) {
...
 void *
 createMemList(const nixl_remote_meta_dlist_t &dlist, size_t worker_id, nixlUcxWorker &worker) {
+    using namespace std::chrono_literals;
+
     const device_mem_vector_t elements = createElements(dlist, worker_id);
     const memListParams params{elements};
 
     ucp_device_remote_mem_list_h handle{nullptr};
     ucs_status_t status;
+    bool check_timeout{true};
```
ColinNV: Since unified initialization didn't work out in C++ I'd use `check_timeout = true` for the standard cases.

src/plugins/ucx/mem_list.cpp
```diff
@@ -141,19 +149,33 @@ createElements(const T &dlist, size_t worker_id = 0) {
...
+    const auto timeout = nixl::config::getValueDefaulted("NIXL_UCX_TIMEOUT_WARNING", 5'000ms);
+
+    const auto start = std::chrono::steady_clock::now();
     while ((status = ucp_device_remote_mem_list_create(params.get(), &handle)) ==
            UCS_ERR_NOT_CONNECTED) {
-        worker.progress();
+        if (check_timeout && ((std::chrono::steady_clock::now() - start) > timeout)) {
+            NIXL_WARN << "Still waiting to create device memory list after " << timeout.count()
+                      << " ms; retrying";
+            check_timeout = false;
```
ColinNV: Increase the timeout to occasionally warn again?

ColinNV (dismissed):

src/plugins/ucx/mem_list.cpp
rakhmets: Changed to `std::string`.

src/plugins/ucx/mem_list.cpp
rakhmets: changed to `= true`.

src/plugins/ucx/mem_list.cpp
rakhmets: Previously it was agreed, that a single warning is enough, and I have just restored previous behavior.

rakhmets: Added a timeout increase to print warning again.

src/plugins/ucx/mem_list.cpp
```diff
@@ -141,19 +148,33 @@ createElements(const T &dlist, size_t worker_id = 0) {
...
 createMemList(const nixl_remote_meta_dlist_t &dlist, size_t worker_id, nixlUcxWorker &worker) {
+    using namespace std::chrono_literals;
+
     const device_mem_vector_t elements = createElements(dlist, worker_id);
     const memListParams params{elements};
 
     ucp_device_remote_mem_list_h handle{nullptr};
     ucs_status_t status;
+    const auto timeout_warning =
+        nixl::config::getValueDefaulted("NIXL_UCX_TIMEOUT_WARNING", 5'000ms);
```
ColinNV: The name should be changed, better `NIXL_UCX_WARNING_TIMEOUT`, or something else that puts `TIMEOUT` last.

rakhmets: Renamed.

PR #1422, opened 2026-03-13
BINDINGS/RUST: Remove `/usr/include` from Rust bindings build include path

mbrobbel:
## What?

Removes `/usr/include` from the include path for the Rust bindings build.

## Why?

Remove the hard-coded `/usr/include` from the Rust binding build. System headers should come from the compiler’s implicit include paths or configured sysroot, not from a forced host path. This makes the build more portable and avoids accidentally mixing host headers into sysrooted environments, while being redundant on normal toolchains.

## Summary by CodeRabbit

* **Chores**
  * Removed reliance on the system-wide include path during native component builds.
  * Header resolution now uses configured project and NIXL include directories.
  * Updated copyright year information in build metadata.
  * No user-facing functionality or runtime behavior has changed.


tomerg-nvidia (approved):
LGTM

roiedanino (approved):

brminich (approved):

mbrobbel:
@brminich it looks like some required check are not starting. Is there anything I can do to unblock this PR?

PR #1425, opened 2026-03-16
telemetry: lock-free double-buffer, skip-when-full

BatshevaBlack:
Replaces the previous mutex-protected single event queue with a **lock-free double buffer** for telemetry events. One buffer is written by producers (agent threads calling `updateData` / `addXferTime`); the periodic write task drains the other buffer and exports events. When the active buffer reaches capacity it is marked full; producers skip appending until the next drain (no blocking, no mutex).

- **Lower latency on the hot path:** Telemetry is in the data path; a mutex in `updateData`/`addXferTime` can add contention and tail latency under load.
- **Avoid blocking producers:** Under heavy load, if the consumer is slow, the old design could block all producers. This design keeps producers non-blocking: they either append or skip.

Trade-off: under sustained overload, some events are dropped instead of backing up and blocking.

- **Two fixed buffers** (`event_buffers_[2]`): at any time, producers write to the buffer indexed by `write_buffer_index_`; the write task drains the other buffer.
- **Lock-free coordination:** `write_buffer_index_`, `write_idx_`, and `full_` are atomics. Producers use `fetch_add` on `write_idx_` to reserve a slot; if the buffer is already full they return without blocking.
- **Drain step:** In `writeEventHelper()`, the consumer sets `full_` (blocking new writes to the current buffer), swaps the active buffer with `write_buffer_index_`, resets `write_idx_` for the new buffer, drains the old buffer to the exporter, then (conceptually) allows writes again to the new buffer.
- **Skip-when-full:** In `updateData`/`addXferTime`, if `full_.load()` is true or the reserved index is past the buffer size, the call returns immediately and the event is dropped.

## Summary by CodeRabbit

* **Refactor**
  * Telemetry redesigned for lower contention and higher throughput using lock-free double-buffering and batched writes.

* **New Features**
  * Added a consolidated “transfer complete” telemetry event API and a public recording flag to control telemetry state.

* **Bug Fixes / Behavior**
  * Telemetry is only active when an exporter is configured; bookkeeping and per-transfer reporting are gated and will warn when disabled.

* **Behavior Change**
  * Per-event timestamps removed from telemetry events and reader output.

### DIFF
--- a/examples/python/telemetry_reader.py
+++ b/examples/python/telemetry_reader.py
@@ -23,7 +23,6 @@ import os
 import signal
 import sys
 import time
-from datetime import datetime
 
 # Set up logging
 logging.basicConfig(
@@ -34,8 +33,7 @@ logging.basicConfig(
 logger = logging.getLogger(__name__)
 
 # Constants from telemetry_event.h
-TELEMETRY_VERSION = 1
-MAX_EVENT_NAME_LEN = 32
+TELEMETRY_VERSION = 2
 
 # NIXL telemetry categories
 NIXL_TELEMETRY_MEMORY = 0
@@ -47,6 +45,18 @@ NIXL_TELEMETRY_PERFORMANCE = 5
 NIXL_TELEMETRY_SYSTEM = 6
 NIXL_TELEMETRY_CUSTOM = 7
 NIXL_TELEMETRY_MAX = 8
+
+# NIXL telemetry event names (nixl_telemetry_event_name_t)
+AGENT_TX_BYTES = 0
+AGENT_RX_BYTES = 1
+AGENT_TX_REQUESTS_NUM = 2
+AGENT_RX_REQUESTS_NUM = 3
+AGENT_MEMORY_REGISTERED = 4
+AGENT_MEMORY_DEREGISTERED = 5
+AGENT_XFER_TIME = 6
+AGENT_XFER_POST_TIME = 7
+AGENT_ERROR = 8
+AGENT_BACKEND = 9
 
 # Global flag for graceful shutdown
 running = True
@@ -65,10 +75,9 @@ class NixlTelemetryEvent(ctypes.Structure):
 
     _pack_ = 1
     _fields_ = [
-        ("timestamp_us", ctypes.c_uint64),
         ("category", ctypes.c_int),
-        ("event_name", ctypes.c_char * MAX_EVENT_NAME_LEN),
-        ("_padding", ctypes.c_uint32),
+        ("event_name", ctypes.c_uint8),
+        ("_padding", ctypes.c_char * 3),
         ("value", ctypes.c_uint64),
     ]
 
@@ -197,13 +206,6 @@ class SharedRingBuffer:
             os.close(self.file_fd)
 
 
-def format_timestamp(timestamp_us):
-    """Format timestamp in microseconds to readable format"""
-    dt = datetime.fromtimestamp(timestamp_us / 1_000_000)
-    microseconds = timestamp_us % 1_000_000
-    return f"{dt.strftime('%Y-%m-%d %H:%M:%S')}.{microseconds:06d}"
-
-
 def format_bytes(bytes_val):
     """Format bytes to human readable format"""
     units = ["B", "KB", "MB", "GB", "TB"]
@@ -232,13 +234,28 @@ def get_telemetry_category_string(category):
     return category_strings.get(category, f"UNKNOWN_CATEGORY_{category}")
 
 
+def get_telemetry_event_name_string(event_name):
+    """Get string representation of telemetry event name enum"""
+    event_name_strings = {
+        AGENT_TX_BYTES: "agent_tx_bytes",
+        AGENT_RX_BYTES: "agent_rx_bytes",
+        AGENT_TX_REQUESTS_NUM: "agent_tx_requests_num",
+        AGENT_RX_REQUESTS_NUM: "agent_rx_requests_num",
+        AGENT_MEMORY_REGISTERED: "agent_memory_registered",
+        AGENT_MEMORY_DEREGISTERED: "agent_memory_deregistered",
+        AGENT_XFER_TIME: "agent_xfer_time",
+        AGENT_XFER_POST_TIME: "agent_xfer_post_time",
+        AGENT_ERROR: "agent_error",
+        AGENT_BACKEND: "agent_backend",
+    }
+    return event_name_strings.get(event_name, f"unknown_event_{event_name}")
+
+
 def print_telemetry_event(event):
     """Print telemetry event in a formatted way"""
     logger.info("\n=== NIXL Telemetry Event ===")
-    logger.info("Timestamp: %s", format_timestamp(event.timestamp_us))
 
-    # Decode event name
-    event_name = event.event_name.decode("utf-8").rstrip("\x00")
+    event_name = get_telemetry_event_name_string(event.event_name)
     category_str = get_telemetry_category_string(event.category)
 
     logger.info("Category: %s", category_str)
--- a/examples/python/telemetry_reader.py
+++ b/examples/python/telemetry_reader.py
@@ -34,6 +34,7 @@ logger = logging.getLogger(__name__)
 
 # Constants from telemetry_event.h
 TELEMETRY_VERSION = 2
+MAX_EVENT_NAME_LEN = 32
 
 # NIXL telemetry categories
 NIXL_TELEMETRY_MEMORY = 0
@@ -45,18 +46,6 @@ NIXL_TELEMETRY_PERFORMANCE = 5
 NIXL_TELEMETRY_SYSTEM = 6
 NIXL_TELEMETRY_CUSTOM = 7
 NIXL_TELEMETRY_MAX = 8
-
-# NIXL telemetry event names (nixl_telemetry_event_name_t)
-AGENT_TX_BYTES = 0
-AGENT_RX_BYTES = 1
-AGENT_TX_REQUESTS_NUM = 2
-AGENT_RX_REQUESTS_NUM = 3
-AGENT_MEMORY_REGISTERED = 4
-AGENT_MEMORY_DEREGISTERED = 5
-AGENT_XFER_TIME = 6
-AGENT_XFER_POST_TIME = 7
-AGENT_ERROR = 8
-AGENT_BACKEND = 9
 
 # Global flag for graceful shutdown
 running = True
@@ -76,8 +65,7 @@ class NixlTelemetryEvent(ctypes.Structure):
     _pack_ = 1
     _fields_ = [
         ("category", ctypes.c_int),
-        ("event_name", ctypes.c_uint8),
-        ("_padding", ctypes.c_char * 3),
+        ("event_name", ctypes.c_char * MAX_EVENT_NAME_LEN),
         ("value", ctypes.c_uint64),
     ]
 
@@ -235,20 +223,10 @@ def get_telemetry_category_string(category):
 
 
 def get_telemetry_event_name_string(event_name):
-    """Get string representation of telemetry event name enum"""
-    event_name_strings = {
-        AGENT_TX_BYTES: "agent_tx_bytes",
-        AGENT_RX_BYTES: "agent_rx_bytes",
-        AGENT_TX_REQUESTS_NUM: "agent_tx_requests_num",
-        AGENT_RX_REQUESTS_NUM: "agent_rx_requests_num",
-        AGENT_MEMORY_REGISTERED: "agent_memory_registered",
-        AGENT_MEMORY_DEREGISTERED: "agent_memory_deregistered",
-        AGENT_XFER_TIME: "agent_xfer_time",
-        AGENT_XFER_POST_TIME: "agent_xfer_post_time",
-        AGENT_ERROR: "agent_error",
-        AGENT_BACKEND: "agent_backend",
-    }
-    return event_name_strings.get(event_name, f"unknown_event_{event_name}")
+    """Get string representation of telemetry event name"""
+    if isinstance(event_name, bytes):
+        return event_name.decode("utf-8", errors="replace").rstrip("\x00")
+    return str(event_name)
 
 
 def print_telemetry_event(event):
--- a/src/core/agent_data.h
+++ b/src/core/agent_data.h
@@ -128,7 +128,7 @@ class nixlAgentData {
 
         void
         addErrorTelemetry(nixl_status_t err_status) {
-            if (telemetry_) {
+            if (telemetry_ && telemetryEnabled) {
                 telemetry_->updateErrorCount(err_status);
             }
         }
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -118,8 +118,8 @@ nixlXferReqH::updateRequestStats(nixlTelemetry *telemetry_pub,
     }
 
     if (telemetry_pub && (stat_status != NIXL_TELEMETRY_POST)) {
-        telemetry_pub->addPostTime(telemetry.postDuration);
-        telemetry_pub->addXferTime(duration, backendOp == NIXL_WRITE, telemetry.totalBytes);
+        telemetry_pub->addTransferComplete(
+            telemetry.postDuration, duration, backendOp == NIXL_WRITE, telemetry.totalBytes);
     }
 
     NIXL_TRACE << "[NIXL TELEMETRY]: From backend " << engine->getType()
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -154,25 +154,25 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
 
     const auto telemetry_enabled = nixl::config::getValueOptional<bool>(TELEMETRY_ENABLED_VAR);
 
-    if (telemetry_enabled) {
-        if (*telemetry_enabled) {
-            telemetry_ = std::make_unique<nixlTelemetry>(name);
-            telemetryEnabled = telemetry_->recording;
-            if (!telemetryEnabled) {
-                NIXL_WARN << "NIXL_TELEMETRY_ENABLE is set but no telemetry exporter is configured "
-                             "(set NIXL_TELEMETRY_EXPORTER or NIXL_TELEMETRY_DIR); "
-                             "per-transfer telemetry and error export are disabled.";
-            }
-        } else if (config.captureTelemetry) {
-            telemetryEnabled = true;
+    const bool env_on = telemetry_enabled && *telemetry_enabled;
+    const bool config_wants = config.captureTelemetry;
+
+    if (env_on || config_wants) {
+        if (telemetry_enabled && !*telemetry_enabled && config_wants) {
             NIXL_WARN << "NIXL telemetry is enabled through config, "
                          "ignoring the NIXL_TELEMETRY_ENABLE environment variable";
-        } else {
-            NIXL_DEBUG << "NIXL telemetry is disabled";
         }
-    } else if (config.captureTelemetry) {
-        telemetryEnabled = true;
-        NIXL_DEBUG << "Capturing NIXL telemetry based on config (without an output file)";
+        telemetry_ = std::make_unique<nixlTelemetry>(name);
+        telemetryEnabled = telemetry_->recording;
+        if (!telemetryEnabled) {
+            NIXL_WARN << "Telemetry was requested but no exporter is configured "
+                         "(set NIXL_TELEMETRY_EXPORTER or NIXL_TELEMETRY_DIR); "
+                         "per-transfer telemetry and error export are disabled.";
+        } else if (!env_on && config_wants) {
+            NIXL_DEBUG << "Capturing NIXL telemetry based on config";
+        }
+    } else if (telemetry_enabled && !*telemetry_enabled) {
+        NIXL_DEBUG << "NIXL telemetry is disabled";
     }
 }
 
@@ -333,7 +333,7 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     init_params.enableProgTh = data->config_.useProgThread;
     init_params.pthrDelay = data->config_.pthrDelay;
     init_params.syncMode = data->config_.syncMode;
-    init_params.enableTelemetry_ = (data->telemetry_ != nullptr) && data->telemetryEnabled;
+    init_params.enableTelemetry_ = data->telemetryEnabled;
 
     // First, try to load the backend as a plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -154,11 +154,11 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
 
     const auto telemetry_enabled = nixl::config::getValueOptional<bool>(TELEMETRY_ENABLED_VAR);
 
-    const bool env_on = telemetry_enabled && *telemetry_enabled;
-    const bool config_wants = config.captureTelemetry;
+    const bool telemetry_enabled_env = telemetry_enabled && *telemetry_enabled;
+    const bool telemetry_enabled_config = config.captureTelemetry;
 
-    if (env_on || config_wants) {
-        if (telemetry_enabled && !*telemetry_enabled && config_wants) {
+    if (telemetry_enabled_env || telemetry_enabled_config) {
+        if (telemetry_enabled && !*telemetry_enabled && telemetry_enabled_config) {
             NIXL_WARN << "NIXL telemetry is enabled through config, "
                          "ignoring the NIXL_TELEMETRY_ENABLE environment variable";
         }
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -156,8 +156,13 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
 
     if (telemetry_enabled) {
         if (*telemetry_enabled) {
-            telemetryEnabled = true;
             telemetry_ = std::make_unique<nixlTelemetry>(name);
+            telemetryEnabled = telemetry_->recording;
+            if (!telemetryEnabled) {
+                NIXL_WARN << "NIXL_TELEMETRY_ENABLE is set but no telemetry exporter is configured "
+                             "(set NIXL_TELEMETRY_EXPORTER or NIXL_TELEMETRY_DIR); "
+                             "per-transfer telemetry and error export are disabled.";
+            }
         } else if (config.captureTelemetry) {
             telemetryEnabled = true;
             NIXL_WARN << "NIXL telemetry is enabled through config, "
@@ -328,7 +333,8 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     init_params.enableProgTh = data->config_.useProgThread;
     init_params.pthrDelay = data->config_.pthrDelay;
     init_params.syncMode = data->config_.syncMode;
-    init_params.enableTelemetry_ = (data->telemetry_ != nullptr);
+    init_params.enableTelemetry_ =
+        (data->telemetry_ != nullptr) && data->telemetryEnabled;
 
     // First, try to load the backend as a plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
@@ -471,7 +477,7 @@ nixlAgent::registerMem(const nixl_reg_dlist_t &descs,
 
     if (count > 0) {
         // sum all the sizes of the descriptors using std::accumulate
-        if (data->telemetry_) {
+        if (data->telemetry_ && data->telemetryEnabled) {
             uint64_t total_size = std::accumulate(
                 descs.begin(),
                 descs.end(),
@@ -514,7 +520,7 @@ nixlAgent::deregisterMem(const nixl_reg_dlist_t &descs,
             bad_ret = ret;
     }
     if (bad_ret == NIXL_SUCCESS) {
-        if (data->telemetry_) {
+        if (data->telemetry_ && data->telemetryEnabled) {
             uint64_t total_size = std::accumulate(
                 descs.begin(),
                 descs.end(),
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -333,8 +333,7 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     init_params.enableProgTh = data->config_.useProgThread;
     init_params.pthrDelay = data->config_.pthrDelay;
     init_params.syncMode = data->config_.syncMode;
-    init_params.enableTelemetry_ =
-        (data->telemetry_ != nullptr) && data->telemetryEnabled;
+    init_params.enableTelemetry_ = (data->telemetry_ != nullptr) && data->telemetryEnabled;
 
     // First, try to load the backend as a plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -168,7 +168,7 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
             NIXL_WARN << "Telemetry was requested but no exporter is configured "
                          "(set NIXL_TELEMETRY_EXPORTER or NIXL_TELEMETRY_DIR); "
                          "per-transfer telemetry and error export are disabled.";
-        } else if (!env_on && config_wants) {
+        } else if (!telemetry_enabled_env && telemetry_enabled_config) {
             NIXL_DEBUG << "Capturing NIXL telemetry based on config";
         }
     } else if (telemetry_enabled && !*telemetry_enabled) {
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -171,7 +171,7 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
         } else if (!telemetry_enabled_env && telemetry_enabled_config) {
             NIXL_DEBUG << "Capturing NIXL telemetry based on config";
         }
-    } else if (telemetry_enabled && !*telemetry_enabled) {
+    } else {
         NIXL_DEBUG << "NIXL telemetry is disabled";
     }
 }
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -178,7 +178,6 @@ void
 nixlTelemetry::updateData(const std::string &event_name,
                           nixl_telemetry_category_t category,
                           uint64_t value) {
-
     const size_t id = writeIdx_.fetch_add(1);
     if (id >= maxEventsBuffered_) {
         overflowed_.store(true, std::memory_order_relaxed);
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -125,7 +125,7 @@ nixlTelemetry::initializeTelemetry() {
     event_buffers_[0].resize(max_events_buffered_);
     event_buffers_[1].resize(max_events_buffered_);
     write_idx_.store(0);
-    write_buffer_index_.store(0);
+    write_buf_index_.store(0);
 
     const auto run_interval =
         nixl::config::getValueDefaulted(TELEMETRY_RUN_INTERVAL_VAR, DEFAULT_TELEMETRY_RUN_INTERVAL);
@@ -143,14 +143,14 @@ nixlTelemetry::writeEventHelper() {
         return true;
     }
     // Block producers by setting write_idx_=max so they get id>=max and skip.
+    const int w = write_buf_index_.load();
     const size_t num_events_raw = write_idx_.exchange(max_events_buffered_);
-    const int old_write = write_buffer_index_.load();
-    write_buffer_index_.store(1 - old_write);
+    write_buf_index_.store(1 - w);
     write_idx_.store(0);
     const size_t num_events = std::min(num_events_raw, max_events_buffered_);
 
     for (size_t i = 0; i < num_events; ++i) {
-        exporter_->exportEvent(event_buffers_[old_write][i]);
+        exporter_->exportEvent(event_buffers_[w][i]);
     }
 
     return true;
@@ -184,8 +184,7 @@ nixlTelemetry::updateData(const std::string &event_name,
     if (id >= max_events_buffered_) {
         return;
     }
-    const int idx = write_buffer_index_.load();
-    event_buffers_[idx][id] =
+    event_buffers_[write_buf_index_.load()][id] =
         nixlTelemetryEvent(std::chrono::duration_cast<std::chrono::microseconds>(
                                std::chrono::system_clock::now().time_since_epoch())
                                .count(),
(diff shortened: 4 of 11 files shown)


src/core/telemetry/telemetry.cpp
```diff
@@ -133,17 +133,18 @@ nixlTelemetry::initializeTelemetry() {
 
 bool
 nixlTelemetry::writeEventHelper() {
-    std::vector<nixlTelemetryEvent> next_queue;
-    // assume next buffer will be the same size as the current one
-    next_queue.reserve(exporter_->getMaxEventsBuffered());
-    {
-        std::lock_guard<std::mutex> lock(mutex_);
-        events_.swap(next_queue);
+    if (!exporter_ || max_events_buffered_ == 0) {
```
bilkovsky: In the initializeTelemetry we rule out such case. This check is redundant

src/core/telemetry/telemetry.cpp
```diff
@@ -121,6 +121,13 @@ nixlTelemetry::initializeTelemetry() {
 
     NIXL_DEBUG << "NIXL telemetry is enabled with exporter: " << *exporter_name;
 
+    max_events_buffered_ = buffer_size;
+    event_buffers_[0].resize(max_events_buffered_);
+    event_buffers_[1].resize(max_events_buffered_);
+    write_idx_.store(0);
+    write_buffer_index_.store(0);
```
bilkovsky: What I had in mind is not keeping buffer index, but two pointers reader*, writer* each points on one of the buffers, and those pointers are being swapped by writeEventHelper()

src/core/telemetry/telemetry.cpp
```diff
@@ -227,22 +249,37 @@ nixlTelemetry::updateMemoryDeregistered(uint64_t memory_deregistered) {
 
 void
 nixlTelemetry::addXferTime(std::chrono::microseconds xfer_time, bool is_write, uint64_t bytes) {
+    if (!exporter_ || max_events_buffered_ == 0) {
```
bilkovsky: Remove

src/core/telemetry/telemetry.cpp
```diff
@@ -170,14 +179,27 @@ void
...
-    std::lock_guard<std::mutex> lock(mutex_);
-    events_.emplace_back(std::chrono::duration_cast<std::chrono::microseconds>(
-                             std::chrono::system_clock::now().time_since_epoch())
-                             .count(),
-                         category,
-                         event_name,
-                         value);
+    if (!exporter_ || max_events_buffered_ == 0) {
+        return;
+    }
```
bilkovsky: Remove

src/core/telemetry/telemetry.cpp
```diff
@@ -170,14 +179,27 @@ void
...
+    const int idx = write_buffer_index_.load();
+    event_buffers_[idx][id] =
+        nixlTelemetryEvent(std::chrono::duration_cast<std::chrono::microseconds>(
+                               std::chrono::system_clock::now().time_since_epoch())
+                               .count(),
+                           category,
+                           event_name,
+                           value);
+    if (id + 1 == max_events_buffered_) {
+        full_.store(true);
```
bilkovsky: What if we do swap and set buffer to 0 right before this line? It seems for me like we are going to skip next batch in that case

src/core/telemetry/telemetry.cpp
BatshevaBlack: when the NIXL_TELEMETRY_DIR isn't set there is a seg fault - which means the check is crucial for when the var isn't set

src/core/telemetry/telemetry.cpp
```diff
@@ -170,14 +179,27 @@ void
...
+    if (!exporter_ || max_events_buffered_ == 0) {
+        return;
+    }
+    if (full_.load()) {
+        return;
+    }
+    const size_t id = write_idx_.fetch_add(1);
+    if (id >= max_events_buffered_) {
+        return;
+    }
```
bilkovsky: Is not second check sufficient and we can get rid of full_ flag?

src/core/telemetry/telemetry.cpp
BatshevaBlack: when the NIXL_TELEMETRY_DIR isn't set there is a seg fault - which means the check is crucial for when the var isn't set

src/core/telemetry/telemetry.cpp
BatshevaBlack: when the NIXL_TELEMETRY_DIR isn't set there is a seg fault - which means the check is crucial for when the var isn't set

src/core/telemetry/telemetry.cpp
BatshevaBlack: implemented the pointer for write, read is within the function

src/core/telemetry/telemetry.cpp
bilkovsky: In all cases that behavior is not OK. I'd suggest to change it

bilkovsky:
@coderabbitai review

bilkovsky:
@coderabbitai full review

src/core/telemetry/telemetry.cpp
BatshevaBlack: fixed

src/core/telemetry/telemetry.cpp
```diff
@@ -87,6 +87,17 @@ getExporterName() {
...
 }
 
+void
+fillTelemetryEventSlot(nixlTelemetryEvent &slot,
+                       nixl_telemetry_category_t category,
+                       const char *event_name,
+                       uint64_t value) noexcept {
+    slot.category_ = category;
+    strncpy(slot.eventName_, event_name, MAX_EVENT_NAME_LEN - 1);
+    slot.eventName_[MAX_EVENT_NAME_LEN - 1] = '\0';
```
alexanderbilk: We don't have backend events and all our even names are hard coded. Maybe save a copy and pass event_names as enum? With helper function to translate to a string for exporter

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +146,21 @@ nixlTelemetry::initializeTelemetry() {
     writeTask_.interval_ = run_interval;
     writeTask_.enabled_ = true;
     registerPeriodicTask(writeTask_);
+    recording = true;
```
alexanderbilk, line 136: Not sure getting the point of it? Seems redundant from the first glance

src/core/telemetry/telemetry.cpp
```diff
@@ -234,15 +259,35 @@ nixlTelemetry::addXferTime(std::chrono::microseconds xfer_time, bool is_write, u
...
+
+    const size_t id_req = writeIdx_.fetch_add(1);
+    if (id_req >= maxEventsBuffered_) {
+        return;
+    }
+    fillTelemetryEventSlot(eventBuffers_[writeBufIndex_.load()][id_req],
+                           time,
+                           nixl_telemetry_category_t::NIXL_TELEMETRY_TRANSFER,
+                           requests_name,
+                           1);
```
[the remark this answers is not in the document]
alexanderbilk: @BatshevaBlack Agree here, besides that, why to make 3 atomic operations instead of one?

src/core/telemetry/telemetry_event.h
```diff
@@ -55,29 +55,25 @@ telemetryCategoryStr(const nixl_telemetry_category_t &category);
...
         eventName_[MAX_EVENT_NAME_LEN - 1] = '\0';
     }
 
-    nixlTelemetryEvent(uint64_t timestamp_us,
-                       nixl_telemetry_category_t category,
+    nixlTelemetryEvent(nixl_telemetry_category_t category,
                        const std::string &event_name,
                        uint64_t value) noexcept
-        : nixlTelemetryEvent(timestamp_us, category, event_name.c_str(), value) {}
+        : nixlTelemetryEvent(category, event_name.c_str(), value) {}
```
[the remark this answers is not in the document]
alexanderbilk: @BatshevaBlack +

src/core/agent_data.h
```diff
@@ -128,7 +128,7 @@ class nixlAgentData {
 
         void
         addErrorTelemetry(nixl_status_t err_status) {
-            if (telemetry_) {
+            if (telemetry_ && telemetryEnabled) {
```
alexanderbilk, line 131: Again seems redundant. If we need 2 fields to reflect same idea something has to be changed

alexanderbilk (commented):
@BatshevaBlack Please address comments. And fix documentation to reflect changes

src/core/telemetry/telemetry.cpp
BatshevaBlack: recording tells the agent whether telemetry actually initialized with a working exporter. The telemetry object is always created, but if no exporter is configured (no NIXL_TELEMETRY_DIR or NIXL_TELEMETRY_EXPORTER), initializeTelemetry() returns early and recording stays false. The agent checks telemetry_->recording to skip all telemetry work on the data path when there's nothing to export to.

src/core/agent_data.h
BatshevaBlack: Not redundant - telemetry_ can be non-null while telemetryEnabled is false. This happens when telemetry is requested but no exporter is configured: the telemetry object is created, but recording stays false so telemetryEnabled is false. We need both checks because telemetry_ existing doesn't mean it's actually active.

src/core/telemetry/telemetry.cpp
BatshevaBlack: done

src/core/agent_data.h
alexanderbilk: Is there any use case when telemetry is getting collected but not getting recorded?

src/core/telemetry/telemetry.cpp
```diff
@@ -167,91 +183,131 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
+    const auto bytes_name = is_write ? nixl_telemetry_event_name_t::AGENT_TX_BYTES :
+                                       nixl_telemetry_event_name_t::AGENT_RX_BYTES;
+    const auto requests_name = is_write ? nixl_telemetry_event_name_t::AGENT_TX_REQUESTS_NUM :
+                                          nixl_telemetry_event_name_t::AGENT_RX_REQUESTS_NUM;
+
+    const size_t id_base = writeIdx_.fetch_add(3);
+    if (id_base + 2 >= maxEventsBuffered_) {
+        return;
+    }
+    auto *arr = eventBuffers_[writeBufIndex_.load()];
```
guy-ealey-morag:
```suggestion
    auto &arr = eventBuffers_[writeBufIndex_.load()];
```

I got a compilation error for this line

```
../../src/core/telemetry/telemetry.cpp:260:52: error: unable to deduce ‘auto*’ from ‘((nixlTelemetry*)this)->nixlTelemetry::eventBuffers_[((nixlTelemetry*)this)->nixlTelemetry::writeBufIndex_.std::atomic<int>::<anonymous>.std::__atomic_base<int>::load(std::memory_order_seq_cst)]’
  260 |     auto *arr = eventBuffers_[writeBufIndex_.load()];
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
../../src/core/telemetry/telemetry.cpp:260:52: note:   mismatched types ‘auto*’ and ‘std::vector<nixlTelemetryEvent>’
  260 |     auto *arr = eventBuffers_[writeBufIndex_.load()];
      | 
```

src/core/telemetry/telemetry_event.h
```diff
@@ -45,39 +43,73 @@ enum class nixl_telemetry_category_t {
...
 
 /**
  * @struct nixlTelemetryEvent
  * @brief A structure to hold individual telemetry event data for cyclic buffer storage
  */
 struct nixlTelemetryEvent {
-    uint64_t timestampUs_;
     nixl_telemetry_category_t category_; // Main event category for filtering
-    char eventName_[MAX_EVENT_NAME_LEN]; // Detailed event name/identifier
+    nixl_telemetry_event_name_t eventName_; // Detailed event name/identifier
```
guy-ealey-morag:
```suggestion
    nixl_telemetry_event_type_t eventType_;
```

Now that the event name is not stored as a string, it is more appropriate to rename it to "event type".
This is also relevant to all other places in the code that reference it.

examples/python/telemetry_reader.py
```diff
@@ -232,13 +234,28 @@ def get_telemetry_category_string(category):
     return category_strings.get(category, f"UNKNOWN_CATEGORY_{category}")
 
 
+def get_telemetry_event_name_string(event_name):
+    """Get string representation of telemetry event name enum"""
+    event_name_strings = {
```
guy-ealey-morag: The dictionary should be defined at the module scope to avoid recreating the entire dict for every call of `get_telemetry_event_name_string`

This is also applicable for the dict in `get_telemetry_category_string`

src/core/nixl_agent.cpp
```diff
@@ -118,8 +118,8 @@ nixlXferReqH::updateRequestStats(nixlTelemetry *telemetry_pub,
     }
 
     if (telemetry_pub && (stat_status != NIXL_TELEMETRY_POST)) {
-        telemetry_pub->addPostTime(telemetry.postDuration);
-        telemetry_pub->addXferTime(duration, backendOp == NIXL_WRITE, telemetry.totalBytes);
```
guy-ealey-morag, line 122: It looks like `addPostTime` and `addXferTime` are not used anywhere, should they be removed?

src/core/telemetry/telemetry.h
```diff
@@ -72,20 +71,33 @@ class nixlTelemetry {
...
+               nixl_telemetry_category_t category,
+               uint64_t value);
     bool
     writeEventHelper();
     std::unique_ptr<nixlTelemetryExporter> exporter_;
     std::unique_ptr<sharedRingBuffer<nixlTelemetryEvent>> buffer_;
-    std::vector<nixlTelemetryEvent> events_;
-    std::mutex mutex_;
+    // Double buffer: writeBufIndex_ is 0|1 — same role as colleague's write* -> arr. Swap = one
+    // store.
```
guy-ealey-morag: Can you explain the second part of the comment?
`same role as colleague's write* -> arr. Swap = one store.`

src/core/telemetry/telemetry.cpp
```diff
@@ -87,6 +87,16 @@ getExporterName() {
     return defaultTelemetryPlugin;
 }
 
+void
+fillTelemetryEventSlot(nixlTelemetryEvent &slot,
```
guy-ealey-morag: This should be a member function of `nixlTelemetryEvent`

src/core/telemetry/telemetry.h
```diff
@@ -72,20 +71,33 @@ class nixlTelemetry {
...
+    std::vector<nixlTelemetryEvent> eventBuffers_[2];
```
guy-ealey-morag: I think it could be cleaner if the 4 attributes of the double buffer would be encapsulated in a new class.
Then that class can expose function with meaningful names to improve readability and robustness.

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +145,21 @@ nixlTelemetry::initializeTelemetry() {
...
-    }
 
-    for (auto &event : next_queue) {
-        // if full, ignore
-        exporter_->exportEvent(event);
+    // Block producers by setting writeIdx_=max so they get id>=max and skip.
+    const int w = writeBufIndex_.load();
+    const size_t num_events_raw = writeIdx_.exchange(maxEventsBuffered_);
+    writeBufIndex_.store(1 - w);
+    writeIdx_.store(0);
```
guy-ealey-morag:
```suggestion
    const size_t num_events_raw = writeIdx_.exchange(maxEventsBuffered_);
    const int w = writeBufIndex_.fetch_xor(1);
    writeIdx_.store(0);
```

reading and inverting of `w` can be done in one atomic operation

src/core/telemetry/telemetry.cpp
```diff
@@ -167,91 +183,131 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
-                         category,
-                         event_name,
-                         value);
+
+    const size_t id = writeIdx_.fetch_add(1);
+    if (id >= maxEventsBuffered_) {
+        return;
+    }
+
+    fillTelemetryEventSlot(eventBuffers_[writeBufIndex_.load()][id], category, event_name, value);
```
guy-ealey-morag:
```suggestion
    const size_t id = writeIdx_.fetch_add(1);
    if (id >= maxEventsBuffered_) {
        return;
    }
    
    // possible race here

    fillTelemetryEventSlot(eventBuffers_[writeBufIndex_.load()][id], category, event_name, value);
```

There is a possible race if the consumer drains the events between the `writeIdx_.fetch_add` and `writeBufIndex_.load()` (the writeBufIndex may change to the second buffer while the writeIdx belongs to the first one).

Is there a promise that it won't happen here? I think it's not very robust.

I suggest looking into either have two separate `writeIdx` (one for each buffer), or alternatively pack `writeIdx_` and `writeBufIndex_` into a single atomic variable and use masking to separate the values.

This is relevant for the other functions that call `writeIdx_.fetch_add` as well

guy-ealey-morag:
Should we log a warning when telemetry is skipped?
Maybe log once when it happens to not spam the logs, but let the user know it happened.
Alternatively, we could set a boolean flag and read it after the drain to log a special telemetry event that indicates that some are missing

guy-ealey-morag:
Why were the event timestamps removed?

src/core/telemetry/telemetry.cpp
```diff
@@ -121,6 +131,12 @@ nixlTelemetry::initializeTelemetry() {
 
     NIXL_DEBUG << "NIXL telemetry is enabled with exporter: " << *exporter_name;
 
+    maxEventsBuffered_ = buffer_size;
+    eventBuffers_[0].resize(maxEventsBuffered_);
+    eventBuffers_[1].resize(maxEventsBuffered_);
```
guy-ealey-morag: We can change the type of the buffers to `std::unique_ptr<nixlTelemetryEvent[]>` instead of `std::vector<nixlTelemetryEvent>` and then we can initialize them using:

```
eventBuffers_[0] = std::make_unique<nixlTelemetryEvent[]>(buffer_size);
eventBuffers_[1] = std::make_unique<nixlTelemetryEvent[]>(buffer_size);
```

It's still heap allocated, but it also shows that the size is constant (possibly avoiding some overhead as well)

src/core/telemetry/telemetry_event.h
```diff
@@ -45,39 +43,73 @@ enum class nixl_telemetry_category_t {
...
+    AGENT_XFER_POST_TIME = 7,
+    AGENT_ERROR = 8,
+    AGENT_BACKEND = 9,
+};
+
 namespace nixlEnumStrings {
 std::string
 telemetryCategoryStr(const nixl_telemetry_category_t &category);
+
+inline std::string
```
ColinNV: Suggest `[[nodiscard]] constexpr std::string_view toString(nixl_telemetry_event_name_t name) noexcept {`.

src/core/telemetry/telemetry_event.h
```diff
@@ -45,39 +43,73 @@ enum class nixl_telemetry_category_t {
...
+        return "agent_memory_deregistered";
+    case nixl_telemetry_event_name_t::AGENT_XFER_TIME:
+        return "agent_xfer_time";
+    case nixl_telemetry_event_name_t::AGENT_XFER_POST_TIME:
+        return "agent_xfer_post_time";
+    case nixl_telemetry_event_name_t::AGENT_ERROR:
+        return "agent_error";
+    case nixl_telemetry_event_name_t::AGENT_BACKEND:
+        return "agent_backend";
+    default:
```
ColinNV: The default case should be handled outside of the `switch`.

ColinNV (changes requested):
This PR does two things, it changes the event names from a string to an enum, and it changes the data structure; can these be separated into two PRs?

And are we sure the new lock-free data structure is worth the complexity? As it stands `writeEventHelper()` only does a vector swap, the vector is pre-allocated, so all other functions only hold the mutex for very simple operations. Given that an uncontended mutex lock stays in userland...

BatshevaBlack:
> Why were the event timestamps removed? @guy-ealey-morag

 Per-event system_clock::now() calls add overhead on the hot path. This timestamp does not add any value since it's the time when event was recorded, not when happend

src/core/telemetry/telemetry.cpp
BatshevaBlack: We prefer to keep the lock-free path as lightweight as possible, so we accept this race. In the unlikely event it fires (nanosecond window vs. 100ms drain interval), at most one event is misplaced per thread per cycle - acceptable for lossy telemetry that's aggregated at 100ms intervals.

src/core/telemetry/telemetry.h
BatshevaBlack: Agreed, will refactor into a dedicated class in a follow-up PR to keep this one focused on the core mechanism.

src/core/telemetry/telemetry.cpp
BatshevaBlack: Thanks for your review!
done

examples/python/telemetry_reader.py
BatshevaBlack: I've moved all these changes to a new PR as Colin required, I addressed this point there.

src/core/telemetry/telemetry_event.h
BatshevaBlack: Addressed in new PR

src/core/nixl_agent.cpp
BatshevaBlack: done

src/core/telemetry/telemetry.cpp
BatshevaBlack: done thanks

src/core/telemetry/telemetry.cpp
BatshevaBlack: done thanks

src/core/telemetry/telemetry.cpp
BatshevaBlack: done

src/core/telemetry/telemetry_event.h
BatshevaBlack: Addressed in new PR

src/core/telemetry/telemetry_event.h
BatshevaBlack: Addressed in new PR

BatshevaBlack:
@ColinNV Thanks for your review!
> This PR does two things, it changes the event names from a string to an enum, and it changes the data structure; can these be separated into two PRs?

Done — the enum naming changes have been moved to a separate branch/PR (telemetry-enum-names). This PR now only contains the double-buffer data structure changes.

> And are we sure the new lock-free data structure is worth the complexity? As it stands `writeEventHelper()` only does a vector swap, the vector is pre-allocated, so all other functions only hold the mutex for very simple operations. Given that an uncontended mutex lock stays in userland...

This change addresses a real performance gap we observed - enabling telemetry caused a significant drop in transfer performance. The mutex on the hot path was the bottleneck: telemetry updates are called per-transfer by every thread posting RDMA operations, and under high concurrency the mutex caused cache-line bouncing and contention even when technically "uncontended."

With the lock-free double buffer, producers only do a single fetch_add per event, eliminating the contention. This closed the performance gap between telemetry-on and telemetry-off.

src/core/telemetry/telemetry.cpp
guy-ealey-morag: The alternatives that solve this race are also lightweight, so why not have a robust solution?

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,24 @@ nixlTelemetry::initializeTelemetry() {
...
 
 bool
 nixlTelemetry::writeEventHelper() {
-    std::vector<nixlTelemetryEvent> next_queue;
-    // assume next buffer will be the same size as the current one
-    next_queue.reserve(exporter_->getMaxEventsBuffered());
-    {
-        std::lock_guard<std::mutex> lock(mutex_);
-        events_.swap(next_queue);
+
```
guy-ealey-morag:
```suggestion
```

src/core/telemetry/telemetry.cpp
```diff
@@ -170,17 +178,18 @@ void
...
                           uint64_t value) {
-    // agent can be multi-threaded
-    std::lock_guard<std::mutex> lock(mutex_);
-    events_.emplace_back(std::chrono::duration_cast<std::chrono::microseconds>(
-                             std::chrono::system_clock::now().time_since_epoch())
-                             .count(),
-                         category,
-                         event_name,
-                         value);
+
```
guy-ealey-morag, line 186:
```suggestion
```

src/core/nixl_agent.cpp
```diff
@@ -154,20 +154,25 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
...
     const auto telemetry_enabled = nixl::config::getValueOptional<bool>(TELEMETRY_ENABLED_VAR);
 
-    if (telemetry_enabled) {
-        if (*telemetry_enabled) {
-            telemetryEnabled = true;
-            telemetry_ = std::make_unique<nixlTelemetry>(name);
-        } else if (config.captureTelemetry) {
-            telemetryEnabled = true;
+    const bool env_on = telemetry_enabled && *telemetry_enabled;
+    const bool config_wants = config.captureTelemetry;
```
guy-ealey-morag: I think the name "config_wants" is a bit confusing and not typical.
maybe rename to have both vars consistent names?

for example: `telemetry_enabled_env` and `telemetry_enabled_config`

src/core/telemetry/telemetry.cpp
BatshevaBlack: @guy-ealey-morag thanks!
Good point. Rather than patching this with packed atomics, I'm planning to move to a buffer-per-producer design in a follow-up PR, which eliminates shared atomic contention entirely. The consumer would drain all per-producer buffers on each cycle. Will address this in the next iteration. since we want to push this fix to solve the performance issue.
Is this ok for now?

src/core/agent_data.h
guy-ealey-morag: I have a suggestion to solve the redundancy of having `telemetry_`, `telemetryEnabled` and `recording`.

When creating the `telemetry_` it might fail to create an exporter (resulting in `recording = false`) which then results in `telemetryEnabled = false`.
At this point it doesn't make sense to keep the `_telemetry` object alive. We can set it to `nullptr` and use it as an indication that telemetry is disabled without the need for the boolean flags (`if (telemetry_) { ... }` instead of `if (telemetryEnabled) { ... }`)

In addition, instead of creating the object and then resetting it on failure, I suggest creating a factory function that checks for the exporter before creating the object and then returns a valid object or `nullptr` on failure (the constructor should be private), for example:

```c++
std::unique_ptr<nixlTelemetry>
nixlTelemetry::create(const std::string &agent_name) {
   bool success = ... // try to create the exporter
   if (!success) return nullptr;
   return std::unique_ptr<nixlTelemetry>(...);
}
``` 

Let me know what you think @alexanderbilk @BatshevaBlack @ColinNV

src/core/nixl_agent.cpp
BatshevaBlack: @guy-ealey-morag done thanks!

src/core/agent_data.h
BatshevaBlack: I think it's a good suggestion — using a factory function and treating telemetry_ != nullptr as the single source of truth is cleaner. This is a broader refactoring though (touches the constructor, callers, and the boolean flags), so I'll address it in a follow-up PR to keep this one focused on the double-buffer mechanism.

src/core/nixl_agent.cpp
```diff
@@ -154,20 +154,25 @@ nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &con
...
+        telemetryEnabled = telemetry_->recording;
+        if (!telemetryEnabled) {
+            NIXL_WARN << "Telemetry was requested but no exporter is configured "
+                         "(set NIXL_TELEMETRY_EXPORTER or NIXL_TELEMETRY_DIR); "
+                         "per-transfer telemetry and error export are disabled.";
+        } else if (!telemetry_enabled_env && telemetry_enabled_config) {
+            NIXL_DEBUG << "Capturing NIXL telemetry based on config";
+        }
+    } else if (telemetry_enabled && !*telemetry_enabled) {
+        NIXL_DEBUG << "NIXL telemetry is disabled";
```
```suggestion
    } else {
        NIXL_DEBUG << "NIXL telemetry is disabled";
```

Last thing, shouldn't this be always logged? (regardless if the env var was set to false or not set at all)

guy-ealey-morag (dismissed):

guy-ealey-morag (dismissed):

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,23 @@ nixlTelemetry::initializeTelemetry() {
...
 bool
 nixlTelemetry::writeEventHelper() {
-    std::vector<nixlTelemetryEvent> next_queue;
-    // assume next buffer will be the same size as the current one
-    next_queue.reserve(exporter_->getMaxEventsBuffered());
-    {
-        std::lock_guard<std::mutex> lock(mutex_);
-        events_.swap(next_queue);
+    // Block producers by setting writeIdx_=max so they get id>=max and skip.
+    const size_t num_events_raw = writeIdx_.exchange(maxEventsBuffered_);
```
ColinNV: This can pick up up to 3 bogus events, `addTransferComplete()` advances `writeIdx_` by 4 but doesn't write to any available slots on overflow.

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,23 @@ nixlTelemetry::initializeTelemetry() {
...
+    const int w = writeBufIndex_.fetch_xor(1);
+    writeIdx_.store(0);
+    const size_t num_events = std::min(num_events_raw, maxEventsBuffered_);
+
+    if (overflowed_.exchange(false, std::memory_order_relaxed)) {
```
ColinNV: Isn't `overflowed_` redundant and can be completely removed? Just check `num_events_raw` against `maxEventsBuffered_`, which already happens in the call to `std::min()`.

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,23 @@ nixlTelemetry::initializeTelemetry() {
...
 
 bool
 nixlTelemetry::writeEventHelper() {
-    std::vector<nixlTelemetryEvent> next_queue;
-    // assume next buffer will be the same size as the current one
-    next_queue.reserve(exporter_->getMaxEventsBuffered());
-    {
-        std::lock_guard<std::mutex> lock(mutex_);
-        events_.swap(next_queue);
+    // Block producers by setting writeIdx_=max so they get id>=max and skip.
```
ColinNV: "Block" sounds like "make them wait".

src/core/telemetry/telemetry.cpp
guy-ealey-morag: According to https://github.com/ai-dynamo/nixl/pull/1425#discussion_r3062756060, the same issue would happen here if a race happens (a slot is reserved but left unpopulated).

I think this must be corrected in this PR.

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,23 @@ nixlTelemetry::initializeTelemetry() {
...
+    const size_t num_events_raw = writeIdx_.exchange(maxEventsBuffered_);
+    const int w = writeBufIndex_.fetch_xor(1);
```
ColinNV: Suggest an unsigned data type for the buffer index.

src/core/telemetry/telemetry.h
```diff
@@ -84,8 +85,13 @@ class nixlTelemetry {
     writeEventHelper();
     std::unique_ptr<nixlTelemetryExporter> exporter_;
     std::unique_ptr<sharedRingBuffer<nixlTelemetryEvent>> buffer_;
-    std::vector<nixlTelemetryEvent> events_;
-    std::mutex mutex_;
+    // Lock-free double buffer: producers write to eventBuffers_[writeBufIndex_],
+    // consumer swaps writeBufIndex_ and drains the inactive buffer.
+    std::unique_ptr<nixlTelemetryEvent[]> eventBuffers_[2];
+    size_t maxEventsBuffered_{0};
+    std::atomic<size_t> writeIdx_{0};
```
ColinNV: `writeIndex_`

src/core/telemetry/telemetry.h
```diff
@@ -84,8 +85,13 @@ class nixlTelemetry {
...
+    std::atomic<int> writeBufIndex_{0};
```
ColinNV: `writeBufferIndex_`

src/core/telemetry/telemetry.h
```diff
@@ -84,8 +85,13 @@ class nixlTelemetry {
     writeEventHelper();
     std::unique_ptr<nixlTelemetryExporter> exporter_;
     std::unique_ptr<sharedRingBuffer<nixlTelemetryEvent>> buffer_;
-    std::vector<nixlTelemetryEvent> events_;
-    std::mutex mutex_;
+    // Lock-free double buffer: producers write to eventBuffers_[writeBufIndex_],
+    // consumer swaps writeBufIndex_ and drains the inactive buffer.
+    std::unique_ptr<nixlTelemetryEvent[]> eventBuffers_[2];
+    size_t maxEventsBuffered_{0};
```
ColinNV: `eventBufferSize_`

src/core/telemetry/telemetry.cpp
```diff
@@ -129,21 +135,23 @@ nixlTelemetry::initializeTelemetry() {
...
+    writeIdx_.store(0);
```
ColinNV: Since we're at this level, if the write buffer index becomes e.g. the high bit of the write index, and a separate copy of it is kept just for this function to use, then a single atomic exchange operation can be used to swap the buffers and reset the write index.

ColinNV (changes requested):

src/core/telemetry/telemetry.cpp
ColinNV: And independent of overflow, a bogus event will be picked up if the producer advanced the write index but didn't fill in the slot yet.

src/core/telemetry/telemetry.cpp
guy-ealey-morag: I think it's similar to what I suggested here https://github.com/ai-dynamo/nixl/pull/1425#discussion_r3021218410 (and it would also solve the race)

ColinNV:
Can the tests be extended to hammer the telemetry with events and check whether everything works as expected?

src/core/telemetry/telemetry.cpp
BatshevaBlack: @ColinNV @guy-ealey-morag 
Thanks for all your comments!
I see since this PR is dealing with quite a few fixes, and some can be merged easily without relation to the atomic implementation discussion- I would like to merge the little fixes first.
would you both be able to review this PR? https://github.com/ai-dynamo/nixl/pull/1516 - a simple fix taken out from the current one.
thanks!!

BatshevaBlack:
> Can the tests be extended to hammer the telemetry with events and check whether everything works as expected?

I will handle it in a new PR

src/core/telemetry/telemetry.h
```diff
@@ -83,9 +82,15 @@ class nixlTelemetry {
     bool
     writeEventHelper();
     std::unique_ptr<nixlTelemetryExporter> exporter_;
-    std::unique_ptr<sharedRingBuffer<nixlTelemetryEvent>> buffer_;
-    std::vector<nixlTelemetryEvent> events_;
-    std::mutex mutex_;
+    // Lock-free double buffer.  writeState_ packs the buffer selector
+    // (bit 63) and write index (bits 0-62) into one atomic so that
+    // producers obtain both from a single fetch_add — no buf/index race.
+    static constexpr size_t BUF_BIT = size_t(1) << 63;
```
ColinNV, line 88: Please use `buffer` instead of `buf`, and `index` instead of `idx`.

src/core/telemetry/telemetry.cpp
```diff
@@ -226,30 +234,28 @@ nixlTelemetry::updateMemoryDeregistered(uint64_t memory_deregistered) {
...
-               nixl_telemetry_category_t::NIXL_TELEMETRY_PERFORMANCE,
-               post_time.count());
+nixlTelemetry::addTransferComplete(std::chrono::microseconds post_time,
+                                   std::chrono::microseconds xfer_time,
+                                   bool is_write,
+                                   uint64_t bytes) {
+    const size_t old = writeState_.fetch_add(4);
+    const unsigned buf = (old & BUF_BIT) ? 1 : 0;
+    const size_t id = old & IDX_MASK;
+    if (id + 4 > eventBufferSize_) return;
```
ColinNV, line 244: This still produces bogus data.

ColinNV (changes requested):

src/core/telemetry/telemetry.h
```diff
@@ -83,9 +82,15 @@ class nixlTelemetry {
...
+    static constexpr size_t IDX_MASK = BUF_BIT - 1;
+    std::unique_ptr<nixlTelemetryEvent[]> eventBuffers_[2];
```
iyastreb, line 90: why not just std::vector<nixlTelemetryEvent>?
eventBufferSize_ would be redundant

src/core/telemetry/telemetry.cpp
```diff
@@ -170,17 +179,15 @@ void
...
                           uint64_t value) {
-    // agent can be multi-threaded
-    std::lock_guard<std::mutex> lock(mutex_);
-    events_.emplace_back(std::chrono::duration_cast<std::chrono::microseconds>(
-                             std::chrono::system_clock::now().time_since_epoch())
-                             .count(),
-                         category,
-                         event_name,
-                         value);
+    const size_t old = writeState_.fetch_add(1);
```
iyastreb, line 182: This reserves a slot to write a new event by producer, right?
Reservation does not mean that the event is actually written
What stops consumer from reading this reserved but uninitialized event?

src/core/telemetry/telemetry.cpp
```diff
@@ -170,17 +179,15 @@ void
...
+    const unsigned buf = (old & BUF_BIT) ? 1 : 0;
+    const size_t id = old & IDX_MASK;
+    if (id >= eventBufferSize_) return;
```
iyastreb, line 185: Could you please explain this check?
Do we just drop data silently?

src/core/telemetry/telemetry.cpp
```diff
@@ -129,23 +133,28 @@ nixlTelemetry::initializeTelemetry() {
...
-    for (auto &event : next_queue) {
-        // if full, ignore
-        exporter_->exportEvent(event);
+    for (size_t i = 0; i < numEvents; ++i) {
+        if (eventBuffers_[buf][i].eventName_[0] == '\0') continue;
+        exporter_->exportEvent(eventBuffers_[buf][i]);
+        eventBuffers_[buf][i].eventName_[0] = '\0';
     }
 
+    nextBufBit_ ^= BUF_BIT;
```
iyastreb, line 157: This operation can be reordered by compiler or CPU (since there are no load/store fences), and may be executed before the loop above

test/gtest/telemetry_test.cpp
```diff
@@ -138,13 +138,14 @@ TEST_F(telemetryTest, TransferBytesTracking) {
     EXPECT_NO_THROW(telemetry.updateErrorCount(nixl_status_t::NIXL_ERR_BACKEND));
```
iyastreb, line 138: lock free algorithms are very hard to get right, so I propose to add extensive tests to guarantee that it works as expected. For example we may have 8-32 producer threads constantly pushing some events with incremental ID in a loop, and consumer should validate them (no dirty reads etc)

PR #1431, opened 2026-03-17
libfabric: fix logging when it retries

fengjica:
Today when retries happen, at 1 per 100 attempts, it floods the log. This change makes exponentially backoff logging until INT_MAX / 2,  and then logging at regular interval because that's already bad enough to indicate a bigger issue.

## What?
Busy retrying at initialization can take 1 - 2 seconds on some platforms. It floods the log when that's done at 1 per 100 attempts. E.g.
I0317 18:48:10.397101  162355 libfabric_rail.cpp:1202] fi_read still retrying EAGAIN on rail 0 after 100 attempts
...
I0317 18:48:12.180008  162355 libfabric_rail.cpp:1202] fi_read still retrying EAGAIN on rail 0 after 6553600 attempts

## Why?
Makes one hard to scroll in logs.

## How?
Logging only at iterations of power of (100 *) 2 until INT_MAX /2 (over 2 billion). If it's still retrying after that point, just let it be.

## Summary by CodeRabbit

* **Refactor & Optimization**
  * Reduced logging verbosity during prolonged retry loops to minimize log spam and surface more actionable messages.
  * Centralized and standardized retry logging with an exponential backoff so entries appear less frequently and at predictable intervals.
  * Adjusted retry counter handling for more consistent and robust retry behavior, improving observability and stability.

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -709,6 +709,28 @@ nixlLibfabricRail::setXferIdCallback(std::function<void(uint32_t)> callback) {
     xferIdCallback = callback;
 }
 
+
+namespace {
+
+// Helper function to handle exponential backoff logging for retry operations
+inline void
+logRetryAttempt(unsigned int attempt,
+                unsigned int &log_threshold,
+                const char *operation_name,
+                uint16_t rail_id) {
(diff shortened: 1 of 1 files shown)


fengjica:
close this as it's not necessary any more. logging updated in https://github.com/ai-dynamo/nixl/pull/1462.

PR #1437, opened 2026-03-18
UTILS/CONFIG: Add config file reading.

ColinNV:
## What?
Add config file reading to the previously established config infrastructure.

## How?
Read a TOML config file from one of three possible places, if there is one, and look for config entries there when no environment variable can be found.

## Summary by CodeRabbit

* **New Features**
  * Added TOML-based configuration file support with automatic precedence: environment variables override user-level config (~/.nixl.cfg), which overrides system-level config (/etc/nixl.cfg).

* **Chores**
  * Updated build dependencies and configuration infrastructure.


src/utils/common/configuration.h
```diff
@@ -33,166 +36,284 @@
...
-    if (const auto opt = getenvOptional(env)) {
+    if (const auto opt = internal::getenvOptional(env)) {
+        internal::warnIgnoreToml(env);
         return traits<type>::convert(*opt);
     }
+
+    if (const auto view = internal::findTomlNode(env)) {
+        return traits<type>::convert(view);
+    }
     return std::nullopt;
```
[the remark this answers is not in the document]
ColinNV: The change from "env only" to "env or TOML" is the whole point of this PR.

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+            return toml::parse_file(file.native());
+        }
+        catch (const std::exception &e) {
+            internal::throw_runtime_error("Error ", e.what(), " reading TOML config file ", file);
+        }
+    }
+
+    [[nodiscard]] toml::table
+    readFile() {
+        // First choice, use $NIXL_CONFIG_FILE as config file if the environment variable is set.
```
guy-ealey-morag:
```suggestion
        // First choice, use environment variable as config file if the environment variable is set.
```

The name of the var is defined above, it's best not to repeat it.
Same for mentioning `~/.nixl.cfg` and `/etc/nixl.cfg` below

src/utils/common/configuration.h
```diff
@@ -33,166 +36,284 @@
...
         try {
             result = traits<std::decay_t<type>>::convert(*opt);
             return NIXL_SUCCESS;
         }
         catch (const std::exception &e) {
-            NIXL_DEBUG << "Unable to convert value '" << *opt << "' from environment variable '"
-                       << env << "' to target type " << typeid(type).name();
+            NIXL_DEBUG << "Unable to convert environment variable '" << env << "' to target type "
+                       << typeid(type).name();
+            return NIXL_ERR_MISMATCH;
```
guy-ealey-morag, line 256: If there is a mismatch the warning from `internal::warnIgnoreToml(env);` is still printed which may be confusing

src/utils/common/configuration.h
```diff
@@ -33,166 +36,284 @@
...
+            throw_runtime_error("Invalid TOML type '", view.type(), "' for integer");
+        }
 
-    [[nodiscard]] static int
-    base(const std::string &value) noexcept {
-        return isHex(value) ? 16 : 10;
-    }
+    private:
+        [[nodiscard]] static bool
+        fitsIn(const int64_t value) noexcept {
```
guy-ealey-morag: dead code?

src/utils/common/configuration.h
```diff
@@ -33,166 +36,284 @@
...
-    }
-};
+        [[nodiscard]] static const char *
+        start(const std::string &value) noexcept {
+            return value.data() + (isHex(value) ? 2 : 0);
+        }
+
+        template<typename T>
+        [[nodiscard]] static bool
+        fits(const T value) noexcept {
```
guy-ealey-morag: This is similar to `in_range` of C++20, consider renaming it

src/utils/common/configuration.h.orig
guy-ealey-morag, line 1: remove this file?

src/utils/common/configuration.h
```diff
@@ -22,8 +22,11 @@
 #include <charconv>
 #include <chrono>
 #include <cstdlib>
+#include <limits>
 #include <filesystem>
+#include <limits>
```
guy-ealey-morag, line 26: included twice

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+ * limitations under the License.
+ */
+
+#include "configuration.h"
+
+namespace nixl::config {
+
+const char *home_filename __attribute__((weak)) = ".nixl.cfg";
+const char *nixl_filename __attribute__((weak)) = "/etc/nixl.cfg";
+const char *nixl_variable __attribute__((weak)) = "NIXL_CONFIG_FILE";
```
guy-ealey-morag: What is the motivation to make these weak?

ColinNV: Didn't think much about it, but in this project we probably don't need to override these via the linker.

src/utils/common/configuration.h
ColinNV: Good point, only print the warning on success.

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+        if (nixl_variable != nullptr) {
+            if (const auto env = internal::getenvOptional(nixl_variable)) {
+                const std::filesystem::path file(std::filesystem::weakly_canonical(*env));
+                NIXL_DEBUG << "Reading nixl config file " << file << " from " << nixl_variable;
+                return readFile(file);
+            }
+        }
+
+        // Second choice, use ~/.nixl.cfg (default) as config file if the file exists.
```
guy-ealey-morag:
```suggestion
        // Second choice, use config file from the home directory if the file exists.
```

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+                    std::filesystem::weakly_canonical(home / home_filename);
+                if (std::filesystem::exists(file)) {
+                    NIXL_DEBUG << "Reading nixl config file " << file;
+                    return readFile(file);
+                }
+                NIXL_DEBUG << "Config file " << file << " does not exist";
+            }
+        }
+
+        // Third choice, use /etc/nixl.cfg (default) as config file if the file exists.
```
guy-ealey-morag: I'm not sure how to correct the text here, but it's best not to mention the full path in the comment

src/utils/common/configuration.h
```diff
@@ -33,166 +35,275 @@
...
-template<typename type, template<typename...> class traits = convertTraits>
+template<typename type, template<typename...> class traits = internal::convertTraits>
 [[nodiscard]] nixl_status_t
 getValueWithStatus(type &result, const std::string &env) {
-    if (const auto opt = getenvOptional(env)) {
+    if (const auto opt = internal::getenvOptional(env)) {
         try {
             result = traits<std::decay_t<type>>::convert(*opt);
+            internal::warnIgnoreToml(env);
+            return NIXL_SUCCESS;
```
guy-ealey-morag: minor: I would move these two lines to under the `catch` to emphasis that the `try` is on the `convert`

guy-ealey-morag:
Is it possible to add unit tests that create toml files and check for correct behavior?

ColinNV:
> Is it possible to add unit tests that create toml files and check for correct behavior?

Absolutely, once it's clear there will be no more changes to the semantics. Until then I did some smoke tests before opening the PR.

src/utils/common/configuration.h
```diff
@@ -33,166 +35,275 @@
...
-template<> struct convertTraits<std::chrono::milliseconds> {
-    [[nodiscard]] static std::chrono::milliseconds
-    convert(const std::string &value) {
-        return std::chrono::milliseconds(convertTraits<uint64_t>::convert(value));
-    }
-};
+        template<typename T>
+        [[nodiscard]] static bool
+        in_range(const T value) noexcept {
+            static_assert(std::is_signed_v<T>);
```
tomerg-nvidia, line 204: Why does T have to be signed?

ColinNV: This `T` is from the internal integer representation of tomlplusplus, which is signed, and is what the following code assumes; the `static_assert` documents and verifies this assumption.

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+            if (const auto env = internal::getenvOptional(nixl_variable)) {
+                const std::filesystem::path file(std::filesystem::weakly_canonical(*env));
+                NIXL_DEBUG << "Reading nixl config file " << file << " from " << nixl_variable;
+                return readFile(file);
+            }
+        }
+
+        // Second choice, use config file in home directory if directory and file exist.
+
+        if (home_filename && *home_filename) {
```
brminich: can we convert this check to assert?

test/unit/utils/common/meson.build
```diff
@@ -26,3 +26,9 @@ p2p_socket = executable('p2p_test',
             dependencies: [nixl_dep, nixl_infra, stream_interface] + cuda_dependencies,
             include_directories: [nixl_inc_dirs, utils_inc_dirs],
             install: true)
+
+configuration = executable('configucation',
```
brminich:
```suggestion
configuration = executable('configuration',
```

test/unit/utils/common/configuration.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+const std::string pid = std::to_string(::getpid());
+const std::string bool_name = "bool" + pid;
+const std::string number_name = "number" + pid;
+const std::string string_name = "string" + pid;
+
+} // namespace
+
+int
+main() {
```
brminich: can we test with just gtest?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    [[nodiscard]] toml::node_view<const toml::node>
+    findTomlNode(const std::string &path) {
+        return findTomlNode(toml::path(path));
+    }
+
+    void
+    warnIgnoreToml(const std::string &path) {
+        if (getConfig().at_path(toml::path(path))) {
+            NIXL_WARN << "Ignoring config file entry '" << path << "' for environment variable";
```
brminich: pls make it debug

src/utils/common/configuration.cpp
ColinNV: The symbols aren't weak anymore, the check can be removed.

test/unit/utils/common/configuration.cpp
ColinNV: I don't think so, if any other test directly or indirectly accesses the configuration first the config file reading will be done then, not now.

brminich: can you create a config file specific just for that gtest?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    [[nodiscard]] toml::table
+    readFile() {
+        // First choice, use config file path and name from environment variable if set.
+
+        if (nixl_variable != nullptr) {
+            if (const auto env = internal::getenvOptional(nixl_variable)) {
+                const std::filesystem::path file(std::filesystem::weakly_canonical(*env));
+                NIXL_DEBUG << "Reading nixl config file " << file << " from " << nixl_variable;
+                return readFile(file);
```
brminich: should we check whether this file exists?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,134 @@
...
+        catch (const std::exception &e) {
+            internal::throwRuntimeError("Error ", e.what(), " reading TOML config file ", file);
+        }
+    }
+
+    [[nodiscard]] toml::table
+    readFile() {
+        // First choice, use config file path and name from environment variable if set.
+
+        if (nixl_variable != nullptr) {
```
brminich: seems it can never be nullptr, why is this check needed?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,134 @@
...
+            const std::filesystem::path file =
+                std::filesystem::weakly_canonical(home / home_filename);
+            if (std::filesystem::exists(file)) {
+                NIXL_DEBUG << "Reading nixl config file " << file;
+                return readFile(file);
+            }
+            NIXL_DEBUG << "Config file " << file << " does not exist";
+        }
+
+        // Third choice, use global config file if it exists.
```
brminich, line 61: can we document this order somewhere?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,134 @@
...
+        NIXL_DEBUG << "Missing environment variable " << name;
+        return std::nullopt;
+    }
+
+    [[nodiscard]] std::string
+    getenvDefaulted(const std::string &name, const std::string &fallback) {
+        if (const char *value = std::getenv(name.c_str())) {
+            NIXL_DEBUG << "Found environment variable " << name;
+            return std::string(value);
+        }
```
brminich, line 101: can it call `getenvOptional` instead?

src/utils/common/configuration.cpp
```diff
@@ -0,0 +1,134 @@
...
+        NIXL_DEBUG << "Missing environment variable " << name;
+        return std::nullopt;
+    }
+
+    [[nodiscard]] std::string
+    getenvDefaulted(const std::string &name, const std::string &fallback) {
+        if (const char *value = std::getenv(name.c_str())) {
```
brminich, line 98: do they really need to be internal given that you are using them in gtest?
or is it a gtest too fine-grained?

src/utils/common/configuration.cpp
ColinNV: Right, it can't be 0 anymore.

src/utils/common/configuration.cpp
ColinNV: What would be the right place, the main README.md? Or a new file in the docs directory?

src/utils/common/configuration.cpp
ColinNV: If the user sets `NIXL_CONFIG_FILE` in the environment and the file doesn't exist then we'll get an exception:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error File could not be opened for reading reading TOML config file "/tmp/foo.toml"
```
I believe that's exactly what we want to happen.

src/utils/common/configuration.cpp
ColinNV: From a program logic point of view yes, but I didn't want two log messages when the default is used, and it didn't pass my threshold to factor out the common code.

src/utils/common/configuration.cpp
ColinNV: I want them `internal` because people should use the functions that also read from the config file.

I'm using them in gtest because I already had the tests from before config files and decided to keep them.

test/unit/utils/common/configuration.cpp
ColinNV: The problem is that when the same process access the config before the config file is created and the variable set, the file created by the testcase will not be used and the test will fail. However since, and as long as, we run each individual testcase separately it will work, so I've now moved the separate test into the gtests. Running the gtests with multiple tests being executed by the same process will now usually fail.

src/utils/common/configuration.cpp
brminich: probably main readme

test/gtest/configuration.cpp
```diff
@@ -215,4 +216,38 @@ TEST(Config, ConvertUnsigned) {
...
+        const auto value = nixl::config::getValue<bool>(bool_name);
+        EXPECT_TRUE(value);
+    }
+    {
+        const auto value = nixl::config::getValue<unsigned>(number_name);
+        EXPECT_EQ(value, 42);
+    }
+    {
+        const auto value = nixl::config::getValue<std::string>(string_name);
+        EXPECT_EQ(value, "hello");
```
brminich: can we also add a test, which would check the order of getting the value? I.e. set some env var also

guy-ealey-morag (approved):

tomerg-nvidia (approved):

brminich (approved):

PR #1438, opened 2026-03-18
Increased unit test coverage for OBJ plugin accelerated engine for Dell ObjectScale

jgoldsch12:
These tests require the cuobjclient library provided by CUDA 13.1 and a cufile.json configuration.

## What?
Adds unit test cases

## Why?
To increase unit test coverage and protect against regression

## Summary by CodeRabbit

* **Tests**
  * Expanded OBJ test suite with new vendor-accelerated RDMA scenarios, broader error-path and edge-case coverage, and a new shared test fixture and mocks for reuse.
  * Added a null-handling test for deregistration and improved parameterized instantiation to separate standard and vendor-specific cases.
  * Build config updated to include vendor-specific test unit conditionally when acceleration support is present.

*Note: No user-facing changes in this release.*


test/gtest/unit/obj/obj.cpp
```diff
@@ -759,4 +765,671 @@ TEST_F(objCrtTestFixture, MixedSizeThreshold) {
...
+    nixl_status_t status = objEngine_->registerMem(mem_desc, DRAM_SEG, metadata);
+    EXPECT_EQ(status, NIXL_ERR_NOT_SUPPORTED);
+    EXPECT_EQ(metadata, nullptr);
+
+    // Same check for VRAM_SEG
+    metadata = nullptr;
+    status = objEngine_->registerMem(mem_desc, VRAM_SEG, metadata);
+    EXPECT_EQ(status, NIXL_ERR_NOT_SUPPORTED);
+    EXPECT_EQ(metadata, nullptr);
+}
```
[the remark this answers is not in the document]
jgoldsch12: cuobjclient.h cannot be included in the test file because it transitively includes cufile.h 
(line 26 of the header), which is not available in the test build environment. The 
HAVE_CUOBJ_CLIENT macro is set when the cuobjclient-13.1 pkg-config package is found, but 
the test uses mocks and doesn't link or include the full cuObject SDK.

The existing code is already the correct approach:

  • The hardcoded constexpr size_t kMaxMemRegSize = 4ULL * 1024 * 1024 * 1024 matches the 
  library's #define CUOBJ_MAX_MEMORY_REG_SIZE (4ULL * 1024 * 1024 * 1024) exactly.
  • The comment explicitly documents the dependency: "This constant is defined in 
  cuobjclient.h; keep in sync if it changes."

test/gtest/unit/obj/obj.cpp
```diff
@@ -759,4 +765,671 @@ TEST_F(objCrtTestFixture, MixedSizeThreshold) {
     testMultiDescriptorWithSizes(NIXL_WRITE, 1048576, 6291456, "-crt-mixed");
 }
 
+#if defined HAVE_CUOBJ_CLIENT
```
aranadive: Maybe its easier to create a separate file with these cuobj Dell tests? The file only gets built if cuobj client library exists.

jgoldsch12: @aranadive I pushed a new change.  Can you take a look?  The main question is what we want to do with the CUOBJ dependent Accel config.

test/gtest/unit/obj/obj_test_base.h
```diff
@@ -0,0 +1,423 @@
...
+
+    // Dell-specific RDMA methods
+    void
+    putObjectRdmaAsync(std::string_view key,
+                       uintptr_t data_ptr,
+                       size_t data_len,
+                       size_t offset,
+                       std::string_view rdma_desc,
+                       put_object_callback_t callback) override {
+        if (rdma_desc.empty()) {
```
aranadive: Should this also reject xfers with data_len = 0 and offset != 0 similar to the dell s3 client?

jgoldsch12: addressed.  Thanks.

test/gtest/unit/obj/obj.cpp
aranadive: Its fine to gate it on the cuobj dependency for now.

aranadive (approved):

PR #1439, opened 2026-03-18
Support OS-assigned ports for metadata listener

guy-ealey-morag:
## What?
Allow NIXL agents to bind their metadata listener to an OS-assigned port (marked with `0`).
Update CI test scripts to use this mechanism instead of `get_next_tcp_port()`, and fetch the assigned port using a file that the selected port is written to.

## Why?
CI tests occasionally fail with `Address already in use` due to a race condition with `get_next_tcp_port()`. OS-assigned ports eliminate the race entirely.


src/utils/stream/metadata_stream.h
```diff
@@ -59,6 +59,11 @@ class nixlMDStreamListener: public nixlMetadataStream {
         nixlMDStreamListener(int port);
         ~nixlMDStreamListener();
 
+        int
+        getPort() const {
```
ColinNV: `[[nodiscard]]` and `noexcept`

examples/python/basic_two_peers.py
```diff
@@ -24,6 +24,12 @@ def parse_args():
         default="initiator",
         help="Local IP in target, peer IP (target's) in initiator",
     )
+    parser.add_argument(
```
ColinNV: `--listen-port-file` or something more descriptive?

src/api/cpp/nixl.h
```diff
@@ -56,6 +56,15 @@ class nixlAgent {
         nixlAgent(nixlAgent&&) noexcept = delete;
         nixlAgent &operator=(nixlAgent&&) noexcept = delete;
 
+        /**
+         * @brief  Get the actual port the metadata listener is bound to.
+         *         Useful when listenPort was set to 0 (OS-assigned port).
+         *
+         * @return The port number, or -1 if no listener is active
+         */
+        int
```
ColinNV: `[[nodiscard]]` and `noexcept` and I'd return a `std::uint16_t` with `0` for "not used" or better yet a `std::optional< std::uint16_t >`, though that's probably no fun with the bindings?

src/core/nixl_agent.cpp
```diff
@@ -231,6 +229,14 @@ nixlAgent::~nixlAgent() {
     }
 }
 
+int
```
ColinNV: `[[nodiscard]]` and `noexcept`

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,6 +86,16 @@ void nixlMDStreamListener::setupListener() {
         throw std::runtime_error("Failed to bind metadata listener socket");
     }
 
+    struct sockaddr_in bound_addr;
+    socklen_t addr_len = sizeof(bound_addr);
+    if (getsockname(socketFd, (struct sockaddr *)&bound_addr, &addr_len) == 0) {
+        port = ntohs(bound_addr.sin_port);
+    } else {
+        NIXL_PERROR << "getsockname failed after successful bind";
```
ColinNV: Is this needed in addition to the exception that will presumably be logged somewhere else?

src/api/cpp/nixl.h
guy-ealey-morag: I will change it to use uint16_t (without optional because of the bindings)

examples/python/basic_two_peers.py
guy-ealey-morag: I'll go with `listen-port-out` to make it obvious it's an output

src/api/cpp/nixl.h
```diff
@@ -56,6 +57,15 @@ class nixlAgent {
...
         nixlAgent &operator=(nixlAgent&&) noexcept = delete;
 
+        /**
+         * @brief  Get the actual port the metadata listener is bound to.
+         *         Useful when listenPort was set to 0 (OS-assigned port).
+         *
+         * @return The port number, or 0 if no listener is active
+         */
+        [[nodiscard]] uint16_t
+        getListenPort() const noexcept;
```
rakhmets: In any other case, I am definitely in favor of adding `[[nodiscard]]` and `noexcept`. But unfortunately, no existing method is marked with these specifiers. If I remember correctly, the idea was to fix it once in all places. In the meantime, add without these specifiers.
@yosefe do we still follow the approach?

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,6 +86,15 @@ void nixlMDStreamListener::setupListener() {
         throw std::runtime_error("Failed to bind metadata listener socket");
     }
 
+    struct sockaddr_in bound_addr;
+    socklen_t addr_len = sizeof(bound_addr);
+    if (getsockname(socketFd, (struct sockaddr *)&bound_addr, &addr_len) == 0) {
+        port = ntohs(bound_addr.sin_port);
+    } else {
+        closeStream();
+        throw std::runtime_error("Failed to retrieve bound port for metadata listener");
```
rakhmets: Maybe add strerror?

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,6 +86,15 @@ void nixlMDStreamListener::setupListener() {
         throw std::runtime_error("Failed to bind metadata listener socket");
     }
 
+    struct sockaddr_in bound_addr;
```
rakhmets:
```suggestion
    sockaddr_in bound_addr;
```

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,6 +86,15 @@ void nixlMDStreamListener::setupListener() {
...
+    socklen_t addr_len = sizeof(bound_addr);
+    if (getsockname(socketFd, (struct sockaddr *)&bound_addr, &addr_len) == 0) {
```
rakhmets:
```suggestion
    if (getsockname(socketFd, reinterpret_cast<sockaddr *>(&bound_addr), &addr_len) == 0) {
```

src/utils/stream/metadata_stream.h
```diff
@@ -59,6 +59,11 @@ class nixlMDStreamListener: public nixlMetadataStream {
         nixlMDStreamListener(int port);
         ~nixlMDStreamListener();
 
+        [[nodiscard]] int
+        getPort() const noexcept {
+            return port;
```
rakhmets: I think it would be better to change the type of the field to `uint16_t`.
Maybe add `TODO`?

src/utils/stream/metadata_stream.h
```diff
@@ -33,7 +33,7 @@
 
 class nixlMetadataStream {
     protected:
-        int                 port;
+        uint16_t port;
```
guy-ealey-morag, line 36: clang-format makes me keep it unaligned for some reason

rakhmets: Yes, it should not be aligned. The alignment was before the clang format was added to CI.

rakhmets (dismissed):

ColinNV (dismissed):

src/api/cpp/nixl.h
```diff
@@ -56,6 +57,15 @@ class nixlAgent {
...
 
+        /**
+         * @brief  Get the actual port the metadata listener is bound to.
+         *         Useful when listenPort was set to 0 (OS-assigned port).
+         *
+         * @return The port number, or 0 if no listener is active
+         */
+        uint16_t
+        getListenPort() const;
+
```
brminich: can we avoid extending API?

brminich (changes requested):

src/api/cpp/nixl.h
guy-ealey-morag: Currently the flow is:
- The agent get a port assigned
- The python script reads it using the API
- The python script writes it to a file
- The bash script reads the file and sets the port for the second script call

Do you have a suggestion how to avoid the API?

We could write the port to a file directly from the C++ code but I don't think it's very clean, and also would make it inconvenient for the C++ tests.

brminich: can we try to get it calling `lsof` os using some other tool?
I'd not want to introduce API for the sake of CI

ColinNV: A log message that includes the port number?

guy-ealey-morag: I will look into it

guy-ealey-morag: @brminich @ColinNV 
I reverted the API changes and used the `ss` cli command instead to find the assigned port

src/utils/stream/metadata_stream.h
```diff
@@ -56,9 +56,14 @@ class nixlMDStreamListener: public nixlMetadataStream {
         void            recvFromClients(int clientSocket);
 
     public:
-        nixlMDStreamListener(int port);
+        nixlMDStreamListener(uint16_t port);
```
rakhmets:
```suggestion
        explicit nixlMDStreamListener(uint16_t port) noexcept;
```

src/utils/stream/metadata_stream.h
```diff
@@ -42,7 +42,7 @@ class nixlMetadataStream {
         void closeStream();
 
     public:
-        nixlMetadataStream(int port);
+        nixlMetadataStream(uint16_t port);
```
rakhmets:
```suggestion
        explicit nixlMetadataStream(uint16_t port) noexcept;
```

src/utils/stream/metadata_stream.h
```diff
@@ -56,9 +56,14 @@ class nixlMDStreamListener: public nixlMetadataStream {
...
     public:
-        nixlMDStreamListener(int port);
+        nixlMDStreamListener(uint16_t port);
         ~nixlMDStreamListener();
 
+        uint16_t
+        getPort() const {
+            return port;
+        }
+
```
rakhmets:
```suggestion
```

rakhmets (dismissed):

ColinNV (dismissed):

rakhmets (dismissed):

guy-ealey-morag:
@brminich @ColinNV @rakhmets 
I found out that getting the listener port with `ss` was unreliable.
Instead I implemented a solution that writes the port to a file path given through an env var.
Please review again.

rakhmets (dismissed):

ColinNV (dismissed):

.gitlab/test_python.sh
```diff
@@ -93,19 +93,41 @@ python3 partial_md_example.py --init-port "$(get_next_tcp_port)" --target-port "
...
+    NIXL_MD_LISTENER_PORT_FILE="$port_fifo" \
+        python3 "$script" --mode="target" --ip=127.0.0.1 --port=0 &
+    target_pid=$!
+
+    if ! port=$(timeout 30 head -n1 "$port_fifo"); then
+        echo "Target (pid=$target_pid) failed to report port within 30s"
+        kill "$target_pid" 2>/dev/null
+        exit 1
+    fi
+    rm -f "$port_fifo"
```
roiedanino: I see why this works, but is there any reason not to use get_next_tcp_port from `scripts/common.sh`?

guy-ealey-morag: The usage of `get_next_tcp_port` caused port collisions in some cases (leading to statistical CI failures), this is solved by the changes in this PR

guy-ealey-morag:
Refactored to read the os-assigned port through the logs instead of using an env var.
@brminich

src/api/cpp/nixl_types.h
```diff
@@ -130,7 +130,7 @@ using nixl_notifs_t = std::unordered_map<std::string, std::vector<nixl_blob_t>>;
 /**
  * @brief A constant to define the default communication port.
  */
-constexpr int default_comm_port = 8888;
+constexpr uint16_t default_comm_port = 8888;
```
ColinNV: Also needs `inline`.

ColinNV: Actually, it looks like it's only used in one place; perhaps replace it with `8888` there and remove this variable?

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,14 +85,30 @@ void nixlMDStreamListener::setupListener() {
...
 
+    const bool os_assigned_port = (port == 0);
+    sockaddr_in bound_addr;
+    socklen_t addr_len = sizeof(bound_addr);
+    if (getsockname(socketFd, reinterpret_cast<sockaddr *>(&bound_addr), &addr_len) == 0) {
+        port = ntohs(bound_addr.sin_port);
+    } else {
+        closeStream();
+        throw std::runtime_error(
+            std::string("Failed to retrieve bound port for metadata listener: ") + strerror(errno));
```
ColinNV: When including an `errno` I like to have the exact name of the function that failed in the message.

src/api/cpp/nixl_types.h
guy-ealey-morag: I'll keep it because it's exposed as part of the api

tomerg-nvidia (dismissed):

src/utils/stream/metadata_stream.cpp
```diff
@@ -86,14 +85,31 @@ void nixlMDStreamListener::setupListener() {
...
     if (listen(socketFd, 128) < 0) {
         NIXL_PERROR << "Listening failed for stream Socket: "
                     << socketFd;
         closeStream();
         throw std::runtime_error("Failed to listen on metadata listener socket");
     }
-    NIXL_DEBUG << "MD listener is listening on port "
-               << port << "...";
+
+    const std::string log_msg = "MD listener is listening on port ";
```
ColinNV, line 107: The exception message above spells out `metadata` instead of abbreviating to `MD`.

.gitlab/test_cpp.sh
```diff
@@ -107,9 +107,9 @@ kill -s INT $telePID
 gtest-parallel --workers=1 --serialize_test_cases ./bin/gtest -- --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port"
 ./bin/test_plugin
 
-# Run NIXL client-server test
-nixl_test_port=$(get_next_tcp_port)
-parallel --line-buffer --halt now,fail=1 ::: "./bin/nixl_test target" "sleep 3 ; ./bin/nixl_test initiator" ::: "127.0.0.1 $nixl_test_port"
```
ColinNV, line 112: Now we're back to not detecting when the background instance fails?

src/utils/stream/metadata_stream.cpp
guy-ealey-morag: I kept the same texts as before, do you want to change them?

Alexey-Rivkin (dismissed):
Pls bump CI_IMAGE_TAG, then re-run with a `/build` comment.

.gitlab/test_cpp.sh
guy-ealey-morag: You're right, we need to design a better solution for this

guy-ealey-morag:
I reverted the last change because it wasn't robust enough, I'll work on improvements for another PR.

src/utils/stream/metadata_stream.cpp
ColinNV: Not necessarily in this PR.

ColinNV (approved):

tomerg-nvidia (approved):

brminich (approved):

PR #1455, opened 2026-03-24
Add Notif Callbacks

ColinNV:
## What?
Add configurable user-defined callbacks for received notifications.

The callbacks are called on the context of the thread that receives the notifications with the understanding that any non-trivial work needs to be handed off to a different thread (this leaves a choice of whether the inter-thread communication overhead is warranted).

A callback can be registered for a "tag", the prefix of the notification message (these prefixes of the notifications do not have a delimiter; the expected main use case is for a small number of equal-length prefixes to be registered).

The set of callbacks is determined at agent creation time as changing later is not a current requirement (this allows the data structure with the callbacks to be `const` meaning it never requires any locking).

In this PR the backend-specific part is implemented and tested for the UCX backend.

## Summary by CodeRabbit

* **New Features**
  * Notification callback system with prefix-based routing and configurable default handler.
  * Backend API exposes an optional notification-callback capability; UCX backend implements and uses it.
  * Configuration now carries default and prefix-specific notification callbacks into backend initialization.

* **Bug Fixes**
  * Backends are rejected when callbacks are configured but the backend lacks notification/callback support.

* **Tests**
  * Added tests validating callback registration, routing, default fallback, and threaded behavior.


src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+
+    void
+    setDefaultCallback(const nixl_notif_callback_t &callback) {
+        default_ = callback;
+    }
+
+    void
+    addCallback(const std::string &prefix, const nixl_notif_callback_t &callback) {
+        assert(callback);
+        assert(!prefix.empty());
```
guy-ealey-morag: This is public API so maybe it's best to always check these asserts (not only during debug)?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+        for (const auto &cb : callbacks_) {
+            if (message.compare(0, cb.prefix.size(), cb.prefix) == 0) {
+                cb.callback(std::move(remote), std::move(message));
+                return;
+            }
+        }
+        callDefault(std::move(remote), std::move(message));
+    }
+
+    size_t common_size_ = 0;
```
guy-ealey-morag:
```suggestion
    size_t commonSize_ = 0;
```

Should be camelCase

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+#include <cassert>
+#include <functional>
+#include <string>
+#include <utility>
+#include <vector>
+
+using nixl_notif_callback_t = std::function<void(std::string &&, std::string &&)>;
+
+struct nixlNotifCallback {
+    nixlNotifCallback(const std::string &prefix, const nixl_notif_callback_t &callback)
```
guy-ealey-morag, line 32: Pass args by value and use move to avoid copying when not needed

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+    callBinarySearch(std::string &&remote, std::string &&message) const {
+        if (message.size() >= common_size_) {
+            const std::string prefix = message.substr(0, common_size_);
+            const auto iter =
+                std::lower_bound(callbacks_.begin(),
+                                 callbacks_.end(),
+                                 message,
+                                 [&](const nixlNotifCallback &l, const std::string &r) {
+                                     return l.prefix < prefix;
+                                 });
```
guy-ealey-morag:
```suggestion
                std::lower_bound(callbacks_.begin(),
                                 callbacks_.end(),
                                 prefix,
                                 [&](const nixlNotifCallback &l, const std::string &r) {
                                     return l.prefix < r;
                                 });
```

I think this would be easier to understand

src/api/cpp/backend/notif_callbacks.h
guy-ealey-morag: rename to `commonPrefixSize_`?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+
+    void
+    call(std::string &&remote, std::string &&message) const {
+        if (callbacks_.empty()) {
+            callDefault(std::move(remote), std::move(message));
+        } else if (common_size_ > 0) {
+            callBinarySearch(std::move(remote), std::move(message));
+        } else {
+            callLinearScan(std::move(remote), std::move(message));
+        }
```
guy-ealey-morag, line 101: Is it possible to have the search functions return an iter and have the actual call done only in one place at the bottom of this func? (separate searching logic from calling logic, avoid `callDefault` from multiple places)

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,133 @@
...
+        } else {
+            common_size_ = 0;
+        }
+    }
+
+    void
+    call(std::string &&remote, std::string &&message) const {
+        if (callbacks_.empty()) {
+            callDefault(std::move(remote), std::move(message));
+        } else if (common_size_ > 0) {
```
guy-ealey-morag: I think it would be beneficial to add an explanation on why binary search can't be done when the prefixes are of different length

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1379,6 +1380,16 @@ int nixlUcxEngine::progress() {
  * Notifications
 *****************************************/
 
+nixlNotifCallbacks
+nixlUcxEngine::setDefaultCallback(nixlNotifCallbacks callbacks) {
+    if (!callbacks.hasDefaultCallback()) {
+        callbacks.setDefaultCallback([engine = this](std::string &&remote, std::string &&message) {
+            engine->appendNotif(std::move(remote), std::move(message));
```
guy-ealey-morag: In the case where someone adds a callback for some prefixes without a default callback, every notification that didn't match any callback will be appended with `appendNotif`.
Is it a use case that we expect would happen?
If not then we should consider failing if callbacks are added without a default

src/core/nixl_agent.cpp
```diff
@@ -356,7 +357,12 @@ nixlAgent::createBackend(const nixl_backend_t &type,
 
     if (backend->supportsRemote()) {
         if (!backend->supportsNotif()) {
-            NIXL_ERROR_FUNC << "backend '" << type << "' supportsRemote but not notifications";
+            NIXL_ERROR_FUNC << "backend '" << type << "' supports remote but not notifications";
+            return NIXL_ERR_BACKEND;
+        }
+
+        if (data->config_.notifCallbacks.hasAnyCallback() && !backend->supportsNotifCallback()) {
+            NIXL_ERROR_FUNC << "backend '" << type << "' does not support notif callbacks";
```
```suggestion
            NIXL_ERROR_FUNC << "backend '" << type << "' does not support notification callbacks";
```

test/gtest/notif_callbacks.cpp
```diff
@@ -0,0 +1,146 @@
...
+        promise.set_value(true);
+    });
+
+    agentPair agents(cfg);
+    agents.genNotif(message1);
+    const auto future = promise.get_future();
+    EXPECT_EQ(future.wait_for(std::chrono::milliseconds(5000)), std::future_status::ready);
+}
+
+TEST(NotifCallbacks, PrefixFixedSizeWithProgressThread) {
```
guy-ealey-morag:
```suggestion
TEST(NotifCallbacks, PrefixBinaryScanWithProgressThread) {
```

Consistent with other test name `PrefixLinearScanWithProgressThread`

test/gtest/notif_callbacks.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+TEST(NotifCallbacks, PrefixLinearScanWithProgressThread) {
+    std::promise<bool> promise;
+
+    nixlAgentConfig cfg;
+    cfg.useProgThread = true;
+    cfg.notifCallbacks.addCallback(
+        "aaa", [](std::string &&remote, std::string &&message) { ADD_FAILURE(); });
+    cfg.notifCallbacks.addCallback("notif", [&](std::string &&remote, std::string &&message) {
```
guy-ealey-morag:
```suggestion
    cfg.notifCallbacks.addCallback(message1_prefix, [&](std::string &&remote, std::string &&message) {
```

set the prefix in its own variable (define next to `message1`)

src/api/cpp/backend/notif_callbacks.h
ColinNV: _And_ it's not remotely performance critical.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Not _remotely_ performance critical, keep things simple?!

src/api/cpp/backend/notif_callbacks.h
ColinNV: Sure, though it's not much of a win.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Now that you say it :-)

src/plugins/ucx/ucx_backend.cpp
ColinNV: The implemented behaviour is what we want for now, all notifications that aren't intercepted by a custom callback should be handled as before.

guy-ealey-morag: Okay, is it something we could easily add to one of the tests?

ColinNV: Several tests already use `getNotifs()` with the UCX backend.

guy-ealey-morag: I mean testing that a notification that was matched with a callback doesn't show up in `getNotifs()`

ColinNV: Gotcha, done.

guy-ealey-morag (dismissed):

src/api/cpp/backend/backend_aux.h
```diff
@@ -53,6 +54,7 @@ class nixlBackendInitParams {
         nixlTime::us_t    pthrDelay;
         nixl_thread_sync_t syncMode;
         bool enableTelemetry_;
+        nixlNotifCallbacks notifCallbacks;
```
iyastreb, line 58: notifCallbacks_ ?

src/api/cpp/backend/backend_engine.h
```diff
@@ -112,6 +112,13 @@ class nixlBackendEngine {
         // pure virtual, and return errors, as parent shouldn't call if supportsNotif is false.
         virtual bool supportsNotif() const = 0;
 
+        // Determine whether a backend supports notification callbacks;
+        // meaningful only when supportsNotif() == true.
+        virtual bool
+        supportsNotifCallback() const {
```
iyastreb, line 117: These "supports" methods and overrides in the derived classes blows up the code.
I think better alternative is to:
- define enum nixlBackendCapability with bit masks entries
- Add bitmask capabilities_ field in the backend engine, that is initialized in ctor
- Have a single `bool supportsCapability(enum capa)` function in the nixlBackendEngine

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+#include <functional>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+using nixl_notif_callback_t = std::function<void(std::string &&, std::string &&)>;
+
+struct nixlNotifCallback {
+    nixlNotifCallback(const std::string &prefix, const nixl_notif_callback_t &callback)
```
iyastreb, line 32: maybe pass callback by value and move

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+        : prefix(prefix),
+          callback(callback) {}
+
+    std::string prefix;
```
iyastreb, line 36: both can be const?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+            commonPrefixSize_ = prefix.size();
+        } else if ((commonPrefixSize_ > 0) && (commonPrefixSize_ == prefix.size())) {
+            std::sort(callbacks_.begin(), callbacks_.end());
+        } else {
+            commonPrefixSize_ = 0;
+        }
+    }
+
+    void
+    call(std::string &&remote, std::string &&message) const {
```
iyastreb: just pass by value seems sufficient here

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+    std::string prefix;
+    nixl_notif_callback_t callback;
+};
+
+[[nodiscard]] inline bool
+operator<(const nixlNotifCallback &l, const nixlNotifCallback &r) noexcept {
+    return l.prefix < r.prefix;
+}
+
+class nixlNotifCallbacks {
```
iyastreb, line 61: I think some short documentation would be beneficial. For example describe callback semantics: is it allowed to register multiple callbacks with the same prefix, how many callbacks are called on call etc

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+
+    [[nodiscard]] iterator
+    findLinearScan(const std::string &message) const {
+        return std::find_if(callbacks_.begin(), callbacks_.end(), [&](const nixlNotifCallback &cb) {
+            return isPrefixOf(cb.prefix, message);
+        });
+    }
+
+    size_t commonPrefixSize_ = 0;
+    nixl_notif_callback_t default_;
```
iyastreb, line 182: do we need to store it separately?
Can't it be just a callback with empty prefix?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    [[nodiscard]] iterator
+    findBinarySearch(const std::string &message) const {
+        if (message.size() < commonPrefixSize_) {
+            return callbacks_.end();
+        }
+
+        const std::string prefix = message.substr(0, commonPrefixSize_);
```
iyastreb: can we do it without string allocation?
Maybe using string_view

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+        if (!callback) {
+            throw std::runtime_error(
+                "Empty notification callback function not allowed with prefix");
+        }
+
+        if (prefix.empty()) {
+            throw std::runtime_error("Empty notification callback prefix is not allowed");
+        }
+
+        if (hasOverlap(prefix)) {
```
iyastreb, line 118: why hasOverlap is needed, can't we just reuse findCallback?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+    [[nodiscard]] static bool
+    isPrefixOf(const std::string &prefix, const std::string &string) noexcept {
+        return (string.size() >= prefix.size()) &&
+            (std::memcmp(prefix.data(), string.data(), prefix.size()) == 0);
+    }
+
+    [[nodiscard]] bool
+    hasOverlap(const std::string &prefix) const noexcept {
+        for (const auto &cb : callbacks_) {
+            if (isPrefixOf(prefix, cb.prefix) || isPrefixOf(cb.prefix, prefix)) {
```
iyastreb, line 132: why do we check twice?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+    void
+    addCallback(const std::string &prefix, const nixl_notif_callback_t &callback) {
+        checkNewCallback(prefix, callback);
+
+        callbacks_.emplace_back(prefix, callback);
+
+        if (callbacks_.size() == 1) {
+            commonPrefixSize_ = prefix.size();
+        } else if ((commonPrefixSize_ > 0) && (commonPrefixSize_ == prefix.size())) {
+            std::sort(callbacks_.begin(), callbacks_.end());
```
iyastreb: do we need to update commonPrefixSize_ here?

src/api/cpp/backend/backend_engine.h
ColinNV: I fully agree, but we already have `supportsNotif()` and have to keep it for backwards compatibility; I didn't want to have different mechanisms for different capabilities...?

src/api/cpp/backend/notif_callbacks.h
ColinNV: This is called like 10 times at startup, i.e. performance is irrelevant. I wanted to keep it simple.

src/api/cpp/backend/notif_callbacks.h
ColinNV: This is the case where the new prefix size is the same as all previous ones' sizes.

src/api/cpp/backend/notif_callbacks.h
ColinNV: To detect both `f` vs. `foo` and `foo` vs. `f`.

src/api/cpp/backend/notif_callbacks.h
ColinNV: It breaks the common size thing; I also find it clearer this way, and we can skip the find operation when the vector is empty which will presumably still be a very common case.

src/api/cpp/backend/backend_aux.h
ColinNV: These are public members of a struct/record.

src/api/cpp/backend/notif_callbacks.h
ColinNV: It's a struct/record with only public data members, I usually keep things mutable.

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            commonPrefixSize_ = 0;
+        }
+    }
+
+    void
+    call(std::string &&remote, std::string &&message) const {
+        if (const iterator iter = findCallback(message); iter != callbacks_.end()) {
```
tvegas1: would it make sense to use two lines here?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    [[nodiscard]] iterator
+    findBinarySearch(const std::string &message) const {
+        if (message.size() < commonPrefixSize_) {
+            return callbacks_.end();
+        }
+
+        const std::string prefix = message.substr(0, commonPrefixSize_);
+        const auto iter = std::lower_bound(
```
tvegas1, line 162: why not using some map for such search with known prefix size?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,169 @@
...
+            [](const nixlNotifCallback &l, const std::string &r) { return l.prefix < r; });
+
+        if ((iter == callbacks_.end()) || (prefix != iter->prefix)) {
+            return callbacks_.end();
+        }
+        return iter;
+    }
+
+    [[nodiscard]] iterator
+    findLinearScan(const std::string &message) const {
```
tvegas1, line 175: in general needs documentation for user. maybe even setting time complexity expectations wrt this case with linear search with many callbacks added.

src/api/cpp/backend/notif_callbacks.h
ColinNV: To check in both directions.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Simpler lines vs. reduced scope of variables...

src/api/cpp/backend/notif_callbacks.h
ColinNV: For better locality.

src/api/cpp/backend/backend_engine.h
iyastreb: Well, if we keep it like this, then it will never improve
Maybe we can introduce this enum now, and in the subsequent PR we can refactor existing NIXL backends to use enum instead of `supportsRemote, supportsLocal, supportsNotif`.
We can still keep `supportNotif` in the API for backward compatibility, but no need to override it anymore in NIXL engines, and all new capabilities are implemented in a clean way

src/api/cpp/backend/notif_callbacks.h
iyastreb: imo it must be immutable, because this struct is used in MT environment.
These notifications are created in one thread, but might be handled in another one. And making immutable data structure means we can safely use it concurrently, because no one can change it

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,187 @@
...
+    // there is no delimiter.
+    // When registering multiple callbacks with prefixes, these prefixes
+    // must not be prefixes of each other.
+    // The expected primary use case of all prefixes having the same length
+    // is optimized better (logarithmic) than the generic case (linear).
+    // If use cases with a large set of registered prefixes arise things
+    // can be optimized further.
+
+    void
+    addCallback(const std::string &prefix, const nixl_notif_callback_t &callback) {
```
iyastreb: Is it supposed to be used concurrently? Or all the callbacks are expected to be set up during init phase?
If it might be updated concurrently, then mutex is needed
If not, then maybe mention it in the doc?

src/api/cpp/backend/notif_callbacks.h
iyastreb: quite unusual in NIXL code, but not a big deal

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,187 @@
...
+
+#include <algorithm>
+#include <cstring>
+#include <functional>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+using nixl_notif_callback_t = std::function<void(std::string &&, std::string &&)>;
```
iyastreb: maybe simply:
```suggestion
using nixl_notif_callback_t = std::function<void(std::string, std::string)>;
```

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,187 @@
...
+    [[nodiscard]] iterator
+    findLinearScan(const std::string &message) const {
+        return std::find_if(callbacks_.begin(), callbacks_.end(), [&](const nixlNotifCallback &cb) {
+            return isPrefixOf(cb.prefix, message);
+        });
+    }
+
+    size_t commonPrefixSize_ = 0;
+    nixl_notif_callback_t default_;
+    std::vector<nixlNotifCallback> callbacks_;
```
iyastreb, line 183: Maybe I'm missing something, but I don't really understand the semantics of prefixes.
Why can't we just use std::unordered_map<string, callback> and have a constant lookup time? What use cases it does not cover?

src/api/cpp/backend/notif_callbacks.h
ColinNV: The `nixlNotifCallbacks` instance that might be used by multiple threads _is_ declared `const`, which the applies to all instances of `nixlNotifCallback` it contains.

src/api/cpp/backend/notif_callbacks.h
ColinNV: The agent receives a `nixlNotifCallbacks` instance during construction, it is not changed later. I can add a clarifying comment.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Moving a `std::string` is not free, this function does not have many call sites but will be called relatively frequently.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Ok, perhaps it's more fitting in this code base to separate.

src/api/cpp/backend/notif_callbacks.h
ColinNV: For the general case, we can't use a map because the prefix is not delimited in the message, i.e. we can't extract the prefix and then go searching. In the fixed-length case I didn't want to use a hash map because it only supports heterogeneous lookup from C++20 on, the hash map has a larger constant overhead with the hash (and either the hash is simple, so we don't know whether we are always hitting the linear case, or it's complex, then it adds overhead), and we are talking about like 10 or so entries typically.

src/api/cpp/backend/notif_callbacks.h
ColinNV: Should we not keep the fast path efficient?

src/api/cpp/backend/backend_engine.h
brminich: it's api, minimal change is preferred. Let's not introduce unrelated API changes in this PR

src/api/cpp/backend/notif_callbacks.h
iyastreb: we should, I just assume that moving string is free, but I don't have a strong opinion

src/api/cpp/backend/notif_callbacks.h
iyastreb: Ok, I see, but in this case binary lookup is redundant according to this design
Let's remove it then and keep only linear?

src/api/cpp/backend/notif_callbacks.h
ColinNV: The `std::move()` is free, but the move-constructor isn't.

src/api/cpp/backend/notif_callbacks.h
ColinNV: I see your point, but it's simple enough and even for 10-20 entries the binary will be faster. For now I wanted the common size case to be reasonably fast without too much work, and the other case to at least work correctly.

guy-ealey-morag (dismissed):

aranadive:
@ColinNV please add nixlbench support as well in the PR.

tvegas1 (dismissed):

brminich:
> @ColinNV please add nixlbench support as well in the PR.

let's target it for a separate PR, as this one is already quite large

src/api/cpp/nixl_params.h
```diff
@@ -68,6 +71,15 @@ struct nixlAgentConfig {
      */
     std::chrono::microseconds etcdWatchTimeout = kDefaultEtcdWatchTimeout;
 
+    /**
+     * @var Callback functions to be called on received notifs.
+     *      The map keys are prefixes, when a received notification
+     *      begins with a prefix the corresponding function will be
+     *      called. Not all backends support notification callbacks.
+     *      The empty prefix can be used to set a default handler.
```
mkhazraee, line 79: Maybe say "otherwise, they can be fetched by getNotifs" or something on that line to clarify the previous path is not blocked.

I wonder if we should add if a remote backend supports notifications but not tags, and this field is not empty, it would result in an error. The behavior makes sense, just wondering if we should put some comments here or becomes too long.

mkhazraee:
LGTM as long as the polling getNotifs is the fallback for a notification that didn't match (I think it is). Other than that just a minor comment to update a comment about what that map means.

mkhazraee (dismissed):

src/api/cpp/nixl_params.h
ColinNV: Comment improved.

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,191 @@
...
+// must not be prefixes of each other.
+// The expected primary use case of all prefixes having the same length
+// is optimized better (logarithmic) than the generic case (linear).
+// If use cases with a large set of registered prefixes arise things
+// can be optimized further.
+
+// The default callback is invoked when no callback with a matching
+// prefix was registered.
+
+class nixlNotifCallbacks {
```
rakhmets, line 61: What do you think about removing implementation from the public header? Wouldn't it be better to have an interface here?

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,191 @@
...
+public:
+    nixlNotifCallbacks() = default;
+
+    [[nodiscard]] bool
+    hasAnyCallback() const noexcept {
```
rakhmets: I haven't found the usage of the method? Is it needed? Can we remove it, or you have plans to use it in another PR?

src/api/cpp/nixl_types.h
```diff
@@ -304,4 +306,17 @@ using nixl_xfer_telem_t = nixlXferTelemetry;
...
+ *
+ *        The first string is the name of the remote agent.
+ *        The second string is the received notification.
+ *        Notification callbacks are called on the thread that received the notif,
+ *        usually the progress thread. Any non-trivial tasks MUST be handed off to
+ *        a separate thread by the callback. In particular callbacks SHOULD NOT call
+ *        functions on any NIXL agent, and MUST NOT call functions on the agent that
+ *        received the notif for which they were called.
+ */
+using nixl_notif_callback_t = std::function<void(std::string &&, std::string &&)>;
```
rakhmets: Wouldn't it better to pass a structure instead? So the code will document the parameters, not the comment.

src/api/cpp/backend/notif_callbacks.h
ColinNV: I'm so used to templated code that this is often my default :-) It's also simpler and gives the compiler the possibility to inline all functions. Do you think it needs to change?

src/api/cpp/backend/notif_callbacks.h
ColinNV: Good catch, it was needed but the last change made it more or less unnecessary. I't still keep it for completeness and future use.

src/api/cpp/backend/notif_callbacks.h
rakhmets: What worries me first of all is that we may face the same problems as in other APIs. When every implementation change requires API changes. But on the other hand, it's hard for me to understand if this would look alien in the current state of the API. Therefore, this comment is more of a question than a call to action.

guy-ealey-morag (dismissed):

test/gtest/notif_callbacks.cpp
```diff
@@ -0,0 +1,191 @@
...
+            const auto status = agent.loadRemoteMD(md, name);
+            EXPECT_EQ(status, NIXL_SUCCESS);
+            EXPECT_EQ(name, src_name);
+        }
+    }
+
+    nixlAgent agent;
+    nixlBackendH *backend = nullptr;
+    nixl_mem_list_t memories;
+    nixl_b_params_t params;
```
rakhmets: These are local vars of `testAgent::createBackend`.

test/gtest/notif_callbacks.cpp
```diff
@@ -0,0 +1,191 @@
...
+        EXPECT_EQ(args.remoteAgent, name1);
+        EXPECT_EQ(args.notifMessage, message1);
+        promise.set_value(true);
+    });
+    cfg.notifCallbacks.try_emplace("zzzzzzz", [](nixlNotifCallbackArgs &&) { ADD_FAILURE(); });
+
+    agentPair agents(cfg);
+    agents.genNotif(message1);
+    const auto future = promise.get_future();
+    EXPECT_EQ(future.wait_for(std::chrono::milliseconds(5000)), std::future_status::ready);
```
rakhmets: Maybe `using namespace std::chrono_literals` and `5'000ms`?

src/api/cpp/backend/notif_callbacks.h
rakhmets: I would prefer to remove the unused method. Or at least add usage to the tests.

yosefe (commented):
API LGTM

test/gtest/notif_callbacks.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+const std::string name1 = "Agent";
+const std::string name2 = "Guard";
+
+struct testAgent {
+    testAgent(const std::string &name, const nixlAgentConfig &cfg) : agent(name, cfg) {}
+
+    void
+    createBackend() {
+        nixl_mem_list_t memories;
```
guy-ealey-morag, line 37:
```suggestion
        nixl_mem_list_t mems;
```

From what I found this var is always called `mems` across the codebase, I would rename for consistency

guy-ealey-morag (dismissed):

src/api/cpp/backend/notif_callbacks.h
```diff
@@ -0,0 +1,186 @@
...
+
+        if (commonPrefixSize_ > 0) {
+            return findBinarySearch(message);
+        } else {
+            return findLinearScan(message);
+        }
+    }
+
+    [[nodiscard]] iterator
+    findBinarySearch(const std::string &message) const {
```
iyastreb, line 156: As you described we expect a very low number of callbacks, so maybe keep just a linear scan?

iyastreb (dismissed):

tvegas1 (dismissed):

mkhazraee (dismissed):

mkhazraee (approved):

mkhazraee (dismissed):

tvegas1 (dismissed):

guy-ealey-morag (dismissed):

PR #1114, opened 2025-12-12
Fix issue related to Mooncake Backend data corruption in CI 

magicYang1573:
## What?
Fix data corruption bug in mooncake backend. Mooncake backend test in CI is passed now.

## Why?
Mooncake backend cannot pass NIXL CI due to data corruption issue. 

## How?
When NIXL opens segment in Mooncake TransferEngine, the metadata in the last segment allocation and deallocation is not released. An additional flush is added.

PR #1493, opened 2026-04-02
fix(mooncake): fix segment handle leak and improve transfer status handling

staryxchen:
## What?

Fix two bugs in the Mooncake backend plugin:

1. **Segment handle leak in `disconnect()` and destructor**: `loadRemoteConnInfo()` calls `openSegment()` and stores the returned `segment_id`, but `disconnect()` was a no-op and the destructor never called `closeSegment()`. Each connected remote agent leaked one segment handle for the lifetime of the engine.

2. **Silent success on terminal transfer failures in `checkXfer()`**: `STATUS_CANCELED` and `STATUS_TIMEOUT` are both terminal states in Mooncake's transfer engine with no possibility of recovery, but they were not handled in `checkXfer()` and fell through to return `NIXL_SUCCESS`, silently hiding the underlying failure. The loop also had a correctness issue where an early `return NIXL_IN_PROG` could discard a previously observed failure.

## Why?

- The segment handle leak grows with the number of distinct remote agents connected over the engine lifetime, which is relevant in dynamic inference clusters where prefill/decode workers are added and removed.

- `STATUS_CANCELED` is set when an RDMA endpoint is reset or destroyed (e.g. remote disconnect, link error). `STATUS_TIMEOUT` is set by the worker monitor thread when a slice exceeds the configured timeout (default 10 s), after which the endpoint is disabled. Treating these as success causes the caller to silently receive corrupted or incomplete data.

> **Note:** A related data-corruption fix (using `openSegmentNoCache` instead of `openSegment` in `loadRemoteConnInfo`) was proposed in #1114 but has not yet been merged. That fix is orthogonal to this PR and should be merged independently.

## How?

**Fix 1 — segment handle leak (`disconnect()` and destructor):**
- Implement `disconnect()`: call `closeSegment()` on the stored `segment_id` and remove the agent from `connected_agents_`.
- In the destructor, iterate over any remaining `connected_agents_` and call `closeSegment()` as a safety net for cases where `disconnect()` was not called explicitly.

**Fix 2 — `checkXfer()` status handling:**
- Add `STATUS_CANCELED` and `STATUS_TIMEOUT` to the failure condition alongside `STATUS_FAILED`.
- Refactor the scan loop to collect `has_failed` and `has_pending` independently across all requests before deciding the return value, so a failure observed before a still-pending request is not discarded by an early return.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable cleanup of remote agent connections so associated segments are closed during shutdown.
  * Disconnects now close agent segments immediately upon removal, preventing lingering connections.
  * Transfer-tracking now correctly distinguishes pending vs terminal failures (including cancellations and timeouts), avoiding false "in progress" reports and ensuring accurate completion status.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/nixl/pull/1493" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


staryxchen:
I noticed that #1114, which fixes a related data corruption issue (using `openSegmentNoCache` instead of `openSegment` in `loadRemoteConnInfo`), has been open for a while without activity.

Merging that fix would also unlock the Mooncake unit test (`mooncake_backend_test`) that is currently commented out in CI ([`test_cpp.sh`](https://github.com/ai-dynamo/nixl/blob/main/.gitlab/test_cpp.sh#L103)), providing better test coverage for future Mooncake backend changes.

If the community would find it helpful, I'd be happy to incorporate that fix into this PR as well — just let me know!

staryxchen:
Hi @brminich @ovidiusm 👋

Just a friendly ping — all CI checks are passing on this PR. It fixes a segment handle leak in the Mooncake backend's `disconnect()` and improves `checkXfer()` status aggregation.

Would either of you have a chance to review when you get a moment?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -299,13 +315,24 @@ nixl_status_t
...
+        //   configured timeout (default 10 s). The endpoint is then disabled,
+        //   making retries futile. Terminal, no retry.
+        if (rc || status.status == STATUS_FAILED || status.status == STATUS_CANCELED ||
+            status.status == STATUS_TIMEOUT) {
             has_failed = true;
-        else if (status.status == STATUS_PENDING || status.status == STATUS_WAITING)
-            return NIXL_IN_PROG;
+        } else if (status.status == STATUS_PENDING || status.status == STATUS_WAITING) {
+            has_pending = true;
+        }
```
guy-ealey-morag, line 353: `STATUS_INVALID` is not handled here and it falls through.
I also suggest adding an `else` block to catch any unhandled status values for robustness.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -114,10 +118,22 @@ nixlMooncakeEngine::connect(const std::string &remote_agent) {
...
+        if (it == connected_agents_.end()) {
+            return NIXL_SUCCESS;
+        }
+        segment_id = it->second.segment_id;
+        // Erase under the lock so concurrent postXfer/genNotif calls will see
+        // the agent as gone and return NIXL_ERR_INVALID_PARAM before using
+        // the segment handle.
+        connected_agents_.erase(it);
+    }
+    closeSegment(engine_, segment_id);
```
guy-ealey-morag: check return value?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -97,6 +97,10 @@ nixlMooncakeEngine::getSupportedMems() const {
 
 // Through parent destructor the unregister will be called.
 nixlMooncakeEngine::~nixlMooncakeEngine() {
+    // Close all open remote segments before destroying the engine.
+    for (auto &[agent, info] : connected_agents_) {
+        closeSegment(engine_, info.segment_id);
```
guy-ealey-morag: check return value?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -114,10 +118,22 @@ nixlMooncakeEngine::connect(const std::string &remote_agent) {
     return NIXL_SUCCESS;
 }
 
-// TODO We purposely set this function as empty.
-// Will be changed to follow NIXL's paradigm after refactoring Mooncake Transfer Engine.
 nixl_status_t
 nixlMooncakeEngine::disconnect(const std::string &remote_agent) {
+    segment_id_t segment_id;
```
guy-ealey-morag, line 126: I see that in `mooncake_backend.h:116` the type used for `segment_id` is `int`.
Should we change it there to `segment_id_t` as well?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -114,10 +118,22 @@ nixlMooncakeEngine::connect(const std::string &remote_agent) {
...
+    {
+        const std::lock_guard<std::mutex> lock(mutex_);
+        auto it = connected_agents_.find(remote_agent);
```
guy-ealey-morag:
```suggestion
        const auto it = connected_agents_.find(remote_agent);
```

staryxchen:
Hi @guy-ealey-morag Thanks for the thorough review! All five comments have been addressed in commit 5fff1ba. Please take another look when you get a chance. Thanks!

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -97,6 +97,10 @@ nixlMooncakeEngine::getSupportedMems() const {
 
 // Through parent destructor the unregister will be called.
 nixlMooncakeEngine::~nixlMooncakeEngine() {
+    // Close all open remote segments before destroying the engine.
+    for (const auto &[agent, info] : connected_agents_) {
+        (void)closeSegment(engine_, info.segment_id);
```
guy-ealey-morag: warn on failure? (It's preferred not to ignore error codes)

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -114,10 +118,22 @@ nixlMooncakeEngine::connect(const std::string &remote_agent) {
...
+        if (it == connected_agents_.end()) {
+            return NIXL_SUCCESS;
+        }
+        segment_id = it->second.segment_id;
+        // Erase under the lock so concurrent postXfer/genNotif calls will see
+        // the agent as gone and return NIXL_ERR_INVALID_PARAM before using
+        // the segment handle.
+        connected_agents_.erase(it);
+    }
+    (void)closeSegment(engine_, segment_id);
```
guy-ealey-morag: return error on failure?

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -114,10 +118,22 @@ nixlMooncakeEngine::connect(const std::string &remote_agent) {
...
     return NIXL_SUCCESS;
```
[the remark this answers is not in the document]
guy-ealey-morag: Is this race relevant @staryxchen ?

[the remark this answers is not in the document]
staryxchen: > Is this race relevant @staryxchen ?

@guy-ealey-morag I think the race exists in theory but is safe in practice for two reasons:
1. The higher-level callers (e.g. nixl_agent, nixl_listener) are expected to drain in-flight transfers before calling disconnect(). So under normal usage the window doesn't arise.
2. Even if closeSegment() is called while transfers are still in-flight, the mooncake transfer engine marks those transfers as STATUS_CANCELED (the underlying RDMA endpoint gets reset/destroyed). Our checkXfer() already treats STATUS_CANCELED as a terminal failure, so the caller will see NIXL_ERR_BACKEND rather than a hang or use-after-free.

That said, a fully robust solution would require a large refactoring and probably better suited for a follow-up, not this bug-fix PR.

src/plugins/mooncake/mooncake_backend.cpp
```diff
@@ -136,6 +157,15 @@ nixlMooncakeEngine::loadRemoteConnInfo(const std::string &remote_agent,
...
+    // Close the previous segment if this agent was already connected,
+    // so we don't leak the old handle on reconnect.
+    auto it = connected_agents_.find(remote_agent);
+    if (it != connected_agents_.end()) {
+        if (closeSegment(engine_, it->second.segment_id)) {
+            NIXL_WARN << "loadRemoteConnInfo: failed to close old segment " << it->second.segment_id
+                      << " for agent " << remote_agent;
+        }
+    }
     connected_agents_[remote_agent].segment_id = segment_id;
```
guy-ealey-morag: With a small refactor it's possible to avoid the second lookup:

```
    auto [it, inserted] = connected_agents_.try_emplace(remote_agent);
    if (!inserted) {
        if (closeSegment(engine_, it->second.segment_id)) {
            NIXL_WARN << "loadRemoteConnInfo: failed to close old segment " << it->second.segment_id
                      << " for agent " << remote_agent;
        }
    }
    it->second.segment_id = segment_id;
```

guy-ealey-morag (approved):

staryxchen:
Hi @guy-ealey-morag, there are two failed CI jobs, but I can't see the detail. Could you check if this is related to this PR?

guy-ealey-morag:
> Hi @guy-ealey-morag, there are two failed CI jobs, but I can't see the detail. Could you check if this is related to this PR?

It looks like it's failing due to an ongoing IT issue, I will restart the tests when it's fixed

PR #1496, opened 2026-04-06
fix(OBJ) Make OBJ plugin queryMem launch all HeadObject requests in parallel

jgoldsch12:
Fixes #1494 

DefaultObjEngineImpl::queryMem checks object existence by calling checkObjectExists synchronously in a loop — one S3 HeadObject request per descriptor, waiting for each to complete before issuing the next. For a list of N descriptors, this serializes N network round-trips to S3.

The OBJ plugin already uses async patterns for putObjectAsync and getObjectAsync, but checkObjectExists had no async counterpart.

Fix: Add checkObjectExistsAsync to the iS3Client interface with implementations for both the standard S3 and S3 CRT clients. Refactor queryMem to launch all HeadObject requests in parallel using futures/ promises, then wait for all to complete. Remove the now-dead synchronous checkObjectExists.

Tested using nixl_object_test.  With 256 objects the queryMem phase took 300ms in the current code.  After this fix, it took 15ms.

## Summary by CodeRabbit

* **New Features**
  * Object existence checks are now asynchronous and deliver results via callbacks.
  * Query timing is displayed for object queries.

* **Bug Fixes / Behavior**
  * Remote request errors during existence checks are reported as request failures instead of throwing.
  * Bulk queries run concurrent async existence checks and surface backend errors when checks fail.

* **Tests**
  * Added comprehensive async tests and enhanced test harness to simulate delays, per-key outcomes, ordering, and executor fallback.


src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -158,25 +159,95 @@ DefaultObjEngineImpl::deregisterMem(nixlBackendMD *meta) {
...
+        std::shared_ptr<std::atomic<bool>> completed;
+    };
+
+    std::vector<std::future<void>> futures;
+    futures.reserve(descs.descCount());
+    std::vector<PerDescriptorState> states;
+    states.reserve(descs.descCount());
+    std::atomic<bool> has_error{false};
+
+    constexpr auto kQueryTimeout = std::chrono::seconds(30);
```
ovidiusm, line 181: Minor: Should this be configurable?

ovidiusm (dismissed):

src/plugins/obj/s3/engine_impl.cpp
```diff
@@ -158,25 +159,95 @@ DefaultObjEngineImpl::deregisterMem(nixlBackendMD *meta) {
...
+                    catch (...) {
+                    }
+                }
+            }
+        }
         NIXL_ERROR << "Failed to query memory: " << e.what();
         return NIXL_ERR_BACKEND;
     }
 
+    // Wait for all async operations with timeout fallback
```
vvenkates27, line 229: Is there a use-after-free race in the exception path in the drain loop?

The drain loop waits with wait_for(kQueryTimeout) and, on timeout, calls states[j].completed->exchange(true). But exchange only wins the "ownership" race atomically — it does not stop a callback that already passed its own exchange check but hasn't yet written to resp[i].      
  Scenario:       
                                                                                                                                                                                                                                                                                        
  1. Callback calls completed->exchange(true) → returns false (callback owns it, will write)                                                                                                                                                                                            
  2. wait_for times out before the promise is set
  3. Catch block calls completed->exchange(true) → returns true (callback already owns it), so the catch block skips the set_value() branch                                                                                                                                             
  4. Catch block returns NIXL_ERR_BACKEND, destroying the stack frame — resp and has_error are freed                                                                                                                                                                                    
  5. Callback writes to now-freed resp[i] → UB/crash   

We probably need to wait in the cathc block?
                                                                                                                                                                                                                                                                                        
  // In catch block, after all exchange() calls:                                                                                                                                                                                                                                        
  for (size_t j = 0; j < futures.size(); ++j)
      futures[j].wait(); // blocking — but exchange(true) already set, callback will set_value() soon         
Or maybe a temp shared_vec for resp might be easier to handle such races?

src/plugins/obj/s3_crt/client.cpp
```diff
@@ -137,17 +138,48 @@ awsS3CrtClient::getObjectAsync(std::string_view key,
...
 }
 
-bool
-awsS3CrtClient::checkObjectExists(std::string_view key) {
-    Aws::S3Crt::Model::HeadObjectRequest request;
-    request.WithBucket(bucketName_).WithKey(Aws::String(key));
+void
+awsS3CrtClient::checkObjectExistsAsync(std::string_view key, check_object_callback_t callback) {
+    // Heap-allocate the request for the same reason as other async methods:
+    // the SDK stores a raw pointer to it used in callbacks that fire after this function returns
```
vvenkates27, line 144: The comment says "the SDK stores a raw pointer to it used in callbacks." For HeadObjectAsync specifically, the CRT SDK copies the request internally before the call returns (unlike some lower-level CRT bindings). If confirmed, the heap allocation is wasted overhead on every    
  existence check. Please verify against the SDK version in use — the standard S3 client's checkObjectExistsAsync passes the request as a local stack variable without heap allocation.

vvenkates27 (commented):
Thanks for the PR. The idea of parallel. querying of headObject looks good to me. Just need to make sure we don't have races. Added a couple on this regard.

src/plugins/obj/s3_crt/client.cpp
jgoldsch12: I've addressed this comment.  You are correct.

src/plugins/obj/s3/engine_impl.cpp
jgoldsch12: Thank you for catching this.  Yes, this race exists and I've added the defensive code and tests.

vvenkates27 (approved):

aranadive (approved):

src/plugins/obj/s3/engine_impl.cpp
aranadive: Maybe as a follow up if query is really seems laggy.

PR #1497, opened 2026-04-06
libfabric: Update control messages from SEND/RECV to WRITE

anshumang:
## What?
Update libfabric backend control messages (notifications) from SEND/RECV (`fi_senddata`/`fi_recvmsg`) to RDMA WRITE with immediate data (`fi_writedata`).

Also fixes a bug in `test/nixl/agent_example.cpp` where `getPluginParams` was hardcoded to `"UCX"` instead of using the CLI backend argument.

## Why?
The SEND/RECV notification path requires pre-posting 1024 receive buffers per rail, consuming receive queue entries and adding complexity to the completion routing. RDMA WRITE with immediate data:
- Eliminates pre-posted receive buffers for notifications
- Unifies the completion path (both data transfers and notifications now arrive via `FI_REMOTE_WRITE`)
- Reduces receive queue pressure on EFA

## How?
**Receiver side:** Each rail allocates a notification receive buffer (ring buffer: 1024 slots × 8KB) registered with `FI_REMOTE_WRITE`. The buffer address and rkey are exchanged during connection metadata serialization (SerDes). `processRemoteWriteCompletion` now dispatches on `msg_type` in immediate data — `MSG_TRANSFER` for data, `MSG_NOTIFICTION` for notifications.

**Sender side:** `postControlMessage` uses `fi_writedata` to write serialized notification data into a remote slot (indexed by `xfer_id % NUM_SLOTS`), replacing `fi_senddata`.

**Cleanup:** Pre-posted recv pool removed. `FI_RECV` completion path kept as legacy fallback with warning.


dmvevents:
Filed #1656 with a related concern: on EFA SRD, `fi_writedata` (RDMA WRITE *with immediate data*) is silently emulated as a `send/recv` pair because EFA SRD does not support the 32-bit immediate-data field on `IBV_WR_RDMA_WRITE`. Wire-level effect: `tx_bytes` increments, but `rdma_write_bytes` HW counter does not. Verified today on 2× p5.48xlarge with EFA installer 1.48.0 / libfabric 2.4.0amzn3.0 — `tx_bytes +544 KB`, `rdma_write_bytes +0` on both pods after a cross-node FI_EP_RDM transfer.

That means the unification this PR introduces ("both data transfers and notifications now arrive via `FI_REMOTE_WRITE`") will land on EFA RDM as "both data transfers and notifications now go through the same SEND/RECV emulation pool" — which doubles the pressure on the receiver buffer pool that already exhausts at ~240 in-flight ops, rather than reducing receive queue pressure as the PR body intends.

Suggest changing the unified primitive from `fi_writedata` to `fi_write` (no immediate) on the data path, with flow control / notification routing carried via a separate small-message channel (or via the legacy `fi_recvmsg` notification path retained as a configurable fallback). On P5+ EFA, `fi_write` maps directly to `IBV_WR_RDMA_WRITE` and is native — so this would actually deliver the "unified FI_REMOTE_WRITE" goal at the wire level.

Full reproducer + cross-references (Monarch PR #3391 runtime-detect precedent, ofiwg/libfabric#12253) in #1656. Happy to send a follow-up patch on top of this PR once a direction is agreed — flagging now since it's still draft.

cc @anshumang

PR #1502, opened 2026-04-07
NIXLBENCH: Implement runtime without ETCD

ColinNV:
## What?
Implements a "runtime" for NIXLBench that allows running two instances with direct socket communication instead of ETCD

## Why?
Provides a rapid solution for running NIXLBench without ETCD; the desired target solution of using NIXL's metadata exchange mechanisms will be implemented after the necessary groundwork is finished.

## How?
Uses ASIO for a simple two-instance "runtime". Designed to allow both instances to use the same command line (first attempts to `bind()` to the supplied address, if that fails it does a `connect()` instead). Updates the test script to also run all two-instance tests with the new runtime.

TODO: Update the documentation once the code is stable.

## Summary by CodeRabbit

* **New Features**
  * Added an ASIO runtime backend with CLI flags for address/port; ASIO linked into non-CUDA builds.

* **Tests**
  * Expanded/reordered test matrix to run ASIO two-worker scenarios before etcd; added parallel ASIO and etcd two-worker runners.

* **Bug Fixes**
  * Always perform runtime/barrier synchronization and corrected role/synchronization behavior for storage/backends.

* **Chores**
  * CI/build now installs ASIO system dependency and updated CI image tag.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/nixl/pull/1502" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -458,12 +474,19 @@ xferBenchConfig::loadParams(void) {
...
+            if (etcd_endpoints.empty()) {
+                // For non-storage backends, set default ETCD endpoint
+                etcd_endpoints = "http://localhost:2379";
+                std::cout << "Using default ETCD endpoint for non-storage backend: "
+                          << etcd_endpoints << std::endl;
+            }
+        } else if (runtime_type == xferbench_rt_asio) {
+            std::cout << "Using address " << asio_address << " port " << asio_port
+                      << " for ASIO runtime" << std::endl;
+        }
```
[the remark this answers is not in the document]
ColinNV: The lower-case `std::string_view` is the preferred style for new code.

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -0,0 +1,317 @@
...
+
+    void recvHead() {
+        asio::async_read(*socket_, asio::buffer(&head_, sizeof(head_)), [this](const asio::error_code &ec, const std::size_t done) {
+            if (ec.value()) {
+                throw std::runtime_error("ASIO Read head failed: " + ec.message());
+            }
+            temp_.type = asio_msg_type_t(head_ >> 56);
+            temp_.data.clear();
+            temp_.data.resize(head_ & 0x1ffffffffull);
+            recvData();
```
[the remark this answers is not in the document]
ColinNV: This is intentional to not allocate enormous chunks of memory; the bitmask might be further reduced; expected use-cases are all very far below this value.

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -64,6 +65,7 @@
 
 // Runtime types
 #define XFERBENCH_RT_ETCD "ETCD"
+constexpr inline std::string_view xferbench_rt_asio = "ASIO";
```
guy-ealey-morag: I find it weird that `XFERBENCH_RT_ETCD` and `xferbench_rt_asio` use different standards, can we make them the same? (either define or constexpr)

benchmark/nixlbench/src/worker/worker.cpp
```diff
@@ -96,16 +106,19 @@ static xferBenchRT *createRT(int *terminate) {
     }
 #endif
 
-    std::cerr << "Invalid runtime: " << xferBenchConfig::runtime_type << std::endl;
+    if ((total == 2) && (xferBenchConfig::runtime_type == xferbench_rt_asio)) {
+        return new xferBenchAsioRT(xferBenchConfig::asio_address, xferBenchConfig::asio_port);
+    }
+
+    std::cerr << "Invalid runtime: " << xferBenchConfig::runtime_type << " for total number "
+              << total << " -- only ETCD runtime supports more than 2" << std::endl;
```
guy-ealey-morag:
```suggestion
    if (xferBenchConfig::runtime_type == xferbench_rt_asio) {
        if (total != 2) {
            // error
        }
        return new xferBenchAsioRT(xferBenchConfig::asio_address, xferBenchConfig::asio_port);
    }

    // revert to original error
```

I think it's best to have the new error only under the `xferbench_rt_asio` case, it's cleaner and would print a better message when giving a runtime that doesn't exist.

benchmark/nixlbench/src/worker/worker.cpp
```diff
@@ -128,7 +141,8 @@ xferBenchWorker::xferBenchWorker() {
     int rank = rt->getRank();
 
     // For storage backends without ETCD, always act as initiator
```
guy-ealey-morag: Update text to say something like "When using ETCD runtime.." to stress it's runtime specific

benchmark/nixlbench/meson.build
```diff
@@ -169,7 +172,7 @@ configure_file(
     install_dir: get_option('includedir') / 'nixlbench'
 )
 
-deps = [nixl_lib, nixl_build, nixl_serdes, openmp_dep, gflags_dep, tomlplusplus_dep]
+deps = [nixl_lib, nixl_build, nixl_serdes, openmp_dep, gflags_dep, tomlplusplus_dep, asio_dep]
```
guy-ealey-morag, line 175: Does it make asio a requirement?
We can make it optional by implementing logic similar to `etc_available`

benchmark/nixlbench/src/worker/worker.cpp
```diff
@@ -96,16 +106,19 @@ static xferBenchRT *createRT(int *terminate) {
...
+    }
+
+    std::cerr << "Invalid runtime: " << xferBenchConfig::runtime_type << " for total number "
+              << total << " -- only ETCD runtime supports more than 2" << std::endl;
     exit(EXIT_FAILURE);
 }
 
-int xferBenchWorker::synchronize() {
-    // For storage backends without ETCD, no synchronization needed
-    if (xferBenchConfig::isStorageBackend() && xferBenchConfig::etcd_endpoints.empty()) {
```
guy-ealey-morag, line 106: Why is it okay to remove this condition?
Should it be updated like the other condition to: 
```
xferBenchConfig::isStorageBackend() && xferBenchConfig::etcd_endpoints.empty() && (xferBenchConfig::runtime_type == XFERBENCH_RT_ETCD)
```

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -0,0 +1,338 @@
...
+    recvHead() {
+        asio::async_read(*socket_,
+                         asio::buffer(&head_, sizeof(head_)),
+                         [this](const asio::error_code &ec, const std::size_t done) {
+                             if (ec.value()) {
+                                 throw std::runtime_error("ASIO Read head failed: " + ec.message());
+                             }
+                             temp_.type = asio_msg_type_t(head_ >> 56);
+                             temp_.data.clear();
+                             temp_.data.resize(head_ & 0x1ffffffffull);
```
guy-ealey-morag: The size is encoded in the first 56 bits but this mask is only 33 bits

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -0,0 +1,338 @@
...
+        if (acceptor_) {
+            acceptor_->async_accept(*sock, handler);
+        } else {
+            sock->async_connect(endpoint_, handler);
+        }
+        timer_.async_wait(timeout);
+    }
+
+    [[nodiscard]] std::unique_ptr<asio::ip::tcp::acceptor>
+    attemptAcceptor() noexcept {
```
guy-ealey-morag:
```suggestion
    attemptAcceptor() {
```

Can `open` `set_option` or `listen` throw exceptions?

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -120,6 +123,15 @@ NB_ARG_STRING(etcd_endpoints,
               "",
               "ETCD server endpoints for communication (optional for storage backends)");
 
+NB_ARG_STRING(asio_address,
+              "127.0.0.1",
+              "Address for direct socket communication for 2 instances with ASIO runtime");
+
+// Should be UINT16 but that's not available
```
guy-ealey-morag, line 130: I found that it should possible to implement something like this to force the input to be valid:

```
NB_ARG_UINT32(asio_port,
              12345,
              "Port for direct socket communication for 2 instances with ASIO runtime");

static bool validateAsioPort(const char *flagname, std::uint32_t value) {
    if (value <= std::numeric_limits<std::uint16_t>::max()) {
        return true;
    }
    std::cerr << "Invalid value for --" << flagname << ": " << value
              << " (must be <= 65535)" << std::endl;
    return false;
}
DEFINE_validator(asio_port, &validateAsioPort);
```

benchmark/nixlbench/src/runtime/asio_runtime.h
ColinNV: I'm hoping somebody can confirm that 24bits are enough, then we can change to a 32bit integer for `head_`. This was just for now to prevent allocating terabytes of memory.

benchmark/nixlbench/meson.build
ColinNV: I don't think this requirement is a problem, it just needs libasio-dev package installed. As I understand it, the annoyance with ETCD wasn't so much the compile-time dependency as the run-time dependency on the ETCD server.

benchmark/nixlbench/src/runtime/asio_runtime.h
ColinNV: On second thought I'll just change it, from what I see this is used for at most a few KB.

ColinNV: Changing to 32bits.

benchmark/nixlbench/src/utils/utils.cpp
ColinNV: Works, thanks.

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -215,7 +217,12 @@ class xferBenchAsioRT : public xferBenchRT {
 
     void
     postSend(const asio_msg_type_t type, const void *data, const std::size_t size) {
-        const std::size_t head = size | (size_t(type) << 56);
+        if (size > 0x00ffffff) {
+            throw std::runtime_error("Runtime message size " + std::to_string(size) +
+                                     " exceeds 16MB-1 limit");
+        }
+
+        const std::uint32_t head = size | (std::uint32_t(type) << 24);
```
guy-ealey-morag: small suggestion:
It would be more robust to define `24` and the mask `0x00ffffff` in a `constexpr` at the top to not have them used as literals in two places. Maybe even calculate the mask from the number of bits instead of using a literal.

guy-ealey-morag (dismissed):

guy-ealey-morag (dismissed):

guy-ealey-morag (dismissed):

benchmark/nixlbench/src/worker/worker.cpp
ColinNV: When the `if` fires the `rt` is just the dummy rt that does nothing, no harm in calling its dummy `barrier()` function.

.gitlab/test_nixlbench.sh
```diff
@@ -66,9 +66,19 @@ else
...
+s.bind((\"\", 0))
+print(s.getsockname()[1])
+s.close()
+"
+}
+
 run_nixlbench_two_workers_asio() {
     args="$@"
-    asio_port=$(get_next_tcp_port)
+    asio_port=$(get_random_tcp_port)
```
guy-ealey-morag, line 81: Is it possible to have the port auto-assigned inside nixlbench instead of having to choose a random port here?
(This could be the default behavior when the `--asio-port` arg is missing)

ColinNV: Not without making everything a lot more complicated -- there is no other communication channel between the two instances. One could start using a file again, but that would also mean giving the two instances different command line parameters. As implemented they both try to `bind()` to the same address, only one will succeed (one will fail either because the address was already bound to by the other instance, or, when running on another node, because the machine doesn't have the address to bind to) and the other will `connect()` instead.

guy-ealey-morag (dismissed):

.gitlab/build.sh
```diff
@@ -101,6 +101,7 @@ else
                                  libprotobuf-dev \
                                  libcpprest-dev \
                                  libaio-dev \
+                                 libasio-dev \
```
aranadive: Why is this required if asio is already a subproject in nixlbench?

ColinNV: NIXL and NIXLBench have ASIO as a subproject to fall back on, if it's already available they don't need to use the fallback. Do you think it's better to remove (again) and rely on the subprojects?

aranadive: To me relying only on one version (either from subprojects or distro package) makes sense. NIXL relies only on subproject asio, lets do the same here.

ColinNV: Done.

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -0,0 +1,346 @@
...
+
+        try {
+            result->bind(endpoint_);
+        }
+        catch (...) {
+            std::cout << "ASIO runtime bind() error -- using connect() instead" << std::endl;
+            return {};
+        }
+
+        result->listen();
```
aranadive: Why isnt this inside a try block? This could fail with address is in use for the second worker since REUSEADDR is set to true (bind will be successful) - two nixlbench instances could be running on same node.

aranadive: Lets address this separately as a bug.

aranadive: CI is failing because of this, so needs to be fix in this PR.

ColinNV: The `REUSEADDR` for TCP shouldn't allow two sockets to bind to the same address, unlike UDP. I was expecting the `bind()` to fail when two instances are on the same host, not the `listen()`. Moved the `listen()` inside the try-catch block now.

guy-ealey-morag (dismissed):

.ci/jenkins/lib/build-matrix.yaml
```diff
@@ -43,7 +43,7 @@ env:
   TEST_TIMEOUT: 30
   UCX_TLS: "^shm"
   STORAGE_DRIVER: 'overlay'
-  CI_IMAGE_TAG: "20260414-3"
+  CI_IMAGE_TAG: "20260415-1"
```
aranadive: Is this change to ci-image-tag required now if the build script update was dropped?

benchmark/nixlbench/src/worker/worker.cpp
```diff
@@ -97,16 +106,22 @@ createRT(std::atomic<int> *terminate) {
     }
 #endif
 
+    if (xferBenchConfig::runtime_type == XFERBENCH_RT_ASIO) {
+        if (total != 2) {
+            std::cerr << "Invalid total " << total << " for ASIO runtime -- supports only 2"
+                      << std::endl;
```
aranadive, line 112: We should exit with failure here too?

ColinNV: Yes, already added.

benchmark/nixlbench/src/runtime/asio_runtime.h
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    int
+    barrier(const std::string &barrier_id) override {
+        const std::string barrier_name = barrier_id + "/" + std::to_string(++barrier_);
+        postSend(asio_msg_type_t::BARRIER, barrier_name.data(), barrier_name.size());
+        recvWait(asio_msg_type_t::BARRIER,
+                 [&](const std::string &data) { return data == barrier_name; });
+        return 0;
+    }
```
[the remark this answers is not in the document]
aranadive: I think this is fine for the simpler socket-based runtime.

aranadive (approved):

PR #1510, opened 2026-04-09
libfabric: fix active rail tracking

fengjica:
Fix active rail tracking in libfabric backend.

## What?
Today there is no reference counting to correctly keep track of the status of rails. In addition to fixing that, I add a unit test with some code refactoring to avoid duplicating the common test code.

## Why?
It is a bug without this mechanism.

## Summary by CodeRabbit

* **New Features**
  * Active-rail tracking now uses per-rail reference counts; rails become inactive only when their count reaches zero.

* **Tests**
  * Added a comprehensive unit test covering refcount transitions, edge cases, memory-registration interactions, and multi-rail behavior.

* **Chores**
  * Added shared test stubs/helpers and a new test executable to support the refcounting tests.

### DIFF
--- a/src/utils/libfabric/libfabric_rail_manager.cpp
+++ b/src/utils/libfabric/libfabric_rail_manager.cpp
@@ -712,9 +712,11 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
             for (size_t j = 0; j < i; ++j) {
                 const size_t cleanup_idx = selected_rails[j];
                 if (mr_list_out[cleanup_idx]) {
-                    rails_[cleanup_idx]->deregisterMemory(mr_list_out[cleanup_idx]);
+                    if (rails_[cleanup_idx]->deregisterMemory(mr_list_out[cleanup_idx]) ==
+                        NIXL_SUCCESS) {
+                        decRailActive(cleanup_idx);
+                    }
                     mr_list_out[cleanup_idx] = nullptr;
-                    decRailActive(cleanup_idx);
                 }
             }
             return NIXL_ERR_INVALID_PARAM;
@@ -731,9 +733,11 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
             for (size_t j = 0; j < i; ++j) {
                 const size_t cleanup_idx = selected_rails[j];
                 if (mr_list_out[cleanup_idx]) {
-                    rails_[cleanup_idx]->deregisterMemory(mr_list_out[cleanup_idx]);
+                    if (rails_[cleanup_idx]->deregisterMemory(mr_list_out[cleanup_idx]) ==
+                        NIXL_SUCCESS) {
+                        decRailActive(cleanup_idx);
+                    }
                     mr_list_out[cleanup_idx] = nullptr;
-                    decRailActive(cleanup_idx);
                 }
             }
             return status;
@@ -772,11 +776,12 @@ nixlLibfabricRailManager::deregisterMemory(const std::vector<size_t> &selected_r
 
         if (mr_list[rail_idx]) {
             nixl_status_t status = rails_[rail_idx]->deregisterMemory(mr_list[rail_idx]);
-            if (status != NIXL_SUCCESS) {
+            if (status == NIXL_SUCCESS) {
+                decRailActive(rail_idx);
+            } else {
                 NIXL_ERROR << "Failed to deregister memory on rail " << rail_idx;
                 overall_status = status;
             }
-            decRailActive(rail_idx);
         }
     }
 
@@ -889,9 +894,6 @@ nixlLibfabricRailManager::postControlMessage(ControlMessageType msg_type,
 
     NIXL_DEBUG << "Sending control message type " << msg_type_value << " agent_idx=" << agent_idx
                << " XFER_ID=" << xfer_id << " imm_data=" << imm_data << " on rail " << rail_id;
-
-    // Mark rail 0 as active so its CQ gets progressed
-    incRailActive(rail_id);
 
     // Use rail 0 for notifications
     nixl_status_t status = rails_[rail_id]->postSend(imm_data, dest_addr, req);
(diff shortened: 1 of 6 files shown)


src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -741,7 +741,7 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
         key_list_out[rail_idx] = key;
 
         // Mark rail as active for progress tracking optimization
-        markRailActive(rail_idx);
+        incRailActive(rail_idx);
```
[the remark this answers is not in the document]
fengjica: Fixed.

test/unit/utils/libfabric/libfabric_mock_stubs.h
```diff
@@ -0,0 +1,261 @@
...
+ */
+
+/*
+ * Shared libfabric stub ops for unit tests that need to construct a
+ * nixlLibfabricRailManager without real hardware.  Include this header
+ * in exactly one translation unit per test executable (it contains
+ * definitions, not just declarations).
+ */
+#ifndef NIXL_TEST_LIBFABRIC_MOCK_STUBS_H
+#define NIXL_TEST_LIBFABRIC_MOCK_STUBS_H
```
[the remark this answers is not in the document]
fengjica: Fixed.

test/unit/utils/libfabric/rail_active_refcount_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+    nixlLibfabricRailManager mgr(0);
+    TEST_ASSERT(mgr.getNumRails() == NUM_FAKE_RAILS,
+                "expected " + std::to_string(NUM_FAKE_RAILS) + " rails");
+
+    int res;
+    if ((res = testIncDecBasic(mgr)) != 0) return res;
+    if ((res = testDecNonExistent(mgr)) != 0) return res;
+    if ((res = testIncInvalidRailId(mgr)) != 0) return res;
+    if ((res = testClearActiveRails(mgr)) != 0) return res;
+    if ((res = testMultipleRailsIndependent(mgr)) != 0) return res;
```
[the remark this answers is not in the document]
fengjica: fixed.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -741,7 +743,7 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
...
```
anshumang, line 796: Since rail 0 is already unconditionally polled in `progressActiveRails()` and CQ completions don't distinguish 
control sends, do we need this `incRailActive()` here?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -774,7 +776,7 @@ nixlLibfabricRailManager::deregisterMemory(const std::vector<size_t> &selected_r
                 NIXL_ERROR << "Failed to deregister memory on rail " << rail_idx;
                 overall_status = status;
             }
-            markRailInactive(rail_idx);
+            decRailActive(rail_idx);
```
anshumang: If `deregisterMemory()` fails and the MR is still live, should we skip the `decRailActive()` here to avoid dropping the rail from the active set prematurely?

anshumang (changes requested):

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -1131,30 +1135,31 @@ nixlLibfabricRailManager::deserializeRailEndpoints(
...
-nixlLibfabricRailManager::markRailInactive(size_t rail_id) {
+nixlLibfabricRailManager::decRailActive(size_t rail_id) {
     std::lock_guard<std::mutex> lock(active_rails_mutex_);
-    size_t erased = active_rails_.erase(rail_id);
-    if (erased > 0) {
-        NIXL_DEBUG << "Marked rail " << rail_id
-                   << " as inactive (total active: " << active_rails_.size() << ")";
+    auto it = active_rails_.find(rail_id);
+    if (it != active_rails_.end()) {
+        it->second--;
```
yexiang-aws: nit: refcount`it->second` is size_t, if there's a bug over decrement it, it might underflow, do you want to add an assert `refcount==0` there before decrementing it?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -714,6 +714,7 @@ nixlLibfabricRailManager::registerMemory(void *buffer,
                 if (mr_list_out[cleanup_idx]) {
                     rails_[cleanup_idx]->deregisterMemory(mr_list_out[cleanup_idx]);
                     mr_list_out[cleanup_idx] = nullptr;
+                    decRailActive(cleanup_idx);
```
yexiang-aws: both here and line 736, we would need to check the status of `deregisterMemory` and if it failed, we should not decrement

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: After looking at the code, I see that another incRailActive(0) in postControlMessage() is actually unnecessary, and will indefinitely increment it in every postControlMessage(). Removed that one.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Makes sense.

Fixed in the next revision.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Fixed.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica: Make sense.

Fixed.

anshumang (approved):

yexiang-aws (approved):

PR #1516, opened 2026-04-12
Fix unbounded growth of in-memory telemetry event buffer

BatshevaBlack:
The events_ vector grew without limit between periodic flushes. Enforce the configured buffer size (NIXL_TELEMETRY_BUFFER_SIZE) by dropping new events when the buffer is full (drop-newest policy).

## What?
Enforce the configured buffer size (`NIXL_TELEMETRY_BUFFER_SIZE`, default 4096) on the in-memory `events_` vector in `nixlTelemetry`, dropping new events when the buffer is full.

## Why?
The `events_` vector in `nixlTelemetry` grows without bound between periodic flushes. Both `updateData()` and `addXferTime()` unconditionally call `emplace_back()` with no size check. Under burst conditions (or if events are produced faster than the flush interval drains them), this leads to unbounded memory growth. The `buffer_size` config was read but only forwarded to the exporter — it was never used to limit the in-memory staging vector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry now strictly honors the configured buffer capacity; once full, additional events are dropped instead of growing the buffer.
  * Related performance/transfer events are now blocked from being enqueued when doing so would exceed capacity, preventing buffer overruns and reducing unexpected memory growth.

### DIFF
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -98,6 +98,8 @@ nixlTelemetry::initializeTelemetry() {
         throw std::invalid_argument("Telemetry buffer size cannot be 0");
     }
 
+    maxBufferedEvents_ = buffer_size;
+
     const std::optional<std::string> exporter_name = getExporterName();
 
     if (!exporter_name) {
@@ -172,6 +174,9 @@ nixlTelemetry::updateData(const std::string &event_name,
                           uint64_t value) {
     // agent can be multi-threaded
     std::lock_guard<std::mutex> lock(mutex_);
+    if (events_.size() >= maxBufferedEvents_) {
+        return;
+    }
     events_.emplace_back(std::chrono::duration_cast<std::chrono::microseconds>(
                              std::chrono::system_clock::now().time_since_epoch())
                              .count(),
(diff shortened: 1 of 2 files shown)


src/core/telemetry/telemetry.cpp
```diff
@@ -172,6 +174,9 @@ nixlTelemetry::updateData(const std::string &event_name,
                           uint64_t value) {
     // agent can be multi-threaded
     std::lock_guard<std::mutex> lock(mutex_);
+    if (events_.size() >= maxBufferedEvents_) {
+        return;
+    }
```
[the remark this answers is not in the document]
BatshevaBlack: we'll implement a counter for lossed events later on, after we fix the perf gap we have. will be done in a different PR

src/core/telemetry/telemetry.cpp
ColinNV: Isn't this the same as `reserve(maxBufferedEvents_)`?

src/core/telemetry/telemetry.cpp
```diff
@@ -98,6 +98,8 @@ nixlTelemetry::initializeTelemetry() {
         throw std::invalid_argument("Telemetry buffer size cannot be 0");
     }
 
+    maxBufferedEvents_ = buffer_size;
```
ColinNV: We've now got `buffer_size`, `maxBufferedEvents_`, `DEFAULT_TELEMETRY_BUFFER_SIZE` and `getMaxEventsBuffered()` as names for the same thing...

src/core/telemetry/telemetry.cpp
BatshevaBlack: @ColinNV 
fixed thanks!

ColinNV (approved):

ovidiusm (approved):

alexanderbilk (approved):

PR #1517, opened 2026-04-12
nixl_ep: Add elastic test to CI

lishapira:
Add elastic EP test for expansion, fault tolerance, and contraction.

1. Build nixl_ep when GPUs are available by adding -Dbuild_nixl_ep=true in build.sh.
2. Run the elastic EP test in test_python.sh with a 4-process plan (5-minute timeout) that validates three scenarios in a single run:
- Expansion: 2 ranks -> 4 ranks
- Fault tolerance: rank 2 killed, remaining ranks continue
- Contraction: 4 ranks -> 2 ranks
3. Add silent-failure checks in elastic.py:
- Assert that rank failures only occur for ranks marked for kill in the plan (catches unexpected crashes).
- Verify the SIGTERM exit count matches the plan's expected kills (catches cases where the kill mechanism fails silently).
- Fix failure-detection log to say "expected rank failures" instead of "unexpected", since unexpected ones now trigger anassertion.
- Add get_total_killed_ranks() helper to Plan class.

## Summary by CodeRabbit

* **New Features**
  * GPU-aware build option enabled when a GPU is detected to adjust compilation
  * CI now runs an additional GPU-gated elastic endpoint test for device scenarios

* **Improvements**
  * Refined signal handling and stricter accounting of worker terminations in elastic tests
  * Added utility to compute total killed ranks and new test fixtures covering expansion/fault/contraction scenarios

* **Chores**
  * CI image tag updated for pipeline runs


itayalroy:
Did not review yet, but I'm triggering the CI so you can test it

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -647,15 +656,25 @@ def main():
...
+        if p.exitcode == -signal.SIGTERM:
+            sigterm_count += 1
+        elif p.exitcode != 0:
             failed.append((i, p.exitcode))
     if failed:
         raise RuntimeError(
             f"Worker processes failed: {', '.join(f'worker {i} (exit code {code})' for i, code in failed)}"
         )
 
+    # Verify that exactly the expected number of ranks were killed via SIGTERM.
```
itayalroy: Could we move this logic out of `elastic.py`? This isn’t a reliable check for the general case. I’d prefer if `elastic.py` exposed an optional debug flag to report process exit codes (and possibly more info, whatever you think is necesarry), and then `test_python.sh` (or some wrapper pytest) could verify for its specific plan that we received a single SIGTERM.

.gitlab/test_python.sh
```diff
@@ -70,6 +70,16 @@ $pip3 install --break-system-packages zmq
 
 start_etcd_server "/nixl/python_ci"
 
+if $HAS_GPU ; then
+    echo "==== Running elastic EP test ===="
+    EP_SRC_DIR="examples/device/ep"
+    NIXL_BUILD_DIR=${NIXL_BUILD_DIR:-nixl_build}
+    PYTHONPATH="${NIXL_BUILD_DIR}/${EP_SRC_DIR}:${EP_SRC_DIR}/tests:${EP_SRC_DIR}/tests/elastic${PYTHONPATH:+:$PYTHONPATH}" \
+    timeout 300 python3 ${EP_SRC_DIR}/tests/elastic/elastic.py \
```
ofirfarjun7: As discussed please also add basic plan and run all tests with nvlink/rdma variants

examples/device/ep/tests/elastic/elastic.py
lishapira: move validation checks in elastic behind --validate-plan flag.
Do you think it's enough?

.gitlab/test_python.sh
lishapira: Added basic plane and run both plans with NVLink and RDMA variants.

PR #1519, opened 2026-04-12
Prepare nixl_xfer_op_t extension

ColinNV:
## What?
Prepare the code base for `enum nixl_xfer_op_t` being extended with more operations

## Why?
Will be required by upcoming change for send/recv operations

## Summary by CodeRabbit

* **Refactor**
  * Telemetry and internal helpers now use an explicit transfer-operation type (enum) for direction attribution instead of a boolean.

* **Bug Fixes**
  * Engines/backends enforce valid read/write operations earlier (assertion-based checks), reducing undetected invalid-operation paths.
  * Transfer operation reporting uses numeric enum values for out-of-range cases instead of fixed sentinel text.

* **Tests**
  * Tests updated to call telemetry with the operation-type enum.


src/utils/common/util.h
```diff
@@ -14,11 +14,35 @@
...
+[[nodiscard]] constexpr bool
+weAreTarget(const nixl_xfer_op_t operation) noexcept {
+    switch (operation) {
+    case NIXL_READ:
+        return true;
+    case NIXL_WRITE:
+        return false;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
ColinNV: The compiler should issue a warning when a new enum case is added without it being handled in a switch statement, ergo no "silent" transfer.

src/core/nixl_agent.cpp
```diff
@@ -40,17 +41,19 @@ static const std::vector<std::vector<std::string>> illegal_plugin_combinations =
...
 
 std::string nixlEnumStrings::xferOpStr (const nixl_xfer_op_t &op) {
-    static std::array<std::string, 2> nixl_op_str = {"READ", "WRITE"};
-    if (op<NIXL_READ || op>NIXL_WRITE)
-        return "BAD_OP";
+    static const std::array<std::string, 2> nixl_op_str = {"READ", "WRITE"};
+    if ((op < NIXL_READ) || (op > NIXL_WRITE)) {
+        return std::to_string(std::underlying_type_t<nixl_xfer_op_t>(op));
+    }
     return nixl_op_str[op];
```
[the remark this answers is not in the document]
ColinNV: One of the enums will be changed again soon, can use that opportunity to change these functions. Ideally they would return `std::string_view`, though that change might not be possible for compatibility reasons.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1098,6 +1099,11 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        nixlBackendReqH* &handle,
                                        const nixl_opt_b_args_t* opt_args) const
 {
+    if ((operation != NIXL_WRITE) && (operation != NIXL_READ)) {
+        NIXL_ERROR << "Operation is not read or write";
+        return NIXL_ERR_INVALID_PARAM;
+    }
```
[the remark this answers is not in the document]
ColinNV: In the first step the new functions are to be proof-of-concepted without the thread pool engine.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1193,9 +1204,18 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
...
+        switch (operation) {
+        case NIXL_READ:
+            ret = ep.read(raddr, rmd->getRkey(worker_id), laddr, lmd->mem, lsize, req);
+            break;
+        case NIXL_WRITE:
+            ret = ep.write(laddr, lmd->mem, raddr, rmd->getRkey(worker_id), lsize, req);
+            break;
+        default:
+            std::terminate(); // UNREACHABLE
+        }
```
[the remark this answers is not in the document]
ColinNV: Will be changed in the next PR.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1020,7 +1020,17 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
...
+    case NIXL_READ:
+        op_type = nixlLibfabricReq::READ;
+        break;
+    case NIXL_WRITE:
+        op_type = nixlLibfabricReq::WRITE;
+        break;
+    default:
+        NIXL_ERROR << "Invalid operation for libfabric backend: " << int(operation);
+        return NIXL_ERR_INVALID_PARAM;
+    }
```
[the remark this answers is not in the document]
ColinNV: Not a problem here because we already know that the default error handling is correct for future operations unless this piece of code is extended, too.

src/utils/common/util.h
```diff
@@ -14,11 +14,35 @@
...
+
+#include "nixl_types.h"
 
 #define CONCAT(a, b) CONCAT_0(a, b)
 #define CONCAT_0(a, b) a ## b
 #define UNIQUE_NAME(name) CONCAT(name, __COUNTER__)
 
-#endif /* UTIL_H */
+[[nodiscard]] constexpr bool
+weAreSource(const nixl_xfer_op_t operation) noexcept {
```
iyastreb: I would propose more common name, like nixlXferIsSource
maybe implement as one liner:
```
return operation == NIXL_WRITE

File: src/plugins/ucx/ucx_backend.cpp [hunk 7]
```diff
@@ -1193,9 +1204,18 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
 
         ++result.size;
         nixlUcxReq req;
-        nixl_status_t ret = operation == NIXL_READ ?
-            ep.read(raddr, rmd->getRkey(worker_id), laddr, lmd->mem, lsize, req) :
-            ep.write(laddr, lmd->mem, raddr, rmd->getRkey(worker_id), lsize, req);
+        nixl_status_t ret = NIXL_ERR_INVALID_PARAM;
+
+        switch (operation) {
```
iyastreb: I think switch is heavier than a single if
Maybe just reuse weAreSource here?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1178,6 +1184,11 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
                                   size_t end_idx) {
     batchResult result = {NIXL_SUCCESS, 0, nullptr};
 
+    if ((operation != NIXL_READ) && (operation != NIXL_WRITE)) {
```
iyastreb: This is not a public API, no need to validate args here
We may need to validate in postXfer instead

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1098,6 +1099,11 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
                                        nixlBackendReqH* &handle,
                                        const nixl_opt_b_args_t* opt_args) const
 {
+    if ((operation != NIXL_WRITE) && (operation != NIXL_READ)) {
```
iyastreb: maybe add nixlXferOpIsValid function of something?
So that if we extend operation one day you don't have to change code in many places

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1193,9 +1204,18 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
...
+
+        switch (operation) {
+        case NIXL_READ:
+            ret = ep.read(raddr, rmd->getRkey(worker_id), laddr, lmd->mem, lsize, req);
+            break;
+        case NIXL_WRITE:
+            ret = ep.write(laddr, lmd->mem, raddr, rmd->getRkey(worker_id), lsize, req);
+            break;
+        default:
+            std::terminate(); // UNREACHABLE
```
iyastreb: redundant, operation is already validated

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -20,6 +20,7 @@
 #include "serdes/serdes.h"
 #include "common/nixl_log.h"
 
+#include <exception>
```
iyastreb: I think it's redundant

src/plugins/posix/posix_backend.cpp
```diff
@@ -28,11 +28,16 @@
 
 namespace {
 bool
-isValidPrepXferParams(const nixl_xfer_op_t &operation,
+isValidPrepXferParams(const nixl_xfer_op_t operation,
                       const nixl_meta_dlist_t &local,
                       const nixl_meta_dlist_t &remote,
                       const std::string &remote_agent,
                       const std::string &local_agent) {
+    if ((operation != NIXL_READ) && (operation != NIXL_WRITE)) {
```
iyastreb: maybe just validate it once in common agent code before triggering prepXfer?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1020,7 +1020,17 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
     NIXL_DEBUG << "Processing " << desc_count
                << " descriptors using optimized single-pass approach";
 
-    op_type = (operation == NIXL_WRITE) ? nixlLibfabricReq::WRITE : nixlLibfabricReq::READ;
+    switch (operation) {
```
iyastreb: I would keep a single if as before

ColinNV: This backend doesn't validate `operation` in `prepXfer()`.

src/plugins/ucx/ucx_backend.cpp
ColinNV: There's a catch for `std::exception` in this file, but this is not the right PR for adding it.

src/utils/common/util.h
ColinNV: It's not implemented as one-liner because if this PR is merged there will be more cases in the future. As for the name, I'll add a namespace and...

src/plugins/ucx/ucx_backend.cpp
ColinNV: That doesn't really help because it's not clear whether the new ops would use this code path or a new one.

src/plugins/posix/posix_backend.cpp
ColinNV: So have the agent check the operation, and then do an assert here just for debug builds? Makes sense.

src/plugins/uccl/uccl_backend.cpp
```diff
@@ -538,20 +545,12 @@ nixlUcclEngine::postXfer(const nixl_xfer_op_t &operation,
...
-    case NIXL_WRITE: {
+    } else {
         result = uccl_engine_write_vector(
             conn, mr_ids, addr_v, size_v, uccl_handle->fifo_items, lcnt, &transfer_id);
-        break;
-    }
-    default:
-        NIXL_ERROR << "Unsupported operation type: " << operation;
-        return NIXL_ERR_INVALID_PARAM;
     }
```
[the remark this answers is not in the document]
ColinNV: The assert is more for documentation, there are now checks in class nixlAgent before calling a plugin's prepXfer() or postXfer() function.

src/utils/common/util.h
```diff
@@ -14,11 +14,49 @@
...
+    case NIXL_READ:
+    case NIXL_WRITE:
+        return true;
+    }
+    return false;
+}
+
+[[nodiscard]] constexpr bool
+localIsSource(const nixl_xfer_op_t operation) noexcept {
+    switch (operation) {
```
iyastreb, line 40: minor: still one liner is better (and faster) option imo, at least until we extend this enum

iyastreb (approved):

src/utils/common/util.h
ColinNV: The optimizer should be good enough to produce the same assembly output. And this will only be merged if an extension of the enum is imminent.

ColinNV:
These changes will be part of the later PR that needs them.

PR #1520, opened 2026-04-12
nixl_ep: Make GPU timeouts configurable 

itayalroy:
EP GPU timeouts were hardcoded to 100 seconds, which surfaced in multiple contexts.

- vLLM fault-tolerance architecture discussions asked whether nixl_ep could support dynamic kernel-side timeout configuration, and noted that some systems may need sub-second timeouts.
- CI fault-tolerance tests take too long when failures are only detected after a 100 second timeout.

Make it configurable. Start with a buffer-scoped timeout, and add per-operation timeout controls incrementally if needed.

## Summary by CodeRabbit

* **New Features**
  * Configurable GPU communication timeout: Buffer now accepts timeout_ms (default 30,000 ms). Runtime timeout values are applied to GPU kernels, barrier synchronization, and dispatch/notify/combine paths; CLI test gains --timeout-ms to control this.

* **Documentation**
  * Constructor and barrier docs updated to explain timeout behavior and barrier timeout semantics.

### DIFF
--- a/examples/device/ep/csrc/kernels/api.cuh
+++ b/examples/device/ep/csrc/kernels/api.cuh
@@ -32,7 +32,7 @@ namespace nixl_ep {
 
 namespace intranode {
 
-void barrier(int** barrier_signal_ptrs, int rank, int num_nvl_ranks, cudaStream_t stream);
+void barrier(int** barrier_signal_ptrs, int rank, int num_nvl_ranks, uint64_t timeout_cycles, cudaStream_t stream);
 
 }  // namespace intranode
 
@@ -105,6 +105,7 @@ void notify_dispatch(const int* num_tokens_per_rank,
                      cudaStream_t stream,
                      int64_t num_rdma_bytes,
                      int64_t num_nvl_bytes,
+                     uint64_t timeout_cycles,
                      bool low_latency_mode,
                      gpu_nixl_ctx nixl_ctx);
 
@@ -144,6 +145,7 @@ void dispatch(void* recv_x,
               bool is_cached_dispatch,
               cudaStream_t stream,
               int num_channels,
+              uint64_t timeout_cycles,
               bool low_latency_mode,
               gpu_nixl_ctx nixl_ctx);
 
@@ -167,6 +169,7 @@ void cached_notify(int hidden_int4,
                    cudaStream_t stream,
                    int64_t num_rdma_bytes,
                    int64_t num_nvl_bytes,
+                   uint64_t timeout_cycles,
                    bool is_cached_dispatch,
                    bool low_latency_mode,
                    gpu_nixl_ctx nixl_ctx);
@@ -199,6 +202,7 @@ void combine(cudaDataType_t type,
              int num_ranks,
              cudaStream_t stream,
              int num_channels,
+             uint64_t timeout_cycles,
              bool low_latency_mode,
              gpu_nixl_ctx nixl_ctx);
 
@@ -224,6 +228,7 @@ void dispatch(void* packed_recv_x, void* packed_recv_x_scales,
               int num_tokens, int hidden, int num_max_dispatch_tokens_per_rank,
               int num_topk, int num_experts, int rank, int num_ranks,
               bool use_fp8, bool round_scale, bool use_ue8m0,
+              uint64_t timeout_cycles,
               void* workspace, int num_device_sms,
               cudaStream_t stream, int phases, nixl_ep::gpu_nixl_ctx nixl_ctx);
 
@@ -236,11 +241,11 @@ void combine(void* combined_x,
              uint64_t* next_clean, int num_next_clean_int,
              int num_combined_tokens, int hidden, int num_max_dispatch_tokens_per_rank,
              int num_topk, int num_experts, int rank, int num_ranks,
-             bool use_logfmt,
+             bool use_logfmt, uint64_t timeout_cycles,
              void* workspace, int num_device_sms,
              cudaStream_t stream, int phases, bool zero_copy, nixl_ep::gpu_nixl_ctx nixl_ctx);
 
-void barrier(gpu_nixl_ctx nixl_ctx, int* mask_buffer_ptr, cudaStream_t stream);
+void barrier(gpu_nixl_ctx nixl_ctx, int* mask_buffer_ptr, uint64_t timeout_cycles, cudaStream_t stream);
 
 void query_mask_buffer(int* mask_buffer_ptr, int num_ranks, int* output_mask_tensor, cudaStream_t stream);
 
--- a/examples/device/ep/csrc/kernels/nixl_ep_ht.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ht.cu
@@ -156,6 +156,7 @@ __global__ void notify_dispatch(const int* num_tokens_per_rank,
                                 void** buffer_ptrs,
                                 int** barrier_signal_ptrs,
                                 int rank,
+                                uint64_t timeout_cycles,
                                 nixl_ep::gpu_nixl_ctx nixl_ctx) {
     auto sm_id = static_cast<int>(blockIdx.x);
     auto thread_id = static_cast<int>(threadIdx.x), warp_id = thread_id / 32, lane_id = get_lane_id();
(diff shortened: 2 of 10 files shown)


examples/device/ep/csrc/kernels/api.cuh
ofirfarjun7, line 39: why not add fields here for barrier/notify/dispatch/combine timeout?

itayalroy: gpu_nixl_ctx is just a container for NIXL resources. Timeouts aren’t passed through NIXL APIs, they’re part of the kernel logic, so putting them in nixl_gpu_ctx didn't make sense IMO, even if it would make the diff smaller. We followed DeepEP’s approach for non-NIXL kernel parameters by passing them explicitly instead of putting them in a struct. We can discuss changing this (there are pros and cons), but I wouldn’t do it in this PR.

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -624,6 +626,12 @@ def main():
         action="store_true",
         help="Disable NVLink communication for low-latency kernels",
     )
+    parser.add_argument(
+        "--timeout-ms",
+        type=float,
```
rakhmets: I think the type should be non-negative integers.

```Python
def non_negative_int(value):
    ivalue = int(value)
    if ivalue < 0:
        raise argparse.ArgumentTypeError("must be a non-negative integer")
    return ivalue
...
    parser.add_argument(
        "--timeout-ms",
        type=non_negative_int,
        default=DEFAULT_TIMEOUT_MS,
        help="GPU timeout in milliseconds",
    )
```

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -105,6 +113,9 @@ void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_nvl_bytes
     cudaDeviceProp device_prop = {};
     CUDA_CHECK(cudaGetDeviceProperties(&device_prop, device_id));
     num_device_sms = device_prop.multiProcessorCount;
+    device_clock_rate_khz = device_prop.clockRate;
```
rakhmets: This field was removed from `cudaDeviceProp` in CUDA13.
It can be obtained by:
```C++
cudaDeviceGetAttribute(&device_clock_rate_khz, cudaDevAttrClockRate, device_id);
```

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -59,6 +62,7 @@ def __init__(
         group: Optional[dist.ProcessGroup] = None,
         comm: Optional["mpi4py.MPI.Comm"] = None,
         tcp_store_group: Optional[dist.TCPStore] = None,
+        timeout_ms: float = DEFAULT_TIMEOUT_MS,
```
rakhmets: I think we can change timeout_ms and DEFAULT_TIMEOUT_MS to int now.
And in C++ too.

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: thanks, good catch

examples/device/ep/nixl_ep/buffer.py
itayalroy: done

ofirfarjun7 (dismissed):

examples/device/ep/csrc/nixl_ep.hpp
```diff
@@ -107,6 +108,8 @@ struct Buffer {
     // Device info and communication
     int device_id;
     int num_device_sms;
+    int device_clock_rate_khz = 0;
```
rakhmets: `device_clock_rate_khz` is a local variable of `Buffer::init`, can be removed from the class private fields.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -102,9 +109,10 @@ void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_nvl_bytes
     num_rdma_ranks = std::max(1, num_ranks / NUM_MAX_NVL_PEERS), num_nvl_ranks = std::min(num_ranks, NUM_MAX_NVL_PEERS);
 
     // Get device info
-    cudaDeviceProp device_prop = {};
-    CUDA_CHECK(cudaGetDeviceProperties(&device_prop, device_id));
-    num_device_sms = device_prop.multiProcessorCount;
+    CUDA_CHECK(cudaDeviceGetAttribute(&num_device_sms, cudaDevAttrMultiProcessorCount, device_id));
+    CUDA_CHECK(cudaDeviceGetAttribute(&device_clock_rate_khz, cudaDevAttrClockRate, device_id));
+    EP_HOST_ASSERT(device_clock_rate_khz > 0);
```
rakhmets: There is no need to check the variable twice.
You can do the check here, do static_cast to local variable, and pass `uint64_t` to `milliseconds_to_sycles`. Or pass `int` value without checking it here, and check it inside `milliseconds_to_sycles`.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -58,6 +58,12 @@ static void sleep_ms(int milliseconds) {
     std::this_thread::sleep_for(std::chrono::milliseconds(milliseconds));
 }
 
+static uint64_t milliseconds_to_cycles(int milliseconds, int device_clock_rate_khz) {
```
rakhmets: Please place it to anonymous namespace instead of declaring it as a `static`.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -68,8 +74,9 @@ void Buffer::update_memory_buffers(int num_ranks, int num_experts_per_rank, int6
     }
 }
 
-Buffer::Buffer(int rank, bool explicitly_destroy, bool low_latency_mode):
+Buffer::Buffer(int rank, bool explicitly_destroy, bool low_latency_mode, int timeout_ms):
         low_latency_mode(low_latency_mode),
+        timeout_ms(timeout_ms),
```
rakhmets: You can declare the class field as `uint64_t`, add a helper method to anonymous namespace that asserts in case of negative value, and do `static_cast` to avoid asserts at the later stage.

examples/device/ep/csrc/kernels/nixl_ep_ht.cu
```diff
@@ -913,7 +917,7 @@ __global__ void __launch_bounds__(((kNumDispatchRDMASenderWarps + 1 + NUM_MAX_NV
                 }
 
                 // Timeout check
-                if (clock64() - start_time > NUM_TIMEOUT_CYCLES) {
+                if (clock64() - start_time > timeout_cycles) {
```
rakhmets, line 920: This check is repeated many times. Consider place this into a separate method in this or follow up PR.

examples/device/ep/csrc/nixl_ep.hpp
itayalroy: done

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Done, validating it once in ctor and passing uint64_t onwards

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Done, although I'm not sure why it matters

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: done

examples/device/ep/csrc/kernels/nixl_ep_ht.cu
itayalroy: I agree that it should be a helper method, but this is DeepEP code, I think it's outside the scope of this PR to refactor it, let's postpone to a follow up pr

rakhmets (approved):

ofirfarjun7 (approved):

examples/device/ep/csrc/nixl_ep.cpp
rakhmets: It's mentioned in CPP Core Guidelines.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities

And we also added this to our code style document:
https://github.com/ai-dynamo/nixl/blob/main/docs/CodeStyle.md#anonymous-namespaces

`static` has limitations. Anonymous namespaces have better extensibility, e.g., when adding helper types.

PR #1522, opened 2026-04-13
Remove timestamp field from telemetry events

BatshevaBlack:
## What?
Removes the per-event microsecond timestamp (timestampUs_) from nixlTelemetryEvent, updates all constructors and call sites (core telemetry, backend engine, buffer readers), and bumps TELEMETRY_VERSION to 2 so shared-memory ring buffers reject mismatched layouts. Example readers (C++ and Python) no longer print or decode a per-event timestamp field.

## Why?
Recording a timestamp on every event added cost (syscall / clock read and 8 bytes per event) without any exporter or downstream consumer using it. Dropping the field reduces hot-path overhead and shrinks each event. 

## Summary by CodeRabbit

* **Breaking Changes**
  * Telemetry schema bumped to version 2.
  * Per-event timestamps removed; records now include only category, event name, and value. CSV/export/output reflect the new format.

* **Documentation**
  * Docs updated to remove per-event timestamp field and show revised example output and behavior.

* **Examples**
  * Example readers updated to stop displaying timestamps and match the new telemetry format.

* **Tests**
  * Unit tests adjusted to no longer assert per-event timestamps.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -60,7 +60,7 @@ The **Shared Memory Buffer** plug-in, contains the data per transaction event, w
 
 ### Telemetry Details
 
-- Events are recorded after the corresponding operation completes. The shared buffer does not attach a per-event timestamp; ordering still reflects enqueue order.
+- Events are recorded after the corresponding operation completes and appear in order.
 - Current design allows silent telemetry loss.
 - Current design does not support selective telemetry(e.g per category). All the telemetry events could be either ON or OFF.
 
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -60,7 +60,7 @@ The **Shared Memory Buffer** plug-in, contains the data per transaction event, w
 
 ### Telemetry Details
 
-- Events are recorded after the corresponding operation completes and appear in order.
+- Producers append events under a mutex, consumers read them in **ring insertion order**.
 - Current design allows silent telemetry loss.
 - Current design does not support selective telemetry(e.g per category). All the telemetry events could be either ON or OFF.
 
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -18,7 +18,6 @@ Custom telemetry exporter plug-ins can be created according to [src/plugins/tele
 ### Event Structure
 
 Each telemetry event contains:
-- **Timestamp**: Microsecond precision timestamp
 - **Category**: Event category for filtering and aggregation
 - **Event Name**: Descriptive name/identifier for the event
 - **Value**: Numeric value associated with the event
@@ -61,7 +60,7 @@ The **Shared Memory Buffer** plug-in, contains the data per transaction event, w
 
 ### Telemetry Details
 
-- Timestamp is done after the operation completion.
+- Events are recorded after the corresponding operation completes. The shared buffer does not attach a per-event timestamp; ordering still reflects enqueue order.
 - Current design allows silent telemetry loss.
 - Current design does not support selective telemetry(e.g per category). All the telemetry events could be either ON or OFF.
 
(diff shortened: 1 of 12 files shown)


alexanderbilk (commented):
Fix https://github.com/ai-dynamo/nixl/blob/main/docs/telemetry.md accordingly 
The rest seems fine

alexanderbilk (dismissed):

docs/telemetry.md
```diff
@@ -61,7 +60,7 @@ The **Shared Memory Buffer** plug-in, contains the data per transaction event, w
 
 ### Telemetry Details
 
-- Timestamp is done after the operation completion.
+- Events are recorded after the corresponding operation completes. The shared buffer does not attach a per-event timestamp; ordering still reflects enqueue order.
```
guy-ealey-morag:
```suggestion
- Events are recorded after the corresponding operation completes and appear in order.
```

I think it's redundant to document that something is not there, we should focus on the things that exist in the code.

src/core/telemetry/telemetry_event.h
```diff
@@ -55,29 +55,25 @@ telemetryCategoryStr(const nixl_telemetry_category_t &category);
...
         eventName_[MAX_EVENT_NAME_LEN - 1] = '\0';
     }
 
-    nixlTelemetryEvent(uint64_t timestamp_us,
-                       nixl_telemetry_category_t category,
+    nixlTelemetryEvent(nixl_telemetry_category_t category,
                        const std::string &event_name,
                        uint64_t value) noexcept
-        : nixlTelemetryEvent(timestamp_us, category, event_name.c_str(), value) {}
+        : nixlTelemetryEvent(category, event_name.c_str(), value) {}
```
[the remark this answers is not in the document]
BatshevaBlack: not related to the PR should be handled seperatly

src/api/cpp/telemetry/telemetry_plugin.h
```diff
@@ -24,10 +24,15 @@
...
 
-enum class nixl_telemetry_plugin_api_version : unsigned int { V1 = 1 };
+enum class nixl_telemetry_plugin_api_version : unsigned int {
+    V1 = 1,
+    V2 = 2,
+};
 
 inline constexpr nixl_telemetry_plugin_api_version nixlTelemetryPluginApiVersionV1 =
     nixl_telemetry_plugin_api_version::V1;
+inline constexpr nixl_telemetry_plugin_api_version nixlTelemetryPluginApiVersionV2 =
```
ColinNV: Why use these variables and not directly the enum values?

src/core/nixl_plugin_manager.cpp
```diff
@@ -186,10 +186,9 @@ telemetryLoader(void *handle, const std::string &plugin_path) {
         return nullptr;
     }
 
-    // Check API version
-    if (plugin->api_version != nixlTelemetryPluginApiVersionV1) {
+    if (plugin->api_version != nixlTelemetryPluginApiVersionV2) {
         NIXL_ERROR << "Plugin API version mismatch for " << plugin_path << ": expected "
-                   << static_cast<int>(nixlTelemetryPluginApiVersionV1) << ", got "
+                   << static_cast<int>(nixlTelemetryPluginApiVersionV2) << ", got "
```
ColinNV: The `enum` uses `unsigned`.

ColinNV (dismissed):
Btw, getting the current time on Linux is often a userspace operation without system call overhead. Of course if the timestamp is never needed we should still get rid of it.

ColinNV (dismissed):

src/plugins/telemetry/README.md
guy-ealey-morag: Update to `2025-2026`

guy-ealey-morag (approved):

ColinNV (approved):

alexanderbilk:
/ok to test [ae39e41](https://github.com/ai-dynamo/nixl/commit/ae39e41dda7a6c1086aa79822c01ec23a0e2e7d8)

alexanderbilk (approved):

mkhazraee:
I'm not involved in the details of telemetry implementation, but rather surprised by title of this PR. The whole point of telemetry was to measure post and completion time, are we doing it in another way? Or in some update this capability was dropped? This sounds off to me.

alexanderbilk:
@mkhazraee  post and completion time are getting exported as event values. What we are getting rid of is redundant timestamps of 'When event was captured". We have no real use of it, while turned out it has quite an impact on performance in some cases

brminich (approved):

PR #1523, opened 2026-04-13
Refactor manylinux dockerfile to allow faster UCX & NIXL bisection

ovidiusm:
## What?
Moved UCX and NIXL build steps at the end so that incremental builds during bisection are much faster.
Added --ucx-ref parameter and removed obsolete --ucx-upstream
Removed pip install step since it is just slowing things down for no purpose, this base image does not support running on GPU so no tests are possible.

## Why?
Need the ability to bisect both NIXL and UCX code.

## How?
ARGs must be placed where needed and not on top of Dockerfile since this prevents incremental builds.

## Summary by CodeRabbit

* **Build & Infrastructure**
  * Build scripts now accept a user-provided UCX git reference instead of a fixed upstream ref, improving flexibility.
  * Container build sequence restructured and includes an extra verification step during UCX source checkout.
  * The build no longer automatically installs produced package wheels into the build image.


contrib/build-container.sh
```diff
@@ -192,7 +196,7 @@ show_help() {
     echo "  [--build-type [debug|release] to select build type (default: release)]"
     echo "  [--tag tag for image]"
     echo "  [--python-versions python versions to build for, comma separated]"
-    echo "  [--ucx-upstream use ucx master branch]"
+    echo "  [--ucx-ref ucx git reference (branch, tag, or sha)]"
     echo "  [--build-nixl-ep build NIXL with NIXL EP support (uses latest UCX master)]"
```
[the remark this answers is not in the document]
ovidiusm: fixed

contrib/Dockerfile.manylinux
```diff
@@ -372,5 +373,3 @@ RUN IFS=',' read -ra PYTHON_VERSIONS <<< "$WHL_PYTHON_VERSIONS" && \
 RUN if [ "$(echo $CUDA_VERSION | cut -d. -f1)" = "12" ]; then \
       cp build/src/bindings/python/nixl-meta/nixl*.whl dist/; \
     fi
-
-RUN uv pip install dist/nixl*cp${DEFAULT_PYTHON_VERSION//./}*.whl build/src/bindings/python/nixl-meta/nixl*.whl
```
brminich, line 376: why removed?

contrib/Dockerfile.manylinux
```diff
@@ -321,6 +298,30 @@ RUN export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | c
 # Upgrade setuptools to latest version for compatibility with PEP 639 (license format)
 RUN uv pip install --upgrade 'setuptools>=80.9.0'
 
+ARG UCX_REF="v1.21.x"
```
brminich, line 301: why is this move needed?

contrib/Dockerfile.manylinux
```diff
@@ -321,6 +298,29 @@ RUN export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | c
...
+RUN cd /usr/local/src && \
+     git clone https://github.com/openucx/ucx.git && \
+     cd ucx && 			     \
+     git checkout $UCX_REF &&	     \
```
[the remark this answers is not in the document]
ovidiusm: fixed

contrib/Dockerfile.manylinux
ovidiusm: If the variable is set on top, and we change the value, it will rebuild everything. Dockerfile ARGs should not be placed on top in general.

contrib/Dockerfile.manylinux
ovidiusm: It should not have been installed in the first place. This image does not support any tests, it is for building only. GPU support is not available.

contrib/build-container.sh
```diff
@@ -119,9 +119,13 @@ get_options() {
...
-        --ucx-upstream)
-            # Master branch (v1.20) also containing EFA SRD support
-            UCX_REF=9d2b88a1f67faf9876f267658bd077b379b8bb76
+        --ucx-ref)
+            if [ "$2" ]; then
+                UCX_REF=$2
+                shift
+            else
+                missing_requirement $1
+            fi
```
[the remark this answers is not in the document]
ovidiusm: I see your point but it would require a full cleanup of this entire script. Should be fixed separately.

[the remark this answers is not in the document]
ovidiusm: Sure

contrib/Dockerfile.manylinux
```diff
@@ -321,6 +298,30 @@ RUN export UV_INDEX="https://download.pytorch.org/whl/cu$(echo $CUDA_VERSION | c
 # Upgrade setuptools to latest version for compatibility with PEP 639 (license format)
 RUN uv pip install --upgrade 'setuptools>=80.9.0'
 
+ARG UCX_REF="v1.21.x"
+RUN cd /usr/local/src && \
+     git clone https://github.com/openucx/ucx.git && \
```
guy-ealey-morag, line 303: Would adding `--depth=1` be beneficial here? (shallow clone)

ovidiusm: We do not do that in any of our dockerfiles. UCX clones very quickly. We also have a separate checkout step that needs to work

guy-ealey-morag (approved):

aranadive (approved):

PR #1527, opened 2026-04-14
PLUGINS/UCX: Apply configuration for EFA only

tvegas1:
Apply configuration for UCX only, tested on AWS/EFA setup.

## Summary by CodeRabbit

* **Bug Fixes**
  * UCX configuration for address version and component limits is now applied only when EFA-capable hardware is present, avoiding incorrect defaults on systems without EFA.
  * Component/socket limit enforcement now considers both UCX version and EFA hardware presence to ensure correct behavior.
  * UCX version compatibility warnings are emitted independently of the hardware gating for more reliable reporting.


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -455,9 +454,15 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
...
-    if (ucpVersion_ >= UCP_VERSION(1, 19)) {
-        config.modify("MAX_COMPONENT_MDS", "32");
-    } else {
+    const auto &hw_info = nixl::hwInfo::instance();
+    if (hw_info.numIbDevices == 0 && hw_info.numEfaDevices != 0) {
+        config.modify("ADDRESS_VERSION", "v2");
+        if (ucpVersion_ >= UCP_VERSION(1, 19)) {
+            config.modify("MAX_COMPONENT_MDS", "32");
+        }
+    }
```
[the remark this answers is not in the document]
tvegas1: this is expected, on mixed-hardware nodes (if any) we remove the configuration

brminich (dismissed):

src/plugins/ucx/ucx_utils.cpp
guy-ealey-morag, line 463: What is the reason that these two configs should be applied only for EFA setups?

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -455,9 +454,15 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
         config.modify("RC_GDA_NUM_CHANNELS", std::to_string(num_device_channels));
     }
 
-    if (ucpVersion_ >= UCP_VERSION(1, 19)) {
-        config.modify("MAX_COMPONENT_MDS", "32");
-    } else {
+    const auto &hw_info = nixl::hwInfo::instance();
+    if (hw_info.numIbDevices == 0 && hw_info.numEfaDevices != 0) {
+        config.modify("ADDRESS_VERSION", "v2");
```
guy-ealey-morag, line 459: This check looks only at the hardware and ignores any specific selections done in `UCX_NET_DEVICES`,
So for example in a setup with IB devices & EFA devices, if `UCX_NET_DEVICES` selects only EFA devices then this check is `false` even though only EFAs are used.
Is it okay in this case or should the check be adapted?

src/plugins/ucx/ucx_utils.cpp
tvegas1: AWS/EFA can have more HCA/MD (32) than default max component mds of 16, "v1" does not support packing such count of interfaces addresses.

src/plugins/ucx/ucx_utils.cpp
tvegas1: I think in practice the only presence of EFA interfaces is enough to conclude that we want v2/32 options, tbh I am not aware of EFA with non-EFA IB use-case tbh, @brminich do we want to remove `numIbDevices == 0` check?

brminich: im also not aware about such systems
it is up to you, you can remove

tvegas1: removed

brminich (approved):

PR #1528, opened 2026-04-14
Remove default port from device names in UCX context

guy-ealey-morag:
## What?
Don't append a default port `:1` to device names when building the UCX config

## Why?
From UCX 1.21 device names can be specified without a port number (https://github.com/openucx/ucx/pull/11142)
For previous versions the logic stays the same

## Summary by CodeRabbit

* **Refactor**
  * Device-list formatting for NET_DEVICES now adapts to UCX version, altering how device suffixes are applied.
  * Improves compatibility with newer UCX releases and streamlines configuration handling for device selection.
  * Internal logic and comments simplified; no changes to public interfaces or exported entities.


rakhmets (approved):

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -434,14 +437,14 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
...
-        for (const auto &dev : devs) {
-            devs_str += dev + ":1,";
-        }
-        devs_str.pop_back(); // to remove odd comma after the last device
+        // Before UCX 1.21 each device name needs an explicit port suffix
+        const absl::string_view suffix = (ucpVersion_ < UCP_VERSION(1, 21)) ? ":1" : "";
+        const std::string devs_str =
+            absl::StrJoin(devs, ",", [suffix](std::string *out, absl::string_view dev) {
+                absl::StrAppend(out, dev, suffix);
+            });
```
brminich, line 447: maybe we can wait till 1.21 will be the oldest supported version?

guy-ealey-morag: I'm worried that it might be forgotten.
I think it would be best to merge it now and when 1.21 becomes the oldest supported then we can remove the special handling (it would be easy to find it with a Ctrl+F)

ColinNV (approved):

src/plugins/ucx/ucx_utils.cpp
brminich: can we add a test?
im afraid of bringing some regression, while old syntax is anyway acceepted with new UCX versions

tvegas1: Do we have oldest supported UCX version in NIXL somewhere?

PR #1531, opened 2026-04-15
telemetry: replace string event names with enum type

BatshevaBlack:
## What?
Replaces per-telemetry-event string storage with a fixed nixl_telemetry_event_type_t enum (uint8_t) and renames the field from eventName_ to eventType_. Events are still described as strings where needed via nixlEnumStrings::telemetryEventTypeStr() (constexpr string_view).

## Why?
Save string copy on the hot path

## Summary by CodeRabbit

* **New Features**
  * Telemetry now uses a strongly-typed event type classification for consistent event identities.

* **Bug Fixes**
  * Event buffering/reporting switched from freeform names to typed event types; unmapped statuses no longer emit events.

* **Documentation**
  * Event field label changed to "Event type"; CSV and Prometheus docs updated to reflect type-based output.

* **Tests**
  * Tests updated to assert typed telemetry event identifiers.

* **Examples**
  * Example readers now print mapped event-type strings instead of raw event-name text.

### DIFF
--- a/examples/python/telemetry_reader.py
+++ b/examples/python/telemetry_reader.py
@@ -67,7 +67,6 @@ AGENT_ERR_NOT_SUPPORTED = 16
 AGENT_ERR_REMOTE_DISCONNECT = 17
 AGENT_ERR_CANCELED = 18
 AGENT_ERR_NO_TELEMETRY = 19
-AGENT_BACKEND = 20
 
 # Global flag for graceful shutdown
 running = True
@@ -263,7 +262,6 @@ _EVENT_TYPE_STRINGS = {
     AGENT_ERR_REMOTE_DISCONNECT: "agent_err_remote_disconnect",
     AGENT_ERR_CANCELED: "agent_err_canceled",
     AGENT_ERR_NO_TELEMETRY: "agent_err_no_telemetry",
-    AGENT_BACKEND: "agent_backend",
 }
 
 
--- a/examples/python/telemetry_reader.py
+++ b/examples/python/telemetry_reader.py
@@ -55,8 +55,19 @@ AGENT_MEMORY_REGISTERED = 4
 AGENT_MEMORY_DEREGISTERED = 5
 AGENT_XFER_TIME = 6
 AGENT_XFER_POST_TIME = 7
-AGENT_ERROR = 8
-AGENT_BACKEND = 9
+AGENT_ERR_NOT_POSTED = 8
+AGENT_ERR_INVALID_PARAM = 9
+AGENT_ERR_BACKEND = 10
+AGENT_ERR_NOT_FOUND = 11
+AGENT_ERR_MISMATCH = 12
+AGENT_ERR_NOT_ALLOWED = 13
+AGENT_ERR_REPOST_ACTIVE = 14
+AGENT_ERR_UNKNOWN = 15
+AGENT_ERR_NOT_SUPPORTED = 16
+AGENT_ERR_REMOTE_DISCONNECT = 17
+AGENT_ERR_CANCELED = 18
+AGENT_ERR_NO_TELEMETRY = 19
+AGENT_BACKEND = 20
 
 # Global flag for graceful shutdown
 running = True
@@ -240,7 +251,18 @@ _EVENT_TYPE_STRINGS = {
     AGENT_MEMORY_DEREGISTERED: "agent_memory_deregistered",
     AGENT_XFER_TIME: "agent_xfer_time",
     AGENT_XFER_POST_TIME: "agent_xfer_post_time",
-    AGENT_ERROR: "agent_error",
+    AGENT_ERR_NOT_POSTED: "agent_err_not_posted",
+    AGENT_ERR_INVALID_PARAM: "agent_err_invalid_param",
+    AGENT_ERR_BACKEND: "agent_err_backend",
+    AGENT_ERR_NOT_FOUND: "agent_err_not_found",
+    AGENT_ERR_MISMATCH: "agent_err_mismatch",
+    AGENT_ERR_NOT_ALLOWED: "agent_err_not_allowed",
+    AGENT_ERR_REPOST_ACTIVE: "agent_err_repost_active",
+    AGENT_ERR_UNKNOWN: "agent_err_unknown",
+    AGENT_ERR_NOT_SUPPORTED: "agent_err_not_supported",
+    AGENT_ERR_REMOTE_DISCONNECT: "agent_err_remote_disconnect",
+    AGENT_ERR_CANCELED: "agent_err_canceled",
+    AGENT_ERR_NO_TELEMETRY: "agent_err_no_telemetry",
     AGENT_BACKEND: "agent_backend",
 }
 
--- a/examples/python/telemetry_reader.py
+++ b/examples/python/telemetry_reader.py
@@ -34,7 +34,6 @@ logger = logging.getLogger(__name__)
 
 # Constants from telemetry_event.h
 TELEMETRY_VERSION = 2
-MAX_EVENT_NAME_LEN = 32
 
 # NIXL telemetry categories
 NIXL_TELEMETRY_MEMORY = 0
@@ -46,6 +45,18 @@ NIXL_TELEMETRY_PERFORMANCE = 5
 NIXL_TELEMETRY_SYSTEM = 6
 NIXL_TELEMETRY_CUSTOM = 7
 NIXL_TELEMETRY_MAX = 8
+
+# NIXL telemetry event types (nixl_telemetry_event_type_t)
+AGENT_TX_BYTES = 0
+AGENT_RX_BYTES = 1
+AGENT_TX_REQUESTS_NUM = 2
+AGENT_RX_REQUESTS_NUM = 3
+AGENT_MEMORY_REGISTERED = 4
+AGENT_MEMORY_DEREGISTERED = 5
+AGENT_XFER_TIME = 6
+AGENT_XFER_POST_TIME = 7
+AGENT_ERROR = 8
+AGENT_BACKEND = 9
 
 # Global flag for graceful shutdown
 running = True
@@ -65,8 +76,8 @@ class NixlTelemetryEvent(ctypes.Structure):
     _pack_ = 1
     _fields_ = [
         ("category", ctypes.c_int),
-        ("event_name", ctypes.c_char * MAX_EVENT_NAME_LEN),
-        ("_padding", ctypes.c_uint32),
+        ("event_type", ctypes.c_uint8),
+        ("_padding", ctypes.c_char * 3),
         ("value", ctypes.c_uint64),
     ]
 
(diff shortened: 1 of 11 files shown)


src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -124,7 +124,7 @@ nixlTelemetryPrometheusExporter::registerGauge(const std::string &name,
 nixl_status_t
 nixlTelemetryPrometheusExporter::exportEvent(const nixlTelemetryEvent &event) {
     try {
-        const std::string event_name(event.eventName_);
+        const std::string event_name(nixlEnumStrings::telemetryEventTypeStr(event.eventType_));
```
ColinNV, line 129: Perhaps add a "TODO" to switch to `string_view` and use heterogeneous lookup for `counters_` and `gauges_` if/when we switch to C++20.

src/core/telemetry/telemetry.cpp
```diff
@@ -166,81 +165,88 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
 }
 
 void
-nixlTelemetry::updateData(const std::string &event_name,
+nixlTelemetry::updateData(nixl_telemetry_event_type_t event_type,
```
ColinNV, line 168: Perhaps for a separate PR, since the event has category before type it would be nice that were the case here, too.

ColinNV (dismissed):
LGTM -- as long as we don't need the flexibility of the string.

src/api/cpp/backend/backend_engine.h
```diff
@@ -62,12 +62,12 @@ class nixlBackendEngine {
...
 
         void
-        addTelemetryEvent(const std::string &event_name, uint64_t value) {
+        addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) {
             if (!enableTelemetry_) return;
             if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) return;
             std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
             telemetryEvents_.emplace_back(
-                nixl_telemetry_category_t::NIXL_TELEMETRY_BACKEND, event_name, value);
+                nixl_telemetry_category_t::NIXL_TELEMETRY_BACKEND, event_type, value);
```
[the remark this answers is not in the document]
BatshevaBlack: This locking/size() issue is pre-existing in addTelemetryEvent and is orthogonal to the telemetry enum change (string event name → nixl_telemetry_event_type_t). I’d rather not expand this PR’s scope with mutex ordering here;

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +43,53 @@ enum class nixl_telemetry_category_t {
...
+    AGENT_RX_BYTES = 1,
+    AGENT_TX_REQUESTS_NUM = 2,
+    AGENT_RX_REQUESTS_NUM = 3,
+    AGENT_MEMORY_REGISTERED = 4,
+    AGENT_MEMORY_DEREGISTERED = 5,
+    AGENT_XFER_TIME = 6,
+    AGENT_XFER_POST_TIME = 7,
+    AGENT_ERROR = 8,
+    AGENT_BACKEND = 9,
+};
```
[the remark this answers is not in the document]
BatshevaBlack: Agree this is a real follow-on (reserved backend/custom event ordinals + export behavior). Out of scope for this PR

src/core/telemetry/telemetry.cpp
BatshevaBlack: Agree it would read nicer; I’ll reorder updateData and all call sites in a small follow-up to keep this PR focused

alexanderbilk (dismissed):

src/core/telemetry/telemetry.cpp
```diff
@@ -166,81 +165,88 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
+    events_.emplace_back(nixl_telemetry_category_t::NIXL_TELEMETRY_TRANSFER, bytes_type, bytes);
+    events_.emplace_back(nixl_telemetry_category_t::NIXL_TELEMETRY_TRANSFER, requests_type, 1);
 }
 
 void
 nixlTelemetry::addPostTime(std::chrono::microseconds post_time) {
-    updateData("agent_xfer_post_time",
+    updateData(nixl_telemetry_event_type_t::AGENT_XFER_POST_TIME,
                nixl_telemetry_category_t::NIXL_TELEMETRY_PERFORMANCE,
                post_time.count());
```
guy-ealey-morag:
```suggestion
               static_cast<uint64_t>(post_time.count()));
```

For consistency with other funcs here

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +43,53 @@ enum class nixl_telemetry_category_t {
...
+
 namespace nixlEnumStrings {
 std::string
 telemetryCategoryStr(const nixl_telemetry_category_t &category);
+
+[[nodiscard]] constexpr std::string_view
+telemetryEventTypeStr(nixl_telemetry_event_type_t name) noexcept {
```
guy-ealey-morag:
```suggestion
telemetryEventTypeStr(nixl_telemetry_event_type_t type) noexcept {
```

examples/python/telemetry_reader.py
```diff
@@ -34,7 +34,6 @@
 
 # Constants from telemetry_event.h
 TELEMETRY_VERSION = 2
```
guy-ealey-morag, line 36: The version is not increased because it was increased in the other PR, correct?

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +43,53 @@ enum class nixl_telemetry_category_t {
     NIXL_TELEMETRY_CUSTOM = 7, // Custom/user-defined events
 };
 
+/**
+ * @enum nixl_telemetry_event_type_t
+ * @brief Enumerates all known telemetry event types, avoiding per-event string copies.
```
guy-ealey-morag:
```suggestion
 * @brief Enumerates all known telemetry event types.
```

The last part references the code that is now deleted so I think it's irrelevant.

src/core/telemetry/telemetry.cpp
```diff
@@ -166,81 +165,88 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
                rx_requests_num);
 }
 
 void
 nixlTelemetry::updateErrorCount(nixl_status_t error_type) {
-    updateData(
-        nixlEnumStrings::statusStr(error_type), nixl_telemetry_category_t::NIXL_TELEMETRY_ERROR, 1);
+    updateData(nixl_telemetry_event_type_t::AGENT_ERROR,
+               nixl_telemetry_category_t::NIXL_TELEMETRY_ERROR,
+               static_cast<uint64_t>(error_type));
```
guy-ealey-morag, line 212: This function is called `updateErrorCount` so I think the goal was to log "1" per error.
Now it logs the `error_type` cast to `uint64_t` instead. 
(and wouldn't it result in weird values? the error types are negative and would "wrap" around when being cast to unsigned)
Do you know how error telemetry events are being used after they're logged?

Also, previously the error name would be reflected in the telemetry event name through `nixlEnumStrings::statusStr(error_type)` and now it's missing and may be less intuitive.

@ColinNV

guy-ealey-morag (changes requested):

examples/python/telemetry_reader.py
BatshevaBlack: yes

src/core/telemetry/telemetry.cpp
ColinNV: Agree, this is a bit strange. Using `error_type` as value only works if post-processing does a histogram on the values. Otherwise `nixl_telemetry_event_type_t` should have one value for every `nixl_status_t` value to then count `1`s again.

BatshevaBlack:
@guy-ealey-morag - I don't manage to solve the requested change on one of the comments- can you close it please?
thanks!

src/core/telemetry/telemetry_event.h
```diff
@@ -56,17 +57,62 @@ enum class nixl_telemetry_event_type_t : uint8_t {
...
+        return nixl_telemetry_event_type_t::AGENT_ERR_REMOTE_DISCONNECT;
+    case NIXL_ERR_CANCELED:
+        return nixl_telemetry_event_type_t::AGENT_ERR_CANCELED;
+    case NIXL_ERR_NO_TELEMETRY:
+        return nixl_telemetry_event_type_t::AGENT_ERR_NO_TELEMETRY;
+    case NIXL_SUCCESS:
+    case NIXL_IN_PROG:
+        return std::nullopt;
+    }
+    return std::nullopt;
```
guy-ealey-morag: Reaching here means that a new type of error was not added here.
I think we should probably fail or at least log something to raise it as an issue to the developer.

src/core/telemetry/telemetry.cpp
```diff
@@ -207,9 +207,11 @@ nixlTelemetry::updateRxRequestsNum(uint32_t rx_requests_num) {
 
 void
 nixlTelemetry::updateErrorCount(nixl_status_t error_type) {
-    updateData(nixl_telemetry_event_type_t::AGENT_ERROR,
-               nixl_telemetry_category_t::NIXL_TELEMETRY_ERROR,
-               static_cast<uint64_t>(error_type));
+    const auto event_type = nixlTelemetryEventTypeForStatus(error_type);
+    if (!event_type) {
```
guy-ealey-morag: We should not reach here with a non-error status, I think it would be cleaner to assert it instead of ignoring.
`NIXL_ASSERT_ALWAYS(error_type < 0)`
(Alternatively, this can also be done inside `nixlTelemetryEventTypeForStatus`)

BatshevaBlack:
@guy-ealey-morag - thanks! 
resolved comments

src/core/telemetry/telemetry.cpp
```diff
@@ -34,6 +33,41 @@
 using namespace std::chrono_literals;
 namespace fs = std::filesystem;
 
+std::optional<nixl_telemetry_event_type_t>
+nixlTelemetryEventTypeForStatus(nixl_status_t s) {
+    switch (s) {
+    case NIXL_SUCCESS:
+    case NIXL_IN_PROG:
+        return std::nullopt;
```
guy-ealey-morag:
```suggestion
nixl_telemetry_event_type_t
nixlTelemetryEventTypeForStatus(nixl_status_t s) {
    switch (s) {
```

It is already asserted that the value is negative so no need for optional value here.

src/core/telemetry/telemetry.cpp
```diff
@@ -166,83 +200,94 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
 void
 nixlTelemetry::updateErrorCount(nixl_status_t error_type) {
-    updateData(
-        nixlEnumStrings::statusStr(error_type), nixl_telemetry_category_t::NIXL_TELEMETRY_ERROR, 1);
+    NIXL_ASSERT_ALWAYS(static_cast<int>(error_type) < 0)
+        << "nixlTelemetry::updateErrorCount expects a negative nixl_status_t error code";
+    const auto event_type = nixlTelemetryEventTypeForStatus(error_type);
+    NIXL_ASSERT_ALWAYS(event_type.has_value())
+        << "nixlTelemetryEventTypeForStatus missing mapping for nixl_status_t "
+        << static_cast<int>(error_type);
```
guy-ealey-morag:
```suggestion
```

This assert is already covered by the two others, can be removed

guy-ealey-morag (dismissed):

guy-ealey-morag (dismissed):

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +44,89 @@ enum class nixl_telemetry_category_t {
...
+
+[[nodiscard]] std::optional<nixl_telemetry_event_type_t>
+nixlTelemetryEventTypeForStatus(nixl_status_t s);
+
 namespace nixlEnumStrings {
 std::string
 telemetryCategoryStr(const nixl_telemetry_category_t &category);
+
+[[nodiscard]] constexpr std::string_view
+telemetryEventTypeStr(nixl_telemetry_event_type_t type) noexcept {
```
ColinNV: Minor: `type` can be `const`.

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +44,89 @@ enum class nixl_telemetry_category_t {
     NIXL_TELEMETRY_CUSTOM = 7, // Custom/user-defined events
 };
 
+/**
+ * @enum nixl_telemetry_event_type_t
+ * @brief Enumerates all known telemetry event types.
+ */
+enum class nixl_telemetry_event_type_t : uint8_t {
```
ColinNV: We aren't saving space in `nixlTelemetryEvent` by making this smaller than 32-bits, and probably never will as it also contains a 64-bit value. Might as well make this 32-bit for future extensions?

src/core/telemetry/telemetry_event.h
```diff
@@ -45,9 +44,89 @@ enum class nixl_telemetry_category_t {
...
+    AGENT_ERR_NOT_FOUND = 11,
+    AGENT_ERR_MISMATCH = 12,
+    AGENT_ERR_NOT_ALLOWED = 13,
+    AGENT_ERR_REPOST_ACTIVE = 14,
+    AGENT_ERR_UNKNOWN = 15,
+    AGENT_ERR_NOT_SUPPORTED = 16,
+    AGENT_ERR_REMOTE_DISCONNECT = 17,
+    AGENT_ERR_CANCELED = 18,
+    AGENT_ERR_NO_TELEMETRY = 19,
+    AGENT_BACKEND = 20,
```
ColinNV: Would BACKEND_SOMETHING be a better fit? Will we need BACKEND_ERR_foo for all errors? Or should we remove it until we re-introduce backend telemetry?

src/core/telemetry/telemetry.cpp
```diff
@@ -34,6 +33,38 @@
 using namespace std::chrono_literals;
 namespace fs = std::filesystem;
 
+nixl_telemetry_event_type_t
```
ColinNV: Minor: `[[nodiscard]]`.

ColinNV (dismissed):

guy-ealey-morag (dismissed):

ColinNV (approved):

alexanderbilk (approved):

ovidiusm (approved):

PR #1534, opened 2026-04-15
python: recognize Intel XPU devices as VRAM

yafshar:
## What?
- Add Intel XPU device support in the Python API by recognizing `xpu:N` device strings.
- Map XPU devices to `VRAM_SEG` alongside CUDA devices.
- Replace CUDA/CPU ternary logic with explicit `if/elif/else` handling in:
  - `nixlXferDList` (single-tensor and list paths)
  - `nixlRegDList` (single-tensor and list paths)

## Why?
- PyTorch reports Intel XPU devices as `xpu:N`, not `cuda:N`.
- Without this change, XPU tensors are misclassified as DRAM, leading to incorrect registration and transfer behavior.
- Correct classification is required for VRAM handling and fast transfer paths in Dynamo/NIXL.

## How?
- Detect `xpu` device strings in Python descriptor helpers and classify them as VRAM-backed.
- Use explicit device classification instead of CUDA/CPU ternary logic.
- Keep changes localized to Python descriptor helpers.

## Compatibility and risk
- No API surface changes.
- No behavior changes for CPU or CUDA paths.
- Change is limited to Python-side device classification logic.

## Summary by CodeRabbit

* **New Features**
  * XPU support added so tensor operations, transfers, and memory registrations recognize CPU, CUDA, and XPU devices.
  * Device-type inference improved to determine memory type consistently from device/holder objects.

* **Bug Fixes**
  * Empty transfer/registration requests now validated and return a clear error instead of proceeding.
  * Safer handling of empty or invalid descriptor inputs prevents unexpected indexing errors and fails early with diagnostics.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/nixl/pull/1534" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


src/api/python/_api.py
```diff
@@ -904,6 +906,16 @@ def check_remote_metadata(
...
 
+    def get_mem_type_from_device(self, device_or_holder) -> str:
+        device = getattr(device_or_holder, "device", device_or_holder)
+        device_type = getattr(device, "type", None)
+
+        if device_type == "cuda":
+            return "cuda"
+        if device_type == "xpu":
+            return "xpu"
+        return "cpu"
```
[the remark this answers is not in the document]
yafshar: Thanks for the suggestion. We intentionally keep unknown/unsupported device types on a CPU fallback path in get_mem_type_from_device for backward compatibility and permissive behavior in mixed environments. So we are not changing this to raise on unknown device.type.

yafshar:
@mkhazraee @brminich quick ping for visibility and following up on last week’s discussion. Would appreciate review/feedback on #1534 and #1536 when you have a chance.

src/api/python/_api.py
```diff
@@ -912,6 +914,16 @@ def check_remote_metadata(
         else:
             return False
 
+    def get_mem_type_from_device(self, device_or_holder) -> str:
```
mkhazraee: Let's make this method return one of these 5 values, better matching the c++ world:
"DRAM", "VRAM", "FILE", "BLOCK", "OBJ"
The only two places that cuda/cpu was used was this internal call. It was leftovers from an initial code we had, and we forgot to clean it up. 

The usage sounds like more of get_mem_type_from_tensor_type, maybe we can shrink the name to get_mem_type for now, and add a type for the input to be exactly the tensor_type output? Also if there is a better way of extracting tensor type directly from tensor, maybe we can pass the tensor itself.

src/api/python/_api.py
```diff
@@ -238,6 +239,7 @@ def __init__(
             "OBJ": nixlBind.OBJ_SEG,
             "cpu": nixlBind.DRAM_SEG,
             "cuda": nixlBind.VRAM_SEG,
+            "xpu": nixlBind.VRAM_SEG,
```
mkhazraee: With the change I mentioned, we don't need this, and we can even remove cpu/cuda from here.

src/api/python/_api.py
```diff
@@ -933,22 +945,20 @@ def get_xfer_descs(
...
+        with contextlib.suppress(TypeError):
+            descs_len = len(descs)
+
         if isinstance(descs, nixlBind.nixlXferDList):
             return descs
-        elif isinstance(descs, nixlBind.nixlRegDList):
+        if isinstance(descs, nixlBind.nixlRegDList):
             logger.error("RegList type detected for transfer, please use XferList")
             new_descs = None
-        elif isinstance(descs[0], tuple):
```
mkhazraee, line 962: AI bot seems to barely have moved the code around, using the suppress, but it didn't make the code any cleaner. Maybe it added just one check for non-empty list? Not sure even that. Am I missing some functionality change?

yafshar: You are right that most of this part is structural, but there is a concrete behavior change around the tuple branch.

The tuple path now runs only after a safe length probe, an explicit empty-input check, and a list/tuple type check. This avoids touching `descs[0]` too early, which previously could raise `IndexError` or `TypeError` for empty or non-indexable inputs before validation.

So for valid list/tuple tuple-descriptor inputs, behavior is unchanged. For empty or wrong-container inputs, behavior is now deterministic (None plus logged error) rather than failing on early indexing. It also intentionally narrows accepted containers to list/tuple.

src/api/python/_api.py
dsocek: Hi @mkhazraee 
Renaming to `get_mem_type()` and returning `"DRAM" / "VRAM"` makes it much cleaner.

But, removing `"cpu"` and `"cuda"` from `nixl_mems` would break dynamo’s `nixl_connect` on current main.

It currently passes those strings as `mem_type` through the tuple-list path in three places:
* https://github.com/ai-dynamo/dynamo/blob/main/lib/bindings/python/src/dynamo/nixl_connect/__init__.py#L348
* https://github.com/ai-dynamo/dynamo/blob/main/lib/bindings/python/src/dynamo/nixl_connect/__init__.py#L355
* https://github.com/ai-dynamo/dynamo/blob/main/lib/bindings/python/src/dynamo/nixl_connect/__init__.py#L1166

Keeping it for backward compatibility would avoid requiring a synchronized release across both repos.

Happy to put up a companion dynamo PR to migrate to canonical names, but until that lands, the aliases should probably stick around.

I think the safest path is to first land a dynamo PR that migrates to the canonical names, and only then drop the aliases here.

That said, unless we can confirm no other repos depend on "cpu" / "cuda", it may still be worth keeping them as aliases for backward compatibility to avoid cross-repo breakage.

dsocek: FYI @mkhazraee @yafshar, I opened a PR in Dynamo repo to address use of canonical segment names (VRAM/DRAM) instead of device-specific strings:
ai-dynamo/dynamo#9073

mkhazraee: Fair, we can add to be deprecated, or just let them be, not a big deal.

src/api/python/_api.py
mkhazraee: Oh I see what you mean, let's just separate it to another PR with clear clarification, and we land the current PR separately.

src/api/python/_api.py
```diff
@@ -933,6 +937,9 @@ def check_remote_metadata(
         else:
             return False
 
+    def get_mem_type(self, tensor: torch.Tensor) -> Literal["DRAM", "VRAM"]:
```
mkhazraee: If input is just tensor, then it should be tensor_mem_type or so, I expected 5 potential outputs, but fine if we keep the tensor as input. The current get_device handles XPU case properly too?

src/api/python/_api.py
```diff
@@ -257,8 +258,11 @@ def __init__(
             "FILE": nixlBind.FILE_SEG,
             "BLOCK": nixlBind.BLK_SEG,
             "OBJ": nixlBind.OBJ_SEG,
+            # TODO(nixl): Remove legacy aliases after downstream callers migrate
+            # to canonical mem_type names ("DRAM"/"VRAM") in follow-up PRs.
             "cpu": nixlBind.DRAM_SEG,
             "cuda": nixlBind.VRAM_SEG,
+            "xpu": nixlBind.VRAM_SEG,
```
mkhazraee: Why there is xpu still here? Any external code that depends on it? There shouldn't be. Also instead of the TODO and details, just let's add // to be deprecated at the end of cpu and cuda lines.

src/api/python/_api.py
yafshar: Tensor inference is intentionally limited to DRAM and VRAM. Memory types such as FILE, BLOCK, and OBJ are supported only for descriptor-based inputs where mem_type is explicitly specified. For device handling, CPU tensors (get_device == -1) map to DRAM, while non‑CPU devices, including XPU, are treated as accelerators and routed through the VRAM path.

src/api/python/_api.py
yafshar: I will remove the xpu alias and will replace the TODO block with inline deprecation notes

src/api/python/_api.py
yafshar: Makes sense, agreed. I’ll keep this PR focused, and move the structural/input-validation cleanup (safe length probe, empty/container checks, tuple-branch ordering) into a separate PR with clear motivation. I’ll update this PR accordingly.

src/api/python/_api.py
mkhazraee: The reason I'm asking this is now that it's only for tensor, and it's a single line, why not to simplify the PR to just change this line:
mem_type = "cuda" if str(descs.device).startswith("cuda") else "cpu"
to
mem_type = "DRAM" if tensor.get_device() == -1 else "VRAM"

Agreed cuda/cpu was not the best, it was a leftover from early development, but if that's just a one liner fixed, why to separate it into a method. Previously I was thinking of making it more generic with 5 outputs, but agreed for now it's just tensor type. And IIRC in dynamo that cuda/cpu is passed, xpu would've some issues, and that line should be changed too anyways, this method is not getting called there so doesn't fix it.

yafshar: I’ll inline the DRAM/VRAM mapping directly at the four tensor call sites and remove the helper method. This keeps behavior the same for this PR and leaves non-tensor paths unchanged.

yafshar:
@mkhazraee All feedback on this PR is now addressed, and I’ve kept it scoped to the requested change.
I’ll follow up with a separate PR for the structural/input‑validation cleanup you mentioned. This one should be ready to merge.

mkhazraee (dismissed):

brminich (dismissed):

src/api/python/_api.py
```diff
@@ -981,7 +981,7 @@ def get_xfer_descs(
                 new_descs = None
         elif isinstance(descs, torch.Tensor):
             if descs.is_contiguous():
-                mem_type = "cuda" if str(descs.device).startswith("cuda") else "cpu"
+                mem_type = "DRAM" if descs.get_device() == -1 else "VRAM"
```
guy-ealey-morag:
```suggestion
                mem_type = _tensor_mem_type(descs)
```

This trinary operator repeats here 4 times, I suggest putting it in a static helper function
```python
@staticmethod
def _tensor_mem_type(tensor: torch.Tensor) -> str:
    return "DRAM" if tensor.get_device() == -1 else "VRAM"
```

src/api/python/_api.py
```diff
@@ -257,8 +257,8 @@ def __init__(
             "FILE": nixlBind.FILE_SEG,
             "BLOCK": nixlBind.BLK_SEG,
             "OBJ": nixlBind.OBJ_SEG,
-            "cpu": nixlBind.DRAM_SEG,
-            "cuda": nixlBind.VRAM_SEG,
+            "cpu": nixlBind.DRAM_SEG,  # to be deprecated
+            "cuda": nixlBind.VRAM_SEG,  # to be deprecated
```
guy-ealey-morag:
```suggestion
            "cpu": nixlBind.DRAM_SEG,  # deprecated
            "cuda": nixlBind.VRAM_SEG,  # deprecated
```

`deprecated` means that it's discouraged from use (but still works), so I would say it's already deprecated.
Alternatively, it can say `to be removed`.

src/api/python/_api.py
yafshar: Thanks for the suggestion. I tried refactoring it into a helper, but reverted to keeping it inline to minimize changes and avoid extra indirection for a simple, readable ternary. Happy to switch back if you prefer the helper version.

yafshar: @guy-ealey-morag done!

guy-ealey-morag (approved):

yafshar:
@mkhazraee @guy-ealey-morag If there are no further comments, may we merge this?

brminich (approved):

PR #1536, opened 2026-04-15
NIXL/UCX: enforce VRAM memtype hint query behavior and add tests

yafshar:
## What?
This PR hardens UCX VRAM memtype hint handling and aligns documentation and
tests with runtime behavior.

- Enforce fail-fast behavior when an explicit `ucx_vram_memtype_hint` is set
  but UCX context memory types cannot be queried.
- Keep `auto` and `none` as non-failing modes that proceed without hinting
  when memory-type queries are unavailable.
- Document the exact supported values for `ucx_vram_memtype_hint`:
  `auto`, `none`, `cuda`, `cuda-managed`, `rocm`, `ze-device`.
- Clarify case-sensitive parsing and failure behavior for explicit hints.
- Add gtests to verify default parameter exposure and rejection of
  case-mismatched values (e.g. `CUDA`).
- Suppress expected error logs in negative-path tests to avoid triggering
  global log-failure checks.

## Why?
Previous behavior around explicit VRAM memtype hints was not fully documented
or covered by tests, which could lead to confusion and fragile failure paths.

This change improves correctness and predictability by:
- Making explicit-hint behavior deterministic when UCX cannot report
  memory-type capabilities.
- Ensuring documentation reflects actual accepted values and parsing rules.
- Adding regression coverage for default and invalid-input paths.

## How?
- Update UCX memory-type resolution to require a successful UCX context
  memory-type query for explicit hint modes.
- Keep `auto` and `none` permissive when queries are unavailable.
- Update backend and Python documentation to reflect exact values and
  failure semantics.
- Add UCX backend parameter tests covering:
  - default `ucx_vram_memtype_hint` exposure (`auto`)
  - rejection of case-mismatched hint values
- Add log-ignore guards for expected error paths in negative tests.

## Testing performed
- Ran the targeted gtest suite via the Meson test wrapper.
- Result: **12 tests from 3 suites passed, 0 failed**.

## Credits
Co-authored-by: Daniel Socek <daniel.socek@intel.com>

## Summary by CodeRabbit

* **Documentation**
  * Added docs for UCX backend initialization parameters and how to override them; documented ucx_error_handling_mode and ucx_vram_memtype_hint, accepted values, defaults, case-sensitivity, and failure semantics.

* **New Features**
  * Added ucx_vram_memtype_hint backend option to control VRAM memory-type hinting (auto, none, explicit accelerator hints; default: auto). Hint is resolved at backend init and applied to VRAM registrations when available.

* **Tests**
  * Added tests verifying ucx_vram_memtype_hint default, case-sensitive handling, and behavior when explicit hints are unsupported or unqueryable.

### DIFF
--- a/docs/BackendGuide.md
+++ b/docs/BackendGuide.md
@@ -161,6 +161,24 @@ When user asks for a specific backend for its name, alongside a list of paramete
 
 In this step, if the plugin supports talking to remote agents, the required connection data for other agents to talk to it is acquired through **getConnInfo** in SB API. And/or if it supports within node transfers, a **connection** call to itself is called, as some backends might require that.
 
+#### UCX backend options (common)
+
+The UCX plugin exposes initialization options through `getPluginParams("UCX", ...)` in C++
+or `get_plugin_params("UCX")` in Python. Two commonly used options are:
+
+| Option key | Default | Description |
+| ---------- | ------- | ----------- |
+| `ucx_error_handling_mode` | `peer` | UCX endpoint error handling policy. `peer` enables peer failure reporting; `none` disables it. |
+| `ucx_vram_memtype_hint` | `auto` | Controls how NIXL hints UCX memory type during VRAM registration. Supported policies are `auto`, `none`, and explicit accelerator policies (CUDA/CUDA managed/ROCm/ZE device). |
+
+Notes:
+
+* `ucx_error_handling_mode` and `ucx_vram_memtype_hint` are independent options. You do not need to set one to use the other.
+* `ucx_vram_memtype_hint=auto` is the recommended default for general use.
+* Matching for `ucx_vram_memtype_hint` values is case-sensitive.
+* If an explicit VRAM hint is configured but not supported by the current UCX context,
+  backend creation fails with an invalid-argument error.
+
 ### Make connections (optional):
 
 When a connection is requested to an remote agent, which is possible if the remote agent’s metadata is already loaded, the local Agent would look for common backend plugins between itself and the remote agent, and for each of them initiate a connect by using the **connect** API in SB API of such backends.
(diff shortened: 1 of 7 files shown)


brminich (commented):
why is it needed? Isn't it enough to check if GPU memory type is not detected as host?
I'd not expectd that rocm memory can be detected as cuda

yafshar:
> why is it needed? Isn't it enough to check if GPU memory type is not detected as host?
> I'd not expectd that rocm memory can be detected as cuda

This PR is needed beyond any single platform: memory-type auto-detection can be unreliable or ambiguous when allocations come from external runtimes/allocators, cross-context boundaries, or incomplete cache coverage. In those cases, relying on automatic detection alone can misclassify VRAM and fail registration.
So instead of changing UCX global slowpath behavior (which has broader performance/correctness tradeoffs), we allow explicit memtype hints from the integration layer. The init-time fail-fast check validates that the requested explicit hint is actually supported by the current UCX context (from ucp_context_query memory_types), while the host-detection check remains a separate runtime guard on the mapped allocation.
In short: one check validates configured capability early, the other validates runtime outcome.

This NIXL change was motivated by the discussion in UCX PR 11332: https://github.com/openucx/ucx/pull/11332. We are not changing UCX slowpath/cache semantics here; instead, we add an integration-level explicit memtype hint with early capability validation to reduce reliance on auto-detection in ambiguous runtime environments.

yafshar:
Hi @brminich, @mkhazraee just pinging for visibility. It’s been open for a few weeks and I wanted to check if you had a chance to look, or if there’s anything I can do to help move it forward.

brminich:
@yafshar sorry for the delay, can you pls fix merge conflicts?

PR #1545, opened 2026-04-17
PLUGINS/UCX: Clean up enums.

ColinNV:
## What?
Clean up some enums and related functions in the UCX plugin.

## Why?
- Consistent naming.
- Consistent semantics.
- Type safety.
- Smaller source files.

## Summary by CodeRabbit

* **Refactor**
  * Introduced strongly-typed UCX enums and migrated internal APIs to use them for safer, clearer backend interactions.
* **Bug Fixes**
  * Centralized UCX→internal status mapping with improved handling and warnings for unknown statuses.
* **Tests**
  * Updated unit tests to use the new typed callback identifiers.
* **Chores**
  * Build system updated to include the new compilation unit supporting these changes.

### DIFF
--- a/src/plugins/ucx/meson.build
+++ b/src/plugins/ucx/meson.build
@@ -21,9 +21,9 @@ ucx_backend_sources = ['config.cpp',
                        'mem_list.cpp',
                        'rkey.cpp',
                        'ucx_backend.cpp',
+                       'ucx_enums.cpp',
                        'ucx_plugin.cpp',
-                       'ucx_utils.cpp',
-                       'ucx_enums.cpp']
+                       'ucx_utils.cpp']
 
 ucx_backend_dependencies = [asio_dep,
                             cuda_dep,
--- a/src/plugins/ucx/meson.build
+++ b/src/plugins/ucx/meson.build
@@ -22,7 +22,8 @@ ucx_backend_sources = ['config.cpp',
                        'rkey.cpp',
                        'ucx_backend.cpp',
                        'ucx_plugin.cpp',
-                       'ucx_utils.cpp']
+                       'ucx_utils.cpp',
+                       'ucx_enums.cpp']
 
 ucx_backend_dependencies = [asio_dep,
                             cuda_dep,
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -122,7 +122,7 @@ public:
     release() {
         // TODO: Error log: uncompleted requests found! Cancelling ...
         for (nixlUcxReq req : requests_) {
-            const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+            const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
             if (ret == NIXL_IN_PROG) {
                 // TODO: Need process this properly.
                 // it may not be enough to cancel UCX request
@@ -150,7 +150,7 @@ public:
         /* If last request is incomplete, return NIXL_IN_PROG early without
          * checking other requests */
         nixlUcxReq req = requests_.back();
-        const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+        const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
         if (ret == NIXL_IN_PROG) {
             return NIXL_IN_PROG;
         } else if (ret != NIXL_SUCCESS) {
@@ -162,7 +162,7 @@ public:
         size_t incomplete_reqs = 0;
         nixl_status_t out_ret = NIXL_SUCCESS;
         for (nixlUcxReq req : requests_) {
-            const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+            const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
             if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
                 worker->reqRelease(req);
             } else if (ret == NIXL_IN_PROG) {
@@ -375,7 +375,7 @@ private:
 
 nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
     : nixlUcxEngine(init_params) {
-    if (!nixlUcxMtLevelIsSupported(nixl_ucx_mt_t::WORKER)) {
+    if (!nixlUcxMtLevelIsSupported(nixl::ucx::mt_mode_t::WORKER)) {
         throw std::invalid_argument("UCX library does not support multi-threading");
     }
 
@@ -865,7 +865,7 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
 
     auto &uw = uws.front();
     workerAddr = uw->epAddr();
-    uw->regAmCallback(nixl_ucx_cb_op_t::NOTIF_STR, notifAmCb, this);
+    uw->regAmCallback(nixl::ucx::am_cb_op_t::NOTIF_STR, notifAmCb, this);
 }
 
 nixl_mem_list_t nixlUcxEngine::getSupportedMems () const {
@@ -1393,7 +1393,7 @@ nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
         }
     };
 
-    return ep->sendAm(nixl_ucx_cb_op_t::NOTIF_STR,
+    return ep->sendAm(nixl::ucx::am_cb_op_t::NOTIF_STR,
                       nullptr,
                       0,
                       (void *)buffer->data(),
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -122,7 +122,7 @@ public:
     release() {
         // TODO: Error log: uncompleted requests found! Cancelling ...
         for (nixlUcxReq req : requests_) {
-            nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+            const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
             if (ret == NIXL_IN_PROG) {
                 // TODO: Need process this properly.
                 // it may not be enough to cancel UCX request
@@ -150,7 +150,7 @@ public:
         /* If last request is incomplete, return NIXL_IN_PROG early without
          * checking other requests */
         nixlUcxReq req = requests_.back();
-        nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+        const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
         if (ret == NIXL_IN_PROG) {
             return NIXL_IN_PROG;
         } else if (ret != NIXL_SUCCESS) {
@@ -162,7 +162,7 @@ public:
         size_t incomplete_reqs = 0;
         nixl_status_t out_ret = NIXL_SUCCESS;
         for (nixlUcxReq req : requests_) {
-            nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
+            const nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
             if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
                 worker->reqRelease(req);
             } else if (ret == NIXL_IN_PROG) {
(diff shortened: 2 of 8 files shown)


src/plugins/ucx/ucx_enums.cpp
```diff
@@ -0,0 +1,62 @@
...
+    case UCS_ERR_ENDPOINT_TIMEOUT:
+        return NIXL_ERR_REMOTE_DISCONNECT;
+    case UCS_ERR_INVALID_PARAM:
+        return NIXL_ERR_INVALID_PARAM;
+    case UCS_ERR_CANCELED:
+        return NIXL_ERR_CANCELED;
+    default:
+        NIXL_WARN << "Unexpected UCX error: " << ucs_status_string(t);
+        return NIXL_ERR_BACKEND;
+    }
```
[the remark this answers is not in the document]
ColinNV: I agree, this is from pre-existing code. Presumably somebody wanted to optimize the common case. When switching to C++20 the UCS_OK case can be folded into the switch and `[[likely]]` be used to mark it as the most common case.

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,134 @@
...
+    case nixl_ucx_ep_state_t::CONNECTED:
+        return NIXL_SUCCESS;
+    case nixl_ucx_ep_state_t::FAILED:
+        return NIXL_ERR_REMOTE_DISCONNECT;
+    case nixl_ucx_ep_state_t::NULL_:
+    case nixl_ucx_ep_state_t::DISCONNECTED:
+        return NIXL_ERR_BACKEND;
+    }
+    return NIXL_ERR_BACKEND;
+}
```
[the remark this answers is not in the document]
ColinNV: If the enum is extended the compiler will warn about it. The trailing return is because this particular function is not supposed to enforce any invariants or value restrictions.

test/unit/plugins/ucx/ucx_am_test.cpp
```diff
@@ -63,7 +63,8 @@ main() {
     uint64_t buffer;
     int ret, i;
 
-    unsigned check_cb_id = 1, rndv_cb_id = 2;
+    const auto check_cb_id = nixl_ucx_cb_op_t(42);
+    const auto rndv_cb_id = nixl_ucx_cb_op_t(43);
```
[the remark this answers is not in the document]
ColinNV: Adding values to an enum that are only used for testing is not a good idea, then production code would have to handle them in all switch statements!

Your concern might be valid in general, but not in this particular case.

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,123 @@
...
+#include <string_view>
+#include <type_traits>
+
+#include <ucs/type/status.h>
+
+#include "nixl_types.h"
+
+inline constexpr std::string_view nixl_ucx_invalid = "INVALID";
+
+enum class nixl_ucx_mt_t { SINGLE, CTX, WORKER };
```
guy-ealey-morag:
```suggestion
enum class nixl_ucx_mt_t { SINGLE, CTX, WORKER, };
```

Suggestion: improve formatting (would list them on separate lines)
Applicable to other enums in this file as well

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,123 @@
...
+        return "CONNECTED";
+    case nixl_ucx_ep_state_t::FAILED:
+        return "FAILED";
+    case nixl_ucx_ep_state_t::DISCONNECTED:
+        return "DISCONNECTED";
+    }
+    return nixl_ucx_invalid;
+}
+
+enum class nixl_ucx_cb_op_t { NOTIF_STR };
```
guy-ealey-morag: I find it confusing that this type is an enum with one value but it is also being used as a general op code with arbitrary values. (for example `42`, `43` in the test code)
Maybe separate the enum type from the variable type used in the general code?

src/plugins/ucx/ucx_enums.cpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rakhmets:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,123 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rakhmets:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <ostream>
+#include <string_view>
+#include <type_traits>
+
+#include <ucs/type/status.h>
+
+#include "nixl_types.h"
+
+inline constexpr std::string_view nixl_ucx_invalid = "INVALID";
```
rakhmets: We can use `namespace nixl::ucx` in newly created files to gradually switch to using namespaces in all files.

src/plugins/ucx/ucx_enums.h
```diff
@@ -0,0 +1,123 @@
...
+        return "SINGLE";
+    case nixl_ucx_mt_t::CTX:
+        return "CTX";
+    case nixl_ucx_mt_t::WORKER:
+        return "WORKER";
+    }
+    return nixl_ucx_invalid;
+}
+
+enum class nixl_ucx_ep_state_t { NULL_, CONNECTED, FAILED, DISCONNECTED };
```
rakhmets: Maybe `UNINITIALIZED` would be better than `NULL_`.

src/plugins/ucx/ucx_enums.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "common/nixl_log.h"
+
+#include "ucx_enums.h"
```
rakhmets: It's not regulated by our code style. However, I'd prefer to put `"name.h"` as a first include file in `"name.cpp"`.

src/plugins/ucx/ucx_backend.h
```diff
@@ -37,8 +37,7 @@
 #include "mem_list.h"
 #include "rkey.h"
 #include "ucx_utils.h"
-
-enum ucx_cb_op_t { NOTIF_STR };
+#include "ucx_enums.h"
```
rakhmets, line 39: This is also not described in code style. But I prefer to arrange the includes in alphabetical order.

src/plugins/ucx/meson.build
```diff
@@ -22,7 +22,8 @@ ucx_backend_sources = ['config.cpp',
                        'rkey.cpp',
                        'ucx_backend.cpp',
                        'ucx_plugin.cpp',
-                       'ucx_utils.cpp']
+                       'ucx_utils.cpp',
+                       'ucx_enums.cpp']
```
rakhmets: `'ucx_enums.cpp'` after `'ucx_backend.cpp'`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -122,7 +122,7 @@ class nixlUcxBackendH : public nixlBackendReqH {
     release() {
         // TODO: Error log: uncompleted requests found! Cancelling ...
         for (nixlUcxReq req : requests_) {
-            nixl_status_t ret = ucx_status_to_nixl(ucp_request_check_status(req));
+            nixl_status_t ret = ucxStatusToNixlStatus(ucp_request_check_status(req));
```
rakhmets: Please add `const`, since you are changing the line anyway.
The same for other lines in this file.

src/plugins/ucx/ucx_enums.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+std::ostream &
+operator<<(std::ostream &os, const nixl_ucx_cb_op_t t) {
+    toStream(os, t);
+    return os;
+}
+
+nixl_status_t
+ucxStatusToNixlStatus(const ucs_status_t t) {
+    if (t == UCS_OK) {
```
rakhmets: I think it makes sense to add: `__builtin_expect`, as the function is used in fast path.

src/plugins/ucx/ucx_enums.h
ColinNV: That's how I formatted it, but the clang-format CI job didn't agree.

ColinNV: @michal-shalev Since you worked on the coding style, wdyt?

src/plugins/ucx/ucx_enums.h
ColinNV: The other values are only for the test, I don't see that as a problem?

src/plugins/ucx/ucx_enums.h
guy-ealey-morag: Even with the trailing comma?
When I run clang-format it changes to this:
```c++
enum class nixl_ucx_mt_t {
    SINGLE,
    CTX,
    WORKER,
};
```

src/plugins/ucx/ucx_enums.h
guy-ealey-morag: Usually the tests imply how it should normally be used.
Real code would use only the listed values?

src/plugins/ucx/ucx_enums.h
ColinNV: I didn't use a trailing comma, will try.

src/plugins/ucx/ucx_enums.h
ColinNV: Real code wouldn't call this function directly, it's a unit test on a lower layer?

src/plugins/ucx/ucx_enums.h
ColinNV: Ok, though the gradual switch might need some help...

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -258,7 +234,7 @@ nixlUcxEp::sendAm(unsigned msg_id,
     }
 
     const ucs_status_ptr_t request =
-        ucp_am_send_nbx(eph, msg_id, hdr, hdr_len, buffer, len, &param);
+        ucp_am_send_nbx(eph, unsigned(msg_id), hdr, hdr_len, buffer, len, &param);
```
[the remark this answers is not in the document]
ColinNV: The `unsigned` was chosen to match the type expected by `ucp_am_send_nbx`.

src/plugins/ucx/ucx_enums.h
rakhmets: Yes, but we can start with UCX plugin. `config.[h|cpp]`, `mem_list.[h|cpp]`, `rkey.[h|cpp]` are using `nixl::ucx` namespace.
I think we can use it for the entire file. So, `nixl_ucx_mt_t` -> `mt_t`, `nixl_ucx_ep_state_t` -> `ep_state_t`, `nixl_ucx_cb_op_t` -> `cb_op_t`.

ColinNV: Ok, I'll put `ucx_enums.h` in the namespace and hope that we get away with it :-)

rakhmets (approved):

test/unit/plugins/ucx/ucx_am_test.cpp
```diff
@@ -63,7 +63,8 @@ main() {
     uint64_t buffer;
     int ret, i;
 
-    unsigned check_cb_id = 1, rndv_cb_id = 2;
+    const auto check_cb_id = nixl::ucx::am_cb_op_t(42);
+    const auto rndv_cb_id = nixl::ucx::am_cb_op_t(43);
```
brminich, line 67: why need to change values?

ColinNV: The test uses two distinct values, at the moment the enum only has one. The test also is a unit test on a level that is normally not exposed, I think in this case inventing additional values to test the underlying functionality is fine.

brminich (approved):

src/plugins/ucx/ucx_enums.h
brminich: yes, it is confusing, but the old valus were also out of the scope.
btw, @ColinNV do we really run this am test in CI?

ColinNV: Erm, it's not called from `test_cpp.sh`, but it works for me. Should I add it?

brminich: let's remove this test in a separate PR, it looks incorrect

PR #1546, opened 2026-04-17
Plugin Manager: Defer plugin load to the point where they are actually used

ovidiusm:
## What?

Defer plugin `dlopen` from discovery time to first use. Plugins found on disk are now recorded by name only; the actual `dlopen` happens when `createBackend()` or `getPluginParams()` is called for that specific plugin.

## Why?

https://github.com/vllm-project/vllm/issues/39521

Creating a `nixl_agent` crashes with a segfault on some systems (RHEL 9 / Rocky 9 bare-metal DGX H100) due to eager plugin loading.

During `nixlPluginManager` construction, every plugin `.so` in the plugins directory is `dlopen`'d, including plugins the user never requested (e.g. LIBFABRIC on InfiniBand systems). The `dlopen` triggers transitive dependency initialization (`libnuma` -> `numa_hook.so`) which crashes before any NIXL code even executes. The Python `nixl_agent.__init__` further aggravates this by calling `getPluginParams()` on every discovered plugin, forcing a second `dlopen` attempt for each.

## How?

**C++ (`plugin_manager.h`, `nixl_plugin_manager.cpp`, `nixl_agent.cpp`):**
- Add `discovered_backend_plugins_` and `discovered_telemetry_plugins_` sets to track plugin names found on disk without loading them.
- Add `explicit_plugin_paths_` map to preserve exact `.so` paths from the plugin list file, so `loadBackendPlugin()` tries those first before scanning directories.
- `discoverBackendPlugin()` / `discoverTelemetryPlugin()`: record the plugin name instead of calling `loadBackendPlugin()`.
- `loadPluginsFromList()`: same, record names and paths instead of `dlopen`'ing.
- Add `getAvailBackendPluginNames()` which returns loaded + discovered names. `getAvailPlugins()` calls this instead of `getLoadedBackendPluginNames()`.
- `getLoadedBackendPluginNames()` / `getLoadedTelemetryPluginNames()` remain unchanged (loaded only).

**Python (`_api.py`):**
- Remove the `for plugin in self.plugin_list: getPluginParams(plugin)` loop from `nixl_agent.__init__`.
- Add `_load_plugin_params()` helper that lazily populates the cache on first access.
- `get_plugin_params()` and `get_plugin_mem_types()` call the helper before returning.

No API changes. `get_plugin_list()` returns the same names. `create_backend()`, `getPluginParams()`, and `instantiate_all=True` all work identically. The `dlopen` just happens on demand instead of unconditionally at init.

## Summary by CodeRabbit

* **Performance**
  * Plugin parameters are now loaded on-demand instead of at startup, reducing initialization time.

* **Improvements**
  * Plugin availability reporting now includes both loaded and discovered plugins for more complete visibility.
  * Discovery of plugins is separated from actual loading, allowing explicit paths to be respected when present.
  * Improved warning-level logging when plugin parameter loading fails to aid troubleshooting.


src/core/nixl_agent.cpp
```diff
@@ -133,22 +133,43 @@ nixlDlistH::nixlDlistH(const std::string &remote_agent, descs_t &&descs)
...
+}
+
+} // namespace
+
 nixlAgentData::nixlAgentData(const std::string &name, const nixlAgentConfig &config)
     : name_(name),
       config_(config),
-      lock(config.syncMode) {
+      useEtcd(detectEtcd()),
+      lock(effectiveSyncMode(config.syncMode, useEtcd || config.useListenThread)) {
```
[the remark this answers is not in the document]
ovidiusm: Not correct, the baclend synchronization is different. Only SYNC_RW triggers WORKER mode. Both SYNC_NONE and SYNC_STRICT result in SINGLE.

src/api/python/_api.py
```diff
@@ -299,6 +307,8 @@ def get_plugin_mem_types(self, backend: str) -> list[str]:
     """
 
     def get_plugin_params(self, backend: str) -> dict[str, str]:
+        if backend not in self.plugin_b_options:
```
brminich, line 309: minor:
it can be simplified a bit
```
 if backend not in self.plugin_b_options:
        self._load_plugin_params(backend)
 if backend not in self.plugin_b_options:
        logger.warning("Plugin %s is not available to get its parameters.", backend)
        return {}
 return self.plugin_b_options[backend]
```

src/core/plugin_manager.h
```diff
@@ -160,6 +165,11 @@ class nixlPluginManager {
         loaded_backend_plugins_;
     std::map<nixl_telemetry_plugin_t, std::shared_ptr<const nixlTelemetryPluginHandle>>
         loaded_telemetry_plugins_;
+    // Plugins discovered on disk but not yet dlopen'd
+    std::set<nixl_backend_t> discovered_backend_plugins_;
+    std::set<nixl_telemetry_plugin_t> discovered_telemetry_plugins_;
```
brminich: maybe you can use existing maps with nullptr as indication that plugin is discovered?

src/core/nixl_plugin_manager.cpp
```diff
@@ -375,6 +378,21 @@ nixlPluginManager::loadBackendPlugin(const std::string &plugin_name) {
         return it->second;
     }
 
+    // Try the explicit path from the plugin list file first
+    auto path_it = explicit_plugin_paths_.find(plugin_name);
+    if (path_it != explicit_plugin_paths_.end()) {
+        const std::string &plugin_path = path_it->second;
+        if (std::filesystem::exists(plugin_path)) {
+            auto plugin_handle = loadPluginFromPath(plugin_path, backendLoader);
```
brminich, line 386: where do we load telemetry plugins now?

brminich (commented):
is some gtest possible?

src/core/plugin_manager.h
ovidiusm: `loaded_backend_plugins_` values are always valid shared_ptrs that callers dereference without null checks.

Separate sets keep the semantics clear: loaded_backend_plugins_ = loaded and ready, discovered_backend_plugins_ = name seen on disk but not loaded. Mixing this logic with nullptr guards would complicate every lookup site and leak the abstraction

src/core/nixl_plugin_manager.cpp
ovidiusm: You are right, the telemetry changes were dead code. Telemetry is already using deferred plugin load (decoupled discover and load) so we do not have to add bookkeeping for it

ovidiusm: Telemetry plugin load is called from nixlTelemetry::initializeTelemetry() when an agent has telemetry enabled

brminich (approved):

src/core/plugin_manager.h
ColinNV: What about merging `discovered_backend_plugins_` and `explicit_plugin_paths_`?

src/core/nixl_plugin_manager.cpp
```diff
@@ -570,6 +586,23 @@ nixlPluginManager::getLoadedBackendPluginNames() {
     return names;
 }
 
+std::vector<nixl_backend_t>
+nixlPluginManager::getAvailBackendPluginNames() {
+    lock_guard lg(lock);
```
ColinNV: `const`

src/core/nixl_plugin_manager.cpp
```diff
@@ -457,9 +475,11 @@ void
 nixlPluginManager::discoverBackendPlugin(const std::string &filename) {
     if (startsWith(filename, backendPluginPrefix) && endsWith(filename, kPluginSuffix)) {
         std::string plugin_name = extractPluginName(filename, backendPluginPrefix);
-        auto plugin = loadBackendPlugin(plugin_name);
-        if (plugin) {
-            NIXL_INFO << "Discovered and loaded backend plugin: " << plugin_name;
+
+        lock_guard lg(lock);
```
ColinNV: `const`

ColinNV (approved):

src/core/plugin_manager.h
ovidiusm: It's an interesting idea, technically it's possible but my instinct is that they serve different purposes, so I would keep them separate

brminich (approved):

PR #1547, opened 2026-04-17
gtest: allow to test without UCX

oliviergaraud:
## What?
Don't condition gtest on UCX availability.

## Why?
The gtest suite was only built when UCX was enabled, which meant OBJ plugin unit tests were silently skipped on builds without UCX.

## Summary by CodeRabbit

* **Tests**
  * Modified test build configuration to always include the gtest test suite during test builds.
  * Refined conditional dependencies for hardware warning tests to ensure proper test execution when required components are available.

### DIFF
--- a/test/gtest/meson.build
+++ b/test/gtest/meson.build
@@ -81,7 +81,7 @@ else
     device_api_dep = []
 endif
 
-if ucx_dep.found()
+if ucx_dep.found() and is_variable('ucx_backend_interface')
     gtest_sources += 'hw_warning_test.cpp'
     ucx_hw_warning_dep = [ucx_backend_interface, ucx_dep]
 else
--- a/test/meson.build
+++ b/test/meson.build
@@ -16,6 +16,4 @@
 subdir('nixl')
 subdir('unit')
 
-if enabled_plugins.get('UCX') and ucx_dep.found()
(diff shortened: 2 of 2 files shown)


test/gtest/meson.build
```diff
@@ -81,7 +81,7 @@ else
     device_api_dep = []
 endif
 
-if ucx_dep.found()
+if ucx_dep.found() and is_variable('ucx_backend_interface')
```
brminich, line 84: this specific test is relevant when UCX is present, why is this change needed?

oliviergaraud: UCX libraries can be installed on the system but the UCX plugin might not be built (e.g., disabled via build options)

brminich (approved):

oliviergaraud:
Hi @brminich, I can see that [nixl-ci-gpu](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/1485/pipeline/) failed but I cannot see the results, so I don't know how to act on that.
Do you have advice ?

oliviergaraud:
@aranadive, @brminich is there something I can do regarding the check that failed ?

PR #1551, opened 2026-04-18
Fix NIXL_ERR_NOT_FOUND when deregistering FILE/BLK/OBJ with len==0

flpanbin:
## What?
Fix deregisterMem returning `NIXL_ERR_NOT_FOUND` for `FILE_SEG`, `BLK_SEG`, and `OBJ_SEG` descriptors that were registered with `len == 0`.

In `nixlLocalSection::addDescList`, zero-length descriptors of these types are internally stored with `len = SIZE_MAX`("File has no range limit"). However, `nixlLocalSection::remDescList` performed lookups using the original `len == 0`, causing getIndex to fail because the stored and queried lengths no longer match.

This change aligns the lookup logic in `remDescList` with the storage logic in `addDescList`.

## Why?
When a plugin or application registers a file-like segment without knowing its exact size upfront (e.g., `len = 0`), NIXL core silently upgrades the stored length to `SIZE_MAX`. During cleanup, `deregisterMem` fails because `remDescList` cannot find the original (`addr=0, len=0`) descriptor in a list where it was stored as (`addr=0, len=SIZE_MAX`).

## How?
In `remDescList`, before each `target.getIndex()` call, construct a temporary `nixlBasicDesc` query from the user-provided element. If the memory type is `FILE_SEG`, `BLK_SEG`, or `OBJ_SEG` and `query.len == 0`, set `query.len = SIZE_MAX` so that the lookup matches the internal representation created by `addDescList`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Zero-length queries for file, block, and object segments now consistently match stored entries by address and identifier; DRAM segment behavior unchanged. This reduces incorrect "not found" results for those queries.

* **Refactor**
  * Centralized query normalization so lookups behave uniformly across code paths.

* **Tests**
  * Added unit tests covering matching and non-matching zero-length query scenarios.

### DIFF
--- a/src/infra/mem_section.h
+++ b/src/infra/mem_section.h
@@ -119,6 +119,20 @@ private:
 using nixl_sec_dlist_t = nixlSecDescList;
 using section_map_t = std::map<section_key_t, nixlSecDescList>;
 
+/**
+ * @brief Normalize a section descriptor for file-like segments.
+ *
+ * For BLK_SEG, OBJ_SEG, and FILE_SEG, a zero-length registration is stored
+ * internally with len = SIZE_MAX (unlimited range).
+ */
+inline nixlBasicDesc
+normalizeSecDesc(const nixlBasicDesc &desc, nixl_mem_t type) {
+    if ((type == BLK_SEG || type == OBJ_SEG || type == FILE_SEG) && desc.len == 0) {
+        return nixlBasicDesc(desc.addr, SIZE_MAX, desc.devId);
+    }
+    return desc;
+}
+
 class nixlMemSection {
     protected:
         std::array<backend_set_t, FILE_SEG+1>         memToBackend;
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -228,13 +228,21 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
     // First check if the mem_elms are present in the list,
     // don't deregister anything in case any is missing.
     for (auto & elm : mem_elms) {
-        int index = target.getIndex(elm);
+        nixlBasicDesc query = elm;
+        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) || (nixl_mem == FILE_SEG)) &&
+            (query.len == 0))
+            query.len = SIZE_MAX;
+        int index = target.getIndex(query);
         if (index < 0)
             return NIXL_ERR_NOT_FOUND;
     }
 
     for (auto & elm : mem_elms) {
-        int index = target.getIndex(elm);
+        nixlBasicDesc query = elm;
+        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) || (nixl_mem == FILE_SEG)) &&
+            (query.len == 0))
+            query.len = SIZE_MAX;
+        int index = target.getIndex(query);
         // Already checked, elm should always be found. Can add a check in debug mode.
         backend->deregisterMem(target[index].metadataP);
         target.remDesc(index);
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -181,10 +181,7 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
             }
         }
 
-        *lp = mem; // Copy the basic desc part
-        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) ||
-             (nixl_mem == FILE_SEG)) && (lp->len==0))
-            lp->len = SIZE_MAX; // File has no range limit
+        *lp = normalizeSecDesc(mem, nixl_mem); // Copy the basic desc part
 
         local_batch.push_back(local_sec);
 
--- a/src/infra/nixl_descriptors.cpp
+++ b/src/infra/nixl_descriptors.cpp
@@ -446,8 +446,10 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
 
 int
 nixlSecDescList::getIndex(const nixlBasicDesc &query) const {
-    auto itr = std::lower_bound(this->descs.begin(), this->descs.end(), query);
-    if (itr == this->descs.end() || static_cast<const nixlBasicDesc &>(*itr) != query)
+    const nixlBasicDesc adjusted_query = normalizeSecDesc(query, this->getType());
+
+    auto itr = std::lower_bound(this->descs.begin(), this->descs.end(), adjusted_query);
+    if (itr == this->descs.end() || static_cast<const nixlBasicDesc &>(*itr) != adjusted_query)
         return NIXL_ERR_NOT_FOUND;
     return static_cast<int>(itr - this->descs.begin());
 }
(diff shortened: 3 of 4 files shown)


flpanbin:
Hi @brminich, Could you please take a look this PR?

guy-ealey-morag:
Hi @flpanbin , 
I think this solution is not robust enough because it requires modifying the `len` before every call to `getIndex`.
Ideally `getIndex` itself should know how to handle this without needing a pre-processing stage such that every call would take this edge case into account, or alternatively a wider refactor would allow unlimited length without changing the `len` attr directly.

Also, I think a new test should be added to verify this edge case (a test that would fail without this fix).

src/infra/nixl_memory_section.cpp
```diff
@@ -228,13 +228,21 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
     // First check if the mem_elms are present in the list,
     // don't deregister anything in case any is missing.
     for (auto & elm : mem_elms) {
-        int index = target.getIndex(elm);
+        nixlBasicDesc query = elm;
+        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) || (nixl_mem == FILE_SEG)) &&
+            (query.len == 0))
+            query.len = SIZE_MAX;
+        int index = target.getIndex(query);
```
ColinNV: `const`

src/infra/nixl_memory_section.cpp
```diff
@@ -228,13 +228,21 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
     // First check if the mem_elms are present in the list,
     // don't deregister anything in case any is missing.
     for (auto & elm : mem_elms) {
-        int index = target.getIndex(elm);
+        nixlBasicDesc query = elm;
+        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) || (nixl_mem == FILE_SEG)) &&
```
ColinNV: Please use braces.

ColinNV (commented):
Would be great with a small unit test.

flpanbin:
> Hi @flpanbin , I think this solution is not robust enough because it requires modifying the `len` before every call to `getIndex`. Ideally `getIndex` itself should know how to handle this without needing a pre-processing stage such that every call would take this edge case into account, or alternatively a wider refactor would allow unlimited length without changing the `len` attr directly.
> 
> Also, I think a new test should be added to verify this edge case (a test that would fail without this fix).

@guy-ealey-morag Thanks for the detailed feedback. I have updated PR to address it. Please take another look when you are free.

src/infra/nixl_memory_section.cpp
flpanbin: Thanks for your review, I have updated code.

src/infra/mem_section.h
```diff
@@ -119,6 +119,21 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
...
 
+/**
+ * @brief Normalize a section descriptor query for file-like segments.
+ *
+ * For BLK_SEG, OBJ_SEG, and FILE_SEG, a zero-length registration is stored
+ * internally with len = SIZE_MAX (unlimited range). This helper applies the
+ * same transformation to a query so that lookups remain symmetric with storage.
+ */
+inline nixlBasicDesc
+normalizeSecQuery(const nixlBasicDesc &q, nixl_mem_t type) {
```
guy-ealey-morag:
```suggestion
 * @brief Normalize a section descriptor for file-like segments.
 *
 * For BLK_SEG, OBJ_SEG, and FILE_SEG, a zero-length registration is stored
 * internally with len = SIZE_MAX (unlimited range).
 */
inline nixlBasicDesc
normalizeSecDesc(const nixlBasicDesc &desc, nixl_mem_t type) {
```

The term "query" is relevant only for searching (`getIndex`), and may be confusing when it's used on insert.
I would keep the docs generic so they easily apply to both use-cases.

src/infra/nixl_memory_section.cpp
```diff
@@ -181,10 +181,7 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
             }
         }
 
-        *lp = mem; // Copy the basic desc part
-        if (((nixl_mem == BLK_SEG) || (nixl_mem == OBJ_SEG) ||
-             (nixl_mem == FILE_SEG)) && (lp->len==0))
-            lp->len = SIZE_MAX; // File has no range limit
+        *lp = normalizeSecQuery(mem, nixl_mem); // Copy the basic desc part
```
guy-ealey-morag: This is definitely an improvement.
Looking at the code again I think that the normalization should happen one layer deeper at `addDescs`/`addDesc`.
The reason for it is that this invariant should be kept by `nixlSecDescList`, and it would be fail-safe to not assume that the input is already given normalized.

Something like:
```c++
bool
nixlSecDescList::usesUnboundedLen() const noexcept {
    return type == BLK_SEG || type == OBJ_SEG || type == FILE_SEG;
}

void
nixlSecDescList::normalize(nixlSectionDesc &desc) const noexcept {
    if (!usesUnboundedLen()) return;
    if (d.len == 0) d.len = SIZE_MAX;
}

void
nixlSecDescList::normalizeBatch(std::vector<nixlSectionDesc> &batch) const noexcept {
    if (!usesUnboundedLen()) return;
    for (auto &d : batch) {
        if (d.len == 0) d.len = SIZE_MAX;
    }
}

void
nixlSecDescList::addDescs(std::vector<nixlSectionDesc> batch, order ord) {
    if (batch.empty()) return;

    if (batch.size() == 1) {
        // It's more efficient to insert a single element directly
        addDesc(std::move(batch[0]));
        return;
    }

    normalizeBatch(batch);

    if (ord == order::SORTED) {
        NIXL_ASSERT(std::is_sorted(batch.begin(), batch.end()));
    } else {
        std::sort(batch.begin(), batch.end());
    }

    addSortedDescs(std::move(batch));
}
```

(And similarly in `addDesc` for the single insertion case)

This is a sketch, let me know what you think

guy-ealey-morag: (This may be out of scope for this PR)

flpanbin: Thanks for the suggestion. I agree that pushing the normalization down into `nixlSecDescList::addDesc/addDescs` would be a cleaner design. However, since that touches the insertion API surface and could affect other callers, I'd prefer to keep this PR focused on the getIndex fix. Happy to follow up with a separate refactor PR if you'd like.

src/infra/mem_section.h
flpanbin: done.

src/infra/nixl_memory_section.cpp
guy-ealey-morag: A follow up PR would be great

guy-ealey-morag (approved):

flpanbin:
Hi @guy-ealey-morag Thanks for your approval and the valuable feedback! The PR appears to be in a blocked state. Do we need additional approvals from other maintainers before merging, or is there something else blocking the merge?

iyastreb (approved):

flpanbin:
Hi @brminich, Could you please take a look this PR again?

flpanbin:
Hi @guy-ealey-morag @iyastreb, Thank you both for approving this PR! Could you please help move this forward by merging it? Your approval is greatly appreciated. Thanks!

PR #1557, opened 2026-04-20
nixl_ep: Adding ep tests to dl-gpu CI

lishapira:
Integrates nixl_ep elastic tests into CI by adding elastic coverage to the existing nixl-ci-dl-gpu Jenkins pipeline. RDMA support is not enabled yet and is planned for a future update.

Changes:
- CI: Add “Run DL EP elastic tests” to test-dl-matrix.yaml and build the PR image with BUILD_NIXL_EP=true.
- test_ep.sh: EP CI script that runs elastic.py on 4 GPUs with no_expansion.json and expansion_fault_contraction.json to exercise nixl_ep elastic scale-up/scale-down.
- Build: Enable nixl_ep builds from CI, copy DOCA GPUNetIO headers, and use an EP-local buildtype=release override to avoid CUDA -G/nvlink register-count failures.
- elastic.py: Add phase-local failure validation behind --validate-phase-failures and change failure log wording.
- Declare nixl_ep.topk_idx_t in the meta-dispatcher TYPE_CHECKING block so mypy can see the runtime-exported attribute.
 

## Summary by CodeRabbit

* **New Features**
  * Elastic Parallel GPU test capability with optional strict phase-failure validation and NVLink/transport toggles.
  * Configurable build flag to enable the GPU endpoint feature and conditional DOCA GPUNetIO header handling during builds.

* **Tests**
  * Added new elastic GPU test plans and a CI script to run EP elastic tests.

* **Chores**
  * Updated CI image tags and pipeline configuration to include the new build flag and test stage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1557?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.gitlab/test_ep.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rakhmets:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

lishapira:
/ok to test https://github.com/ai-dynamo/nixl/commit/c40ba34a5ed3ca904b3ea588060b08c25f4adddc

lishapira:
/ok to test https://github.com/ai-dynamo/nixl/commit/0830214a8f43c693b62b22ee66c6064d66d2174e

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -577,16 +577,23 @@ def worker(torch_rank: int, args: argparse.Namespace):
...
+            # Reject failures not marked for kill in the plan.
+            unexpected_failures = newly_failed_ranks - set(ranks_to_kill)
+            assert (
+                not unexpected_failures
+            ), f"rank {global_rank}, local_rank={local_rank} phase {plan.get_phase()}: unexpected failures {unexpected_failures}"
+
         if len(newly_failed_ranks) > 0:
             print(
-                f"global_rank={global_rank}, local_rank={local_rank} -> detected unexpected rank failures: {newly_failed_ranks}, cleaning up...",
+                f"global_rank={global_rank}, local_rank={local_rank} -> detected expected rank failures: {newly_failed_ranks}, cleaning up...",
```
itayalroy:
```suggestion
                f"global_rank={global_rank}, local_rank={local_rank} -> detected rank failures: {newly_failed_ranks}, cleaning up...",
```

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -577,16 +577,23 @@ def worker(torch_rank: int, args: argparse.Namespace):
...
+        # Query mask buffer to detect rank failures and clean them up
         buffer.query_mask_buffer(mask_status)
         newly_failed_ranks = set()
         for r in range(current_num_ranks):
             if mask_status[r].item() != 0 and r in remote_ranks:
                 newly_failed_ranks.add(r)
 
+        if args.validate_plan:
+            # Reject failures not marked for kill in the plan.
+            unexpected_failures = newly_failed_ranks - set(ranks_to_kill)
```
itayalroy: If we're already validating failed ranks let's do it according to the plan phase. If we detected failure of rank X on phase 1 and it is supposed to be killed in phase 2 then something is wrong

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -645,6 +652,11 @@ def main():
         default=DEFAULT_TIMEOUT_MS,
         help="GPU timeout in milliseconds (non-negative integer)",
     )
+    parser.add_argument(
+        "--validate-plan",
```
itayalroy: validate plan really does not suggest what this flag is actually doing, we need a better name for it, or maybe even consider not making it a flag as it is mainly aimed for CI, not users.

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -667,16 +679,26 @@ def main():
...
             failed.append((i, p.exitcode))
     if failed:
         raise RuntimeError(
             f"Worker processes failed: {', '.join(f'worker {i} (exit code {code})' for i, code in failed)}"
         )
 
+    if args.validate_plan:
+        # Verify that exactly the expected number of ranks were killed via SIGTERM.
+        plan = Plan(args.plan, rank=0)
+        expected_kills = plan.get_total_killed_ranks()
```
itayalroy: I don't think this will work for this plan:
```
(1,2), (1,-2), (1,2,3)
```
Rank 3 joins after rank 2 is killed, so it will not see any failures, therefore `sigterm_count` will be 0 and `expected_kills` will be 1.

Also in general counting total num failures is not very safe

examples/device/ep/meson.build
```diff
@@ -92,6 +97,7 @@ nixl_ep_cuda_args = [
     '-DHAVE_CUDA',
     '-DTORCH_EXTENSION_NAME=nixl_ep_cpp',
     '--expt-relaxed-constexpr',  # Allow calling constexpr __host__ functions from __device__ functions
+    '-rdc=false',  # Override global -rdc=true to avoid nvlink register count errors with nixlPut
```
itayalroy: Can you clarify why is this necessary? I managed to build for all CUDA archs w/o it

itayalroy (changes requested):

examples/device/ep/tests/elastic/elastic.py
lishapira: changed

examples/device/ep/tests/elastic/elastic.py
lishapira: renamed to --validate-phase-failures

examples/device/ep/tests/elastic/elastic.py
lishapira: changed to phase-local validation

examples/device/ep/meson.build
lishapira: Without it I hit nvlink errors:
  nvlink error: entry function '...combine<...>...'
  with max regcount of 96 calls function 'nixlPut<WARP>' with regcount of 215
and similarly for dispatch kernels capped at 128 regs.
The EP kernels are compiled with a per-kernel register cap of 96 or 128 registers/thread.
Under -rdc=true, nvlink enforces callee regcount ≤ caller max regcount across the call boundary. With nixlPut compiling to 215 registers as a separately-linked symbol, it cannot be called from kernels capped at 96/128.
-rdc=false on EP only inlines nixlPut into the EP kernel TU, so there’s no cross-call register check

lishapira:
/ok to test https://github.com/ai-dynamo/nixl/commit/c693b8518cada9b951439dcee16fff99759d7d2c

lishapira:
/ok to test https://github.com/ai-dynamo/nixl/commit/c693b8518cada9b951439dcee16fff99759d7d2c

examples/device/ep/meson.build
lishapira: changed

.gitlab/test_ep.sh
```diff
@@ -0,0 +1,105 @@
...
+    echo "/sys/module/nvidia_peermem/version not found"
+fi
+
+if [ -f /sys/module/nv_peer_mem/version ]; then
+    cat /sys/module/nv_peer_mem/version
+else
+    echo "/sys/module/nv_peer_mem/version not found"
+fi
+
+start_etcd_server "/nixl/python_ci"
```
itayalroy: elastic.py does not use etcd, no need to launch an etcd server

examples/device/ep/meson.build
```diff
@@ -120,7 +126,7 @@ nixl_ep_rpath += ':' + join_paths(nixl_lib_dir, 'plugins')
 # For now, nixl ep cannot be built with -G due to register usage limits
 nixl_ep_override_options = []
 if get_option('buildtype') == 'debug'
-    nixl_ep_override_options = ['optimization=3']
+    nixl_ep_override_options = ['buildtype=custom', 'optimization=3']
```
itayalroy: explain?

lishapira: NIXL CI is configured with global Meson --buildtype=debug.
For CUDA targets, Meson buildtype=debug adds -G, which enables CUDA device debug.
For now, NIXL EP cannot be built with -G due to register usage limits.
Using buildtype=custom for the EP target overrides the global buildtype=debug for that target, so Meson contributes no buildtype-specific CUDA flags for EP, including no -G.
We could use another buildtype such as release, but I thought custom fits better because we explicitly add -g separately and define optimization=3.

src/bindings/python/nixl-meta/nixl_ep/__init__.py
```diff
@@ -75,5 +75,8 @@ def _load_ep_module() -> str:
         setattr(sys.modules[__name__], attr, getattr(_pkg, attr))
 
 if TYPE_CHECKING:
+    import torch
```
lishapira, line 78: topk_idx_t is exported dynamically by the nixl_ep meta-dispatcher, so it exists at runtime but mypy cannot infer it. Declaring it as torch.dtype under TYPE_CHECKING makes the static type surface match the runtime API.

This was added after mypy reported: Module has no attribute "topk_idx_t" in elastic.py.

.gitlab/build.sh
```diff
@@ -358,6 +358,31 @@ else
...
 fi # PRE_INSTALLED_ENV end
 
+# PRE_INSTALLED_ENV skips apt bootstrap, so DOCA GPUNetIO dev headers are absent;
+# nixl_ep CUDA builds need them (UCX device code pulls doca_gpunetio_*.cuh).
+if [ -n "${PRE_INSTALLED_ENV}" ] && [ "${BUILD_NIXL_EP}" = "true" ] && $HAS_GPU && [ -d "${CUDA_HOME}" ]; then
+    if ! dpkg-query -W -f='${Status}' libdoca-sdk-gpunetio-dev 2>/dev/null | grep -q "ok installed"; then
+        echo "PRE_INSTALLED_ENV: installing DOCA GPUNetIO dev packages for nixl_ep build"
+        ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi)
+        MELLANOX_OS="$(. /etc/lsb-release; echo "${DISTRIB_ID}${DISTRIB_RELEASE}" | tr '[:upper:]' '[:lower:]' | tr -d .)"
+        wget --tries=3 --waitretry=5 --no-verbose \
```
dpressle: we already have doca 3.3.0 in the base image see line 178 why we need different doca version? why duplicate the doca install logic? IMO it should be the same in the same place

.gitlab/build.sh
```diff
@@ -358,6 +358,31 @@ else
     )
 fi # PRE_INSTALLED_ENV end
 
+# PRE_INSTALLED_ENV skips apt bootstrap, so DOCA GPUNetIO dev headers are absent;
```
dpressle: I dont understand why we need to run this every PR (-n "${PRE_INSTALLED_ENV}") as far as i see this whole block should be part of base image and be located in the if block that ends in line 359

dpressle (changes requested):

.gitlab/build.sh
lishapira: changed

.gitlab/build.sh
lishapira: changed

dpressle (dismissed):

.gitlab/test_ep.sh
```diff
@@ -0,0 +1,103 @@
...
+
+export LD_LIBRARY_PATH=${INSTALL_DIR}/lib:${INSTALL_DIR}/lib/$ARCH-linux-gnu:${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins:/usr/local/lib:$LD_LIBRARY_PATH
+export CPATH=${INSTALL_DIR}/include::$CPATH
+export PATH=${INSTALL_DIR}/bin:$PATH
+export PKG_CONFIG_PATH=${INSTALL_DIR}/lib/pkgconfig:$PKG_CONFIG_PATH
+export NIXL_PLUGIN_DIR=${INSTALL_DIR}/lib/$ARCH-linux-gnu/plugins
+export NIXL_PREFIX=${INSTALL_DIR}
+export NIXL_DEBUG_LOGGING=yes
+# Force eager protocol; UCX skips GPU-direct RNDV path.
+export UCX_RNDV_THRESH=inf
```
itayalroy: is this needed? why?

examples/device/ep/meson.build
```diff
@@ -18,6 +18,11 @@ if not cuda_dep.found()
     subdir_done()
 endif
 
+if not ucx_gpu_device_api_available
```
itayalroy: This seems weird, why does nixl ep directly check if ucx device api is available? nixl ep should depend on nixl device api, not ucx.
Is this change needed for this PR? if it is just some defensive code, let's remove it from this PR

examples/device/ep/meson.build
itayalroy: I would rather use release here as this is how we build it for release

itayalroy (commented):
Overall lgtm, some minor comments

.gitlab/test_ep.sh
```diff
@@ -0,0 +1,103 @@
...
+        unset NIXL_ETCD_ENDPOINTS NIXL_ETCD_PEER_URLS NIXL_ETCD_NAMESPACE
+        unset UCX_NET_DEVICES  # let UCX auto-select GPU-capable transport
+        # Force NVLink-only transports.
+        if [[ "$extra_flags" != *--disable-ll-nvlink* ]]; then
+            export UCX_TLS=cuda_copy,cuda_ipc,sm,self
+        fi
+        PYTHONPATH="${NIXL_BUILD_DIR}/${EP_SRC_DIR}:${EP_SRC_DIR}/tests:${EP_SRC_DIR}/tests/elastic${PYTHONPATH:+:$PYTHONPATH}" \
+        timeout 300 python3 ${EP_SRC_DIR}/tests/elastic/elastic.py \
+            --plan "$plan_file" \
+            --num-processes 4 --num-topk 4 --validate-phase-failures $extra_flags
```
itayalroy: Can we run some more interesting params? let's do 32 experts per rank, topk 8, num tokens 256

lishapira: changed

examples/device/ep/meson.build
lishapira: removed

.gitlab/test_ep.sh
lishapira: removed

itayalroy (dismissed):

itayalroy (approved):

dpressle (approved):

PR #1565, opened 2026-04-22
PLUGINS/UCX: Improve progress and notif handling.

ColinNV:
## What?
- Simplify and optimize notif handling in the UCX backend.
- Add a `progressLoop()` function and call it consistently.

## How?
- Remove the second `notif_list_t` from some NIXL UCX engines.
- Simplify `getNotifs()` to a single `std::vector::swap()` with lock.
- Change the return type of `progress()` to match that of `ucp_worker_progress()`.
- Add `progressLoop()` functions that loop until zero, and use them where possible.
- Consistently call `progressLoop()` in `getNotifs()` where applicable.
- Remove the unused member variable `nixlUcxEngine:: progressThreadEnabled_`.

## Summary by CodeRabbit

* **Refactor**
  * Unified notification queuing into a single, thread-safe path for more reliable and consistent delivery.
  * Replaced busy-wait progress polling with a centralized progress loop to reduce CPU spin and improve responsiveness.
  * Simplified engine threading/state variants for more predictable backend behavior and consistent notification retrieval.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1398,6 +1398,7 @@ nixlUcxEngine::getConnection(const std::string &remote_agent) const {
 
 void
 nixlUcxEngine::appendNotif(std::string &&remote_name, std::string &&msg) {
+    // In the "no progress thread" case the lock in nixlAgent is sufficient.
     notifList_.emplace_back(std::move(remote_name), std::move(msg));
 }
 
@@ -1432,6 +1433,7 @@ nixlUcxEngine::getNotifs(notif_list_t &notif_list) {
 
     progressLoop();
 
+    // In the "no progress thread" case the lock in nixlAgent is sufficient.
     notifList_.swap(notif_list);
     return NIXL_SUCCESS;
 }
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -393,7 +393,6 @@ nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
 
     const std::lock_guard lock(notifMutex_);
     notifList_.swap(notif_list);
-    notifList_.reserve(notif_list.size());
     return NIXL_SUCCESS;
 }
 
@@ -778,7 +777,6 @@ nixlUcxThreadPoolEngine::getNotifs(notif_list_t &notif_list) {
 
     const std::lock_guard lock(notifMutex_);
     notifList_.swap(notif_list);
-    notifList_.reserve(notif_list.size());
     return NIXL_SUCCESS;
 }
 
@@ -1435,7 +1433,6 @@ nixlUcxEngine::getNotifs(notif_list_t &notif_list) {
     progressLoop();
 
     notifList_.swap(notif_list);
-    notifList_.reserve(notif_list.size());
     return NIXL_SUCCESS;
 }
 
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -30,15 +30,6 @@
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_split.h"
 #include <asio.hpp>
-namespace {
-    void moveNotifList(notif_list_t &src, notif_list_t &tgt)
-    {
-        if (src.size() > 0) {
-            std::move(src.begin(), src.end(), std::back_inserter(tgt));
-            src.clear();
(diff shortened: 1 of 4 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -392,23 +380,20 @@ nixlUcxThreadEngine::~nixlUcxThreadEngine() {
...
+    if (!notif_list.empty()) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
 
-    getNotifsImpl(notif_list);
-    const std::lock_guard<std::mutex> lock(notifMtx_);
-    moveNotifList(notifPthr_, notif_list);
+    const std::lock_guard lock(notifMutex_);
+    notifList_.swap(notif_list);
+    notifList_.reserve(notif_list.size());
```
iyastreb: reserve here is suspicious
We may call this function many times before it actually returns a notification
So each time we reserve (means allocate) an array just for nothing?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -777,26 +761,24 @@ nixlUcxThreadPoolEngine::sendXferRange(const nixl_xfer_op_t &operation,
...
-        progress();
+        progressLoop();
     }
 
-    getNotifsImpl(notif_list);
-    std::lock_guard<std::mutex> lock(notifMutex_);
-    moveNotifList(notifThread_, notif_list);
+    const std::lock_guard lock(notifMutex_);
+    notifList_.swap(notif_list);
+    notifList_.reserve(notif_list.size());
```
iyastreb: same here

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1437,18 +1426,16 @@ nixlUcxEngine::notifAmCb(void *arg, const void *header,
...
 
-nixl_status_t nixlUcxEngine::getNotifs(notif_list_t &notif_list)
-{
-    if (!notif_list.empty()) return NIXL_ERR_INVALID_PARAM;
+    progressLoop();
 
-    while (progress())
-        ;
-    getNotifsImpl(notif_list);
+    notifList_.swap(notif_list);
```
guy-ealey-morag, line 1437: no need for the lock here?

ColinNV: This is the `nixlUcxEngine` without a progress thread, and all calls to the engine are under protection of the mutex in `nixlAgent`.

guy-ealey-morag: Okay, I suggest adding a note about this in the code

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -820,8 +800,7 @@ nixlUcxEngine::create(const nixlBackendInitParams &init_params) {
 
 nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     : nixlBackendEngine(&init_params),
-      sharedWorkerIndex_(1),
-      progressThreadEnabled_(init_params.enableProgTh) {
+      sharedWorkerIndex_(1) {
```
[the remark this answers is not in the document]
ColinNV: You are right, the description was wrong, the variable that was unused and removed is called `progressThreadEnabled_`.

iyastreb (approved):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -777,26 +760,23 @@ nixlUcxThreadPoolEngine::sendXferRange(const nixl_xfer_op_t &operation,
...
 nixlUcxThreadPoolEngine::getNotifs(notif_list_t &notif_list) {
-    if (!notif_list.empty()) return NIXL_ERR_INVALID_PARAM;
+    if (!notif_list.empty()) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
 
     if (!sharedThread_) {
-        progress();
+        progressLoop();
     }
```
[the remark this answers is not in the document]
ColinNV: If this is an issue it's pre-existing.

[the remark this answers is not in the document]
iyastreb: yes, it's a known limitation
Ideally we must split shared and dedicated workers, and this way we can set NO_LOCK mode for dedicated ones.
But this is not very trivial task, because during connect we need to register MDs for all workers.
I think we may skip this task for now, as it does not harm much

guy-ealey-morag (approved):

brminich (approved):

PR #1567, opened 2026-04-22
Python wheel packaging: dedup symlinked files

ovidiusm:
## What?

Deduplicate UCX transport plugin shared libraries in the wheel to prevent double initialization at runtime.

## Why?

UCX plugins are installed with standard versioned symlink chains (e.g. `libuct_ib.so` → `libuct_ib.so.0` → `libuct_ib.so.0.0.0`). Python wheels are ZIP files and do not support symlinks, so when these plugins are packaged, the symlinks become independent regular files.

With symlinks on a normal filesystem, all names resolve to the same file and `dlopen` returns the same handle. Without symlinks, each copy is a distinct file and the dynamic linker can load the same library multiple times under different names. For example, `libuct_ib.so.0` gets pulled in via `DT_NEEDED` from `libuct_rdmacm`, while `libuct_ib.so.0.0.0` is loaded separately by UCX's module scanner. This creates two IB component instances, causing a `wireup.c` assertion failure due to MD/component mismatch.

As a side effect, every UCX plugin was also tripled in the wheel (30 files instead of 10), adding ~5 MB to the compressed wheel size.

## How?

`copytree()` now groups `.so` files by `(st_dev, st_ino)` and keeps only one file per group: the real (non-symlink) file, which is the fully-versioned `.so.X.Y.Z` name that UCX's `module.c` dlopens.

After copying, a second `patchelf --replace-needed` pass rewrites any inter-plugin `DT_NEEDED` entries that referenced removed symlink names (e.g. `libuct_ib.so.0` → `libuct_ib.so.0.0.0`) so dependencies resolve via RPATH regardless of load order.

### Before

```
       60561  nixl_cu12.libs/ucx/libucm_cuda.so
       60561  nixl_cu12.libs/ucx/libucm_cuda.so.0
       60561  nixl_cu12.libs/ucx/libucm_cuda.so.0.0.0
       43081  nixl_cu12.libs/ucx/libuct_cma.so
       43081  nixl_cu12.libs/ucx/libuct_cma.so.0
       43081  nixl_cu12.libs/ucx/libuct_cma.so.0.0.0
      174465  nixl_cu12.libs/ucx/libuct_cuda.so
      174465  nixl_cu12.libs/ucx/libuct_cuda.so.0
      174465  nixl_cu12.libs/ucx/libuct_cuda.so.0.0.0
      365641  nixl_cu12.libs/ucx/libuct_ib.so
      365641  nixl_cu12.libs/ucx/libuct_ib.so.0
      365641  nixl_cu12.libs/ucx/libuct_ib.so.0.0.0
       77073  nixl_cu12.libs/ucx/libuct_ib_efa.so
       77073  nixl_cu12.libs/ucx/libuct_ib_efa.so.0
       77073  nixl_cu12.libs/ucx/libuct_ib_efa.so.0.0.0
      532361  nixl_cu12.libs/ucx/libuct_ib_mlx5.so
      532361  nixl_cu12.libs/ucx/libuct_ib_mlx5.so.0
      532361  nixl_cu12.libs/ucx/libuct_ib_mlx5.so.0.0.0
       64617  nixl_cu12.libs/ucx/libuct_ib_mlx5_gda.so
       64617  nixl_cu12.libs/ucx/libuct_ib_mlx5_gda.so.0
       64617  nixl_cu12.libs/ucx/libuct_ib_mlx5_gda.so.0.0.0
       93977  nixl_cu12.libs/ucx/libuct_rdmacm.so
       93977  nixl_cu12.libs/ucx/libuct_rdmacm.so.0
       93977  nixl_cu12.libs/ucx/libuct_rdmacm.so.0.0.0
     6552265  nixl_cu12.libs/ucx/libucx_perftest_cuda.so
     6552265  nixl_cu12.libs/ucx/libucx_perftest_cuda.so.0
     6552265  nixl_cu12.libs/ucx/libucx_perftest_cuda.so.0.0.0
       27169  nixl_cu12.libs/ucx/libucx_perftest_mad.so
       27169  nixl_cu12.libs/ucx/libucx_perftest_mad.so.0
       27169  nixl_cu12.libs/ucx/libucx_perftest_mad.so.0.0.0
(30 files)
```

### After

```
       60561  nixl_cu12.libs/ucx/libucm_cuda.so.0.0.0
       43081  nixl_cu12.libs/ucx/libuct_cma.so.0.0.0
      174465  nixl_cu12.libs/ucx/libuct_cuda.so.0.0.0
      365641  nixl_cu12.libs/ucx/libuct_ib.so.0.0.0
       81185  nixl_cu12.libs/ucx/libuct_ib_efa.so.0.0.0
      548761  nixl_cu12.libs/ucx/libuct_ib_mlx5.so.0.0.0
       76969  nixl_cu12.libs/ucx/libuct_ib_mlx5_gda.so.0.0.0
       98225  nixl_cu12.libs/ucx/libuct_rdmacm.so.0.0.0
     6552265  nixl_cu12.libs/ucx/libucx_perftest_cuda.so.0.0.0
       27169  nixl_cu12.libs/ucx/libucx_perftest_mad.so.0.0.0
(10 files)
```

## Summary by CodeRabbit

* **Improvements**
  * Improved deduplication of UCX/NIXL plugin shared libraries during wheel builds to consolidate identical binaries and reduce package size.
  * Updated plugin dependency references inside wheels to use canonical library filenames instead of symlink variants for more consistent runtime linkage.

* **Reliability**
  * Enhanced error handling during wheel repair: dependency rewrite failures now surface clear failures rather than silently continuing.
  * Removed the wheel-build option for plugin soname suffix handling; the repair step now uses a simplified, standardized invocation.


contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,65 @@ def get_lib_deps(lib_path):
...
+            real_src = os.path.realpath(os.path.join(root, kept))
+            dst_file = os.path.join(dst_dir, kept)
+            shutil.copy2(real_src, dst_file)
             copied_files.append(dst_file)
-    return copied_files
+            for removed in group[1:]:
+                dedup_map[removed] = kept
+                print(f"Deduplicated symlink: {removed} -> {kept}")
+
+    return copied_files, dedup_map
```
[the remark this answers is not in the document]
ovidiusm: This does not make any sense. Within a single directory, filenames are unique by definition. Across directories, the same symlink name (e.g. libuct_ib.so) maps to the same versioned name (e.g. libuct_ib.so.0.0.0), so overwrites are idempotent. There can be no conflict. Adding a dead assertion just to guard against something that can't happen adds noise.

contrib/wheel_add_ucx_plugins.py
```diff
@@ -261,6 +305,21 @@ def add_plugins(wheel_path, sys_plugins_dir, install_dirname):
...
+            if os.path.isfile(fname) and ".so" in fname:
+                for old_name, new_name in dedup_map.items():
+                    ret = os.system(
+                        f"patchelf --replace-needed '{old_name}' '{new_name}' {fname}"
+                    )
+                    if ret != 0:
+                        raise RuntimeError(
+                            f"Failed to replace {old_name} with {new_name} in {fname}"
+                        )
+
```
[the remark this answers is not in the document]
ovidiusm: The loop is intentionally separate since it is handling different situations, the logic should not be merged.

Plugin bundling is done by listing a directory that may contain symlinks. It needs special treatment.

Dependency bundling is done by following binary dependency chains, which do not seem impacted by symlinks.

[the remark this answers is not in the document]
ovidiusm: > Verification gap (concern #1). After the dedup --replace-needed rewrites, there is no dependency re-check, so a broken RPATH (where the kept real name isn't reachable) would only surface at install/import time. A minimal guard would be re-running get_lib_deps(fname) after the rewrites and checking for None paths — similar to what the first loop already does.

Since we only replace dependencies that we store in the map, which is created by the fs tree copy, I do not see the point of this check. Think about it, it would just duplicate the same code. There is no value in that.

The dependency check in the other place was done to make sure there were no missing dependencies that could have been caused by reasons outside our control, such as ld config or library path issues.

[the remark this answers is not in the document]
ovidiusm: > ".so" in fname on an absolute path (concern #3, Line 313). fname comes from copied_files, which holds absolute paths (e.g. /tmp/xyz/nixl.libs/ucx/libuct_ib.so.0.0.0). If any intermediate directory in the temp path happened to contain .so in its name, the check would spuriously pass. The safe fix is ".so" in os.path.basename(fname).

Theoretically possible, practically not. tempfile.mkdtemp() generates paths like /tmp/tmpXXXXXXXX not /tmp/tmpXXXXXXXX.so

contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,63 @@ def get_lib_deps(lib_path):
...
     """
     Copy a tree of files from @src directory to @dst directory.
-    Similar to shutil.copytree, but returns a list of all files copied.
+    Deduplicates shared-library files (``.so``) that share the same inode
+    (symlinks/hardlinks to the same underlying file).  For each group of
+    duplicate names only the real (non-symlink) file is kept -- typically the
+    fully-versioned name such as ``libuct_ib.so.0.0.0``.  This prevents the
+    dynamic linker from treating what used to be symlinks as separate
+    libraries (different inodes), which would cause components to be
+    initialised multiple times.  Non-``.so`` files are always copied as-is.
```
brminich: can you pls use 1 space after each `.`

contrib/wheel_add_ucx_plugins.py
```diff
@@ -261,6 +303,21 @@ def add_plugins(wheel_path, sys_plugins_dir, install_dirname):
...
 
+    # Replace inter-plugin DT_NEEDED entries that reference removed symlink
+    # names (e.g. libuct_ib.so.0) with the kept real file name
+    # (e.g. libuct_ib.so.0.0.0).
+    if dedup_map:
+        for fname in copied_files:
+            if os.path.isfile(fname) and ".so" in fname:
+                for old_name, new_name in dedup_map.items():
+                    ret = os.system(
+                        f"patchelf --replace-needed '{old_name}' '{new_name}' {fname}"
```
brminich: missed `''`?
```suggestion
                        f"patchelf --replace-needed '{old_name}' '{new_name}' '{fname}'"
```

contrib/wheel_add_ucx_plugins.py
```diff
@@ -261,6 +303,21 @@ def add_plugins(wheel_path, sys_plugins_dir, install_dirname):
                     ):
                         raise RuntimeError(f"Library {libname} not loaded by {fpath}")
 
+    # Replace inter-plugin DT_NEEDED entries that reference removed symlink
+    # names (e.g. libuct_ib.so.0) with the kept real file name
+    # (e.g. libuct_ib.so.0.0.0).
+    if dedup_map:
+        for fname in copied_files:
+            if os.path.isfile(fname) and ".so" in fname:
```
brminich: the guard filters to .so files, but it doesn't check whether old_name is actually in that file's DT_NEEDED. So patchelf is still invoked for every (fname, old_name) pair where fname happens to be a .so, even when fname has no dependency on old_name.
Whether that causes a problem depends entirely on patchelf version behavior. Modern patchelf (0.14+) exits 0 silently when the name isn't found — in which case this is just wasteful but harmless. Older patchelf exits non-zero, which would trigger the RuntimeError incorrectly. If the build environment guarantees a recent patchelf, this isn't a bug.

ovidiusm: We have patchelf 0.17 in base image. Also, I would prefer to leave the logic for patchelf if it is already there instead of us trying to reimplement it in the script

contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,63 @@ def get_lib_deps(lib_path):
...
+
+            try:
+                stat_info = os.stat(src_file)
+                key = (stat_info.st_dev, stat_info.st_ino)
+            except OSError:
+                dst_file = os.path.join(dst_dir, file)
+                shutil.copy2(src_file, dst_file)
+                copied_files.append(dst_file)
+                continue
+            inode_groups.setdefault(key, []).append(file)
```
guy-ealey-morag: Can you add comments here? 
I find it hard to follow what `os.stat` is doing and what are the return values, and what it means when there is an exception?

contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,63 @@ def get_lib_deps(lib_path):
...
+                  "libuct_ib.so.0": "libuct_ib.so.0.0.0"}``.
     """
     copied_files = []
+    dedup_map = {}
     for root, dirs, files in os.walk(src):
         rel_path = os.path.relpath(root, src)
         dst_dir = os.path.join(dst, rel_path)
         os.makedirs(dst_dir, exist_ok=True)
+
+        inode_groups = {}
```
guy-ealey-morag:
```suggestion
        inode_groups = collections.defaultdict(list)
```

Then appending to the list looks like:
`inode_groups[key].append(file)`

Also add:
`import collections`

contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,63 @@ def get_lib_deps(lib_path):
...
         for file in files:
```
guy-ealey-morag, line 212:
```python
        for file in files:
            src_file = os.path.join(root, file)

            if ".so" in file:
                try:
                    stat_info = os.stat(src_file)
                    key = (stat_info.st_dev, stat_info.st_ino)
                    inode_groups[key].append(file)
                    continue
                except OSError:
                    pass

            dst_file = os.path.join(dst_dir, file)
            shutil.copy2(src_file, dst_file)
            copied_files.append(dst_file)
```

Suggestion to reorder code to avoid repetition of the file copy

contrib/wheel_add_ucx_plugins.py
guy-ealey-morag:
```python
if not os.path.isfile(fname) or ".so" not in fname:
	continue
...
```

This allows for less nesting in the code.

contrib/wheel_add_ucx_plugins.py
```diff
@@ -261,6 +303,21 @@ def add_plugins(wheel_path, sys_plugins_dir, install_dirname):
...
+    # (e.g. libuct_ib.so.0.0.0).
+    if dedup_map:
+        for fname in copied_files:
+            if os.path.isfile(fname) and ".so" in fname:
+                for old_name, new_name in dedup_map.items():
+                    ret = os.system(
+                        f"patchelf --replace-needed '{old_name}' '{new_name}' '{fname}'"
+                    )
+                    if ret != 0:
+                        raise RuntimeError(
```
guy-ealey-morag: May be overkill, but did you consider capturing stdout/stderr to log them in case of failure?
Something like:
```python
result = subprocess.run(
                    ["patchelf", "--replace-needed", old_name, new_name, fname],
                    capture_output=True,
                    text=True,
                )
                if result.returncode != 0:
                    logger.error(
                        "patchelf --replace-needed %s %s %s failed (exit %d)\n"
                        "stdout: %s\nstderr: %s",
                        old_name,
                        new_name,
                        fname,
                        result.returncode,
                        result.stdout.strip(),
                        result.stderr.strip(),
                    )
                    raise RuntimeError(
                        f"Failed to replace {old_name} with {new_name} in {fname}"
                    )
```

contrib/wheel_add_ucx_plugins.py
```diff
@@ -163,21 +163,63 @@ def get_lib_deps(lib_path):
...
         dst_dir = os.path.join(dst, rel_path)
         os.makedirs(dst_dir, exist_ok=True)
+
+        inode_groups = {}
         for file in files:
             src_file = os.path.join(root, file)
-            dst_file = os.path.join(dst_dir, file)
-            shutil.copy2(src_file, dst_file)
+
+            if ".so" not in file:
```
guy-ealey-morag: I suggest making this check more robust, currently it would match files that are not intended like `so.suffix` or `.sol`. (may be unlikely but would be annoying to debug if it happens).

```python
import re
_SHARED_LIB_RE = re.compile(r"\.so(\.\d+)*$")
def _is_shared_lib(name: str) -> bool:
    return bool(_SHARED_LIB_RE.search(name))
```

another (simplified) check we can do is:
```python
file.endswith(".so") or ".so." in file
```

ovidiusm: Thanks, addressed as you suggested

contrib/wheel_add_ucx_plugins.py
ovidiusm: Thanks, addressed

contrib/wheel_add_ucx_plugins.py
ovidiusm: Done

contrib/wheel_add_ucx_plugins.py
ovidiusm: Done

contrib/wheel_add_ucx_plugins.py
ovidiusm: Makes sense. I reorganized the code

contrib/wheel_add_ucx_plugins.py
```diff
@@ -17,17 +17,28 @@
...
 import shutil
+import subprocess
 import tempfile
 import zipfile
 
 logger = logging.getLogger(__name__)
 
+# Matches shared-library names such as ``libfoo.so``, ``libfoo.so.0`` and
+# ``libfoo.so.0.0.0`` while rejecting unrelated names like ``foo.sol``.
+_SHARED_LIB_RE = re.compile(r"\.so(\.\d+)*$")
```
ColinNV, line 35: Should this also start with `"^lib"`?

ovidiusm: Not all the `so` files we package start with lib

contrib/wheel_add_ucx_plugins.py
ovidiusm: Reduced the nesting

contrib/wheel_add_ucx_plugins.py
```diff
@@ -231,55 +279,76 @@ def add_plugins(
...
                     )
-                    ret = os.system(
-                        f"patchelf --replace-needed {libname} {packaged_name} {fpath}"
+        # Check that there is no breakage introduced in the patched lib
+        logger.debug("Checking that %s loads", fpath)
+        original_deps = get_lib_deps(os.path.join(sys_plugins_dir, fname))
+        for libname, libpath in get_lib_deps(fpath).items():
+            if libpath is None:
+                if libname not in original_deps or original_deps[libname] is not None:
+                    raise RuntimeError(f"Library {libname} not loaded by {fpath}")
```
guy-ealey-morag, line 323: Shouldn't this check come after the dedup step?

Claude suggested replacing this check with the following loop placed before `create_wheel`:
```python
    # Check that no breakage was introduced once every lib is fully patched
    # (rpath + name_map replacements + dedup DT_NEEDED rewrites). This runs
    # last so it validates the final state, not an intermediate one where
    # deduplicated symlink names still dangle.
    for fname in copied_files:
        basename = os.path.basename(fname)
        if not os.path.isfile(fname) or not _is_shared_lib(basename):
            continue
        logger.debug("Checking that %s loads", fname)
        original_deps = get_lib_deps(os.path.join(sys_plugins_dir, basename))
        for libname, libpath in get_lib_deps(fname).items():
            if libpath is None and (
                libname not in original_deps or original_deps[libname] is not None
            ):
                raise RuntimeError(f"Library {libname} not loaded by {fname}")
```

ovidiusm: I think Claude does not understand the intent.

I added the check to this stage to make sure that we do not miss any dependencies in this specific stage, that come from extracting elf dependencies, which is a higher risk than resolving the symlinks. A few releases ago we had an issue in this area.

The check is not required nor essential to the functionality, but I would prefer to keep it here.

guy-ealey-morag (approved):

ColinNV (approved):

PR #1568, opened 2026-04-22
TEST/UCX: Make thread safe.

ColinNV:
## What?
Makes the multi-threaded `ucx_backend_multi` thread safe.

## Why?
It was using `volatile` for thread synchronization which is always wrong and possibly leading to spurious failures of some CI jobs.

## Summary by CodeRabbit

* **Tests**
  * Replaced unsafe busy-wait primitives with atomic synchronization for more reliable multi-threaded test runs.
  * Introduced configurable progress-driving mode and repeated iterations to better exercise thread interactions and timing.
  * Adjusted test lifecycle and failure checks for more consistent start/stop behavior and clearer failure reporting across runs.

### DIFF
--- a/test/unit/plugins/ucx/ucx_backend_multi.cpp
+++ b/test/unit/plugins/ucx/ucx_backend_multi.cpp
@@ -22,10 +22,6 @@
 #include "ucx_backend.h"
 #include "test_utils.h"
 
-// TODO: Find out what this comment means:
-// Temporarily while fixing CI/CD pipeline
-// #define USE_PTHREAD false
-
 std::atomic<bool> ready[2];
 std::atomic<bool> done[2];
 std::atomic<bool> disconnect[2];
--- a/test/unit/plugins/ucx/ucx_backend_multi.cpp
+++ b/test/unit/plugins/ucx/ucx_backend_multi.cpp
@@ -33,8 +33,7 @@ std::atomic<bool> disconnect[2];
 std::string conn_info[2];
 
 void
-test_thread(const unsigned id, const bool progress_thread)
-{
+test_thread(const unsigned id, const bool progress_thread) {
     const std::string my_name = "Agent" + std::to_string(id);
     const std::string other = "Agent" + std::to_string(1 - id);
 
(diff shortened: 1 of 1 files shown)


test/unit/plugins/ucx/ucx_backend_multi.cpp
```diff
@@ -14,88 +14,113 @@
...
-volatile bool disconnect[2]  = {false, false};
-std::string conn_info[2];
+// #define USE_PTHREAD false
 
-void test_thread(int id)
-{
-    nixlBackendInitParams init_params;
-    nixl_b_params_t       custom_params;
-    nixl_status_t         ret;
+std::atomic<bool> ready[2];
```
iyastreb, line 25: It seems NIXL will use C++20 (https://github.com/ai-dynamo/nixl/pull/1571)
So maybe just use std::barrier, at least for 2 stages?

ColinNV: Yes, but the switch to C++20 might take some time; I'd like to get this fix in as soon as possible.

test/unit/plugins/ucx/ucx_backend_multi.cpp
```diff
@@ -14,88 +14,113 @@
...
-
+// TODO: Find out what this comment means:
 // Temporarily while fixing CI/CD pipeline
-#define USE_PTHREAD false
-
-volatile bool ready[2]  = {false, false};
-volatile bool done[2]  = {false, false};
-volatile bool disconnect[2]  = {false, false};
-std::string conn_info[2];
+// #define USE_PTHREAD false
```
brminich: afai, it defines whether we will invoke ucx progress or not in the test thread (so probably it means use progress thread or not)

test/unit/plugins/ucx/ucx_backend_multi.cpp
```diff
@@ -14,88 +14,113 @@
...
 
     th1.join();
     th2.join();
+
+    std::cout << "Multithread test done\n";
+}
+
+int
+main() {
+    for (unsigned i = 0; i < 12; ++i) {
```
brminich, line 118: why 12?

test/unit/plugins/ucx/ucx_backend_multi.cpp
brminich: maybe remove alltogether?

test/unit/plugins/ucx/ucx_backend_multi.cpp
ColinNV: Random number to run it a few times without taking too much time.

test/unit/plugins/ucx/ucx_backend_multi.cpp
ColinNV: Yes. My question was wrt. the pipeline fixing.

ColinNV: Removed.

test/unit/plugins/ucx/ucx_backend_multi.cpp
```diff
@@ -14,88 +14,109 @@
...
+    disconnect[id].store(true);
     //wait for other
-    while(!disconnect[!id]);
+    while (!disconnect[!id].load())
+        ;
 
-    if (!USE_PTHREAD) ucx->progress();
+    if (!progress_thread) {
+        ucx->progress();
+    }
```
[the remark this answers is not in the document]
ColinNV: If this is really is an issue it's pre-existing.

brminich (approved):

PR #1573, opened 2026-04-23
PLUGINS/UCX: Remove indirection.

ColinNV:
## What?
Remove a level of pointer indirection in `nixlUcxPublicMetadata`.

## Why?
It was superfluous.

The vector in class `nixlUcxPublicMetadata` doesn't need to keep unique pointers, especially since `rkey` is nothing more than a unique pointer itself.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized and simplified connection metadata construction and initialization.
  * Switched to immutable metadata storage and removed runtime mutation of rkey collections.
  * Streamlined transfer request handling and tightened local variable constness.
  * Cleaned up internal includes and minor whitespace for maintainability.

### DIFF
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -78,22 +78,19 @@ class nixlUcxPrivateMetadata : public nixlBackendMD {
 // A public metadata has to implement put, and only has the remote metadata
 class nixlUcxPublicMetadata : public nixlBackendMD {
 public:
-    nixlUcxPublicMetadata() : nixlBackendMD(false) {}
+    nixlUcxPublicMetadata() = delete;
+    nixlUcxPublicMetadata(const ucx_connection_ptr_t &conn,
+                          std::vector<nixl::ucx::rkey> &&rkeys);
 
-    [[nodiscard]] ucp_rkey_h
-    rawRkey(const size_t id) const noexcept {
-        return rkeys_[id].get();
+    [[nodiscard]] const nixl::ucx::rkey &
+    getRkey(const size_t id) const {
+        return rkeys_[id];
     }
 
-    void
-    addRkey(const nixlUcxEp &ep, const void *rkey_buffer) {
-        rkeys_.emplace_back(ep, rkey_buffer);
-    }
-
-    ucx_connection_ptr_t conn;
+    const ucx_connection_ptr_t conn;
 
 private:
-    std::vector<nixl::ucx::rkey> rkeys_;
+    const std::vector<nixl::ucx::rkey> rkeys_;
 };
 
 class nixlUcxEngine : public nixlBackendEngine {
--- a/src/plugins/ucx/ucx_backend.h
+++ b/src/plugins/ucx/ucx_backend.h
@@ -81,19 +81,19 @@ public:
     nixlUcxPublicMetadata() : nixlBackendMD(false) {}
 
     [[nodiscard]] const nixl::ucx::rkey &
-    getRkey(size_t id) const {
-        return *rkeys_[id];
+    getRkey(const size_t id) const noexcept {
+        return rkeys_[id];
     }
 
     void
     addRkey(const nixlUcxEp &ep, const void *rkey_buffer) {
-        rkeys_.emplace_back(std::make_unique<nixl::ucx::rkey>(ep, rkey_buffer));
+        rkeys_.emplace_back(ep, rkey_buffer);
     }
 
     ucx_connection_ptr_t conn;
 
 private:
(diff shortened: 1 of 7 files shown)


src/plugins/ucx/ucx_backend.h
```diff
@@ -81,19 +81,19 @@ class nixlUcxPublicMetadata : public nixlBackendMD {
...
-        rkeys_.emplace_back(std::make_unique<nixl::ucx::rkey>(ep, rkey_buffer));
+        rkeys_.emplace_back(ep, rkey_buffer);
     }
 
     ucx_connection_ptr_t conn;
 
 private:
-    std::vector<std::unique_ptr<nixl::ucx::rkey>> rkeys_;
+    std::vector<nixl::ucx::rkey> rkeys_;
 };
```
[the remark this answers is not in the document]
ColinNV: The basic concern is valid, however since all callers of getRkey() only keep copies of getRkey(...).get() and not the result of getRkey() itself everything is safe. The interface can however be improved by making getRkey() call get() on what was its result, that's a much better "hardening" than what is suggested here.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -254,7 +254,7 @@ nixlUcxEp::sendAm(nixl::ucx::am_cb_op_t msg_id,
 
 nixl_status_t
 nixlUcxEp::read(uint64_t raddr,
-                const nixl::ucx::rkey &rkey,
+                const ucp_rkey_h rkey,
```
rakhmets: maybe without `const` to match the type of parameter of `ucp_get_nbx`?

src/plugins/ucx/ucx_backend.h
```diff
@@ -80,20 +80,20 @@ class nixlUcxPublicMetadata : public nixlBackendMD {
 public:
     nixlUcxPublicMetadata() : nixlBackendMD(false) {}
 
-    [[nodiscard]] const nixl::ucx::rkey &
-    getRkey(size_t id) const {
-        return *rkeys_[id];
+    [[nodiscard]] ucp_rkey_h
+    rawRkey(const size_t id) const noexcept {
```
rakhmets: I think `noexcept` can only be used when `id` is checked to be smaller than `rkeys_.size()`.

src/plugins/ucx/ucx_utils.cpp
rakhmets: `#include "rkey.h"` can be removed.

src/plugins/ucx/ucx_backend.h
ColinNV: The line below uses `operator[]`, not `at()`.

src/plugins/ucx/ucx_utils.cpp
ColinNV: This is just saying that `read()` won't modify `rkey`, the pointer.

src/plugins/ucx/ucx_backend.h
rakhmets: I'm not really sure if I like the new interface better than the old one.
Maybe it can be solved in a slightly different way:
replace `addRkey` with ctor, and move vector to the new ctor, and make the vector const.

rakhmets: `vector::operator[]` doesn't have `noexcept` specifier. Accessing a nonexistent element is undefined behavior.

ColinNV: Because `addRkey()` is only used shortly after construction in the same function? Right.

ColinNV: I don't see how the `noexcept` makes things worse, but I can remove it again.

rakhmets: Yes, it is.

src/plugins/ucx/ucx_backend.h
```diff
@@ -78,22 +78,18 @@ class nixlUcxPrivateMetadata : public nixlBackendMD {
...
-        rkeys_.emplace_back(std::make_unique<nixl::ucx::rkey>(ep, rkey_buffer));
-    }
-
-    ucx_connection_ptr_t conn;
+    const ucx_connection_ptr_t conn;
 
 private:
-    std::vector<std::unique_ptr<nixl::ucx::rkey>> rkeys_;
+    const std::vector<nixl::ucx::rkey> rkeys_;
 };
```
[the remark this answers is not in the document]
ColinNV: This took a different direction, now the vector `rkeys_` is `const` so there won't be any reallocations.

rakhmets (approved):

brminich (approved):

PR #1576, opened 2026-04-23
NIXL_EP: fallback to cudaMalloc if vmm allocation failed.

ofirfarjun7:
## What?
Fallback to cudaMalloc if allocation with vmm failed

## Why?
Allocation with VMM API can fail if server is not running Nvidia IMEX service.

## Summary by CodeRabbit

* **Refactor**
  * Optimized virtual memory management allocation strategy with improved fallback handling for fabric-backed memory operations.


examples/device/ep/csrc/vmm.cpp
```diff
@@ -139,35 +139,42 @@ vmm_region::vmm_region(size_t size) {
...
+            vmm_mapped_ = true;
 
-    if (cuMemAddressReserve(&ptr_, size_, 0, 0, 0) != CUDA_SUCCESS) {
-        release();
-        throw std::runtime_error("Failed to reserve CUDA virtual address");
-    }
+            if (cuMemSetAccess(ptr_, size_, &ctx.access_desc, 1) != CUDA_SUCCESS) {
+                release();
+                throw std::runtime_error("Failed to set CUDA memory access");
+            }
```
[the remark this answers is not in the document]
ofirfarjun7: We fallback here only if cuMemCreate failed because according to CUDA docs it needs at least one opned imex channel and we didn't check this requirement in the `cuda_alloc_ctx` creation. For the rest because we check requirements in ctx open, if they fail we don't want to fallback.

examples/device/ep/csrc/vmm.cpp
```diff
@@ -139,35 +139,42 @@ vmm_region::vmm_region(size_t size) {
...
-        size_ = size;
-        is_cuda_malloc_ = true;
-        if (cuMemAlloc(&ptr_, size) != CUDA_SUCCESS) {
-            throw std::runtime_error("cuMemAlloc fallback failed");
-        }
-        return;
-    }
-
-    size_ = nixl_ep::align_up<size_t>(size, ctx.granularity);
+    if (ctx.fabric_supported) {
```
rakhmets, line 142: Please consider adding a separate method. E.g.:
```C++
bool
vmm_region::tryVmmFabricAllocation(cuda_alloc_ctx &ctx, size_t size) {
    if (!ctx.fabric_supported) {
        return false;
    }
    ...
    if (mem_create_status != CUDA_SUCCESS) {
        ...
        return false;
    }
    ...
    return true;
}

vmm_region::vmm_region(size_t size) {
    ...
    if (tryVmmFabricAllocation(ctx, size)) {
        return;
    }
    ...
    if (cuMemAlloc(&ptr_, size) != CUDA_SUCCESS) {
    ...
}
```

examples/device/ep/csrc/vmm.cpp
```diff
@@ -139,35 +139,42 @@ vmm_region::vmm_region(size_t size) {
...
+        size_ = nixl_ep::align_up<size_t>(size, ctx.granularity);
+
+        const CUresult mem_create_status = cuMemCreate(&handle_, size_, &ctx.prop, 0);
+        if (mem_create_status != CUDA_SUCCESS) {
+            handle_ = 0;
+            warn_cu_api(
```
rakhmets: I'm not sure if fabric memory is unavailable on the user's setup, we should issue a message starting with the word `WARNING`.
We can issue the message starting with `INFO`, or `DEBUG`. Or even silently switch to using `cuMemAlloc`. As the only downside of this fallback is that MNNVL won't work.

ofirfarjun7: I case user expect mnvvl to be used and we just silently switch to cudaMalloc he might be confused.
I can change it to info/debug.

rakhmets: In the case when creating fabric memory is not possible on the target system, this will be a message that cannot be hidden.
I think we shouldn't print:
```
WARNING: vmm_region failed to cuMemCreate (falling back to cuMemAlloc; e.g. check IMEX service if applicable): ...
```
As it will be printed on all systems w/o MNNVL. And the suggestion in this case is not relevant.

rakhmets (approved):

examples/device/ep/csrc/vmm.cpp
```diff
@@ -74,100 +182,20 @@ vmm_region::vmm_region(size_t size) {
...
-        if (cuMemAlloc(&ptr_, size) != CUDA_SUCCESS) {
-            throw std::runtime_error("cuMemAlloc fallback failed");
-        }
+    if (vmmFabricAllocation(ctx, size)) {
         return;
+    } else if (ctx.fabric_supported) {
+        std::cerr << "INFO: " << k_vmm_ctx
+                      << " - CUDA VMM fabric allocation failed; falling back to "
+                         "cuMemAlloc\n";
+        ctx.fabric_supported = false;
```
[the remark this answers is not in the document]
rakhmets: In general, the comment is valid. However, this code is called only during an initialization stage in a single thread.

examples/device/ep/csrc/vmm.cpp
```diff
@@ -25,42 +25,151 @@ namespace {
...
+        if (cuMemGetAllocationGranularity(
+                &granularity, &prop, CU_MEM_ALLOC_GRANULARITY_MINIMUM) != CUDA_SUCCESS) {
+            throw std::runtime_error("Failed to get CUDA allocation granularity");
+        }
+
+        access_desc.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+        access_desc.location.id = device;
+        access_desc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+        fabric_supported = true;
+    }
```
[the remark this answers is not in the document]
rakhmets: that's a valid point. however, we only have single device in a process.

PR #1579, opened 2026-04-24
PLUGINS/UCX: Clean up nixlUcxBackendH.

ColinNV:
## What?
Clean up (mostly the interface of) class `nixlUcxBackendH`.

## How?
- Rename `nixlUcxBackendH` to `nixlUcxBackendReqH` to make the name reflect that it IS-A `nixlBackendReqH`.
- Remove the getter `notification()` and make the private member that it provided mutable access to `public`.
- Pass the `std::shared_ptr` to `append()` via const-reference instead of by value.
- Rename the private data members to align with the coding style.
- Update lines affected by the class rename to use `static_cast` instead of C-style cast.
- Update lines affected by previous changes to use correct name style.
- Sprinkle some `const`, `[[nodiscard]]` and `noexcept` where appropriate.
- Clean up occurrences of `auto` in `ucx_backend.cpp` following some review comments.

## Summary by CodeRabbit

* **Refactor**
  * Request-handle implementation reorganized for clearer ownership and safer usage.
  * Internal state and handle APIs standardized with stricter qualifiers for safer callers.
  * Notification flow simplified to a direct optional transfer state for simpler post/check semantics.
  * Call sites across engine and threadpool updated for consistent request-handle construction and usage.


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -51,38 +52,33 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+
     std::optional<Notif> notif;
 
-    nixl_status_t
+    [[nodiscard]] nixl_status_t
     checkConnection(nixl_status_t status = NIXL_SUCCESS) const {
         NIXL_ASSERT(!connections_.empty());
         for (const auto &conn : connections_) {
-            nixl_status_t conn_status = conn->getEp(worker_id)->checkTxState();
+            nixl_status_t conn_status = conn->getEp(workerId_)->checkTxState();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -35,13 +35,14 @@
  * Backend request management
 *****************************************/
 
-class nixlUcxBackendH : public nixlBackendReqH {
+class nixlUcxBackendReqH : public nixlBackendReqH {
```
rakhmets, line 38: I would even consider moving the class(es) to a separate file(s). Since this file is already 1.5k lines long.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -99,13 +95,13 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
-    const std::set<ucx_connection_ptr_t> &
-    getConnections() const {
+    [[nodiscard]] const std::set<ucx_connection_ptr_t> &
+    getConnections() const noexcept {
         return connections_;
     }
 
     virtual bool
-    isComposite() const {
+    isComposite() const noexcept {
```
rakhmets, line 112: nodiscard?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -408,9 +405,9 @@ struct nixlUcxBackendSharedState;
  * performed by a dedicated worker thread of threadpool. It holds a shared state
  * with the main request to track its completion status and control the lifetime.
  */
-class nixlUcxChunkBackendH : public nixlUcxBackendH {
+class nixlUcxChunkBackendH : public nixlUcxBackendReqH {
```
rakhmets: inherited classes should probably be renamed as well.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -522,14 +519,14 @@ class nixlUcxCompositeBackendH : public nixlUcxBackendH {
...
-    isComposite() const override {
+    isComposite() const noexcept override {
         return true;
     }
 
     nixl_status_t
     release() override {
         NIXL_TRACE << *this << " releasing";
-        nixl_status_t status = nixlUcxBackendH::release();
+        nixl_status_t status = nixlUcxBackendReqH::release();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -711,7 +708,7 @@ nixlUcxThreadPoolEngine::sendXferRange(const nixl_xfer_op_t &operation,
                                        nixlBackendReqH *handle,
                                        size_t start_idx,
                                        size_t end_idx) const {
-    nixlUcxBackendH *int_handle = static_cast<nixlUcxBackendH *>(handle);
+    nixlUcxBackendReqH *int_handle = static_cast<nixlUcxBackendReqH *>(handle);
```
rakhmets: you can use `auto` here as the type is mentioned on the right.
(the same for the other similar lines).

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1084,7 +1081,7 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
 
     const auto worker_id = getWorkerId(opt_args);
     /* TODO: try to get from a pool first */
-    auto *ucx_handle = new nixlUcxBackendH(getWorker(worker_id).get(), worker_id);
+    auto *ucx_handle = new nixlUcxBackendReqH(getWorker(worker_id).get(), worker_id);
```
rakhmets: `*` is redundant

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -51,38 +52,33 @@ class nixlUcxBackendH : public nixlBackendReqH {
             : agent(remote_agent),
               payload(msg) {}
     };
+
     std::optional<Notif> notif;
 
-    nixl_status_t
+    [[nodiscard]] nixl_status_t
     checkConnection(nixl_status_t status = NIXL_SUCCESS) const {
```
rakhmets: Please move this method back to a private scope.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -174,18 +170,19 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
+        NIXL_ASSERT(worker_ == nullptr || worker == nullptr);
+        worker_ = worker;
+        workerId_ = worker_id;
     }
 
-    nixlUcxWorker *
-    getWorker() const {
-        return worker;
+    [[nodiscard]] nixlUcxWorker *
+    getWorker() const noexcept {
```
rakhmets, line 179: It looks like setWorker and getWorker are internal stuff. Please consider place them into protected scope.

src/plugins/ucx/ucx_backend.cpp
rakhmets: probably, not in this pr...

ColinNV: Agree on making some files smaller; those PRs should do nothing else.

src/plugins/ucx/ucx_backend.cpp
ColinNV: Yes for `setWorker()`.

src/plugins/ucx/ucx_backend.cpp
ColinNV: I like to keep it in these cases so that it's immediately apparent to the reader that we're always dealing with a pointer.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -99,42 +103,42 @@ class nixlUcxBackendH : public nixlBackendReqH {
...
 
-    virtual bool
-    isComposite() const {
+    [[nodiscard]] virtual bool
+    isComposite() const noexcept {
         return false;
     }
 
-    virtual nixl_status_t
+    [[nodiscard]] virtual nixl_status_t
```
rakhmets: This one should return `void`.

src/plugins/ucx/ucx_backend.cpp
rakhmets: imo, having new on the same line is quite enough for this purpose.
and for me, there is also enough information here:
```C++
auto int_handle = static_cast<nixlUcxBackendReqH *>(handle);
```
, to clearly state that the type will be a pointer.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1133,7 +1131,8 @@ nixl_status_t nixlUcxEngine::estimateXferCost (const nixl_xfer_op_t &operation,
         std::chrono::microseconds msg_duration;
         std::chrono::microseconds msg_err_margin;
         nixl_cost_t msg_method;
-        nixl_status_t ret = rmd->conn->getEp(workerId)->estimateCost(lsize, msg_duration, msg_err_margin, msg_method);
+        nixl_status_t ret = rmd->conn->getEp(worker_id)->estimateCost(
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1163,16 +1162,16 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
         uint64_t raddr = static_cast<uint64_t>(remote[i].addr);
         NIXL_ASSERT(lsize == remote[i].len);
 
-        auto lmd = static_cast<nixlUcxPrivateMetadata *>(local[i].metadataP);
-        auto rmd = static_cast<nixlUcxPublicMetadata *>(remote[i].metadataP);
+        auto *lmd = static_cast<nixlUcxPrivateMetadata *>(local[i].metadataP);
+        auto *rmd = static_cast<nixlUcxPublicMetadata *>(remote[i].metadataP);
```
rakhmets: I don't think that these changes are necessary.

ColinNV: Ok.

src/plugins/ucx/ucx_backend.cpp
ColinNV: Not saying you're wrong, I just prefer the `*` in this case.

rakhmets: What confuses me the most is that this leads to ambiguity: in some cases we don't put `*`, in others we do. In my opinion, in the current state of the code, the option without `*` is more common. This is also confirmed by the changes in the lines where `nixlUcxPrivateMetadata*` and `nixlUcxPublicMetadata*` are used. Thus, I think the option without using `*` would be preferable. In addition to the reasons given in my previous comment, it is also an extra symbol that has no effect in this case.

rakhmets: I wouldn't mind if `auto` was followed by `*` everywhere in similar situations in current state of the code.

ColinNV: Ok, I've removed the asterisk wherever it was not needed, and cleaned up all occurrences of `auto` in `ucx_backend.cpp`.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -457,10 +454,10 @@ struct nixlUcxBackendSharedState {
 };
 
 void
-nixlUcxChunkBackendH::complete(nixl_status_t status) {
+nixlUcxChunkBackendReqH::complete(const nixl_status_t status) {
     NIXL_ASSERT(sharedState_.get() != nullptr);
     if (status != NIXL_SUCCESS) {
-        nixlUcxBackendH::release();
+        (void)nixlUcxBackendReqH::release();
```
```suggestion
        nixlUcxBackendReqH::release();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -470,11 +467,11 @@ nixlUcxChunkBackendH::complete(nixl_status_t status) {
...
 nixl_status_t
-nixlUcxChunkBackendH::status() {
+nixlUcxChunkBackendReqH::status() {
     // First check if entire request was cancelled or failed
     nixl_status_t status = sharedState_->status.load();
     if (status == NIXL_SUCCESS) {
-        status = nixlUcxBackendH::status();
+        status = nixlUcxBackendReqH::status();
     }
     return status;
```
rakhmets: optional:
```C++
    const nixl_status_t status = sharedState_->status.load();
    if (status == NIXL_SUCCESS) {
        return nixlUcxBackendReqH::status();
    }
    return status;
```
, or
```C++
    const nixl_status_t status = sharedState_->status.load();
    if (status != NIXL_SUCCESS) {
        return status;
    }
    return nixlUcxBackendReqH::status();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -514,42 +511,40 @@ class nixlUcxCompositeBackendH : public nixlUcxBackendH {
...
+    [[nodiscard]] nixl_status_t
     status() override {
         getWorker()->progressLoop();
 
         if (sharedState_->pendingReqs.load()) {
             return NIXL_IN_PROG;
         }
 
-        nixl_status_t status = nixlUcxBackendH::status();
+        nixl_status_t status = nixlUcxBackendReqH::status();
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1278,7 +1272,7 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
     if (opt_args && opt_args->hasNotif) {
         if (ret == NIXL_SUCCESS) {
             nixlUcxReq req;
-            auto rmd = (nixlUcxPublicMetadata *)remote[0].metadataP;
+            const auto rmd = (nixlUcxPublicMetadata *)remote[0].metadataP;
```
```suggestion
            const auto rmd = static_cast<nixlUcxPublicMetadata *>(remote[0].metadataP);
```

src/plugins/ucx/ucx_utils.h
rakhmets: Please remove.

rakhmets (dismissed):

rakhmets (dismissed):

rakhmets (dismissed):

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -523,13 +523,13 @@ nixlUcxWorker::epAddr() {
...
-absl::StatusOr<std::unique_ptr<nixlUcxEp>>
+std::unique_ptr<nixlUcxEp>
 nixlUcxWorker::connect(void *addr, std::size_t size) {
     try {
         return std::make_unique<nixlUcxEp>(worker.get(), addr, err_handling_mode_);
     }
-    catch (const std::exception &e) {
-        return absl::UnavailableError(e.what());
+    catch (const std::exception &) {
+        return {};
```
brminich, line 533: The exception message is silently dropped here. If `ucp_ep_create` fails (bad address, unreachable peer), the caller gets `nullptr` with no diagnostics. Consider adding a log line before returning:
```cpp
    NIXL_ERROR << "UCX connect failed: " << e.what();
    return {};
}
```

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -35,54 +35,58 @@
...
+        workerId_ = worker_id;
+    }
+
 public:
-    nixlUcxBackendH(nixlUcxWorker *worker, size_t worker_id)
-        : worker(worker),
-          worker_id(worker_id) {}
+    // Notification to be sent after completion of all requests
+    struct Notif {
+        std::string agent;
```
iyastreb: these fields may be const

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -35,54 +35,58 @@
...
+        nixl_blob_t payload;
 
-    auto &
-    notification() {
-        return notif;
-    }
+        Notif(const std::string &remote_agent, const nixl_blob_t &msg)
```
iyastreb, line 71: this ctor = default one, so redundant

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1306,9 +1300,9 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
 
 nixl_status_t nixlUcxEngine::checkXfer (nixlBackendReqH* handle) const
 {
-    nixlUcxBackendH *intHandle = (nixlUcxBackendH *)handle;
-    auto& notif = intHandle->notification();
-    nixl_status_t handle_status = intHandle->status();
+    const auto int_handle = static_cast<nixlUcxBackendReqH *>(handle);
+    auto &notif = int_handle->notif;
+    const nixl_status_t handle_status = int_handle->status();
```
iyastreb, line 1304: maybe change a bit the flow here:
```
const nixl_status_t handle_status = int_handle->status();
if (handle_status == IN_PROG) { return handle_status; }

auto notif = int_handle.release(); // EXTRACT notif from handle - for all cases below

File: src/plugins/ucx/ucx_backend.cpp (hunk 19 above)
Comment (ColinNV) [reply src/plugins/ucx/ucx_backend.cpp:71]: It's not exactly the same, the `emplace()` won't work without it.

File: src/plugins/ucx/ucx_backend.cpp (hunk 20 above)
Comment (ColinNV) [reply src/plugins/ucx/ucx_backend.cpp:1304]:
```
auto notif = int_handle.release(); // EXTRACT notif from handle - for all cases below
```

I don't understand this line, `int_handle` is a plain pointer.

Comment (iyastreb) [reply src/plugins/ucx/ucx_backend.cpp:1304]: I meant just move the ownership once (for all cases below except IN_PROG), smth like
```
auto notif = std::move(int_handle->notif.value());
int_handle->notif.reset();
```

Comment (ColinNV) [reply src/plugins/ucx/ucx_backend.cpp:1304]: Ok, got it, if you could double check.

File: src/plugins/ucx/ucx_backend.cpp [hunk 21]
```diff
@@ -1306,45 +1300,46 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
...
 
-    if ((handle_status != NIXL_SUCCESS) || !notif.has_value()) {
-        if (handle_status != NIXL_IN_PROG) { // error flow
-            notif.reset();
-        }
+    if ((handle_status == NIXL_IN_PROG) || !int_handle->notif) {
+        return handle_status;
+    }
 
+    if (handle_status != NIXL_SUCCESS) {
```
iyastreb: you can do this block after `int_handle->notif.reset();`
This is error flow, no need to optimize it

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1306,45 +1300,46 @@ nixlUcxEngine::postXfer(const nixl_xfer_op_t &operation,
...
+        int_handle->notif.reset();
         return handle_status;
     }
 
-    ucx_connection_ptr_t conn = getConnection(notif->agent);
+    const nixlUcxBackendReqH::Notif notif(std::move(int_handle->notif).value());
+    int_handle->notif.reset();
+
+    const ucx_connection_ptr_t conn = getConnection(notif.agent);
     if (!conn) {
```
iyastreb: This is a hot path, so I propose to add `unlikely` to all the error handling branches

iyastreb (approved):

rakhmets (approved):

brminich (approved):

PR #1584, opened 2026-04-25
nixl_ep: Preserve low-latency CUDA graphs across scale up/down

itayalroy:
Store the low-latency NIXL GPU context in stable device memory and pass that pointer to LL kernels. This keeps the kernel argument stable when replacing memory views on `connect_ranks`/`disconnect_ranks`.

Also add low-latency-only connect_ranks(..., activate=False), allowing new ranks to be connected while remaining masked until explicitly activated.

This is important for vLLM elastic EP and fault tolerance- elastic EP needs to re-use existing CUDA graphs and serve with the old group post connect_ranks until the new ranks are ready to serve, and FT needs to re-use existing CUDA graphs post rank recovery.

Validated no performance degradation for both NVLink and RDMA paths on 1-node and 2-node runs.

## Summary by CodeRabbit

* **Documentation**
  * Updated peer-connection docs to reflect new activation option and behavior in low-latency vs high-throughput modes.

* **New Features**
  * Added an optional activate parameter to control whether newly connected peers are immediately activated (defaults to true); in low-latency mode callers can defer activation, while deferred activation is disallowed in high-throughput mode.

### DIFF
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -74,7 +74,8 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
          int num_tokens, int num_max_dispatch_tokens_per_rank,
          int num_topk, int num_experts, int rank, int num_ranks,
          int num_warp_groups, int num_warps_per_group,
-         bool round_scale, uint64_t timeout_cycles, int phases, nixl_ep::gpu_nixl_ctx nixl_ctx) {
+         bool round_scale, uint64_t timeout_cycles, int phases, nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr) {
+    auto nixl_ctx = *nixl_ctx_ptr;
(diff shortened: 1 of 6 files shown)


jeynmann (approved):

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -617,7 +618,8 @@ combine(void* combined_x,
         int num_max_dispatch_tokens_per_rank,
         int num_experts, int rank, int num_ranks,
         int num_warp_groups, int num_warps_per_group,
-        uint64_t timeout_cycles, int phases, bool zero_copy, nixl_ep::gpu_nixl_ctx nixl_ctx) {
+        uint64_t timeout_cycles, int phases, bool zero_copy, nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr) {
```
ofirfarjun7, line 621: change to const in next PR

ofirfarjun7 (approved):

PR #1587, opened 2026-04-26
libfabric: Fix log messages for NUMA-aware rail selection policy

oa-aws:
Fix log messages for NUMA-aware rail selection policy when provider does not provide PCIe bus details.

## What?
Fix the info messages printed by NUMA-aware rail selection when libfabric provider does not give PCIe bus details for network devices.

## Why?
As of now, then log message being printed is somewhat misleading, stating that only one EFA device is present, and therefore default rail selection policy will be used. The policy selection is correct but from the wrong reasons. The correct reason is that there is PCIe information available for any device, and therefore NUMA aware rail selection should not be used (hence the selection of the default policy, which uses all rails).

## How?
fi_getinfo() result is checked for fi_info->nic->bus_attr to determine whether PCIe details are present for the used libfabric provider. If none found, then an appropriate info message is printed, something like: Using default (all) rail selection policy for DRAM memory type when provider is "tcp" and no PCIe bus data is available.

## Summary by CodeRabbit

* **Improvements**
  * Network rail selection now checks for PCIe NIC bus information and falls back to an all-rail DRAM policy when PCIe details are unavailable.
  * NUMA-aware rail selection is retained when PCIe information is present.
  * Logging and tracing improved to clarify selection paths and fallback reasons.

* **New Features**
  * Added a topology query to detect presence of PCIe NIC information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1587)


src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -195,6 +195,48 @@ nixlLibfabricTopology::isValidDevice(const std::string &efa_device) const {
...
+
+    fi_freeinfo(info);
+    fi_freeinfo(hints);
+    if (hasDevice) {
+        NIXL_TRACE << "Found at least one NIC with PCIe bus info";
+    } else {
+        NIXL_TRACE << "Did not find any NIC device with PCIe info";
+    }
+    return NIXL_SUCCESS;
+}
```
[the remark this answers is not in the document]
oa-aws: This refactoring is not planned for this PR.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -249,58 +249,84 @@ nixlLibfabricRailManager::init(const nixl_b_params_t &custom_params) {
...
+    // does. in order to avoid duplicate tests, we first check if the provider is not EFA
+    bool pcie_device_found = true;
+    if (topology->getProviderName().compare("efa") != 0) {
+        nixl_status_t res = topology->hasPCIeDevices(pcie_device_found);
+        if (res != NIXL_SUCCESS) {
+            NIXL_ERROR << "Failed to verify whether a device with PCIe bus information exists, "
+                          "when provider is "
+                       << topology->getProviderName();
+            return res;
+        }
```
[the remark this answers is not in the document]
oa-aws: Fixed

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -195,6 +195,48 @@ nixlLibfabricTopology::isValidDevice(const std::string &efa_device) const {
...
+    }
+
+    device_found = false;
+    for (struct fi_info *cur = info; cur; cur = cur->next) {
+        if (cur->domain_attr && cur->domain_attr->name && cur->nic && cur->nic->bus_attr) {
+            // this is enough, we have bus_attr
+            device_found = true;
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
oa-aws: Fixed

src/utils/libfabric/libfabric_topology.h
```diff
@@ -237,6 +237,15 @@ class nixlLibfabricTopology {
...
     static const uint16_t INVALID_NUMA_NODE_ID = UINT16_MAX;
 
+    /**
+     * @brief Queries whether there is any NIC with PCIe info.
+     * @param[out] device_found Upon seccussful execution, returns whether a device with PCIe bus
+     * details was found.
+     * @return The operation's result.
+     */
+    nixl_status_t
+    hasPCIeDevices(bool &device_found) const;
```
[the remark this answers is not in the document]
oa-aws: Fixed

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -195,6 +195,48 @@ nixlLibfabricTopology::isValidDevice(const std::string &efa_device) const {
     return std::find(all_devices.begin(), all_devices.end(), efa_device) != all_devices.end();
 }
 
+nixl_status_t
+nixlLibfabricTopology::hasPCIeDevices(bool &hasDevice) {
```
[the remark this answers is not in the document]
oa-aws: Already fixed

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -195,6 +195,50 @@ nixlLibfabricTopology::isValidDevice(const std::string &efa_device) const {
...
+    }
+
+    // Configure hints for the discovered provider
+    // This ensures consistency between device discovery and PCIe mapping
+    hints->fabric_attr->prov_name = strdup(provider_name.c_str());
+
+    int ret = fi_getinfo(FI_VERSION(1, 18), NULL, NULL, 0, hints, &info);
+    if (ret) {
+        NIXL_ERROR << "fi_getinfo failed for PCIe mapping with provider " << provider_name << ": "
+                   << fi_strerror(-ret);
```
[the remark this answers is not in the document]
oa-aws: Fixed

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -195,6 +195,49 @@ nixlLibfabricTopology::isValidDevice(const std::string &efa_device) const {
...
+    hints = fi_allocinfo();
+    if (!hints) {
+        NIXL_ERROR << "Failed to allocate fi_info for PCIe device search";
+        return NIXL_ERR_BACKEND;
+    }
+
+    // Configure hints for the discovered provider
+    hints->fabric_attr->prov_name = strdup(provider_name.c_str());
+
+    int ret = fi_getinfo(FI_VERSION(1, 18), NULL, NULL, 0, hints, &info);
```
yexiangd: can we not introduce another `fi_getinfo` ? in stead, in the construction of nixlLibfabricTopology, you can update and have a `has_pcie_devices_` bool member, , the n here, `hasPcieDevices()` only checks that value

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -249,58 +249,84 @@ nixlLibfabricRailManager::init(const nixl_b_params_t &custom_params) {
...
+        // because these are the only ones that matter for avoiding PCIe congestion (the instance
+        // types we deal with have either all rails connected to PCIe switches, or have only one EFA
+        // that is either not attached to a NUMA node, as in C-series, or not attached to a PCIe
+        // switch, as in G-series), so we test rail count against topology->getTotalNicCount()
+        // (which represents the total number of "usable" rails, i.e. that are attached to a PCIe
+        // switch) and not against topology->getAllDevices().size()
+        else if (max_rails < topology->getTotalNicCount()) {
+            // bandwidth does not exceed total machine capacity, so use NUMA-aware rail selection
+            // policy
+            NIXL_TRACE << "Using NUMA-aware rail selection policy for DRAM memory type";
```
yexiangd: NIXL_TRACE -> NIXL_INFO

src/utils/libfabric/libfabric_topology.cpp
oa-aws: The problem is that fi_getinfo() is called within buildPcieToLibfabricMapping() only when provider is efa, but I need to address cases where provider is not necessarily efa. Perhaps we can instead do a refactoring so there will be only one place in code where we call fi_getinfo (simple visitor interface of array returned from fi_getinfo(), maybe support lambda syntax for convenience).

yexiangd: I agree that a full refactoring isn't needed in this PR. But we can still avoid the extra `fi_getinfo` without a visitor pattern. I was referring to the simplest approach: call `buildPcieToLibfabricMapping()` for all providers (not just EFA), and have it set `has_pcie_devices_` as a side effect. Then `hasPcieDevices()` is just a getter — no new fi_getinfo at all. But up-to you if you want to address it in a fast follow-up

oa-aws: Ok, calling buildPcieToLibfabricMapping() for all providers and saving has_pcie_devices_ as a side effect.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -75,15 +76,21 @@ nixlLibfabricTopology::discoverTopology() {
...
-        if (status != NIXL_SUCCESS) {
+    // build the topology mapping regardless of provider
+    status = buildPcieToLibfabricMapping();
+    if (status != NIXL_SUCCESS) {
+        if (provider_name == "efa") {
             NIXL_ERROR << "Failed to build PCIe to Libfabric mapping - this is required for EFA "
                           "topology discovery";
-            return status;
+        } else {
+            NIXL_ERROR << "Failed to build PCIe to libfabric mapping (provider: " << provider_name
```
fengjica, line 86: From the code, it is also required for non-efa provider. Why differentiate the error message if the behavior is the same? Or should the return statement be moved inside the efa path?

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -249,58 +249,73 @@ nixlLibfabricRailManager::init(const nixl_b_params_t &custom_params) {
...
+    // when a provider does not provide PCIe data, there is no need to employ PCIe congestion
+    // avoidance mechanisms, which is one major part of what the NUMA-aware rail selection policy
+    // does. in order to avoid duplicate tests, we first check if the provider is not EFA
+    if (!topology->hasPcieDevices()) {
+        // either an error occurred or there is no PCIe bus data, in either case we use the
+        // default policy
+        NIXL_INFO << "Using default (all) rail selection policy for DRAM memory type when "
+                     "provider is \""
+                  << topology->getProviderName() << "\" and no PCIe bus data is available";
         dram_rail_selection_policy_ = std::make_unique<nixlLibfabricAllRailSelectionPolicy>();
```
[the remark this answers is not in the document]
fengjica: This comment from AI makes sense.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -75,15 +76,21 @@ nixlLibfabricTopology::discoverTopology() {
...
+                       << ")";
         }
+        return status;
+    }
+    // For EFA devices, build PCIe to Libfabric device mapping and full topology
+    if (provider_name == "efa") {
+        // Build PCIe to Libfabric device mapping
```
fengjica: Remove this comment line '// Build PCIe to Libfabric device mapping'?

src/utils/libfabric/libfabric_topology.cpp
oa-aws: Since the function now is being called also when provider is not necessarily EFA, I wanted to make sure I retain the same error message for the case of EFA. If that is not required I can use a single error message for all providers.

src/utils/libfabric/libfabric_topology.h
```diff
@@ -124,6 +124,7 @@ class nixlLibfabricTopology {
     NicInfoMap nic_info_map;
     size_t avg_nic_speed; // average NIC speed
     size_t avg_nic_upstream_speed; // average NIC upstream link speed
+    bool has_pcie_devices;
```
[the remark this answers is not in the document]
oa-aws: Fixing

src/utils/libfabric/libfabric_rail_manager.cpp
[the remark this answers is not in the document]
oa-aws: Fixing

src/utils/libfabric/libfabric_topology.cpp
oa-aws: Sure

src/utils/libfabric/libfabric_topology.cpp
fengjica: It appears that the behavior is also changed, for the non-efa code path. Before this PR, it will always go to the else part ("// For TCP/sockets devices, bypass complex topology discovery"). Now if buildPcieToLibfabricMapping is not successful, it will return (at line 89). Right?

fengjica (approved):

yexiangd (approved):

src/utils/libfabric/libfabric_topology.cpp
oa-aws: Reviewed and agreed to keep it this way

oa-aws:
Can someone (Nvidia/NIXL team) please share nixl-ci-dl-gpu failed test logs? Thanks.

PR #1588, opened 2026-04-27
fix(nixlbench): fix stats computation

oliviergaraud:
## What?

Fix incorrect iteration counts in printStats.

Throughput is computed using total iterations (all threads combined).
Latency used total iterations instead of per-thread iterations.

Since wall time covers all threads running in parallel, throughput needs the total count, while latency (per-thread sequential) needs the per-thread count.

## Release Notes

* **Bug Fixes**
  * Improved accuracy of benchmark statistics calculation by correctly distinguishing between total and per-thread iteration counts, ensuring latency measurements are properly computed across different block size scenarios.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -1122,10 +1122,12 @@ xferBenchUtils::printStats(bool is_target,
     double avg_latency = 0, throughput_gb = 0;
     double totalbw = 0;
 
-    int num_iter = xferBenchConfig::num_iter;
+    int total_iter = xferBenchConfig::num_iter;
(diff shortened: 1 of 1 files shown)


ofer (approved):

benchmark/nixlbench/src/utils/utils.cpp
aranadive, line 553: @oliviergaraud please add a check here that num_iter % num_threads should be 0.

oliviergaraud: @aranadive What behavior do you expect?
Doesn't the check on next lines make this condition always true?

aranadive (approved):

PR #1590, opened 2026-04-27
Nixlbench fix deallocate memory order

oliviergaraud:
## What?
Fix memory management issues in deallocateMemory/allocateMemory:

deallocateMemory deregistered IOVs in the wrong order:
remote IOVs (OBJ/BLK/FILE) must be deregistered before local IOVs,
and for OBJ the deregistration must happen before cleanup to avoid use-after-free.
+
memset in allocateMemory use iov.len instead of buffer_size.
This could be wrong if the actual allocation size differed, but the original was not a bug as in this loop the allocation was to buffer_size anyways.

## Summary by CodeRabbit

* **Bug Fixes**
  * Initialize each I/O buffer using its exact length to prevent partial writes and ensure correct validation.
  * Close backing file resources exactly once to avoid leaks.
  * Reorder deallocation so remote registrations are torn down before related local cleanup, reducing use-after-free risks.

* **Refactor**
  * Cleanup logic reorganized to enforce strict ordering and more robust resource tracking.


ofer (dismissed):

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -990,52 +990,53 @@ xferBenchNixlWorker::deallocateMemory(std::vector<std::vector<xferBenchIOV>> &io
...
+            case VRAM_SEG:
+                cleanupBasicDescVram(iov);
+                break;
+            default:
+                std::cerr << "Unsupported mem type: " << seg_type << std::endl;
+                exit(EXIT_FAILURE);
+            }
         }
     }
 }
```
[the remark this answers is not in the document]
aranadive: @oliviergaraud lets add the suggested comment here (a shortened version mentioning the order is fine).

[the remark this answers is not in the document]
oliviergaraud: @aranadive
done

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -992,53 +991,65 @@ xferBenchNixlWorker::deallocateMemory(std::vector<std::vector<xferBenchIOV>> &io
...
     } else if (xferBenchConfig::isStorageBackend()) {
         for (auto &iov_list : remote_iovs) {
-            for (auto &iov : iov_list) {
-                cleanupBasicDescFile(iov);
-            }
             nixl_reg_dlist_t desc_list(FILE_SEG);
             iovListToNixlRegDlist(iov_list, desc_list);
             CHECK_NIXL_ERROR(agent->deregisterMem(desc_list, &opt_args), "deregisterMem failed");
         }
+        // Close each backing fd exactly once, after all deregistrations complete.
```
iyastreb, line 1145: Maybe we can use RAII instead of calling cleanup/deregistration explictly?
This way it's automatically closed and destroyed in case of exceptions
It can be done similar to nixlAlloc

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -992,53 +991,65 @@ xferBenchNixlWorker::deallocateMemory(std::vector<std::vector<xferBenchIOV>> &io
...
+        for (auto &iov_list : remote_iovs) {
+            for (auto &iov : iov_list) {
+                if (closed_fds.insert(iov.devId).second) {
+                    cleanupBasicDescFile(iov);
+                }
+            }
+        }
+    }
+
+    for (auto &iov_list : iov_lists) {
```
iyastreb, line 1156: Can we use RAII here as well?

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
oliviergaraud: @iyastreb This is a more significant change.
I wish we can move forward with this fix and handle the RAII part in #1689

iyastreb: ok

iyastreb (approved):

ofer:
@oliviergaraud please rebase to pick up some CI fixes

oliviergaraud:
> @oliviergaraud please rebase to pick up some CI fixes

done

PR #1596, opened 2026-04-28
telemetry: fix prometheus metric registration

e-eygin:
CounterEntry and GaugeEntry destructors removed metrics from the shared Registry when aggregate-assignment temporaries were destroyed during initialization. Keep those entries as plain cached pointers, remove this agent's metric rows only from the exporter destructor, and add regression coverage for scrape visibility and counter updates.

## What?

## Why?

## How?
architecture, approach, etc._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Prometheus telemetry registration and cleanup to avoid duplicate/overwritten metrics and ensure stable removal on failure/shutdown.

* **Tests**
  * Added comprehensive Prometheus telemetry tests verifying metric families, exporter labels (agent, category, hostname), and cumulative counter accuracy under load.


ColinNV (commented):
How about instead:

- Change `counters_[name] = {&family, &metric};` and `gauges_[name] = {&family, &metric};` to use `try_emplace` instead of `operator[]` (giving the two structs an appropriate c'tor if needed).
- Make `CounterEntry` and `GaugeEntry` non-copyable and non-movable (unless move is needed, that could be done easily and cleanly).

This gets rid of the root cause, the superfluous copy?

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -154,7 +156,7 @@ nixlTelemetryPrometheusExporter::registerCounter(const std::string &name,
     auto &family = prometheus::BuildCounter().Name(name + "_total").Help(help).Register(*registry_);
     auto &metric =
         family.Add({{"category", category}, {"hostname", hostname_}, {"agent_name", agent_name_}});
-    counters_[name] = {&family, &metric};
+    counters_.try_emplace(name, &family, &metric);
```
ColinNV: Looks good, now perhaps a `NIXL_ASSERT` for debug builds to assert that a new entry was indeed created, which also documents the expectation that we never expect an overwrite.

ovidiusm (approved):

PR #1597, opened 2026-04-28
nixl: bulk descriptor removal and empty-section cleanup

benlwalker:
Two related cleanups to the descriptor removal paths in `nixl_memory_section`:

1. **`nixl: erase empty sections in removeLocalData`** — `removeLocalData` removed
   matching descriptors but left the empty `sectionMap` / `memToBackend` entries in
   place, so long-running register/deregister workloads grew these maps without
   bound. Mirrors the cleanup already done in `remDescList`.

2. **`nixl: bulk removal for descriptor lists`** — `remDescList` and
   `removeLocalData` previously called `target.remDesc(index)` once per
   descriptor, giving O(N*M) behavior when the backend held N entries and the
   caller was removing M of them. Adds `nixlSecDescList::remDescs(sorted_indices)`
   which compacts the vector in a single pass, mirroring the existing batched
   `addDescs()` on the add side.

## Summary by CodeRabbit

* **Refactor**
  * Optimized descriptor removal to perform bulk index-based removals and metadata deregistration in a single operation, replacing per-item loops to improve performance and reduce work when deleting multiple descriptors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1597?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/infra/mem_section.h
```diff
@@ -87,6 +87,9 @@ class nixlSecDescList : public nixlDescList<nixlSectionDesc> {
     void
     addDescs(nixlSecDescList &&other);
 
+    void
+    remDescs(std::vector<int> sorted_indices);
```
ColinNV: Looks like `remDescs()` doesn't mutate `sorted_indices`, should be passed by const-reference.

src/infra/nixl_descriptors.cpp
```diff
@@ -443,6 +443,26 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
...
+    NIXL_ASSERT(std::is_sorted(sorted_indices.begin(), sorted_indices.end()));
+
+    auto &vec = this->descs;
+    size_t ri = 0;
+    size_t write = 0;
+    for (size_t read = 0; read < vec.size(); ++read) {
+        if (ri < sorted_indices.size() && static_cast<int>(read) == sorted_indices[ri]) {
+            ++ri;
+        } else {
+            if (write != read) vec[write] = std::move(vec[read]);
```
ColinNV: Please use braces, and a separate line for the body.

src/infra/nixl_descriptors.cpp
```diff
@@ -443,6 +443,26 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
     addDescs(std::move(other.descs), order::SORTED);
 }
 
+void
+nixlSecDescList::remDescs(std::vector<int> sorted_indices) {
+    if (sorted_indices.empty()) return;
+
+    NIXL_ASSERT(std::is_sorted(sorted_indices.begin(), sorted_indices.end()));
+
+    auto &vec = this->descs;
```
ColinNV: Does it not compile when removing this and using `descs` instead of `vec`?

src/infra/nixl_memory_section.cpp
```diff
@@ -450,12 +452,22 @@ nixlRemoteSection::removeLocalData(const nixl_reg_dlist_t &mem_elms, nixlBackend
...
+        if (index >= 0) indices.push_back(index);
+    }
+
+    for (int idx : indices)
+        backend.unloadMD(target[idx].metadataP);
+
+    std::sort(indices.begin(), indices.end());
+    target.remDescs(std::move(indices));
+
+    if (target.isEmpty()) {
```
ColinNV, line 470: Aren't both of the data structures on which `erase()` is called limited in size by the number of memory types times the number of backends?

src/infra/nixl_descriptors.cpp
guy-ealey-morag: It's a pattern that is used extensively in this file, I think it's okay to keep it

src/infra/nixl_descriptors.cpp
```diff
@@ -443,6 +443,26 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
     addDescs(std::move(other.descs), order::SORTED);
 }
 
+void
+nixlSecDescList::remDescs(std::vector<int> sorted_indices) {
```
guy-ealey-morag:
```suggestion
nixlSecDescList::remDescs(std::vector<int> indices, order ord) {
    if (ord == order::SORTED) {
        NIXL_ASSERT(std::is_sorted(indices.begin(), indices.end()));
    } else {
        std::sort(indices.begin(), indices.end());
    }
```

I suggest making it similar to `addDescs`.
It's more robust and also moves the sorting to inside the function, making the callers cleaner

src/infra/nixl_memory_section.cpp
```diff
@@ -227,18 +227,20 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
...
-        int index = target.getIndex(elm);
-        // Already checked, elm should always be found. Can add a check in debug mode.
-        backend->deregisterMem(target[index].metadataP);
-        target.remDesc(index);
-    }
+    for (int idx : indices)
+        backend->deregisterMem(target[idx].metadataP);
+
+    std::sort(indices.begin(), indices.end());
+    target.remDescs(std::move(indices));
```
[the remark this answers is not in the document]
benlwalker: Done

src/infra/nixl_memory_section.cpp
```diff
@@ -450,12 +452,23 @@ nixlRemoteSection::removeLocalData(const nixl_reg_dlist_t &mem_elms, nixlBackend
...
-        }
+        if (index >= 0)
+            indices.push_back(index);
+    }
+
+    for (int idx : indices)
+        backend.unloadMD(target[idx].metadataP);
+
+    std::sort(indices.begin(), indices.end());
+    target.remDescs(std::move(indices));
```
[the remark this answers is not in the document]
benlwalker: Done

src/infra/mem_section.h
benlwalker: I went with Guy's suggestion to mirror addDescs instead, so remDescs now takes ownership of the index vector and sorts internally unless the caller passes order::SORTED.

src/infra/nixl_descriptors.cpp
benlwalker: Done

src/infra/nixl_descriptors.cpp
benlwalker: I have no preference but the rest of this file uses aliases like this.

src/infra/nixl_memory_section.cpp
benlwalker: The maps are bounded. I updated the commit message to avoid claiming unbounded growth. I kept the cleanup because this mirrors remDescList, and otherwise serializeSections still iterates the empty section and emits an empty record on each metadata exchange. If we should retain empty sections, that should be a separate change that updates both paths consistently.

src/infra/nixl_descriptors.cpp
benlwalker: Done

src/infra/nixl_memory_section.cpp
```diff
@@ -227,18 +227,22 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
...
     }
 
-    for (auto & elm : mem_elms) {
-        int index = target.getIndex(elm);
-        // Already checked, elm should always be found. Can add a check in debug mode.
-        backend->deregisterMem(target[index].metadataP);
-        target.remDesc(index);
-    }
+    std::sort(indices.begin(), indices.end());
+    indices.erase(std::unique(indices.begin(), indices.end()), indices.end());
```
guy-ealey-morag: Why is this step necessary? are we expecting duplicate indices?

guy-ealey-morag:
@benlwalker  It looks like some unrelated code was pushed to this branch

src/infra/nixl_descriptors.cpp
```diff
@@ -443,6 +443,32 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
...
+            ++ri;
+        } else {
+            if (write != read) {
+                vec[write] = std::move(vec[read]);
+            }
+            ++write;
+        }
+    }
+    vec.resize(write);
+}
```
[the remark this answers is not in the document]
benlwalker: `remDescs` is an internal helper on `nixlSecDescList` with two call sites, both in `nixl_memory_section.cpp`, that already filter invalid indices before calling. `addDescs` (the pattern this now mirrors) does no range validation either, so adding it here would be asymmetric within the same class. Out-of-range or duplicate entries are non-corrupting: they silently fall through the compaction loop. `remDesc` (singular) throws because it is a public template on `nixlDescList` exposed via the bindings; the audience and constraints differ.

src/infra/nixl_descriptors.cpp
```diff
@@ -443,6 +443,33 @@ nixlSecDescList::addDescs(nixlSecDescList &&other) {
     addDescs(std::move(other.descs), order::SORTED);
 }
 
+void
+nixlSecDescList::remDescs(std::vector<int> indices, order ord) {
```
iyastreb: In this particular case it's better to pass vector by const ref, because we never need to make a copy
So usage will be easier, no move needed
Also pls consider vector<size_t>, as it's more appropriate

benlwalker: Switched to `std::vector<size_t>`. Kept it by value rather than `const&` since the unsorted path sorts in place, and taking ownership lets the caller `std::move` its temporary indices vector in (both current call sites do).

iyastreb (dismissed):

guy-ealey-morag (dismissed):

brminich:
seems the issues are relevant
```
[2026-05-21T08:05:07.507Z] [52/111] UcxGdsLoadMultiplePluginInstantiation/LoadMultiplePluginsTestFixture.SimpleLifeCycleTest/0 (436 ms)
[2026-05-21T08:05:14.110Z] [53/111] ucx/TestErrorHandling.BasicXfer/0 (6084 ms)
[2026-05-21T08:05:14.110Z] [0;33mNote: Google Test filter = ucx/TestErrorHandling.BasicXfer/0
[2026-05-21T08:05:14.110Z] [m[0;32m[==========] [mRunning 1 test from 1 test suite.
[2026-05-21T08:05:14.110Z] [0;32m[----------] [mGlobal test environment set-up.
[2026-05-21T08:05:14.110Z] [0;32m[----------] [m1 test from ucx/TestErrorHandling
[2026-05-21T08:05:14.110Z] [0;32m[ RUN      ] [mucx/TestErrorHandling.BasicXfer/0
[2026-05-21T08:05:14.110Z] [53/111] ucx/TestErrorHandling.BasicXfer/0 killed by signal SIGSEGV (6084 ms)
[20
```
@benlwalker

benlwalker:
The crash is a double-free in `removeLocalData` triggered by a latent bug in `TestErrorHandling`: the fixture reuses the same `Agent` instances across `testXfer<WRITE>()` and `testXfer<READ>()`, and `MemDesc::init()` appends to `m_dlist` without clearing it first. The second `init()` therefore registers the same memory range twice on the fresh agent, and the matching `deregisterMem` calls into the new bulk path with two identical descriptors.

Pushed a third commit that clears `m_dlist` in `MemDesc::init()` so a re-init never re-registers the same memory. I also rebased the series onto current `main` to drop the merge commit.

iyastreb (approved):

brminich (approved):

PR #1600, opened 2026-04-29
AGENT: Move enum functions etc.

ColinNV:
## What?
- Move the enum-to-string functions from `nixl_agent.cpp` to `nixl_enum_strings.cpp`.
- Eliminate a constant in `nixl_agent.cpp` that was only used once.
- Put `illegal_plugin_combinations` in an anonymous namespace.
- Improve a `nixlAgent` exception message.
- Change all moved enum-to-string functions to provoke a compile-time error when the enums are extended (and no new case is added).

## Summary by CodeRabbit

* **Refactor**
  * Moved internal enum-to-string utilities into a dedicated module for clearer organization and maintainability.
* **Chore**
  * Build updated to include the new module so its utilities are compiled with the project.
* **Bug Fix**
  * Invalid-name error message clarified to explicitly state that names must be non-empty.

### DIFF
--- a/src/core/nixl_enum_strings.cpp
+++ b/src/core/nixl_enum_strings.cpp
@@ -24,9 +24,7 @@ namespace nixlEnumStrings {
 
 std::string
 memTypeStr(const nixl_mem_t &mem) {
-    static const std::array strings = {
-        "DRAM_SEG", "VRAM_SEG", "BLK_SEG", "OBJ_SEG", "FILE_SEG"
-    };
+    static const std::array strings = {"DRAM_SEG", "VRAM_SEG", "BLK_SEG", "OBJ_SEG", "FILE_SEG"};
 
     if ((mem < 0) || (mem >= strings.size())) {
         return "BAD_SEG";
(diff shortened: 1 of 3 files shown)


src/core/nixl_enum_strings.cpp
```diff
@@ -0,0 +1,81 @@
...
+        return "BAD_SEG";
+    }
+    return strings[mem];
+}
+
+std::string
+xferOpStr(const nixl_xfer_op_t &op) {
+    static const std::array strings = {"READ", "WRITE"};
+
+    if ((op < 0) || (op >= strings.size())) {
```
tomerg-nvidia: nit: if you are already not copying it as is, consider using a switch over the enum instead of having an array. I think it reads just as good as this strings array but also you don't have to maintain an array with the same order as in `nixl_xfer_op_t`. Same goes for `memTypeStr`

tomerg-nvidia (dismissed):

src/core/nixl_enum_strings.cpp
ColinNV: The main advantage is that the compiler will complain when a new enum value is added without a case.

tomerg-nvidia (dismissed):

guy-ealey-morag (approved):

tomerg-nvidia (approved):

brminich (approved):

PR #1601, opened 2026-04-29
fix(plugins): namespace meson variables to prevent cross-plugin leakage.

oliviergaraud:
## What?
Namespace all plugin-local meson variables (`plugin_deps`, `compile_defs`, `compile_flags`, `plugin_link_args`)
with a plugin-specific prefix across all plugin `meson.build` files following the pattern already established by `ucx` and partially by `libfabric`.

## Why?
Meson variables defined in one plugin's `meson.build` are visible to all subsequently-built plugins.
Generic variable names like `plugin_deps`, `compile_defs`, and `compile_flags` were reused across plugins,

causing two classes of bugs:           

1. **Silent cross-contamination:** When multiple plugins are enabled, later plugins inherit dependencies and compile definitions from earlier ones.
For example, enabling both POSIX and OBJ causes OBJ to silently pick up POSIX's `-DHAVE_LINUXAIO` flags and `file_utils_interface` dependency.
2. **Build failures in isolation:** Several plugins (OBJ, Azure Blob, GDS, GDS_MT, Gusli, HF3FS) reference `compile_defs` or `compile_flags` without ever defining them, relying on leakage from a previously-built plugin.                                                                                                                                                       
   
This bug has existed since the initial plugin commits (#479 for OBJ, #234 for POSIX) but was masked because POSIX always ran unconditionally.
Since #1086 introduced the `has_posix_plugin` guard, on systems where POSIX is disabled (either explicitely disabled or no `libaio`, `liburing`, or POSIX AIO), `plugin_deps` is never defined and downstream plugins fail with `Unknown variable "plugin_deps"`.

## How?
- Variables that are only used within a single plugin are prefixed with the plugin name (e.g. `posix_plugin_deps`, `obj_compile_defs`, `libfabric_compile_flags`).
- For plugins that never had compile defs/flags of their own (azure_blob, cuda_gds, gds_mt, gusli, hf3fs), the undefined variable references are replaced with empty lists.
- No functional change to any plugin's actual dependencies or compile flags, only scoping.  

## Summary by CodeRabbit

* **Chores**
  * Standardized per-plugin build configuration and renamed plugin-specific build variables for clearer isolation.
  * Static plugin builds no longer inherit shared build flags; shared builds consistently include position-independent settings.
  * Updated several plugin build declarations and expanded SPDX year ranges.
  * Core dependency declaration now propagates downstream library dependencies.

* **Tests**
  * Some plugin test targets no longer propagate the broader plugin dependency list.

### DIFF
--- a/src/plugins/gusli/meson.build
+++ b/src/plugins/gusli/meson.build
@@ -24,14 +24,14 @@ if 'GUSLI' in static_plugins
         dependencies: gusli_deps,
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: false,
-        cpp_args: compile_defs + compile_flags,
+        cpp_args: [],
         name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
 else
   gusli_backend_lib = shared_library('GUSLI', cpp_sources,
         dependencies: gusli_deps,
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: true,
-        cpp_args: compile_defs + ['-fPIC'],
+        cpp_args: ['-fPIC'],
         name_prefix: 'libplugin_',  # Custom prefix for plugin libraries
         install_dir: plugin_install_dir,
         install_rpath: '$ORIGIN/..')
--- a/src/plugins/obj/meson.build
+++ b/src/plugins/obj/meson.build
@@ -37,8 +37,6 @@ if not is_variable('obj_utils_interface')
 endif
 
 obj_plugin_deps = [nixl_infra, nixl_common_dep, obj_utils_interface, dependency('asio', required: true)]
-obj_compile_defs = []
-obj_compile_flags = []
 
 if cuobj_dep.found()
     message('Found CUObjClient Library. Enabling S3 Accelerated engines')
@@ -62,7 +60,7 @@ if 'OBJ' in static_plugins
         'OBJ',
         obj_sources,
         dependencies: obj_plugin_deps,
-        cpp_args: obj_compile_defs + obj_compile_flags,
+        cpp_args: [],
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: false,
         name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
@@ -71,7 +69,7 @@ else
         'OBJ',
         obj_sources,
         dependencies: obj_plugin_deps,
-        cpp_args: obj_compile_defs + obj_compile_flags + ['-fPIC'],
+        cpp_args: ['-fPIC'],
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: true,
         name_prefix: 'libplugin_',  # Custom prefix for plugin libraries
--- a/src/plugins/obj/meson.build
+++ b/src/plugins/obj/meson.build
@@ -36,9 +36,9 @@ if not is_variable('obj_utils_interface')
     subdir_done()
 endif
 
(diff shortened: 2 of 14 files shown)


ofer (dismissed):

lluki:
Could we also unify `defs` and `flags` while we are at it ? We could move to `flags` only: `obj_compile_defs` seems unused, and POSIX could move to `_flags` trivially

src/plugins/uccl/meson.build
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 cpp = meson.get_compiler('cpp')
-compile_flags = []
+uccl_compile_flags = []
```
aranadive: Lets clean up this vars too if not required by default.

src/plugins/uccl/meson.build
```diff
@@ -23,15 +23,15 @@ if 'UCCL' in static_plugins
                dependencies: [nixl_infra, serdes_interface, cuda_dep, libuccl_p2p, nixl_common_dep],
                include_directories: nixl_inc_dirs,
                install: false,
-               cpp_args : compile_flags,
+               cpp_args : uccl_compile_flags,
```
aranadive: Remove for simplicity

src/plugins/uccl/meson.build
```diff
@@ -23,15 +23,15 @@ if 'UCCL' in static_plugins
...
                name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
 else
     uccl_backend_lib = shared_library('UCCL',
                'uccl_backend.cpp', 'uccl_backend.h', 'uccl_plugin.cpp',
                dependencies: [nixl_infra, serdes_interface, cuda_dep, libuccl_p2p, nixl_common_dep],
                include_directories: [nixl_inc_dirs, utils_inc_dirs],
                install: true,
-               cpp_args : compile_flags + ['-fPIC'],
+               cpp_args : uccl_compile_flags + ['-fPIC'],
```
aranadive: Just use the flag directly

src/plugins/obj/meson.build
```diff
@@ -36,9 +36,9 @@ if not is_variable('obj_utils_interface')
     subdir_done()
 endif
 
-plugin_deps = [nixl_infra, nixl_common_dep, obj_utils_interface, dependency('asio', required: true)]
-compile_defs = []
-compile_flags = []
+obj_plugin_deps = [nixl_infra, nixl_common_dep, obj_utils_interface, dependency('asio', required: true)]
+obj_compile_defs = []
+obj_compile_flags = []
```
aranadive: These two can be removed.

src/plugins/gusli/meson.build
```diff
@@ -24,14 +24,14 @@ if 'GUSLI' in static_plugins
         dependencies: gusli_deps,
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: false,
-        cpp_args: compile_defs + compile_flags,
+        cpp_args: [],
```
aranadive, line 27: Lets make things consistent across plugins then if namespacing. Either keep flags var around for all plugins or remove if they are empty. My suggestion is to remove and use flags directly if by default they are empty.

oliviergaraud: done

oliviergaraud:
Hi @ofer 
thanks for keeping an eye on my PRs is there something I can do to help with the failing tests ?

PR #1602, opened 2026-04-29
fix(nixlbench): fix double evaluation in CHECK_NIXL_ERROR macro

oliviergaraud:
## What?
Fix the `CHECK_NIXL_ERROR` macro in nixlbench to evaluate its `result` argument only once.

## Why?
The macro expanded its result parameter twice: once in the condition and once in the error message.

When the argument is a function call (e.g. `agent->registerMem()`), the function executes twice on failure.
This can cause double side effects (double registration, double memory allocation, etc.) and the printed error code may differ from the one that actually triggered the error path.

## How?
Capture the `result` argument into a local `nixl_status_t` variable before testing and printing, following the standard C macro hygiene pattern. 

## Summary by CodeRabbit

* **Chores**
  * Improved internal error reporting so error messages display descriptive status names instead of numeric codes.
  * No change to functionality or behavior; error handling outcomes remain the same.

### DIFF
--- a/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
+++ b/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
@@ -40,12 +40,14 @@
 #define ROUND_UP(value, granularity) \
     ((((value) + (granularity) - 1) / (granularity)) * (granularity))
 
-#define CHECK_NIXL_ERROR(result, message)                                                       \
-    do {                                                                                        \
-        if (0 != result) {                                                                      \
-            std::cerr << "NIXL: " << message << " (Error code: " << result << ")" << std::endl; \
-            exit(EXIT_FAILURE);                                                                 \
-        }                                                                                       \
+#define CHECK_NIXL_ERROR(result, message)                                                     \
+    do {                                                                                      \
+        const nixl_status_t _r = (result);                                                    \
+        if (0 != _r) {                                                                        \
+            std::cerr << "NIXL: " << message << " (" << nixlEnumStrings::statusStr(_r) << ")" \
+                      << std::endl;                                                           \
+            exit(EXIT_FAILURE);                                                               \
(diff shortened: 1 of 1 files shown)


benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -40,12 +40,13 @@
...
 
-#define CHECK_NIXL_ERROR(result, message)                                                       \
-    do {                                                                                        \
-        if (0 != result) {                                                                      \
-            std::cerr << "NIXL: " << message << " (Error code: " << result << ")" << std::endl; \
-            exit(EXIT_FAILURE);                                                                 \
-        }                                                                                       \
+#define CHECK_NIXL_ERROR(result, message)                                                   \
+    do {                                                                                    \
+        nixl_status_t _r = (result);                                                        \
```
guy-ealey-morag:
```suggestion
        const nixl_status_t _r = (result);                                                  \
```

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -40,12 +40,13 @@
...
+        if (0 != _r) {                                                                      \
+            std::cerr << "NIXL: " << message << " (Error code: " << _r << ")" << std::endl; \
```
guy-ealey-morag:
```suggestion
            std::cerr << "NIXL: " << message << " (" << nixlEnumStrings::statusStr(_r) << ")" << std::endl; \
```

Small upgrade, we could print the error name instead of just a numeric value

oliviergaraud: done

guy-ealey-morag (approved):

brminich (approved):

PR #1581, opened 2026-04-24
Add LIBFABRIC postXfer thread pool

aknvda:
## Summary
- add a configurable LIBFABRIC descriptor-posting thread pool for `postXfer()`
- consume `num_threads` for LIBFABRIC from backend options, Python config, and nixlbench
- advertise LIBFABRIC post-thread options and document defaults

Fixes #1514

## Validation
- `git diff --check`
- `python3 -m py_compile src/api/python/_api.py`
- `ninja -C build-nolibfabric`
- `meson test -C build-nolibfabric`
- Docker LIBFABRIC-only build with `libfabric-dev 1.17.0`, `libhwloc-dev`, and `libnuma-dev`: `meson setup build-libfabric -Dbuildtype=debug -Dbuild_tests=false -Dbuild_examples=false -Denable_plugins=LIBFABRIC -Dinstall_headers=false && ninja -C build-libfabric`
- AWS/EFA e2e validation on two separate GB200/EFA nodes:
  - built libfabric 1.21.0 with the EFA provider enabled
  - built NIXL from this PR with `-Denable_plugins=LIBFABRIC`
  - ran `nixlbench --backend=LIBFABRIC` with DRAM WRITE, `--check_consistency`, `--progress_threads=4`, block size 4096, batch size 2048
  - cross-node run completed successfully on both ranks: `AWS/EFA cross-node LIBFABRIC nixlbench validation passed`

## Notes
- LIBFABRIC-enabled compile was validated in Docker.
- A Docker GTest compile also verified `test/gtest/plugin_manager.cpp`; full UCX+GTest build stopped later on Debian UCX 1.13 missing newer UCX symbols used by this repo.
- POSIX fixes were moved out of this PR and are tracked separately in #1605.

## Summary by CodeRabbit

* **New Features**
  * Optional parallel transfer posting for Libfabric with configurable thread pool and batch-size threshold; CUDA context access synchronized.

* **Bug Fixes**
  * Backend initialization now consistently forwards thread-count settings to Libfabric and ensures thread-pool teardown.

* **Documentation**
  * Runtime docs updated with new Libfabric controls: num_threads and split_batch_size.

* **Tests**
  * Added/expanded tests for Libfabric and POSIX backend options and lifecycle.

PR #1605, opened 2026-05-01
Fix POSIX backend queue fallback handling

aknvda:
## What
- Guard POSIX postXfer against an unavailable I/O queue.
- Treat POSIX backend initialization as failed when the selected queue type cannot create a queue.
- Let POSIX tests use the backend compiled default queue instead of forcing Linux AIO.
- Make POSIX request-handle release null-safe and delete the allocated handle.

## Why
Some build environments have POSIX AIO available but not Linux AIO/liburing. The existing test path forced Linux AIO and could hit a null queue, causing a SIGSEGV in nixlPosixBackendReqH::postXfer.

## Testing
- git diff --check
- meson setup --wipe build-posix -Dbuildtype=debug -Dbuild_tests=true -Dbuild_examples=false -Denable_plugins=POSIX
- ninja -C build-posix
- meson test -C build-posix nixl:posix_plugin_test --print-errorlogs

Split out from the LIBFABRIC postXfer thread-pool PR (#1581) so the POSIX fix can be reviewed independently: https://github.com/ai-dynamo/nixl/pull/1581

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved POSIX backend initialization with additional checks, clearer error logging, and fail-fast behavior when required I/O queues are unavailable.
  * Safer request cleanup with early-return on missing handles to prevent invalid operations.

* **Tests**
  * POSIX tests now default to the compiled queue; non-io_uring paths report "default" instead of "AIO" and show clearer guidance for explicit io_uring.
  * Added compile-gated helpers to validate supported POSIX queue options.

* **Chores**
  * Test build passes conditional compile-time flags for POSIX test variants.
* **Chores**
  * CI test matrix updated to target a new Slurm head node for job allocation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1605?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/plugins/posix/posix_backend.cpp
```diff
@@ -235,9 +240,10 @@ nixlPosixEngine::nixlPosixEngine(const nixlBackendInitParams *init_params)
...
                                               getKernelQueueSize(init_params->customParams))),
       io_queue_lock_(init_params->syncMode) {
-    if (io_queue_type_.empty()) {
+    if (io_queue_type_.empty() || !io_queue_) {
         initErr = true;
-        NIXL_ERROR << "Failed to initialize POSIX backend - no supported io queue type found";
+        NIXL_ERROR << "Failed to initialize POSIX backend - unsupported or unavailable io queue "
+                   << "type requested: " << io_queue_type_;
         return;
     }
```
guy-ealey-morag: I suggest splitting this to two checks with different error logs to distinct between the two issues.

src/plugins/posix/posix_backend.cpp
```diff
@@ -332,9 +338,13 @@ nixlPosixEngine::checkXfer(nixlBackendReqH *handle) const {
...
 nixl_status_t
 nixlPosixEngine::releaseReqH(nixlBackendReqH *handle) const {
+    if (!handle) {
+        return NIXL_SUCCESS;
+    }
+
     try {
         auto &posix_handle = castPosixHandle(handle);
-        posix_handle.~nixlPosixBackendReqH();
+        delete &posix_handle;
```
guy-ealey-morag:
```suggestion
        (void)castPosixHandle(handle); // Runtime validation
        delete handle;
```

The casting is not required for `delete`, so this change better shows that `castPosixHandle` is used for runtime validation

src/plugins/posix/posix_backend.cpp
aknvda: Fixed in 77635d6: POSIX backend initialization now checks `io_queue_type_.empty()` and `!io_queue_` separately, with distinct error messages for no supported queue type vs. an unavailable requested queue type.

src/plugins/posix/posix_backend.cpp
aknvda: Fixed in 77635d6: `releaseReqH()` now uses `(void)castPosixHandle(handle)` only for runtime validation, then deletes the base `nixlBackendReqH*` directly. This relies on the virtual base destructor and makes the validation intent explicit.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -217,8 +217,7 @@ read_write_test (int num_transfers,
         params["use_uring"] = "true";
         params["use_aio"] = "false";
     } else {
-        // Explicitly request AIO
-        params["use_aio"] = "true";
+        // Use the backend's compiled default queue. Some builds have POSIX AIO but not Linux AIO.
```
aranadive: Which builds are these? Our plan is to deprecate POSIX AIO since its sub-optimal compared to both Linux AIO / URING.

aknvda: This came from a minimal Ubuntu 22.04 POSIX-only build where Linux AIO and liburing were not installed, but POSIX AIO was available through libc/librt. In that setup, the POSIX plugin can still build using the POSIX AIO fallback but the test was forcing Linux AIO and hit an unavailable queue path. I am not trying to promote POSIX AIO here. This just keeps the current fallback safe while it is still supported. If POSIX AIO is being deprecated then I can instead update the test to skip/fail clearly when neither Linux AIO nor io_uring is available.

aranadive: @aknvda lets update this to fail clearly on start rather than a backend error.

aknvda: Updated in 9741147: the POSIX test now gets the compiled queue support flags from Meson and fails immediately at startup with a clear message if only POSIX AIO is available, or if io_uring is requested without liburing. I also updated the help/comments so POSIX AIO is not presented as the normal test fallback.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -60,6 +60,37 @@ namespace {
...
+    }
+
+    void
+    print_unsupported_test_queue_error(bool use_uring) {
+        std::cerr << "Unsupported POSIX test queue configuration: ";
+        if (use_uring) {
+            std::cerr << "io_uring was requested, but this build does not include liburing support."
+                      << std::endl;
+        } else {
+            std::cerr << "this build does not include Linux AIO or io_uring support." << std::endl;
```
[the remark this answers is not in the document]
aknvda: Fixed in 87b9a84d: applied clang-format 19 to the changed C++ lines and verified locally with the same clang-format-diff command used by CI. The GitHub clang-format check is now passing.

ofer (dismissed):

aranadive (dismissed):

src/plugins/posix/posix_backend.cpp
```diff
@@ -201,6 +201,11 @@ nixlPosixBackendReqH::checkXfer() {
 
 nixl_status_t
 nixlPosixBackendReqH::postXfer() {
+    if (!io_queue_) {
```
iyastreb: since it's hot path, maybe add unlikely

src/plugins/posix/posix_backend.cpp
```diff
@@ -332,9 +343,13 @@ nixlPosixEngine::checkXfer(nixlBackendReqH *handle) const {
...
 nixl_status_t
 nixlPosixEngine::releaseReqH(nixlBackendReqH *handle) const {
+    if (!handle) {
+        return NIXL_SUCCESS;
+    }
+
     try {
-        auto &posix_handle = castPosixHandle(handle);
-        posix_handle.~nixlPosixBackendReqH();
+        (void)castPosixHandle(handle);
```
iyastreb: btw dynamic_cast relies on RTTI and is quite heavy call, and I see it's called on hot path in 3 places
I'm not sure that castPosixHandle is needed at all, because it's a responsibility of nixlAgent to keep the correct handle mapping. Other backends don't do such checks too

src/plugins/posix/posix_backend.cpp
aknvda: Fixed in 1256f2b0: the defensive missing-queue guard now uses `__builtin_expect(!io_queue_, 0)`.

src/plugins/posix/posix_backend.cpp
aknvda: Fixed in 1256f2b0: POSIX request-handle paths no longer use RTTI. `castPosixHandle()` now does a null check plus `static_cast`, and `releaseReqH()` deletes the base handle directly after the null check.

src/plugins/posix/posix_backend.cpp
```diff
@@ -69,7 +69,7 @@ castPosixHandle(nixlBackendReqH *handle) {
     if (!handle) {
         throw nixlPosixBackendReqH::exception("received null handle", NIXL_ERR_INVALID_PARAM);
     }
-    return dynamic_cast<nixlPosixBackendReqH &>(*handle);
+    return *static_cast<nixlPosixBackendReqH *>(handle);
```
iyastreb:
```suggestion
    return static_cast<nixlPosixBackendReqH &>(*handle);
```

src/plugins/posix/posix_backend.cpp
```diff
@@ -332,16 +343,12 @@ nixlPosixEngine::checkXfer(nixlBackendReqH *handle) const {
 
 nixl_status_t
 nixlPosixEngine::releaseReqH(nixlBackendReqH *handle) const {
-    try {
-        auto &posix_handle = castPosixHandle(handle);
-        posix_handle.~nixlPosixBackendReqH();
+    if (!handle) {
```
iyastreb: NIXL_ASSERT(handle != nullptr);

src/plugins/posix/posix_backend.cpp
aknvda: Fixed in 85333903: changed `castPosixHandle()` to cast the already-checked base reference with `static_cast<nixlPosixBackendReqH &>(*handle)`.

iyastreb (dismissed):

aranadive (dismissed):

dpressle:
@ci-triage-agent

dpressle:
👀 Investigating [`nixl-ci-dl-gpu` build 709](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-dl-gpu/detail/nixl-ci-dl-gpu/709/pipeline/)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-dl-gpu` · commit `9bce96dc`

I now have a comprehensive picture of the failure. Here is the full diagnosis:

---

**Summary:** `TestTransfer.remoteMDFromSocket` fails across all 4 UCX test variants (ucx, ucx_no_pt, ucx_threadpool, ucx_threadpool_no_pt) with exit code 42 — CUDA DMA-buf registration via `ibv_reg_dmabuf_mr` on `mlx5_7` returns error 524 (`ENOTRECOVERABLE`), which cascades into `registerMem` failure → remote MD not found → transfer and notification assertions all fail.

---

**Root cause:**

The failure is a **hardware/driver-level DMA-buf registration error on `mlx5_7`** specific to the test node `gb200-nvl4-ts2-89`. The exact UCX error is:

```
ibv_reg_dmabuf_mr(address=0xfcc99fa00000, length=16384, ...) failed: Unknown error 524
UCX ERROR failed to register address ... (cuda) length 16384 dmabuf_fd 518 on md[7]=mlx5_7: Input/output error
```

Error 524 is `ENOTRECOVERABLE` — a Linux kernel errno meaning the mutex/state is irrecoverably broken, which in the context of `ibv_reg_dmabuf_mr` indicates the DMA-buf file descriptor is stale, the CUDA DMA-buf export capability is broken, or the `mlx5_7` port has a bad peer-memory/DMA-buf mapping on this particular GB200-NVL4 node.

Because `ucp_mem_map` fails, `nixlAgent::registerMem` returns `NIXL_ERR_BACKEND`. The initiator agent (`agent_0`) publishes no valid remote-key metadata through the socket listener. When the responder (`agent_1`) calls `loadRemoteMD` via the socket, it receives metadata with no registered memory, and returns `NIXL_ERR_NOT_FOUND`. This causes the three cascading gtest assertion failures:
- `test_transfer.cpp:254` — `checkRemoteMD` never returns `NIXL_SUCCESS`
- `test_transfer.cpp:402` — `createXferReq` returns `-4` (`NIXL_ERR_NOT_FOUND`)
- `test_transfer.cpp:288` — final transfer status check fails
- `test_transfer.cpp:330` — notification count is 0 instead of 1

The failure is **consistent across all 4 UCX test variants** (all hit the same `md[7]=mlx5_7` path with same error), which strongly points to an **environment/hardware problem on `gb200-nvl4-ts2-89`** rather than a code regression. Importantly, the most recent UCX plugin change (PR #1637, commit `71e68170`, `UCX_MAX_HCA_PER_GPU = "auto"`) is the most plausible code-side trigger: setting `MAX_HCA_PER_GPU=auto` on UCX ≥1.21 lets UCX automatically assign more HCAs (including `mlx5_7`) per GPU. On this node, `mlx5_7` appears to not support DMA-buf (or its DMA-buf path is broken), so it was previously excluded by the prior fixed limit, but is now included.

---

**Implicated commit:** `71e68170bb8a` — *tomerg-nvidia* — `PLUGINS/UCX: set UCX_MAX_HCA_PER_GPU to auto (#1637)` (2026-05-15)

**File:** `src/plugins/ucx/ucx_utils.cpp:33` (`config.modify("MAX_HCA_PER_GPU", "auto");`)

**Suggested fix:**

Two complementary actions are recommended:

1. **Short-term / CI fix:** In `ucx_utils.cpp`, guard the `MAX_HCA_PER_GPU=auto` setting behind a check for DMA-buf support, or revert to a conservative fixed value (e.g., `"4"`) on GB200/ARM64 nodes until the `mlx5_7` DMA-buf issue is resolved on the test node. Alternatively, add `UCX_IB_DMABUF_SUPPORT=no` as a fallback for memory registration failures (`ibv_reg_dmabuf_mr` error 524) in the `nixlUcxContext::memReg` path.

2. **Infrastructure fix (preferred):** Investigate `gb200-nvl4-ts2-89` — confirm that `mlx5_7`'s DMA-buf/peer-memory kernel module state is healthy (check `nvidia-peermem` / `nv_peer_mem` driver status and `libibverbs` capabilities on that port). The error `ENOTRECOVERABLE (524)` from `ibv_reg_dmabuf_mr` often indicates a stale/hung peer-memory state that can be cleared with a CUDA context reset or driver reload.

**Related:**
- PR #1637: https://github.com/ai-dynamo/nixl/pull/1637 — `PLUGINS/UCX: set UCX_MAX_HCA_PER_GPU to auto` (the enabling change)
- PR #1507: CI bump UCX v1.20.x → v1.21.x (which activates the UCX ≥1.21 code path)

dpressle:
@ci-triage-agent

dpressle:
👀 Investigating [`nixl-ci-gpu` build 2016](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/2016/pipeline/)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-gpu` · commit `8c0d3c98`

I now have all the evidence needed for a confident diagnosis. Here is the full analysis:

---

**Summary:** POSIX backend test hangs for ~42 minutes after the last benchmark run, triggering the CI pipeline wall-clock timeout (exit code 143).

**Root cause:** The hang is caused by a **broken queue-fallback path in `nixlPosixIOQueue::instantiate()` / `getDefaultIoQueueType()`** introduced by PR #1577 ("Use liburing wrap", commit `652c8560`). Here is the chain of events:

1. **`getDefaultIoQueueType()`** (`io_queue.cpp:78–90`) uses compile-time `#ifdef` priority: `HAVE_LINUXAIO` → `HAVE_LIBURING` → `HAVE_POSIXAIO`. On the CI GPU node (`mizu01`), at least one of these flags selects a queue type string (e.g. `"URING"` or `"AIO"`).
2. **`instantiate()`** (`io_queue.cpp:57–75`) performs an **exact-string match** against the `factories[]` array. The `factories` array is also `#ifdef`-guarded. If the two `#ifdef` trees evaluate differently — or if the `UCCL` backend test exercise path ends up calling the POSIX engine with a queue type whose factory was compiled out — `instantiate()` returns `nullptr`.
3. When `io_queue_` is `nullptr`, **`nixlPosixEngine` constructor** sets `initErr = true` and logs `"Failed to initialize POSIX backend"` — but the engine object is still returned and used.
4. Because `nixlPosixBackendReqH` stores a **reference** (not a copy) to `io_queue_` (`io_queue_.h:166`), and `io_queue_` is `nullptr`, any subsequent call to **`checkXfer()` → `io_queue_->poll()`** dereferences a null pointer, silently hanging or spinning in an undefined state rather than returning an error to the polling loop.
5. The benchmark/test that exercises the POSIX backend (running after the last visible `kill -9 3393329` at `04:29:17`) enters an infinite poll loop with no output — producing **42 minutes of silence** until the Jenkins `timeout()` step fires at `05:11:21` and kills the job with SIGTERM (exit code 143).

The specific smoking-gun bug is in `io_queue.cpp`: `instantiate()` returns `nullptr` on a queue-type mismatch, and neither the engine constructor nor the request-handle code guards against a null `io_queue_` before dereferencing it in `checkXfer()`. PR #1605 ("Fix POSIX backend queue fallback handling") is the open fix for exactly this scenario.

**Implicated commit:** `652c8560a118` — Adit Ranadive, "Use liburing wrap (#1577)" (2026-05-18)

**File:**
- `src/plugins/posix/io_queue.cpp:57–74` — `instantiate()` returns `nullptr` without fallback
- `src/plugins/posix/posix_backend.cpp:231–244` — constructor does not abort on `nullptr` queue; engine proceeds in broken state
- `src/plugins/posix/posix_backend.cpp:189–199` — `checkXfer()` calls `io_queue_->poll()` with no null-guard

**Suggested fix:**
1. **Merge PR #1605** ("Fix POSIX backend queue fallback handling") — this is the targeted fix already in review.
2. As a defensive hardening step regardless, add a null-guard in `checkXfer()`:
   ```cpp
   nixl_status_t nixlPosixBackendReqH::checkXfer() {
       if (!io_queue_) return NIXL_ERR_BACKEND; // guard against null queue
       if (num_confirmed_ios_ == queue_depth_) return NIXL_SUCCESS;
       ...
   }
   ```
3. In `nixlPosixEngine` constructor, treat a `nullptr` return from `instantiate()` as a fatal error and immediately set `initErr = true` with an explicit log — which is already done, but the engine should also refuse `prepXfer`/`postXfer`/`checkXfer` calls when `initErr` is set, preventing the null dereference path entirely.
4. In `io_queue.cpp::instantiate()`, consider **iterating the factory list for a fallback** rather than requiring an exact match, or logging a clear fatal error when the requested type is not compiled in.

**Related:** [PR #1605 – Fix POSIX backend queue fallback handling](https://github.com/ai-dynamo/nixl/pull/1605), [PR #1577 – Use liburing wrap](https://github.com/ai-dynamo/nixl/pull/1577)

brminich:
@aknvda can you pls fix the conflicts?

aknvda:
Fixed in b53218d: merged current main into this branch and resolved the POSIX test conflicts. Verified with `ninja -C build-posix` and `meson test -C build-posix posix_plugin_test --print-errorlogs`.

ovidiusm (approved):

PR #1606, opened 2026-05-01
fix: multiple evaluation in nixlbench macro

weizhoublue:
## What?
I fixed the confirmed double-evaluation bugs in three benchmark macros by storing result in a local temporary and using that value for the check and error message

## Why?
these macros used result more than once inside the macro body. When callers pass a function call as result, that function is
executed multiple times during macro expansion.

## How?
capturing result once avoids repeated execution and makes the macro behave correctly for expressions with side effects.

## Release Notes

* **Bug Fixes**
  * Improved robustness of benchmark diagnostic error reporting to avoid duplicated evaluations and ensure accurate error codes.
  * Updated GPU/CUDA diagnostic checks for more reliable and consistent error messages during benchmarking and troubleshooting.

### DIFF
--- a/benchmark/nixlbench/src/utils/neuron.h
+++ b/benchmark/nixlbench/src/utils/neuron.h
@@ -41,12 +41,13 @@ neuronMemcpy(void *dest, const void *src, size_t count, neuronMemcpyKind kind);
 int
 neuronMemset(void *addr, int val, size_t count);
 
-#define CHECK_NEURON_ERROR(result, message)                                                       \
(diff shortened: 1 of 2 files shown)


guy-ealey-morag (approved):

tomerg-nvidia (approved):

ColinNV (commented):
Why not convert the macros to functions and avoid the issue altogether?

brminich (approved):

PR #1610, opened 2026-05-04
PLUGINS/UCX: Do not allow emulated RMA protocols

tvegas1:
## What?
Disable emulated RMA protocols in UCX plugin. An explicit error will be triggered if some non-zcopy protocols would have to be used.

## Why?
We want obvious failure when zcopy protocols are not usable.

## How?
Leverage added option start in UCX 1.21.

## Summary by CodeRabbit

* **Chores**
  * Tuned UCX communication for newer UCX versions: disabled protocol emulation, adjusted inline response handling, and refined channel settings—improving stability and network performance for distributed workloads.


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -418,6 +418,7 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
         config.modifyAlways("NET_DEVICES", devs_str.c_str());
     }
 
+    config.modify("PROTO_EMULATION_ENABLE", "n");
```
brminich: can you post it to the branch for 1.21 below?

tvegas1: done

brminich (dismissed):

iyastreb (dismissed):

brminich (dismissed):

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -428,6 +428,9 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
     config.modify("RCACHE_MAX_UNRELEASED", "1024");
 
     if (ucpVersion_ >= UCP_VERSION(1, 21)) {
+        config.modify("IB_TX_INLINE_RESP", "0");
```
iyastreb, line 461: IB_TX_INLINE_RESP is 0 by default anyway, right?

iyastreb (dismissed):

src/plugins/ucx/ucx_utils.cpp
tvegas1: it is not in rc base:
````
src/uct/ib/rc/base/rc_iface.c:  {UCT_IB_CONFIG_PREFIX, "RX_INLINE=64;TX_INLINE_RESP=64;RX_QUEUE_LEN=4095;SEG_SIZE=8256", NULL,

PR #1617, opened 2026-05-06
UCX: fail on invalid UCX_TLS for GPU support

brminich:
## What?
This PR adds NIXL-side validation for common invalid `UCX_TLS` settings in the UCX backend.

  What is covered:
   - Fail backend creation when NVIDIA GPU(s) are present, `UCX_TLS` is explicitly set without `cuda` or `cuda_copy`, and the created UCX context does not expose CUDA memory support.

## Summary by CodeRabbit

- **New Features**
  - Added validation for UCX TLS configurations when NVIDIA GPU support is available.
  - Recognizes supported CUDA transport settings, including `cuda`, `cuda_copy`, escaped `\cuda_copy`, and `all`.
  - Improved detection of CUDA and RDMA hardware capabilities.

- **Bug Fixes**
  - Initialization now fails early with a clear error when required CUDA support is unavailable.
  - Hardware capability query failures are now reported instead of silently ignored.
  - Improved reliability of UCX memory and communication operations.

### DIFF
--- a/src/plugins/ucx/config.cpp
+++ b/src/plugins/ucx/config.cpp
@@ -16,12 +16,96 @@
  */
 #include "config.h"
 
+#include <cctype>
 #include <stdexcept>
+#include <string>
 
 #include "common/configuration.h"
+#include "common/hw_info.h"
 #include "common/nixl_log.h"
+#include "ucx_utils.h"
 
 namespace nixl::ucx {
+namespace {
+
+bool
+hasInvalidDenyListSyntax(std::string_view tls) {
+    bool can_start_deny_list = true;
+
+    for (const char c : tls) {
+        if (std::isspace(static_cast<unsigned char>(c)) && can_start_deny_list) {
+            continue;
+        }
+
+        if (c == '^') {
+            if (!can_start_deny_list) {
+                return true;
+            }
+
+            can_start_deny_list = false;
+            continue;
+        }
+
+        can_start_deny_list = false;
+    }
+
+    return false;
+}
+
+bool
+isDenyList(std::string_view tls) {
+    for (const char c : tls) {
+        if (std::isspace(static_cast<unsigned char>(c))) {
+            continue;
+        }
+
+        return c == '^';
+    }
+
+    return false;
+}
+
+bool
+isCudaSupportToken(std::string_view token) {
+    return (token == "cuda") || (token == "cuda_copy");
+}
+
+bool
+hasCudaSupportToken(std::string_view tls) {
+    size_t pos = isDenyList(tls) ? tls.find('^') + 1 : 0;
+
+    while (pos < tls.size()) {
+        size_t end = tls.find(',', pos);
+        if (end == std::string_view::npos) {
+            end = tls.size();
+        }
+
+        while ((pos < end) && std::isspace(static_cast<unsigned char>(tls[pos]))) {
+            ++pos;
+        }
+
+        while ((end > pos) && std::isspace(static_cast<unsigned char>(tls[end - 1]))) {
+            --end;
+        }
+
+        if (isCudaSupportToken(tls.substr(pos, end - pos))) {
+            return true;
+        }
+
+        pos = end + 1;
+    }
+
+    return false;
+}
+
+bool
+tlsEnablesCudaSupport(std::string_view tls) {
+    const bool has_cuda_support = hasCudaSupportToken(tls);
+    return (!isDenyList(tls) && has_cuda_support) || (isDenyList(tls) && !has_cuda_support);
+}
+
+} // namespace
+
 void
 config::modify(std::string_view key, std::string_view value) const {
     const auto env_val = nixl::config::getValueOptional<std::string>("UCX_" + std::string(key));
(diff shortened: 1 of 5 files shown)


guy-ealey-morag:
Should we implement the check that `^` appears only at the beginning in UCX instead?

src/plugins/ucx/ucx_utils.h
```diff
@@ -201,6 +202,9 @@ class nixlUcxContext {
 [[nodiscard]] bool
 nixlUcxMtLevelIsSupported(const nixl::ucx::mt_mode_t) noexcept;
 
+[[nodiscard]] std::optional<bool>
+nixlUcpContextSupportsMemoryType(ucp_context_h ctx, ucs_memory_type_t mem_type) noexcept;
```
rakhmets: This should be a method of `nixlUcxContext`.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -452,11 +466,24 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
...
     if (status != UCS_OK) {
         throw std::runtime_error("Failed to create UCX context: " +
                                  std::string(ucs_status_string(status)));
     }
+
+    try {
+        config.validateTlsCudaSupport(new_ctx);
+    }
+    catch (...) {
+        ucp_cleanup(new_ctx);
```
rakhmets: We need to refactor `nixlUcxContext` first.
We need to get rid of usage of a raw pointer (`ucp_context_h`). It should be replaced by `std::unique_ptr`. Please check other classes: worker, config, rkey.

https://github.com/ai-dynamo/nixl/pull/1241#discussion_r2717042442

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -364,6 +364,20 @@ nixlUcxMtLevelIsSupported(const nixl::ucx::mt_mode_t mt_type) noexcept {
     std::terminate();
 }
 
+std::optional<bool>
+nixlUcpContextSupportsMemoryType(ucp_context_h ctx, ucs_memory_type_t mem_type) noexcept {
+    ucp_context_attr_t attr = {
+        .field_mask = UCP_ATTR_FIELD_MEMORY_TYPES,
+    };
+    const auto status = ucp_context_query(ctx, &attr);
```
rakhmets: I still think that this is an exceptional case.
If query method failed on a successfully created context, it seems to a quite weird case. And with exception here you can get rid of `std::optional`.
I had this comment earlier.

https://github.com/ai-dynamo/nixl/pull/1241#discussion_r2759598955
https://github.com/ai-dynamo/nixl/pull/1241#discussion_r2759819710

src/plugins/ucx/config.cpp
```diff
@@ -17,11 +17,31 @@
...
+
+    bool
+    tlsEnablesCudaSupport(std::string_view tls) {
+        const bool deny_list = !tls.empty() && tls.front() == '^';
+        if (deny_list) {
+            tls.remove_prefix(1);
+        }
+
+        const std::string tokens = "," + std::string(tls) + ",";
+        const bool has_cuda_support = (tls == "all") ||
```
iyastreb: strictly speaking all can be also checked against ",all,"
so maybe:
```
const auto has_cuda_support = std::ranges::any_of(
    std::array{",all,", ",cuda,", ",cuda_copy,"},
    [&](auto t) { return tokens.find(t) != std::string::npos; });
```

iyastreb (dismissed):
I thought we want to append "cuda_copy" to the UCX_TLS
But validation + error is safer option, LGTM

src/plugins/ucx/config.cpp
```diff
@@ -47,6 +67,21 @@ config::modifyAlways(std::string_view key, std::string_view value) const {
     }
 }
 
+void
+config::validateTlsCudaSupport() const {
+    const auto tls = nixl::config::getValueOptional<std::string>("UCX_TLS");
+    if (!tls || tlsEnablesCudaSupport(*tls) || nixl::hwInfo::instance().numNvidiaGpus == 0) {
```
guy-ealey-morag:
```suggestion
    if (nixl::hwInfo::instance().numNvidiaGpus == 0 || !tls || tlsEnablesCudaSupport(*tls)) {
```

Check hw first to skip the tls check if it's not relevant (short circuit)

src/plugins/ucx/config.cpp
```diff
@@ -16,12 +16,34 @@
...
+#include <array>
 #include <stdexcept>
+#include <string>
 
 #include "common/configuration.h"
+#include "common/hw_info.h"
 #include "common/nixl_log.h"
 
 namespace nixl::ucx {
+namespace {
```
rakhmets, line 28: An unnamed namespace with auxiliary functions is usually located outside any namespace.

src/plugins/ucx/config.cpp
```diff
@@ -16,12 +16,34 @@
...
+
+    bool
+    tlsEnablesCudaSupport(std::string_view tls) {
+        const bool deny_list = !tls.empty() && tls.front() == '^';
+        if (deny_list) {
+            tls.remove_prefix(1);
+        }
+
+        const std::string tokens = "," + std::string(tls) + ",";
+        const auto has_cuda_support = std::ranges::any_of(
```
rakhmets: maybe use bool here, as it's used for another local var instead of auto.

test/gtest/hw_warning_test.cpp
```diff
@@ -45,6 +45,48 @@ class HardwareWarningTest : public ::testing::Test {
     }
 };
 
+class UcxTlsValidationTest : public ::testing::Test {
+protected:
+    gtest::ScopedEnv envHelper_;
```
rakhmets: It seems that this should be a local variable of a function, not a class field.

src/plugins/ucx/config.cpp
brminich: fixed

src/plugins/ucx/config.cpp
brminich: fixed

src/plugins/ucx/config.cpp
brminich: fixed

test/gtest/hw_warning_test.cpp
brminich: fixed

src/plugins/ucx/config.cpp
```diff
@@ -47,6 +67,21 @@ config::modifyAlways(std::string_view key, std::string_view value) const {
...
+    if (!tls || tlsEnablesCudaSupport(*tls) || nixl::hwInfo::instance().numNvidiaGpus == 0) {
+        return;
+    }
+
+    const std::string error = "Invalid UCX_TLS=" + *tls +
+        " for NIXL UCX backend: NVIDIA GPU(s) are present, "
+        "but this setting does not enable CUDA memory support. Add cuda_copy for "
+        "basic GPU support, or cuda to also include NVLink support.";
+    NIXL_ERROR << error;
+    throw std::runtime_error(error);
```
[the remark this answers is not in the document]
brminich: it will still be a non fatal warning if UCX_TLS is not set, but GPU memory is not recognized by UCX dues to some reason (wrong env)

rakhmets (approved):

iyastreb (approved):

PR #1621, opened 2026-05-07
UTILS: Exception throw utility function.

ColinNV:
## What?
Make the utility function `throwRuntimeError` from the config utility generally/separately available.

## Why?
Can and will be used in other places; will not remain the only exception-related utility.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal error handling for configuration routines, centralizing how runtime errors are produced and ensuring consistent, clearer error messages.
  * No changes to configuration behavior, public interfaces, or loading precedence; end-user functionality remains unchanged.

### DIFF
--- a/src/utils/common/exception.h
+++ b/src/utils/common/exception.h
@@ -1,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef NIXL_SRC_UTILS_COMMON_EXCEPTION_H
+#define NIXL_SRC_UTILS_COMMON_EXCEPTION_H
+
+#include <sstream>
+#include <stdexcept>
+#include <utility>
+
+namespace nixl {
+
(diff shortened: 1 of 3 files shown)


src/utils/common/exception.h
```diff
@@ -0,0 +1,37 @@
...
+#include <stdexcept>
+#include <utility>
+
+namespace nixl {
+
+template<typename... Ts>
+[[noreturn]] void
+throwRuntimeError(Ts &&...ts) {
+    std::ostringstream oss;
+    (void)(oss << ... << ts);
```
rakhmets: Is this `(void)` really needed? I think it can be removed.

rakhmets: Shouldn't we use std::forward for `ts` here?

ColinNV: I don't think it would make any difference.

rakhmets: Ok, I wasn't sure about this. Most likely, in practice, this will not affect the behavior in any way.

src/utils/common/exception.h
```diff
@@ -0,0 +1,37 @@
...
+#include <stdexcept>
+#include <utility>
+
+namespace nixl {
+
+template<typename... Ts>
+[[noreturn]] void
+throwRuntimeError(Ts &&...ts) {
+    std::ostringstream oss;
+    (oss << ... << ts);
```
rakhmets, line 31:
```suggestion
    oss << ... << ts;
```

ColinNV: It's a fold expression, it needs the brackets!

rakhmets (approved):

src/utils/common/exception.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <sstream>
+#include <stdexcept>
+#include <utility>
+
+namespace nixl {
+
+template<typename... Ts>
+[[noreturn]] void
+throwRuntimeError(Ts &&...ts) {
```
iyastreb, line 29: Does it make sense to also log NIXL error?
Ideally, if possible, we could also log __FILE__ and __LINE__, might be useful for debugging

ColinNV: Since caught exceptions are usually logged that would lead to duplicate log messages. And you don't always know at the throw site whether it warrants a debug, info, warning or error message.

brminich (approved):

PR #1627, opened 2026-05-08
fix: surface register memory error

jordan-wu-97:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed memory registration to now properly surface errors that were previously ignored, improving error reporting and system reliability.


tomerg-nvidia (approved):
LGTM

tomerg-nvidia:
This is a good change however it seems to break some of the existing tests. There's this error in the logs:
```
nixl_agent.cpp:422] registerMem: no available backends for mem type 'DRAM_SEG'
```

PR #1625, opened 2026-05-07
refactor(core): add metadata manager - phase 1

aschwartz12:
## What?
This PR moves metadata coordination into a dedicated `nixlMetadataManager` with a southbound `nixlMetadataBackend` interface.

Main updates:
- Add `nixlMetadataManager` with worker thread, command queue, name-to-UUID cache, and peer registry.
- Route Tier 3 metadata calls through the manager while preserving existing single-peer P2P behavior.
- Add `nixlMetadataBackend` as the backend contract, with P2P socket and ETCD implementations.
- Add `nixl::UUIDv7` generation and emit `AgentName` + `AgentUUID` in metadata blobs, with fallback to the old `Agent` tag.
- Add public metadata API types in `src/api/cpp/nixl_md.h`.
- Add `nixlAgentConfig::heartbeatInterval` for metadata liveness and heartbeat work.
- Add gtest coverage for metadata format, compatibility fallback, UUIDv7, and ETCD key formats.
- Use neutral single-peer helper names and standalone implementation comments.

## Why?
This creates the foundation for the metadata manager architecture while keeping existing Tier 1/3 wire behavior unchanged. The code now supports both list-shaped backend operations and existing single-peer APIs, so the refactor can land without changing current API behavior.

## How?
The manager owns metadata coordination, request queuing, backend instances, and metadata identity state. P2P and ETCD are implemented behind the shared backend contract: P2P handles direct single-peer NIXLCOMM requests, while ETCD preserves agent-name-key compatibility and adds UUID-backed key formation.

## Summary by CodeRabbit

* **New Features**
  * Metadata manager with P2P and optional ETCD backends for metadata exchange, background sync, watches, and invalidation
  * Stable UUIDv7-based agent identity included in metadata
  * Background worker thread and request queue for metadata operations
* **Tests**
  * New tests for metadata format, UUIDv7 shape/ordering, and key construction
* **Chores**
  * Build updates to include UUIDv7 support and conditional backend wiring

PR #1559, opened 2026-04-20
Agent: Use correct synchronization with etcd metadata comm thread

ovidiusm:
## What?
Fix a data race in `nixlAgentData` that sometimes causes `partial_md_example.py --etcd` to crash with SIGABRT.

## Why?
When etcd is enabled, the C++ constructor starts a communication thread that shares agent data structures (`remoteSections_`, `remoteBackends_`, …) with the caller. However, the Python API set `agent_config.syncMode` to `NIXL_THREAD_SYNC_NONE` (no-op locks) only for listener thread, not checking for the etcd comm thread. Concurrent reads/writes on unprotected STL containers caused heap corruption and SIGABRT.

## How?
* Move `useEtcd` before lock in the member declaration order so it is initialized first.
* Add `detectEtcd()` helper to compute useEtcd in the initializer list.
* Add `effectiveSyncMode()` helper that upgrades `SYNC_NONE` to `SYNC_STRICT` when a comm thread will be started (etcd or listen-based).
* The lock member is now initialized with the effective sync mode, eliminating the data race.

## Summary by CodeRabbit

* **Refactor**
  * Simplified agent startup and synchronization by centralizing remote-coordination detection and the decision to run a communication thread, and tightened initialization order for related flags.

* **Bug Fix**
  * Fixed runtime gating and lifecycle handling for remote coordination: thread creation/shutdown, dispatch guards, and debug logging now behave consistently to improve stability and diagnostics.

PR #1635, opened 2026-05-13
file: support path-based file registration in all FILE_SEG backends

lluki:
## What?

Allow callers of the northbound API to declare files by path (string) instead of as file descriptor.

## Why?

SGLang's HiCache issues 128 os.open() calls per flush from Python; GIL contention makes each batch take ~415ms. Moving the opens to C++ gives ~24x batch-open speedup in our mock and ~6-12x end-to-end warm-phase TTFT speedup in hibench against Qwen-32B.

## How?

Allow callers to declare files by path in nixlBlobDesc::metaInfo:

    metaInfo := <modes>:<path>     # path-mode (new)
              | <anything else>    # legacy fd-in-devId
    modes    := <access>[,<flag>]*
    access   := "ro" | "rw"
    flag     := "direct" | "sync" | "noatime" | "create"

Backends open the file in registerMem and close it in deregisterMem. Wired in all four file-aware plugins via a shared helper at src/utils/file/file_path_mode.{h,cpp} (parsePathMeta + nixlFilePathMD owned-fd RAII base):

  - POSIX  
  - HF3FS 
  - CUDA_GDS 
  - GDS_MT

Strictly additive: unknown access/flag tokens fall through to legacy mode, so existing fd-in-devId callers are unaffected.

## Testing 

I ran the smoke test included in this PR on: 

* POSIX : tested
* GDS: tested (on dlcluster, viking-prod machines)
* GDS-MT: tested (on dlcluster, viking-prod machines)
* H3FS : tested against a 3FS cluster. However, #1645 is necessary to compile NIXL with h3fs. 

The smoke test registers some memory then transfers from a file. At the end, it verifies no stale file handles.

### Mock 
In a single-file repro that imitates the SGLang access pattern minimally:

- One main thread busy-loops on a non-yielding boolean flag (mimicking the scheduler's CPU usage).
- A "nixl-connector" thread runs 16 iterations of (open 128 files ->  `register_memory` -> `READ` xfer → cleanup) against `/raid/hicache/*.bin` (256 KB each).
- An env var `NIXL_MOCK_MODE=fd|path` switches between the legacy fd-based registration and our proposed path-mode, so the two are directly comparable on the same box and the same workload.

The mock reports per-iteration, per-128-batch, and per-individual-`os.open` timings. The 128-batch number is the one we want to improve.

| Phase            | fd mode (legacy) | path mode (new) | Δ |
|------------------|------------------|-----------------|---|
| Per-128-batch open | ~415 ms        | **~17 ms**      | **24× faster** |
| Per-iteration    | ~684 ms          | **~55 ms**      | **12× faster** |
| Total wall-clock | ~12.0 s          | **~1.9 s**      | **6× faster**  |

With path mode, the 128 sequential Python-level `os.open()` calls collapse into a single `register_memory` call where the opens happen in C++, thus we collapse the 128 GIL re-acquires to just one. Note, this works because the `register_memory` is already batched (takes a list of descriptors), without this batching we would not be able to improve! 

### Integrated in SGLang

PoC integration of the path mode to HiCache NIXL for KV Cache loading from SSD. It leads to significant TTFT improvements with `tp=1` (one GPU). Changes on the SGLang NIXL connector are necessary to use the path mode, just integrating this patched NIXL version will not change anything (it will still use the FD mode).

TTFT with `tp=1`:
image

It is a bit less significant in `tp>1` because the worker communication adds more GIL yield points.

## Summary by CodeRabbit

* **New Features**
  * Path-mode file registration: register files by path+access-mode (backend opens/closes files) across GPU and POSIX backends.

* **Documentation**
  * Comprehensive docs for path-mode vs fd-mode and backend file-lifecycle expectations.

* **Tests**
  * New path-mode smoke/integration tests exercising registration, write/read roundtrips, and FD lifecycle.

* **Chores**
  * CI updated to gracefully skip path-mode smoke tests when runtime plugins or binaries are unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1635)

### DIFF
--- a/src/plugins/cuda_gds/gds_backend.cpp
+++ b/src/plugins/cuda_gds/gds_backend.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -16,9 +16,11 @@
  */
 #include <cassert>
 #include <cufile.h>
+#include <absl/strings/str_format.h>
 #include "gds_backend.h"
 #include "gds_utils.h"
 #include "common/nixl_log.h"
+#include "file/file_path_mode.h"
 #include "file/file_utils.h"
 #include <unordered_map>
 #include <memory>
@@ -95,67 +97,53 @@ nixl_status_t nixlGdsEngine::registerMem(const nixlBlobDesc &mem,
                                          const nixl_mem_t &nixl_mem,
                                          nixlBackendMD* &out)
 {
-    nixl_status_t status = NIXL_SUCCESS;
-    nixlGdsMetadata *md = new nixlGdsMetadata();
+    if (nixl_mem == FILE_SEG) {
+        std::unique_ptr<nixlGdsMetadata> md;
+        try {
+            md = std::make_unique<nixlGdsMetadata>(nixl::FileFd(mem.devId, mem.metaInfo));
+        }
+        catch (const std::system_error &e) {
+            NIXL_ERROR << "GDS path-mode open failed: " << e.what();
+            return NIXL_ERR_BACKEND;
+        }
+        int fd = md->file_fd.fd();
+
+        if (auto it = gds_file_map.find(fd); it != gds_file_map.end()) {
+            md->handle = it->second;
+            md->handle.size = mem.len;
+            md->handle.metadata = mem.metaInfo;
+        } else {
+            nixl_status_t s = gds_utils->registerFileHandle(fd, mem.len, mem.metaInfo, md->handle);
+            if (s != NIXL_SUCCESS) {
+                return s; // ~FileFd via unique_ptr drop closes the owned fd
+            }
+            gds_file_map[fd] = md->handle;
+        }
+        out = md.release();
+        return NIXL_SUCCESS;
+    }
+
+    auto md = std::make_unique<nixlGdsMetadata>();
     md->type = nixl_mem;
-    cudaError_t error_id;
 
-    switch (nixl_mem) {
-        case FILE_SEG: {
-            // Check if we already have a file handle for this devId
-            auto it = gds_file_map.find(mem.devId);
-            if (it != gds_file_map.end()) {
-                md->handle = it->second;
-                md->handle.size = mem.len;
-                md->handle.metadata = mem.metaInfo;
-                break;
-            }
-
-            status = gds_utils->registerFileHandle(mem.devId, mem.len,
-                                                   mem.metaInfo, md->handle);
-            if (status == NIXL_SUCCESS) {
-                gds_file_map[mem.devId] = md->handle;
-            }
-            break;
+    if (nixl_mem == VRAM_SEG) {
+        if (cudaError_t e = cudaSetDevice(mem.devId); e != cudaSuccess) {
+            NIXL_ERROR << "cudaSetDevice returned " << cudaGetErrorString(e) << " for device ID "
+                       << mem.devId;
+            return NIXL_ERR_BACKEND;
         }
-
-        case VRAM_SEG: {
-            error_id = cudaSetDevice(mem.devId);
-            if (error_id != cudaSuccess) {
-                NIXL_ERROR << "cudaSetDevice returned " << cudaGetErrorString(error_id)
-                          << " for device ID " << mem.devId;
-                delete md;
-                return NIXL_ERR_BACKEND;
-            }
-            status = gds_utils->registerBufHandle((void *)mem.addr, mem.len, 0);
-            if (status == NIXL_SUCCESS) {
-                md->buf.base = (void *)mem.addr;
-                md->buf.size = mem.len;
-            }
-            break;
-        }
-
-        case DRAM_SEG: {
-            status = gds_utils->registerBufHandle((void *)mem.addr, mem.len, 0);
-            if (status == NIXL_SUCCESS) {
-                md->buf.base = (void *)mem.addr;
-                md->buf.size = mem.len;
-            }
-            break;
-        }
-
-        default:
-            status = NIXL_ERR_BACKEND;
-            break;
+    } else if (nixl_mem != DRAM_SEG) {
+        return NIXL_ERR_BACKEND;
     }
 
-    if (status != NIXL_SUCCESS) {
-        delete md;
-        return status;
+    if (nixl_status_t s = gds_utils->registerBufHandle((void *)mem.addr, mem.len, 0);
+        s != NIXL_SUCCESS) {
+        return s;
     }
-
-    out = (nixlBackendMD*)md;
-    return status;
+    md->buf.base = (void *)mem.addr;
+    md->buf.size = mem.len;
+    out = md.release();
+    return NIXL_SUCCESS;
 }
 
 nixl_status_t nixlGdsEngine::deregisterMem (nixlBackendMD* meta)
@@ -164,11 +152,21 @@ nixl_status_t nixlGdsEngine::deregisterMem (nixlBackendMD* meta)
     if (md->type == FILE_SEG) {
         gds_utils->deregisterFileHandle(md->handle);
         gds_file_map.erase(md->handle.fd);
-        // No need to close fd since we're not opening files
     } else {
         gds_utils->deregisterBufHandle(md->buf.base);
     }
-    delete md;  // Clean up the metadata object
+    delete md;
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlGdsEngine::resolveFileHandle(const nixlMetaDesc &d, gdsFileHandle &fh) const {
+    auto *md = static_cast<const nixlGdsMetadata *>(d.metadataP);
+    if (!md || md->type != FILE_SEG) {
+        NIXL_ERROR << "GDS: missing FILE_SEG metadata at xfer time";
+        return NIXL_ERR_NOT_FOUND;
+    }
+    fh = md->handle;
     return NIXL_SUCCESS;
 }
 
@@ -220,13 +218,10 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
             total_size = remote[i].len;
             base_offset = (size_t)local[i].addr;
 
-            auto it = gds_file_map.find(local[i].devId);
-            if (it == gds_file_map.end()) {
-                NIXL_ERROR << "File handle not found";
+            if (resolveFileHandle(local[i], fh) != NIXL_SUCCESS) {
                 delete gds_handle;
                 return NIXL_ERR_NOT_FOUND;
             }
-            fh = it->second;
         } else {
             base_addr = (void*)local[i].addr;
             if (!base_addr) {
@@ -236,13 +231,10 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
             total_size = local[i].len;
             base_offset = (size_t)remote[i].addr;
 
-            auto it = gds_file_map.find(remote[i].devId);
-            if (it == gds_file_map.end()) {
-                NIXL_ERROR << "File handle not found";
+            if (resolveFileHandle(remote[i], fh) != NIXL_SUCCESS) {
                 delete gds_handle;
                 return NIXL_ERR_NOT_FOUND;
             }
-            fh = it->second;
         }
 
         // Split large transfers into multiple requests
--- a/src/plugins/cuda_gds/gds_backend.h
+++ b/src/plugins/cuda_gds/gds_backend.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -28,15 +28,17 @@
 #include <mutex>
 #include "gds_utils.h"
 #include "backend/backend_engine.h"
+#include "file/file_path_mode.h"
 
-class nixlGdsMetadata : public nixlBackendMD {
-    public:
-        gdsFileHandle handle;
-        gdsMemBuf buf;
-        nixl_mem_t type;
+class nixlGdsMetadata : public nixlFilePathMD {
+public:
+    gdsFileHandle handle;
+    gdsMemBuf buf;
+    nixl_mem_t type;
 
-        nixlGdsMetadata() : nixlBackendMD(true) { }
-        ~nixlGdsMetadata() { }
+    nixlGdsMetadata() = default;
+
+    explicit nixlGdsMetadata(nixl::FileFd &&fd) : nixlFilePathMD(std::move(fd)), type(FILE_SEG) {}
 };
 
 class GdsTransferRequestH {
@@ -97,9 +99,13 @@ class nixlGdsEngine : public nixlBackendEngine {
 
         nixlGdsIOBatch* getBatchFromPool(unsigned int size) const;
         void returnBatchToPool(nixlGdsIOBatch* batch) const;
-        nixl_status_t createAndSubmitBatch(const std::vector<GdsTransferRequestH>& requests,
-                                           size_t start_idx, size_t batch_size,
-                                           std::vector<nixlGdsIOBatch*>& batch_list) const;
+        nixl_status_t
+        createAndSubmitBatch(const std::vector<GdsTransferRequestH> &requests,
+                             size_t start_idx,
+                             size_t batch_size,
+                             std::vector<nixlGdsIOBatch *> &batch_list) const;
+        nixl_status_t
+        resolveFileHandle(const nixlMetaDesc &d, gdsFileHandle &fh) const;
         nixl_status_t createBatches(const nixl_xfer_op_t &operation,
                                    const nixl_meta_dlist_t &local,
                                    const nixl_meta_dlist_t &remote,
--- a/src/plugins/gds_mt/gds_mt_backend.cpp
+++ b/src/plugins/gds_mt/gds_mt_backend.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -32,6 +32,7 @@
 #include "common/nixl_log.h"
 #include "gds_mt_backend.h"
 #include "gds_mt_utils.h"
+#include "file/file_path_mode.h"
 #include "file/file_utils.h"
 #include <taskflow/taskflow.hpp>
 #include <unordered_map>
@@ -158,7 +159,6 @@ runCuFileOp (GdsMtTransferRequestH *req, std::atomic<nixl_status_t> *overall_sta
 nixl_status_t
 extractTransferParams (const nixlMetaDesc &mem_desc,
                        const nixlMetaDesc &file_desc,
-                       const std::unordered_map<int, std::weak_ptr<gdsMtFileHandle>> &file_map,
                        void *&base_addr,
                        size_t &total_size,
                        size_t &base_offset,
@@ -167,15 +167,17 @@ extractTransferParams (const nixlMetaDesc &mem_desc,
     total_size = mem_desc.len;
     base_offset = (size_t)file_desc.addr;
 
-    auto it = file_map.find (file_desc.devId);
-    if (it == file_map.end()) {
-        NIXL_ERROR << "GDS_MT: error: file metadata not found";
+    const auto *md = static_cast<const nixlGdsMtMetadata *>(file_desc.metadataP);
+    if (!md) {
+        NIXL_ERROR << "GDS_MT: missing FILE_SEG metadata at xfer time";
         return NIXL_ERR_NOT_FOUND;
     }
-
-    auto handle = it->second.lock();
-    NIXL_ASSERT (handle);
-    cu_fhandle = handle->cu_fhandle;
+    auto *file_data = std::get_if<FileSegData>(&md->data_);
+    if (!file_data || !file_data->handle) {
+        NIXL_ERROR << "GDS_MT: file metadata is not a FILE_SEG variant";
+        return NIXL_ERR_NOT_FOUND;
+    }
+    cu_fhandle = file_data->handle->cu_fhandle;
     return NIXL_SUCCESS;
 }
 } // namespace
@@ -200,26 +202,35 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
                               nixlBackendMD *&out) {
     switch (nixl_mem) {
     case FILE_SEG: {
-        auto it = gds_mt_file_map_.find (mem.devId);
-        std::shared_ptr<gdsMtFileHandle> handle;
-        if (it != gds_mt_file_map_.end()) {
-            handle = it->second.lock();
-            if (handle) {
-                out = new nixlGdsMtMetadata (handle);
-                return NIXL_SUCCESS;
-            }
-            gds_mt_file_map_.erase (it);
-        }
-
+        nixl::FileFd file_fd;
         try {
-            handle = std::make_shared<gdsMtFileHandle> (mem.devId);
+            file_fd = nixl::FileFd(mem.devId, mem.metaInfo);
         }
-        catch (const std::exception &e) {
-            NIXL_ERROR << "GDS_MT: failed to create file handle: " << e.what();
+        catch (const std::system_error &e) {
+            NIXL_ERROR << "GDS_MT: path-mode open failed: " << e.what();
             return NIXL_ERR_BACKEND;
         }
-        gds_mt_file_map_[mem.devId] = handle;
-        out = new nixlGdsMtMetadata (handle);
+        int fd = file_fd.fd();
+
+        std::shared_ptr<gdsMtFileHandle> handle;
+        if (auto it = gds_mt_file_map_.find(fd); it != gds_mt_file_map_.end()) {
+            handle = it->second.lock();
+            if (!handle) {
+                gds_mt_file_map_.erase(it);
+            }
+            // Cache hit: drop file_fd (~FileFd closes any duplicate owned fd).
+        }
+        if (!handle) {
+            try {
+                handle = std::make_shared<gdsMtFileHandle>(std::move(file_fd));
+            }
+            catch (const std::exception &e) {
+                NIXL_ERROR << "GDS_MT: failed to create file handle: " << e.what();
+                return NIXL_ERR_BACKEND;
+            }
+            gds_mt_file_map_[fd] = handle;
+        }
+        out = new nixlGdsMtMetadata(std::move(handle));
         return NIXL_SUCCESS;
     }
 
@@ -251,11 +262,11 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
 
 nixl_status_t
 nixlGdsMtEngine::deregisterMem (nixlBackendMD *meta) {
-    std::unique_ptr<nixlGdsMtMetadata> md ((nixlGdsMtMetadata *)meta);
+    std::unique_ptr<nixlGdsMtMetadata> md((nixlGdsMtMetadata *)meta);
 
     if (auto *file_data = std::get_if<FileSegData> (&md->data_)) {
         if (file_data->handle) {
-            int key = file_data->handle->fd;
+            int key = file_data->handle->file_fd.fd();
             md.reset(); // Release metadata first
 
             auto it = gds_mt_file_map_.find (key);
@@ -263,7 +274,7 @@ nixlGdsMtEngine::deregisterMem (nixlBackendMD *meta) {
                 gds_mt_file_map_.erase (it);
             }
 
-            // No need to close fd since we're not opening files
+            // owned fds: closed by ~FileFd (RAII) on last shared_ptr drop.
         }
     }
 
@@ -304,7 +315,6 @@ nixlGdsMtEngine::prepXfer (const nixl_xfer_op_t &operation,
         if (is_local_file) {
             param_status = extractTransferParams (remote[i],
                                                   local[i],
-                                                  gds_mt_file_map_,
                                                   base_addr,
                                                   total_size,
                                                   base_offset,
@@ -312,7 +322,6 @@ nixlGdsMtEngine::prepXfer (const nixl_xfer_op_t &operation,
         } else {
             param_status = extractTransferParams (local[i],
                                                   remote[i],
-                                                  gds_mt_file_map_,
                                                   base_addr,
                                                   total_size,
                                                   base_offset,
--- a/src/plugins/hf3fs/hf3fs_backend.cpp
+++ b/src/plugins/hf3fs/hf3fs_backend.cpp
@@ -24,6 +24,7 @@
 #include "hf3fs_backend.h"
 #include "hf3fs_log.h"
 #include "common/nixl_log.h"
+#include "file/file_path_mode.h"
 #include "file/file_utils.h"
 
 #define NUM_CQES 1024
@@ -185,11 +186,18 @@ nixl_status_t nixlHf3fsEngine::registerMem (const nixlBlobDesc &mem,
         break;
     }
     case FILE_SEG: {
-        int fd = mem.devId;
+        std::unique_ptr<nixlHf3fsFileMetadata> md;
+        try {
+            md = std::make_unique<nixlHf3fsFileMetadata>(nixl::FileFd(mem.devId, mem.metaInfo));
+        }
+        catch (const std::system_error &e) {
+            HF3FS_LOG_RETURN(NIXL_ERR_BACKEND,
+                             absl::StrFormat("HF3FS path-mode open failed: %s", e.what()));
+        }
+        int fd = md->file_fd.fd();
 
         // if the same file is reused - no need to re-register
-        auto it = hf3fs_file_set.find(fd);
-        if (it == hf3fs_file_set.end()) {
+        if (hf3fs_file_set.find(fd) == hf3fs_file_set.end()) {
             int ret = 0;
             status = hf3fs_utils->registerFileHandle(fd, &ret);
             if (status != NIXL_SUCCESS) {
@@ -199,11 +207,10 @@ nixl_status_t nixlHf3fsEngine::registerMem (const nixlBlobDesc &mem,
             hf3fs_file_set.insert(fd);
         }
 
-        nixlHf3fsFileMetadata *md = new nixlHf3fsFileMetadata();
         md->handle.fd = fd;
         md->handle.size = mem.len;
         md->handle.metadata = mem.metaInfo;
-        out = (nixlBackendMD *)md;
+        out = md.release();
         break;
     }
     default:
@@ -299,8 +306,8 @@ nixl_status_t nixlHf3fsEngine::prepXfer (const nixl_xfer_op_t &operation,
     }
 
     for (int i = 0; i < file_cnt; i++) {
-        // Get file descriptor from the proper list
-        int file_descriptor = (*file_list)[i].devId;
+        auto file_md = static_cast<nixlHf3fsFileMetadata *>((*file_list)[i].metadataP);
+        int file_descriptor = file_md->file_fd.fd();
         addr = (void*) (*mem_list)[i].addr;
         size = (*mem_list)[i].len;
         offset = (size_t) (*file_list)[i].addr;  // Offset in file
--- a/src/plugins/hf3fs/hf3fs_backend.h
+++ b/src/plugins/hf3fs/hf3fs_backend.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -28,6 +28,7 @@
 #include <thread>
 #include "hf3fs_utils.h"
 #include "backend/backend_engine.h"
+#include "file/file_path_mode.h"
 
 class nixlHf3fsShmException : public std::runtime_error {
 public:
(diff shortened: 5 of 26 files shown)


dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/25922857045/job/76196183208)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `680c451c`

I now have a complete, confident diagnosis. Here is the full report:

---

**Summary:** PR #1635 (`path-mode`) failed the automated PR Size Check because it modifies **609 lines** of existing files, exceeding the repo's hard limit of **500 lines**.

**Root cause:** The workflow script counts lines added to modified files (`git show --numstat --diff-filter=M`, excluding `subprojects/*`). PR #1635 — "support path-based file registration in all FILE_SEG backends" — accumulates 609 net-added lines across modified files, which is 109 lines over the 500-line threshold enforced by the check. The job script explicitly `exit 1`s on this condition:
```
❌ PR size check failed!
This PR adds 609 lines (excluding subprojects), which exceeds the maximum of 500 lines.
Process completed with exit code 1.
```
No test crash, no hang, no infrastructure issue — this is a pure policy enforcement failure.

**Implicated commit:** `680c451cb4ed819596441f753cc23d26bf891f75` (branch: `path-mode`, merged ref: `4efb72fa24e60e98b4138a6af239f4ad82c947ad`)

**File:** The PR Size Check workflow script (`.github/workflows/` — the step "Calculate detailed breakdown of changes")

**Suggested fix:** The PR author has two clean options:
1. **Split the PR** — Break PR #1635 into two or more smaller PRs, each under 500 modified lines. For example, split backend-by-backend or separate the core path-registration logic from the per-backend wiring.
2. **Refactor for compactness** — Review the 609-line diff for any boilerplate that can be shared/deduplicated (e.g., common helper functions, macros, or base-class methods) to bring the modification count below 500 lines.

> ⚠️ **Do NOT raise the 500-line limit** as the fix — this is an intentional policy gate to encourage well-scoped, reviewable PRs. The right action is to reduce the diff size.

**Related:** https://github.com/ai-dynamo/nixl/pull/1635

dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/25924120085/job/76200624304)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `a223b11f`

The diagnosis is complete and unambiguous. Here is the full report:

---

**Summary:** The "PR Size Check" workflow failed because PR #1635 (`path-mode` branch) adds **639 lines** of code changes to existing files, exceeding the enforced hard limit of **500 lines**.

**Root cause:** The CI job runs `git show --numstat --pretty="" --diff-filter=M -- . ':(exclude)subprojects/*'` and sums all lines added to modified files. For this PR (commit `a223b11ff8fde86355f19945180ce9620f4e3fb9`), that total came to **639 lines**, which triggers the `if [[ "$LINES_CHANGED" -gt 500 ]]` guard and exits with code 1. The check is a policy gate, not a code defect — the PR is simply too large per the project's contribution guidelines.

```
❌ PR size check failed!
This PR adds 639 lines (excluding subprojects), which exceeds the maximum of 500 lines.
Process completed with exit code 1.
```

**Implicated commit:** `a223b11ff8fde86355f19945180ce9620f4e3fb9` (PR #1635 author on branch `path-mode`)

**File:** `.github/workflows` (the PR Size Check workflow) — policy enforced via `git show --numstat` diff analysis

**Suggested fix:** The PR author has three options:
1. **Split the PR** — Divide the 639-line changeset into two or more PRs, each staying under the 500-line ceiling. For example, split the "path-based file registration" feature into a core infrastructure PR and a follow-on PR for individual backend implementations.
2. **Reduce scope** — Trim the current PR (e.g., remove generated files, refactor large blocks, or defer non-essential changes to a follow-up PR) so the modified-file line count drops below 500.
3. **Request a policy exception** — If the reviewers/maintainers agree the change is genuinely atomic and cannot be split, a maintainer can bypass or override the check. The workflow check itself could also be updated to use `--diff-filter=AM` (added + modified) or adjust the threshold, but that requires a deliberate policy decision.

> ⚠️ Note: there is also a deprecation warning that `actions/checkout@v3` runs on Node.js 20, which will be removed from runners on September 16, 2026. Consider bumping it to `actions/checkout@v4` while the PR is being revised.

**Related:** [PR #1635 — file: support path-based file registration in all FILE_SEG backends](https://github.com/ai-dynamo/nixl/pull/1635)

dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/25925180658/job/76204411329)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `7fe1c5bf`

I have a clear and complete picture. Here is the diagnosis:

---

**Summary:** PR #1635 (`path-mode` branch) failed the "PR Size Check" because it modifies **824 lines** in existing files, exceeding the enforced 500-line limit.

**Root cause:** The workflow at `.github/workflows/pr-size-check.yml` counts lines added to **modified** (not newly-created) files via `git show --numstat --diff-filter=M`, then hard-fails with `exit 1` if the count exceeds 500. PR #1635 introduced 824 changed lines across modified files (subprojects excluded), triggering the guard. The logic itself worked correctly — this is a genuine policy violation, not a flaky test or infrastructure issue.

Key log lines:
```
❌ PR size check failed!
This PR adds 824 lines (excluding subprojects), which exceeds the maximum of 500 lines.
Process completed with exit code 1.
```

**Implicated commit:** `7fe1c5bf606cde760c5cc76791539487340d1a4c` (branch: `path-mode`, PR author)

**File:** `.github/workflows/pr-size-check.yml:23` — the `if [[ "$LINES_CHANGED" -gt 500 ]]` threshold

**Suggested fix:** The PR author has two options:
1. **Split the PR** — divide the 824-line change into two or more smaller PRs, each staying under the 500-line threshold (preferred, as this respects the policy's intent of keeping reviews manageable).
2. **Request a policy exception** — if the change is genuinely indivisible (e.g. a large mechanical refactor), the team can temporarily raise or bypass the limit by adding a `# noqa-size` label/override mechanism to the workflow, or by bumping the threshold for this PR only via a workflow dispatch input. This should be a conscious team decision, not a silent workaround.

Note: there is also a minor housekeeping issue — `actions/checkout@v3` is deprecated in favor of Node.js 24-compatible versions; upgrading to `actions/checkout@v4` would silence the runner warning before the September 2026 deadline.

**Related:** https://github.com/ai-dynamo/nixl/pull/1635

dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/25926490476/job/76209025132)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `1ac7d824`

---

**Summary:** PR #1635 (`path-mode` branch) fails the **PR Size Check / clang-format** job because 4 modified C/C++ files contain style violations that `clang-format-19` would reformat.

**Root cause:** The PR introduced new code in 4 files that do not comply with the project's `.clang-format` rules. Specifically, `clang-format-diff-19` reports three categories of violations:

| File | Lines | Violation |
|---|---|---|
| `src/plugins/gds_mt/gds_mt_utils.cpp` | ~60 | Single-statement `if` body must use braces on separate lines — `if (owned && fd >= 0) ::close(fd);` must become a 3-line `{ ... }` block |
| `test/unit/plugins/cuda_gds/nixl_gds_path_mode_test.cpp` | ~144–145 | Line-break-before-assignment in `nixl_status_t r = agent.createXferReq(...)` should be collapsed to one line (fits within column limit) |
| `test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp` | ~326–327 | Same pattern — `nixl_status_t r =\n    agent.createXferReq(...)` should be on one line |
| `test/unit/plugins/hf3fs/nixl_hf3fs_test.cpp` | ~199–200 | `std::cerr` stream chain broken in a way clang-format would reflow onto two lines differently |

The CI script runs `git diff … | clang-format-diff-19 -p1 -style=file …` and exits with code **1** whenever the diff output is non-empty, indicating unformatted changes.

**Implicated commit:** `1ac7d8242e0c36475c117f02c75ee79eae10383b` (PR #1635 author, branch `path-mode`)

**File:**
- `src/plugins/gds_mt/gds_mt_utils.cpp` : line ~60
- `test/unit/plugins/cuda_gds/nixl_gds_path_mode_test.cpp` : line ~144
- `test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp` : line ~326
- `test/unit/plugins/hf3fs/nixl_hf3fs_test.cpp` : line ~199

**Suggested fix:** Run `clang-format-19` (or the project's formatting helper) on all four files locally and commit the result. The exact changes needed are shown in the CI diff:

```bash
# From the repo root, reformat only the changed files:
clang-format-19 -i -style=file \
  src/plugins/gds_mt/gds_mt_utils.cpp \
  test/unit/plugins/cuda_gds/nixl_gds_path_mode_test.cpp \
  test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp \
  test/unit/plugins/hf3fs/nixl_hf3fs_test.cpp

git add -u
git commit -m "style: apply clang-format-19 fixes for path-mode PR"
git push
```

The three concrete edits required are:
1. **`gds_mt_utils.cpp:60`** — expand `if (owned && fd >= 0) ::close(fd);` into a brace-enclosed block.
2. **`nixl_gds_path_mode_test.cpp:144` and `nixl_gds_mt_test.cpp:326`** — merge the broken two-line `nixl_status_t r =\n    agent.createXferReq(...)` back onto a single line.
3. **`nixl_hf3fs_test.cpp:199`** — reflow the `std::cerr` stream expression to match the formatter's preferred column-break position.

**Related:** none

dpressle:
👀 Investigating [`Run Pre-Commit Hooks`](https://github.com/ai-dynamo/nixl/actions/runs/25926971321/job/76210722709)…

dpressle:
👀 Investigating [`Copyright Checks`](https://github.com/ai-dynamo/nixl/actions/runs/25926971321/job/76210722709)…

dpressle:
🤖 **CI Triage Agent** — `Run Pre-Commit Hooks` · commit `db1c567e`

I now have all the evidence needed for a complete diagnosis. Here is the full report:

---

**Summary:** The `pre-commit` CI job failed because `subprojects/tomlplusplus-3.4.0/vendor/catch.hpp` had a **trailing newline removed** by the commit on branch `path-mode`, which caused the `end-of-file-fixer` hook to flag it (files covered by `types_or: [c, c++, …]` must end with exactly one newline).

**Root cause:** Commit `db1c567e0b84849f2c0acb665147561780d3cae6` (branch `path-mode`, PR #1635) reformatted several files under `subprojects/tomlplusplus-3.4.0/`. As part of that reformatting, the final trailing newline was **deleted** from `vendor/catch.hpp`:

```diff
- // end catch.hpp
  #endif // TWOBLUECUBES_SINGLE_INCLUDE_CATCH_HPP_INCLUDED
-           ← blank line removed here
```

The `.pre-commit-config.yaml` `end-of-file-fixer` hook (from `pre-commit/pre-commit-hooks@v4.4.0`) requires that all C/C++ files end with exactly one newline. Removing it causes the hook to auto-fix and exit with code 1, failing the whole job.

The same commit also reformatted several Python tool files (`generate_single_header.py`, `generate_windows_test_targets.py`, `utils.py`, `version.py`, `requirements.txt`) under `subprojects/tomlplusplus-3.4.0/tools/` — converting tabs→spaces (Black-style). Those changes themselves are consistent with what `black` and `isort` would produce, so they don't individually cause failures; but **the missing EOF newline on `catch.hpp` is the sole exit-code-1 trigger**.

**Implicated commit:** `db1c567e0b84849f2c0acb665147561780d3cae6` (branch `path-mode`, PR #1635)

**File:** `subprojects/tomlplusplus-3.4.0/vendor/catch.hpp` — last line (line 17975), trailing newline deleted

**Suggested fix:** Re-add a single trailing newline to the end of `subprojects/tomlplusplus-3.4.0/vendor/catch.hpp`. The simplest one-liner:
```bash
echo "" >> subprojects/tomlplusplus-3.4.0/vendor/catch.hpp
```
Or run `pre-commit run end-of-file-fixer --files subprojects/tomlplusplus-3.4.0/vendor/catch.hpp` locally and commit the auto-fix. Going forward, if `subprojects/` vendored files are not meant to be linted, consider adding an `exclude` pattern in `.pre-commit-config.yaml` for `subprojects/` (similar to the `codespell` hook's existing `exclude` list).

**Related:** none

dpressle:
🤖 **CI Triage Agent** — `Copyright Checks` · commit `db1c567e`

Now the picture is completely clear. The workflow runs `pre-commit` **only on modified files** in the PR. The PR #1635 (`path-mode` branch, commit `db1c567e`) introduced or modified files in `subprojects/tomlplusplus-3.4.0/` — in particular Python tools files and `vendor/catch.hpp` — that fail several pre-commit hooks: **`black`** (Python code style), **`isort`** (Python import order), and **`end-of-file-fixer`** (trailing newline).

---

**Summary:** Pre-commit hooks (`black`, `isort`, and `end-of-file-fixer`) fail on newly added/modified files in `subprojects/tomlplusplus-3.4.0/` introduced by PR #1635.

**Root cause:** PR #1635 (branch `path-mode`, commit `db1c567e`) added or modified vendored `subprojects/tomlplusplus-3.4.0/` Python tool files (`tools/generate_single_header.py`, `tools/generate_windows_test_targets.py`, `tools/version.py`, `tools/utils.py`, `tools/requirements.txt`) and `vendor/catch.hpp`. These files use tab-indentation and single-quoted strings (not `black`-compliant), have unsorted imports (not `isort`-compliant), and `vendor/catch.hpp` has a trailing newline that `end-of-file-fixer` wants to remove. The `.pre-commit-config.yaml` runs `black`, `isort`, and `end-of-file-fixer` against all modified files in the PR without excluding the `subprojects/` vendored directory, so the unformatted third-party Python files trip all three hooks.

**Implicated commit:** `db1c567e0b84849f2c0acb665147561780d3cae6` (PR #1635 author, branch `path-mode`)

**File:**
- `subprojects/tomlplusplus-3.4.0/tools/generate_single_header.py` — `black` + `isort` violations
- `subprojects/tomlplusplus-3.4.0/tools/generate_windows_test_targets.py` — `black` + `isort` violations
- `subprojects/tomlplusplus-3.4.0/tools/version.py` — `black` + `isort` violations
- `subprojects/tomlplusplus-3.4.0/tools/utils.py` — `black` + `isort` violations
- `subprojects/tomlplusplus-3.4.0/vendor/catch.hpp` — `end-of-file-fixer` (trailing newline)

**Suggested fix:** Two complementary options:

1. **(Preferred — exclude the vendored directory)** Add an exclusion for `subprojects/` to `.pre-commit-config.yaml` so third-party code is never linted. Add a top-level `exclude` key:
   ```yaml
   exclude: ^subprojects/
   ```
   This is the right long-term fix since vendored code should not be reformatted by the project's own style tools.

2. **(Alternative — for the PR itself)** If `subprojects/tomlplusplus-3.4.0/` is genuinely new code added by this PR rather than truly vendored, run `black`, `isort`, and fix the trailing newline in `catch.hpp` locally before pushing:
   ```bash
   black subprojects/tomlplusplus-3.4.0/tools/
   isort --profile black --filter-files subprojects/tomlplusplus-3.4.0/tools/
   # Remove trailing newline from catch.hpp
   ```

Option 1 is strongly recommended since the diff shows these are clearly third-party files (tomlplusplus upstream tooling).

**Related:** [PR #1635](https://github.com/ai-dynamo/nixl/pull/1635)

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26020264108/job/76479827654)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `4549a84e`

I have everything needed for a confident diagnosis. Here is the complete breakdown:

---

**Summary:** Clang Format Check fails on PR #1635 (`path-mode` branch) — 5 files have single-statement control-flow bodies without braces, and one has a function-call argument list split across lines in a style not matching `.clang-format` rules.

**Root cause:** The commit `4549a84e96ad89dff0ce3bb2fc793fc87df87bf3` introduced new code across 5 files that violates the project's `clang-format-19` style rules. The checker (`clang-format-diff-19`) produced a non-empty diff and exited with code 1. There are two categories of violations:

1. **Braceless single-statement `if`/`for`/`while`/`else` bodies** — the `.clang-format` config enforces `InsertBraces: true` (or equivalent), requiring all control-flow bodies to be wrapped in `{}`. Every occurrence of the pattern `if (...) single_stmt;` was flagged. Affected locations:
   - `src/plugins/gds_mt/gds_mt_utils.cpp` line ~62 — `if (owned && fd >= 0) ::close(fd);`
   - `test/unit/plugins/cuda_gds/nixl_gds_test.cpp` lines ~208, ~214, ~322 — braceless `if`s and a braceless init-statement `if`
   - `test/unit/plugins/path_mode_common.h` lines ~40, ~107, ~117–119, ~128–130, ~134, ~136, ~150–151 — braceless `for`, `while`, and `if`/`else` chains
   - `test/unit/plugins/path_mode_common_cuda.h` lines ~26, ~29–30 — braceless `if` and `for`
   - `test/unit/plugins/posix/nixl_posix_test.cpp` lines ~778, ~786–787, ~806 — braceless `if` returns

2. **Function-call argument alignment** — `nixl_test::runPathModeSmoke(...)` in `nixl_gds_test.cpp` ~line 200 had each argument on its own line aligned under the opening `(`, but clang-format wants the arguments packed onto as few lines as possible after the opening of the call.

**Implicated commit:** `4549a84e96ad89dff0ce3bb2fc793fc87df87bf3` (PR #1635 author, branch `path-mode`)

**File:**
- `src/plugins/gds_mt/gds_mt_utils.cpp` : line 62
- `test/unit/plugins/cuda_gds/nixl_gds_test.cpp` : lines 199–214, 321–322
- `test/unit/plugins/path_mode_common.h` : lines 38–40, 95–136, 148–151
- `test/unit/plugins/path_mode_common_cuda.h` : lines 24–32
- `test/unit/plugins/posix/nixl_posix_test.cpp` : lines 775–788, 806

**Suggested fix:** Run `clang-format-19` (with `-style=file`) in-place on every modified file before pushing. The exact command that will fix all violations is:

```bash
# From the repo root — reformat only the files flagged by the check
clang-format-19 -style=file -i \
  src/plugins/gds_mt/gds_mt_utils.cpp \
  test/unit/plugins/cuda_gds/nixl_gds_test.cpp \
  test/unit/plugins/path_mode_common.h \
  test/unit/plugins/path_mode_common_cuda.h \
  test/unit/plugins/posix/nixl_posix_test.cpp
```

Then `git add` and `git commit --amend` (or push a follow-up fixup commit). Alternatively, add a pre-commit hook or editor integration that runs `clang-format` on save so these are caught locally before CI.

**Related:** [PR #1635](https://github.com/ai-dynamo/nixl/pull/1635), [PR #461](https://github.com/ai-dynamo/nixl/pull/461) (original GDS-MT backend introduction)

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26084070062/job/76692586739)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `4461be84`

---

**Summary:** Clang Format Check fails on PR #1635 — a multi-line boolean condition in `test/unit/plugins/path_mode_common.h` (line ~116) is split across two lines in a way that violates the project's 100-column limit rule enforced by `clang-format-19`.

**Root cause:** The file `test/unit/plugins/path_mode_common.h` (introduced by this PR) contains the following hand-wrapped `if` condition:

```cpp
// Current code (as written in PR):
if (xfer(NIXL_READ) == NIXL_SUCCESS &&
    mem.verify(buf_a, kPatternA, size) &&
        mem.verify(buf_b, kPatternB, size)) {
```

`clang-format-19` with `ColumnLimit: 100` and `BreakBeforeBinaryOperators: None` (operators stay at end of line) wants to collapse the first two sub-expressions onto a single line because they fit within 100 columns, producing:

```cpp
// clang-format expected:
if (xfer(NIXL_READ) == NIXL_SUCCESS && mem.verify(buf_a, kPatternA, size) &&
    mem.verify(buf_b, kPatternB, size)) {
```

The diff from the CI log is explicit:
```diff
-            if (xfer(NIXL_READ) == NIXL_SUCCESS &&
-                mem.verify(buf_a, kPatternA, size) &&
+            if (xfer(NIXL_READ) == NIXL_SUCCESS && mem.verify(buf_a, kPatternA, size) &&
                 mem.verify(buf_b, kPatternB, size)) {
```

The `.clang-format` config key `ColumnLimit: 100` and `BreakBeforeBinaryOperators: None` drive this reformatting — the manually inserted line break after `NIXL_SUCCESS &&` is not permitted because the combined expression fits within 100 characters.

**Implicated commit:** `4461be844afe67d399a3eadbece1655025bca529` (PR #1635 author, branch `path-mode`)

**File:** `test/unit/plugins/path_mode_common.h:116`

**Suggested fix:** Run `clang-format-19` locally and commit the result. The one-liner fix is:

```bash
clang-format-19 -i -style=file test/unit/plugins/path_mode_common.h
git add test/unit/plugins/path_mode_common.h
git commit --amend --no-edit   # or a new commit
```

Specifically, change line ~116 from:
```cpp
// Before
if (xfer(NIXL_READ) == NIXL_SUCCESS &&
    mem.verify(buf_a, kPatternA, size) &&
        mem.verify(buf_b, kPatternB, size)) {
```
to:
```cpp
// After (what clang-format produces)
if (xfer(NIXL_READ) == NIXL_SUCCESS && mem.verify(buf_a, kPatternA, size) &&
    mem.verify(buf_b, kPatternB, size)) {
```

It is also worth running `clang-format-19` across **all** files changed in this PR (the CI listed 17 changed C/C++ files) to catch any other formatting issues before re-pushing:
```bash
clang-format-19 -i -style=file \
  src/plugins/cuda_gds/gds_backend.{cpp,h} \
  src/plugins/gds_mt/gds_mt_backend.cpp \
  src/plugins/gds_mt/gds_mt_utils.{cpp,h} \
  src/plugins/hf3fs/hf3fs_backend.{cpp,h} \
  src/plugins/posix/posix_backend.{cpp,h} \
  src/utils/file/file_path_mode.{cpp,h} \
  test/unit/plugins/{path_mode_common.h,path_mode_common_cuda.h} \
  test/unit/plugins/cuda_gds/nixl_gds_test.cpp \
  test/unit/plugins/gds_mt/nixl_gds_mt_test.cpp \
  test/unit/plugins/hf3fs/nixl_hf3fs_test.cpp \
  test/unit/plugins/posix/nixl_posix_test.cpp
```

**Related:** [PR #1635 — file: support path-based file registration in all FILE_SEG backends](https://github.com/ai-dynamo/nixl/pull/1635)

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -203,6 +222,26 @@ nixl_status_t nixlGdsEngine::prepXfer (const nixl_xfer_op_t &operation,
     // Determine if local is the file segment
     bool is_local_file = (local.getType() == FILE_SEG);
 
+    // Resolve via metadataP (path-mode opens fd inside registerMem, so the
+    // xfer-time devId doesn't match gds_file_map's fd key); fall back to
+    // devId lookup for fd-mode descriptors.
+    auto resolve_fh = [&](const nixlMetaDesc &d, gdsFileHandle &fh) -> nixl_status_t {
```
ofer: This should be a private member method rather than an anonymous function

src/plugins/hf3fs/hf3fs_backend.cpp
```diff
@@ -299,8 +324,9 @@ nixl_status_t nixlHf3fsEngine::prepXfer (const nixl_xfer_op_t &operation,
     }
 
     for (int i = 0; i < file_cnt; i++) {
-        // Get file descriptor from the proper list
-        int file_descriptor = (*file_list)[i].devId;
+        // Read fd from metadataP; in path-mode devId is 0 (path is in metaInfo).
+        auto file_md = (nixlHf3fsFileMetadata *)(*file_list)[i].metadataP;
```
ofer: What happens if there is no metadata and it's the file descriptor as devId case?

src/plugins/hf3fs/hf3fs_backend.h
ofer, line 50: I see a repeat in all of these file backends, why not make a nixlFileSegBackendMD that has an owned and path?

lluki: Right, good point. I think a second about it, maybe I can restructure a bit such that your comment below becomes an "obviously correct" code.

src/plugins/cuda_gds/gds_backend.cpp
lluki: agree, will do

src/plugins/hf3fs/hf3fs_backend.cpp
lluki: There is always metadata also in the `devId = fd` case. Assuming the registration will be called. But I will fix the nullpointer exception a invalid API usage would cause with returning an error code.

See these two locations where it gets set in the legacy case:
 * 1: https://github.com/ai-dynamo/nixl/pull/1635/changes/BASE..0d2106d2bbc28dcb70cc9b7eacc785d88a21808e#diff-a1a94f7f40b4730076dcd6fd1ae8356404491e3a9e41526d26c0d503ef265ae2R204
 * 2: https://github.com/ai-dynamo/nixl/pull/1635/changes/BASE..0d2106d2bbc28dcb70cc9b7eacc785d88a21808e#diff-a1a94f7f40b4730076dcd6fd1ae8356404491e3a9e41526d26c0d503ef265ae2R204

ofer: As prepXfer can be called indirectly by the user, they could assign the devid in the transfer lists instead of using the metadata given by the register mem, it could cause a bug

lluki: I see.

What would be your suggestion for fixing this? What is the API contract with the caller? Do we only require a previously registered `devId` but not metadata? Then we maintain an internal `devId -> metadata` map and ignore the passed metadata?

lluki:
If you want to have another look @ofer - I refactored it such that the backends figure out what is the right FD to use. Then, store it in metadata. From there on, only rely on the metadata. I also tried your suggestion of having a common superclass, it is possible but requires a lot of refactoring. I think the current approach is quite nice.

src/plugins/cuda_gds/gds_backend.h
```diff
@@ -28,14 +28,21 @@
 #include <mutex>
 #include "gds_utils.h"
 #include "backend/backend_engine.h"
+#include "file/file_path_mode.h"
 
 class nixlGdsMetadata : public nixlBackendMD {
```
ofer: why aren't you inheriting the nixlFilePathMD?

src/utils/file/file_path_mode.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+    std::string modes = s.substr(0, colon);
+    std::string path = s.substr(colon + 1);
+    if (path.empty()) {
+        return std::nullopt;
+    }
+
+    std::vector<std::string> tokens = absl::StrSplit(modes, ',');
+    if (tokens.empty()) {
+        return std::nullopt;
```
ofer: add a debug log to help developers see if they put in a bad string into metadata

src/plugins/gds_mt/gds_mt_backend.cpp
```diff
@@ -200,26 +202,35 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
...
+        }
+        catch (const std::system_error &e) {
+            NIXL_ERROR << "GDS_MT: path-mode open failed: " << e.what();
+            return NIXL_ERR_BACKEND;
+        }
+        int fd = file_fd.fd();
+
         std::shared_ptr<gdsMtFileHandle> handle;
-        if (it != gds_mt_file_map_.end()) {
+        if (auto it = gds_mt_file_map_.find(fd); it != gds_mt_file_map_.end()) {
```
ofer, line 216: Why would there be empty file handles?

lluki: Hm, i think this is the default case - you register a new FD then the `gds_mt_file_map_` is empty and a new entry is inserted. If a second registration on the same `devId` is performed, the handle is dropped (and destructed/closed, if it was the last instance). Maybe the second case should be an error, but I didnt want to change the behavior for existing clients.

src/plugins/cuda_gds/gds_backend.h
lluki: Good point. will do

src/utils/file/file_path_mode.cpp
lluki: Will do, some callers i think do provide a pure path even in the legacy devId mode. In that case the path is stored (in GDS backends and h3fs) but never used - at least as far as i can see. To not make too many debug prints, I suggest printing only when the string generally looks like a path mode string: Starts with `ro` or `rw` and contains a colon.

ofer (dismissed):

src/utils/file/file_path_mode.cpp
ofer: Sounds good

lluki:
CI failures seem unrelated to this PR. POSIX path mode test passed on CI.

dpressle:
@ci-triage-agent

dpressle:
👀 Investigating [`nixl-ci-gpu` build 1933](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/1933/pipeline/)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-gpu` · commit `773875ec`

I now have sufficient evidence for a confident diagnosis. Here is the full analysis:

---

## Analysis

### Timeline from the log

| Time | Event |
|---|---|
| `10:16:18.073` | gtest suite starts (109 tests) |
| `10:17:51.140` | `[13/109] LocalNonLocalMDExchange` completes (18645 ms) |
| `10:18:01.423` | `[14/109] EtcdSendLocalAndFetchRemote` **starts** |
| `10:18:37.796` | WebSocket `close 1006 / unexpected EOF` — **the build pod dies** |
| — | `EtcdSendLocalAndFetchRemote` never printed a completion line |

The gap between `10:18:01` and `10:18:37` is **36 seconds of silence**, and the test never finished. The test itself contains **multiple `sleep_for(sleep_time)` calls where `sleep_time = std::chrono::seconds(10)`** plus three full etcd round-trip sequences. The test requires a live, reachable etcd endpoint. When etcd is unreachable or the `NIXL_ETCD_ENDPOINTS` variable is set but the endpoint is not responding, the etcd `Watch` call can silently block far beyond the intended 10-second timeout, causing the container pod to hit its wall-clock limit.

The test **does** have a `VERIFY_ETCD_MODE()` guard (lines 581), but that guard only skips when `NIXL_ETCD_ENDPOINTS` is **empty or unset**. If the variable is set to an endpoint that is currently **down, unreachable, or overloaded**, the guard passes and the test proceeds, after which the etcd `Watch` inside `fetchRemoteMD` hangs indefinitely instead of timing out cleanly.

The Jenkins pipeline then interprets the pod crash as a WebSocket close 1006, and the overall `parallel` stage returns exit code 1.

---

**Summary:** `MetadataExchangeTestFixture.EtcdSendLocalAndFetchRemote` hung inside an etcd `Watch` call, causing the Jenkins pod to exhaust its wall-clock budget and die with WebSocket close 1006.

**Root cause:** The `EtcdSendLocalAndFetchRemote` test (line 580 of `test/gtest/metadata_exchange.cpp`) calls `fetchRemoteMD()` which issues an etcd `Watch`. When the etcd endpoint is set but not responding, the `Watch` call blocks indefinitely — the 10-second `sleep_time` periods accumulate and the test never completes. The `VERIFY_ETCD_MODE()` guard at line 581 only skips when `NIXL_ETCD_ENDPOINTS` is empty/unset; it does not guard against a configured-but-unreachable endpoint. The CI runner ran out of wall time (pod killed by Kubernetes → WebSocket EOF) while waiting inside this hung test.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` — ovidiusm, *"Agent: Use correct synchronization with etcd metadata comm thread (#1559)"* (most recent commit touching the etcd comm-worker path)

**File:** `test/gtest/metadata_exchange.cpp:580` (`EtcdSendLocalAndFetchRemote`) and `src/core/` etcd comm-worker

**Suggested fix:**
1. **Immediate / CI reliability:** Add a connectivity pre-check in `VERIFY_ETCD_MODE()` — after verifying the env var is set, attempt a short-timeout probe (e.g. a TCP connect or a `grpc::Channel` health check with a 2-second deadline) to the etcd endpoint. If it fails, `GTEST_SKIP()` rather than hanging.
2. **Defensive fix in the etcd watch path:** Ensure the `Watch`/`fetchRemoteMD` implementation enforces an absolute deadline (not just the `sleep_time` in the test) so that a non-responsive etcd endpoint causes a clean timeout error rather than an indefinite block.
3. **CI infra check:** Verify the etcd service on the CI cluster is healthy and the `NIXL_ETCD_ENDPOINTS` endpoint is reachable from the `nixl-ci-gpu` pod. If the service is flapping, this test will continue to cause intermittent pod kills.

**Related:** [#1625 — refactor(core): add metadata manager - phase 1](https://github.com/ai-dynamo/nixl/pull/1625), [#1559 — Agent: Use correct synchronization with etcd metadata comm thread](https://github.com/ai-dynamo/nixl/pull/1559) — none directly track this hang.



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `2392fed4-d38d-48f9-9a31-a6ec7bd49718` in the triage console for the audit trail.

mkhazraee:
Great that we're working on this again, @lluki if you haven't already, please take a look at the initial implementations Vish worked on:
https://github.com/ai-dynamo/nixl/pull/279
https://github.com/ai-dynamo/nixl/pull/353

One point is we shouldn't have create/delete in nixl APIs, we don't allocate/deallocate in NIXL. Opening a file is equivalent to giving access to that file to NIXL, which matches the NIXL registration concept, even if under the hood an in case of write we're adding data, that's abstracted away by the underlying FS.

Regarding RO/RW, are you enforcing them during xfer req creation too? I didn't see any changes in agent cpp file, maybe it should go there, not sure the best place.

Finally, just realized the GDS_MT doesn't have any readme file, while IIRC that's the main one being used. Maybe good to replicate/move the readme there.

lluki:
Hi Moein, thanks for looking into this

> Great that we're working on this again, @lluki if you haven't already, please take a look at the initial implementations Vish worked on: #279 #353

Will look into it!

> One point is we shouldn't have create/delete in nixl APIs, we don't allocate/deallocate in NIXL. Opening a file is equivalent to giving access to that file to NIXL, which matches the NIXL registration concept, even if under the hood an in case of write we're adding data, that's abstracted away by the underlying FS.

I think you're right, I wouldn't consider the "write-extends-file" behavior an allocation. This PR supports passing  `O_CREAT` which creates a new file if it doesnt exist. On the other hand, the OBJ semantics are similar... Performance wise, it would be good to keep `create` in: In the GIL contention problem of SGLang, every syscall is expensive. If I have to create the file first, we loose some performance on the SET path. 

Another performance consideration I thought about but decided against: Call `ftruncate` on registration such that the FS can pre-allocate blocks. But then we'd have to change the registration API even further. Also, we might introduce some new synchronization problem: The file will for a short moment be visible with the correct size and completely 0-filled. 


> Regarding RO/RW, are you enforcing them during xfer req creation too? I didn't see any changes in agent cpp file, maybe it should go there, not sure the best place.

I don't do that. We just launch the write on a file descriptor that would be open in read-only mode and propagate the error from the file system back to the caller. There's also some problems with that error propagation, see my other PR https://github.com/ai-dynamo/nixl/pull/1697.

I'll suggest a suitable check for this in the agent.
 
> Finally, just realized the GDS_MT doesn't have any readme file, while IIRC that's the main one being used. Maybe good to replicate/move the readme there.

Hm, not sure I'm the right person and this is the right PR :-) I mean I can give it a try but not sure I'm the best person to do that.

lluki:
I had claude draft a change for the early check on `createXferReq`:  https://github.com/ai-dynamo/nixl/commit/7661d3b81f73fb0f16eacbe3bb2225e7b67e2414 . I dont know if its worth it, it has to add quite a bit of boilerplate for catching just one case. My suggestion is to not add it and rely on the filesystem error propagation

lluki:
Some comments of looking through the discussions in https://github.com/ai-dynamo/nixl/pull/279 https://github.com/ai-dynamo/nixl/pull/353


* Addition of nixlFileUtils supporting Unlink/Stat etc. helpers
	* This PR doesn't add any helpers. Clients have to use plain POSIX for non-transfers (such as `stat`)

* Request of RAII (https://github.com/ai-dynamo/nixl/pull/353#discussion_r2103017593)
	* This PR uses RAII
	
* Discussion of always include O_CREAT https://github.com/ai-dynamo/nixl/pull/353/changes#r2217776168
	* This PR requires explicit passing of `create` -> `O_CREAT`, that makes the the problem pointed out by the comment not applicable.
	* For my primary use-case in SGLang HiCache, always passing O_CREAT would be fine. But I prefer to have it explicit: The use-case of writing to a pre-existing file is valid. In this case it would be good to catch user errors and not accidentally create files.
	* I just added to the test-suite a check on non-existing files.

lluki:
Sorry, I ( /Claude) made a big mess with the experimental branch and I accidentally force-pushed a bit too often. 
Here is a compare link that compares what Ofer/Moein have reviewed with the current branch:

12e7607dc175aafbdb0c031702f36cd987a85d8d   -> 9465244c4afed031d3d1c4f942e7b47c935bf3a1 :  [Compare](https://github.com/ai-dynamo/nixl/compare/12e7607dc175aafbdb0c031702f36cd987a85d8d..9465244c4afed031d3d1c4f942e7b47c935bf3a1) 

I just added another test to the testcase...

ofer (approved):

aranadive:
@lluki please add support for this option in the PR in the CI tests - .gitlab/test_cpp.sh.
Also add nixlbench support for this one. We will need QA to test the path option correctly. Its a new feature for FILE_SEG API - separate PR is fine.

lluki:
Hi @aranadive 

Tests are already implemented. No nixlbench support yet. Is nixlbench ok in a separate PR? Something specific you are looking for other than "it works" ? I could try to simulate the SGLang behavior, but we would need a python run in nixlbench, not sure we have or want that.

Test that is run in CI:

https://github.com/ai-dynamo/nixl/pull/1635/changes#diff-3b3eeee19f735ac4da2f779b8b808466e066a3a127a6cee88f4bc461d84700e1R148-R206  (and similar)

Given that i added them to the existing test binaries, they should all run in CI too. For example here's the output for POSIX of the last CI run on this PR:

https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/rest/organizations/jenkins/pipelines/nixl-ci-non-gpu/runs/1714/nodes/472/steps/479/log/?start=0

These are the relevant lines:

```
[2026-06-03T16:44:08.307Z] + ./bin/nixl_posix_test -n 128 -s 1048576
[2026-06-03T16:44:08.307Z] 
[2026-06-03T16:44:08.307Z] Total events read: 40
[2026-06-03T16:44:08.307Z] Final buffer size: 0 events
[2026-06-03T16:44:08.307Z] NIXL POSIX Plugin Test
[2026-06-03T16:44:08.307Z] parsePathMeta: OK
[2026-06-03T16:44:08.307Z] O_CREAT path-mode: OK
[2026-06-03T16:44:08.307Z] E0603 16:44:07.866465    2943 posix_backend.cpp:265] POSIX path-mode open failed: nixl::FileFd("/tmp/nixl_posix_path_mode_smoke.bin.missing"): No such file or directory
[2026-06-03T16:44:08.307Z] E0603 16:44:07.866537    2943 nixl_agent.cpp:468] registerMem: registration failed for the specified or all potential backends
[2026-06-03T16:44:08.307Z] POSIX path-mode roundtrip OK (baseline=9)
```

So they already run and even succeed!

ovidiusm (approved):

PR #1637, opened 2026-05-13
PLUGINS/UCX: set UCX_MAX_HCA_PER_GPU to auto

tomerg-nvidia:
## What?
Se the default in setting for `UCX_MAX_HCA_PER_GPU` to `auto`. This will cause UCX to register GPU memory only on closest HCAs.

## Why?
Currently UCX was registering every memory buffer on every MD, which is consuming a lot of ICM memory.

## Summary by CodeRabbit

* **Chores**
  * Automatic GPU resource optimization enabled for systems using UCX 1.21 and later to improve compatibility and performance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1637)

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -427,6 +427,10 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
         config.modify("RC_GDA_NUM_CHANNELS", std::to_string(num_device_channels));
     }
 
+    if (ucpVersion_ >= UCP_VERSION(1, 21)) {
+        config.modify("MAX_HCA_PER_GPU", "auto");
+    }
+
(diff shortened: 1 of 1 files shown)


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -427,6 +427,10 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
         config.modify("RC_GDA_NUM_CHANNELS", std::to_string(num_device_channels));
     }
 
+    if (ucpVersion_ >= UCP_VERSION(1, 21)) {
```
brminich: no need to define separate if, can you pls move it to another if above?

tomerg-nvidia: done

brminich (approved):

PR #1640, opened 2026-05-14
telemetry/doca: add nixl_common_dep so tomlplusplus resolves

e-eygin:
## What?

The DOCA telemetry plugin's `meson.build` was missing `nixl_common_dep` in its dependency list. This PR adds it — a one-line change — mirroring what the Prometheus plugin already does in `src/plugins/telemetry/prometheus/meson.build` (line 51).

Without this, the DOCA plugin currently fails to compile against `main` with:

```
fatal error: toml++/toml.hpp: No such file or directory
```

## Why?

Since commit `e539a11e` ("UTILS/CONFIG: Add config file reading.") `src/utils/common/configuration.h` directly includes `<toml++/toml.hpp>`. `doca_exporter.cpp` includes `common/configuration.h`, so the DOCA plugin now transitively requires `tomlplusplus` on its include path. `nixl_common_dep` already carries `tomlplusplus_dep` (see `src/utils/common/meson.build:16,28,55`), so declaring it on the DOCA plugin is the canonical fix and matches the Prometheus plugin's pattern.

### Why didn't CI catch this?

`contrib/Dockerfile` installs `doca-sdk-gpunetio` and `libdoca-sdk-verbs-dev` but **not** `libdoca-sdk-telemetry-exporter-dev`. So the `cpp.has_header('doca_telemetry_exporter.h', ...)` guard in `src/plugins/telemetry/doca/meson.build` line 22 short-circuits, `doca_found` stays `false`, and the entire `shared_library('libtelemetry_exporter_doca', ...)` block is skipped in CI. The toml++ leak is therefore invisible to CI. Anyone with the DOCA telemetry-exporter SDK installed hits this immediately on a clean rebuild — and the day someone updates `contrib/Dockerfile` to install the package (a known pending item to make CI actually exercise the DOCA plugin), this PR will be a prerequisite for keeping CI green.

## How?

One-line change in `src/plugins/telemetry/doca/meson.build`:

```diff
-        dependencies: [nixl_infra, absl_log_dep, doca_dep],
+        dependencies: [nixl_infra, nixl_common_dep, absl_log_dep, doca_dep],
```

### Test plan

Verified locally with `libdoca-sdk-telemetry-exporter-dev` 3.3.0109 + `collectx-clxapidev` 1.24.3 installed:

- `ninja -C build install` — clean rebuild, DOCA plugin compiles and installs successfully, no warnings.
[...]

PR #1649, opened 2026-05-17
telemetry: remove redundant category field from events

e-eygin:
## What?

This PR removes the redundant telemetry category field from `nixlTelemetryEvent`.

The event now carries only:

- `eventType_`
- `value_`

It also removes category-related runtime surfaces:

- Removes `nixl_telemetry_category_t`.
- Removes `nixlEnumStrings::telemetryCategoryStr`.
- Drops the `category` label from Prometheus and DOCA telemetry metrics.
- Updates C++ and Python telemetry readers.
- Updates telemetry unit tests.
- Updates `docs/telemetry.md` and telemetry exporter READMEs.
- Bumps `TELEMETRY_VERSION` from `2` to `3`.

This also resolves the separate `nixlTelemetry::updateData` argument-order cleanup request: once category is removed, both `nixlTelemetryEvent` construction and `updateData` use `(event_type, value)`.

Depends on #1640. Please review only commit `a330177` for now. After #1640 merges, this branch will be rebased onto `ai-dynamo/main` so the PR diff contains only the category-removal change.

## Why?

`category_` is fully derivable from `eventType_`.

For example:

- `AGENT_TX_BYTES` is always a transfer metric.
- `AGENT_MEMORY_REGISTERED` is always a memory metric.
- `AGENT_XFER_TIME` is always a timing metric.

Because of that, storing category in every telemetry event and exporting it as a metric label is redundant. The metric/event name already carries the meaning.

Removing it also deletes the category string conversion path and avoids emitting an extra `category=...` label on every Prometheus / DOCA metric sample.

`TELEMETRY_VERSION` is bumped because `nixlTelemetryEvent` is serialized into the shared-memory cyclic buffer by binary layout. Removing `category_` changes field offsets, so readers must reject old buffer files instead of silently decoding them incorrectly. This follows the existing precedent from the timestamp-field removal, which bumped the version from `1` to `2`.

## How?

Core telemetry changes:

- `nixlTelemetryEvent` now stores only `eventType_` and `value_`.
- `nixlTelemetry::updateData` now takes only `(event_type, value)`.
- Event producer call sites no longer pass category.
- Backend telemetry plumbing now constructs events with `(event_type, value)`.

Exporter changes:

- Prometheus registers metric instances with only `hostname` and `agent_name` labels.
- DOCA registers only the per-event `agent_name` label; `hostname` remains a constant source label.
- Exporters dispatch based on the event type / registered metric maps instead of `event.category_`.

Reader/docs/test changes:

- C++ and Python telemetry readers no longer print category.
- Python reader layout now matches the C++ event layout (`uint32_t event_type`, padding, `uint64_t value`).
- Tests were updated to assert the new layout, including `TELEMETRY_VERSION == 3` and `sizeof(nixlTelemetryEvent) == 16`.
- Documentation no longer lists categories as part of the telemetry event model.
- `docs/telemetry.md` now documents the `TELEMETRY_VERSION` policy.

Validation:

- `ninja -C build install`
- `build/test/gtest/gtest --gtest_filter='telemetryTest.*:prometheusTelemetryTest.*'`
  - 15/15 passed
- Prometheus smoke test:
  - agent metrics appeared with `agent_name` and `hostname`
  - no `category=` label was present

## Summary by CodeRabbit

* **Refactor**
  * Telemetry events simplified: category removed, event-type-focused schema and version bump; exporters and metric registration updated to operate without category labels.

* **Documentation**
  * Telemetry docs and exporter READMEs updated to reflect new schema, event-type naming, sample output, and environment-variable/table formatting.

* **Examples**
  * Example readers updated to expect the new format and stop emitting event category.

* **Tests**
  * Unit and integration tests adjusted to validate the streamlined event structure.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1649?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/25993374095/job/76403448367)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `a3301772`

I have all the evidence I need. Here is the complete diagnosis:

---

**Summary:** Clang Format Check failed on PR #1649 — two function definitions in `prometheus_exporter.cpp` have parameter lists manually wrapped across two lines in a style that `clang-format-19` reformats onto a single line.

**Root cause:** The PR (`telemetry-remove-category-field`, commit `a3301772`) removed the `category` parameter from `registerCounter` and `registerGauge`. After removal, the two remaining parameters (`name` and `help`) fit comfortably on one line within the column limit, so `clang-format-19` collapses the manually broken two-line signatures into one. The committed code still has them split across two lines, which disagrees with what `clang-format` produces:

```diff
--- src/plugins/telemetry/prometheus/prometheus_exporter.cpp  (before formatting)
+++ src/plugins/telemetry/prometheus/prometheus_exporter.cpp  (after formatting)
@@ -131,8 +131,7 @@
 void
-nixlTelemetryPrometheusExporter::registerCounter(const std::string &name,
-                                                 const std::string &help) {
+nixlTelemetryPrometheusExporter::registerCounter(const std::string &name, const std::string &help) {

-nixlTelemetryPrometheusExporter::registerGauge(const std::string &name,
-                                               const std::string &help) {
+nixlTelemetryPrometheusExporter::registerGauge(const std::string &name, const std::string &help) {
```

The old three-parameter signatures were wide enough to require wrapping; now that `category` is gone the shorter two-parameter signatures must be on one line per the project's `.clang-format` rules.

**Implicated commit:** `a3301772961f66c18e5e75d5aeaeeebdb89acfd8` (branch `telemetry-remove-category-field`, PR #1649)

**File:** `src/plugins/telemetry/prometheus/prometheus_exporter.cpp` — lines ~153–154 and ~156–157 (the `registerCounter` and `registerGauge` function definitions)

**Suggested fix:** Run `clang-format-19` locally on the file (or just apply the diff above) to collapse those two split signatures onto single lines:

```cpp
// Change from:
nixlTelemetryPrometheusExporter::registerCounter(const std::string &name,
                                                 const std::string &help) {
// To:
nixlTelemetryPrometheusExporter::registerCounter(const std::string &name, const std::string &help) {

// Change from:
nixlTelemetryPrometheusExporter::registerGauge(const std::string &name,
                                               const std::string &help) {
// To:
nixlTelemetryPrometheusExporter::registerGauge(const std::string &name, const std::string &help) {
```

The quickest way to fix it is to run `clang-format-19 -i -style=file src/plugins/telemetry/prometheus/prometheus_exporter.cpp` and amend/push the commit.

**Related:** none

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/25994091112/job/76405412461)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `531a5f83`

---

**Summary:** `clang-format` check fails on PR #1649 — single-statement `if` body in `src/api/cpp/backend/backend_engine.h` must use braces per the project's style rules.

**Root cause:** The commit `531a5f83` introduced (or retained) a braceless single-line `if` guard in `addTelemetryEvent()`:

```cpp
// CURRENT (fails clang-format-19):
if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) return;

// REQUIRED by .clang-format style:
if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) {
    return;
}
```

The `clang-format-diff-19` tool applied to the diff of this PR detected the violation and produced a non-empty patch, causing the job to exit with code 1. The two earlier `if (!enableTelemetry_) return;` guards on the lines immediately above use the same braceless style, but those lines were not part of this PR's diff so they were not flagged — only the newly added/modified line was checked. Note that the lock/queue-check lines appear to have been reordered in this PR (queue-size check is now before the mutex lock), which is what put the line into the diff and triggered the linter.

**Implicated commit:** `531a5f83ba39b38301d382608bff1105b0db0b03` (PR #1649, branch `telemetry-remove-category-field`)

**File:** `src/api/cpp/backend/backend_engine.h` — the `addTelemetryEvent` method, specifically the `if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE)` guard (around line 68 in the file).

**Suggested fix:** Expand the braceless `return` into a braced block as `clang-format-19` requires:

```cpp
void
addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) {
    if (!enableTelemetry_) return;
    std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
    if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) {
        return;
    }
    telemetryEvents_.emplace_back(event_type, value);
}
```

The fastest way to do this locally is to run:
```bash
clang-format-19 -i -style=file src/api/cpp/backend/backend_engine.h
```
and then commit the result. This will auto-apply exactly the diff the CI checker expected.

**Related:** #1531 (telemetry enum refactor by BatshevaBlack, last touch to this file), #1522 (Remove timestamp field from telemetry events) — no existing issue specifically for this clang-format violation.

src/api/cpp/backend/backend_engine.h
```diff
@@ -64,10 +64,9 @@ class nixlBackendEngine {
         void
         addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) {
```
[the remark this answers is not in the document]
e-eygin: I verified this against both main and the parent commit. The addTelemetryEvent(nixl_telemetry_event_type_t event_type, uint64_t value) signature already existed before this PR, so there is no API/ABI signature break here. This PR only changes the method body to construct the new two-field nixlTelemetryEvent. The actual compatibility break is the shared-memory event layout, which is why TELEMETRY_VERSION was bumped to 3 and documented.

e-eygin:
@ColinNV, @ovidiusm please review

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -70,6 +66,64 @@ docaTimestamp() noexcept {
     return ts;
 }
 
+[[nodiscard]] bool
```
ColinNV: Nitpick: Can be `constexpr`.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -247,49 +301,22 @@ nixl_status_t
...
-            if (result != DOCA_SUCCESS) {
-                NIXL_ERROR << "Failed to add gauge: " << result;
-                return NIXL_ERR_UNKNOWN;
-            }
-            break;
         }
-        case nixl_telemetry_category_t::NIXL_TELEMETRY_MEMORY: {
-            const char *label_values[] = {docaExporterMemoryCategory, agent_name_.c_str()};
+
+        if (isGaugeEvent(event.eventType_)) {
```
ColinNV: `else if` because counter and gauge are mutually exclusive?

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: fixed.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: fixed.

src/core/telemetry/telemetry_event.h
```diff
@@ -18,30 +18,14 @@
...
 #include "nixl_types.h"
 
 constexpr char TELEMETRY_BUFFER_SIZE_VAR[] = "NIXL_TELEMETRY_BUFFER_SIZE";
 constexpr char TELEMETRY_RUN_INTERVAL_VAR[] = "NIXL_TELEMETRY_RUN_INTERVAL";
 
-constexpr inline int TELEMETRY_VERSION = 2;
-
-/**
- * @enum nixl_telemetry_category_t
- * @brief An enumeration of main telemetry event categories for easy filtering and aggregation
```
guy-ealey-morag, line 33: Wouldn't removing the categories make filtering harder to do?
For example filter for all memory operations, all errors, etc.

e-eygin: Yes, category-level filtering becomes less direct, but the category was already fully derivable from `eventType_`, so keeping it in every event duplicated state.

For the current built-in events, filtering can still be done by event-type family:

- errors: `AGENT_ERR_*` / `agent_err_*`
- transfer: `AGENT_TX_*` and `AGENT_RX_*`
- memory: `AGENT_MEMORY_*`
- timing: `AGENT_XFER_*`

The goal of this PR is to avoid carrying and exporting redundant information on every event/sample. If we later want convenient category-style filtering, I think the better approach is to add a computed helper such as `telemetryEventTypeCategory(eventType)` or `telemetryEventTypeIsError(eventType)`, rather than storing another field in the event wire format.

guy-ealey-morag: In this case I think it's best to document the different prefixes events should have according to their categories (where the categories table was documented before)

guy-ealey-morag: Also, I think that the prefix for timing events is not indicative, I would rename `agent_xfer_time` to `agent_time_xfer` and `agent_xfer_post_time` to `agent_time_post` (or similar)

e-eygin: Good point. I added an `Event Type Naming` section in `docs/telemetry.md` where the old category table used to be.

e-eygin: I agree the timing prefix can be improved, but I’d keep that out of this PR. Renaming `agent_xfer_time` / `agent_xfer_post_time` changes exported metric names and can break downstream dashboards/scrapers, so it should be handled as a separate metrics-naming PR with docs and migration notes. This PR only removes the redundant stored category field and documents current event-type prefix filtering.

e-eygin: @AlexNvd please comment

AlexNvd: > Also, I think that the prefix for timing events is not indicative, I would rename `agent_xfer_time` to `agent_time_xfer` and `agent_xfer_post_time` to `agent_time_post` (or similar)

Guy, the names, indeed, were set much time ago. As they are already in use, we must be very careful when making changes affecting customers.

e-eygin:
@brminich please review

ColinNV (approved):

src/api/cpp/backend/backend_engine.h
```diff
@@ -63,11 +63,14 @@ class nixlBackendEngine {
...
+            if (!enableTelemetry_) {
+                return;
+            }
             std::lock_guard<std::mutex> lock(telemetryEventsMutex_);
-            telemetryEvents_.emplace_back(
-                nixl_telemetry_category_t::NIXL_TELEMETRY_BACKEND, event_type, value);
+            if (telemetryEvents_.size() >= MAX_TELEMETRY_QUEUE_SIZE) {
+                return;
+            }
+            telemetryEvents_.emplace_back(event_type, value);
```
iyastreb, line 73: Since we don't care about individual samples, and just store the resulting counters, why do we use vector at all?
We can simply use an array of atomic pointers maybe:
`std::array<std::atomic<uint64_t>, kNumTelemetryEvents> counters_;`

This way we get:
- no limit on queue size
- no need to lock, just increment atomic counter with relaxed ordering
- no memory allocation/free

src/core/telemetry/telemetry.cpp
```diff
@@ -201,66 +201,52 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
...
-                          nixl_telemetry_category_t category,
-                          uint64_t value) {
+nixlTelemetry::updateData(nixl_telemetry_event_type_t event_type, uint64_t value) {
     // agent can be multi-threaded
     std::lock_guard<std::mutex> lock(mutex_);
     if (events_.size() >= maxBufferedEvents_) {
         return;
     }
-    events_.emplace_back(category, event_type, value);
+    events_.emplace_back(event_type, value);
```
iyastreb, line 210: If we keep events in a vector without timestamps, probably we don't need ordering at all?
In that case one alternative would be to store just an array of atomic counters
No lock, no max queue size, no memory alloc

src/api/cpp/backend/backend_engine.h
AlexNvd: Currently, the default exporter (used, for example for testing) does expose individual events rather than counters.

In general, your suggestion is reasonable, and we already consider a wider change (that will break this backward compatibility) - but it was not approved for development yet.

src/core/telemetry/telemetry.cpp
AlexNvd: We do need ordering, as for a request there are, at least, two events: one created in the beginning and the other at the end. These may need to be correlated.

As I wrote earlier, if a decision is made to break backward compatibility, there is an arch change that will remove the event queues altogether.

src/api/cpp/backend/backend_engine.h
iyastreb: This PR already introduces quite a major change to telemetry and updates TELEMETRY_VERSION, so I think this is a right moment to have this change too.

AlexNvd: Regarding this PR, let's keep its scope to the specific change.

As for the "major change" - @yosefe  , following our meeting, please consider also Ilya's suggestions towards your decision of refactoring Telemetry to remove events and collect data using counters.

e-eygin: This is out of scope for this PR. The PR is a focused cleanup that removes the redundant `category_` field and updates the directly affected exporters, docs, and tests.

Changing backend telemetry storage from an event vector to atomic counters would be a separate telemetry redesign with different semantics, so I don’t want to mix it into this cleanup.

src/core/telemetry/telemetry.cpp
e-eygin: This is a larger telemetry storage redesign and I’d keep it out of this PR.

Even without timestamps, the current telemetry layer is still an event stream: the buffer exporter preserves per-event samples in insertion order, and some metrics use last-event value semantics. Replacing the vector with atomic counters would change that model from event collection to pre-aggregation.

This PR is only removing the redundant `category_` field and updating the directly affected code/docs/tests.

iyastreb (approved):

brminich (approved):

PR #1658, opened 2026-05-20
nixlbench: derive storage taxonomy from supported mem types

benlwalker:
## What?
Re-implement isStorageBackend() and isObjStorageBackend() by querying supported memory types rather than a hard coded list of backend names.

## Why?
Long term, we'd like to not require nixlbench changes each time a new backend is added. This is one small piece of that puzzle.

## How?
isStorageBackend() and isObjStorageBackend() compared the configured backend string against a hardcoded set of names, and nixl_worker.cpp maintained a separate allowlist of "supported" backend strings. Adding a new backend or registering an existing plugin under a different name required touching both lists in the bench.

Cache nixl_mem_list_t for the configured backend in xferBenchConfig. Populate it pre-instantiation via a transient discovery nixlAgent and getPluginParams (so the predicate works before the worker exists, for the runtime / role / metadata-exchange gates), then refresh it from getBackendParams after createBackend so plugins that add mem types at runtime (LIBFABRIC adds VRAM_SEG when CUDA HMEM is detected) are reflected. Rewrite the predicates against the cached mem set and add isBlkStorageBackend().

## Summary by CodeRabbit

* **Refactor**
  * Added runtime discovery and caching of backend memory-segment capabilities (via NIXL plugin discovery) so supported storage segment types are detected automatically.
  * Backend capability checks now rely on discovered segment types instead of hard-coded backend identifiers.
  * Initialization validates discovered capabilities and fails early if none are found; capabilities are refreshed after backend creation.
  * Added explicit block-storage detection and tightened readback/consistency logic to only treat true storage-capable backends as storage-backed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1658?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -35,6 +35,7 @@
 #include <fcntl.h>
 #include <filesystem>
 #include <gflags/gflags.h>
+#include <nixl.h>
 
 #include "runtime/etcd/etcd_rt.h"
 #include "utils/neuron.h"
@@ -312,6 +313,28 @@ int xferBenchConfig::gusli_max_simultaneous_requests = 0;
 std::string xferBenchConfig::gusli_config_file = "";
 std::string xferBenchConfig::gusli_device_byte_offsets = "";
 std::string xferBenchConfig::gusli_device_security = "";
+nixl_mem_list_t xferBenchConfig::backend_mems = {};
+
+namespace {
+// Discover the memory segment types supported by the configured backend plugin.
+// Uses a transient nixlAgent to call getPluginParams; the plugin manager is a
+// singleton so this is cheap and does not affect the real agent created later
+// by the worker.
+nixl_mem_list_t
+discoverPluginMems(const std::string &backend) {
+    nixl_mem_list_t mems;
+    if (backend.empty()) {
+        return mems;
+    }
+    nixlAgent discovery_agent("nixlbench_discovery_" + std::to_string(::getpid()),
+                              nixlAgentConfig());
+    nixl_b_params_t params;
+    if (discovery_agent.getPluginParams(backend, mems, params) != NIXL_SUCCESS) {
+        mems.clear();
+    }
+    return mems;
+}
+} // namespace
 
 int
 xferBenchConfig::parseConfig(int argc, char *argv[]) {
@@ -378,6 +401,9 @@ xferBenchConfig::loadParams(void) {
     // Only load NIXL-specific configurations if using NIXL worker
     if (worker_type == XFERBENCH_WORKER_NIXL) {
         backend = NB_ARG(backend);
+        // Discover supported memory types so the storage/object/block predicates work
+        // before the real agent is created (the worker later refreshes from the engine).
+        backend_mems = discoverPluginMems(backend);
         enable_pt = NB_ARG(enable_pt);
         progress_threads = NB_ARG(progress_threads);
         device_list = NB_ARG(device_list);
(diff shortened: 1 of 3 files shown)


benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -744,20 +777,27 @@ xferBenchConfig::parseDeviceList() {
...
+    return false;
+}
+
+bool
+xferBenchConfig::isBlkStorageBackend() {
+    for (auto m : backend_mems) {
+        if (m == BLK_SEG) return true;
+    }
+    return false;
+}
```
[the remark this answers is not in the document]
benlwalker: Done

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -926,8 +972,7 @@ xferBenchUtils::checkConsistency(std::vector<std::vector<xferBenchIOV>> &iov_lis
 
             len = iov.len;
 
-            if (xferBenchConfig::isStorageBackend() ||
-                xferBenchConfig::backend == XFERBENCH_BACKEND_GPUNETIO) {
+            if (xferBenchConfig::isStorageBackend()) {
```
ofer, line 1035: GPUNETIO is not a storage backend, why can you remove it from the exceptional case?

brminich: @benlwalker?

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -300,6 +301,35 @@ int xferBenchConfig::gusli_max_simultaneous_requests = 0;
...
+    }
+    try {
+        nixlAgent discovery_agent("nixlbench_discovery_" + std::to_string(::getpid()),
+                                  nixlAgentConfig());
+        nixl_b_params_t params;
+        if (discovery_agent.getPluginParams(backend, mems, params) != NIXL_SUCCESS) {
+            mems.clear();
+        }
+    }
+    catch (const std::exception &e) {
```
iyastreb: what kind of exceptions is it expected to catch? I don't see getPluginParams throws user exceptions
getPluginParams returns error code, so it must not throw exceptions
We should handle either error code or expection, but not both

And catching runtime exceptions (std::bad_alloc etc) does not make much sense here - the same way we can put try-catch in each and every function, which just adds a runtime cost for no benefit. IMO better let them propagate

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -776,20 +809,33 @@ xferBenchConfig::parseDeviceList() {
...
+        if (m == OBJ_SEG) {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool
+xferBenchConfig::isBlkStorageBackend() {
+    for (auto m : backend_mems) {
```
iyastreb: Maybe more concise lookup? And for other functions too
```
    return std::any_of(backend_mems.begin(), backend_mems.end(),
                       [](nixl_mem_t m) { return m == BLK_SEG; });
```

benchmark/nixlbench/src/utils/utils.cpp
benlwalker: GPUNETIO reports {DRAM_SEG, VRAM_SEG}, so isStorageBackend() is false for it and it now flows through the network branch, which handles every op/seg combination. The previous special case forced it into the storage branch, whose WRITE path calls pread(iov.devId, ...) — correct when devId is a file fd, but undefined for GPUNETIO where devId is a CUDA device id. Removing it is a correctness fix.

benchmark/nixlbench/src/utils/utils.cpp
benlwalker: Done

benchmark/nixlbench/src/utils/utils.cpp
benlwalker: Done

iyastreb (dismissed):

iyastreb (approved):

PR #1662, opened 2026-05-21
Log AM error

iyastreb:
## What?
https://nvbugspro.nvidia.com/bug/6114610
Log error when AM fails

nixlUcxEngine::genNotif is a fire-and-forget API by construction

Making nixlUcxEngine::genNotif synchronous would require blocking the calling thread on UCX worker progress until the active message completes, which is not acceptable on the notification fast path and would also need a non-trivial refactor of the engine's threading model. Since genNotif returns no caller-visible request handle, there is no channel left to propagate an asynchronous send failure back to the caller after the function returns NIXL_SUCCESS. As a best-effort first fix, we can log the failure status in the AM completion callback, so transport errors such as UCS_ERR_ENDPOINT_TIMEOUT become visible in NIXL logs instead of being silently dropped. This does not prevent the hang on the CTX side, but it gives operators a clear signal of which notifications were lost and why.

## Why?
Improve diagnostics on the caller side

## Summary by CodeRabbit

* **Chores**
  * Enhanced error reporting for network communication failures with detailed diagnostic information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1662?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -200,6 +200,10 @@ using nixl_ucx_am_cb_ctx_ptr_t = std::unique_ptr<nixl_ucx_am_cb_ctx_t>;
 
 void
 nixlUcxEp::sendAmCallback(void *request, ucs_status_t status, void *user_data) {
+    if (status != UCS_OK) {
+        NIXL_ERROR << "UCX AM send failed with status " << status << " ("
+                   << ucs_status_string(status) << ")";
+    }
     auto ctx = static_cast<nixl_ucx_am_cb_ctx_t *>(user_data);
     ctx->second(request, ctx->first);
     delete ctx;


brminich (approved):

ColinNV (approved):

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -200,6 +200,10 @@ using nixl_ucx_am_cb_ctx_ptr_t = std::unique_ptr<nixl_ucx_am_cb_ctx_t>;
 
 void
 nixlUcxEp::sendAmCallback(void *request, ucs_status_t status, void *user_data) {
+    if (status != UCS_OK) {
+        NIXL_ERROR << "UCX AM send failed with status " << status << " ("
```
rakhmets, line 204: I think it should be warning. As you mentioned in the description we cannot return any status. So the client's code cannot handle this error.
This case is unexpected, but we continue to execute the code. Thus, it should be a warning.

iyastreb: Well, if we reach this point, then the client application hangs and can't recover from that error.
Imo it's 100% error, if it happens we can't progress and app needs to be restarted with another config

rakhmets: I understand your point. But this conclusion comes from particular scenario and use case. However, from the perspective of this part of code it doesn't seem to be correct.
My point is that we should print an error only if client code can react in any way. All other unexpected cases should be printed as warnings. And warnings are also always visible. Thus it will have the same effect ("Improve diagnostics on the caller side"), as the error.

rakhmets: The reason I want this to be a warning is so that we follow simple and clear rules when using logging.
Unexpected behavior and client code can react based on returned status or raised exception, then log an error. Otherwise, e.g. during the destruction of resources, issue a warning. So that we can avoid situation where developer has to think about how critical the case is when choosing the right logging level.

iyastreb: I still believe this should be an error, because this is a critical error that client app can't handle or recover from.
But I let @brminich to decide

brminich: imo the severity corresponds to error, but it seems to be not that easy to pass some status to the user

PR #1678, opened 2026-05-25
[WiP] Send Recv API and UCX implementation

ColinNV:
## What?
This PR implements the new send/recv API functions on the NIXL agent and in the UCX backend.

## Done
- Extend operations from `read`/`write` to `read`/`write`/`send`/`recv`.
- New agent function `createTagXferReq()`.
- New backend function `supportsSendRecv()`.
- New backend function `prepTagXfer()` for `send`/`recv`.
- New backend function `postTagXfer()` for `send`/`recv`.
- UCX backend implementation of new backend functions.
- Agent-level smoke test.
- UCX backend test for good cases.
- Python bindings and basic tests.

## Todo
- Cost estimation functions **do not** work with `send`/`recv`.
- Rust bindings.
- Agent-level tests, more UCX backend tests, telemetry tests.
- Audit object lifecycle handling in all good and bad cases.
- Reviews and a cleanup.

The UCX backend implementation handles send-before-recv and recv-before-send, it can queue multiple requests for the same remote-agent/tag combination, and it handles the multi-descriptor case; however operation boundaries are not yet enforced, e.g. if the initiator posts a send for size 100, the target posts a recv for sizes 100 and 200, and the initiator posts another send for size 200, all with the same tag, then the operation(s) will complete successfully.


src/core/telemetry/telemetry.cpp
```diff
@@ -264,11 +265,14 @@ nixlTelemetry::updateMemoryDeregistered(uint64_t memory_deregistered) {
 }
 
 void
-nixlTelemetry::addXferTime(std::chrono::microseconds xfer_time, bool is_write, uint64_t bytes) {
-    const auto bytes_type = is_write ? nixl_telemetry_event_type_t::AGENT_TX_BYTES :
-                                       nixl_telemetry_event_type_t::AGENT_RX_BYTES;
-    const auto requests_type = is_write ? nixl_telemetry_event_type_t::AGENT_TX_REQUESTS_NUM :
-                                          nixl_telemetry_event_type_t::AGENT_RX_REQUESTS_NUM;
+nixlTelemetry::addXferTime(std::chrono::microseconds xfer_time,
+                           const nixl_xfer_op_t operation,
```
mkhazraee, line 294: I really like passing the full op, just back then we were trying to really optimize the perf, hence passed a single flag. Just let's make sure we don't add considerable overhead to the telemetry.

src/plugins/ucx/ucx_backend.h
```diff
@@ -280,6 +303,34 @@ class nixlUcxEngine : public nixlBackendEngine {
                        size_t start_idx,
                        size_t end_idx);
 
+    nixl_status_t
+    prepTagSend(const nixl_meta_dlist_t &local,
```
mkhazraee, line 307: Should we use ucx utils instead of the main ucx_backend header? Maybe not practical, just this is getting too long, few extra functions relative to backend are fine, but maybe better not to make this too different than the base SB API and use utils for that.

src/plugins/azure_blob/azure_blob_backend.cpp
```diff
@@ -43,10 +44,7 @@ isValidPrepXferParams(const nixl_xfer_op_t &operation,
                       const nixl_meta_dlist_t &remote,
                       const std::string &remote_agent,
                       const std::string &local_agent) {
-    if (operation != NIXL_WRITE && operation != NIXL_READ) {
-        NIXL_ERROR << absl::StrFormat("Error: Invalid operation type: %d", operation);
-        return false;
-    }
+    NIXL_ASSERT(nixl::isReadWrite(operation));
```
mkhazraee, line 48: Didn't we remove asserts in release mode? This might cause buggy behavior, it's a dynamic case, and user might call incorrectly and cause errors. Or there is another part that catches it if assert parts are compiled out?

src/api/cpp/backend/backend_engine.h
```diff
@@ -108,10 +108,11 @@ class nixlBackendEngine {
         // pure virtual, and return errors, as parent shouldn't call if supportsNotif is false.
         virtual bool supportsNotif() const = 0;
 
-    // Determine whether a backend supports send and recv operations
-    virtual bool supportsSendRecv() const {
-        return false;
-    }
+        // Determine whether a backend supports send and recv operations
+        virtual bool
+        supportsSendRecv() const {
```
mkhazraee, line 116: As we discussed, let's add a check for it that it can only be true if supportsRemote is also true. Same way we have a supportsRemote requires supportsNotif.

src/core/transfer_request.h
```diff
@@ -59,23 +61,47 @@ class nixlXferReqH {
...
-    const nixl_xfer_op_t backendOp;
-    nixl_status_t status = NIXL_ERR_NOT_POSTED;
+class nixlXferReqSR : public nixlXferReqH {
+public:
+    const std::string tag;
 
-    nixl_xfer_telem_t telemetry;
+    nixlXferReqSR(const std::string &remote_agent,
+                  const nixl_xfer_op_t operation,
+                  const std::string &tag,
```
itayalroy, line 97: Tags are somewhat of an advanced use case, some existing send/recv users (EPLB in vLLM for instance) don’t use them, I think we should consider making tags optional or providing a default tag.

mkhazraee: Disagreed. Without tag or a gather all means staging, we kept the notion of zero copy of nixl for send/recv too. Idea being we simplify the process of doing the exchange, and then do a write, to each side just doing the send or recv. But the core idea of zero copy shouldn't be changed. Unless you have another idea how to achieve the wildcard style.

src/api/cpp/backend/backend_engine.h
ColinNV: Done.

src/plugins/azure_blob/azure_blob_backend.cpp
ColinNV: This is just a "paranoia level" check, the Agent should never call this function with a different operation.

src/plugins/ucx/ucx_backend.h
ColinNV: I agree that some files here are now too large, but I'm already shuffling a lot around and would like keep additional refactorings separate.

src/core/telemetry/telemetry.cpp
ColinNV: I don't think the single flag was an optimization.

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27012968618/job/79720851148)…

dpressle:
👀 Investigating [`Run Pre-Commit Hooks`](https://github.com/ai-dynamo/nixl/actions/runs/27012968618/job/79720851148)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `253b8322`

I have enough information from the logs. The full diff in the CI log is conclusive. Here is the complete diagnosis:

---

**Summary:** `black` formatter failed on `test/python/test_nixl_api.py` — multi-line call expressions were formatted in a style that `black` collapses to single lines.

**Root cause:** Commit `[REDACTED:Hex High Entropy String]` (branch `send_recv`, PR #1678) introduced a new test function `test_tag_xfer_send_recv` in `test/python/test_nixl_api.py` with function calls hand-wrapped across multiple lines in a way that does **not** conform to `black`'s formatting rules. Specifically, two patterns were written in expanded form when `black` requires them on a single line (because each call fits within the 88-character line-length limit):

```python
# As committed (black rejects):
send_descs = agent1.get_xfer_descs(
    [(addr1, mem_size, 0)], mem_type="DRAM"
)
recv_descs = agent2.get_xfer_descs(
    [(addr2, mem_size, 0)], mem_type="DRAM"
)
recv_handle = agent2.initialize_tag_xfer(
    "RECV", recv_descs, tag, agent1.name
)
send_handle = agent1.initialize_tag_xfer(
    "SEND", send_descs, tag, agent2.name
)

# What black reformats them to (correct):
send_descs = agent1.get_xfer_descs([(addr1, mem_size, 0)], mem_type="DRAM")
recv_descs = agent2.get_xfer_descs([(addr2, mem_size, 0)], mem_type="DRAM")
recv_handle = agent2.initialize_tag_xfer("RECV", recv_descs, tag, agent1.name)
send_handle = agent1.initialize_tag_xfer("SEND", send_descs, tag, agent2.name)
```

All other hooks (`mypy`, `isort`, `flake8`, `codespell`, etc.) passed — only `black` failed, and it modified exactly 1 file.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `send_recv`, PR #1678)

**File:** `test/python/test_nixl_api.py` — lines around `test_tag_xfer_send_recv` (diff hunk at `@@ -215,20 +215,12 @@`)

**Suggested fix:** Apply `black`'s reformatting locally before pushing. The author simply needs to run:
```bash
black test/python/test_nixl_api.py
```
...and then commit the result. The exact 4 function calls need to be collapsed onto single lines as shown in the diff above. To avoid this in the future, the author can install the pre-commit hook locally with `pre-commit install` so `black` runs automatically before every commit.

**Related:** PR #1678 (`send_recv` branch) — no pre-existing issues found for this formatting violation.



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `d1aad676-0e74-4630-b920-a6135d62c201` in the triage console for the audit trail.

dpressle:
🤖 **CI Triage Agent** — `Run Pre-Commit Hooks` · commit `253b8322`

The logs provide a clear and complete picture of the failure. Here is the full diagnosis:

---

**Summary:** `black` formatter reformatted `test/python/test_nixl_api.py` — the PR was submitted without running `black` first.

**Root cause:** The `black` pre-commit hook **failed** because `test/python/test_nixl_api.py` was not formatted according to `black`'s style. Specifically, several multi-line function calls in the `test_tag_xfer_send_recv` function were broken across lines in a way that `black` disagrees with. Black collapsed them into single lines:

- `agent1.get_xfer_descs([(addr1, mem_size, 0)], mem_type="DRAM")` (was split over 3 lines)
- `agent2.get_xfer_descs([(addr2, mem_size, 0)], mem_type="DRAM")` (was split over 3 lines)
- `agent2.initialize_tag_xfer("RECV", recv_descs, tag, agent1.name)` (was split over 3 lines)
- `agent1.initialize_tag_xfer("SEND", send_descs, tag, agent2.name)` (was split over 3 lines)

All other hooks (`mypy`, `isort`, `flake8`, `codespell`, and all general checks) **passed**.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1678 author)

**File:** `test/python/test_nixl_api.py` — around lines 218–228 (the `test_tag_xfer_send_recv` function, in the area that adds and transfers descriptors)

**Suggested fix:** Run `black` locally on the file before pushing:
```bash
black test/python/test_nixl_api.py
```
Then commit the reformatted file and push again. The diff black applied is exactly what needs to be committed — those 8 lines of multi-line call syntax need to be collapsed to 4 single-line calls as shown in the CI diff. You can also install the pre-commit hooks locally (`pre-commit install`) to catch this automatically on every future commit.

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `c80ea586-b779-4ccc-8bf6-3ab9b7847a37` in the triage console for the audit trail.

src/api/cpp/backend/backend_engine.h
```diff
@@ -149,6 +155,28 @@ class nixlBackendEngine {
...
+                    const nixl_opt_b_args_t *opt_args = nullptr) const {
+            return NIXL_ERR_BACKEND;
+        }
+
+        // Posting a request, which completes the async handle creation and posts it
+        virtual nixl_status_t
+        postTagXfer(nixl_xfer_op_t operation,
+                    const nixl_meta_dlist_t &local,
+                    const std::string &tag,
+                    const std::string &remote_agent,
```
rakhmets, line 174: Why do we need these parameters in both prep and post functions?
Can we prepare smth for one operation, but post for another? The same question about the metadata. 
Do we need tag, remote_agent at prep?

ColinNV: I'm mirroring `prepXfer()` / `postXfer()` with my assumptions being that (a) it depends on the backend which of the functions really needs the parameters, and (b) since `nixlXferReqRW` and `nixlXferReqSR` already store these values it's prudent to supply them to the backend in both calls.

(If they were only supplied in `prepTagXfer()` the backend would have to store anything it also needs in `postTagXfer()`.)

ColinNV:
> 👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27012968618/job/79720851148)…

@dpressle These are creating a lot of noise for trivial formatting issues where the fix can simply be taken from the logs?

PR #1684, opened 2026-05-27
EXAMPLES/DEVICE/EP: Removed method w/o definition.

rakhmets:
## What?
Deleted unused `clean_buffer` method that is declared but not defined.

## Summary by CodeRabbit

* **Chores**
  * Removed internal buffer cleaning utility declarations and implementations across kernel, C++ header, and Python wrapper layers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1684?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/device/ep/csrc/kernels/api.cuh
+++ b/examples/device/ep/csrc/kernels/api.cuh
@@ -211,10 +211,6 @@ void combine(cudaDataType_t type,
 
 // EP kernels
 namespace ep_kernels {
-void clean_buffer(int* clean_0, int num_clean_int_0,
-                              int* clean_1, int num_clean_int_1,
-                              int rank, int num_ranks, int* mask_buffer, int* sync_buffer,
-                              cudaStream_t stream);
 
 void dispatch(void* packed_recv_x, void* packed_recv_x_scales,
               int* packed_recv_src_info, int64_t* packed_recv_layout_range,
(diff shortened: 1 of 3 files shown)


ebarilanM:
@itayalroy do we know that this API not called in VLLM/SGLANG?

ebarilanM (commented):
@itayalroy what was the motivation ? is it safe also remove the buffer api?

rakhmets:
> @itayalroy do we know that this API not called in VLLM/SGLANG?

> @itayalroy what was the motivation ? is it safe also remove the buffer api?

The method doesn't have a definition. It's just a declaration. There is no binding for the method. Thus, it cannot be called by any framework.

rakhmets:
> > @itayalroy do we know that this API not called in VLLM/SGLANG?
> 
> > @itayalroy what was the motivation ? is it safe also remove the buffer api?
> 
> The method doesn't have a definition. It's just a declaration. There is no binding for the method. Thus, it cannot be called by any framework.

Calling the method raises `AttributeError`.

itayalroy (approved):
@ebarilanM this is a leftover from DeepEP, they used it to support low_latency_dispatch/combine() after ht_dispatch/combine() since ht does not clean low latency counters. We don't support ht + ll on the same buffer anyways

ebarilanM (approved):

guy-ealey-morag (approved):

iyastreb (approved):

PR #1685, opened 2026-05-27
CI: cap port pool below kernel ephemeral range

Alexey-Rivkin:
## What?
Cap the CI port pool below Linux ephemeral range, and modulo the concurrent-id sources by the available slot count.

## Why?
NIXL CI flakes with `Failed to bind metadata listener socket` on both platforms:
- Jenkins: https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/1850/pipeline/261/
- GitHub Actions (AWS EFA): https://github.com/ai-dynamo/nixl/actions/runs/26509466812/attempts/2?pr=1658

The CI port pool overlaps Linux default `ip_local_port_range` (32768-60999), so the kernel ephemeral allocator can grab a port in our pool between the listener probe in `get_next_tcp_port` and the test's `bind()`.

## How?
Capping at 32500 keeps every bucket below the ephemeral floor. Modulo on the id sources prevents high values (observed 20, 27, 46, 53 on Jenkins) from landing above the cap.

## Summary by CodeRabbit

* **Chores**
  * Reduced the TCP port allocation range used by CI to limit available ports and adjust concurrency-based port selection.
  * Revised CI concurrency handling so concurrent IDs are normalized before selecting ports to avoid collisions.
  * Updated CI pipeline image tags used for build and test jobs to newer versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/nixl/pull/1685?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.ci/scripts/common.sh
+++ b/.ci/scripts/common.sh
@@ -22,7 +22,7 @@
 # Set initial port number for client/server applications to be updated with
 # function below
 #
-tcp_port_range=1000
+tcp_port_range=300
 min_port_number=10500
 max_port_number=32500
 
--- a/.ci/scripts/common.sh
+++ b/.ci/scripts/common.sh
@@ -24,18 +24,10 @@
 #
 tcp_port_range=1000
 min_port_number=10500
-max_port_number=65535
+max_port_number=32500
 
-# GITLAB CI
-if [ -n "$CI_CONCURRENT_ID" ]; then
-    nixl_concurrent_id=$CI_CONCURRENT_ID
-# Jenkins CI
-elif [ -n "$EXECUTOR_NUMBER" ]; then
-    nixl_concurrent_id=$EXECUTOR_NUMBER
-else
-    # Fallback to random number if both CI_CONCURRENT_ID and EXECUTOR_NUMBER are not set
-    nixl_concurrent_id=$((RANDOM % $(((max_port_number - min_port_number) / tcp_port_range))))
-fi
+tcp_port_slots=$(((max_port_number - min_port_number) / tcp_port_range))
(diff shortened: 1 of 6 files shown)


ovidiusm (dismissed):

brminich (dismissed):

.ci/scripts/common.sh
```diff
@@ -24,18 +24,10 @@
...
-if [ -n "$CI_CONCURRENT_ID" ]; then
-    nixl_concurrent_id=$CI_CONCURRENT_ID
-# Jenkins CI
-elif [ -n "$EXECUTOR_NUMBER" ]; then
-    nixl_concurrent_id=$EXECUTOR_NUMBER
-else
-    # Fallback to random number if both CI_CONCURRENT_ID and EXECUTOR_NUMBER are not set
-    nixl_concurrent_id=$((RANDOM % $(((max_port_number - min_port_number) / tcp_port_range))))
-fi
+max_port_number=32500
```
roiedanino, line 27: (32500 - 10500) / 1000 = 22
This means we will have collisions when more than 22 executors are running simultaneously
Is that safe?

I would suggest changing tcp_port_range to 300, which should be more than enough per runner, and give us a bigger headroom, and help us avoid collisions

Alexey-Rivkin: right call!
3ca09d4ed8d2194cabc7cb4b578119eb395ccd2f

roiedanino (dismissed):

ovidiusm (approved):

PR #1689, opened 2026-05-28
refactor(nixlbench): use RAII for memory deregistration and fd cleanup

oliviergaraud:
## What?
Replace explicit deregister/cleanup loops in `deallocateMemory()` with RAII wrappers
that automatically handle resource teardown.

- `xferFileState` now owns its fd (close-on-destroy, move-only)
- New `NixlMemRegion` class wraps registered memory regions and calls `deregisterMem` + per-IOV cleanup in its destructor
- `deallocateMemory()` just clears the RAII containers
- `createFileFds` error paths simplified (RAII closes open fds automatically)

## Why?
From this discussion: https://github.com/ai-dynamo/nixl/pull/1590#discussion_r3315879165
Resources (registered memory regions, file descriptors) were managed through manual cleanup calls scattered across `deallocateMemory()`, with backend-specific branching and fd deduplication logic.
This was fragile in the face of exceptions and required careful ordering that was easy to get wrong.

## How?
Two RAII types handle the two resource categories:

- **`xferFileState`**: move-only struct that calls `::close(fd)` in its destructor. Eliminates the `unordered_set<int> closed_fds` dedup pattern, each `xferFileState` owns exactly one fd, so `remote_fds.clear()` closes them all exactly once.
 - **`NixlMemRegion`**: owns a registered IOV list along with an agent pointer, backend handle, segment type, and an optional per-IOV cleanup function.  `release()` calls `deregisterMem` then invokes the cleanup on each IOV.  Stored as class members (`remote_regs_`, `local_regs_`), cleared in order in `deallocateMemory()` to preserve the remote-before-local invariant.

The virtual interface (`allocateMemory` / `deallocateMemory`) is unchanged. `allocateMemory` still returns `vector<vector<xferBenchIOV>>` and additionally populates the internal RAII guards; `deallocateMemory` ignores its parameter and clears the guards.

## Summary by CodeRabbit

* **Refactor**
  * Memory management moved to RAII-style region objects for automatic, safer cleanup and simpler bookkeeping.
  * Local and remote registrations consolidated into moveable region entries, reducing manual deregistration paths.
  * File-descriptor ownership converted to move-only RAII to simplify lifecycle handling.
* **Bug Fixes**
  * File-open failures now return immediately to avoid inconsistent state.
  * Deregistration and cleanup made more reliable to prevent resource leaks and double-cleanup.
* **Chores**
  * Build sources updated to include the new RAII region implementation.


iyastreb:
will review once https://github.com/ai-dynamo/nixl/pull/1590 gets merged

benchmark/nixlbench/src/worker/nixl/nixl_worker.h
```diff
@@ -19,20 +19,95 @@
...
+        }
+        return *this;
+    }
+
+    xferFileState(const xferFileState &) = delete;
+    xferFileState &
+    operator=(const xferFileState &) = delete;
+};
+
+class NixlMemRegion {
```
rakhmets: I suggest to place new classes/structs to separate files.

benchmark/nixlbench/src/worker/nixl/nixl_worker.h
```diff
@@ -19,20 +19,95 @@
...
-    int fd;
-    uint64_t file_size;
-    uint64_t offset;
+    int fd = -1;
+    uint64_t file_size = 0;
+    uint64_t offset = 0;
+
+    xferFileState() = default;
+
+    xferFileState(int fd, uint64_t file_size, uint64_t offset)
```
rakhmets: noexcept?

benchmark/nixlbench/src/worker/nixl/nixl_worker.h
```diff
@@ -19,20 +19,95 @@
...
+        : fd(std::exchange(o.fd, -1)),
+          file_size(o.file_size),
+          offset(o.offset) {}
+
+    xferFileState &
+    operator=(xferFileState &&o) noexcept {
+        if (this != &o) {
+            if (fd >= 0) {
+                ::close(fd);
+            }
```
rakhmets: Please put it to a private method to avoid code duplication.

benchmark/nixlbench/src/worker/nixl/nixl_worker.h
```diff
@@ -19,20 +19,95 @@
...
+        return iovs_;
+    }
+
+    std::vector<xferBenchIOV> &
+    iovs() {
+        return iovs_;
+    }
+
+    void
+    release();
```
rakhmets: This method should be private.

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
```diff
@@ -0,0 +1,139 @@
...
+        if (fd >= 0) {
+            ::close(fd);
+        }
+    }
+};
+
+// RAII wrapper around a NIXL memory registration: deregisters the memory and
+// runs the per-IOV cleanup on destruction.
+class NixlMemRegion {
+    nixlAgent *agent_ = nullptr;
```
iyastreb: Why to allow null agent?
I think this field should be ref

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
```diff
@@ -0,0 +1,139 @@
...
+            std::cerr << "NIXL: " << message << " (" << nixlEnumStrings::statusStr(_r) << ")" \
+                      << std::endl;                                                           \
+            exit(EXIT_FAILURE);                                                               \
+        }                                                                                     \
+    } while (0)
+
+// Convert vector of xferBenchIOV to nixl_reg_dlist_t
+inline void
+iovListToNixlRegDlist(const std::vector<xferBenchIOV> &iov_list, nixl_reg_dlist_t &dlist) {
+    nixlBlobDesc desc;
```
iyastreb: maybe resize/reserve dlist in advance to avoid extra re-allocations

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
```diff
@@ -0,0 +1,139 @@
...
+        const nixl_status_t _r = (result);                                                    \
+        if (0 != _r) {                                                                        \
+            std::cerr << "NIXL: " << message << " (" << nixlEnumStrings::statusStr(_r) << ")" \
+                      << std::endl;                                                           \
+            exit(EXIT_FAILURE);                                                               \
+        }                                                                                     \
+    } while (0)
+
+// Convert vector of xferBenchIOV to nixl_reg_dlist_t
+inline void
```
iyastreb: more C++ way would be to return nixl_reg_dlist_t from the function, unless you need to merge several iov lists into a single dlist

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
oliviergaraud: indeed

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
oliviergaraud: done

benchmark/nixlbench/src/worker/nixl/nixl_mem_region.h
oliviergaraud: :ok_hand:

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -1115,103 +1087,61 @@ xferBenchNixlWorker::allocateMemory(int num_threads) {
...
             }
         }
 
-        nixl_reg_dlist_t desc_list(seg_type);
-        iovListToNixlRegDlist(iov_list, desc_list);
+        nixl_reg_dlist_t desc_list = iovListToNixlRegDlist(iov_list, seg_type);
         CHECK_NIXL_ERROR(agent->registerMem(desc_list, &opt_args), "registerMem failed");
-        iov_lists.push_back(iov_list);
+
+        std::function<void(xferBenchIOV &)> local_cleanup;
```
iyastreb: This is the only part I'm not comfortable with. Injecting cleanup function explicitly looks a bit weird, maybe we can find a way to set/keep this info during init?

oliviergaraud: Thanks for your useful review @iyastreb 
I hope that's better now

iyastreb (approved):

PR #1690, opened 2026-05-28
feat(core): add opt-in nixlMDManager (PR 1)

aschwartz12:
# Metadata Manager PR1 Summary

Adds an internal, agent-owned `nixlMDManager`, a `nixlMetadataBackend` contract,
and the first P2P backend. The public `nixlAgent` metadata API is unchanged:
callers still use the same `sendLocalMD`, `sendLocalPartialMD`, `fetchRemoteMD`,
and `invalidateLocalMD` calls with `nixl_opt_args_t`.

Before this PR, P2P metadata exchange was implemented inline in `nixlAgent` and
the comm-worker socket path. This PR factors that behavior into a backend class
and lets `NIXL_USE_MD_MANAGER` route P2P metadata calls through the new manager.
The default path remains unchanged when the env var is unset.

## What?

Adds the internal manager/backend layer for metadata exchange and implements the
P2P backend. The manager is owned by `nixlAgentData` and is not part of the
installed public API.

## Why?

The old code keeps metadata transport details inside the agent and comm worker.
This PR creates the backend boundary needed to move those details out while
preserving the existing user-facing API and behavior.

## How?

Manager - `nixlMDManager` routes metadata exchange calls to the active backend.
It is created with the agent data when metadata exchange needs the comm thread.

Backend - `nixlMetadataBackend` defines the transport boundary. The P2P backend
reuses `getLocalMD` / `getLocalPartialMD` for serialization and enqueues
`SOCK_SEND`, `SOCK_FETCH`, or `SOCK_INVAL` work on the existing comm thread.

Routing - With `NIXL_USE_MD_MANAGER` set, P2P calls that include `ipAddr` route
through the manager. Without it, the old inline path runs as before.

Scope - No public API change, no new threads, no new sockets. The PR changes the
internal shape of P2P metadata exchange without changing caller-visible
semantics.


src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,151 @@
...
+    nixlMDManager(nixlMDManager &&) = delete;
+    nixlMDManager(const nixlMDManager &) = delete;
+    nixlMDManager &
+    operator=(nixlMDManager &&) = delete;
+    nixlMDManager &
+    operator=(const nixlMDManager &) = delete;
+
+private:
+    explicit nixlMDManager(nixlAgent &agent) noexcept : agent_(agent) {}
+    friend class nixlAgent;
```
iyastreb: we already have too many friends in NIXL, I would suggest to minimize its usage

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,151 @@
...
+    struct Peer {
+        std::string ip;
+        std::uint16_t port;
+    };
+
+    [[nodiscard]] bool
+    lookupPeer(const std::string &agent_name, Peer &out) const;
+
+    nixlAgent &agent_;
+    mutable std::mutex mu_;
```
iyastreb: mutex_

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,125 @@
...
+    args.ipAddr = ip;
+    args.port = static_cast<int>(port);
+    return args;
+}
+
+} // namespace
+
+bool
+nixlMDManager::lookupPeer(const std::string &agent_name, Peer &out) const {
+    std::lock_guard<std::mutex> lk(mu_);
```
iyastreb: what's nice about nixlData is that nixlLock can be configured to different modes (no lock, rw, strict).
Should we use the same approach here? Or even should we reuse nixlData lock?

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,125 @@
...
+                              std::uint16_t port) {
+    if (agent_name.empty() || ip.empty()) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+    // 0 means "use the listener-side default", matching the existing
+    // nixl_opt_args_t::port default.
+    const std::uint16_t resolved_port =
+        (port == 0) ? static_cast<std::uint16_t>(default_comm_port) : port;
+    std::lock_guard<std::mutex> lk(mu_);
+    peers_[agent_name] = Peer{ip, resolved_port};
```
iyastreb: should we report an error if peer exists already?

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,125 @@
...
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlMDManager::sendLocalMD(const std::string &agent_name) const {
+    Peer peer;
+    if (!lookupPeer(agent_name, peer)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+    const auto args = makePeerArgs(peer.ip, peer.port);
```
iyastreb: there is a locking gap between lookup and send - is it always safe?

src/core/agent_data.h
```diff
@@ -67,6 +68,12 @@ class nixlAgentData {
         const nixlAgentConfig config_;
         const bool useEtcd_;
         const bool needsCommThread_;
+        // Opt-in: gated by the NIXL_MD_MANAGER environment variable. When
+        // false, nixlAgent::getMDManager() reports NIXL_ERR_NOT_SUPPORTED
+        // and no nixlMDManager is ever constructed.
+        const bool mdManagerEnabled_;
+        std::mutex mdManagerMutex_;
```
iyastreb: Maybe we don't need mutex
Just create mdManager in nixlAgent ctor if config is set

src/core/agent_data.h
```diff
@@ -67,6 +68,12 @@ class nixlAgentData {
         const nixlAgentConfig config_;
         const bool useEtcd_;
         const bool needsCommThread_;
+        // Opt-in: gated by the NIXL_MD_MANAGER environment variable. When
+        // false, nixlAgent::getMDManager() reports NIXL_ERR_NOT_SUPPORTED
+        // and no nixlMDManager is ever constructed.
+        const bool mdManagerEnabled_;
```
iyastreb: If we create manager in ctor, then no need to store this field

src/api/cpp/nixl_md_manager.h
aschwartz12: Done. 
Dropped the friend and made the ctor public, so the agent constructs it with make_unique and there's no special access. The ctor still takes a nixlAgent& and grants no capability beyond what the agent already exposes.

src/api/cpp/nixl_md_manager.h
aschwartz12: Renamed to mutex_

src/core/nixl_md_manager.cpp
aschwartz12: Good point.
Identical re-registration is now treated as a harmless duplicate, but rebinding a name to a different {ip, port} returns NIXL_ERR_NOT_ALLOWED. To move a peer, callers unregisterMDPeer first (which also invalidates the metadata we sent to the old address), avoiding a split-brain peer. 
Added unit tests for both paths.

src/core/nixl_md_manager.cpp
aschwartz12: Safe by design. lookupPeer copies {ip, port} under the lock, so we never hold the mutex across the network call (which we want to avoid). If a concurrent unregister/re-register changes the registry in that gap, the worst case is we send to the just-removed address, which is acceptable. Added a comment explaining this.

src/core/agent_data.h
aschwartz12: Done. The manager is built once in the ctor, so the lazy-init mutex is gone and getMDManager() just returns the pointer (or NIXL_ERR_NOT_SUPPORTED if null).

src/core/agent_data.h
aschwartz12: Removed. The NIXL_MD_MANAGER check now lives in the ctor and mdManager_ != nullptr is the single source of truth.

src/core/nixl_md_manager.cpp
aschwartz12: I considered both.  nixl_md_manager.h is an installed public header (install_headers in meson.build), so using nixlLock would leak absl into the public API. Reusing nixlData's lock would re-introduce exactly the nixlAgent coupling/friendship we just removed. Given the registry is small and only guards a local map, a plain std::mutex keeps the header dependency-free. If we want configurable modes later, we can hide a nixlLock behind a pimpl so absl stays out of the header — happy to do that as a follow-up if you prefer.

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,190 @@
...
+    explicit nixlMDManager(nixlAgent &agent) noexcept : agent_(agent) {}
+
+    /**
+     * @brief Register a P2P peer's reachable address.
+     *
+     * Re-registering the same {ip, port} is a no-op; rebinding a name to a
+     * different address is rejected (unregisterMDPeer first).
+     *
+     * @param agent_name Logical agent name.
+     * @param ip         IPv4 dotted-decimal address.
```
iyastreb: Is it strictly for IPv4?
If so, maybe we can use dedicated integer based type to represent the address?

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,190 @@
...
+private:
+    struct Peer {
+        std::string ip;
+        std::uint16_t port;
+    };
+
+    [[nodiscard]] bool
+    lookupPeer(const std::string &agent_name, Peer &out) const;
+
+    nixlAgent &agent_;
```
iyastreb: can it be const?

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,141 @@
...
+nixlMDManager::registerMDPeer(const std::string &agent_name,
+                              const std::string &ip,
+                              std::uint16_t port) {
+    if (agent_name.empty() || ip.empty()) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+    // 0 means "use the listener-side default", matching the existing
+    // nixl_opt_args_t::port default.
+    const std::uint16_t resolved_port =
+        (port == 0) ? static_cast<std::uint16_t>(default_comm_port) : port;
```
iyastreb: cast seems redundant

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,141 @@
...
+ * limitations under the License.
+ */
+#include "nixl_md_manager.h"
+
+#include "nixl.h"
+
+namespace {
+
+nixl_opt_args_t
+makePeerArgs(const std::string &ip, std::uint16_t port, const nixl_opt_args_t *base = nullptr) {
```
iyastreb: Does it make sense to pass Peer struct here?
Please use move semantics to avoid redundant copies

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return agent_.fetchRemoteMD(agent_name, &args);
+}
+
+nixl_status_t
+nixlMDManager::invalidateLocalMD(const std::string &agent_name) const {
+    Peer peer;
+    if (!lookupPeer(agent_name, peer)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+    const auto args = makePeerArgs(peer.ip, peer.port);
```
iyastreb: std::move peer, and in other places too

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto it = peers_.find(agent_name);
+    if (it == peers_.end()) {
+        return false;
+    }
+    out = it->second;
+    return true;
+}
+
+nixl_status_t
+nixlMDManager::registerMDPeer(const std::string &agent_name,
```
iyastreb: For all events - maybe we can add logging, at DEBUG or trace level?

aschwartz12: Added. Control-plane events log at NIXL_DEBUG (register/unregister/send/fetch/invalidate, including the rejection paths). checkRemoteMD is NIXL_TRACE instead because it's typically polled in a loop and DEBUG would be noisy.

src/core/nixl_md_manager.cpp
aschwartz12: Done - invalidateLocalMD now does makePeerArgs(std::move(peer.ip), peer.port), and I applied the same to sendLocalMD, sendLocalPartialMD, and fetchRemoteMD. The only place I kept a copy is unregisterMDPeer, where peer.ip is reused afterward in the compare-then-erase guard.

src/core/nixl_md_manager.cpp
aschwartz12: I kept it taking the fields rather than the Peer struct on purpose. Peer is a private nested type, so a helper taking it would have to be a member of nixlMDManager, which would pull this purely-internal helper into nixl_md_manager.h — an installed public header we've deliberately kept minimal (same reason we avoided nixlLock/absl there). Keeping it a .cpp-local free function avoids that. I did address the actual concern: it now takes the ip by value and std::moves it in, so the redundant copy is gone.

src/core/nixl_md_manager.cpp
aschwartz12: Right, removed.

src/api/cpp/nixl_md_manager.h
aschwartz12: Four of the five forwarded agent methods are already const, but nixlAgent::fetchRemoteMD is non-const, so agent_ can't be a const nixlAgent& without a build error.

src/api/cpp/nixl_md_manager.h
aschwartz12: It is effectively IPv4-only today - the listener uses inet_pton(AF_INET, ...). 
I kept std::string rather than an integer type for two reasons: 
 1. the underlying API we wrap (nixl_opt_args_t::ipAddr) is a string, so an integer would just be converted back. 
 2. a uint32_t would hard-lock us to IPv4 and break the day the listener gains AF_INET6. 
 
 To address the real concern I now validate with inet_pton and reject bad addresses early.

src/core/nixl_agent.cpp
```diff
@@ -1710,6 +1715,18 @@ nixlAgent::checkRemoteMD (const std::string remote_name,
     return NIXL_ERR_NOT_FOUND;
 }
 
+nixl_status_t
```
iyastreb: imo just returning a pointer is enough (nullptr if not supported), but up to you

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+nixl_status_t
+nixlMDManager::fetchRemoteMD(const std::string &agent_name) const {
+    Peer peer;
+    if (!lookupPeer(agent_name, peer)) {
+        return NIXL_ERR_NOT_FOUND;
+    }
+    const auto args = makePeerArgs(std::move(peer.ip), peer.port);
+    NIXL_DEBUG << "fetchRemoteMD: " << agent_name;
```
iyastreb: I think for all traces it would be useful to print local agent name first

iyastreb (commented):
LGTM apart from minor comments

src/core/nixl_agent.cpp
aschwartz12: Kept the nixl_status_t + out-param form for consistency - every agent getter/factory uses it

src/core/nixl_md_manager.cpp
aschwartz12: Done - all traces now start with [<local_agent>]. The manager only had a nixlAgent& with no public name accessor, so I pass the name into the ctor instead of adding a public getName() just for logging.

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26880160891/job/79277830623)…

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26880154375/job/79277807993)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `3d31c0c8`

All the evidence needed is in the logs. Here is the complete diagnosis:

---

**Summary:** `clang-format-19` requires the constructor initializer list in `nixlMDManager`'s constructor to be split onto separate lines, but the PR wrote it on a single line.

**Root cause:** In `src/api/cpp/nixl_md_manager.h` (around line 57), the constructor initializer list is written as a single-line form:
```cpp
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent), localName_(std::move(local_name)) {}
```
`clang-format-19`, using the project's `.clang-format` style file, requires each initializer to be on its own line when they don't all fit within the column limit:
```cpp
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent),
      localName_(std::move(local_name)) {}
```
The `clang-format-diff-19` script detected this discrepancy, produced a non-empty diff, and exited with code 1, failing the check.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1690 author)

**File:** `src/api/cpp/nixl_md_manager.h:57`

**Suggested fix:** In `src/api/cpp/nixl_md_manager.h`, split the constructor initializer list so each member initializer is on its own line. The simplest fix is to apply `clang-format-19` directly before pushing:
```bash
clang-format-19 -i -style=file src/api/cpp/nixl_md_manager.h
```
This will rewrite the constructor as:
```cpp
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent),
      localName_(std::move(local_name)) {}
```
Then commit and push the reformatted file. All other files touched by the PR (`nixl.h`, `agent_data.h`, `nixl_agent.cpp`, `nixl_md_manager.cpp`, `test/gtest/md_manager.cpp`) passed the format check — only `nixl_md_manager.h` needs fixing.

**Related:** https://github.com/ai-dynamo/nixl/pull/1690



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `48dfe1e5-6a4d-4cb0-b85c-6403b545ec97` in the triage console for the audit trail.

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `acb7cfce`

All the evidence is in hand. Here is the complete diagnosis:

---

**Summary:** Clang Format Check fails on PR #1690 — constructor initializer list in `nixl_md_manager.h` is not formatted per project style.

**Root cause:** The `nixlMDManager` constructor at line 57 of `src/api/cpp/nixl_md_manager.h` places both initializer-list members on a single line:
```cpp
// CURRENT (fails clang-format-19):
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent), localName_(std::move(local_name)) {}
```
`clang-format-19` (enforced by the project's `.clang-format` config) requires each initializer-list entry to be on its own line when they don't all fit on one line within the column limit:
```cpp
// REQUIRED (clang-format-19 output):
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent),
      localName_(std::move(local_name)) {}
```
The `clang-format-diff-19` tool detected this discrepancy and exited with code 1, failing the CI step.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1690 branch `metadata_manager_pr1`)

**File:** `src/api/cpp/nixl_md_manager.h:57`

**Suggested fix:** In `src/api/cpp/nixl_md_manager.h`, split the constructor initializer list so each member initializer is on its own line, exactly as `clang-format-19` prescribes:
```cpp
nixlMDManager(nixlAgent &agent, std::string local_name) noexcept
    : agent_(agent),
      localName_(std::move(local_name)) {}
```
The quickest way to apply this (and catch any other formatting issues across all PR-touched files) is to run locally:
```bash
# From repo root, on the branch
git diff -U0 HEAD^1 HEAD -- src/api/cpp/nixl_md_manager.h \
  src/api/cpp/nixl.h src/core/agent_data.h src/core/nixl_agent.cpp \
  src/core/nixl_md_manager.cpp test/gtest/md_manager.cpp \
git add -u && git commit --amend --no-edit
```

**Related:** [PR #1690 — feat(core): add opt-in nixlMDManager (PR 1)](https://github.com/ai-dynamo/nixl/pull/1690)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `c1d781b9-7f94-402e-b35d-79160599905b` in the triage console for the audit trail.

iyastreb (dismissed):
LGTM
@mkhazraee This is new API, could you please check?

src/core/agent_data.h
```diff
@@ -67,6 +68,9 @@ class nixlAgentData {
         const nixlAgentConfig config_;
         const bool useEtcd_;
         const bool needsCommThread_;
+        // Constructed in the nixlAgent ctor when NIXL_MD_MANAGER is set;
+        // null (and getMDManager reports NIXL_ERR_NOT_SUPPORTED) otherwise.
+        std::unique_ptr<nixlMDManager> mdManager_;
```
ColinNV: Prefer this to be `const`, even if it means creating a small utility function that can be called in the member initializer list.

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <arpa/inet.h>
+#include <utility>
+
+namespace {
+
+// Build the nixl_opt_args_t for a peer. Takes ip by value and moves it in to
+// avoid a redundant copy. Kept .cpp-local rather than passing the manager's
+// private Peer struct, which would pull this helper into the installed header.
+nixl_opt_args_t
```
ColinNV: `[[nodiscard]]`

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+    args.ipAddr = std::move(ip);
+    args.port = port;
+    return args;
+}
+
+} // namespace
+
+bool
+nixlMDManager::lookupPeer(const std::string &agent_name, Peer &out) const {
+    std::lock_guard<std::mutex> lk(mutex_);
```
ColinNV: `const std::lock_guard lock(mutex_);`

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,196 @@
...
+    const std::string self_name_;
```
ColinNV: Could and should this be replaced by something like this?
```
[[nodiscard]] const std::string &
nixlAgent::getName() const noexcept;
```

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+    if (s != NIXL_SUCCESS) {
+        NIXL_DEBUG << "[" << self_name_ << "] unregisterMDPeer: invalidate for " << agent_name
+                   << " failed: " << s;
+        return s;
+    }
+    // Compare-then-erase: guard against a concurrent registerMDPeer
+    // having replaced the entry while invalidate was in flight.
+    std::lock_guard<std::mutex> lk(mutex_);
+    auto it = peers_.find(agent_name);
+    if (it != peers_.end() && it->second.ip == peer.ip && it->second.port == peer.port) {
```
ColinNV: This should be implemented?
```
[[nodiscard]] bool
operator==( const Peer& l, const Peer& r ) noexcept;
```

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,196 @@
...
+#include <unordered_map>
+#include <utility>
+
+class nixlAgent;
+
+/**
+ * @brief Transport selected by the manager. Currently always P2P;
+ *        ETCD is not yet supported.
+ */
+enum class nixl_md_backend_t {
```
ColinNV: No chance making this "open", i.e. without an `enum` and having the md backend report its type as string?

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,196 @@
...
+
+    nixlMDManager(nixlMDManager &&) = delete;
+    nixlMDManager(const nixlMDManager &) = delete;
+    nixlMDManager &
+    operator=(nixlMDManager &&) = delete;
+    nixlMDManager &
+    operator=(const nixlMDManager &) = delete;
+
+private:
+    struct Peer {
```
ColinNV: How MD backend specific is this?

src/core/agent_data.h
aschwartz12: Done. Moved construction into nixlAgentData's initializer list, made mdManager_ const, and kept the manager ctor side-effect-free.

src/core/nixl_md_manager.cpp
aschwartz12: Done

src/core/nixl_md_manager.cpp
aschwartz12: Done

src/api/cpp/nixl_md_manager.h
aschwartz12: Done. Since the agent name is already part of the public constructor/API identity, I added getName() and removed the manager's duplicate name copy.

src/core/nixl_md_manager.cpp
aschwartz12: Done. Added operator== for the peer type and now use it in the erase guard.

src/api/cpp/nixl_md_manager.h
aschwartz12: Done. getBackend() now reports the backend name as a string view, so future backends do not require extending a public enum.

src/api/cpp/nixl_md_manager.h
aschwartz12: The {ip, port} is P2P socket addressing, but the struct is really just the value of a name-keyed registry. When we add ETCD in the next PR, that backend reuses the same name -> Peer registry and simply ignores the address fields (it keys metadata by agent name in the store), so the struct itself doesn't need to change. That's why I kept it generic Peer rather than P2PPeer.

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    nixlMDManager(nixlMDManager &&) = delete;
+    nixlMDManager(const nixlMDManager &) = delete;
+    nixlMDManager &
+    operator=(nixlMDManager &&) = delete;
+    nixlMDManager &
+    operator=(const nixlMDManager &) = delete;
+
+private:
```
mkhazraee: I think it would be good to use a pImpl style like nixl, to keep the main header clean.

src/api/cpp/nixl.h
```diff
@@ -561,6 +572,21 @@ class nixlAgent {
...
+         * is built in the agent ctor and the returned pointer is owned by the
+         * agent; do not delete it. `out` is set to `nullptr` on the disabled
+         * path.
+         *
+         * @param  out [out]     Pointer to the agent-owned nixlMDManager,
+         *                       or `nullptr` if the manager is disabled.
+         * @return nixl_status_t NIXL_SUCCESS or NIXL_ERR_NOT_SUPPORTED.
+         */
+        nixl_status_t
+        getMDManager(nixlMDManager *&out);
```
mkhazraee: I'm not sure we need this. User instantiates the md manager, right? So they have the handle to it. I'm thinking md manager is an external class feeding nixl agent, nixl agent doesn't need to know about it or keep any state.

aschwartz12: Agreed in principle - the manager only ever calls public nixlAgent methods, so the agent doesn't need to own it or hold state. Proposing to drop getMDManager/mdManager_/the NIXL_MD_MANAGER gate and let the user construct nixlMDManager(agent, ...) directly (which also lets me delete the self_name ctor param, since agent.getName() is valid post-construction). Flagging one cross-review tension: this reverses the earlier "build it in the agent ctor and hand out a pointer" direction from @iyastreb  - want to confirm you both prefer the user-owned model before I make the change.

src/api/cpp/nixl_md_manager.h
aschwartz12: Done - moved the private surface (Peer, lookupPeer, and the agent_/mutex_/peers_ members) into a core-internal nixlMDManagerData in the .cpp, mirroring nixlAgent/nixlAgentData. The installed header now keeps only the public API plus a single data_ pointer and forward-declares nixlMDManagerData; each method forwards to data_->. No behavior change.

src/api/cpp/nixl.h
iyastreb: I fully agree to remove `getMDManager`: we need just one way of constructing things, and this simplifies the code

aschwartz12: Done - went user-owned. Dropped getMDManager, mdManager_, and the NIXL_MD_MANAGER gate. Opt-in is now just nixlMDManager(agent).

src/api/cpp/nixl.h
```diff
@@ -560,7 +569,6 @@ class nixlAgent {
         nixl_status_t
         checkRemoteMD (const std::string remote_name,
                        const nixl_xfer_dlist_t &descs) const;
-
```
iyastreb, line 563: minor: this can be reverted

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,183 @@
...
+#include "nixl_types.h"
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <string_view>
+
+class nixlAgent;
+// All registry/mutex state lives here; defined in nixl_md_manager.cpp so this
+// installed public header stays free of those details (pImpl).
```
iyastreb: minor: I think comment for forward declarations makes no sense
what if definition evolve/change

src/core/nixl_md_manager.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+// pImpl: the registry/mutex state and all logic live here, so the installed
+// nixl_md_manager.h carries only the public API plus a `data` pointer (mirrors
+// nixlAgent/nixlAgentData). nixlMDManager methods forward to the matching
+// method below.
+class nixlMDManagerData {
+public:
+    explicit nixlMDManagerData(nixlAgent &agent) noexcept : agent_(agent) {}
+
+    [[nodiscard]] nixl_status_t
```
iyastreb: that's a bit weird way of implementing pimpl - it just duplicates the owner interface.
It can be just a struct which stores all the fields, and maybe providing some helpers/utils, but not implementing the core logic. Core logic can still be implemented in the owner as it was before

src/api/cpp/nixl.h
aschwartz12: done

src/api/cpp/nixl_md_manager.h
aschwartz12: Removed. Agreed - documenting the definition's internals from the forward declaration just goes stale; kept the one note on the data_ member instead.

src/core/nixl_md_manager.cpp
aschwartz12: Good point, that was over-engineered. I've made nixlMDManagerData a plain state struct now - it just holds agent/mutex/peers and the Peer type, plus lookupPeer as a shared helper.

src/api/cpp/nixl.h
```diff
@@ -26,6 +26,7 @@
 #include "nixl_descriptors.h"
 #include <chrono>
 #include <memory>
+#include <string>
```
iyastreb: this change is needed?

src/core/agent_data.h
```diff
@@ -63,80 +63,127 @@ using nixl_socket_peer_t = std::pair<std::string, int>;
 
 using nixl_socket_map_t = std::map<nixl_socket_peer_t, int>;
 
-class nixlAgentData {
-    private:
```
ColinNV, line 67: Erm, if you could only reformat new or changed lines, otherwise it's hard to see which lines did in fact change.

src/core/nixl_agent.cpp
```diff
@@ -1426,7 +1436,64 @@ nixlAgent::genNotif(const std::string &remote_agent,
...
 nixl_status_t
-nixlAgent::getLocalMD (nixl_blob_t &str) const {
+nixlAgentData::getLocalMD(nixl_blob_t &str) {
+    size_t conn_cnt;
+    nixl_backend_t nixl_backend;
+    nixl_status_t ret;
+
+    NIXL_LOCK_GUARD(lock);
+    // connMd_ was populated when the backend was created
+    conn_cnt = connMd_.size();
```
ColinNV: `const size_t conn_cnt = connMd_.size();` instead of `conn_cnt` above.

src/core/nixl_agent.cpp
```diff
@@ -1426,7 +1436,64 @@ nixlAgent::genNotif(const std::string &remote_agent,
...
+
+    if (conn_cnt == 0) { // Error, no backend supports remote
+        NIXL_ERROR_FUNC << "no backends support remote operations";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    nixlSerDes sd;
+    ret = sd.addStr("Agent", name_);
+    // Always returns SUCCESS, serdes class logs errors if necessary
+    if (ret) {
```
ColinNV: Better `if (ret != NIXL_SUCCESS)` for when we switch to `enum class`.

src/core/nixl_agent.cpp
```diff
@@ -1470,10 +1539,105 @@ nixlAgent::getLocalMD (nixl_blob_t &str) const {
...
+                                 nixl_blob_t &str,
+                                 const nixl_opt_args_t *extra_params) {
+    backend_list_t tmp_list;
+    backend_list_t *backend_list;
+    nixl_status_t ret;
+
+    NIXL_LOCK_GUARD(lock);
+
+    if (!extra_params || extra_params->backends.size() == 0) {
+        if (descs.descCount() != 0) {
```
ColinNV: `if (!descs.isEmpty()) {`

src/core/nixl_md_manager.h
```diff
@@ -0,0 +1,100 @@
...
+     * @param extra_params Operational args (e.g. `ipAddr`/`port` for P2P).
+     */
+    [[nodiscard]] nixl_status_t
+    invalidateLocalMD(const nixl_opt_args_t *extra_params = nullptr) const;
+
+    /**
+     * @brief Name of the active metadata backend (e.g. "P2P").
+     */
+    [[nodiscard]] std::string_view
+    getBackend() const noexcept;
```
ColinNV: Perhaps `backendName()`?

src/core/agent_data.h
aschwartz12: Yes, sorry. 
Fixed it.

src/core/nixl_agent.cpp
aschwartz12: Done - declared and initialized in one const statement.

src/core/nixl_agent.cpp
aschwartz12: Done

src/core/nixl_agent.cpp
aschwartz12: Done

src/core/nixl_md_manager.h
aschwartz12: Done

PR #1693, opened 2026-05-28
nixl_ep: Refactor rank and expert semantics

itayalroy:
num_ranks/num_experts had mixed meanings across different contexts. Replace them with explicit fields that have well-defined meanings:
  - `max_num_ranks`: Preallocated rank capacity
  - `remote_ranks`: Array of connected ranks
  - `active_ranks`: Array of explicitly activated ranks (via `connect_ranks(..., activate=True)` or `update_mask_buffer`)
  - `active_rank_bound`: max(active_ranks) + 1
  - `num_experts_per_rank`: static number of physical experts per rank

Additionally, make `connect_ranks(..., activate=False)` establish connections without changing `active_ranks` or `active_rank_bound` (instead of wrongly modifying `num_ranks` earlier)

Tested thoroughly with both `elastic.py` and vLLM EEP

## Summary by CodeRabbit

* **New Features**
  * Added public mask-update capability and host-side tracking of active ranks for elastic rank handling.

* **API Changes**
  * Dispatch/combine now accept an active-rank bound + experts-per-rank parameterization; internal buffer/layouts updated to use the active-range model.

* **Deprecations**
  * Removed the legacy mask-clean API.

* **Documentation**
  * Clarified buffer sizing, rank-capacity semantics, and mask behavior in docstrings.

### DIFF
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -81,8 +81,8 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
     const auto warp_id = thread_id / 32, lane_id = get_lane_id();
     const auto num_sms = static_cast<int>(gridDim.x);
     const auto num_warps = num_warp_groups * num_warps_per_group;
-    const auto num_local_experts = num_experts_per_rank;
-    const auto active_expert_bound = active_rank_bound * num_local_experts;
+    const int num_local_experts = num_experts_per_rank;
+    const int active_expert_bound = active_rank_bound * num_local_experts;
     const auto warp_group_id = warp_id / num_warps_per_group;
     const auto sub_warp_id = warp_id % num_warps_per_group;
     const auto responsible_expert_idx = sm_id * num_warp_groups + warp_group_id;
@@ -628,8 +628,8 @@ combine(void* combined_x,
     const auto thread_id = static_cast<int>(threadIdx.x);
     const auto num_threads = __shfl_sync(0xffffffff, static_cast<int>(blockDim.x), 0);
     const auto warp_id = __shfl_sync(0xffffffff, thread_id / 32, 0), lane_id = get_lane_id();
-    const auto num_local_experts = num_experts_per_rank;
-    const auto active_expert_bound = active_rank_bound * num_local_experts;
+    const int num_local_experts = num_experts_per_rank;
+    const int active_expert_bound = active_rank_bound * num_local_experts;
     const auto warp_group_id = warp_id / num_warps_per_group;
     const auto sub_warp_id = warp_id % num_warps_per_group;
     const auto responsible_expert_idx = sm_id * num_warp_groups + warp_group_id;
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -72,7 +72,7 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
          int* atomic_counter_per_expert, int* atomic_finish_counter_per_expert,
          uint64_t* next_clean, int num_next_clean_int,
          int num_tokens, int num_max_dispatch_tokens_per_rank,
-         int num_topk, int num_experts, int rank, int num_ranks,
+         int num_topk, int active_rank_bound, int num_experts_per_rank, int rank,
          int num_warp_groups, int num_warps_per_group,
          bool round_scale, uint64_t timeout_cycles, int phases, nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr) {
     auto nixl_ctx = *nixl_ctx_ptr;
@@ -81,7 +81,8 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
     const auto warp_id = thread_id / 32, lane_id = get_lane_id();
     const auto num_sms = static_cast<int>(gridDim.x);
     const auto num_warps = num_warp_groups * num_warps_per_group;
-    const auto num_local_experts = num_experts / num_ranks;
+    const auto num_local_experts = num_experts_per_rank;
+    const auto active_expert_bound = active_rank_bound * num_local_experts;
     const auto warp_group_id = warp_id / num_warps_per_group;
     const auto sub_warp_id = warp_id % num_warps_per_group;
     const auto responsible_expert_idx = sm_id * num_warp_groups + warp_group_id;
@@ -175,13 +176,14 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
 
             // Issue NIXL sends
             if (dst_expert_idx >= 0) {
+                EP_DEVICE_ASSERT(dst_expert_idx < active_expert_bound);
                 int slot_idx = lane_id == 0 ? atomicAdd(atomic_counter_per_expert + dst_expert_idx, 1) : 0;
                 slot_idx = __shfl_sync(0xffffffff, slot_idx, 0);
                 const auto dst_rank = dst_expert_idx / num_local_experts;
                 const auto dst_expert_local_idx = dst_expert_idx % num_local_experts;
                 const auto src_ptr = reinterpret_cast<uint64_t>(rdma_x_src_idx);
                 const auto dst_ptr = reinterpret_cast<uint64_t>(rdma_recv_x) +
-                                     dst_expert_local_idx * num_ranks * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
+                                     dst_expert_local_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                                      rank * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                                      slot_idx * num_bytes_per_msg;
                 if (not is_rank_masked<true>(mask_buffer_ptr, dst_rank)) {
@@ -215,14 +217,14 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
             // Notify before executing `int_p`
             __syncwarp();
             #pragma unroll
-            for (int i = lane_id; i < num_experts; i += 32)
+            for (int i = lane_id; i < active_expert_bound; i += 32)
                 atomic_add_release_global(atomic_finish_counter_per_expert + i, FINISHED_SUM_TAG);
         }
 
         // This SM should be responsible for some destination experts, read `topk_idx` for them
         int expert_count[kNumMaxWarpGroups] = {0};
         const auto expert_begin_idx = sm_id * num_warp_groups;
-        const auto expert_end_idx = min(expert_begin_idx + num_warp_groups, num_experts);
+        const auto expert_end_idx = min(expert_begin_idx + num_warp_groups, active_expert_bound);
 
         // Per lane count
         #pragma unroll 8
@@ -245,14 +247,14 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
     __syncthreads();
 
     // Issue count sends
-    if (responsible_expert_idx < num_experts and sub_warp_id == 0 and lane_id == 0) {
+    if (responsible_expert_idx < active_expert_bound and sub_warp_id == 0 and lane_id == 0) {
         const auto dst_rank = responsible_expert_idx / num_local_experts;
         const auto dst_expert_local_idx = responsible_expert_idx % num_local_experts;
         const auto num_tokens_sent = shared_num_tokens_sent_per_expert[responsible_expert_idx - sm_id * num_warp_groups];
 
         // Wait local sends issued and send expert counts
         while (ld_acquire_global(atomic_finish_counter_per_expert + responsible_expert_idx) != FINISHED_SUM_TAG * 2);
-        auto dst_ptr = reinterpret_cast<uint64_t>(rdma_recv_count + dst_expert_local_idx * num_ranks + rank);
+        auto dst_ptr = reinterpret_cast<uint64_t>(rdma_recv_count + dst_expert_local_idx * active_rank_bound + rank);
         if (not is_rank_masked(mask_buffer_ptr, dst_rank)) {
             void* dst_p2p_ptr = p2p_ptr_get(nixl_ctx, dst_ptr, dst_rank);
             if (dst_p2p_ptr == 0) {
@@ -283,18 +285,18 @@ DISPATCH_RECV:
         cg::this_grid().sync();
 
     // Receiving and packing
-    if (responsible_expert_idx < num_experts) {
+    if (responsible_expert_idx < active_expert_bound) {
         const auto src_rank = responsible_expert_idx / num_local_experts;
         const auto local_expert_idx = responsible_expert_idx % num_local_experts;
         const auto rdma_recv_x_uint8 = static_cast<uint8_t*>(rdma_recv_x) +
-                local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
+                local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                 src_rank * num_max_dispatch_tokens_per_rank * num_bytes_per_msg;
         const auto recv_x_int4 = static_cast<int4*>(packed_recv_x) +
-                local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * hidden_int4;
-        const auto recv_src_info = packed_recv_src_info + local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank;
-        const auto recv_range = packed_recv_layout_range + local_expert_idx * num_ranks;
+                local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * hidden_int4;
+        const auto recv_src_info = packed_recv_src_info + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank;
+        const auto recv_range = packed_recv_layout_range + local_expert_idx * active_rank_bound;
         const auto num_aligned_scales = align_up<int>(num_scales, sizeof(float) / sizeof(scale_t));
-        const auto recv_x_scales = static_cast<scale_t*>(packed_recv_x_scales) + local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * num_aligned_scales;
+        const auto recv_x_scales = static_cast<scale_t*>(packed_recv_x_scales) + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * num_aligned_scales;
 
         // Shared between sub-warps in warp groups
         __shared__ int shared_num_recv_tokens[kNumMaxWarpGroups], shared_recv_token_begin_idx[kNumMaxWarpGroups];
@@ -307,7 +309,7 @@ DISPATCH_RECV:
             auto start_time = clock64();
             uint64_t wait_recv_cost = 0;
             if (not is_rank_masked(mask_buffer_ptr, src_rank)) {
-                while ((num_recv_tokens = ld_acquire_sys_global(rdma_recv_count + local_expert_idx * num_ranks + src_rank)) ==
+                while ((num_recv_tokens = ld_acquire_sys_global(rdma_recv_count + local_expert_idx * active_rank_bound + src_rank)) ==
                            0                                                               // data not arrived
                        && (wait_recv_cost = clock64() - start_time) <= timeout_cycles       // not timeout
                 )
@@ -366,7 +368,7 @@ DISPATCH_RECV:
                 const auto num_elems_per_pack = static_cast<int>(sizeof(packed_t) / sizeof(scale_t));
                 const auto token_idx = recv_token_begin_idx + i;
                 const auto token_stride = num_elems_per_pack;
-                const auto pack_stride = num_ranks * num_max_dispatch_tokens_per_rank * num_elems_per_pack;
+                const auto pack_stride = active_rank_bound * num_max_dispatch_tokens_per_rank * num_elems_per_pack;
                 if (lane_id < num_scales) {
                     const auto pack_idx = lane_id / num_elems_per_pack;
                     const auto elem_idx = lane_id % num_elems_per_pack;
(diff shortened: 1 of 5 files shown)


examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -451,40 +469,43 @@ void Buffer::connect_ranks(const std::vector<int>& remote_ranks_list, const std:
...
+        CUDA_CHECK(cudaDeviceSynchronize());
+    }
 
-    CUDA_CHECK(cudaDeviceSynchronize());
+    if (activate) {
+        for (int remote_rank : remote_ranks_list) {
+            if (remote_rank != rank)
+                update_mask_buffer(remote_rank, false);
+        }
+    }
```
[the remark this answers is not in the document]
itayalroy: Host `active_ranks` may diverge from device rank mask (as the latter is updated by comm kernels on faults), and `activate=True` should ensure both host/device states are updated

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -160,7 +175,7 @@ void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_nvl_bytes
         *moe_recv_rdma_counter = -1;
     }
 
-    EP_HOST_ASSERT(max_experts_per_rank > 0);
+    EP_HOST_ASSERT(num_experts_per_rank > 0);
```
rakhmets: Looks like this could be moved to the beginning of the function.
And then the class field `Buffer::num_experts_per_rank` can be `unsigned` instead of `int`. Since there is no point in negative values for this field.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -81,7 +81,8 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
     const auto warp_id = thread_id / 32, lane_id = get_lane_id();
     const auto num_sms = static_cast<int>(gridDim.x);
     const auto num_warps = num_warp_groups * num_warps_per_group;
-    const auto num_local_experts = num_experts / num_ranks;
+    const auto num_local_experts = num_experts_per_rank;
+    const auto active_expert_bound = active_rank_bound * num_local_experts;
```
rakhmets: In these lines, the use of `auto` doesn't improve, but worsens the readability of the code.
Please replace `auto` by the actual type (`int` in these cases) in modified lines.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -879,6 +899,8 @@ Buffer::ht_combine(const torch::Tensor& x, const std::optional<torch::Tensor>& t
                           const torch::Tensor& combined_rdma_head, const torch::Tensor& combined_nvl_head,
                           const Config& config, std::optional<EventHandle>& previous_event, bool async, bool allocate_on_comm_stream) {
     EP_HOST_ASSERT(!low_latency_mode && "ht_combine() requires high-throughput mode (low_latency_mode=false)");
+    const int num_ranks = max_num_ranks;
+    EP_HOST_ASSERT(num_ranks > 0);
```
rakhmets: I think it's better to check it once. You can check that num_ranks is greater than 0, and then set `max_num_ranks`. So, you can change the type to `unsigned`, and also use local variable of the same type.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -171,6 +186,9 @@ void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_nvl_bytes
     mask_buffer_ptr = static_cast<int *>(m_mask_alloc->ptr());
     CUDA_CHECK(cudaMemset(mask_buffer_ptr, 0xff, num_mask_buffer_bytes));
     CUDA_CHECK(cudaMemset(mask_buffer_ptr + rank, 0, sizeof(int)));
+    active_ranks.assign(max_num_ranks, false);
+    active_ranks[rank] = true;
+    _refresh_active_rank_bound();
```
rakhmets:
```suggestion
    active_rank_bound = rank + 1;
```

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -83,15 +83,30 @@ Buffer::Buffer(int rank, bool explicitly_destroy, bool low_latency_mode, int tim
             EP_HOST_ASSERT(timeout_ms >= 0);
             return static_cast<uint64_t>(timeout_ms);
         }()),
-        rank(rank), num_ranks(1),
+        rank(rank),
         explicitly_destroy(explicitly_destroy),
         comm_stream(at::cuda::getStreamFromPool(true)) {}
 
+bool Buffer::_is_rank_connected(int rank_id) const {
+    EP_HOST_ASSERT(rank_id >= 0 and rank_id < max_num_ranks);
```
rakhmets: Not sure if this `assert` is useful. The method just returns false.

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Moved it to the beginning, but I kept the field as int for now because public API, kernel wrappers/params, tensor EPLayout etc all still use int. Changing only this class field to unsigned would add lots of casting. If we want to move this to unsigned, I think it should be done in a separate PR that will update all of the above consistently.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: done

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Moved to the beginning, same as the above reg. unsigned, should do it thoroughly, else it will introduce too much casting, and I think it is outside the scope of this PR

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: done

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: removed it

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1015,26 +1035,27 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
     EP_HOST_ASSERT(topk_idx.dim() == 2 and topk_idx.is_contiguous());
     EP_HOST_ASSERT(x.size(0) == topk_idx.size(0) and x.size(0) <= num_max_dispatch_tokens_per_rank);
     EP_HOST_ASSERT(topk_idx.scalar_type() == c10::CppTypeToScalarType<topk_idx_t>::value);
-    EP_HOST_ASSERT(num_experts % num_ranks == 0);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    EP_HOST_ASSERT(active_rank_bound > 0);
```
rakhmets: This assert is not very useful here. Because it checks class field, not a function argument. The better way is to check the value right after changing it. In this case the error will be reported in problematic place. E.g. private method set_active_rank_bound that asserts non-positive values, and use it in `Buffer::_refresh_active_rank_bound`, and in `Buffer::init`


The same issue is `Buffer::combine`, `Buffer::get_next_combine_buffer`.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1015,26 +1035,27 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
...
+    EP_HOST_ASSERT(num_experts == active_expert_bound && "num_experts must equal active_rank_bound * num_experts_per_rank");
```
rakhmets: It doesn't look right. `dispatch` method has `num_experts` parameter. The method only works if the passed parameter is equal to `Buffer::active_rank_bound * Buffer::num_experts_per_rank`. `num_experts` parameter lost meaning due to this assert.

itayalroy: It never had any meaning, num experts is not configurable, it is derived from num ranks and num experts per rank, unfortunately we took it from deep_ep and cannot break backwards compatibility, the least we can do is validate that users actually pass the expected value here.

rakhmets: Before the changes the function used `num_experts` parameter to calculate `num_local_experts` local variable. And the function accepted different values from the caller. The only limitation was dividing without remainder by `Buffer::num_ranks`.

Now the function doesn't use `num_experts` parameter. And the assert is redundant in this function. It can be removed.

itayalroy: `num_local_experts` must match `num_experts_per_rank` value passed to `update_memory_buffers` on init. It should not be derived from user input per dispatch/combine, since this value is fixed and cannot change. In this PR I just enforce it instead of hoping users will pass the correct `num_experts`. This parameter was wrongfully added by DeepEP to dispatch/combine calls and we inherited it sadly

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: done

examples/device/ep/csrc/nixl_ep.cpp
rakhmets: `num_experts` parameter is not used anymore in this function.
```suggestion
                            int num_max_dispatch_tokens_per_rank, int,
```
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f9-unused-parameters-should-be-unnamed

And please remove the assert that checks `num_experts` at line 1150.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1217,7 +1243,11 @@ Buffer::combine(const torch::Tensor& x, const torch::Tensor& topk_idx, const tor
 
 torch::Tensor
 Buffer::get_next_combine_buffer(int num_max_dispatch_tokens_per_rank, int hidden, int num_experts) const {
```
rakhmets:
```suggestion
Buffer::get_next_combine_buffer(int num_max_dispatch_tokens_per_rank, int hidden, int) const {
```

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1217,7 +1243,11 @@ Buffer::combine(const torch::Tensor& x, const torch::Tensor& topk_idx, const tor
...
-    int max_num_experts = max_num_ranks * max_experts_per_rank;
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    EP_HOST_ASSERT(active_rank_bound > 0);
+    EP_HOST_ASSERT(num_experts == active_expert_bound && "num_experts must equal active_rank_bound * num_experts_per_rank");
```
rakhmets:
```suggestion
```

examples/device/ep/csrc/nixl_ep.cpp
rakhmets: Yes, I got it.
Please remove this assert, and unname unused parameter.

examples/device/ep/csrc/nixl_ep.hpp
rakhmets: Can we remove `num_experts` parameter from the C++ API (`dispatch`, `combine`, `get_next_combine_buffer`), and keep it in PYTHON API?

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: done

examples/device/ep/csrc/nixl_ep.hpp
itayalroy: I removed `num_experts` from the C++ API and made it optional in Python. This is safe because it was the last required positional dispatch parameter, so existing callers won’t break.

After we release a NIXL version with this change, vLLM/SGLang can stop passing this parameter. Once those integrations are updated, a future NIXL release can remove `num_experts` from the Python API entirely.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -72,7 +72,7 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
          int* atomic_counter_per_expert, int* atomic_finish_counter_per_expert,
          uint64_t* next_clean, int num_next_clean_int,
          int num_tokens, int num_max_dispatch_tokens_per_rank,
-         int num_topk, int num_experts, int rank, int num_ranks,
+         int num_topk, int active_rank_bound, int num_experts_per_rank, int rank,
```
rakhmets:
```suggestion
         int num_topk, int active_rank_bound, int num_local_experts, int rank,
```
, then you can remove extra line 84.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -625,7 +628,8 @@ combine(void* combined_x,
     const auto thread_id = static_cast<int>(threadIdx.x);
     const auto num_threads = __shfl_sync(0xffffffff, static_cast<int>(blockDim.x), 0);
     const auto warp_id = __shfl_sync(0xffffffff, thread_id / 32, 0), lane_id = get_lane_id();
-    const auto num_local_experts = num_experts / num_ranks;
+    const int num_local_experts = num_experts_per_rank;
```
rakhmets: This line can be removed, if you rename the parameter to `num_local_experts`.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -215,14 +217,14 @@ dispatch(void* packed_recv_x, void* packed_recv_x_scales,
...
-            for (int i = lane_id; i < num_experts; i += 32)
+            for (int i = lane_id; i < active_expert_bound; i += 32)
                 atomic_add_release_global(atomic_finish_counter_per_expert + i, FINISHED_SUM_TAG);
         }
 
         // This SM should be responsible for some destination experts, read `topk_idx` for them
         int expert_count[kNumMaxWarpGroups] = {0};
         const auto expert_begin_idx = sm_id * num_warp_groups;
-        const auto expert_end_idx = min(expert_begin_idx + num_warp_groups, num_experts);
+        const auto expert_end_idx = min(expert_begin_idx + num_warp_groups, active_expert_bound);
```
rakhmets, line 226:
```suggestion
        const int expert_end_idx = min(expert_begin_idx + num_warp_groups, active_expert_bound);
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -283,18 +285,18 @@ DISPATCH_RECV:
...
-                local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
+                local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * num_bytes_per_msg +
                 src_rank * num_max_dispatch_tokens_per_rank * num_bytes_per_msg;
         const auto recv_x_int4 = static_cast<int4*>(packed_recv_x) +
-                local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * hidden_int4;
-        const auto recv_src_info = packed_recv_src_info + local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank;
-        const auto recv_range = packed_recv_layout_range + local_expert_idx * num_ranks;
+                local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * hidden_int4;
+        const auto recv_src_info = packed_recv_src_info + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank;
+        const auto recv_range = packed_recv_layout_range + local_expert_idx * active_rank_bound;
```
rakhmets, line 296:
```suggestion
        const int* recv_src_info = packed_recv_src_info + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank;
        const int64_t* recv_range = packed_recv_layout_range + local_expert_idx * active_rank_bound;
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -366,7 +368,7 @@ DISPATCH_RECV:
                 const auto num_elems_per_pack = static_cast<int>(sizeof(packed_t) / sizeof(scale_t));
                 const auto token_idx = recv_token_begin_idx + i;
                 const auto token_stride = num_elems_per_pack;
-                const auto pack_stride = num_ranks * num_max_dispatch_tokens_per_rank * num_elems_per_pack;
+                const auto pack_stride = active_rank_bound * num_max_dispatch_tokens_per_rank * num_elems_per_pack;
```
rakhmets, line 370:
```suggestion
                const int pack_stride = active_rank_bound * num_max_dispatch_tokens_per_rank * num_elems_per_pack;
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -394,25 +396,26 @@ void dispatch(void* packed_recv_x, void* packed_recv_x_scales,
...
-    const int num_warp_groups = ceil_div(num_experts, num_device_sms);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    const int num_warp_groups = ceil_div(active_expert_bound, num_device_sms);
     const int num_warps_per_group = 32 / num_warp_groups;
     EP_HOST_ASSERT(num_warp_groups > 0 and num_warps_per_group > 0);
     EP_HOST_ASSERT(kNumMaxTopK + 1 <= num_warp_groups * num_warps_per_group);
 
     const auto num_warps = num_warp_groups * num_warps_per_group;
-    const auto num_sms = ceil_div(num_experts, num_warp_groups);
+    const auto num_sms = ceil_div(active_expert_bound, num_warp_groups);
```
rakhmets, line 411:
```suggestion
    const int num_sms = ceil_div(active_expert_bound, num_warp_groups);
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -394,25 +396,26 @@ void dispatch(void* packed_recv_x, void* packed_recv_x_scales,
...
     EP_HOST_ASSERT(num_topk <= kNumMaxTopK);
 
     // Workspace checks
     auto atomic_counter_per_expert = static_cast<int*>(workspace);
-    auto atomic_finish_counter_per_expert = atomic_counter_per_expert + num_experts;
-    EP_HOST_ASSERT(num_experts * sizeof(int) * 2 <= NUM_WORKSPACE_BYTES);
+    auto atomic_finish_counter_per_expert = atomic_counter_per_expert + active_expert_bound;
```
rakhmets, line 416:
```suggestion
    int* atomic_finish_counter_per_expert = atomic_counter_per_expert + active_expert_bound;
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -665,20 +669,20 @@ combine(void* combined_x,
...
     }
 
     // Issue NIXL sends
-    if (responsible_expert_idx < num_experts) {
+    if (responsible_expert_idx < active_expert_bound) {
         const auto dst_rank = responsible_expert_idx / num_local_experts;
         const auto local_expert_idx = responsible_expert_idx % num_local_experts;
         const auto global_expert_idx = rank * num_local_experts + local_expert_idx;
-        const auto layout = __ldg(layout_range + local_expert_idx * num_ranks + dst_rank);
+        const auto layout = __ldg(layout_range + local_expert_idx * active_rank_bound + dst_rank);
```
rakhmets, line 678:
```suggestion
        const int64_t* layout = __ldg(layout_range + local_expert_idx * active_rank_bound + dst_rank);
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -665,20 +669,20 @@ combine(void* combined_x,
...
         const auto local_x = static_cast<const int4*>(x) +
-                local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank * hidden_bf16_int4;
-        const auto local_src_info = src_info + local_expert_idx * num_ranks * num_max_dispatch_tokens_per_rank;
+                local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank * hidden_bf16_int4;
+        const auto local_src_info = src_info + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank;
```
rakhmets, line 681:
```suggestion
        const int* local_src_info = src_info + local_expert_idx * active_rank_bound * num_max_dispatch_tokens_per_rank;
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -1009,18 +1015,19 @@ void combine(void* combined_x,
...
-    const int num_warp_groups = ceil_div(num_experts, num_device_sms);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    const int num_warp_groups = ceil_div(active_expert_bound, num_device_sms);
     const int num_warps_per_group = 32 / num_warp_groups;
     const int num_recv_per_sm = ceil_div(num_combined_tokens, num_device_sms);
     EP_HOST_ASSERT(num_warp_groups > 0 and num_warps_per_group > 0 and num_recv_per_sm >= 0);
 
     const auto num_warps = num_warp_groups * num_warps_per_group;
-    const auto num_sms = max(ceil_div(num_experts, num_warp_groups),
+    const auto num_sms = max(ceil_div(active_expert_bound, num_warp_groups),
```
rakhmets, line 1028:
```suggestion
    const int num_sms = max(ceil_div(active_expert_bound, num_warp_groups),
```

rakhmets (dismissed):
Minor comments.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: done

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: done

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: `nixl_ep_ll.cu` has 147 `auto`s, so this is a general practice in this file. If we want to improve typing, I’m not against it, but let’s make a focused PR for that since I think it’s outside the scope of this PR. Resolving the `auto` -> explicit type comments for now.

rakhmets (approved):

PR #1694, opened 2026-05-28
Update CA cert discovery

anjaliratnam-msft:
## What?
This PR is addressing a concern with finding the CA certs by automatically checking a list of known CA bundle paths across common Linux distributions. Users are still able to override this by setting the `ca_bundle` backend parameter or setting the `AZURE_CA_BUNDLE` environment variable.

## Why?
This PR addresses a comment made in #1329 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS/SSL certificate handling for Azure Blob connections: the client now probes for common system CA bundles when none are configured, improving secure connection reliability across environments.

* **Chores**
  * Unified sourcing of connection string and CA bundle to the central configuration system while preserving empty-value behavior; no external interface changes.


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26589286441/job/78343512063)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `a8e7376d`

I now have a complete and confident diagnosis. Here is the full report:

---

**Summary:** Clang Format Check failed on PR #1694 (`ca-cert-update`) due to misaligned inline comments in `azure_blob_client.cpp`.

**Root cause:** Commit `[REDACTED:Hex High Entropy String]` introduced a CA certificate path discovery block in `src/plugins/azure_blob/azure_blob_client.cpp` (around line 101) where the author manually padded trailing inline comments with extra spaces to vertically align them. `clang-format-19` does **not** preserve that manual alignment — it strips the extra padding and places a single space before each `//` comment. The diff reported by `clang-format-diff-19` is precise:

```diff
-    "/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo
-    "/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+    "/etc/ssl/certs/ca-certificates.crt", // Debian/Ubuntu/Gentoo
+    "/etc/pki/tls/certs/ca-bundle.crt", // Fedora/RHEL 6
     "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
-    "/etc/ssl/cert.pem"                                  // Alpine Linux
+    "/etc/ssl/cert.pem" // Alpine Linux
```

The three lines with shorter string literals have extra trailing spaces before their `//` comments to align them with the longest entry (`/etc/pki/ca-trust/...`). clang-format disallows this alignment style, causing exit code 1.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1694, branch `ca-cert-update`)

**File:** `src/plugins/azure_blob/azure_blob_client.cpp` — lines ~101–108 (the `ca_info` path-discovery initializer list)

**Suggested fix:** Remove the manual whitespace padding from the three inline comments so they each have exactly one space before `//`. The corrected block should read:

```cpp
for (const auto *path : {
        "/etc/ssl/certs/ca-certificates.crt", // Debian/Ubuntu/Gentoo
        "/etc/pki/tls/certs/ca-bundle.crt", // Fedora/RHEL 6
        "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
        "/etc/ssl/cert.pem" // Alpine Linux
    }) {
```

Alternatively, run `clang-format-19 -i src/plugins/azure_blob/azure_blob_client.cpp` locally (with the project's `.clang-format` style file active) before pushing, to auto-correct the formatting in one step.

**Related:** [PR #1694 — Update CA cert discovery](https://github.com/ai-dynamo/nixl/pull/1694)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `71fceb77-4959-4565-9967-ec058f147ba6` in the triage console for the audit trail.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,9 +96,27 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
...
+
+    {
         Azure::Core::Http::CurlTransportOptions curlOptions;
-        curlOptions.CAInfo = caBundle;
+        const char *ca_info = caBundle.empty() ? nullptr : caBundle.c_str();
+
+        if (!ca_info) {
+            for (const auto *path : {
+                     "/etc/ssl/certs/ca-certificates.crt", // Debian/Ubuntu/Gentoo
+                 }) {
```
kyleknap: For this, I'm thinking:

1. Let's avoid doing a for loop and just look for this one specific location
2. Add a comment saying why we have it there and hint that we could potentially remove it. I think something along the lines of the following would work 
```
Libcurl currently looks for CA bundles based on the platform that is built (currently CentOS) and not the
platform that it is running on. This means it will not look for certs in the correct location on Ubuntu. This
is a workaround to make sure we check for Ubuntu certs before falling back to libcurl's default location.
In the future, we can remove this check if we find a way to build libcurl to search for certs in a more
cross-distro compatible way.
```

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,9 +96,27 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
...
+                if (FILE *f = fopen(path, "r")) {
```
kyleknap: Instead of trying to read the file, I'm thinking we just check whether the file exists (e.g., `std::filesystem::exists`). Mainly it will make it a bit simpler and we don't have to worry about managing file handles as in most cases if the file is there it is most likely going to be readable as it is the distros managing the certs at that location.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,9 +96,27 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
     options.Telemetry.ApplicationId = "azpartner-nixl/0.1.0";
 
     std::string caBundle = ::getCaBundle(custom_params);
```
kyleknap, line 112: Instead of updating the `if` statement could we add this additional fallback logic to the `getCaBundle` to help consolidate all CA bundle related resolution logic to one place?

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -64,7 +65,9 @@ getConnectionString(nixl_b_params_t *custom_params) {
         }
     }
     const char *env_conn = std::getenv("AZURE_STORAGE_CONNECTION_STRING");
-    if (env_conn && env_conn[0] != '\0') return std::string(env_conn);
+    if (env_conn && env_conn[0] != '\0') {
+        return std::string(env_conn);
```
kyleknap: Was there a particular reason we needed to update this line? If we do need to update it, it may make sense to update it to the pattern of the rest of these that use a `nixl::config` helper as I noticed this was the only parameter did not get updated when this file was updated and don't think it was intentional.

kyleknap (commented):
Looks good. I just had some smaller suggestions.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,9 +96,27 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
...
-        curlOptions.CAInfo = caBundle;
+        const char *ca_info = caBundle.empty() ? nullptr : caBundle.c_str();
+
+        if (!ca_info) {
+            for (const auto *path : {
+                     "/etc/ssl/certs/ca-certificates.crt", // Debian/Ubuntu/Gentoo
+                 }) {
+                if (FILE *f = fopen(path, "r")) {
+                    fclose(f);
+                    ca_info = path;
```
kyleknap: Also if we do detect the file's existence and decide to use it, we should add a logging statement using the `NIXL_DEBUG` macro so it is easier to tell if we resolved to this CA bundle based on this hardcoded path or that is what libcurl is defaulting to.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,9 +108,15 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
...
+
+    {
         Azure::Core::Http::CurlTransportOptions curlOptions;
-        curlOptions.CAInfo = caBundle;
+        const char *ca_info = caBundle.empty() ? nullptr : caBundle.c_str();
+
+        if (ca_info) {
+            curlOptions.CAInfo = ca_info;
+        }
+
```
kyleknap: Is it possible to not update these lines at all? Mainly if we don't make changes here, we can keep it localized to `getCaBundle()`.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -78,7 +78,22 @@ getCaBundle(nixl_b_params_t *custom_params) {
...
+    // Libcurl currently looks for CA bundles based on the platform that is built (currently CentOS)
+    // and not the platform that it is running on. This means it will not look for certs in the
+    // correct location on Ubuntu. This is a workaround to make sure we check for Ubuntu certs
+    // before falling back to libcurl's default location. In the future, we can remove this check if
+    // we find a way to build libcurl to search for certs in a more cross-distro compatible way.
+    if (std::filesystem::exists("/etc/ssl/certs/ca-certificates.crt")) {
+        NIXL_DEBUG << "Using CA bundle: /etc/ssl/certs/ca-certificates.crt";
+        return "/etc/ssl/certs/ca-certificates.crt";
+    }
```
kyleknap: Can we set the location value once and reuse it and be a little more descriptive in the debug message to better indicate why we are using it? e.g.:
```cpp
const std::string ubuntu_ca_bundle = "/etc/ssl/certs/ca-certificates.crt";
if (std::filesystem::exists(ubuntu_ca_bundle)) {
    NIXL_DEBUG << "Using detected CA bundle at: " << ubuntu_ca_bundle;
    return ubuntu_ca_bundle;
}
```

kyleknap (commented):
It's looking good. Just a couple more suggestions.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -93,6 +109,7 @@ azureBlobClient::azureBlobClient(nixl_b_params_t *custom_params,
     options.Telemetry.ApplicationId = "azpartner-nixl/0.1.0";
 
     std::string caBundle = ::getCaBundle(custom_params);
+    Azure::Core::Http::CurlTransportOptions curlOptions;
```
kyleknap: Let's remove this line as well. It does not appear to be needed.

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -78,7 +78,23 @@ getCaBundle(nixl_b_params_t *custom_params) {
...
+    }
+
+    // Libcurl currently looks for CA bundles based on the platform that is built (currently CentOS)
+    // and not the platform that it is running on. This means it will not look for certs in the
+    // correct location on Ubuntu. This is a workaround to make sure we check for Ubuntu certs
+    // before falling back to libcurl's default location. In the future, we can remove this check if
+    // we find a way to build libcurl to search for certs in a more cross-distro compatible way.
+    const std::string ubuntu_ca_bundle = "/etc/ssl/certs/ca-certificates.crt";
+    if (std::filesystem::exists(ubuntu_ca_bundle)) {
+        NIXL_DEBUG << "Using detected CA bundle at: " << ubuntu_ca_bundle;
```
kyleknap, line 94: Looking more into this, it might make sense to use `NIXL_INFO` macro instead? Mainly it seems to better fit the [contributing guidelines](https://github.com/ai-dynamo/nixl/blob/main/CONTRIBUTING.md#error-handling) as "configuration loaded" and when building from the container, which installs the wheels, I'm only seeing this log message if `NIXL_INFO` is used. @aranadive Thoughts?

kyleknap (commented):
Looks good to me! :ship: I also was able to confirm that we are now leveraging this new code path to detect CA certs on Ubuntu. @aranadive this should be good to review on your end.

kyleknap (approved):

src/plugins/azure_blob/azure_blob_client.cpp
```diff
@@ -78,7 +78,23 @@ getCaBundle(nixl_b_params_t *custom_params) {
...
+    if (!ca_bundle.empty()) {
+        return ca_bundle;
+    }
+
+    // Libcurl currently looks for CA bundles based on the platform that is built (currently CentOS)
+    // and not the platform that it is running on. This means it will not look for certs in the
+    // correct location on Ubuntu. This is a workaround to make sure we check for Ubuntu certs
+    // before falling back to libcurl's default location. In the future, we can remove this check if
+    // we find a way to build libcurl to search for certs in a more cross-distro compatible way.
+    const std::string ubuntu_ca_bundle = "/etc/ssl/certs/ca-certificates.crt";
```
aranadive: Maybe define this as inline constexpr std::string_view within the azureblobclient class? So its scoped within the client.

aranadive (dismissed):

dpressle:
@ci-triage-agent

dpressle:
👀 Investigating [`nixl-ci-dl-gpu` build 705](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-dl-gpu/detail/nixl-ci-dl-gpu/705/pipeline/)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-dl-gpu` · commit `9f39806c`

I now have a complete picture of the failure. Here is the full diagnosis:

---

**Summary:** `TestTransfer.remoteMDFromSocket` fails across all 4 UCX parameterizations because `ibv_reg_dmabuf_mr` returns **error 524** (`EREMOTEIO`), causing `ucp_mem_map` to fail for CUDA VRAM buffers on `mlx5_7`, which prevents memory registration and cascades into `loadRemoteMD` returning `NIXL_ERR_NOT_FOUND`.

---

**Root cause:**

The failure is a **hardware/driver-level DMA-buf registration error**, not a code logic bug. The chain is:

1. `nixlUcxContext::memReg()` calls `ucp_mem_map()` on CUDA VRAM addresses using the DMA-buf path (`ibv_reg_dmabuf_mr`).
2. UCX's `ib_md.c` calls `ibv_reg_dmabuf_mr(address=..., length=16384, access=0x10000f)` on `mlx5_7` and gets **`Unknown error 524`** (`EREMOTEIO` — i.e. the kernel returned an I/O error from the DMA-buf subsystem). This reproduces identically across **all three test runs** on host `gb200-nvl4-ts2-102`.
3. Because `ucp_mem_map` fails, `nixl_agent.cpp:468` reports `registerMem: registration failed for the specified or all potential backends`.
4. With no successfully registered memory, the sending agent never advertises valid remote keys. When the receiving side calls `loadRemoteMD` via the socket listener (`nixl_listener.cpp:680`), it gets `NIXL_ERR_NOT_FOUND` because the memory descriptors in the exchanged blob are invalid/empty.
5. This propagates to:
   - `test_transfer.cpp:254` — `wait_until_true(checkRemoteMD)` → `false`
   - `test_transfer.cpp:402` — `createXferReq` → `-4` (`NIXL_ERR_NOT_FOUND`)
   - `test_transfer.cpp:288/330` — transfer and notification counts both fail

A key observation: the warning **"2 IB device(s) were detected, but accelerated IB support was not found! Performance may be degraded"** appears immediately before the `ibv_reg_dmabuf_mr` failure in every test run. This points directly to PR **#1637** (`PLUGINS/UCX: set UCX_MAX_HCA_PER_GPU to auto`, merged 2026-05-15, commit `71e68170bb8a`), which introduced the line:

```cpp
// src/plugins/ucx/ucx_utils.cpp ~line 113
config.modify("MAX_HCA_PER_GPU", "auto");   // only for UCX >= 1.21
```

Setting `MAX_HCA_PER_GPU=auto` changes how UCX assigns IB HCAs to GPUs. On this specific `gb200-nvl4-ts2-102` node, the `auto` heuristic appears to assign `mlx5_7` — an HCA that **does not support DMA-buf** for the CUDA memory type — as the primary HCA for that GPU. The previous default (a fixed value or a different HCA selection) avoided `mlx5_7`, so DMA-buf registration succeeded. The CI was bumped to UCX v1.21.x (PR #1507, 2026-04-26), so the version guard `ucpVersion_ >= UCP_VERSION(1, 21)` is now always true in the test environment.

---

**Implicated commit:** `71e68170bb8a` — *tomerg-nvidia*, "PLUGINS/UCX: set UCX_MAX_HCA_PER_GPU to auto (#1637)"

**File:** `src/plugins/ucx/ucx_utils.cpp` ~line 113 (the `config.modify("MAX_HCA_PER_GPU", "auto")` call)

**Suggested fix:**

1. **Immediate / safe**: Revert or guard the `MAX_HCA_PER_GPU=auto` setting with an additional check to exclude HCAs that don't support DMA-buf/RDMA to GPU memory. One approach is to fall back to the previous default if `warnAboutHardwareSupportMismatch()` detects `UCS_MEMORY_TYPE_RDMA` is not available:
   ```cpp
   if (ucpVersion_ >= UCP_VERSION(1, 21)) {
       // Only set auto if accelerated IB (DMA-buf RDMA) is confirmed available
       if (hw_info.numIbDevices > 0 && /* rdma mem type supported */) {
           config.modify("MAX_HCA_PER_GPU", "auto");
       }
   }
   ```
2. **Investigate node state**: On `gb200-nvl4-ts2-102`, check whether `mlx5_7`'s peer memory / DMA-buf kernel module is loaded and functional (`ibdev2netdev`, `nvidia-peermem` module status, `dmesg` for DMA-buf errors). Error 524 from the kernel usually means the peer-memory or DMA-buf driver returned `EIO` — this could also indicate a transient driver/hardware issue on this specific node.
3. **If it's a node issue**: Re-run the build on a healthy node to confirm, then file a ticket to remediate `gb200-nvl4-ts2-102`.

**Related:** [PR #1637](https://github.com/ai-dynamo/nixl/pull/1637) (UCX_MAX_HCA_PER_GPU=auto), [PR #1507](https://github.com/ai-dynamo/nixl/pull/1507) (UCX bump to v1.21.x)

src/plugins/azure_blob/azure_blob_client.cpp
aranadive: Lets leave at DEBUG for now. @kyleknap is it not showing up if NIXL_DEBUG is used?

kyleknap: @aranadive no it was not. It was strange that basically when using `NIXL_DEBUG` from within the plugin, it would not emit any debug-level messages even after setting `NIXL_LOG_LEVEL=DEBUG`, even though the `configuration.cpp` related logs seemed to be debug-level too but still printed out e.g.,
```
I0529 23:26:07.185265     262 configuration.cpp:92] Missing environment variable AZURE_STORAGE_CONNECTION_STRING
```

I noticed that this warning was being emitted and seemed correlated to when I would have expected the debug-related from the plugin to be emitted:
```
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
```

The only thing that got it working was switching the line to `NIXL_INFO`. Have not dug into this past that. Not sure if there is a setup/configuration we are missing? Thoughts on what it could be?

I do think it would be good to be emitting some type of log here so that we can differentiate in whether our logic chose the CA bundle location or delegated it to libcurl.

aranadive (approved):

PR #1697, opened 2026-05-29
fix(posix): surface and recover from short writes in all io queues

### LINKED ISSUE #1696: POSIX backend (libaio & io_uring): ENOSPC short writes silently reported as DONE (data loss)
## Summary

## Affected code

```cpp

        return NIXL_ERR_BACKEND;
    }
    completed_ios.push_back(io);
}
```

```cpp
}
```

```cpp
    ...
    return NIXL_ERR_BACKEND;
}
```

## Reproduction

```
```

Minimal reproducer:

```python
import os, resource

size = 1 << 20
nixl_utils.ba_buf(src, size)

agent.transfer(h)
    pass
```

## Environment

## Suggested fix

lluki:
## What

The POSIX backend's libaio (`AIO`, the Linux default) and io_uring (`URING`) I/O queues treated any **non-negative** AIO completion result as a fully successful transfer. When a write runs out of space, the kernel commonly reports a **short write** -- a positive byte count smaller than the requested length -- rather than a negative errno. That short count slipped past the error check, so the transfer was reported as `DONE` while only part of the buffer reached disk: silent data loss with no error surfaced to the caller.

For libaio there was an additional latent bug: `io_event.res` is an `unsigned long`, so the existing `events[i].res < 0` check is always false -- even a fully failed write was never detected.

This PR makes the POSIX backend both **detect** short/failed writes and **cleanly recover** from them across all three io queues.

Fixes #1696.

Note, unlike previous versions of this PR the current version does not include the fix for double dereg anymore. We have separate tracking bugs now for these issues: #1749 and #1792

## Detect

Treat a completion as a fault when its result doesn't equal the requested length (covers both a negative errno and a short write):

- `linux_aio_io_queue.cpp` -- read `events[i].res` as signed (`long`) and compare to `iocb->u.c.nbytes`.
- `io_uring_io_queue.cpp` -- compare `cqe->res` to the requested `len`.
- `posix_aio_io_queue.cpp` already did this (`ret != aio_nbytes`).

## Recover (drain to completion)

The io queue is a **single instance shared across all transfers on an engine**, so the previous "early-return `NIXL_ERR_BACKEND` on the first bad completion" left the queue dirty -- un-reaped in-flight ios and stale completions that the *next* transfer would misattribute (and, since the failed request handle had been released, could reference freed state).

The error path now mirrors the success path instead of short-circuiting:

- A per-io fault is **reported through the completion callback** (a non-zero `error` argument), and the io is recycled exactly like a successful one. `doCheckCompleted` keeps draining; it never early-returns for a per-io fault.
- The request handle records the fault (`any_failed_`), and **`checkXfer()` only renders the verdict once every io of the transfer has terminated** -- returning `NIXL_ERR_BACKEND` if any faulted, else `NIXL_SUCCESS`. (A queue `poll()` returning negative is now reserved for submission / `io_getevents`-level failures.)

This guarantees that after a partial fault the shared queue is fully drained and clean for the next transfer -- no leaked pool entries, no stale completions, no freed-handle references. It also covers all completion-time failures uniformly, not just short writes (a negative `-EIO` recovers identically).

## Behavior change on reads

This PR makes a completion whose transferred byte count differs from the requested length a hard failure (`NIXL_ERR_BACKEND`) in the `libaio` and `io_uring` queues, matching what the POSIX-aio queue already did. Previously these two queues only flagged a negative result (and for `libaio` that check was effectively dead code, since `io_event.res` is an unsigned long), so a short read, for example a READ whose `(offset, len)` extends to or past end-of-file and returns fewer than `len` bytes, was silently reported as DONE. With this change such a short read now surfaces as an error. This is intentional and consistent across all three queues, but callers that rely on partial/short reads (reading a region beyond EOF) will now see an error instead of a truncated-but-successful transfer; transfers are treated as exact-length, all-or-nothing.

## Test

Adds `test_short_write_enospc` to `test/unit/plugins/posix/nixl_posix_test.cpp`. It loops over every available io queue (`use_aio`, `use_uring`, `use_posix_aio`), skipping any the build/runtime doesn't provide, and over **both single- and multi-descriptor (8) transfers**, and for each:

1. forces a short write deterministically by lowering the **soft** `RLIMIT_FSIZE` below the transfer size (no special filesystem or privileges; hard limit preserved so it can be restored),
2. asserts the transfer surfaces an error rather than `DONE`, and
3. asserts a subsequent **normal** transfer on the **same agent** still succeeds.

The multi-descriptor case is what exercises the drain path: every io in the batch faults, so the queue must reap and recycle all of them and stay usable. All queues pass detect + recover for both 1 and 8 descriptors; the existing read/write + repost test is unaffected.

### DIFF
--- a/src/plugins/posix/io_uring_io_queue.cpp
+++ b/src/plugins/posix/io_uring_io_queue.cpp
@@ -117,14 +117,15 @@ nixlPosixIOQueueUring::doCheckCompleted(void) {
     io_uring_for_each_cqe(&uring, head, cqe) {
         int res = cqe->res;
         nixlPosixIoUringIO *io = reinterpret_cast<nixlPosixIoUringIO *>(io_uring_cqe_get_data(cqe));
+        int error = (res < 0 || static_cast<size_t>(res) != io->len_);
+        if (error) {
+            NIXL_ERROR << absl::StrFormat(
+                "IO operation incomplete: result %d, expected %zu", res, io->len_);
+        }
         if (io->clb_) {
-            io->clb_(io->ctx_, res, 0);
+            io->clb_(io->ctx_, error ? 0 : static_cast<uint32_t>(res), error);
         }
         free_ios_.push_back(io);
-        if (res < 0) {
-            NIXL_ERROR << absl::StrFormat("IO operation failed: %s", nixl_strerror(-res));
-            return NIXL_ERR_BACKEND;
-        }
         count++;
         if (count == MAX_IO_CHECK_COMPLETED_BATCH_SIZE) {
             break;
--- a/src/plugins/posix/io_uring_io_queue.cpp
+++ b/src/plugins/posix/io_uring_io_queue.cpp
@@ -33,6 +33,7 @@ public:
     nixlPosixIOQueueDoneCb clb_;
     void *ctx_;
     struct io_uring_sqe *sqe_;
+    bool in_flight_; // submitted to the kernel, not yet reaped (scopes cancel)
 };
 
 class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
@@ -51,6 +52,8 @@ public:
             void *ctx) override;
     virtual nixl_status_t
     poll(void) override;
+    virtual nixl_status_t
+    cancel(void *ctx) override;
     virtual ~nixlPosixIOQueueUring() override;
 
 protected:
@@ -98,6 +101,7 @@ nixlPosixIOQueueUring::post(void) {
         }
 
         io_uring_sqe_set_data(sqe, io);
+        io->in_flight_ = true;
     }
 
     int ret = io_uring_submit(&uring);
@@ -117,14 +121,24 @@ nixlPosixIOQueueUring::doCheckCompleted(void) {
     io_uring_for_each_cqe(&uring, head, cqe) {
         int res = cqe->res;
         nixlPosixIoUringIO *io = reinterpret_cast<nixlPosixIoUringIO *>(io_uring_cqe_get_data(cqe));
+        // cancel SQEs carry a null sentinel user_data; reap the completion but
+        // do not treat it as an io.
+        if (!io) {
+            count++;
+            if (count == MAX_IO_CHECK_COMPLETED_BATCH_SIZE) {
+                break;
+            }
+            continue;
+        }
         int error = (res < 0 || static_cast<size_t>(res) != io->len_);
         if (error) {
-            NIXL_ERROR << absl::StrFormat(
+            NIXL_DEBUG << absl::StrFormat(
                 "IO operation incomplete: result %d, expected %zu", res, io->len_);
         }
         if (io->clb_) {
             io->clb_(io->ctx_, error ? 0 : static_cast<uint32_t>(res), error);
         }
+        io->in_flight_ = false;
         free_ios_.push_back(io);
         count++;
         if (count == MAX_IO_CHECK_COMPLETED_BATCH_SIZE) {
@@ -164,6 +178,7 @@ nixlPosixIOQueueUring::enqueue(int fd,
     io->read_ = read;
     io->clb_ = clb;
     io->ctx_ = ctx;
+    io->in_flight_ = false;
 
     ios_to_submit_.push_back(io);
 
@@ -178,6 +193,31 @@ nixlPosixIOQueueUring::poll(void) {
     }
 
     return doCheckCompleted();
+}
+
+nixl_status_t
+nixlPosixIOQueueUring::cancel(void *ctx) {
+    // drop this transfer's un-submitted ios
+    nixlPosixIOQueueImpl<nixlPosixIoUringIO>::cancel(ctx);
+
+    // best-effort: cancel this transfer's in-flight ios. The cancel SQE carries a null
+    // sentinel (skipped in doCheckCompleted); the target io completes and is reaped normally.
+    bool submit_cancels = false;
+    for (auto &io : ios_) {
+        if (io.in_flight_ && io.ctx_ == ctx) {
+            struct io_uring_sqe *sqe = io_uring_get_sqe(&uring);
+            if (!sqe) {
+                break; // SQ full; the remaining in-flight ios drain on their own
+            }
+            io_uring_prep_cancel(sqe, &io, 0);
+            io_uring_sqe_set_data(sqe, nullptr);
+            submit_cancels = true;
+        }
+    }
+    if (submit_cancels) {
+        io_uring_submit(&uring);
+    }
+    return NIXL_SUCCESS;
 }
 
 nixlPosixIOQueueUring::~nixlPosixIOQueueUring() {
--- a/src/plugins/posix/linux_aio_io_queue.cpp
+++ b/src/plugins/posix/linux_aio_io_queue.cpp
@@ -160,15 +160,17 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
         struct iocb *iocb = events[i].obj;
         nixlPosixLinuxAioIO *io = (nixlPosixLinuxAioIO *)iocb->data;
 
-        if (events[i].res < 0) {
-            NIXL_ERROR << "AIO operation failed: " << events[i].res;
-            return NIXL_ERR_BACKEND;
+        // io_event.res is unsigned long: read as signed to catch a negative errno;
+        // a short completion (res != len) is a fault too.
+        long res = static_cast<long>(events[i].res);
+        const int error = (res < 0 || static_cast<unsigned long>(res) != iocb->u.c.nbytes);
+        if (error) {
+            NIXL_ERROR << absl::StrFormat(
+                "AIO operation incomplete: result %ld, expected %lu", res, iocb->u.c.nbytes);
         }
-
         if (io->clb_) {
-            io->clb_(io->ctx_, events[i].res, 0);
+            io->clb_(io->ctx_, error ? 0 : static_cast<uint32_t>(res), error);
         }
-
         completed_ios.push_back(io);
     }
 
--- a/src/plugins/posix/linux_aio_io_queue.cpp
+++ b/src/plugins/posix/linux_aio_io_queue.cpp
@@ -165,7 +165,7 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
         long res = static_cast<long>(events[i].res);
         const int error = (res < 0 || static_cast<unsigned long>(res) != iocb->u.c.nbytes);
         if (error) {
-            NIXL_ERROR << absl::StrFormat(
+            NIXL_DEBUG << absl::StrFormat(
                 "AIO operation incomplete: result %ld, expected %lu", res, iocb->u.c.nbytes);
         }
         if (io->clb_) {
--- a/src/plugins/posix/posix_aio_io_queue.cpp
+++ b/src/plugins/posix/posix_aio_io_queue.cpp
@@ -131,35 +131,33 @@ nixlPosixIOQueueAIO::doCheckCompleted(void) {
     }
 
     int num_ios = std::min(MAX_IO_CHECK_COMPLETED_BATCH_SIZE, (int)ios_in_flight_.size());
-    for (auto it = ios_in_flight_.begin(); it != ios_in_flight_.end();) {
+    for (auto it = ios_in_flight_.begin(); it != ios_in_flight_.end() && num_ios > 0;) {
         nixlPosixAioIO *io = *it;
         int status = aio_error(&io->aio_);
-        if (status == 0) {
-            ssize_t ret = aio_return(&io->aio_);
-            if (ret < 0 || ret != static_cast<ssize_t>(io->aio_.aio_nbytes)) {
-                NIXL_ERROR << "aio_return failed: " << nixl_strerror(-ret);
-                ios_in_flight_.push_front(io);
-                return NIXL_ERR_BACKEND;
-            }
-            if (io->clb_) {
-                io->clb_(io->ctx_, ret, 0);
-            }
-            it = ios_in_flight_.erase(it);
-            free_ios_.push_back(io);
-        } else if (status == EINPROGRESS) {
-            return NIXL_IN_PROG;
-        } else {
+        if (status == EINPROGRESS) {
+            ++it; // still pending; check the rest of the batch
+            continue;
+        }
+
+        int error;
+        ssize_t ret = 0;
+        if (status != 0) {
             NIXL_ERROR << "aio_error failed: " << nixl_strerror(-status);
-            ios_in_flight_.push_front(io);
-            return NIXL_ERR_BACKEND;
+            error = 1;
+        } else {
+            ret = aio_return(&io->aio_);
+            error = (ret < 0 || ret != static_cast<ssize_t>(io->aio_.aio_nbytes));
+            if (error) {
+                NIXL_ERROR << "aio_return incomplete: " << ret << " expected "
+                           << io->aio_.aio_nbytes;
+            }
         }
-
-        it++;
-
+        if (io->clb_) {
+            io->clb_(io->ctx_, error ? 0 : static_cast<uint32_t>(ret), error);
+        }
+        it = ios_in_flight_.erase(it);
+        free_ios_.push_back(io);
         num_ios--;
-        if (num_ios == 0) {
-            break;
-        }
     }
 
     return ios_in_flight_.empty() ? NIXL_SUCCESS : NIXL_IN_PROG;
--- a/src/plugins/posix/posix_aio_io_queue.cpp
+++ b/src/plugins/posix/posix_aio_io_queue.cpp
@@ -142,13 +142,13 @@ nixlPosixIOQueueAIO::doCheckCompleted(void) {
         int error;
         ssize_t ret = 0;
         if (status != 0) {
-            NIXL_ERROR << "aio_error failed: " << nixl_strerror(-status);
+            NIXL_DEBUG << "aio_error failed: " << nixl_strerror(-status);
             error = 1;
         } else {
             ret = aio_return(&io->aio_);
             error = (ret < 0 || ret != static_cast<ssize_t>(io->aio_.aio_nbytes));
             if (error) {
-                NIXL_ERROR << "aio_return incomplete: " << ret << " expected "
+                NIXL_DEBUG << "aio_return incomplete: " << ret << " expected "
                            << io->aio_.aio_nbytes;
             }
         }
--- a/src/plugins/posix/posix_backend.cpp
+++ b/src/plugins/posix/posix_backend.cpp
@@ -130,6 +130,7 @@ nixlPosixBackendReqH::nixlPosixBackendReqH(const nixl_xfer_op_t &op,
       remote(rem),
       queue_depth_(loc.descCount()),
       num_confirmed_ios_(queue_depth_),
+      any_failed_(false),
       io_queue_(io_queue) {
     NIXL_ASSERT(local.descCount());
     NIXL_ASSERT(remote.descCount());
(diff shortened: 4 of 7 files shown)


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26642760371/job/78520023486)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `55c9b0a3`

The file is new on the PR branch, so it's not available in the base tree — that's fine. The log already contains the full diff needed to diagnose. Here is the complete diagnosis:

---

**Summary:** `clang-format-19` reports three style violations in the new test file `test/unit/plugins/posix/nixl_posix_enospc_test.cpp` on PR #1697, causing the Clang Format Check to exit with code 1.

**Root cause:** The PR introduced `test/unit/plugins/posix/nixl_posix_enospc_test.cpp` with three formatting issues that violate the project's `.clang-format` rules:

1. **Lines 55–56 — Trailing comment alignment:** Two comments were padded to align with extra spaces:
   ```cpp
   // before (violates style):
   const size_t cap = page_size;            // file may grow to one page
   const size_t transfer_size = 4 * page_size;  // we try to write four

   // after (required by clang-format):
   const size_t cap = page_size; // file may grow to one page
   const size_t transfer_size = 4 * page_size; // we try to write four
   ```

2. **Line ~152 — Single-statement `if` must use braces:**
   ```cpp
   // before:
   if (treq) agent.releaseXferReq(treq);

   // after:
   if (treq) {
       agent.releaseXferReq(treq);
   }
   ```

3. **Line ~176 — Single-statement `if` must use braces:**
   ```cpp
   // before:
   if (r == 1 || r == 2) failures++;

   // after:
   if (r == 1 || r == 2) {
       failures++;
   }
   ```

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `chore/posix-enospc-short-write`, PR #1697)

**File:** `test/unit/plugins/posix/nixl_posix_enospc_test.cpp` — lines ~55–56, ~152, ~176

**Suggested fix:** Run `clang-format-19` in-place on the file before pushing — either manually or via a pre-commit hook:
```bash
clang-format-19 -i -style=file test/unit/plugins/posix/nixl_posix_enospc_test.cpp
```
This will automatically:
- Strip the extra alignment spaces before the inline comments on lines 55–56
- Wrap the two single-statement `if` bodies (lines ~152 and ~176) in braces

Then `git add` and amend/push the commit.

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `8550e118-3fb8-41db-9df8-218dc16d97da` in the triage console for the audit trail.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -160,8 +160,12 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
         struct iocb *iocb = events[i].obj;
         nixlPosixLinuxAioIO *io = (nixlPosixLinuxAioIO *)iocb->data;
 
-        if (events[i].res < 0) {
-            NIXL_ERROR << "AIO operation failed: " << events[i].res;
+        // io_event.res is an unsigned long, so read it as signed to see a
+        // negative errno; a short completion (res != len) must fail too.
+        long res = static_cast<long>(events[i].res);
```
lluki, line 165: The kernel [UAPI declares io_event.res as a signed](https://github.com/torvalds/linux/blob/master/include/uapi/linux/aio_abi.h#L60-L65) __s64 - the syscall return value, i.e. bytes transferred or a negative errno - but libaio's userspace header redeclares the [same field as an unsigned long](https://github.com/yugabyte/libaio/blob/master/src/libaio.h#L142-L147) .  The `static_cast<long>` simply restores the kernel's intended signedness so the `< 0 / != len` comparison behaves correctly.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -160,8 +160,12 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
...
+        if (res < 0 || static_cast<unsigned long>(res) != iocb->u.c.nbytes) {
+            NIXL_ERROR << absl::StrFormat(
+                "AIO operation incomplete: result %ld, expected %lu", res, iocb->u.c.nbytes);
             return NIXL_ERR_BACKEND;
```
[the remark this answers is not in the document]
lluki: Not true. completion handler should not be invoked in error case.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -731,6 +733,134 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+short_write_case(const std::string &test_files_dir_path_abs_path,
+                 const std::string &queue,
+                 const std::string &param_key) {
+    print_segment_title(phase_title("Short-write / ENOSPC: " + queue));
+
+    const size_t cap = page_size;
+    const size_t transfer_size = 4 * page_size;
+
+    nixl_b_params_t params;
+    params[param_key] = "true";
```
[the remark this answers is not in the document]
lluki: The params are just default initialized one line above, how does this comment make any sense?

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -731,6 +733,134 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+    agent.deregisterMem(file_for_posix);
+    agent.deregisterMem(dram_for_posix);
+
+    if (status >= 0) {
+        std::cerr << queue << ": short write reported as success -- ENOSPC not propagated"
+                  << std::endl;
+        return 1;
+    }
+    std::cout << queue << ": short write correctly surfaced as error" << std::endl;
+    return 0;
```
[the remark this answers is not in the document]
lluki: As long as we transport an error to the caller, we should be good. no need to overdo it

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -121,8 +121,10 @@ nixlPosixIOQueueUring::doCheckCompleted(void) {
             io->clb_(io->ctx_, res, 0);
         }
         free_ios_.push_back(io);
-        if (res < 0) {
-            NIXL_ERROR << absl::StrFormat("IO operation failed: %s", nixl_strerror(-res));
+        // short completion (res < len) must fail too, e.g. ENOSPC mid-write
+        if (res < 0 || static_cast<size_t>(res) != io->len_) {
+            NIXL_ERROR << absl::StrFormat(
+                "IO operation incomplete: result %d, expected %zu", res, io->len_);
             return NIXL_ERR_BACKEND;
```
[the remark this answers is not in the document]
lluki: Good point. Adressed it.

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -131,35 +131,33 @@ nixlPosixIOQueueAIO::doCheckCompleted(void) {
     }
 
     int num_ios = std::min(MAX_IO_CHECK_COMPLETED_BATCH_SIZE, (int)ios_in_flight_.size());
-    for (auto it = ios_in_flight_.begin(); it != ios_in_flight_.end();) {
+    for (auto it = ios_in_flight_.begin(); it != ios_in_flight_.end() && num_ios > 0;) {
```
ofer, line 134: Since MAX_IO.. > 0, if ios_in_flight = 0, it should still immediately reach .end(), why is num_ios > 0 needed?

ofer: Never mind, I see the code below

src/plugins/posix/posix_backend.cpp
ofer, line 136: initialize any_failed_(false) in the constructor as well

src/plugins/posix/posix_backend.h
```diff
@@ -35,6 +35,7 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
     const nixl_meta_dlist_t &remote; // Remote memory descriptor list
     const int queue_depth_; // Queue depth for async I/O
     int num_confirmed_ios_; // Number of confirmed IOs
+    bool any_failed_ = false; // Set if any io of the current transfer failed
```
ofer: set any_failed_ to false in the constructor (like the others were initialized as well) for consistency

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -117,14 +117,15 @@ nixlPosixIOQueueUring::doCheckCompleted(void) {
...
+        }
         if (io->clb_) {
-            io->clb_(io->ctx_, res, 0);
+            io->clb_(io->ctx_, error ? 0 : static_cast<uint32_t>(res), error);
         }
         free_ios_.push_back(io);
-        if (res < 0) {
-            NIXL_ERROR << absl::StrFormat("IO operation failed: %s", nixl_strerror(-res));
-            return NIXL_ERR_BACKEND;
-        }
```
ofer, line 127: Removing this changes behavior, even though the error goes through the callback, still need to check if the new continue till complete then report the error is desired instead of reporting an error and cancelling quickly

ofer: You should return an error if there's a failure in transfer, you can wait for all of the completion events to drain and throw the error below, but the error should be returned.

lluki: I deliberately removed this check. With this check, on error, we get a corrupted queue that is never able to perform transfers again: An early return here never calls `io_uring_cq_advance` at line 136. So the next `doCheckCompleted` wil hit the failing element again - with no way for the caller to repair this problem. 

The PR also adds a regression test that checks that a northbound API client can after a failure retry and has a working transfer. 

But yes, I can't 100% promise that there is no caller that relies on the behavior that "fails once implies fails always".

lluki:
Worth a consideration: Should NIXL handle partial writes and retries internally or surface it to the caller? This PR surfaces it as an error. With this PR, the caller can then resubmit the whole transfer batch, but with the NIXL API, the caller can not learn about partial writes/reads. The caller just sees `NIXL_ERR_BACKEND`. That means there is no way to resume a partial transfer and a potentially expensive transfer has to be done again.

src/plugins/posix/posix_backend.h
lluki: done

src/plugins/posix/posix_backend.cpp
lluki: done

src/plugins/posix/io_uring_io_queue.cpp
lluki: Ok, we had a chat yesterday - Ofer's request is to keep the early error-out path, but to make sure the agent stays usable. To get both, we need to cancel the remaining requests:

 1) In `ios_to_submit_` we keep track of requests that have not been sent to the backend/OS. These are easy to cancel. However, we submit batches of 64 requests. In many cases everything will be batch-submitted in one go to the backend.
 2) There are requests that have been submitted to the OS already because they were in the same batch. Hence, we need a backend specific cancellation mechanism
     * uring: Has one
     * aio: Doesnt' have one

In both cases, i need to poll the backend until the already submitted requests come back. In uring, they will return quickly with `ECANCEL` (or similar). In AIO we have to wait until the submitted batch completes (or fails). My suggestion is to wait for these events here. What do you think @ofer , is that OK ? Or do we need to defer the poll/cleanup of the failing requests to another time/function (i.e. next time `poll()` is called, so a subsequent transfer would reap them)? Maybe `releaseXferReq` is a candidate, but i think it blocks the Northbound client in the same way as early-reaping

lluki: Ok, this is even more complicated. If we do asynchronous drain of the co-requests of a failing requests, we violate the client API: if `checkXfer` returns `NIXL_ERR_BACKEND` the client contract says the buffers wont be touched anymore. But if the cancellation and drain is asynchronous, there might still be READs pending that write into the user's buffer. I dont think there is a correct way to early signal failure without draining the other requests.

I would suggest to drain as efficient as possible. That is: Clear the unsubmitted reqs from `ios_to_submit_`, submit cancellations if the backend supports (uring does, no op in AIO). But then, as much as we hate it - we have to drain the requests to make sure we get no stale buffer accesses.

ofer: Waiting is problematic, can we cancel and use the current poll to wait for the cancellation statuses to complete??

lluki: This is backend dependent: On `uring` there is a cancel functionality and we can do this. On `aio` there is not.

Imho, the correct approach if we absolutely dont want to wait is tell the user which entries failed and which are still in flight. Then we are essentially at this proposal https://github.com/ai-dynamo/nixl/pull/1274 ... But i guess because that proposal introduces a new status code `NIXL_IN_PROG_WITH_ERR` it can't be done in a backwards compatible way.

ofer: We can return a NIXL_IN_PROG_WITH_ERR and not return the list of failures, just continue till they're all cancelled then return the backend error

lluki: I added a second commit implementing the "cancel as fast as possible" : https://github.com/ai-dynamo/nixl/pull/1697/commits/90866715f7394f6c5509bf4a3c73832d0b093eec

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26956122300/job/79533697349)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `90866715`

The logs provided complete clarity. Here is the full diagnosis:

---

**Summary:** Clang Format Check failed on PR #1697 — `test/unit/plugins/posix/nixl_posix_test.cpp` has three formatting violations detected by `clang-format-19`.

**Root cause:** The new test code introduced in commit `[REDACTED:Hex High Entropy String]` (branch `chore/posix-enospc-short-write`) contains three blocks in `test/unit/plugins/posix/nixl_posix_test.cpp` that do not conform to the project's `.clang-format` style rules:

1. **Lines ~896–897** — A lambda capture list is formatted with a manual line-break that `clang-format` wants to reflow differently:
   ```cpp
   // Current (non-conforming)
   auto add = [&](int n, size_t sz, int fd, nixl_xfer_dlist_t &dram_x,
                  nixl_xfer_dlist_t &file_x) -> int {
   // clang-format wants
   auto add =
       [&](int n, size_t sz, int fd, nixl_xfer_dlist_t &dram_x, nixl_xfer_dlist_t &file_x) -> int {
   ```

2. **Lines ~924–925** — Two single-line `if` statements without braces that must have braces:
   ```cpp
   // Current (non-conforming)
   if (add(a_desc, a_size, fd_a, dram_a, file_a) != 0) return 1;
   if (add(b_desc, b_size, fd_b, dram_b, file_b) != 0) return 1;
   // clang-format wants
   if (add(a_desc, a_size, fd_a, dram_a, file_a) != 0) {
       return 1;
   }
   if (add(b_desc, b_size, fd_b, dram_b, file_b) != 0) {
       return 1;
   }
   ```

3. **Lines ~1010–1011** — Two more single-line `if` statements without braces:
   ```cpp
   // Current (non-conforming)
   if (rc >= 0) ran++;
   if (rc == 1) failures++;
   // clang-format wants
   if (rc >= 0) {
       ran++;
   }
   if (rc == 1) {
       failures++;
   }
   ```

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch author of `chore/posix-enospc-short-write`)

**File:** `test/unit/plugins/posix/nixl_posix_test.cpp` — approximately lines 896, 924–925, and 1010–1011

**Suggested fix:** Run `clang-format-19` (or `clang-format-diff-19`) in-place on the file to auto-apply the exact diff shown in the CI log, then amend/push:
```bash
# Apply clang-format to the changed lines only (using the diff approach CI uses):
git diff HEAD^1 HEAD -- test/unit/plugins/posix/nixl_posix_test.cpp \

# Or simply reformat the whole file:
clang-format-19 -style=file -i test/unit/plugins/posix/nixl_posix_test.cpp

git add test/unit/plugins/posix/nixl_posix_test.cpp
git commit --amend --no-edit
git push --force-with-lease
```
The three changes are purely cosmetic style fixes (brace wrapping for single-line `if` bodies, and lambda indentation), so there is no functional impact.

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `5563ec22-9a47-44b1-8a2b-a3f71b508ffd` in the triage console for the audit trail.

ofer (dismissed):

src/plugins/posix/posix_backend.h
```diff
@@ -35,6 +35,8 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
     const nixl_meta_dlist_t &remote; // Remote memory descriptor list
     const int queue_depth_; // Queue depth for async I/O
     int num_confirmed_ios_; // Number of confirmed IOs
+    bool any_failed_; // Set if any io of the current transfer failed
```
ColinNV, line 42: Would it make sense to count them instead? It often ends up being the case that a count is useful at some later point...

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -160,15 +160,17 @@ nixlPosixIOQueueLinuxAIO::doCheckCompleted(void) {
         struct iocb *iocb = events[i].obj;
         nixlPosixLinuxAioIO *io = (nixlPosixLinuxAioIO *)iocb->data;
 
-        if (events[i].res < 0) {
-            NIXL_ERROR << "AIO operation failed: " << events[i].res;
-            return NIXL_ERR_BACKEND;
+        // io_event.res is unsigned long: read as signed to catch a negative errno;
+        // a short completion (res != len) is a fault too.
+        long res = static_cast<long>(events[i].res);
+        int error = (res < 0 || static_cast<unsigned long>(res) != iocb->u.c.nbytes);
```
ColinNV: `const`

lluki: done

src/plugins/posix/posix_backend.h
lluki: As of now, i dont think so. We can only return success / error anyway. Long term, I would like to have fine grained error reporting as proposed in #1274

lluki:
This PR currently is currently based on https://github.com/ai-dynamo/nixl/pull/1744 . I need that fix for CI to be green.

lluki:
Note, unlike previous versions of this PR the current version does not include the fix for double dereg anymore. We have separate tracking bugs now for these issues: #1749 and #1792

lluki:
Changed error logging to not flood the CI output: Each failed batch of xfer logs once with `NIXL_ERROR` - the per item log was degraded to `NIXL_DEBUG`

ofer (approved):

lluki:
CI is green at https://github.com/ai-dynamo/nixl/commit/80b6f9f7884cf667db1b347509a1cb62aea5270e

lluki:
@yanok

yanok:
This PR mixes two fixes:

 1. Detecting truncated/short writes as errors — this part looks correct and complete.
 2. Draining instead of short-circuiting after an error — this part is only partially fixed.

 Completion-side errors now go through the callback/accounting path, so those I/Os are drained correctly. But other error paths still return immediately — for example submission / queue-level failures — and can leave already-submitted or already-queued I/Os behind. That leaves the backend in a mixed state: some failures now drain correctly, while others can still return a terminal error before all submitted I/Os for the transfer have completed/accounted, preserving the stale-completion / dangling-context problem this PR is trying to fix.

lluki:
@yanok can you elaborate on this: 

> while others can still return a terminal error before all submitted I/Os for the transfer have completed/accounted, preserving the stale-completion / dangling-context problem this PR is trying to fix.

The idea of the PR is, if one error occurs, the whole batch can be cancelled. In the case of non-uring, we just wait until the submitted requests return (as there is no way to cancel it). In uring we fire a cancellation to submitted requests. For unsubmitted requests (but queued in NIXL) we remove the requests entirely. The queue state stays usable afterwards.

PR #1702, opened 2026-05-31
kvbench: add storage backend abstraction and etcd helpers

x41lakazam:
Foundation for storage-target benchmarking in kvbench. Adds:

- test/storage_backend.py: StorageBackend abstraction with a FilesystemBackend implementation. Manages per-rank shard files, read/write handle preparation, block-size chunking for high io_uring queue depth, and cleanup.
- test/traffic_pattern.py: extends TrafficPattern with StorageOp (READ/WRITE) and storage-specific descriptor sizing.
- runtime/etcd_rt.py: pass-through `timeout_sec` for allgather_obj, alltoall_obj, and all_reduce so callers can override the 600s default; add per-call operation counter to alltoall_obj to avoid in-flight key collisions across consecutive calls.

The benchmark runner that consumes this infrastructure (sequential_custom_traffic_perftest.py and CLI wiring) lands in a follow-up PR stacked on this one.

## What?
_Describe what this PR is doing._

## Why?
_Justification for the PR. If there is an existing issue/bug, please reference it. For
bug fixes, the 'Why?' and 'What?' can be merged into a single item._

## How?
_It is optional, but for complex PRs, please provide information about the design,
architecture, approach, etc._

## Summary by CodeRabbit

* **New Features**
  * Added a storage backend abstraction with a filesystem-backed implementation for benchmarks.
  * Traffic pattern model now supports storage-only workloads and per-rank storage operations.

* **Improvements**
  * ETCD-based coordination is more resilient to eventual-consistency delays via an initialization barrier and bounded retries.
  * Collective operations now accept timeouts, use timeout-bounded polling, and report clearer diagnostics for missing/arrived participants.

PR #1703, opened 2026-05-31
Eshukrun/kvbench storage bench

x41lakazam:
Stacked on #1702, please merge it first

## Summary by CodeRabbit

* **New Features**
  * Storage-aware benchmarking with per-rank read/write operations, storage-only and mixed RDMA+storage patterns, unified `tp_file` support, direct-IO toggles, 4K-aligned buffers, block-splitting/sharding, POSIX/GDS options, and concurrent async handles.
  * Enhanced traffic generation CLI and outputs (hit-rate, FLOPs tuning, memory type, warmup/iteration/isolation controls, richer metadata).

* **Reliability**
  * Improved ETCD-based synchronization: configurable timeouts, more robust rendezvous/polling, and clearer timeout diagnostics.

* **Documentation**
  * Expanded KVBench README with design philosophy, diagrams, updated examples, and quick reference.


dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/27123156858/job/80044921317)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `f392d0e0`

I have everything I need. Here is the full diagnosis:

---

**Summary:** PR #1703 (`eshukrun/kvbench_storage_bench`) failed the "PR Size Check" because it modifies **2,719 lines** in existing files, which is **5.4× the enforced 500-line limit**.

**Root cause:** The CI workflow runs `git show --numstat` on modified files (excluding `subprojects/`) and counts added lines. The check hard-fails with `exit 1` when that count exceeds 500. PR #1703 introduced 2,719 such lines — likely a large new benchmark implementation (`kvbench` / storage bench) — blowing past the threshold by a wide margin.

The exact check that triggered the failure (from the log):
```
❌ PR size check failed!
This PR adds 2719 lines (excluding subprojects), which exceeds the maximum of 500 lines.
Process completed with exit code 1.
```

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (author: eshukrun, branch `eshukrun/kvbench_storage_bench`)

**File:** The PR size check workflow (`.github/workflows/` — the script block containing `LINES_CHANGED=$(git show --numstat ...)` and `if [[ "$LINES_CHANGED" -gt 500 ]]`)

**Suggested fix:** The PR author has two options:

1. **Split the PR** *(preferred/policy-compliant)*: Break PR #1703 into smaller, reviewable chunks of ≤500 modified lines each — e.g. one PR for the core benchmark scaffolding, another for storage-backend integration, another for tests/docs.

2. **Request a policy exemption**: If this is a legitimate large addition (e.g. a wholly new benchmark tool that cannot be meaningfully split), open a discussion with the maintainers to either raise the limit for this PR specifically or add an exception path in the workflow (e.g. a `[skip-size-check]` label for approved oversized PRs). The check also only counts lines in *modified* files (`--diff-filter=M`), so if the code is entirely **new files** (added, not modified), the diff filter may need revisiting — worth double-checking whether the 2,719 lines are in new files vs. existing ones.

**Related:** [PR #1703 — Eshukrun/kvbench storage bench](https://github.com/ai-dynamo/nixl/pull/1703)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `daceca7b-1da5-4828-8280-893b7da66cc9` in the triage console for the audit trail.

dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/27492006971/job/81258830784)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `3044a16f`

The diagnosis is complete and unambiguous. Here's the full breakdown:

---

**Summary:** PR #1703 (`eshukrun/kvbench_storage_bench`) failed the PR Size Check because it adds **2,869 lines** of changes to existing files, far exceeding the enforced **500-line maximum**.

**Root cause:** The "PR Size Check" workflow measures lines added to modified files (excluding `subprojects/`) via `git show --numstat --diff-filter=M`. This PR's changes totaled **2,869 lines** — **5.7× over the 500-line limit** — causing the check to emit `❌ PR size check failed!` and exit with code 1. This is a policy enforcement failure, not a build or test failure. No hang, no infrastructure issue — the code ran to completion and deliberately rejected the PR.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `eshukrun/kvbench_storage_bench`)

**File:** `.github/workflows/` (PR size check workflow) — the policy gate itself is working correctly; the PR is simply too large.

**Suggested fix:** The author (`eshukrun`) should split PR #1703 into smaller, more focused pull requests, each staying under the 500-line threshold. Suggested splits for a `kvbench`/storage bench PR:
1. **Core data structures / interfaces** (new types, enums, base classes)
2. **Storage backend implementation** (the benchmark logic itself)
3. **Tests / benchmarks** (unit tests and benchmark scripts)
4. **Documentation / configuration** (README updates, config files)

If the full PR is genuinely indivisible and the team agrees to waive the size limit for this specific case, a project maintainer can add a bypass label or temporarily adjust the workflow threshold — but splitting is strongly preferred for reviewability.

**Related:** https://github.com/ai-dynamo/nixl/pull/1703



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `51d74d0b-c382-493c-a893-d5589d1b7765` in the triage console for the audit trail.

x41lakazam:
Addressed all 14 open review comments in 2874909b:

- **etcd_rt**: `all_reduce` now namespaces keys per invocation (via `ops_counter`, matching `allgather`/`alltoall`); the `NIXL_ETCD_NAMESPACE` warning is suppressed when a scheduler env var (`SLURM_JOB_ID`/`SLURM_JOBID`/`PMIX_NAMESPACE`) is available for auto-namespacing.
- **main.py**: reject unknown/duplicate TP-file sections; reject `tp_file` mixed with legacy `matrix_file`/`matrix`/`storage`; reuse `parse_size()` for block-size parsing.
- **inference_workload_matgen**: storage-only mode is now driven by the explicit flag (not inferred from an empty decode list); inconsistent `storage_only`/`num_decode_gpus` combos are rejected; `hit_rate` falls back to `0.0` (not `1.0`) when storage is disabled so compute sleep isn't zeroed.
- **sequential_ctp**: honor an explicit `storage_direct_io=False`; skip RDMA verification for storage-only TPs; export max-latency under `*_max_ms` keys (was mislabeled `*_avg_ms`).
- **custom_traffic_perftest**: deregister the default registration if the backend registration fails; re-raise on `get_agent_metadata()` failure instead of silently continuing.
- **storage_backend**: recreate stale shard files whose on-disk size no longer matches the region being registered.

pre-commit (isort/black/flake8/mypy) passes locally.

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -141,6 +621,437 @@ def _get_bufs(self, tp: TrafficPattern):
...
+            if rank_data:
+                logger.info(
+                    f"TP {tp_idx}:\n{tabulate(rank_data, headers=rank_headers, floatfmt='.3f', missingval='-')}"
+                )
+
+    # =========================================================================
+    # MAIN RUN METHOD
+    # =========================================================================
+
     def run(
```
aranadive, line 1166: Please refactor this function for readability: something like _setup() / _run_isolated_phase() / _run_workload_phase() / _postprocess() / _teardown()

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -141,6 +621,437 @@ def _get_bufs(self, tp: TrafficPattern):
...
+            self._run_tp(write_h, blocking=True)
+            result["write_time"] = time.time() - write_start
+
+        return result
+
+    # =========================================================================
+    # RESULTS REPORTING
+    # =========================================================================
+
+    def _print_iteration_results(
```
aranadive: Can we just pass an results object?

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -160,298 +1071,448 @@ def run(
...
+        except TimeoutError as e:
+            logger.error(
+                "[Rank %d] Health check FAILED: some ranks did not start. %s",
+                self.my_rank,
+                e,
+            )
+            raise RuntimeError(
+                f"Health check failed: not all {self.world_size} ranks started. Check node health and container mounts."
+            ) from e
+
```
aranadive, line 1222: There should be a try here to catch the failures in the calls below

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -160,298 +1071,448 @@ def run(
...
+                        storage_write_handles, "write"
+                    )
+
+            # Barrier: sync all ranks after isolated benchmarks
+            # Only first rank runs isolated storage, others wait here
+            if self.my_rank == 0:
+                logger.info(
+                    "[Rank 0] Isolated benchmarks complete, syncing with other ranks"
+                )
+            dist_rt.barrier(timeout_sec=None)
```
aranadive: Why no timeout?

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -16,52 +16,105 @@
...
+        storage_nixl_backend: Optional[str] = None,
+        storage_direct_io: Optional[bool] = None,
+        storage_block_size: int = 0,
+        storage_posix_api: str = "auto",
+        storage_num_handles: int = 1,
     ) -> None:
         """Initialize multi-pattern performance test.
 
         Args:
             traffic_patterns: List of traffic patterns to test simultaneously
```
aranadive, line 151: super.__init()__ since this is a sub-class?

benchmark/kvbench/README.md
```diff
@@ -30,6 +30,40 @@ NIXL KVBench provides two main categories of functionality:
 1. **KVBench Commands**: Test KV cache transfers across various LLM architectures with different access patterns (block and layer approaches)
 2. **CTP Commands**: Custom Traffic Performance Testing for measuring asymmetric traffic patterns using transfer matrices
 
+### Design Philosophy
+
```
aranadive, line 35: Can we add an end-to-end quicktstart? How does the metadata.yaml get passed, then multi-node launch, etcd setup and how to read isolated vs interleaved output are missing.

aranadive (changes requested):

benchmark/kvbench/README.md
```diff
@@ -30,6 +30,40 @@ NIXL KVBench provides two main categories of functionality:
...
+- Storage I/O (read/write per rank)
+- Network transfers (RDMA)
+- Compute simulation (sleep)
+
+### Execution Flow Per Traffic Pattern
+
+```text
+Each Rank (per traffic pattern):
+  ┌─────────────────────────────────────┐
+  │ 1. COMPUTE (sleep)                  │  Simulate reduced compute time
```
aranadive: What KV reuse pattern is this supposed to mimic? For KV Reuse for PD Disagg, the compute should happen after storage read. There should also be a compute step after RDMA transfer to mimic the Decode Compute. The right traffic pattern should be READ, COMPUTE, RDMA TRANSFER, COMPUTE, STORAGE WRITE.

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
x41lakazam: super.init cannot be called from here, and it's not new in this PR (main also don't call it), mainly because the parent class allocates buffers from a single tp matrix and this class h0old many patterns.
This may be a small duplication, but I don't think it's worth over-complicating the code for the sake of design, also the parent class seems not to find a use case to be used and may be removed later.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ explicit_package_bases = true
 
 [tool.isort]
 profile = "black"
+known_first_party = ["nixl"]
```
aranadive, line 48: Why is this required? NIXL isnt an actual part of kvbench pyproject (though in the same tree)

benchmark/kvbench/test/storage_backend.py
aranadive, line 203: The ABC should define get_read_handles and get_write_handles as well.

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -71,37 +186,449 @@ def __init__(
...
+            Number of notifications cleared
+        """
+        # Get all pending notifications and discard them
+        cleared_count = 0
+
+        # For each TP, check for notifications from all possible senders
+        for tp in self.traffic_patterns:
+            expected_senders = self._get_expected_rdma_senders(tp)
+            for sender_rank in expected_senders:
+                notif_tag = f"{tp.id}_{sender_rank}_{self.my_rank}".encode()
```
aranadive: Notifications from warmup, the isolated runs, and the real workload all carry the same tag, so a receiver can't tell them apart — hence _clear_stale_notifications(). Can we add a per-round suffix to the tag makes them distinguishable - it can be done at transfer time without changing handle setup.

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
aranadive: Ok if the parent class isnt serving anything we can remove it in a later PR

benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```diff
@@ -71,37 +186,449 @@ def __init__(
...
+                # Keep checking until no more notifications with this tag
```
aranadive: Lets update the comment here to say tag_is_prefix = false is required for the updated tag.

pyproject.toml
x41lakazam: It is required imo.

Without `known_first_party = ["nixl"]`, isort moves `from nixl._api import ...` and `from nixl.logging import ...` out of their own group and in with the third-party imports, next to `tabulate`. That rewrites two files and fails the pre-commit CI job:

```
Fixing benchmark/kvbench/test/custom_traffic_perftest.py
Fixing benchmark/kvbench/test/sequential_custom_traffic_perftest.py
```

I also tried moving the setting into `benchmark/kvbench/pyproject.toml`, which is where you would expect it to live. That does not work: the pre-commit isort hook runs from the repo root and does not pick up the nested config, so it then rewrote three files instead of two.

Outside kvbench it is a no-op. I checked the 14 other files in the repo that import `nixl`, and isort produces the same result with or without the setting. So it changes nothing else.

One small correction: `nixl` is declared in kvbench's own dependencies, at `benchmark/kvbench/pyproject.toml:31`.

Happy to drop it if you would rather fix the import order in the two kvbench files by hand instead.

aranadive (approved):

PR #1714, opened 2026-06-02
Use C++20 likely/unlikely instead of __builtin_expect

iyastreb:
## What?
Use C++20 likely/unlikely instead of __builtin_expect

## Why?
Prefer C++20 standard constructs

## Summary by CodeRabbit

* **Refactor**
  * Modernized internal branch annotations across core and backend components. No functional or API changes; error handling and behavior are unchanged. No end‑user impact.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -155,7 +155,7 @@ public:
         nixl_status_t out_ret = NIXL_SUCCESS;
         for (nixlUcxReq req : requests_) {
             const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
-            if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
+            if (ret == NIXL_SUCCESS) [[unlikely]] {
                 worker_->reqRelease(req);
             } else if (ret == NIXL_IN_PROG) {
                 if (out_ret == NIXL_SUCCESS) {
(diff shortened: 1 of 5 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -155,7 +155,7 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
         nixl_status_t out_ret = NIXL_SUCCESS;
         for (nixlUcxReq req : requests_) {
             const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
-            if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
+            if (ret == NIXL_SUCCESS) [[unlikely]] {
```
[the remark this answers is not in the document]
iyastreb: right, fixed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -155,7 +155,7 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
         nixl_status_t out_ret = NIXL_SUCCESS;
         for (nixlUcxReq req : requests_) {
             const nixl_status_t ret = nixl::ucx::ucsToNixlStatus(ucp_request_check_status(req));
-            if (__builtin_expect(ret == NIXL_SUCCESS, 0)) {
+            if (ret == NIXL_SUCCESS) [[likely]] {
```
rakhmets, line 158:
```suggestion
            if (ret == NIXL_SUCCESS) [[unlikely]] {
```

iyastreb: No, it was the right suggestion from coderabbit
If we are here, it means that the LAST request had been completed successfully
So we expect that all the previous ones are also completed successfully (the opposite case is probably not possible at all)

guy-ealey-morag (approved):

rakhmets (approved):

aranadive (approved):

brminich (approved):

PR #1716, opened 2026-06-02
telemetry: disable collection when no exporter is created

e-eygin:
## What?

Fixes a half-enabled telemetry state. Previously, requesting telemetry via `NIXL_TELEMETRY_ENABLE=y` (or the `captureTelemetry` agent config) marked telemetry as "enabled" even when no exporter or sink was actually created (e.g. neither `NIXL_TELEMETRY_DIR` nor `NIXL_TELEMETRY_EXPORTER` configured).

This PR makes telemetry active only when a real telemetry object with a usable exporter/sink exists:

* Adds `nixlTelemetry::create()`, which returns `nullptr` when no exporter/sink is configured.
* Makes `data->telemetry_ != nullptr` the single source of truth for whether telemetry is active.
* Removes the redundant `telemetryEnabled` flag.
* Adds a **NOOP exporter** (`NIXL_TELEMETRY_EXPORTER=NOOP`) so the "collect but don't export" mode — useful for measuring the overhead of the event-collection path in isolation — remains available *explicitly*, instead of as an accidental side effect of an unconfigured sink.
* Updates C++ gtest, Python, and Rust binding tests, and `docs/telemetry.md`.

## Why?

In the old behavior, with telemetry "enabled" but no exporter created, `initializeTelemetry()` returned without an exporter while the agent still kept `telemetryEnabled == true`. The datapath then continued to record events that could never be exported — wasted work and a misleading "enabled" state.

Now there is one consistent rule: telemetry is active only if `telemetry_` exists. If telemetry is requested but no sink/exporter is configured, collection is disabled and `getXferTelemetry()` returns `NIXL_ERR_NO_TELEMETRY`.

This removes a deliberate use case — "enabled, no exporter" to gauge collection overhead — but that path was implicit, misleading, and a poor benchmark (the drain task never ran, so the in-memory buffer filled to `NIXL_TELEMETRY_BUFFER_SIZE` and then silently dropped). The NOOP exporter restores the capability cleanly: telemetry stays active (events collected and drained, `getXferTelemetry()` works) while all output is discarded, so collection overhead can be measured without any export/serialization/IO and without the silent half-enabled state.

This is a behavior change for callers that relied on in-band `getXferTelemetry()` with `NIXL_TELEMETRY_ENABLE=y` but no configured sink; they now need to configure a sink (`NIXL_TELEMETRY_DIR` / an exporter), or use `NIXL_TELEMETRY_EXPORTER=NOOP`.

## How?

* `src/core/telemetry/telemetry.{h,cpp}`: add a `create()` factory that resolves the configured exporter up front and returns `nullptr` when no sink is configured (no telemetry instance is built in that case); otherwise it constructs `nixlTelemetry` with the resolved exporter name. The exporter is a `const` member initialized from the member-initializer list, so a constructed `nixlTelemetry` always has a usable exporter; genuine config / plugin-load failures throw.
* `src/core/agent_data.h` / `src/core/nixl_agent.cpp`: drop `telemetryEnabled`; construct telemetry via `nixlTelemetry::create()`; gate all datapath telemetry on `data->telemetry_`.
* NOOP exporter: `src/core/telemetry/noop_exporter.h` + `noop_plugin.cpp`, registered as a static plugin (`NIXL_REGISTER_STATIC_PLUGIN(Telemetry, NOOP)`) like `BUFFER`; `exportEvent()` discards.
* Tests: C++/Python no-sink cases expect `NIXL_ERR_NO_TELEMETRY`; Rust tests use a scoped `EnvGuard`; new `telemetryTest.NoopExporterIsActiveAndWritesNothing` and `telemetryTest.CreateReturnsNullWithoutSink`.
* `docs/telemetry.md`: documents activation rules and `NIXL_TELEMETRY_EXPORTER=NOOP`.

## Test

* `build/test/gtest/gtest --gtest_filter='telemetryTest.*'` — 16/16 (incl. NOOP + no-sink `create()` tests); broader `*Telemetry*` suite green (incl. UCX `TestTransferTelemetry`).
* `pytest test/python/test_nixl_api.py -k xfer_telemetry` — passed
* `cargo test -p nixl-sys --test tests get_xfer_telemetry` — passed
* clang-format-19 (diff-scoped) clean

## Summary by CodeRabbit

* **Documentation**
  * Clarified when telemetry is active: depends on whether an exporter/sink can be created and on the telemetry directory env setting.

* **New Features**
  * Telemetry now initializes via a creation check so it is only active when a usable exporter/sink exists.
  * Added scoped test utilities to safely manage telemetry-related environment variables during tests.

* **Tests**
  * Improved telemetry test isolation with per-test temp directories and updated expectations when no sink/exporter is present.


src/bindings/rust/tests/tests.rs
```diff
@@ -1607,6 +1624,10 @@ fn test_get_xfer_telemetry_from_request() {
 
         println!("Telemetry data from request: {:?}", telemetry);
     }
+
+    env::remove_var("NIXL_TELEMETRY_ENABLE");
+    env::remove_var("NIXL_TELEMETRY_DIR");
+    let _ = std::fs::remove_dir_all(telemetry_dir);
```
roiedanino: Minor: move to some telemetry_cleanup() function to avoid the duplication with 1572-1574

src/core/telemetry/telemetry.h
```diff
@@ -50,6 +50,9 @@ struct periodicTask {
 
 class nixlTelemetry {
 public:
+    static std::unique_ptr<nixlTelemetry>
+    create(const std::string &agent_name);
```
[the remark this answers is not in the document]
e-eygin: fixed.

src/bindings/rust/tests/tests.rs
```diff
@@ -1556,11 +1568,16 @@ fn test_get_xfer_telemetry_success() {
         println!("Telemetry data: {:?}", telemetry);
         println!("Transfer rate: {:.2} MB/s", rate / 1_000_000.0);
     }
+
+    env::remove_var("NIXL_TELEMETRY_ENABLE");
+    env::remove_var("NIXL_TELEMETRY_DIR");
+    let _ = std::fs::remove_dir_all(telemetry_dir);
```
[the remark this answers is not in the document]
e-eygin: fixed.

test/python/test_nixl_api.py
```diff
@@ -246,7 +247,7 @@ def test_incorrect_plugin_env(monkeypatch):
         nixl_agent("bad env agent")
 
 
-def _run_xfer_telemetry_check(agent1, agent2):
+def _run_xfer_telemetry_check(agent1, agent2, expect_telemetry=True):
```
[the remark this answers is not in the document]
e-eygin: fixed.

src/bindings/rust/tests/tests.rs
e-eygin: fixed.

src/bindings/rust/tests/tests.rs
```diff
@@ -40,6 +40,25 @@ fn setup_agent_with_backend(agent: &Agent) -> Result<OptArgs, NixlError> {
...
+    path
+}
+
+fn telemetry_cleanup(telemetry_dir: &str) {
+    env::remove_var("NIXL_TELEMETRY_ENABLE");
+    env::remove_var("NIXL_TELEMETRY_DIR");
+    if let Err(e) = std::fs::remove_dir_all(telemetry_dir) {
+        eprintln!("Warning: failed to remove telemetry directory {telemetry_dir}: {e}");
+    }
+}
```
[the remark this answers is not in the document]
e-eygin: fixed.

src/bindings/rust/tests/tests.rs
```diff
@@ -1664,7 +1690,8 @@ fn test_get_xfer_telemetry_without_telemetry_enabled() {
 
 #[test]
 fn test_get_xfer_telemetry_before_posting() {
-    env::set_var("NIXL_TELEMETRY_ENABLE", "1");
+    let env_guard = EnvGuard::new(["NIXL_TELEMETRY_ENABLE", "NIXL_TELEMETRY_DIR"]);
+    env_guard.set("NIXL_TELEMETRY_ENABLE", "1");
```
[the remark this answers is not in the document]
e-eygin: fixed.

src/bindings/rust/tests/env_guard/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+#![allow(dead_code)]
+
+use std::env;
+use std::sync::{Mutex, MutexGuard, OnceLock};
+
+/// Process-global lock serializing all `EnvGuard` users, so tests that mutate
+/// shared environment variables cannot race under the parallel test runner.
+fn env_lock() -> &'static Mutex<()> {
+    static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
+    LOCK.get_or_init(|| Mutex::new(()))
```
tomerg-nvidia: Is the OnceLock needed here?

e-eygin:

src/bindings/rust/tests/env_guard/mod.rs
e-eygin: You're right — the `OnceLock` wrapper is redundant. `Mutex::new` is `const`, so I
replaced it with a const-initialized static. The mutex itself stays.

dpressle:
@ci-triage-agent http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/[nixl-ci-gpu](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/1960/pipeline/)/detail/nixl-ci-gpu/1960/pipeline/

dpressle:
👀 Investigating [`nixl-ci-gpu` build 1960](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-gpu/detail/nixl-ci-gpu/1960/pipeline/)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-gpu` · commit `67b508e7`

I now have a complete, confident picture of the root cause. Let me synthesize everything:

---

**Summary:** `test_agent` in `test/python/test_nixl_bindings.py` fails with `nixlNoTelemetryError: NIXL_ERR_NO_TELEMETRY` when calling `agent1.getXferTelemetry(handle)`, even though the test explicitly sets `NIXL_TELEMETRY_ENABLE=y` before constructing the agents.

**Root cause:** The telemetry initialization logic introduced in commit `[REDACTED:Hex High Entropy String]` (PR #1716, "telemetry: disable collection when no exporter is created") changed the condition under which `telemetryEnabled` is set. The new code at `src/core/nixl_agent.cpp:140–155` reads:

```cpp
if (telemetry_enabled) {
    if (*telemetry_enabled) {
        telemetryEnabled = true;
        telemetry_ = std::make_unique<nixlTelemetry>(name);  // exporter created
    } else if (config.captureTelemetry) {
        telemetryEnabled = true;   // env var says "no", but config overrides — OK
        NIXL_WARN << "NIXL telemetry is enabled through config, "
                     "ignoring the NIXL_TELEMETRY_ENABLE environment variable";
    } else {
        NIXL_DEBUG << "NIXL telemetry is disabled";
    }
} else if (config.captureTelemetry) {
    telemetryEnabled = true;   // env var absent, config says yes — OK
}
```

The critical change is that the env-var path `*telemetry_enabled == true` now **also requires an exporter** (`telemetry_ = std::make_unique<nixlTelemetry>(name)`). If the build/environment does not have a telemetry exporter plugin available (or if `nixlTelemetry` construction silently fails/does not set `telemetryEnabled`), the flag stays `false`. The CI log confirms this:

```
W0603 12:12:25.818215 nixl_agent.cpp:147] NIXL telemetry is disabled; ignoring telemetry requested through agent config
E0603 12:12:28.148528 nixl_agent.cpp:1162] getXferTelemetry: cannot return values when telemetry is not enabled.
```

The test sets `NIXL_TELEMETRY_ENABLE=y` **in the Python environment** (via `os.environ`) and passes `devices.useProgThread = False` with a default `nixlAgentConfig` — but the test does **not** set `config.captureTelemetry = True`. PR #1716 decoupled enabling the env-var telemetry path from enabling **in-memory capture** (`captureTelemetry`), so the env var alone no longer guarantees `getXferTelemetry()` will work if there is no exporter.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` — PR #1716 ("telemetry: disable collection when no exporter is created")

**File:** `test/python/test_nixl_bindings.py:91` (failing assert) and `src/core/nixl_agent.cpp:140–155` (changed initialization logic)

**Suggested fix:** Two options — choose one based on intent:

1. **Fix the test (preferred for PR #1716):** Update `test_agent()` to also enable in-memory capture via the agent config, so `getXferTelemetry` is valid even without an exporter:
   ```python
   devices = nixl.nixlAgentConfig()
   devices.useProgThread = False
   devices.captureTelemetry = True   # ← add this line
   ```
   And remove or keep the `os.environ["NIXL_TELEMETRY_ENABLE"] = "y"` line (it controls the exporter, not in-memory capture).

2. **Fix the library logic (alternative):** If the intent is that `NIXL_TELEMETRY_ENABLE=y` should always imply `telemetryEnabled = true` (for both exporter and in-memory stats), then restore the `telemetryEnabled = true` assignment regardless of whether the exporter was created successfully, and only skip exporter construction if unavailable — preserving the ability to call `getXferTelemetry()`.

**Related:** [PR #1716 — telemetry: disable collection when no exporter is created](https://github.com/ai-dynamo/nixl/pull/1716)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `6311f659-b15f-4920-97cb-a8e45822d639` in the triage console for the audit trail.

ColinNV:
I remember it being mentioned that "telemetry enabled but no exporter" was useful to measure the overhead of just collecting the events, are we sure this is no longer needed?

e-eygin:
@ColinNV Good catch. You're right that disabling telemetry with no sink also removed the "collect, don't export" overhead-measurement mode. I added an explicit NOOP exporter: set NIXL_TELEMETRY_EXPORTER=NOOP and telemetry stays active (events collected, getXferTelemetry works) but everything is discarded — so you can measure collection overhead in isolation, and it actually drains (the old path silently capped at the buffer size). In 809ce8d8.

src/core/telemetry/telemetry.cpp
```diff
@@ -73,6 +73,15 @@ constexpr std::chrono::milliseconds DEFAULT_TELEMETRY_RUN_INTERVAL = 100ms;
 constexpr size_t DEFAULT_TELEMETRY_BUFFER_SIZE = 4096;
 constexpr const char *defaultTelemetryPlugin = "BUFFER";
 
+std::unique_ptr<nixlTelemetry>
+nixlTelemetry::create(const std::string &agent_name) {
+    auto telemetry = std::make_unique<nixlTelemetry>(agent_name);
+    if (!telemetry->isActive()) {
```
ColinNV: It looks like the `if`-condition is always `false` because `isActive()` is always `true` -- `initializeTelemetry()` throws an exception when `exporter_ == nullptr`.

Perhaps this could be made clearer by `initialieTelemetry()` returning the exporter, declaring `exporter_` as `const`, and using the member initializer list to set it up?

ColinNV (changes requested):

src/core/telemetry/telemetry.cpp
e-eygin: @ColinNV Done in 51bfa6e6 — exporter_ is now a const member built via the member-init list (makeExporter() returns it, or nullptr when no sink is configured), with the periodic export task started from the ctor body. 

Note the if isn't dead: the no-sink path (no NIXL_TELEMETRY_DIR / NIXL_TELEMETRY_EXPORTER) returns nullptr without throwing, which is how "no sink → telemetry disabled" works — the refactor just makes that explicit.

ColinNV: I see. Perhaps the exporter check could be done here instead of in `makeExporter()`, and if the optional is set its content can be passed as `const std::string&` to the `nixlTelemetry` constructor. Then the invariant "telemetry always has an exporter" holds, and we don't create the telemetry just to sometimes destroy it again immediately.

e-eygin: Done in 391b3c58 — create() now resolves the exporter and only builds nixlTelemetry when a sink is configured, passing the name to the ctor. So the "always has an exporter" invariant holds, no construct-then-destroy, and isActive() is gone.

src/core/telemetry/telemetry.cpp
```diff
@@ -122,44 +148,51 @@ getExporterName() {
...
+}
 
-    const std::optional<std::string> exporter_name = getExporterName();
-
-    if (!exporter_name) {
-        return;
-    }
+} // namespace
 
+std::unique_ptr<nixlTelemetryExporter>
```
ColinNV: `[[nodiscard]]`

src/core/telemetry/telemetry.h
```diff
@@ -50,7 +50,18 @@ struct periodicTask {
...
-    explicit nixlTelemetry(const std::string &agent_name);
+    /**
+     * @brief Creates a telemetry instance if an exporter/sink is configured.
+     * @param agent_name Non-empty agent name.
+     * @return A telemetry instance, or nullptr when no exporter/sink is
+     *         configured (telemetry is intentionally disabled).
+     * @throws std::invalid_argument / std::runtime_error on genuine
+     *         configuration or plugin-load errors.
+     */
+    static std::unique_ptr<nixlTelemetry>
```
ColinNV: `[[nodiscard]]`

ColinNV (dismissed):
I'd choose "NOP" over "NOOP", but that's just a personal preference.

ColinNV (dismissed):

ovidiusm (approved):

src/core/telemetry/telemetry.cpp
e-eygin: fixed.

src/core/telemetry/telemetry.cpp
```diff
@@ -73,14 +73,42 @@ constexpr std::chrono::milliseconds DEFAULT_TELEMETRY_RUN_INTERVAL = 100ms;
...
-nixlTelemetry::nixlTelemetry(const std::string &agent_name)
-    : pool_(1),
-      writeTask_(pool_.get_executor(), DEFAULT_TELEMETRY_RUN_INTERVAL, false),
-      agentName_(agent_name) {
-    if (agent_name.empty()) {
-        throw std::invalid_argument("Expected non-empty agent name in nixl telemetry create");
+namespace {
+// Defined below; declared here for use by create() and the constructor's
+// member-initializer list.
+[[nodiscard]] std::optional<std::string>
```
iyastreb, line 79: minor: imo it's better to implement right here, but up to you

src/core/telemetry/telemetry.cpp
```diff
@@ -122,46 +150,51 @@ getExporterName() {
...
-void
-nixlTelemetry::initializeTelemetry() {
-    maxBufferedEvents_ = nixl::config::getValueDefaulted<size_t>(TELEMETRY_BUFFER_SIZE_VAR,
-                                                                 DEFAULT_TELEMETRY_BUFFER_SIZE);
+[[nodiscard]] std::string
+validateAgentName(const std::string &agent_name) {
+    if (agent_name.empty()) {
+        throw std::invalid_argument("Expected non-empty agent name in nixl telemetry create");
+    }
+    return agent_name;
```
iyastreb, line 158: no need to construct a new string, this function may be void
just call it inside ctor body

src/core/telemetry/telemetry.h
```diff
@@ -50,7 +50,18 @@ struct periodicTask {
...
+     * @param agent_name Non-empty agent name.
+     * @return A telemetry instance, or nullptr when no exporter/sink is
+     *         configured (telemetry is intentionally disabled).
+     * @throws std::invalid_argument / std::runtime_error on genuine
+     *         configuration or plugin-load errors.
+     */
+    [[nodiscard]] static std::unique_ptr<nixlTelemetry>
+    create(const std::string &agent_name);
+
+    nixlTelemetry(const std::string &agent_name, const std::string &exporter_name);
```
iyastreb, line 64: if factory method is there, typically it makes sense to make ctor private to force using only a factory

src/core/telemetry/telemetry.cpp
e-eygin: Thanks! These are forward-declared on purpose so create() and the constructor
read top-down at the top of the file, with the small helpers grouped just below.
Happy to inline if you'd prefer, but I'll keep it as-is for that readability reason.

src/core/telemetry/telemetry.cpp
e-eygin: Good point. I went with validate-and-return so that agentName_ is set directly in the member initializer list (keeping validation as a pure function), rather than being assigned in the constructor body—the copy cost is the same either way. I will keep the initializer-list form unless you feel strongly.

src/core/telemetry/telemetry.h
e-eygin: Agreed in principle. The catch is that create() builds the instance via
std::make_unique<nixlTelemetry>(...), which can't reach a private ctor without a
passkey idiom or friend, and the C++/Rust tests construct it directly too. To
avoid that extra machinery I left the ctor public for now, with create() as the
documented entry point -- happy to add a passkey later if we want to enforce
factory-only construction.

tomerg-nvidia (approved):

PR #1718, opened 2026-06-02
API: Zero-copy makeXferReq

iyastreb:
## What?
makeXferReq python binding makes a copy of 2 passed integer vectors: it allocates and copies data into `std::vector<int>` which is accessed in `nixlAgent::makeXferReq`. This PR implements zero-copy by:
- adding makeXferReq overload in nixlAgent
- using it in Python binding

This change is ABI safe

## Why?
Making copy is completely redundant:
- passed Python array is guaranteed to outlive the makeXferReq
- Python array is used just inside makeXferReq and not referenced
- Array is immutable

In certain cases the indices array can be very large, up to several thousands elements. Removing extra copy (while holding a GIL lock) helps to improve performance.

## Summary by CodeRabbit

* **Refactor**
  * Reduced copying and improved performance/memory efficiency when specifying index lists for transfers, yielding faster transfer-request creation.
  * Python bindings streamlined to accept contiguous arrays (e.g., NumPy) and native lists with lower overhead.

* **Bug Fixes**
  * Earlier and stricter input validation in bindings and transfer request creation to prevent invalid/null inputs and improve reliability.


src/api/cpp/nixl.h
```diff
@@ -220,6 +221,16 @@ class nixlAgent {
...
                      const nixl_opt_args_t* extra_params = nullptr) const;
+
+        nixl_status_t
+        makeXferReq(const nixl_xfer_op_t &operation,
+                    const nixlDlistH *local_side,
+                    std::span<const int> local_indices,
+                    const nixlDlistH *remote_side,
+                    std::span<const int> remote_indices,
+                    nixlXferReqH *&req_hndl,
+                    const nixl_opt_args_t *extra_params = nullptr) const;
```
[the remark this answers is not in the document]
iyastreb: done

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/26825926659/job/79093247684)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `3d264cd5`

I have all the evidence needed. Here is the complete diagnosis:

---

**Summary:** `clang-format-19` rejects a too-long `@see` comment line in `src/api/cpp/nixl.h` introduced by PR #1718 (Zero-copy makeXferReq).

**Root cause:** The `Clang Format Check` workflow runs `clang-format-diff-19` against the diff of the PR commit. It found one formatting violation in `src/api/cpp/nixl.h` around line 224–227: a `@see` Doxygen comment on a **single line** exceeds the column limit enforced by the project's `.clang-format` style file. The required fix is to **wrap the long `@see` line** onto two lines at a logical break point (after the first `const std::vector<int>&` argument), exactly as `clang-format-19` prescribes:

```diff
-         * `@see` makeXferReq(const nixl_xfer_op_t&, const nixlDlistH*, const std::vector<int>&, const nixlDlistH*, const std::vector<int>&, nixlXferReqH*&, const nixl_opt_args_t*) const
+         * `@see` makeXferReq(const nixl_xfer_op_t&, const nixlDlistH*, const std::vector<int>&,
+         * const nixlDlistH*, const std::vector<int>&, nixlXferReqH*&, const nixl_opt_args_t*) const
```

The three changed files reported by the check are:
- `src/api/cpp/nixl.h` — **the only file with a format violation** (the long `@see` comment)
- `src/bindings/python/nixl_bindings.cpp`
- `src/core/nixl_agent.cpp`

Only `nixl.h` produced a diff from `clang-format-diff-19`.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1718 head, author unknown from logs)

**File:** `src/api/cpp/nixl.h` — the `@see` tag in the Doxygen block for the span-based `makeXferReq` overload (around line 224–227 in the PR diff context)

**Suggested fix:** In your local branch, apply the wrap that `clang-format` expects — either run `clang-format-19` locally, or manually break the `@see` line after `const std::vector<int>&,` as shown in the diff above, then push. The quickest method:
```bash
# Install clang-format-19 and auto-fix just the changed lines
git diff HEAD~1 HEAD -- src/api/cpp/nixl.h | clang-format-diff-19 -p1 -style=file -i
git add src/api/cpp/nixl.h
git commit --amend --no-edit
```

**Related:** [PR #1718 — Zero-copy makeXferReq](https://github.com/ai-dynamo/nixl/pull/1718)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `0e51f49b-4c54-4ccd-9d3b-b269b021fbd1` in the triage console for the audit trail.

tomerg-nvidia (dismissed):

src/api/cpp/nixl.h
```diff
@@ -220,6 +221,21 @@ class nixlAgent {
...
+        /**
+         * `@brief` Span-based overload of makeXferReq.
+         * `@see` makeXferReq(const nixl_xfer_op_t&, const nixlDlistH*, const std::vector<int>&,
+         * const nixlDlistH*, const std::vector<int>&, nixlXferReqH*&, const nixl_opt_args_t*) const
+         */
+        nixl_status_t
+        makeXferReq(const nixl_xfer_op_t &operation,
+                    const nixlDlistH *local_side,
+                    std::span<const int> local_indices,
+                    const nixlDlistH *remote_side,
```
rakhmets: I think we should also replace pointers by references for `local_side` and `remote_side` parameters in the new API. `makeXferReq` doesn't accept `local_side` and `remote_side` as `nullptr`s.

And then you can remove:
```C++
    if (!local_side || !remote_side) {
        NIXL_ERROR_FUNC << "local or remote side handle is null";
        data->addErrorTelemetry(NIXL_ERR_INVALID_PARAM);
    }
```
from the implementation.

src/api/cpp/nixl.h
```diff
@@ -220,6 +221,21 @@ class nixlAgent {
...
                      nixlXferReqH* &req_hndl,
                      const nixl_opt_args_t* extra_params = nullptr) const;
+
+        /**
+         * `@brief` Span-based overload of makeXferReq.
+         * `@see` makeXferReq(const nixl_xfer_op_t&, const nixlDlistH*, const std::vector<int>&,
+         * const nixlDlistH*, const std::vector<int>&, nixlXferReqH*&, const nixl_opt_args_t*) const
+         */
+        nixl_status_t
+        makeXferReq(const nixl_xfer_op_t &operation,
```
rakhmets:
```suggestion
        makeXferReq(nixl_xfer_op_t operation,
```

iyastreb: ok

src/api/cpp/nixl.h
iyastreb: right

src/bindings/python/nixl_bindings.cpp
```diff
@@ -617,26 +624,24 @@ PYBIND11_MODULE(_bindings, m) {
...
+                    return std::span<const int>(backing);
                 };
 
-                local_indices_vec = init_indices_lambda(local_indices);
-                remote_indices_vec = init_indices_lambda(remote_indices);
+                const auto local_span = indices_to_span(local_indices, local_indices_vec);
+                const auto remote_span = indices_to_span(remote_indices, remote_indices_vec);
 
+                nixlDlistH *local_dlist = reinterpret_cast<nixlDlistH *>(local_side);
+                nixlDlistH *remote_dlist = reinterpret_cast<nixlDlistH *>(remote_side);
```
rakhmets: Minor comment. `auto` can be used here, since the type is mentioned as a cast parameter.

iyastreb: done

src/api/cpp/nixl.h
```diff
@@ -220,6 +221,21 @@ class nixlAgent {
                      const std::vector<int> &remote_indices,
                      nixlXferReqH* &req_hndl,
                      const nixl_opt_args_t* extra_params = nullptr) const;
+
+        /**
+         * `@brief` Span-based overload of makeXferReq.
+         * `@see` makeXferReq(const nixl_xfer_op_t&, const nixlDlistH*, const std::vector<int>&,
+         * const nixlDlistH*, const std::vector<int>&, nixlXferReqH*&, const nixl_opt_args_t*) const
```
rakhmets: I don't think that `@see` will work as a link in this way.
Also "``" around commands should be removed.
```suggestion
         * @overload
         * @brief Span-based overload of makeXferReq.
```

dpressle:
👀 Investigating [`AWS NIXL Validation`](https://github.com/ai-dynamo/nixl/actions/runs/26869888096/job/79242323565)…

dpressle:
🤖 **CI Triage Agent** — `AWS NIXL Validation` · commit `ac7c9dd9`

I have all the evidence needed. Here is the complete diagnosis:

---

**Summary:** `prep_xfer_perf.py array` mode crashes with `ValueError: indices numpy array must be 1D of uint32 or int32` because `np.arange(num_descs)` produces a `uint64` (or platform default `int64`) array but `makeXferReq` now enforces a strict `uint32`/`int32` dtype.

**Root cause:** PR #1718 ("Zero-copy makeXferReq") tightened the C++/Python binding for `makeXferReq` to require that any numpy `indices` array be **1D and of dtype `uint32` or `int32`**. The `perf_test_array` function in `test/python/prep_xfer_perf.py` builds the `indices` array with:
```python
indices = np.arange(num_descs)          # line 106 — dtype defaults to int64 / uint64
```
and then passes it straight through to `make_prepped_xfer` → `makeXferReq` (line 60-62 via `prep_handles`). Because the array is 64-bit, the new dtype check rejects it:
```
ValueError: indices numpy array must be 1D of uint32 or int32
  File "prep_xfer_perf.py", line 60, in prep_handles
    xfer_handle = agent.make_prepped_xfer(
```
The `list` mode path (line 76) builds `indices` as a plain Python `list(range(num_descs))`, which is unaffected. Only the `array` mode path, and only on this new PR branch, hits the error — confirming the regression was introduced by the new binding constraint in PR #1718.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1718, branch `pull-request/1718`)

**File:** `test/python/prep_xfer_perf.py:106`

**Suggested fix:** Explicitly cast the numpy `indices` array to `uint32` (or `int32`) in `perf_test_array` to match the new binding constraint:

```python
# Before (line 106):
indices = np.arange(num_descs)

# After:
indices = np.arange(num_descs, dtype=np.uint32)
```

Alternatively, if the binding is intentionally strict, update the docstring/type-check error message to say `uint32/int32` and add a helper that auto-casts integer arrays in `_api.py`'s `make_prepped_xfer` wrapper so callers don't have to think about it. Either way, the test script and any other callers using `np.arange()` without a dtype will need the same fix.

**Related:** [PR #1718 — Zero-copy makeXferReq](https://github.com/ai-dynamo/nixl/pull/1718)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `3e377933-c84c-47cc-8393-ba6f10825324` in the triage console for the audit trail.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -617,28 +624,27 @@ PYBIND11_MODULE(_bindings, m) {
                                 "indices numpy array must be 1D of uint32 or int32");
```
rakhmets: This is not aligned with https://github.com/ai-dynamo/nixl/blob/871a6294379e7bd32485302f3c86e3043a97b6b7/src/api/python/_api.py#L540

I understand that this mismatch was here before your changes. But maybe we may fix it in this PR.
If we want to accept only `dtype=int32`, then it will simplify the code a little. And also I think there is a potential issue at line 628. Because you cast `uin32_t` to `int`.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -617,28 +624,27 @@ PYBIND11_MODULE(_bindings, m) {
...
+                    backing = indices.cast<std::vector<int>>();
+                    return std::span<const int>(backing);
                 };
 
-                local_indices_vec = init_indices_lambda(local_indices);
-                remote_indices_vec = init_indices_lambda(remote_indices);
+                const auto local_span = indices_to_span(local_indices, local_indices_vec);
+                const auto remote_span = indices_to_span(remote_indices, remote_indices_vec);
+
+                auto local_dlist = reinterpret_cast<nixlDlistH *>(local_side);
```
ColinNV: `const`

iyastreb: fixed

src/bindings/python/nixl_bindings.cpp
iyastreb: This turns out to be a real issue actually
I checked vLLM code and they actually pass int64 array, which is implicitly converted into int32 one by:
```
auto indices_array = indices.cast<py::array_t<uint32_t>>();
```
I will create PR for vLLM to change the type to int32, but before that we cannot remove this cast from NIXL.
I put it into a separate block and added a comment to remove it

tomerg-nvidia (dismissed):

src/api/cpp/nixl.h
mkhazraee, line 219: I suggest to put a "to be deprecated" here, even if we don't do it any time soon. Signaling we're moving to span based APIs.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -693,39 +699,44 @@ PYBIND11_MODULE(_bindings, m) {
...
+                                          std::vector<int> &backing) -> std::span<const int> {
                     if (py::isinstance<py::array>(indices)) {
-                        auto indices_array = indices.cast<py::array_t<uint32_t>>();
+                        const auto indices_array = indices.cast<py::array>();
                         if (indices_array.ndim() != 1)
                             throw std::invalid_argument("indices numpy array must be 1D");
-                        if (!py::dtype::of<uint32_t>().equal(indices_array.dtype()) &&
-                            !py::dtype::of<int32_t>().equal(indices_array.dtype()))
-                            throw std::invalid_argument(
-                                "indices numpy array must be 1D of uint32 or int32");
```
mkhazraee, line 704: We don't need to throw exceptions anymore? We just now check for valid pointers?

src/api/cpp/nixl.h
iyastreb: Yes, I considered this option too
I didn't add it because it may produce warnings at client side (which can be treated as errors, depending on their build settings)
If you think it's ok, I'll add it

iyastreb: For now just adding @deprecated doc
Will add `[[deprecated]]` only if you confirm it's ok

src/bindings/python/nixl_bindings.cpp
iyastreb: These checks were completely pointless, because first we casted array to uint32_t:
`auto indices_array = indices.cast<py::array_t<uint32_t>>();`
and this cast means that we convert original array into new type if they don't match
So after the cast we already have a valid converted array, no need to check it's type.

In the new impl we convert(=cast) only in case of type mismatch.
And if `cast` fails then it throws an exception

brminich (approved):

src/api/cpp/nixl.h
mkhazraee: I meant just in the comment, that's what we did for skipDescMerge. Not full [[deprecated]].

mkhazraee: Just saw it.

mkhazraee (approved):

PR #1723, opened 2026-06-03
CI: wait full build+test for AWS batch completion

Alexey-Rivkin:
## What?

Two fixes to the AWS EFA job-completion wait in `contrib/aws-efa/aws_test.sh`:

1. **Raise the completion timeout from 10m to 60m** - the actual fix for the false failures.
2. **Fail fast on terminal `FAILED`** - a complementary guard so the longer wait can't be abused by a genuinely failed job.

## Why?

Jobs intermittently fail across PRs with `Timeout waiting for status SUCCEEDED after 600s. Final status: RUNNING`.

`kubectl logs -f` returns early on an apiserver HTTP/2 `GOAWAY` (graceful, `ErrCode=NO_ERROR`), after which only 10m was allowed for `SUCCEEDED`. The C++20 switch (#1571) pushed the in-pod build+test past that tail, so healthy jobs timed out while still `RUNNING` and reported a false failure. A green run takes ~35m.

## How?

1. **60m timeout** covers the full build+test even when the log stream drops at the start of `RUNNING`. Real hangs stay capped by the in-pod `timeout` and the job-definition timeout.
2. **`FAILED` early-exit** - with a 60m wait, a genuinely failed job would otherwise idle the full hour. The check applies to both the `RUNNING` and `SUCCEEDED` waits, so a job that dies at startup also exits immediately instead of burning its start timeout.

## Summary by CodeRabbit

* **Bug Fixes**
  * The test script now detects and reports job failures immediately (with accurate elapsed-time reporting) instead of waiting for the original timeout.
  * On failure, verbose shell tracing is re-enabled to improve diagnostic output.
  * Extended the overall completion wait period to cover the full build+test window; notes added about log streaming possibly ending before completion.

### DIFF
--- a/contrib/aws-efa/aws_test.sh
+++ b/contrib/aws-efa/aws_test.sh
@@ -98,6 +98,12 @@ wait_for_status() {
             set -x
             return 0
         fi
+        # Fail fast on terminal FAILED, don't wait out the timeout.
+        if [ "$status" = "FAILED" ]; then
+            echo -e "\nJob reached terminal status FAILED after ${SECONDS}s"
+            set -x
+            return 1
+        fi
         printf "."
         sleep $interval
     done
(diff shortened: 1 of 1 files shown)


contrib/aws-efa/aws_test.sh
dpressle, line 53: if internal timeout for success is 60 min now we should increase the upper layer timeout - it should be nested so something like 65 minutes in the workflow yaml should be good

dpressle (changes requested):

contrib/aws-efa/aws_test.sh
Alexey-Rivkin: Good point!
51fd01d2a26f6154eda7469ef0d52977fd472698

dpressle (approved):

PR #1647, opened 2026-05-15
Add AMD ROCm/HIP build support for nixlbench (follow-up to #1642)

andyluo7:
## Summary

Adds AMD ROCm/HIP build support for **nixlbench** (the canonical NIXL benchmark tool), gated behind a default-off `use_rocm=true` Meson option. Stacks on top of #1642 (which provides the same option for the NIXL library proper).

When `use_rocm=true`, nixlbench builds against ROCm/HIP using the existing CUDA-shaped source code unchanged — translation happens transparently at preprocess time via header-only symbol aliases.

## Approach: header-only symbol translation

`benchmark/nixlbench/src/utils/utils.h` adds an `#ifdef __HIP_PLATFORM_AMD__` block that `#define`s every CUDA runtime + driver API symbol used by nixlbench to its HIP equivalent:

```cpp
// Status types
#define cudaError_t                          hipError_t
#define cudaSuccess                          hipSuccess
#define CUresult                             hipError_t
#define CUDA_SUCCESS                         hipSuccess

// Runtime API
#define cudaMalloc                           hipMalloc
#define cudaMemcpy                           hipMemcpy
#define cudaSetDevice                        hipSetDevice
// ... etc

// Driver API: virtual memory
#define CUdeviceptr                          hipDeviceptr_t
#define CUmemAllocationProp                  hipMemAllocationProp
#define cuMemCreate                          hipMemCreate
#define cuMemMap                             hipMemMap
// ... etc
```

`.cpp` source files stay unchanged with `cu*`/`cuda*` names. `utils.cpp` and `nixl_worker.cpp` had direct `#include <cuda.h>` / `#include <cuda_runtime.h>` lines that bypassed `utils.h`'s translations — those are removed; `utils.h` now provides the right header for both paths.

Two narrow code patches needed beyond pure aliasing:

1. **`uintptr_t → void*` cast** at HIP virtual-memory call sites. CUDA's `cuMemUnmap(CUdeviceptr=ulong, ...)` accepts integer `iov.addr` implicitly; HIP's `hipMemUnmap(void*, ...)` rejects it. New `NIXLB_DEV_PTR()` macro (identity on CUDA, `reinterpret_cast<void*>` on HIP) wraps the call sites.
2. **`[[nodiscard]]` on `hipDrvGetErrorString`** — original macro discards return; `(void)` cast added.

## What's enabled / disabled on ROCm builds

| Component | ROCm path |
|---|---|
| UCX backend (NIXL library) | ✅ via PR #1642 |
[...]

PR #1735, opened 2026-06-05
contrib: unified ROCm/CUDA Dockerfile with ROCm build-check CI and wheel workflow

andyluo7:
## Summary

Restructures `contrib/Dockerfile` into a unified three-stage build (`common-base → rocm-build / cuda-build`) and adds ROCm-specific CI and release infrastructure. Splits the packaging surface out of #1647 per @tvegas1's request.

## What this adds

| File | Purpose |
|---|---|
| `contrib/Dockerfile` | Three-stage build: `common-base` (shared apt deps, Abseil/gRPC, Python tooling) → `rocm-build` (UCX `--with-rocm`, PyTorch ROCm wheels, `nixl_rocm` wheel) / `cuda-build` (DOCA, libfabric, UCX `--with-cuda`, NIXL EP, `nixl` CUDA wheel). UCX pinned to `940c1c1d9` (last clean commit before a `put_offload.c` `-Werror` regression on v1.21.x). |
| `contrib/build-container.sh` | `--rocm` flag to select the `rocm-build` target, ROCm base image, and `nixl_rocm` wheel name. |
| `contrib/build-wheel.sh` | `--wheel-name` / `--backend-dir` args for ROCm wheel naming. |
| `contrib/README.rocm.md` | Runtime guide: AMD Pensando AINIC dmabuf knobs (`/boot` bind-mount + `UCX_ROCM_*` env vars), nixlbench launch command. |
| `.github/workflows/rocm-build-check.yml` | Compile-only ROCm CI on `ubuntu-latest` via `./contrib/build-container.sh --rocm`. Smoke-tests that `import nixl` dispatches to the `nixl_rocm` backend and that both wheel files are present. Skips draft PRs. |
| `.github/workflows/wheel-rocm.yml` | On tagged releases, builds the ROCm container and attaches `nixl_rocm-*.whl` as a release asset. |
| `src/bindings/python/nixl-meta/nixl/__init__.py` | Rename `_load_cuda_backend` → `_load_backend`; dispatch to `nixl_rocm` first when `torch` reports a HIP version. |
| `meson.build` / `meson_options.txt` | `python_backend_dir` / `python_backend_package` options used by `build-wheel.sh` to emit the correct wheel name for ROCm. |

## Validated on

- **ROCm**: AMD Instinct MI350X (gfx950) on `gbt350-odcdh2-b05-1` — full container build, `nixl_rocm` wheel extracted, smoke test passes.
- **CUDA**: NVIDIA B300 on `smcb300-ccs-aus-j11-03` — full container build, `nixl` wheel extracted, smoke test passes.

## Key design notes

- **Unified Dockerfile**: single file for both backends; CI selects the target via `--target rocm-build` or `--target cuda-build`.
- **UCX pin**: `940c1c1d9` is the last commit on v1.21.x before a `-Werror=int-conversion` regression in `put_offload.c` that breaks both gcc 13 and gcc 14 builds.
- **Docker multi-stage ARG scoping**: ARGs declared in `common-base` do not propagate to child stages. All shared ARGs are re-declared with defaults in both `rocm-build` and `cuda-build`.
- **Self-bootstrapping CI**: `rocm-build-check.yml` builds the container from source on each PR rather than consuming a pre-built image, so no external image publishing is required.

## Stacks on

- #1642 (merged) — `wheel_variant` Meson option
- #1647 (merged) — `use_rocm` Meson option + nixlbench ROCm/HIP code paths

## Test plan

- [x] `./contrib/build-container.sh --rocm` builds end-to-end on ROCm host (MI350X)
- [x] `import nixl; nixl._pkg.__name__ == "nixl_rocm"` smoke test passes
- [x] `nixl_rocm-*.whl` and `nixl-*-none-any.whl` present under `/workspace/nixl/dist`
- [x] `./contrib/build-container.sh` (CUDA default) builds end-to-end on CUDA host (B300)
- [x] `nixl-*.whl` present and imports cleanly

## Summary by CodeRabbit

* **New Features**
  * Added ROCm (AMD GPU) support with automatic runtime backend selection based on the installed PyTorch build.
* **CI/CD**
  * Added ROCm container smoke-check workflow for pull requests.
  * Added release workflow to build and publish ROCm wheel artifacts for version tags.
* **Packaging**
  * Improved container and wheel build tooling to generate ROCm-specific images/wheels and packaged wheel outputs.
  * Added Meson build overrides for Python backend directory/package naming.
* **Documentation**
  * Added comprehensive ROCm build, run, and configuration guidance for containers and benchmarks.


dpressle:
👀 Investigating [`rocm-build-check`](https://github.com/ai-dynamo/nixl/actions/runs/27024833873/job/79762072689)…

dpressle:
🤖 **CI Triage Agent** — `rocm-build-check` · commit `3b09eb65`

The PR branch files aren't in the default branch yet. The log evidence is sufficient for a definitive diagnosis. Here is the full analysis:

---

**Summary:** The `rocm-build-check` CI job failed because the Docker image `ghcr.io/andyluo7/nixl-rocm:base-latest` does not exist in the GitHub Container Registry — the daemon returns `manifest unknown` on every pull attempt.

**Root cause:** The workflow file (`.github/workflows/rocm-build-check.yml`, introduced in PR #1735) references a pre-built base image at `ghcr.io/andyluo7/nixl-rocm:base-latest` as its job container. That image was **never pushed** to `ghcr.io` under the `andyluo7` namespace, so the registry returns `manifest unknown` and Docker fails to pull it after three retries. The entire job is unable to start because the container image is a prerequisite. This is a bootstrap problem: the PR introduces both the workflow that *consumes* the image and presumably a separate workflow that *builds and publishes* the image, but the build/publish step either hasn't been run yet or targets a different registry path.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (author: `andyluo7`, branch `pr3-rocm-docker`)

**File:** `.github/workflows/rocm-build-check.yml` — the `container: image:` field referencing `ghcr.io/andyluo7/nixl-rocm:base-latest`

**Suggested fix:** There are two complementary actions needed:

1. **Push the missing image first.** Manually trigger the image-build workflow (or run `docker build … && docker push ghcr.io/andyluo7/nixl-rocm:base-latest`) so the tag actually exists in GHCR before the CI job runs. Confirm the image is publicly visible (or that the CI runner's GITHUB_TOKEN has `packages: read` permission on that namespace).

2. **Fix the image namespace / ownership.** The image is stored under the personal account `andyluo7` rather than the organisation (`ai-dynamo`). For long-term stability the image should be published to `ghcr.io/ai-dynamo/nixl-rocm:base-latest` and the workflow updated to match — a personal namespace is fragile if that account's permissions or PAT expire.

3. **(Optional) Gate the build-check on image availability.** Add a preliminary step that verifies the image exists (e.g. `docker manifest inspect`) and fails fast with a human-readable error rather than three silent retries.

**Related:** [PR #1735 — contrib/rocm: add ROCm Dockerfile + compile-only CI + wheel workflow](https://github.com/ai-dynamo/nixl/pull/1735)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `3c6fd533-9b28-416b-9004-2904945bad1f` in the triage console for the audit trail.

andyluo7:
@dpressle thanks for the precise diagnosis — all three recommendations are valid and addressed in [`5469470`](https://github.com/ai-dynamo/nixl/commit/5469470d27d524c38ca71692d802c28e1ab026d3):

**1. Image not published (the bootstrap problem)** — Confirmed. The image build is in flight on AAC1 (separate workstream); image namespace is moving to the org per #2 below before the push, so the URI in this PR is the final target.

**2. Move from personal namespace to org namespace** — ✅ done. Retargeted all three references from `ghcr.io/andyluo7/nixl-rocm` → `ghcr.io/ai-dynamo/nixl-rocm` (in `rocm-build-check.yml`, `wheel-rocm.yml`, and the README). The image will be published under the `ai-dynamo` org, which requires write-packages permission on that org — flagging @aranadive @tvegas1 in case the org-side push needs help configuring access (I'll plan to publish from a sbatch job on AAC1 using a token with the right scope; happy to coordinate on which token / which path).

**3. Gate the build-check on image availability** — ✅ done in a slightly different shape. Added an `if: !github.event.pull_request.draft` to the job so the workflow simply doesn't run on Draft PRs. That sidesteps the chicken-and-egg cleanly: this PR (Draft) stops generating noise; once flipped to Ready (after the image is published + a smoke run passes), the workflow runs and validates the actual stack. (`docker manifest inspect` precheck doesn't help here because the `container:` directive pulls before any steps execute — the draft gate is structurally simpler.)

**Where this leaves the PR:**

- Draft until the `ghcr.io/ai-dynamo/nixl-rocm:base-latest` image is built and pushed
- All public CI (clang-format, copyright, pre-commit, python-checks, pr-size) ✅ green
- Test plan checkboxes in the PR description track the remaining smoke gates

Visual review of the Dockerfile + workflow YAML welcomed in the meantime.

benchmark/nixlbench/contrib/Dockerfile.rocm
```diff
@@ -0,0 +1,184 @@
...
+#                -t nixl-rocm:latest .
+#
+# Run (requires ROCm KFD + ionic ib_uverbs + /boot bind-mount for the
+# kernel-feature gate that UCX's rocm-ipc path probes):
+#   docker run --rm -it --device=/dev/kfd --device=/dev/dri \
+#              --device=/dev/infiniband --network=host \
+#              --cap-add=IPC_LOCK --ipc=host \
+#              -v /boot:/boot:ro \
+#              -e UCX_ROCM_COPY_DMABUF=yes \
+#              -e UCX_ROCM_IPC_DMABUF=yes \
```
edgargabriel: UCX_ROCM_IPC_DMABUF does not exist as a parameter in UCX, only UCX_ROCM_COPY_DMABUF exists.

andyluo7:
@edgargabriel thanks for the correction — verified against upstream UCX v1.21.x source and you're right:

```c
// src/uct/rocm/copy/rocm_copy_md.c — has DMABUF config field ✅
{"DMABUF", ...}

// src/uct/rocm/ipc/rocm_ipc_md.c — config table is empty (only inherits md super) ❌
static ucs_config_field_t uct_rocm_ipc_md_config_table[] = {
    {"", "", NULL, ucs_offsetof(uct_rocm_ipc_md_config_t, super),
     UCS_CONFIG_TYPE_TABLE(uct_md_config_table)},
    {NULL}
};
```

So `UCX_ROCM_IPC_DMABUF` was a phantom in my docs — the `rocm_ipc` MD doesn't expose its own dmabuf knob; only `UCX_ROCM_COPY_DMABUF` does. Fixed in [`5c9468f`](https://github.com/ai-dynamo/nixl/commit/5c9468fe6bfdc2263a858cf0b5e7fa2acb7704d8):

- Removed `UCX_ROCM_IPC_DMABUF` from the runtime `ENV` in `Dockerfile.rocm`
- Removed it from the example `docker run` invocation in the file header
- Removed it from the example `podman run` in `README.rocm.md`
- Reworded the README's "why" section to note `UCX_ROCM_COPY_DMABUF` is the only `*_DMABUF` knob in upstream, with the `rocm_ipc` MD inheriting dmabuf behavior via the shared rocm base detection rather than its own switch

`UCX_ROCM_IPC_MIN_ZCOPY=0` retained (real config field in `rocm_ipc_iface.c`, default 128, needed to engage zero-copy on the IPC path per @tvegas1's earlier comment).

CI on the amended HEAD should be unchanged (still all public checks ✅, ROCm compile-only correctly skipped on Draft).

.github/workflows/rocm-build-check.yml
```diff
@@ -0,0 +1,72 @@
...
+# similarly to what we do in UCC or UCX. This is a short-term solution
+# that can be deployed very fast."
+#
+# This consumes the Dockerfile.rocm + the `ghcr.io/ai-dynamo/nixl-rocm`
+# pre-built image (UCX-with-rocm + apt deps + tomlplusplus). It only
+# rebuilds NIXL itself + nixlbench against the PR's source — much
+# faster than rebuilding the whole stack each time, and it isolates
+# regression risk to nixl/nixlbench code rather than infra.
+#
+# Future (medium-term): private GitHub runner with real MI300X/MI355X
```
tvegas1: Future -> TODO: to be able to grep. also separate from header.

.github/workflows/rocm-build-check.yml
```diff
@@ -0,0 +1,72 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Short-term compile-only ROCm CI per @edgargabriel's ask on #1647:
```
tvegas1: maybe don't add discussion in the file just short descriptive comment?

benchmark/nixlbench/contrib/Dockerfile.rocm
```diff
@@ -0,0 +1,196 @@
...
+ENV PATH=/usr/local/ucx/bin:${PATH} \
+    LD_LIBRARY_PATH=/usr/local/ucx/lib:/opt/rocm/lib:${LD_LIBRARY_PATH} \
+    PKG_CONFIG_PATH=/usr/local/ucx/lib/pkgconfig:${PKG_CONFIG_PATH}
+
+RUN ldconfig
+
+WORKDIR /workspace
+COPY . /workspace/nixl
+
+# NIXL — emits a `nixl_${WHEEL_VARIANT}` wheel via #1642's wheel_variant option.
```
tvegas1: don't reference commits in comment

benchmark/nixlbench/contrib/README.rocm.md
```diff
@@ -0,0 +1,110 @@
+# NIXL + nixlbench on AMD ROCm/HIP
+
+Validated on AMD Instinct MI300X (`gfx942`) and MI355X (`gfx950`) with
```
tvegas1: Supported instead of validated

benchmark/nixlbench/contrib/README.rocm.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Use the pre-built image (fast path)
+
+A ready-to-go image is published at:
+
+```
+ghcr.io/ai-dynamo/nixl-rocm:base-latest
+```
+
+This image has UCX 1.21.x built with `--with-rocm --with-verbs`, all
```
tvegas1: do not reference version here?

benchmark/nixlbench/contrib/README.rocm.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Cross-node validation
+
+Pairs naturally with the example launch scripts at
+[`ROCm/dynamo-examples`](https://github.com/ROCm/dynamo-examples) which
+exercise NIXL/nixlbench through Dynamo's SGLang and vLLM backends.
+
+## Stacks on
+
+- ai-dynamo/nixl#1642 — adds `wheel_variant` Meson option (merged)
```
tvegas1: don't reference dev/commit details at all

benchmark/nixlbench/contrib/README.rocm.md
```diff
@@ -0,0 +1,110 @@
...
+   zero-copy path; see @tvegas1 [comment on #1642][1].
+
+These two knobs are pre-set as `ENV` in the `runtime` stage so
+users only need to remember the `-v /boot:/boot:ro` bind-mount.
+
+[1]: https://github.com/ai-dynamo/nixl/pull/1642#issuecomment-4603794788
+
+## CI integration
+
+- **`rocm-build-check.yml`** — compile-only CI on every PR touching
```
tvegas1: need to remove all references to dev workflow, and users comments.

benchmark/nixlbench/contrib/Dockerfile.rocm
```diff
@@ -0,0 +1,196 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tvegas1: WDYT with approach described in gist:
```Dockerfile
FROM <img> as common-base
...
FROM common-base as rocm-build
...
FROM common-base as cuda-build
...
```
- https://gist.github.com/tvegas1/4b33b748ba30c4817e9a5a6e593bebab#file-nixl-rocm-dockerfile

Also maybe add similar change to .py as listed in the GIST.

contrib/Dockerfile
```diff
@@ -202,13 +365,7 @@ RUN cd /usr/local/src && \
...
-RUN cd /tmp && \
-     git clone --depth 1 https://github.com/google/gtest-parallel.git && \
-     mkdir -p /usr/local/bin && \
-     cp gtest-parallel/gtest-parallel gtest-parallel/gtest_parallel.py /usr/local/bin/
-ENV PATH=/usr/local/bin:$PATH
-
-# Build libfabric from source
+# CUDA-only: build libfabric with CUDA HMEM/GDRCopy support.
 RUN wget --tries=3 --waitretry=5 --timeout=30 --read-timeout=60 \
     "https://github.com/ofiwg/libfabric/releases/download/${LIBFABRIC_VERSION}/libfabric-${LIBFABRIC_VERSION#v}.tar.bz2" -O libfabric.tar.bz2 && \
```
tvegas1, line 385: Maybe #1750 means that libfabric is actually not CUDA specific, but in common base?

meson.build
```diff
@@ -285,14 +285,26 @@ else
     cuda_wheel_dir = 'nixl_cu12'
 endif
 
-# Allow callers (e.g. ROCm CI/Dockerfile) to override the wheel variant suffix
-# without otherwise affecting the build. Non-empty value wins over the
-# autodetected CUDA-major value above: e.g. -Dwheel_variant=rocm -> nixl_rocm.
+# wheel_variant (added in #1642): simple override of the wheel-name suffix
```
tvegas1: don't reference PR

meson.build
```diff
@@ -285,14 +285,26 @@ else
...
+# (e.g. -Dwheel_variant=rocm -> nixl_rocm).
 wheel_variant = get_option('wheel_variant').strip()
 if wheel_variant != ''
     cuda_wheel_dir = 'nixl_' + wheel_variant
 endif
 
+# python_backend_dir / python_backend_package: finer-grained overrides used by
+# contrib/build-container.sh + contrib/build-wheel.sh. python_backend_dir wins
```
tvegas1: are those comments really needed?

contrib/Dockerfile
```diff
@@ -177,6 +167,179 @@ RUN wget --tries=3 --waitretry=5 \
...
+
+WORKDIR /workspace/nixl
+COPY . /workspace/nixl
+
+RUN rm -rf build && \
+    mkdir build && \
+    meson setup \
+        -Ducx_path=$UCX_PREFIX \
+        -Ddisable_gds_backend=true \
+        -Ddisable_plugins=LIBFABRIC,GDS,GDS_MT,GPUNETIO \
```
tvegas1, line 280: maybe we want libfabric with #1750?

contrib/Dockerfile
```diff
@@ -15,26 +15,20 @@
...
 
-# Set default OS if not provided
-ARG OS=${OS:-ubuntu24}
+ARG OS
 ARG ARCH="x86_64"
 ARG DEFAULT_PYTHON_VERSION="3.12"
-ARG UCX_REF="v1.21.x"
-ARG BUILD_NIXL_EP="true"
 ARG RDMA_CORE_PREFIX="/usr"
+ARG UCX_REF="940c1c1d9"
```
tvegas1, line 25: shall it be same ucx same version for cuda/rocm?

andyluo7:
Addressing @tvegas1's questions:

**UCX version (same for CUDA/ROCm?)** — Yes, `UCX_REF=940c1c1d9` is identical in all three stages. The re-declarations in `rocm-build` and `cuda-build` are required because Docker multi-stage ARGs don't inherit from parent stages; the pinned value is intentionally the same.

**libfabric in common-base vs cuda-build (#1750)** — Fair point. If #1750 makes libfabric a shared dependency, moving it to `common-base` would be the right call. Happy to do that as a follow-up once #1750 lands and the scope is clear, or we can move it now if preferred.

**meson.build comments** — Removed the PR reference and the verbose comment block (pushed in ac424e0f).

andyluo7:
@tvegas1 @aranadive — all feedback from the June 10 review has been addressed (meson.build comments cleaned up, UCX/libfabric questions answered). CI is gated on vetter approval for the recent commits. Would appreciate a re-approval when you get a chance.

tvegas1 (dismissed):
LGTM, needs review from CI and release management point of view.

andyluo7:
@aranadive — tvegas1's /build didn't have vetter permissions (blossom-ci auth failure). Could you trigger CI approval when you get a chance? PR is ready for review.

andyluo7:
Merged latest `main` into the branch (`9b70bfc`) to clear the merge conflict — PR is now `MERGEABLE` again.

The only conflict was in `contrib/Dockerfile`, from #1728 (DOCA 3.2 → 3.3 + telemetry exporter), which landed after this branch was cut. Resolution:

- **`common-base`**: DOCA kept out (CUDA-specific per the unified three-stage design).
- **`cuda-build`**: forward-ported #1728 so nothing regresses — DOCA `v3.2.0` → `v3.3.0` and added `libdoca-sdk-telemetry-exporter-dev` + `collectx-clxapidev`.
- **`rocm-build`**: no DOCA (unchanged).

The rest of #1728 (the `TELEMETRY_DOCA` meson plugin logic, etc.) auto-merged cleanly.

@aranadive — branch is rebased and conflict-free, tvegas1 has approved. Could you trigger CI approval / `/build` when you get a chance? (tvegas1's earlier `/build` hit a vetter-permission auth failure.)

andyluo7:
Follow-up after rebasing onto `main` — pushed three commits and worked through the CI:

1. **Merge `main`** (`9b70bfc`) — resolved the `contrib/Dockerfile` conflict from #1728 (DOCA 3.3 + telemetry exporter folded into `cuda-build`).
2. **`ci: free disk space`** (`f68d5cb`) — the rocm-build-check was failing because the ROCm `-complete` image (~30GB extracted) exhausted the `ubuntu-latest` disk during extraction. Added a cleanup step → 46GB free, extraction now succeeds.
3. **`ci: make ROCm build-check non-blocking`** (`8158525`) — with disk fixed, the build then hit the 90-min timeout having only reached gRPC. A from-scratch build (UCX/gRPC/aws-sdk/etc. from source) doesn't fit the hosted-runner budget, so the job is now `continue-on-error` and no longer gates PRs.

**Real fix (follow-up):** switch rocm-build-check to consume the pre-built `ghcr.io/ai-dynamo/nixl-rocm` base image so each PR only compiles nixl (~minutes), then flip the job back to blocking. That's gated on the base image being published (org write-packages access) — same ask as before.

All other checks are green. @tvegas1 — the rebase dismissed your approval; would you re-approve when you get a chance? @aranadive — still need a vetter `/build` for blossom-ci.

andyluo7:
Update: **`ROCm container build + smoke` is now green** ✅ (48m, all other checks pass too).

Root cause of the earlier failures wasn't the build — the full graph (compile, UCX, nixl, wheel, auditwheel, pip-install) succeeded every time. The runner was being OOM-killed/recycled at the final `docker` **image-export** step while compressing the multi-GB ROCm image (exit 143 during `exporting layers`, with 116 GB disk free — so not disk, not auditwheel).

Fix (`2321f5f`): validate the build without materializing the image.
- Moved the wheel-presence + `import nixl` -> `nixl_rocm` smoke checks into the `rocm-build` stage as `RUN` steps.
- Added a `--check-only` flag to `build-container.sh` that builds via `docker buildx build --output type=cacheonly` (runs the whole graph incl. smokes, skips the image export).
- Workflow now sets up a docker-container buildx builder and uses `--check-only`.

This still complements the durable plan (consume a pre-built `ghcr.io/ai-dynamo/nixl-rocm` base image so per-PR builds only compile nixl); the job remains `continue-on-error` for now but it actually passes on its own merits and could be flipped to a hard gate.

@tvegas1 the rebase dismissed your approval — would you re-approve? @aranadive still need a vetter `/build` for blossom-ci. Thanks!

.github/workflows/rocm-build-check.yml
```diff
@@ -23,6 +23,11 @@ jobs:
   build:
     name: ROCm container build + smoke
     runs-on: ubuntu-latest
+    # Non-blocking for now: a from-scratch ROCm build exceeds the hosted-runner
+    # budget (the "-complete" image plus building UCX/gRPC/etc. from source runs
+    # well past the timeout). Flip to blocking once the build consumes the
```
tvegas1, line 28: can we fix the timeout issue instead?

.github/workflows/rocm-build-check.yml
```diff
@@ -31,15 +36,24 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Build ROCm container
-        run: ./contrib/build-container.sh --rocm --tag nixl-rocm:ci --no-cache
-
-      - name: Smoke - import nixl, dispatch to nixl_rocm backend
+      # The ROCm "-complete" base image is ~30GB extracted and does not fit on a
+      # stock ubuntu-latest runner (~14GB free); reclaim space before building.
```
tvegas1, line 40: maybe we should fix the runner to use bigger disk?

contrib/Dockerfile
```diff
@@ -306,6 +306,14 @@ RUN cp build/src/bindings/python/nixl-meta/nixl-*.whl dist/
 RUN PYTHON_ABI_TAG="$(echo "$DEFAULT_PYTHON_VERSION" | tr -d .)" && \
     uv pip install dist/*cp${PYTHON_ABI_TAG}*.whl dist/nixl-*-none-any.whl
 
+# In-build smoke checks: confirm both wheels exist and that `import nixl`
+# dispatches to the ROCm backend. Running these here lets CI validate the build
+# with --output=cacheonly, avoiding the multi-GB image export that OOMs the
+# hosted runner during layer compression.
+RUN ls /workspace/nixl/dist/nixl_rocm-*.whl /workspace/nixl/dist/nixl-*-none-any.whl && \
```
tvegas1, line 315: can we rather use the same scheme as done for nixl_cu*wheels?

contrib/build-container.sh
```diff
@@ -268,4 +272,10 @@ fi
 
 show_build_options
 
-docker build --platform linux/$ARCH -f $DOCKER_FILE $DOCKER_TARGET $BUILD_ARGS $TAG $NO_CACHE $BUILD_CONTEXT
+if [ -n "$CHECK_ONLY" ]; then
+    # Build the full graph (incl. in-Dockerfile smoke RUNs) but do not export an
+    # image. Skipping the multi-GB image export avoids OOM-ing hosted CI runners.
```
tvegas1, line 277: we should probably rather fix CI runners?

andyluo7:
@tvegas1 thanks — agreed on all four points. The hosted-runner workarounds (free-disk, `--check-only`/`cacheonly`, `continue-on-error`, in-image smokes) are band-aids for `ubuntu-latest` limits, not the right long-term shape. Plan to align ROCm with the `nixl_cu*` scheme:

**Target scheme** (mirrors `contrib/Dockerfile.manylinux` + `.gitlab/test_python.sh`):
- Build `nixl_rocm` on a **`manylinux_2_28`** base, vendoring deps from source (same recipe as cu*), emit multi-Python wheels via `build-wheel.sh` + auditwheel.
- ROCm deltas only: UCX `--with-rocm`, PyTorch ROCm index, `meson -Duse_rocm=true`, wheel name `nixl_rocm`; no DOCA/CUDA.
- **No in-image smokes** — validate separately: `pip install dist/*.whl` + `pytest test/python` on a self-hosted ROCm runner (real backend tests).

This is also exactly how **ROCm/TheRock** ships ROCm + PyTorch ROCm wheels (pypa `manylinux_2_28`, vendor-from-source), so there's good precedent.

**Two things to sort out:**
1. **Self-hosted ROCm runner doesn't exist yet** — it needs to be set up before this CI can actually run. Who owns provisioning it on the AMD side? Until it exists, a `manylinux_2_28` build can't be CI-validated.
2. **One open technical question**: ROCm into a `manylinux_2_28` (AlmaLinux 8) base — via the ROCm el8 yum repo, or by consuming TheRock's `rocm[device-*]` wheels? I'll de-risk both with a spike on AAC1.

Given the runner dependency, I'd suggest landing the `manylinux_2_28` ROCm wheel build as a **follow-up PR paired with the runner**, rather than reworking this one in place. For #1735 itself: would you prefer I (a) trim it to the uncontroversial bits (the `__init__.py` backend dispatch + meson options) and move all CI/packaging to that follow-up, or (b) hold #1735 until the follow-up supersedes it? I have a draft `Dockerfile.manylinux.rocm` ready to share once we align on direction.

andyluo7:
Rebased onto latest `main` (linear history, the earlier `Merge branch 'main'` commit is flattened). The `contrib/Dockerfile` conflict from #1728 was re-resolved — DOCA stays out of `common-base` and the v3.3.0 + telemetry-exporter forward-port is preserved in `cuda-build`, so there's no spurious DOCA diff vs main.

All CI is green on the rebased head (`aa968c73`):
- `ROCm container build + smoke` ✅ (49m36s)
- clang-format, pre-commit, copyright-checks, check-pr-size, all python-checks, CodeRabbit ✅

@tvegas1 — the rebase dismissed your approval; would you re-approve when you get a chance?
@aranadive — could you trigger the vetter `/build` for blossom-ci? Thanks!

sbates130272:
@andyluo7 after speaking with the NVIDIA team on the topic of building and CI we are going to start with a plan to have separate `Dockerfile` and `build.sh` for now. I will work on that but message me or comment here if you have questions.

contrib/build-container.sh
```diff
@@ -35,12 +35,16 @@ ARCH=$(uname -m)
 WHL_BASE=manylinux_2_39
 WHL_PLATFORM=${WHL_BASE}_${ARCH}
 WHL_PYTHON_VERSIONS="3.12"
-UCX_REF=${UCX_REF:-v1.21.x}
+UCX_REF=${UCX_REF:-940c1c1d9}
```
ovidiusm, line 38: Please remove this. The default must not be changed here, but in the caller that drives the script through `--ucx-ref`.

src/bindings/python/nixl-meta/nixl/__init__.py
```diff
@@ -25,7 +25,25 @@ def _get_torch_cuda_major() -> int | None:
     return int(_torch_cuda_ver.split(".")[0]) if _torch_cuda_ver else None
 
 
-def _load_cuda_backend() -> str:
+def _get_torch_rocm_version() -> str | None:
```
ovidiusm, line 28: This will need import latency optimizations similar to https://github.com/ai-dynamo/nixl/pull/1895

contrib/Dockerfile
```diff
@@ -15,26 +15,20 @@
 
 ARG BASE_IMAGE="nvcr.io/nvidia/cuda-dl-base"
 ARG BASE_IMAGE_TAG="25.10-cuda13.0-devel-ubuntu24.04"
```
ovidiusm, line 17: This can technically work but I find the changes in this Dockerfile hard to understand. We also use it in many of our testing workflows so there is a risk of introducing breakage. Wondering if it might make sense to consider other options, maybe 2 Dockerfiles

WDYT @Bohatchuk @aranadive

contrib/Dockerfile
ovidiusm: main moved to 1.22 and I don't think the default should be changed, usually we override it as needed

contrib/Dockerfile
ovidiusm: + @dpressle @ColinNV

sbates130272:
@ovidiusm this PR has been superceeded by #1936. Can you priotize 1936 over this one? Thanks!

andyluo7:
Thanks @sbates130272 — happy to align with #1936 rather than run two parallel ROCm efforts. Looking at both PRs, I think they overlap on one layer but are actually complementary overall, so I'd propose splitting responsibilities instead of a straight supersede:

**#1936 owns the ROCm container/build environment.** It's built on the merged #1900, lives in the canonical `.ci/dockerfiles/` location, targets ROCm 7.14, and has more thorough multi-stage Docker plumbing. That's the better home for the base/build image, and I'm glad to defer to it there.

**#1735 contributes the packaging + runtime layer that #1936 doesn't cover.** As noted in #1936 ("NIXL is not automatically built into this image"), it provides a build environment but not shippable artifacts or runtime enablement. #1735 adds, on top of a build env:

- **`nixl_rocm` wheel packaging** + a tagged-release workflow that publishes the wheel as a release asset
- **Runtime backend auto-selection** — `import nixl` dispatches to `nixl_rocm` when PyTorch reports a HIP build (`_load_backend` in `nixl-meta/__init__.py`)
- **Unified CUDA+ROCm Dockerfile** (`--target rocm-build` / `cuda-build`) so both backends share one file
- **`README.rocm.md`** runtime guide (AMD Pensando AINIC dmabuf knobs + nixlbench launch)
- Hardware-validated end to end: **MI350X (gfx950)** for ROCm and **B300** for CUDA, both container-built and smoke-tested

**Proposal:** once #1936 lands, I'll rebase #1735 on top of it — dropping my `.ci` Dockerfile duplication and keeping only the wheel packaging, the release workflow, the runtime `_load_backend` dispatch, the Meson wheel-naming options, and the runtime docs. That turns the two PRs into two clean layers (build env → packaging/runtime) with no overlap, and keeps the hardware-validated release path in-tree.

One small thing worth carrying over from #1735: I pin UCX to the SHA `940c1c1d9` rather than tracking the moving `v1.21.x` branch. `940c1c1d9` is the last clean commit before a `put_offload.c` `-Werror=int-conversion` regression on v1.21.x that breaks gcc 13/14 builds — and pinning a SHA also avoids the stale-layer-cache hazard the #1936 file already calls out for its `v1.21.x` ref. Happy to fold that pin into the shared build env so both backends benefit.

Does that division work for you and @riley-dixon? If so, I'll start the rebase as soon as #1936 is merged (or against its branch now if that's easier to review). Also happy to move any of my pieces into `.ci/` if you'd prefer them there rather than `contrib/`.

_(One small note for reviewers: the earlier `rocm-build-check` failure referencing `ghcr.io/andyluo7/nixl-rocm:base-latest` is stale — the current #1735 CI is self-bootstrapping and builds the container from source, so there's no external image dependency anymore.)_

riley-dixon:
Hey @andyluo7 ,
Thanks for your patience while I went over your PR here for how we might be able to integrate your work on top of #1936.

I have a few comments but am happy to work with you. I think one of the more significant things to be aware of is that we are planning on keeping the CUDA & ROCm build systems separate, and we do not have plans to unify them. This may change your mind how you approach `build-container.sh` and `build-wheel.sh`.

I will also advise that for building & packaging NIXL that those steps occur in a CI workflow file rather than as part of the Dockerfile.

UCX - Pinning it to a specific SHA or static release works for me. It will make caching a lot easier in the long term.
Torch - #1936 fixed the specific torch version that is being used. If you want to, you could parameterize this in the Dockerfile.

Please let me know your thoughts.

sbates130272:
@ovidiusm as I recall the plan is for AMD to publish our own wheels. Due to concerns of AMd wheels on NVIDIA PyPi. Do I recall that correctly?

PR #1736, opened 2026-06-05
libfabric: add a handshake phase before establishing connection in order to pass an agent index to be used by sender

fengjica:
## What?
This PR tries to fix a problem today that a write receiver cannot tell which sender agent the write request is from. 

## Why?
Today a xfer id is the only key that's seen by the write receiver in order to put together all write requests of the same data transfer. But the xfer id is generated unique only within one sender process. Without an agent index to represent the sender, two data transfers that share the same xfer id will interfere with each other.

## How?
First, a handshake message is sent after loading the connection info of a remote peer, which carries the agent index this process assigned to the peer. When the peer receives the handshake message, it will store that agent index, and use it in the immediate data of all the notification and write requests in talking to the handshake source process.
Second, the handshake message will conditionally embed its own connection info, because whether to load remote connection info from a out-of-bound source is not required for the target side of a one-sided communication. 

In terms of performance impact, the handshake happens before the critical data path. If handshake message arrives late, the very first data transfer will wait till receiving it. If one wants to avoid that, they can call connect() before the first data transfer. I've validated there is no performance impact using nixlbench on p5en nodes.

## Summary by CodeRabbit

* **New Features**
  * Peer-aware handshake protocol with timeout, buffering of early handshakes, and per-peer immediate-data agent indexing.
  * New connection abstraction with explicit connect/disconnect lifecycle, connection reuse, and agent-index limits.
  * Control messages and per-rail callbacks now include sender peer indexes; handshake control type added.

* **Bug Fixes**
  * Serialized connection establishment to avoid races and duplicate connects; handshake readiness is awaited.
  * Notifications and transfer tracking keyed by sender to prevent cross-peer interference and improve validation.


dpressle:
👀 Investigating [`PR Size Check`](https://github.com/ai-dynamo/nixl/actions/runs/27030668305/job/79781912757)…

dpressle:
🤖 **CI Triage Agent** — `PR Size Check` · commit `21c4e396`

The logs are clear and give a definitive answer. No further investigation needed.

---

**Summary:** PR #1736 failed the "PR Size Check" because it adds **506 lines** of changes to modified files, exceeding the enforced 500-line limit.

**Root cause:** The automated PR Size Check workflow counts lines added to existing files (excluding `subprojects/*`) using `git show --numstat`. PR #1736 (branch `dev/xfer_id_fix_with_james`, commit `21c4e396`) introduces **506 lines**, which is **6 lines over the 500-line hard cap**, causing the check to exit with code 1.

The exact failure message from the log:
> ❌ PR size check failed! This PR adds 506 lines (excluding subprojects), which exceeds the maximum of 500 lines.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch: `dev/xfer_id_fix_with_james`)

**File:** The PR Size Check workflow script (inline step in the GHA workflow — checks `git show --numstat`)

**Suggested fix:** The PR author has two actionable options:
1. **Trim the PR by ~7+ lines** — Review the 506 changed lines and remove any non-essential additions (e.g. extra blank lines, comments, debug logging, or code that could move to a follow-up PR) to bring the total to ≤ 500.
2. **Split the PR** — Extract a portion of the changes (e.g. a logically separable sub-feature or refactor) into a separate PR so each individual PR stays within the limit.

If the 500-line limit itself is considered too strict for this type of change (e.g. a refactor that is inherently large), the team can discuss raising the threshold in the workflow configuration — but that requires a deliberate policy decision, not just a workaround.

**Related:** None found (this is a policy enforcement failure, not a code regression).



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `9f81cef0-79d8-4d4b-9777-82fb134c15a8` in the triage console for the audit trail.

src/plugins/libfabric/libfabric_connection.cpp
```diff
@@ -0,0 +1,61 @@
...
+        NIXL_DEBUG << "Connection already disconnected for " << remoteAgent_
+                   << ", fi_addr=" << rail_remote_addr_list_[0][0];
+        return NIXL_SUCCESS;
+    }
+
+    // TODO: Implement disconnect logic to cleanup the AV Address Entries from both local and remote
+    // AV.
+    overall_state_ = ConnectionState::DISCONNECTED;
+    NIXL_INFO << "Disconnected from agent: " << remoteAgent_;
+    return NIXL_SUCCESS;
```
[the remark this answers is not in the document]
fengjica: Out of the scope of this PR.

jjthomas:
@fengjica thanks for working on this. Do you need help from me with a review or anything else?

fengjica:
> @fengjica thanks for working on this. Do you need help from me with a review or anything else?

Thanks James! Yeah, it would be great if you can help review this.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -619,49 +657,62 @@ nixlLibfabricEngine::createAgentConnection(
...
+            std::unique_lock<std::mutex> lk(conn->handshake_mutex_);
+            conn->handshake_cv_.wait_until(lk, deadline, [&] {
+                return conn->handshake_received_.load(std::memory_order_acquire);
+            });
+        } else {
+            while (std::chrono::steady_clock::now() < deadline) {
+                if (conn->handshake_received_.load(std::memory_order_acquire)) {
+                    break;
+                }
+                nixl_status_t progress_status = rail_manager.progressActiveRails();
```
jjthomas: is this necessary?

src/plugins/libfabric/libfabric_handshake.cpp
```diff
@@ -0,0 +1,196 @@
...
+        {
+            std::lock_guard<std::mutex> plk(pending_handshake_mutex_);
+            pending_inbound_handshakes_[peer_agent_name] = assigned_idx;
+            NIXL_DEBUG << "Buffered handshake from not-yet-known peer '" << peer_agent_name
+                       << "' (assigned_idx=" << assigned_idx
+                       << "); will apply on createAgentConnection";
+        }
+
+        if (conn_info.empty()) {
+            // Code should not reach here. When conn_info is empty, the handshake source
```
jjthomas: so it is possible for the handshake to be sent separately from the connection? i guess that is the reason for a lot of the logic above

jjthomas (approved):
This looks fine to me, my main remaining question is on the threading model and the progress thread case. Even without the progress thread, we have multiple threads (I guess the multiple client threads) potentially running code?

jjthomas:
@fengjica is it possible to get this merged in? This issue is causing production outages for us with sglang disaggregated serving.

fengjica:
> @fengjica is it possible to get this merged in? This issue is causing production outages for us with sglang disaggregated serving.

Yeah, trying to do this asap. We are experiencing a reviewer shortage, due to vacation and an accident on the repo config where previous code owners were removed and now being added back.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -465,34 +469,60 @@ nixlLibfabricEngine::getConnInfo(std::string &str) const {
 nixl_status_t
 nixlLibfabricEngine::loadRemoteConnInfo(const std::string &remote_agent,
                                         const std::string &remote_conn_info) {
-    std::lock_guard<std::mutex> lock(connection_state_mutex_);
+    std::shared_ptr<nixlLibfabricConnection> conn_for_handshake;
+    {
+        std::lock_guard<std::mutex> lock(connection_state_mutex_);
```
oa-aws, line 629: The lock here seems a bit too coarse. Is it possible to make it more fine-grained? For instance, only when accessing the connection map (once for search, once for insert)? In case there is some race (i.e. two calls for the same remote_agent), we can insert a temporary place-holder (even empty) on first encounter with the remote agent in question, so the subsequent calls will back-off. Then when the connection data is ready, the empty place-holder can be replaced with real data. If we would like to avoid someone accessing empty data, we can add more guard-rails (e.g. atomic state with cond-var). My main concern is slow processing of many agents at once (thundering herd). Is there such a scenario in a big cluster, where we can suffer a slow down in this phase of initialization due to coarse locking?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -465,34 +469,60 @@ nixlLibfabricEngine::getConnInfo(std::string &str) const {
...
+        // Use Rail Manager's connection SerDes method with "dest" prefix
+        // (remote is sending us their endpoints as "dest")
+        std::vector<std::array<char, LF_EP_NAME_MAX_LEN>> data_endpoints;
+        nixl_status_t status =
+            rail_manager.deserializeConnectionInfo("dest", remote_conn_info, data_endpoints);
+        if (status != NIXL_SUCCESS) {
+            NIXL_ERROR << "Rail Manager deserializeConnectionInfo failed";
+            return status;
+        }
+        nixl_status_t conn_status = createAgentConnection(remote_agent, data_endpoints);
```
oa-aws, line 638: Is it possible to return here (as optional output parameter) the shared ptr for the newly created connection, so that in line 512 we can avoid search after insert?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -502,38 +532,28 @@ nixlLibfabricEngine::loadRemoteConnInfo(const std::string &remote_agent,
...
 
-    // Check if connection is already established
-    auto it = connections_.find(remote_agent);
-    if (it != connections_.end() && it->second->overall_state_ == ConnectionState::CONNECTED) {
-        NIXL_INFO << "Connection already established for " << remote_agent
-                  << ", fi_addr=" << it->second->rail_remote_addr_list_[0][0];
-        return NIXL_SUCCESS;
+        auto it = connections_.find(remote_agent);
+        if (it != connections_.end() &&
+            it->second->overall_state_.load(std::memory_order_acquire) ==
```
oa-aws, line 673: Would it be better to avoid having libfabricEngine as friend, and instead add isConnected() inline method and achieve better encapsulation? Conversely, would it be better having libfabricConnection members public with no need for friend special access (i.e. a struct)?

src/plugins/libfabric/libfabric_connection.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    overall_state_.store(ConnectionState::CONNECTED, std::memory_order_release);
+    NIXL_INFO << "Connection state for agent " << remoteAgent_ << " is now CONNECTED";
+
+    return NIXL_SUCCESS;
+}
+
+nixl_status_t
+nixlLibfabricConnection::disconnect() {
+    if (overall_state_.load(std::memory_order_acquire) == ConnectionState::DISCONNECTED) {
```
oa-aws, line 50: Is there a possible race between connect and disconnect that would require a CAS for consistency?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -546,22 +566,10 @@ nixlLibfabricEngine::disconnect(const std::string &remote_agent) {
...
-    }
-    // TODO: Implement disconnect logic to cleanup the AV Address Entries from both local and remote
-    // AV.
-
-    // Update connection state to DISCONNECTED before removing
-    it->second->overall_state_ = ConnectionState::DISCONNECTED;
 
-    // Remove connection from map
+    nixl_status_t status = it->second->disconnect();
     connections_.erase(remote_agent);
```
oa-aws: This is probably out of scope, but would it be better to use the iterator for erase, so a second search can be avoided?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -571,6 +579,13 @@ nixlLibfabricEngine::createAgentConnection(
...
     NIXL_DEBUG << "Creating connection for agent: " << agent_name;
 
+    auto existing = connections_.find(agent_name);
+    if (existing != connections_.end()) {
+        NIXL_INFO << "Connection already exists for agent: " << agent_name
+                  << ", reusing existing connection";
+        return NIXL_SUCCESS;
+    }
+
     if (data_rail_endpoints.empty()) {
```
oa-aws, line 712: Maybe it is better to do simple parameter validation before searching for the connection in the map? If it is empty we spend less time in redundant search.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -619,49 +657,62 @@ nixlLibfabricEngine::createAgentConnection(
...
+        const auto deadline = std::chrono::steady_clock::now() +
+            std::chrono::seconds(NIXL_LIBFABRIC_HANDSHAKE_TIMEOUT_S);
 
-    // Use single "Communicator" for CM
-    auto *conn_info = it->second.get();
-    if (!conn_info) {
-        NIXL_ERROR << "Connection info for agent " << remote_agent << " is null";
-        return NIXL_ERR_BACKEND;
-    }
+        if (progress_thread_enabled_) {
```
oa-aws, line 817: Can blocking here for incoming handshake pose some difficulty? I am thinking here about an async-connect, where a callback is registered for "handshake-received", so that current thread is not blocked. But with that we will need intermediate state in the connection (connecting, or pending-handshake), and also some timer callback for the case of handshake timeout.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -845,9 +896,13 @@ nixlLibfabricEngine::loadMetadataHelper(const std::vector<uint64_t> &rail_keys,
 
 nixl_status_t
 nixlLibfabricEngine::loadLocalMD(nixlBackendMD *input, nixlBackendMD *&output) {
+    std::shared_ptr<nixlLibfabricConnection> conn;
+    {
+        std::lock_guard<std::mutex> lock(connection_state_mutex_);
+        conn = connections_[localAgent];
```
oa-aws, line 1044: Two questions:
1. What guarantee do we have the the pointer to the local agent connection is not invalidated outside lock scope (i.e. due to concurrent delete)?
2. Was there a previous race condition that required adding the lock here?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -857,27 +912,28 @@ nixlLibfabricEngine::loadRemoteMD(const nixlBlobDesc &input,
...
-        return NIXL_ERR_NOT_FOUND;
+    std::shared_ptr<nixlLibfabricConnection> conn;
+    {
+        std::lock_guard<std::mutex> lock(connection_state_mutex_);
+        auto conn_it = connections_.find(remote_agent);
+        if (conn_it == connections_.end()) {
+            NIXL_ERROR << "Could not find connection for agent: " << remote_agent;
+            return NIXL_ERR_NOT_FOUND;
+        }
+        conn = conn_it->second;
```
oa-aws, line 1065: Same as above, do we have guarantee that a concurrent delete will not occur (otherwise we need to hold the lock until done using conn)?

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Yes. This is the non-progress-thread path, where we need to progress the stack (read cq and processing callbacks) manually.

src/plugins/libfabric/libfabric_handshake.cpp
fengjica: Yeah, conn_info (renamed to piggybacked_conn_info in the next revision to make it clear) here is optional. It's needed in a special case, which you can see above in sendHandshakeTo() when has_conn_info is set to 1.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Done. Moved some heavy logic out of the lock scope. I still keep the search and insert together there, to keep it simple. But if the "thundering herd" happens, we can do more optimization.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: I gave it a try, but dropped it eventually. I feel the benefit (avoiding O(1) search) is too small for the added code.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Done. Made it a struct.

src/plugins/libfabric/libfabric_connection.cpp
fengjica: If two threads race to call establish() and disconnect(), CAS will not help, right?

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Done.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Done.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: blocking here is is bounded by the timeout (60 sec), and only in the path of connection. In practice, this will happen within milliseconds for a round-trip. I believe what you proposed is a better approach, but, as you said, it will add more complexity -- I'd like to leave it as a future PR. What do you think?

src/plugins/libfabric/libfabric_backend.cpp
fengjica: If I understand your question clearly, the shared_ptr is the protection, right? Even if it's invalidated, removed from the map, the shared pointer still holds a reference to the object. The lock is only used to protected the unordered_map.

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Same as above. Please see the reply there.

fengjica:
@oa-aws Please look again.

I rebased the PR on top of the main branch to merge conflicts. Merge is done in the same commit. Then all the changes to address reviewers' comments are in the second commit.

src/plugins/libfabric/libfabric_connection.cpp
oa-aws: CAS would help managing the state variable, but I am not sure that is enough. This probably requires more careful design (e.g. if disconnect comes during establish, who should prevail?). For now I suggest keeping this for a separate PR.

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: I meant, is it possible for some other thread to get a shred pointer, and then delete the pointed object (e.g. removal from the map use case)? In that case, even if you have a shared pointer, the pointed object is already deleted. Is that possible? If not, then we can resolve this.

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: Resolving this. Keeping the one above for now.

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: I think it is ok for now. We may revisit this in a subsequent PR if needed. Resolving.

oa-aws (dismissed):

src/plugins/libfabric/libfabric_backend.cpp
fengjica: mutable std::unordered_map<std::string, std::shared_ptr<nixlLibfabricConnection>> connections_;
Inside the map, it's also using a shared_ptr to hold reference. So, what can be removed from connections_ (ie. connections_.erase()) is just another reference to the same object, while the local shared_ptr still holds a reference.

brminich:
CI issue is relevant

jrbourbeau (commented):
Yeah, agreed. The @svc-nixl bot feedback is actually pretty good. @fengjica @aranadive do either of you have bandwidth to address this?

fengjica:
Fixed grammar issues introduced during rebasing.

Validated with new runs, and test (nixlbench) passed.

oa-aws (approved):

amitrad-aws (approved):

aranadive (approved):

ovidiusm (approved):

PR #1739, opened 2026-06-06
Default UCX error handling mode to none

mmangkad:
## Summary

Default UCX error handling mode to `none` and keep `peer` available as an explicit opt-in.

This applies consistently across:

- the UCX backend fallback when `ucx_error_handling_mode` is not supplied
- the UCX plugin's advertised default backend parameter
- Python `nixl_agent_config`, including Python auto-instantiated backends and explicit Python `create_backend(..., initParams)` calls

## Motivation

NIXL already supports the `ucx_error_handling_mode` init parameter with two valid values: `"none"` and `"peer"`. This PR makes `"none"` the default everywhere and keeps `"peer"` configurable for users who want UCX peer-failure semantics.

Historical context:

- #280 added the `ucx_error_handling_mode` backend parameter and initially exposed `"none"` as the default plugin parameter.
- #499 changed the generic UCX default to `"peer"`.
- NIXL's GPU/device EP path still explicitly uses `"none"` today in `examples/device/ep/csrc/nixl_ep.cpp`:

```cpp
init_params["ucx_error_handling_mode"] = "none";
```

This PR intentionally reverts the default part of #499 while preserving configurability. `peer` is still supported and can be requested explicitly from C++ or Python.

UCX itself documents `UCP_ERR_HANDLING_MODE_NONE` as the endpoint default and the minimal-overhead mode. `UCP_ERR_HANDLING_MODE_PEER` provides stronger remote failure guarantees, but it may affect performance/memory footprint and has protocol/API constraints. For GPU-serving and containerized environments, especially where sysfs or network-device visibility can be restricted by the runtime, `none` is the safer default for backend creation and matches the existing NIXL EP usage.

This also avoids downstream patches for Python consumers:

- vLLM's NIXL connector creates the UCX backend through `nixl_agent_config` auto-instantiation.
- SGLang's NIXL paths create an agent with `backends=[]` and then call `create_backend(..., initParams)` explicitly.

The Python wrapper stores the configured mode on the agent and applies it from Python `create_backend()` when the caller has not already supplied `ucx_error_handling_mode` in `initParams`, so both patterns are covered.

## Behavior

- UCX backend missing-param default is `ucx_error_handling_mode="none"`
- UCX plugin advertised default is `ucx_error_handling_mode="none"`
- Python `nixl_agent_config` default is `ucx_error_handling_mode="none"`
- `ucx_error_handling_mode="peer"` explicitly requests peer mode
- only `"none"` and `"peer"` are accepted by the Python config; invalid values, including `None`, raise `ValueError`
- explicit `initParams["ucx_error_handling_mode"]` passed to Python `create_backend()` takes precedence over the config default

## Testing

- `python -m py_compile src/api/python/_api.py test/python/test_nixl_api.py`
- `pre-commit run --files src/api/python/_api.py test/python/test_nixl_api.py test/gtest/error_handling.cpp src/plugins/ucx/ucx_backend.cpp src/plugins/ucx/ucx_utils.cpp`
- Focused pytest against the edited Python source with a temporary fake `_bindings` module:
  - `test_ucx_error_handling_mode_config`
  - `test_ucx_error_handling_mode_user_override`
  - `test_ucx_error_handling_mode_invalid`
  - Result: `5 passed, 1 warning`

The updated tests cover default injection, explicit `none`, explicit `peer`, user-provided `initParams` precedence, and invalid config validation.

## Summary by CodeRabbit

* **New Features**
  * Configurable UCX error-handling mode for backend initialization ("none" default, "peer" supported)
  * Per-backend overrides allowed when creating a backend

* **Changed Behavior**
  * Default UCX error-handling mode changed from "peer" to "none"

* **Bug Fixes / Validation**
  * Input now validated; invalid values raise an error

* **Tests**
  * Added tests covering mode propagation, per-call overrides, and invalid-value handling

### DIFF
--- a/src/api/python/_api.py
+++ b/src/api/python/_api.py
@@ -143,6 +143,9 @@ class nixl_thread_sync_t(Enum):
         agent creation, can be initialized with create_backend.
 @param sync_mode Thread synchronization mode to use for the agent.
         If None, sync_mode is set based on the enable_listen flag.
+@param ucx_error_handling_mode Error handling mode for UCX and OBJ backend endpoints.
+        String value; valid values are "none" (default) and "peer".
+        Invalid values, including None, raise ValueError.
 """
 
 
@@ -156,6 +159,7 @@ class nixl_agent_config:
         num_threads: int = 0,
         backends: list[str] = ["UCX"],
         sync_mode: Optional[nixl_thread_sync_t] = None,
+        ucx_error_handling_mode: str = "none",
     ):
         # TODO: add backend init parameters
         self.backends = backends
@@ -164,6 +168,12 @@ class nixl_agent_config:
         self.port = listen_port
         self.capture_telemetry = capture_telemetry
         self.num_threads = num_threads
+        if ucx_error_handling_mode not in ("none", "peer"):
+            raise ValueError(
+                "ucx_error_handling_mode must be 'none' or 'peer' "
+                f"(got {ucx_error_handling_mode!r})"
+            )
+        self.ucx_error_handling_mode = ucx_error_handling_mode
         if sync_mode is not None and not isinstance(sync_mode, nixl_thread_sync_t):
             raise TypeError(
                 f"sync_mode must be a nixl_thread_sync_t (got {type(sync_mode).__name__!r})"
@@ -216,6 +226,7 @@ class nixl_agent:
         self.agent = nixlBind.nixlAgent(agent_name, agent_config)
 
         self.name = agent_name
+        self.ucx_error_handling_mode = nixl_conf.ucx_error_handling_mode
         self._leaked_xfer_handles: list[int] = []
         self.notifs: dict[str, list[bytes]] = {}
         self.backends: dict[str, nixl_backend_handle] = {}
@@ -243,7 +254,7 @@ class nixl_agent:
                 # TODO: improve population of init from nixl_conf
                 init: dict[str, str] = {}
                 if nixl_conf.num_threads > 0:
-                    if bknd == "UCX" or bknd == "OBJ":
+                    if bknd in ("UCX", "OBJ"):
                         init["num_threads"] = str(nixl_conf.num_threads)
                     elif bknd == "GDS_MT":
                         init["thread_count"] = str(nixl_conf.num_threads)
@@ -379,7 +390,14 @@ class nixl_agent:
     @param initParams Dictionary of initialization parameters.
     """
 
-    def create_backend(self, backend: str, initParams: dict[str, str] = {}):
+    def create_backend(
+        self, backend: str, initParams: Optional[dict[str, str]] = None
+    ) -> None:
+        if initParams is None:
+            initParams = {}
+        if backend in ("UCX", "OBJ") and "ucx_error_handling_mode" not in initParams:
+            initParams = initParams.copy()
+            initParams["ucx_error_handling_mode"] = self.ucx_error_handling_mode
         self.backends[backend] = self.agent.createBackend(backend, initParams)
 
         (backend_options, mem_types) = self.agent.getBackendParams(
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -817,7 +817,7 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
         num_workers = num_threads + 1;
     }
 
-    ucp_err_handling_mode_t err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+    ucp_err_handling_mode_t err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
     if (const auto opt = nixl::getBackendParamOptional<std::string>(
             custom_params, std::string(nixl_ucx_err_handling_param_name))) {
         err_handling_mode = ucx_err_mode_from_string(*opt);
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -36,7 +36,7 @@ get_ucx_backend_common_options() {
     nixl_b_params_t params = {{"ucx_devices", ""}, {"num_workers", "1"}};
 
     params.emplace(nixl_ucx_err_handling_param_name,
-                   ucx_err_mode_to_string(UCP_ERR_HANDLING_MODE_PEER));
+                   ucx_err_mode_to_string(UCP_ERR_HANDLING_MODE_NONE));
     return params;
 }
 
--- a/test/gtest/error_handling.cpp
+++ b/test/gtest/error_handling.cpp
@@ -23,8 +23,8 @@
 
 namespace gtest {
 namespace nixl {
-    constexpr const char* ucx_err_handling_mode_key  = "ucx_error_handling_mode";
-    constexpr const char* ucx_err_handling_mode_peer = "peer";
+    constexpr const char *ucx_err_handling_mode_key = "ucx_error_handling_mode";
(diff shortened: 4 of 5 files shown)


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27069930639/job/79897226353)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `fb5aff7f`

I have all the evidence needed. The diagnosis is complete and clear.

---

**Summary:** Clang Format Check failed on `test/gtest/error_handling.cpp` due to an inconsistent pointer declaration style — `const char*` (space before `*`) instead of `const char *` (space after `*`).

**Root cause:** The PR added a new constant on line 27 of `test/gtest/error_handling.cpp`:
```cpp
// Current (wrong — fails clang-format):
constexpr const char* ucx_err_handling_mode_none = "none";

// Required (clang-format style):
constexpr const char *ucx_err_handling_mode_none = "none";
```
`clang-format-19` enforces the project's `.clang-format` style, which places the `*` adjacent to the **variable name** (`char *var`), not the type (`char* var`). The new line introduced by commit `fb5aff7` uses the wrong spacing, while the already-existing line directly above it (`ucx_err_handling_mode_key`) has the same style bug — but since that line was not changed in this PR, only the new `ucx_err_handling_mode_none` line was flagged by `clang-format-diff` (which only checks the diff, not the whole file).

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch: `ucx-error-handling-none`, PR #1739)

**File:** `test/gtest/error_handling.cpp:27`

**Suggested fix:** Change the pointer spacing on the offending line:
```cpp
// Before:
constexpr const char* ucx_err_handling_mode_none = "none";

// After:
constexpr const char *ucx_err_handling_mode_none = "none";
```
Additionally, since the immediately preceding line (`ucx_err_handling_mode_key` on line 26 and `ucx_err_handling_mode_peer` on line 27) uses the same wrong style but wasn't caught (it pre-dates this PR's diff), those should be fixed too to avoid future noise:
```cpp
constexpr const char *ucx_err_handling_mode_key  = "ucx_error_handling_mode";
constexpr const char *ucx_err_handling_mode_peer = "peer";
constexpr const char *ucx_err_handling_mode_none = "none";
```
Alternatively, run `clang-format-19 -i test/gtest/error_handling.cpp` locally before pushing to auto-fix all formatting in the file.

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `ddf9d4f9-f71c-4531-963d-e950ba064146` in the triage console for the audit trail.

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27070129475/job/79897748746)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `3e41baf9`

The diagnosis is clear and complete. Here is the full report:

---

**Summary:** Clang format check failed on `test/gtest/error_handling.cpp` due to an extra trailing space in a variable declaration.

**Root cause:** A double-space alignment used before `=` on line 26 of `test/gtest/error_handling.cpp` violates the project's `.clang-format` rules. Specifically, the declaration:
```cpp
constexpr const char *ucx_err_handling_mode_key  = "ucx_error_handling_mode";
```
has two spaces before `=` (likely hand-aligned with the next declaration), whereas `clang-format-19` requires only one space:
```cpp
constexpr const char *ucx_err_handling_mode_key = "ucx_error_handling_mode";
```
The `clang-format-diff-19` tool detected this misalignment and exited with code 1, failing the CI job.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch: `ucx-error-handling-none`, PR #1739)

**File:** `test/gtest/error_handling.cpp:26`

**Suggested fix:** Remove the extra space before `=` on the offending line. Change:
```cpp
constexpr const char *ucx_err_handling_mode_key  = "ucx_error_handling_mode";
```
to:
```cpp
constexpr const char *ucx_err_handling_mode_key = "ucx_error_handling_mode";
```
You can also run `clang-format-19 -i -style=file test/gtest/error_handling.cpp` locally before pushing to auto-fix any remaining formatting issues in that file.

**Related:** PR #1739 (`ucx-error-handling-none`)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `abab1fc2-8658-427a-a678-94c4b46a4d18` in the triage console for the audit trail.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -815,7 +815,7 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     const auto err_handling_mode_it =
         custom_params->find(std::string(nixl_ucx_err_handling_param_name));
     if (err_handling_mode_it == custom_params->end()) {
-        err_handling_mode = UCP_ERR_HANDLING_MODE_PEER;
+        err_handling_mode = UCP_ERR_HANDLING_MODE_NONE;
```
iyastreb: Could you please describe:
- what overhead exactly do you reference?
- with this change, what exactly do we gain? Is it measurable?

I believe for RMA the overhead is exactly 0. Maybe only for DC and error path it can add a couple of CPU cycles

But it may change the lane selection, because actually NONE relaxes the constraints, so more transports may be eligible and more ifaces are opened.
Changing the default behaviour for all NIXL consumers is risky, and should be justified

src/api/python/_api.py
```diff
@@ -155,6 +158,7 @@ def __init__(
         capture_telemetry: bool = False,
         num_threads: int = 0,
         backends: list[str] = ["UCX"],
+        ucx_error_handling_mode: str = "none",
```
iyastreb: If we decide to go on with this change, the arg must be the last one
Otherwise it breaks the API

iyastreb (changes requested):

src/plugins/ucx/ucx_backend.cpp
mmangkad: @iyastreb good points. I should be more precise here. I am not claiming a measured RMA hot path speedup.

By overhead I meant the UCX documented behavior around `PEER`. UCX documents `NONE` as the default mode with minimal performance overhead. For `PEER`, UCX says it provides stronger remote failure guarantees, but may affect performance and memory footprint and disables protocols or APIs that could hang or behave incorrectly under peer failure.

I agree this can affect lane selection. That is actually part of why I think `PEER` should be explicit rather than the default. UCX has checks where peer failure mode adds stricter RNDV memory invalidation requirements, skips choices that are only allowed for `NONE`, and may add keepalive lane handling.

So the gain here is not “RMA is faster by X cycles”. The gain is that NIXL’s default becomes aligned with UCX’s endpoint default and with existing NIXL usage. The device EP path already sets `ucx_error_handling_mode = "none"`, and the UCX worker helper also defaults to `UCP_ERR_HANDLING_MODE_NONE`.

`peer` remains fully supported for users who want those peer failure guarantees. This PR just makes those stronger semantics opt in instead of imposing them on all UCX backend creation by default.

src/api/python/_api.py
mmangkad: Agreed, will move it to the end.

src/plugins/ucx/ucx_backend.cpp
iyastreb: Essentially this changes the default behaviour for all NIXL users (until they adapt their code), which is risky. Personally I'm not seeing much benefits, as performance wise it's the same. I'm also not sure how many NIXL users prefers NONE default. So benefits are questionable, but risk is real..

@brminich What do you think?

PR #1740, opened 2026-06-07
nixlbench: verify storage WRITE consistency via transport read-back

harshavardhana:
## What

`--check_consistency` verifies a storage WRITE by reading the written data back
and comparing it to the initiator's known fill byte. Today that read-back comes
from local storage: object backends download the object to a temporary file and
`pread` it; block/file backends `pread` the device or file directly.

This adds an alternative: read the data back **over the transport** into the
original buffers, and route the storage WRITE case through it under the same
`--check_consistency` flag.

## Why

For a backend whose transfer is a pure memory movement over the network, nothing
is written to a local file or device, so there is no local copy to read back —
the file path doesn't apply. And even where a local copy could be produced,
materializing the object on local storage just to validate it adds disk I/O that
the transfer under test never performs, and ends up checking that copy rather
than the path the data actually took.

Reading the data back over the same transport avoids both. It needs no local
storage, and it verifies the round-trip end to end (buffer → backend → buffer),
which is the path the benchmark is exercising. It also keeps a single
verification path instead of per-backend assumptions about where bytes land on
disk, so a memory-transport backend gets consistency checking without a
local-file or device special case.

## How

- New `xferBenchWorker::verifyWriteByReadback()` (default: unsupported). The NIXL
  worker implements it: fill the local buffers with a poison byte, issue a
  `NIXL_READ` of the same descriptors back into them, then compare to
  `XFERBENCH_INITIATOR_BUFFER_ELEMENT`. The poison ensures a read that moves
  nothing fails rather than passing by coincidence. It reuses
  `prepareTransferDescriptors` / `getRemoteSegType`, so OBJ/FILE/BLK remote
  segments and DRAM/VRAM local buffers are handled the same way as a normal
  transfer.
- `validateTransfer()` now takes the `worker` and routes the storage WRITE case
  to it. READ, target, and non-storage paths are unchanged.

This does change how a storage WRITE is verified (read-back instead of a
local-file/device read). Happy to scope it differently — e.g. gate it behind an
option — if that's preferred.

## Testing

Built NIXL + nixlbench from this branch. Ran `--check_consistency=1` on a storage
WRITE across a block-size sweep: a correct transfer passes, and an injected byte
mismatch is reported and fails the run.

## Summary by CodeRabbit

* **New Features**
  * Added an optional write-verification flow that performs a readback into provided local buffers and compares bytes against expected sentinels.
  * Introduced a worker/back-end hook so storage backends can enable readback-based consistency checks.
  * Implemented the verification for the NIXL backend, including mismatch detection with logged first differing byte and appropriate handling for VRAM-backed buffers.
  * If a backend doesn’t support the hook, write verification remains unsupported by default.

### DIFF
--- a/benchmark/nixlbench/src/main.cpp
+++ b/benchmark/nixlbench/src/main.cpp
@@ -127,7 +127,8 @@ static int processBatchSizes(xferBenchWorker &worker,
             worker.exchangeIOV(local_trans_lists, block_size);
             worker.poll(block_size);
 
-            if (!xferBenchUtils::validateTransfer(false, local_trans_lists, local_trans_lists)) {
+            if (!xferBenchUtils::validateTransfer(
+                    false, worker, local_trans_lists, local_trans_lists)) {
                 return EXIT_FAILURE;
             }
             if (IS_PAIRWISE_AND_SG()) {
@@ -144,7 +145,8 @@ static int processBatchSizes(xferBenchWorker &worker,
                 return 1;
             }
 
-            if (!xferBenchUtils::validateTransfer(true, local_trans_lists, remote_trans_lists)) {
+            if (!xferBenchUtils::validateTransfer(
+                    true, worker, local_trans_lists, remote_trans_lists)) {
                 return EXIT_FAILURE;
             }
 
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -39,6 +39,7 @@
 #include "runtime/etcd/etcd_rt.h"
 #include "utils/neuron.h"
 #include "utils/utils.h"
+#include "worker/worker.h"
 
 // Define command line parameters
 #define NB_ARG_STRING(param_name, def_val, help_text) DEFINE_string(param_name, def_val, help_text)
@@ -573,6 +574,14 @@ xferBenchConfig::loadParams(void) {
     }
 
     if (worker_type == XFERBENCH_WORKER_NVSHMEM) {
+        // The NVSHMEM worker does not implement verifyWriteByReadback(), so a
+        // storage backend (which routes WRITE consistency through it) would fall
+        // back to the base implementation and silently fail. Reject the combo.
+        if (isStorageBackend()) {
+            std::cerr << "Error: NVSHMEM worker does not support storage backends" << std::endl;
+            std::cerr << "Hint: use --worker_type nixl for storage backends" << std::endl;
+            return -1;
+        }
         if (!((XFERBENCH_SEG_TYPE_VRAM == initiator_seg_type) &&
               (XFERBENCH_SEG_TYPE_VRAM == target_seg_type) && (1 == num_threads) &&
               (1 == num_initiator_dev) && (1 == num_target_dev) &&
@@ -1131,6 +1140,7 @@ xferBenchUtils::checkConsistency(std::vector<std::vector<xferBenchIOV>> &iov_lis
 
 bool
 xferBenchUtils::validateTransfer(bool is_initiator,
+                                 xferBenchWorker &worker,
                                  std::vector<std::vector<xferBenchIOV>> &local_lists,
                                  std::vector<std::vector<xferBenchIOV>> &remote_lists) {
     if (!xferBenchConfig::check_consistency) {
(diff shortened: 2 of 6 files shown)


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27097954278/job/79973409067)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `d79f3187`

I have enough information to provide a complete and confident diagnosis. Here is the full report:

---

**Summary:** Clang Format Check failed on PR #1740 (`pr/consistency-readback`) — three code locations in the benchmark files exceed the column width limit and need line-wrapping reformatted to match `clang-format-19` style.

**Root cause:** Commit `[REDACTED:Hex High Entropy String]` introduced new code in `benchmark/nixlbench/src/main.cpp` and `benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp` that violates the project's clang-format style rules. Specifically, `clang-format-19` reports three sets of diffs that need correction:

1. **`benchmark/nixlbench/src/main.cpp` lines ~127 and ~144** — Two calls to `xferBenchUtils::validateTransfer(...)` exceed the column limit and need their arguments split/indented across multiple lines per `clang-format` style:
   ```cpp
   // Current (violates style):
   if (!xferBenchUtils::validateTransfer(false, worker, local_trans_lists, local_trans_lists)) {

   // Required:
   if (!xferBenchUtils::validateTransfer(
           false, worker, local_trans_lists, local_trans_lists)) {
   ```

2. **`benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp` around line 1420** — Three sub-violations:
   - The `verifyWriteByReadback` function definition's parameter list must be reformatted from a 3-line split to a 2-line aligned split.
   - The `for` loop condition (`list_idx < local_lists.size() && list_idx < remote_lists.size()`) must be collapsed from 3 lines to 2 lines.
   - The `CHECK_CUDA_ERROR(cudaMemset(...), ...)` call must be collapsed from 3 lines to 2 lines.

The check ran `clang-format-diff-19 -p1 -style=file` against the diff of the PR merge commit and exited with code 1 because the output was non-empty (i.e., formatting differences were found).

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR author on branch `pr/consistency-readback`)

**File:**
- `benchmark/nixlbench/src/main.cpp` lines ~127, ~144
- `benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp` lines ~1420–1437

**Suggested fix:** Run `clang-format-19` locally on the modified files and commit the results before pushing. The quickest approach:
```bash
# From the repo root — reformat only the files changed in this PR
clang-format-19 -i -style=file \
  benchmark/nixlbench/src/main.cpp \
  benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp

git add benchmark/nixlbench/src/main.cpp \
        benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
git commit -m "style: apply clang-format-19 to benchmark files"
git push
```
Alternatively, apply the exact diff the CI produced (the `--- before / +++ after` hunks shown in the log) as a patch directly.

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `96d93a59-ff06-4b97-b0e8-f51703ee2737` in the triage console for the audit trail.

harshavardhana:
Thanks for the review. Addressed in the latest push:

- **Descriptor counts**: fail fast on a list-count or per-list size mismatch rather than iterating the shorter list.
- **VRAM**: device memory is no longer dereferenced as host — the poison fill and read-back go through helpers (`fillTransferBuffer`/`hostView`) that copy via CUDA/ROCm/Neuron (mirroring `copyVramToHost`) and set the owning device (`cudaSetDevice`/`hipSetDevice`) first.
- **clang-format**: the two earlier formatting findings were from the initial commit and are now resolved.

Rebuilt NIXL + nixlbench locally to confirm it compiles.

harshavardhana:
Rebased onto latest main (no conflicts) and addressed the read-back review comment (bounded wait + preserve terminal status so a transport error isn't misreported as a data mismatch). @aranadive @vvenkates27 — could you trigger `/ok to test` + `/build` and review when you have a moment? Thanks!

harshavardhana:
@aranadive @vvenkates27 — this one's ready for review: all CodeRabbit findings are addressed and the light CI is green. Could you take a review pass and kick off `/ok to test` + `/build` when you get a chance? Thanks!

PR #1741, opened 2026-06-07
obj: add standard-protocol GPU-direct S3-over-RDMA (accelerated=true)

harshavardhana:
## What

Adds a vendor-neutral S3-over-RDMA data path to the object plugin, selected via `accelerated=true` with no `type` (or `type=s3`). It implements the published `x-amz-rdma-*` protocol (the NVIDIA cuObject / aws-c-s3 convention) and needs no per-vendor code — any conformant endpoint works (MinIO AIStor today; AWS S3 if it adopts it). The control plane stays standard HTTP + SigV4 while the object payload moves out-of-band over RDMA (server `RDMA_READ` on PUT, `RDMA_WRITE` on GET; the HTTP body is empty).

This is the preferred GPU-direct path. `accelerated=true, type=dell` selects a vendor-specific engine for servers that speak Dell's `x-rdma-info` header.

## Changes

- **`s3/rdma_protocol.h`** — dependency-free wire helpers (token formatting, reply parsing), covered by a unit test.
- **`s3/rdma.{h,cpp}`** — the cuObject RDMA token API plus a SigV4 `UNSIGNED-PAYLOAD` control plane that carries `x-amz-rdma-token` and reads `x-amz-rdma-reply` / `x-amz-rdma-bytes-transferred`.
- **`s3_accel/generic/`** — the standard-S3 engine, registered in the accel registry under `type=s3`; `accelerated=true` with no `type` resolves to it. RDMA is a capability of the standard S3 client, enabled for this engine.
- **Explicit opt-in** — an RDMA decline/failure is a hard error; no automatic HTTP fallback.
- **Docs** — `RDMA_PROTOCOL.md` documents the server-side wire contract and a vendor compliance checklist; the README gains a GPU-Direct section.
- **`nixl_object_test`** — gains `-a/--accelerated` (standard-S3); keeps `-T/--type` for vendor engines.

## Verification

Validated end-to-end against MinIO AIStor over RoCEv2: `accelerated=true` (no type) and `type=s3` both complete host-DRAM and GPU-direct VRAM PUT/GET with data verification. The protocol layer is covered by a unit test.

> GPU-direct into VRAM requires the GPU and NIC to share a PCIe path that permits peer DMA (see the README's PCIe-ACS note) — a host/fabric property, independent of NIXL.

### DIFF
--- a/src/plugins/obj/RDMA_PROTOCOL.md
+++ b/src/plugins/obj/RDMA_PROTOCOL.md
@@ -1,0 +1,152 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# S3-over-RDMA Wire Protocol
+
+This document specifies the GPU-direct S3-over-RDMA protocol the object plugin
+speaks, so that **any S3-compatible storage vendor can comply on the server side
+without NIXL-specific code**. The control plane stays standard HTTP + SigV4,
+while the object payload moves out-of-band over RDMA. A server that implements
+the contract below works with this client when the caller opts in via
+`accelerated=true` (the standard-S3 engine; no `type`, or `type=s3`).
+
+**Provenance.** The `x-amz-rdma-*` header names and the token layout come
+directly from NVIDIA's published [cuObject documentation][cuobj] (the
+client-side API and the RDMA-enabled GET/PUT workflow). The wire protocol was
+written up as a spec ([aws-c-s3 `RDMA_PROTOCOL_SPEC.md`][spec]) and proposed to
+AWS. It is not yet GA'ed by AWS; however AWS refined it inline with what they
+could eventually implement, so it is a stable, vendor-neutral base to
+standardize on rather than an endpoint-specific convention.
+
+[cuobj]: https://docs.nvidia.com/gpudirect-storage/cuobject/index.html
+[spec]: https://github.com/KiranModukuri/aws-c-s3/blob/nvidia_rdma/RDMA_PROTOCOL_SPEC.md
+
+The standard-S3 engine is the **preferred** GPU-direct path: it follows this
+proposed protocol and needs no per-vendor code, so it works against any
+conformant server. Vendor `type=...` engines exist only for servers that speak a
+vendor's own RDMA headers.
+
+> The client treats RDMA as an explicit opt-in and does **not** auto-fall-back to
+> HTTP today (see the README's "No automatic fallback (yet)"). A compliant server
+> signals an unsupported request with `x-amz-rdma-reply: 501`, but a
+> _non_-compliant server may silently ignore the token and accept a body-less PUT
+> as a 0-byte object — so auto-fallback becomes safe only once the `501` contract
+> below is universal.
+
+The reference server implementation is MinIO AIStor (`internal/rdma`, repo at
+`../aistor`).
+
+## Transport
+
+- RDMA **Dynamically Connected (DC)** transport over **InfiniBand** or
+  **RoCEv2** (cuObject protocol `CUOBJ_PROTO_RDMA_DC_V1`).
+- **GET**: the server issues `RDMA_WRITE`, pushing object bytes directly into the
+  client's registered buffer (host DRAM or GPU VRAM).
+- **PUT** / **UploadPart**: the server issues `RDMA_READ`, pulling bytes directly
+  out of the client's registered buffer.
+- The HTTP request/response body is **empty**; payload never traverses it.
+
+## Headers
+
+| Header                                   | Direction        | Meaning                                                                                                                                                                                                                               |
+| ---------------------------------------- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `x-amz-rdma-token`                       | request          | RDMA negotiation token (see format below). Its presence is what marks a request as RDMA.                                                                                                                                              |
+| `x-amz-content-sha256: UNSIGNED-PAYLOAD` | request          | Body is not signed (there is no body).                                                                                                                                                                                                |
+| `Content-Length: 0`                      | request (PUT)    | No HTTP body; data is delivered via `RDMA_READ`.                                                                                                                                                                                      |
+| `Range: bytes=a-b`                       | request (GET)    | Optional; selects a byte range (server replies `206`).                                                                                                                                                                                |
+| `x-amz-checksum-crc64nvme`               | request/response | Optional CRC64NVME checksum (PUT/UploadPart).                                                                                                                                                                                         |
+| `x-amz-rdma-reply`                       | response         | Outcome marker. **Required on success.** `200`/`204`/`206` = success; `501` (or absent) = server declined RDMA. Under `accelerated=true` the client treats a decline as a hard error (no automatic HTTP fallback today — see README). |
+| `x-amz-rdma-bytes-transferred`           | response (GET)   | Actual bytes moved over RDMA (may be `< requested` for ranged GETs).                                                                                                                                                                  |
+| `ETag`, `x-amz-checksum-crc64nvme`       | response (PUT)   | Standard S3 result metadata.                                                                                                                                                                                                          |
+
+### Token format
+
+```text
+x-amz-rdma-token: <descriptor>:<start_addr_hex>:<size_hex>
+```
+
+- `<descriptor>` — the opaque, fixed-length RDMA descriptor minted by the
+  client's RDMA provider (cuObject `cuMemObjGetRDMAToken`); it encodes the
+  registered buffer's base address, max size, remote key and NIC routing
+  (GID/LID/DCTN). The server splits it off (it is a known fixed length) and
+  forwards exactly those bytes to its own RDMA provider (cuObjServer); it must
+  not interpret them.
+- `<start_addr_hex>` — client buffer start address for this operation, 16-digit
+  lowercase hex.
+- `<size_hex>` — transfer size in bytes, 16-digit lowercase hex.
+
+Appending the per-operation start address and size to the descriptor mirrors
+cuObject's own **IO descriptor** layout (its reference server reads
+`<descriptor><rem_buf_start>…<rem_msg_size>…`); it is not an endpoint-specific
+addition. Because the HTTP body is empty (`Content-Length: 0`), the token is the
+only place these per-request fields can travel. The two trailing fields are
+`:`-separated and base-16; the server splits them off the end and treats the
+remaining prefix as the descriptor. A trailing `;` (some clients append one)
+must be tolerated.
+
+## Request flow (server obligations)
+
+1. **Authenticate** the request with standard SigV4. `x-amz-rdma-token` is part
+   of the signed headers; `x-amz-content-sha256` is `UNSIGNED-PAYLOAD`.
+2. **Detect** an RDMA request by the presence of `x-amz-rdma-token`; parse it
+   into `(descriptor, start_addr, size)`.
+3. **Decide** whether RDMA can be honored (fabric reachable, buffer registrable,
+   object/permissions valid).
+   - **If not** → respond `x-amz-rdma-reply: 501`. You MAY include a normal HTTP
+     error body (do not force `Content-Length: 0`). The `501` signal is what lets
+     a client safely decide what to do; today this client treats it as an error
+     under `accelerated=true` (it does not auto-retry over HTTP — see README), but the
+     signal is required so future clients can fall back transparently.
+4. **Transfer** using the cuObject server APIs (register a local buffer, then):
+   - GET / RANGE_GET → `RDMA_WRITE` the requested bytes into the client buffer.
+   - PUT / UploadPart → `RDMA_READ` the bytes from the client buffer and persist
+     them as the object/part.
+5. **Respond** on success:
+   - `x-amz-rdma-reply: 200` (full GET/PUT), `204` (PUT, no content), or `206`
+     (ranged GET partial).
+   - `x-amz-rdma-bytes-transferred: <n>` for GET.
+   - `Content-Length: 0` and **no HTTP body** (the data went over RDMA — without
+     this the client blocks waiting for body bytes that never arrive).
+   - `ETag` (and `x-amz-checksum-crc64nvme` if requested) for PUT.
+
+## Completion semantics
+
+`x-amz-rdma-reply` is **not** a strict requirement on the PUT success path — it
+is an outcome marker the server is free to use:
+
+- **PUT success:** the server completes the `RDMA_READ` and returns a standard
+  HTTP `200` + `ETag` (the body is empty; data moved out-of-band). The client
+  treats `200` + a non-empty `ETag` as success. AIStor does not set
+  `x-amz-rdma-reply` on this path, and a client must not require it.
+- **GET success:** the server returns `x-amz-rdma-reply: 200/206` and
+  `x-amz-rdma-bytes-transferred` (the authoritative moved-byte count); the HTTP
+  body is empty.
+- **Decline:** `x-amz-rdma-reply: 501` (or its absence on a non-RDMA server)
+  signals the request was not honored over RDMA.
+
+The data path's integrity is covered by RoCEv2 iCRC and, when requested, the
+`x-amz-checksum-crc64nvme` header — not by the (empty) HTTP-body content hash.
+
+## Multipart uploads
+
+UploadPart uses the same body-less, token-carrying PUT with the standard
+`?uploadId=<id>&partNumber=<n>` query parameters (`1 ≤ n ≤ 10000`). The part's
+`ETag`/checksum are returned as usual; CompleteMultipartUpload is an ordinary
+HTTP call.
+
+## Compliance checklist
+
+- [ ] Parse `x-amz-rdma-token` as `<descriptor>:<addr16>:<size16>` (tolerate a
+      trailing `;`).
+- [ ] Authenticate with SigV4 treating the body as `UNSIGNED-PAYLOAD`.
+- [ ] GET ⇒ `RDMA_WRITE` into the client buffer; PUT ⇒ `RDMA_READ` from it.
+- [ ] Always set `x-amz-rdma-reply` (`200`/`204`/`206`) on success, `501` to decline.
+- [ ] Set `x-amz-rdma-bytes-transferred` on GET; force `Content-Length: 0` with
+      no body on every success.
+- [ ] Return `ETag`/checksum for PUT; honor `Range` for GET (`206`).
+- [ ] Fall through to normal HTTP semantics for non-RDMA requests.
+
+A vendor that satisfies this checklist is supported by the NIXL object plugin's
+default S3 client automatically — no NIXL-side engine or plugin is required.
--- a/src/plugins/obj/s3/engine_impl.h
+++ b/src/plugins/obj/s3/engine_impl.h
@@ -7,6 +7,7 @@
 #define OBJ_PLUGIN_S3_ENGINE_IMPL_H
 
 #include "obj_backend.h"
+#include "rdma.h"
 
 class DefaultObjEngineImpl : public nixlObjEngineImpl {
 public:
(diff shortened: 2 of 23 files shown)


dpressle:
👀 Investigating [`Run Pre-Commit Hooks`](https://github.com/ai-dynamo/nixl/actions/runs/27107087273/job/79998085467)…

dpressle:
🤖 **CI Triage Agent** — `Run Pre-Commit Hooks` · commit `2aefadc1`

The file is new (introduced in this PR). The log has given us everything we need. No further investigation required.

---

**Summary:** `pre-commit` codespell hook fails due to a misspelling in `src/plugins/obj/s3/rdma_protocol.h` — the word `unparseable` should be `unparsable`.

**Root cause:** The newly introduced file `src/plugins/obj/s3/rdma_protocol.h` (added on branch `feat/obj-s3-over-rdma-gds`) contains the misspelling `unparseable` at **line 76**. The `codespell` pre-commit hook (exit code 65) caught it and flagged it as a spelling error:
```
src/plugins/obj/s3/rdma_protocol.h:76: unparseable ==> unparsable
```
All other hooks (case conflicts, merge conflicts, end-of-files, trailing whitespace, line endings) passed cleanly. The Python linting hooks (mypy, isort, black, flake8) were skipped as there are no Python files in the changeset.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1741, branch `feat/obj-s3-over-rdma-gds`)

**File:** `src/plugins/obj/s3/rdma_protocol.h:76`

**Suggested fix:** In `src/plugins/obj/s3/rdma_protocol.h` at line 76, replace the misspelled word:
```diff
-  unparseable
+  unparsable
```
Then commit and push the fix. Alternatively, run `pre-commit run codespell --files src/plugins/obj/s3/rdma_protocol.h` locally to auto-verify the correction before pushing.

**Related:** None found — this is a straightforward typo in a new file introduced by PR #1741.



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `661b6714-f55b-4162-88dc-f530d2c5ba84` in the triage console for the audit trail.

harshavardhana:
Thanks for the review. Addressed the actionable findings:

**Fixed**
- **Critical — VRAM readiness mismatch** (`engine_impl.cpp` / `client.cpp`): added a single `iS3Client::supportsRdma()` predicate (`gds` requested **and** cuObject fabric + control plane + executor all ready). The engine now advertises/registers `VRAM_SEG` only when that holds, and the client fails fast (no silent HTTP) when `gds=true` but the fast path isn't ready — so a GPU pointer can never reach `PreallocatedStreamBuf`.
- **Header guards** (`rdma.h`, `rdma_protocol.h`): switched to `NIXL_SRC_PLUGINS_OBJ_...` per CodeStyle.md.
- **constexpr naming**: renamed `k*` constants to `snake_case` per CodeStyle.md line 29.
- **Docs**: fixed the `x-amz-rdma-reply` table row + server-obligation step to reflect *no auto-fallback today* (hard error under `gds=true`); added `text` language tag to the token-format fence; reworded the README "negotiates with proper fallback" claim. Also de-staled the `rdma.h`/`rdma.cpp` comments.
- **Spelling** (`unparseable`→`unparsable`): already fixed in a prior push.

**Skipped (with reason)**
- **`s3_crt/engine_impl.h` header guard** — pre-existing guard, not introduced by this PR; out of scope for this change.
- **`nixl_object_test.cpp` continuation indent** — `clang-format` (the enforced CI gate) produces and passes the current alignment; changing it to base+4 would fail clang-format.
- **NIC-failover retry** (`rdma.cpp`) — the retry re-mints the token but the control-plane HTTP isn't yet bound to the token's NIC source address (AWS SDK doesn't expose per-request interface binding portably). Rather than imply failover works, I corrected the comments to call it a transient retry and documented NIC pinning as a follow-up.

harshavardhana:
Second review pass — all 9 open threads addressed (commit e0cd444) and resolved:

- **`parseRdmaReply` trailing-junk** (`rdma_protocol.h`): switched from `std::stoi` to `std::from_chars` requiring the *entire* value to parse, so `"200xyz"`/`"501x"` are now rejected (→ failure) instead of masquerading as success. Added test cases.
- **Fail fast on `gds=true` without a ready fast path** (`client.cpp`): the constructor now throws when `gds=true` but RDMA isn't fully usable (no cuObjClient build, no executor, fabric down, or invalid control plane) — no more construct-then-hard-fail-every-op.
- **Retry token-mint failures** (`rdma.cpp`): a `nullptr` from `getToken()` now `continue`s the retry loop instead of aborting on the first attempt.
- **NIC steering / dead code** (`rdma.cpp` + `rdma_protocol.h`): removed the unused `parseClientNicFromToken` and all NIC-failover wording. The AWS SDK HTTP client can't bind per-NIC, so the retry is documented honestly as a transient retry; NIC pinning is noted as a future enhancement (no misleading claims, no dead code).
- **Public-before-protected** (`client.h`): reordered so the section order is `public` → `protected` per CodeStyle.md.
- **`obj_key` by const ref** (`engine_impl.cpp`).
- **MD028** (`README.md`): joined the adjacent blockquotes.
- **Stale member rename** (`engine_impl.h`): the flagged `gds_enabled_` member was removed entirely (now gated on the client's `supportsRdma()`); thread was outdated.

Protocol unit tests (including the new trailing-junk cases) pass.

harshavardhana:
@aranadive @vvenkates27 — reworked per our discussion:

- Dropped the separate `gds` flag. The standard-protocol S3-over-RDMA path is now selected by `accelerated=true` with **no `type`** (or `type=s3`); `accelerated=true, type=dell` selects the Dell vendor engine.
- The standard-S3 engine uses the standard `x-amz-rdma-*` headers and needs no per-vendor code.

Validated end-to-end against MinIO AIStor over RoCEv2 — `accelerated=true` (no type) and `type=s3`, host DRAM and GPU-direct VRAM PUT/GET with data verification.

Ready for review — could you `/ok to test` + `/build` when you get a chance? Thanks!

src/plugins/obj/s3/rdma.cpp
```diff
@@ -0,0 +1,505 @@
...
+        // accelerated=true).
+        const std::string reply = resp->HasHeader(amz_rdma_reply) ?
+            std::string(resp->GetHeader(amz_rdma_reply).c_str()) :
+            "";
+        const int reply_code = parseRdmaReply(reply);
+        if (reply_code == static_cast<int>(rdma_not_supported)) {
+            return rdma_not_supported;
+        }
+        if (reply_code != rdma_reply_success && reply_code != rdma_reply_partial_content) {
+            return -1;
```
ofer: print the reply_code then return?

src/plugins/obj/s3/rdma.cpp
```diff
@@ -0,0 +1,505 @@
...
+        if (!ctx.checksum_crc64nvme.empty()) {
+            req->SetHeaderValue("x-amz-checksum-crc64nvme", ctx.checksum_crc64nvme.c_str());
+        }
+
+        impl_->signV4(*req); // manual SigV4 with UNSIGNED-PAYLOAD
+
+        auto resp = impl_->http->MakeRequest(req);
+        if (!resp) {
+            NIXL_ERROR << "rdmaPut: MakeRequest returned null for key=" << ctx.object;
+            return -1;
```
ofer: -1 should be rdma_unknown_error or something else that isn't a magic number

harshavardhana:
Thanks @ofer — your comments have been addressed: the `-1` returns are now a named `rdma_error` constant, and the failure paths log `reply_code` before returning.

src/plugins/obj/RDMA_PROTOCOL.md
```diff
@@ -0,0 +1,133 @@
...
+| `x-amz-rdma-reply`                       | response         | Outcome marker. **Required on success.** `200`/`204`/`206` = success; `501` (or absent) = server declined RDMA. Under `accelerated=true` the client treats a decline as a hard error (no automatic HTTP fallback today — see README). |
+| `x-amz-rdma-bytes-transferred`           | response (GET)   | Actual bytes moved over RDMA (may be `< requested` for ranged GETs).                                                                                                                                                                  |
+| `ETag`, `x-amz-checksum-crc64nvme`       | response (PUT)   | Standard S3 result metadata.                                                                                                                                                                                                          |
+
+### Token format
+
+```text
+x-amz-rdma-token: <descriptor>:<start_addr_hex>:<size_hex>
```
jgoldsch12, line 67: This changes the cuobject descriptor accepted by the cuobjclient and cuobjserver and adds vendor endpoint specific fields which are not part of the cuobject libaries.  If we are looking to have a generic implementation, it should not include additional fields used by one vendor.  Otherwise, this is what the "type" designation is for.

jgoldsch12 (commented):
Hi @harshavardhana Thanks for putting this PR together.  Having a generic implementation which is cross vendor will be very helpful.  I imagine other vendors may want to test with this to ensure that it is truly generic and not designed to work with only one specific endpoint.  Given that there is no open wire protocol specification, implementors have had little to work with and thus chances of interoperability are low.  Thanks for putting this stake in the ground.  I'm interested in seeing this become part of NIXL.

src/plugins/obj/s3_accel/dell/README.md
```diff
@@ -17,36 +17,45 @@ limitations under the License.
 
 # NIXL Dell ObjectScale accelerated S3 engine
 
-This vendor-specific accelerated engine provides S3 over RDMA for Dell ObjectScale.  This engine utilizes the CUDA Toolkit CUObject Client library and the AWS S3 SDK.
+> **Note:** This Dell vendor engine (selected via `accelerated=true, type=dell`)
+> is a vendor-specific engine for servers that use vendor-specific
+> RDMA headers: it uses the `x-rdma-info` header and selects its engine
+> statically at backend creation. The standard-S3 engine
+> (`accelerated=true` with no `type`, or `type=s3`) is preferred because it
+> complies with the agreed S3-over-RDMA protocol, using the standard
```
jgoldsch12: "agreed", I think proposed is probably better?  I haven't seen anything that is part of the official documentation.

src/plugins/obj/s3_accel/generic/engine_impl.cpp
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "engine_impl.h"
+#include "obj_engine_registry.h"
+
+namespace {
+
+// Register the standard-S3 engine under "s3" (the preferred, protocol-compliant
```
jgoldsch12, line 11: "preferred" may be better as "proposed"

src/plugins/obj/s3/engine_impl.h
```diff
@@ -16,8 +17,16 @@ class DefaultObjEngineImpl : public nixlObjEngineImpl {
...
     ~DefaultObjEngineImpl() override;
 
+    // VRAM_SEG (GPU-direct) is advertised only when the standard client's RDMA
+    // fast path is fully ready (accelerated=true generic S3-over-RDMA + cuObject
+    // fabric + control plane + executor). This is the same predicate the client
+    // uses to take the RDMA branch, so a GPU pointer can never reach the HTTP
+    // path.
     nixl_mem_list_t
     getSupportedMems() const override {
+        if (s3Client_ && s3Client_->supportsRdma()) {
```
jgoldsch12, line 27: This is a helpful change.  I tested this for regression with the ObjectScale endpoint and found no issue.

src/plugins/obj/s3/rdma.cpp
```diff
@@ -0,0 +1,512 @@
...
+bool
+SharedCuObjClient::isDeviceMemory(const void *ptr) const {
+    return cuObjClient::getMemoryType(ptr) == CUOBJ_MEMORY_CUDA_DEVICE;
+}
+
+char *
+SharedCuObjClient::getToken(void *ptr, size_t size, size_t offset, cuObjOpType_t op) {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    char *token = nullptr;
+    cuObjErr_t rc = client_->cuMemObjGetRDMAToken(ptr, size, offset, op, &token);
```
jgoldsch12, line 106: This pattern is definitely preferred.

src/plugins/obj/s3/rdma.cpp
```diff
@@ -0,0 +1,512 @@
...
+            uri.AddQueryStringParameter("uploadId", ctx.upload_id.c_str());
+            uri.AddQueryStringParameter("partNumber", std::to_string(ctx.part_number).c_str());
+        }
+
+        auto req =
+            Aws::Http::CreateHttpRequest(uri,
+                                         Aws::Http::HttpMethod::HTTP_PUT,
+                                         Aws::Utils::Stream::DefaultResponseStreamFactoryMethod);
+        req->SetHeaderValue("x-amz-content-sha256", unsigned_payload);
+        req->SetHeaderValue(amz_rdma_token, formatRdmaToken(token, buf_addr, size).c_str());
```
jgoldsch12, line 335: This isn't passing the token, as cuobjserver expects.  This is padding endpoint specific information to the descriptor string.  These values should be moved to a different header for compatibility.

harshavardhana: Thanks for the careful read.

The header names and token layout aren't AIStor inventions — they come straight from NVIDIA's published cuObject material, and the wire protocol around them was written up as a spec: [KiranModukuri/aws-c-s3 `RDMA_PROTOCOL_SPEC.md`](https://github.com/KiranModukuri/aws-c-s3/blob/nvidia_rdma/RDMA_PROTOCOL_SPEC.md). This is not yet GA'ed by AWS, however AWS refined this inline with what they could eventually implement — which is why it's a reasonable vendor-neutral base, and why pinning one concrete format down here (so other vendors can test against it) is useful.

- **Headers**: `x-amz-rdma-token` / `x-amz-rdma-reply` / `x-amz-rdma-bytes-transferred` are taken directly from the [cuObject docs](https://docs.nvidia.com/gpudirect-storage/cuobject/index.html) (the client-side API + the RDMA-enabled GET/PUT workflow), and that spec defines `x-amz-rdma-token` as an **opaque, provider-specific** identifier that "encodes buffer location, size, … and context."
- **The `:start_addr:size` suffix is cuObject's own convention, not endpoint padding.** cuObject's RDMA descriptor (`RDMA_DESC_STR`) is the fixed-length opaque handle — base address, buffer max-size, rkey, LID, DCTN, GID. cuObject then forms its **IO descriptor** by appending the per-operation transfer parameters to that descriptor: NVIDIA's own reference server (`sock_server.cc`) reads `…<RDMA_DESC_STR><REM_BUF_START><LOCAL_BUF_OFFSET><REM_MSG_SIZE><LOCAL_FILE_OFFSET>`, and `cuMemObjGetRDMAToken(ptr, size, buffer_offset, op, …)` takes the size + offset for exactly this. The token here carries the two essential fields (start addr + size).
- The server **splits off** the fixed-length descriptor and forwards only those bytes to its RDMA provider — cuObjServer never sees the suffix — and consumes addr/size itself to set up the transfer. Since the HTTP body is empty (`Content-Length: 0`), the token is the only place to convey the request's start offset and size; moving them to a separate header would diverge from the proposed spec and from cuObject's own IO-descriptor layout.

Keen to keep refining as the spec firms up and other vendors test against it.

src/plugins/obj/s3_accel/dell/README.md
harshavardhana: Good point — changed to "proposed" here and in RDMA_PROTOCOL.md.

src/plugins/obj/s3_accel/generic/engine_impl.cpp
harshavardhana: Protocol wording is now "proposed". "preferred" here just refers to the recommended engine to use, not the protocol's status.

src/plugins/obj/s3/rdma.cpp
jgoldsch12: Thank you for the detailed response.  Two things of notes:  1.  given we are talking about a document in private fork of a repo, it is further reason we should not refer to this as "agreed upon" or "preferred" or "standard".  I appreciate the changes you made to the comments and text.  2.  The x-amz-rdma-* header names are well understood from the documentation and from the proposed wire protocol spec from Kiran.  However, neither specify the format of the descriptor.  The implementation of aws-c-s3 fork does, however, pass the V1 descriptor produced by the cuobject library in x-amz-rdma-token header with only the 7 colon delimited fields, not 9.  I've found no reference to the additional two fields.  And while you make a point about the interface accepting additional fields, it does not change the token produced by the interface.  Adding the two fields is a reasonable implementation decision for an endpoint developer, but I'm missing where the actual header value is documented to include these extra parameters.  This is why I refer to this as being an endpoint specific implementation decision, and not an interoperable use of the proposed wire protocol.  It is very possible I'm missing something here.  I do know that ObjectScale interoperates with the aws-c-s3 fork, so this I why I'm raising this as a concern and a potential issue for NIXL interoperability.

harshavardhana:
Rebased onto latest main (no conflicts) and all review comments are addressed; the lightweight CI checks are green. @aranadive @vvenkates27 — could you kick off `/ok to test` + `/build` and take a review pass when you get a chance? Thanks!

aranadive:
@harshavardhana thanks for submitting the PR. We had further discussions internally on this and a also few other S3 PRs for alignment. We think that splitting this into two PRs would be beneficial for review and making progress for other PRs. It also looks like there is a lot of http header / cuobject-related code that should not included directly in the plugin and it would be helpful to have a refactor of that. Here is a more detailed review / recommended split for this PR:

## Architectural Feedback

RDMA awareness is wired into the base layer (`awsS3Client`, `DefaultObjEngineImpl`, `iS3Client`). Per the PR 1247 layering, it should live entirely in the accel layer (`awsS3AccelClient`, `S3AccelObjEngineImpl`). Specifically:

- **`awsS3Client`** should not have RDMA members, constructors, or dispatch branches — it's the plain HTTP client.
- **`iS3Client`** should not have `supportsRdma()` — that's an accelerator concern on the universal interface.
- **`DefaultObjEngineImpl`** should not have `#ifdef HAVE_CUOBJ_CLIENT` guards or dynamic `VRAM_SEG` — it forces `S3CrtObjEngineImpl` to add a defensive override.
- **`GenericObjEngineImpl` is unnecessary** — `S3AccelObjEngineImpl` *is* the generic standard-protocol engine and should register under `"s3"` directly.
- **~370 lines of reusable infrastructure** (`SharedCuObjClient`, `S3RdmaControlPlane`, `signV4`) are embedded in the S3 plugin but have no plugin-specific dependency. These should be shared utilities so vendor engines can also compose them.

## Minor Issues

1. `formatRdmaToken` uses `%lx` instead of `PRIx64` — not portable.
2. `S3RdmaControlPlane` resolves credentials once at construction — will fail for long-lived backends with rotating IAM session tokens. Document as known limitation.
3. `rdmaGet` with `size==0` would register a 0-byte RDMA region. Add a defensive guard.

## Recommended Split

### PR A: RDMA Transport and AWS Utilities (land first, no plugin changes)

Extract reusable infrastructure into shared utilities. These become the building blocks that both
`S3AccelObjEngineImpl` (the generic standard-protocol engine) and vendor engines (e.g. Dell) can
compose to build their RDMA implementations.

1. Move `SharedCuObjClient` (~72 lines) to `src/utils/rdma/` or `src/utils/cuobj/`. Process-wide RDMA plumbing with no S3 dependency.
2. Move `S3RdmaControlPlane` (~247 lines — SigV4, URI construction, HTTP dispatch, `rdmaPut`, `rdmaGet`) to `src/utils/object/` or `src/utils/aws/`.
3. Simplify `signV4` (~88 lines) or document why the manual implementation is required.
4. Fix `formatRdmaToken` to use `PRIx64`.
5. Move `rdma_protocol.h` (~109 lines) to `src/utils/object/`.
6. Move retry wrappers (`rdmaPutWithRetry`/`rdmaGetWithRetry`, ~51 lines) alongside the control plane.
7. Include `rdma_protocol.cpp` unit tests (~80 lines).
8. Document credential expiry as a known limitation.

### PR B: Wire RDMA into the Accel Layer (depends on PR A)

Integrate the utilities through `awsS3AccelClient` / `S3AccelObjEngineImpl` with zero changes to
base classes. `S3AccelObjEngineImpl` is the generic standard-protocol engine — there is no need
for a separate `GenericObjEngineImpl`.

1. Move RDMA ownership into `awsS3AccelClient` (`s3_accel/client.*`). Construct `SharedCuObjClient` and `S3RdmaControlPlane`, override `putObjectAsync`/`getObjectAsync` for RDMA. Dell overrides header format only. No new client subclass needed.
2. Move RDMA buffer pin/unpin into `S3AccelObjEngineImpl` (`s3_accel/engine_impl.*`). Override `registerMem`/`deregisterMem`. Advertise `VRAM_SEG` in `getSupportedMems()`.
3. Add `VRAM_SEG` validation in `S3AccelObjEngineImpl::prepXfer`, not in `isValidPrepXferParams`.
4. Remove `GenericObjEngineImpl`. Register `S3AccelObjEngineImpl` under `"s3"` directly. `obj_backend.cpp` normalizes missing `type` to `"s3"`.
5. Shouldnt be a need to modify `awsS3Client`, `DefaultObjEngineImpl`, `iS3Client`, or `S3CrtObjEngineImpl`.
6. Update documentation and tests to reflect accel-layer placement.

harshavardhana:
Converting to draft — splitting this per the review. The shared S3-over-RDMA utilities are in #1840; the accel-layer wiring will follow here on top of it once #1840 is merged.

harshavardhana:
Superseded by #2241. Split per the review: the shared S3-over-RDMA utilities landed in #1840, and the accel-layer wiring is now in #2241.

PR #1744, opened 2026-06-08
fix(core): deregister duplicate descriptors without double free

lluki:
#1635 introduces/surfaces a bug if two exact same regions get registered twice: they collapse to the same sort keys so the sequence: `register(A), register(A), deregister([A, A])` causes a double free on the second deregister.

This bugfix makes sure the correct metadata gets freed, even under the presence of duplicate registered regions. This will in turn lead to correct FD closing and no double-frees.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deregistration now performs an all-or-nothing resolution of requested memory descriptors before any cleanup, preventing partial deregistration or unload when descriptors are duplicated or mismatched.

* **Tests**
  * Added unit tests validating deregistration with duplicated descriptors and failing cases with missing descriptors to ensure cleanup runs exactly once per item or not at all on failure.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -208,6 +208,35 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
     return ret;
 }
 
+namespace {
+// Map each element to a distinct entry sharing its descriptor; duplicates share a getIndex match.
+// Returns false (resolving nothing) on any missing/over-counted element, keeping callers
+// all-or-nothing.
+bool
+resolveDuplicateIndices(const nixl_reg_dlist_t &mem_elms,
+                        const nixlSecDescList &target,
+                        std::vector<size_t> &indices) {
+    indices.clear();
+    indices.reserve(mem_elms.descCount());
+    std::map<size_t, size_t> claimed; // first-match index -> entries already taken
+    for (auto &elm : mem_elms) {
+        const int match = target.getIndex(elm);
+        if (match < 0) {
+            return false;
+        }
+        const size_t first = static_cast<size_t>(match);
+        const size_t idx = first + claimed[first]++;
+        // idx must still name the same region as first (mutual covers == equality)
+        if (idx >= static_cast<size_t>(target.descCount()) ||
+            !(target[idx].covers(target[first]) && target[first].covers(target[idx]))) {
+            return false;
+        }
+        indices.push_back(idx);
+    }
+    return true;
+}
+} // namespace
+
 nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
                                              nixlBackendEngine *backend) {
     if (!backend) {
@@ -222,15 +251,9 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
 
     nixlSecDescList &target = it->second;
 
-    // First check if the mem_elms are present in the list,
-    // don't deregister anything in case any is missing.
     std::vector<size_t> indices;
-    indices.reserve(mem_elms.descCount());
-    for (auto &elm : mem_elms) {
-        int index = target.getIndex(elm);
-        if (index < 0)
-            return NIXL_ERR_NOT_FOUND;
-        indices.push_back(static_cast<size_t>(index));
+    if (!resolveDuplicateIndices(mem_elms, target, indices)) {
+        return NIXL_ERR_NOT_FOUND;
     }
 
     for (size_t idx : indices) {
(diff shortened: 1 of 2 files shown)


src/infra/nixl_memory_section.cpp
```diff
@@ -222,15 +222,24 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
...
-        indices.push_back(static_cast<size_t>(index));
+        }
+        const size_t first = static_cast<size_t>(match);
+        const size_t idx = first + claimed[first]++;
+        // idx must still name the same region as first (mutual covers == equality)
+        if (idx >= static_cast<size_t>(target.descCount()) ||
+            !(target[idx].covers(target[first]) && target[first].covers(target[idx]))) {
+            return NIXL_ERR_NOT_FOUND;
+        }
+        indices.push_back(idx);
```
lluki: Good point. Will implement

lluki:
CodeRabbit pointed out (https://github.com/ai-dynamo/nixl/pull/1744#discussion_r3379381682) that the same bug exists in `removeLocalData`. I added my fix there as well. Moved the test code to the gmock based test ones. The new test covers both cases, as the gmock backend defaults to `supportsLocal()==true`.

src/infra/nixl_memory_section.cpp
```diff
@@ -208,6 +208,35 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
     return ret;
 }
 
+namespace {
+// Map each element to a distinct entry sharing its descriptor; duplicates share a getIndex match.
+// Returns false (resolving nothing) on any missing/over-counted element, keeping callers
+// all-or-nothing.
+bool
+resolveDuplicateIndices(const nixl_reg_dlist_t &mem_elms,
```
iyastreb, line 216: maybe just return indices without bool?
empty means error
`std::vector<size_t> findDuplicates(...)`

src/infra/nixl_memory_section.cpp
```diff
@@ -208,6 +208,35 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
...
+                        const nixlSecDescList &target,
+                        std::vector<size_t> &indices) {
+    indices.clear();
+    indices.reserve(mem_elms.descCount());
+    std::map<size_t, size_t> claimed; // first-match index -> entries already taken
```
iyastreb, line 221: Is std::map used on purpose instead of unordered_map?

src/infra/nixl_memory_section.cpp
lluki: Then I would have to handle on the callsite the special case when an empty de-registration is called. I think the function with the current signature is easier to use

src/infra/nixl_memory_section.cpp
lluki: Not on purpose. I dont have any preferences on this. Should i change it? I think it will be mostly used on smallish inputs, then `std::map` is slightly more memory efficient

test/gtest/unit/agent/agent.cpp
```diff
@@ -334,6 +334,96 @@ namespace agent {
...
+        EXPECT_CALL(engine, unloadMD(testing::_)).Times(0);
+
+        nixl_reg_dlist_t bad(DRAM_SEG);
+        bad.addDesc(registered.getDesc());
+        bad.addDesc(registered.getDesc());
+        bad.addDesc(missing.getDesc());
+        EXPECT_NE(agent_->deregisterMem(bad, &extra_params), NIXL_SUCCESS);
+
+        // Verify the Times(0) now -- fixture teardown deregisters the still-registered copies.
+        testing::Mock::VerifyAndClearExpectations(&engine);
```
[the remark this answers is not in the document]
lluki: I added this as 2nd testcase

test/gtest/unit/agent/agent.cpp
```diff
@@ -334,6 +334,97 @@ namespace agent {
...
+    }
```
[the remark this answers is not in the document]
lluki: too complicated

lluki:
I discussed with Ofer, we will go with a different approach. Set the PR to Draft for now

The gist of the new approach will be:
```
1) twice same basicDesc registration should be forbidden
2) Additionally, the path-mode requires unique devid per file
```

lluki:
Closed in favor of #1773

src/infra/nixl_memory_section.cpp
ovidiusm: Actually I also do not understand what this function does, since it seems to do several different things at the moment. What exactly is the current approach and what are the alternatives?

ovidiusm: For example, could we have a simpler deregistration check? I expected that since same-path descriptors map to the same "key" it would be possible to track duplicates without having to analyze everything

PR #1745, opened 2026-06-08
feat(core): add etcd backend for nixlMDManager (PR 2)

aschwartz12:
> [!IMPORTANT]
> **Stacked PR — based on #1690 (PR 1), not `main`.**
> Please **review only the latest commit:** `7d106349`
> Everything below it belongs to #1690 and is reviewed there. Once #1690 merges, this PR's base will be retargeted and the diff will collapse to just this commit.

## What?
Adds an ETCD metadata backend behind `nixlMDManager`, building on the P2P-only manager from #1690.

- Introduces a core-internal `nixlMetadataBackend` contract (`publish` / `fetch` / `remove` / `watch` / `fetchBatch`, plus `name()`) and extracts the existing ETCD client into an `EtcdMetadataBackend` subclass owned by the manager.
- `nixlMDManager` selects the transport at construction: ETCD when `NIXL_ETCD_ENDPOINTS` is set, otherwise the PR 1 P2P path is unchanged. The manager owns key building (`{namespace}/{agent}/{label}`) and dispatches send/sendPartial/fetch/invalidate through the backend.
- Watch-driven invalidation is drained opportunistically on `fetchRemoteMD` / `checkRemoteMD` (no background tick loop).
- `getBackend()` reports the active transport name (`"ETCD"` / `"P2P"`).

## Why?
The metadata manager introduced in #1690 only supported direct P2P socket exchange. Many deployments use a centralized ETCD store for metadata, so the manager needs a pluggable backend path that keys metadata by agent name in a shared store while keeping the same name-based API. This PR adds that path without changing P2P behavior, and structures it so further backends can be added without touching the public API.

## How?
- **Southbound backend abstraction.** The manager dispatches storage ops through `nixlMetadataBackend`. Keys are opaque to the backend (the manager builds them), so backend-specific dependencies (etcd, absl) stay core-internal and never leak into the installed public header — `nixl_md_manager.h` only forward-declares the backend, and `getBackend()` is defined in the `.cpp`.
- **Open backend identity.** `getBackend()` returns a `std::string_view` name reported by the backend itself rather than a closed enum, so adding a backend doesn't require editing a public type (continues the #1690 review decision).
- **Construction-time naming.** The manager receives the agent name via its constructor (`self_name_`) because the ETCD anchor key is built during construction, before the owning `nixlAgent`'s `data` pointer is assigned; `agent_.getName()` is unavailable at that point and is reserved for post-construction use. `mdManager_` remains a `const` member built in `nixlAgentData`'s initializer list, where ETCD's eager connect can throw to fail agent construction (health gate).
- **Shared registry.** The per-name peer registry is reused unchanged across backends; ETCD tracks the same names and ignores the P2P-only `{ip, port}` address fields.
- **Scope.** `nixl_listener.cpp`'s comm-thread ETCD path is intentionally left untouched; the duplicate client is collapsed in PR 3.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in, agent-managed metadata exchange flow that routes metadata operations between direct peer-to-peer and a centralized metadata backend when available.
* **Bug Fixes**
  * Centralized metadata reads now treat “present but empty” values as not found.
  * Centralized metadata watcher timeouts now cancel promptly to avoid late callbacks.
* **Tests**
  * Added automated coverage for both peer-to-peer and centralized metadata workflows.


src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,249 @@
...
+    nixlMDManager &
+    operator=(const nixlMDManager &) = delete;
+
+private:
+    // Registry entry for a tracked agent. ip/port are P2P-only addressing,
+    // unused by centralized backends. `epoch` (from registerEpoch_) stamps
+    // each (re)register so unregister's compare-then-erase won't drop an entry
+    // re-registered during an in-flight invalidate; equality includes it.
+    struct Peer {
+        std::string ip;
```
ColinNV: It's not nice that backend-specific details leak into the general manager. Can this not be abstracted away somehow so that each backend can have its own addressing type and the generic parts handle it as opaque datatype? We're doing the same thing in many places with the other backends. It doesn't make for the nicest APIs, but at least it's flexible and not so "leaky".

src/api/cpp/nixl_md_manager.h
```diff
@@ -0,0 +1,249 @@
...
+    makeKey(const std::string &agent_name, const std::string &label) const;
+
+    // Apply queued remote-invalidation events to the agent's cache. Called
+    // opportunistically from fetchRemoteMD/checkRemoteMD; no-op on P2P.
+    void
+    drainInvalidated() const;
+
+    // ETCD is active when a backend object exists; P2P leaves backend_ null.
+    [[nodiscard]] bool
+    usingEtcd() const noexcept {
```
ColinNV: Same here, we have the generic MD backend interface, but then some hard-coded assumptions on the set of available backends...

ColinNV (changes requested):

PR #1747, opened 2026-06-09
Fix libfabric transfer request lifecycle bugs

### LINKED ISSUE #1722: libfabric backend: releaseReqH() does not delete backend request handle — memory leak
## Summary

`nixlLibfabricEngine::releaseReqH()` does not `delete` the `nixlLibfabricBackendH` object allocated in `prepXfer()`, causing a memory leak on every transfer request lifecycle.

## Environment

- NIXL main branch (latest commit as of 2026-06-03)
- libfabric backend with EFA provider
- Multi-GPU benchmark (8x GPU, 1024 requests/iteration)

## Root Cause

In `src/plugins/libfabric/libfabric_backend.cpp`, `releaseReqH()` simply returns without freeing the handle:

```cpp
nixl_status_t
nixlLibfabricEngine::releaseReqH(nixlBackendReqH *handle) const {
    if (!handle) {
        return NIXL_SUCCESS;
    }
    // Let NIXL framework handle the deletion   <-- INCORRECT: framework does NOT delete
    NIXL_DEBUG << "releaseReqH completed successfully";
    return NIXL_SUCCESS;
}
```

The comment "Let NIXL framework handle the deletion" is **incorrect**. The framework's `~nixlXferReqH()` only calls `engine->releaseReqH(backendHandle)` — it **never** calls `delete backendHandle`. This is confirmed by:

1. `src/core/transfer_request.h` — destructor only calls `releaseReqH`, no `delete`
2. `docs/BackendGuide.md` line 93: _"releaseReqH(): Releases a transfer request handle... **freeing resources**"_
3. Every other backend (UCX, Mooncake, HiXL, HF3FS, OBJ, CUDA GDS) performs `delete handle` inside `releaseReqH()`

## Impact

- Every `postXfer()` call allocates a `new nixlLibfabricBackendH()` (in `prepXfer`)
- The handle is never freed → **unbounded memory growth** proportional to transfer count
- In production disaggregated inference: thousands of transfers/second → rapid OOM

## Proposed Fix

```cpp
nixl_status_t
nixlLibfabricEngine::releaseReqH(nixlBackendReqH *handle) const {
    if (!handle) {
        return NIXL_SUCCESS;
    }

    nixlLibfabricBackendH *lf_handle = static_cast<nixlLibfabricBackendH *>(handle);
    delete lf_handle;

    NIXL_DEBUG << "releaseReqH completed successfully";
    return NIXL_SUCCESS;
}
```

## Additional Related Issues Found

During the same code audit, we also identified several other resource management issues in the libfabric backend:

1. **Control request leak in `postXfer()`** — a `ControlRequestPool` request is allocated unconditionally at the start of `postXfer()` but only released (via send completion callback) when `hasNotif == true`. Without notification, the request permanently leaks from the 256-entry pool.

2. **`received_remote_writes_` unbounded growth** — the `unordered_set<uint32_t>` tracking received XFER_IDs is never cleared, growing indefinitely and causing false positives after XFER_ID wraparound (20-bit counter).

3. **CQ error path does not release failed requests** — when `fi_cq_read` returns error, `fi_cq_readerr` is called for logging but the errored request (identifiable via `err_entry.op_context`) is never returned to the pool.

Happy to submit PRs for any/all of these.

## References

- UCX backend correctly deletes: `src/plugins/ucx/ucx_backend.cpp` (releaseReqH → `delete intHandle`)
- Mooncake backend correctly deletes: `src/plugins/mooncake/mooncake_backend.cpp` (releaseReqH → `delete priv`)
- BackendGuide documentation: `docs/BackendGuide.md` line 93

amitrad-aws:
## What?
Three fixes to the libfabric backend's transfer request lifecycle (Fixes https://github.com/ai-dynamo/nixl/issues/1722)

  1. releaseReqH() now deletes the nixlLibfabricBackendH handle
  2. Removed dead received_remote_writes_ field and unused <unordered_set> include
  3. CQ errors are now propagated to the transfer handle via the completion callback, and the failed request is released back to the pool

## Why?
- Memory leak: Every postXfer() allocates a new nixlLibfabricBackendH that was never freed 
- Dead code: received_remote_writes_ was declared but never used in any .cpp file
- CQ error hang: When fi_cq_read returned an error, the failed request was never released to the pool and the owning handle was never notified — causing transfers to hang indefinitely (and leaking pool entries)

## How?
Memory leak fix: Added delete static_cast<nixlLibfabricBackendH *>(handle) in releaseReqH()

CQ error propagation: Changed completion_callback signature from void() to void(nixl_status_t). On CQ error, the callback is invoked with NIXL_ERR_BACKEND — this increments the completion counter (so is_completed() eventually becomes true) and sets an error flag on the handle. On success, NIXL_SUCCESS is passed as before.
checkXfer() waits for all requests to resolve before returning status. This avoids premature cancellation of in-flight requests on other rails — following the same drain-then-report pattern used by the UCX backend.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transfer failures are now recorded and reported promptly, preventing false successes and stuck requests.
  * Completion errors now trigger failure notifications and reliably release affected requests.
  * Failed transfers no longer continue through success-only completion handling.

* **API Changes**
  * Transfer and control-message completion callbacks now provide explicit status results.

* **Chores**
  * Improved request cleanup and removed redundant transfer-tracking state.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -388,6 +388,7 @@ nixlLibfabricEngine::vramApplyCtxEx(bool &use_cuda_addr_wa) const {
 nixlLibfabricBackendH::nixlLibfabricBackendH(nixl_xfer_op_t op, const std::string &remote_agent)
     : completed_requests_(0),
       submitted_requests_(0),
+      error_status_(NIXL_SUCCESS),
       operation_(op),
       remote_agent_(remote_agent),
       total_notif_msg_len(0) {
@@ -408,6 +409,7 @@ void
 nixlLibfabricBackendH::init_request_tracking(size_t num_requests) {
     submitted_requests_.store(num_requests);
     completed_requests_.store(0);
+    error_status_.store(NIXL_SUCCESS);
     NIXL_DEBUG << "Initialized request tracking for " << num_requests << " requests";
 }
 
@@ -432,6 +434,17 @@ void
 nixlLibfabricBackendH::adjust_total_submitted_requests(size_t actual_count) {
     submitted_requests_.store(actual_count);
     NIXL_DEBUG << "Adjusted total requests to actual count: " << actual_count;
+}
+
+void
+nixlLibfabricBackendH::set_error(nixl_status_t status) {
+    nixl_status_t expected = NIXL_SUCCESS;
+    error_status_.compare_exchange_strong(expected, status);
+}
+
+nixl_status_t
+nixlLibfabricBackendH::get_error_status() const {
+    return error_status_.load();
 }
 
 bool
@@ -1353,7 +1366,12 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             conn->rail_remote_addr_list_,
             imm_agent_idx,
             backend_handle->post_xfer_id,
-            [backend_handle]() { backend_handle->increment_completed_requests(); },
+            [backend_handle](nixl_status_t status) {
+                backend_handle->increment_completed_requests();
+                if (status != NIXL_SUCCESS) {
+                    backend_handle->set_error(status);
+                }
+            }, // Completion callback
             desc_submitted_count,
             desc_idx,
             xfer_base_offset,
@@ -1630,6 +1648,13 @@ nixlLibfabricEngine::checkXfer(nixlBackendReqH *handle) const {
 
     // Then check for completions after processing any pending completions
     if (backend_handle->is_completed()) {
+        // Check if any request completed with error
+        nixl_status_t err = backend_handle->get_error_status();
+        if (err != NIXL_SUCCESS) {
+            NIXL_ERROR << "Transfer completed with CQ error";
+            return err;
+        }
+
         NIXL_DEBUG << "Data transfer completed successfully";
         if (backend_handle->has_notif && backend_handle->operation_ == nixl_xfer_op_t::NIXL_READ) {
             nixl_status_t notif_status = notifSendPriv(backend_handle->remote_agent_,
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -388,6 +388,7 @@ nixlLibfabricEngine::vramApplyCtxEx(bool &use_cuda_addr_wa) const {
 nixlLibfabricBackendH::nixlLibfabricBackendH(nixl_xfer_op_t op, const std::string &remote_agent)
     : completed_requests_(0),
       submitted_requests_(0),
+      error_status_(NIXL_SUCCESS),
       operation_(op),
       remote_agent_(remote_agent),
       total_notif_msg_len(0) {
@@ -408,14 +409,23 @@ void
 nixlLibfabricBackendH::init_request_tracking(size_t num_requests) {
     submitted_requests_.store(num_requests);
     completed_requests_.store(0);
+    error_status_.store(NIXL_SUCCESS);
     NIXL_DEBUG << "Initialized request tracking for " << num_requests << " requests";
 }
 
 void
-nixlLibfabricBackendH::increment_completed_requests() {
-    completed_requests_.fetch_add(1);
-    NIXL_DEBUG << "Request completed, total completed: " << completed_requests_.load() << "/"
-               << submitted_requests_.load();
+nixlLibfabricBackendH::complete_request(nixl_status_t status) {
+    if (status != NIXL_SUCCESS) {
+        nixl_status_t expected = NIXL_SUCCESS;
+        error_status_.compare_exchange_strong(
+            expected, status, std::memory_order_relaxed, std::memory_order_relaxed);
+    }
+    // Release ensures the error store above is visible to any thread that
+    // observes the incremented count via an acquire load in is_completed().
+    completed_requests_.fetch_add(1, std::memory_order_release);
+    NIXL_DEBUG << "Request completed (status=" << status
+               << "), total completed: " << completed_requests_.load(std::memory_order_relaxed)
+               << "/" << submitted_requests_.load(std::memory_order_relaxed);
 }
 
 size_t
@@ -434,10 +444,17 @@ nixlLibfabricBackendH::adjust_total_submitted_requests(size_t actual_count) {
     NIXL_DEBUG << "Adjusted total requests to actual count: " << actual_count;
 }
 
+nixl_status_t
+nixlLibfabricBackendH::get_error_status() const {
+    return error_status_.load(std::memory_order_acquire);
+}
+
 bool
 nixlLibfabricBackendH::is_completed() const {
-    // Transfer is completed when all requests have local completions
-    return completed_requests_.load() == submitted_requests_.load();
+    // Acquire pairs with release in complete_request() — guarantees we see
+    // any error_status_ written before the final increment.
+    return completed_requests_.load(std::memory_order_acquire) ==
+        submitted_requests_.load(std::memory_order_relaxed);
 }
 
 /****************************************
@@ -1353,7 +1370,9 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             conn->rail_remote_addr_list_,
             imm_agent_idx,
             backend_handle->post_xfer_id,
-            [backend_handle]() { backend_handle->increment_completed_requests(); },
+            [backend_handle](nixl_status_t status) {
+                backend_handle->complete_request(status);
+            }, // Completion callback
             desc_submitted_count,
             desc_idx,
             xfer_base_offset,
@@ -1630,6 +1649,13 @@ nixlLibfabricEngine::checkXfer(nixlBackendReqH *handle) const {
 
     // Then check for completions after processing any pending completions
     if (backend_handle->is_completed()) {
+        // Check if any request completed with error
+        nixl_status_t err = backend_handle->get_error_status();
+        if (err != NIXL_SUCCESS) {
+            NIXL_ERROR << "Transfer completed with CQ error";
+            return err;
+        }
+
         NIXL_DEBUG << "Data transfer completed successfully";
         if (backend_handle->has_notif && backend_handle->operation_ == nixl_xfer_op_t::NIXL_READ) {
             nixl_status_t notif_status = notifSendPriv(backend_handle->remote_agent_,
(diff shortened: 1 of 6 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -738,6 +738,17 @@ nixlLibfabricRail::progressCompletionQueue() const {
...
+
+                // Notify the owning handle of the error and release the request
+                if (err_entry.op_context) {
+                    nixlLibfabricReq *req = findRequestFromContext(err_entry.op_context);
+                    if (req && req->in_use) {
+                        if (req->completion_callback) {
+                            req->completion_callback(NIXL_ERR_BACKEND);
+                        }
+                        releaseRequest(req);
+                    }
```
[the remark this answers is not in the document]
amitrad-aws: On RQ completion with error the data isn't valid so there is no way to trace it to a specific transfer.
However, the sender will also get an error to indicate this notificaiton will never get to the target

fengjica (dismissed):

dpressle:
👀 Investigating [`nixl-ci-dl-gpu` build 860](https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-dl-gpu/detail/nixl-ci-dl-gpu/860/pipeline)…

dpressle:
🤖 **CI Triage Agent** — `nixl-ci-dl-gpu` · commit `35f004ba`

I now have a comprehensive picture of the failure. Here is the full diagnosis:

---

**Summary:** `nixl-ci-dl-gpu` build #860 failed with `script returned exit code 1` due to `nixlLibfabricEngine::releaseReqH()` not deleting the backend request handle, causing a memory leak and likely a double-free or use-after-free when the NIXL framework also attempts to manage the same handle — the very bug this PR (#1747) was opened to fix, but the fix is incomplete.

**Root cause:** `releaseReqH()` in `src/plugins/libfabric/libfabric_backend.cpp` (lines 1188–1202) is the identified memory leak site from issue #1722. The implementation explicitly **skips deleting the handle** with the comment *"Let NIXL framework handle the deletion"*, but the NIXL framework **does not** delete the backend-level `nixlBackendReqH*` — that is the contract of `releaseReqH()` itself. The correct fix (mirroring the UCX plugin pattern and the intent of issue #1722) is `delete static_cast<nixlLibfabricBackendH*>(handle)`. 

The test SLURM jobs (`job_id_v1.21.x_860` and `job_id_master_860`) both returned exit code 1, which triggered `parallel --halt now,fail=1`. The SLURM jobs ran the nixlbench UCX benchmark suite (DRAM→DRAM READ/WRITE), which exercises `prepXfer`/`postXfer`/`checkXfer`/`releaseReqH` in a tight loop. With the handle never freed, the process accumulates leaked `nixlLibfabricBackendH` objects; when a sanitizer or allocator guard catches this at process exit (or when the framework touches the unreleased handle), the process crashes with a non-zero exit.

The current `releaseReqH` on `master` (latest commit `[REDACTED:Hex High Entropy String]`) also does not delete the handle — this is the same unpatched state as on the PR branch.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `fix/libfabric-releaseReqH-memleak`, PR author), but the root bug predates it — present since `[REDACTED:Hex High Entropy String]` (Ye Xiang, 2026-05-13) which last touched the libfabric plugin.

**File:** `src/plugins/libfabric/libfabric_backend.cpp:1188–1202`

**Suggested fix:** Replace the stub body of `releaseReqH` with an actual `delete` of the handle:

```cpp
nixl_status_t
nixlLibfabricEngine::releaseReqH(nixlBackendReqH *handle) const {
    if (!handle) {
        return NIXL_SUCCESS;
    }
    // The backend owns the handle allocated in prepXfer — delete it here.
    delete static_cast<nixlLibfabricBackendH *>(handle);
    NIXL_DEBUG << "releaseReqH: handle deleted successfully";
    return NIXL_SUCCESS;
}
```

This mirrors the contract fulfilled by every other backend plugin (e.g., UCX). The comment *"Let NIXL framework handle the deletion"* is incorrect — the framework calls `releaseReqH` as the *only* deletion point for backend request handles. After applying this fix, confirm no double-free by ensuring `prepXfer` always allocates a fresh `nixlLibfabricBackendH` via `new` (it does, at line 925) and that no other code path deletes it.

**Related:**
- Issue #1722: https://github.com/ai-dynamo/nixl/issues/1722 — *libfabric backend: `releaseReqH()` does not delete backend request handle — memory leak*
- PR #1747: https://github.com/ai-dynamo/nixl/pull/1747 — *Fix libfabric transfer request lifecycle bugs*



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `eeaa4274-40a3-421d-91ec-3c2761724a76` in the triage console for the audit trail.

fengjica:
@brminich Could you help rebuild this? Thanks!

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -434,6 +436,17 @@ nixlLibfabricBackendH::adjust_total_submitted_requests(size_t actual_count) {
     NIXL_DEBUG << "Adjusted total requests to actual count: " << actual_count;
 }
 
+void
+nixlLibfabricBackendH::set_error(nixl_status_t status) {
+    nixl_status_t expected = NIXL_SUCCESS;
+    error_status_.compare_exchange_strong(expected, status);
```
oa-aws: Maybe can use load-relaxed and check value to avoid using CAS each time? With single progress thread it doesn't make a difference, but if multiple progress threads are used in the future, it can have an impact.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1353,7 +1366,12 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             conn->rail_remote_addr_list_,
             imm_agent_idx,
             backend_handle->post_xfer_id,
-            [backend_handle]() { backend_handle->increment_completed_requests(); },
+            [backend_handle](nixl_status_t status) {
+                backend_handle->increment_completed_requests();
```
oa-aws: Seems code-rabbit has a point here. There is a race condition. If only the last request failed, user code could theoretically poll for request completion just when number of completed requests reached the total, but before the error flag was raised - and arriving to the wrong conclusion that the request completed successfully when it didn't. So the error status must be set before incrementing the completed count. 

Also, In order to avoid this in other possible places, maybe it is better to put this logic within the backend handle code, pass a status parameter to increment_completed_requests(), and avoid altogether the need for a separate call to set_error()?

src/utils/libfabric/libfabric_rail.cpp
oa-aws, line 1206: Looks like the request should be released if it is not null (as in other places). This could result in request pool exhaustion.

oa-aws, line 1224: Request should be released here too if it is not null?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1271,7 +1282,7 @@ nixlLibfabricRail::drainPostQueue() {
             if (pr.req && pr.req->completion_callback) {
                 // completion is notified also for failed requests
                 // (otherwise counters would never match)
-                pr.req->completion_callback();
+                pr.req->completion_callback(NIXL_ERR_BACKEND);
             }
             continue;
```
oa-aws, line 1296: Request should be released here too (before continue) if it is not null?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -434,6 +436,17 @@ nixlLibfabricBackendH::adjust_total_submitted_requests(size_t actual_count) {
...
 
+void
+nixlLibfabricBackendH::set_error(nixl_status_t status) {
+    nixl_status_t expected = NIXL_SUCCESS;
+    error_status_.compare_exchange_strong(expected, status);
+}
+
+nixl_status_t
+nixlLibfabricBackendH::get_error_status() const {
+    return error_status_.load();
```
oa-aws: Probably not having any impact, but maybe relaxed load is better for caller (without parameter it is seq-cst)?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -408,6 +409,7 @@ void
 nixlLibfabricBackendH::init_request_tracking(size_t num_requests) {
     submitted_requests_.store(num_requests);
     completed_requests_.store(0);
+    error_status_.store(NIXL_SUCCESS);
```
oa-aws, line 412: Probably having low impact, but maybe all stores should be relaxed, except for the last having release semantics? Without parameters they are all seq-cst.

src/plugins/libfabric/libfabric_backend.cpp
amitrad-aws: A specific backend should not be shared between multiple threads transfering.
Also, setting an error is a not a fast path flow, this is not a data path operation

oa-aws: Ok.

amitrad-aws: Done

src/plugins/libfabric/libfabric_backend.cpp
amitrad-aws: Done

src/plugins/libfabric/libfabric_backend.cpp
amitrad-aws: Done

src/plugins/libfabric/libfabric_backend.cpp
amitrad-aws: done

src/utils/libfabric/libfabric_rail.cpp
amitrad-aws: done

amitrad-aws: done

src/utils/libfabric/libfabric_rail.cpp
amitrad-aws: done

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: I can't see the change. Is it committed?

rongbingzhou (approved):

src/plugins/libfabric/libfabric_backend.cpp
amitrad-aws: I didn't change the store themselves, they are only under error case and in init. 
I don't want to make any assumptions in an error case

oa-aws: Ok

oa-aws (approved):

PR #1748, opened 2026-06-09
test/doca: run DOCA telemetry exporter tests in CI and broaden coverage

e-eygin:
## What?
- Run the standalone DOCA telemetry tests (`doca_test`, `doca_nixl_test`) in CI — they currently build but are never executed.
- Align `test/doca-telemetry/meson.build` with how the exporter plugin is gated (pkg-config `dependency()` + `TELEMETRY_DOCA`), replacing the hardcoded `/opt/mellanox/doca` path.
- Broaden exporter coverage: counter (all values), gauge (last value), and multi-agent label distinction.
- Rework the scrape helper into three focused headers: a strict OpenMetrics text parser, an in-memory time-series model + query view, and a thin polling helper.

## Why?
The DOCA telemetry tests build but never run — `.gitlab/test_cpp.sh` invokes explicit binaries (not `meson test`) and omits them — so the DOCA exporter has no end-to-end coverage in CI. The test's meson gate also still uses a hardcoded `/opt/mellanox/doca` header/lib path (with a now-stale comment), inconsistent with the pkg-config gate the exporter plugin uses. Running these tests exercises the exporter end-to-end (`exportEvent` → `flush` → `/metrics` scrape); the stronger, label-aware assertions catch dropped, duplicated, or mis-typed samples that a name-only or last-value check would miss. The scrape parser is deliberately strict, so an exporter-format regression surfaces as a missing series (a failing assertion) instead of being silently absorbed.

## How?
- **`.gitlab/test_cpp.sh`**: invoke `./bin/doca_test` and `./bin/doca_nixl_test`, guarded by `-x` so they no-op on non-DOCA images. No `LD_LIBRARY_PATH` change — the DOCA telemetry libraries resolve via ldconfig (`ld.so.conf.d`).
- **`test/doca-telemetry/meson.build`**: build under `dependency('doca-telemetry-exporter'/'doca-common')` + `enabled_plugins.get('TELEMETRY_DOCA')`, matching `src/plugins/telemetry/meson.build`; honors `-Ddisable_plugins=TELEMETRY_DOCA`.
- **Scrape helper split into three headers** (acyclic layering: model ← parser ← scrape loop):
  - `timeseries.h` — the time-series data model (`labelSet`, `sample`, `seriesId`, `samples`, `seriesMap`) plus the `timeSeries` query view, looked up by metric name **+ label subset**. The view is built from an already-parsed `seriesMap`, decoupling it from parsing.
  - `open_metrics_text_parser.h` — `OpenMetricsTextParser`, a `std::regex`-based Prometheus/OpenMetrics text parser (no new deps). Strict by design: it rejects malformed label blocks (partial parse, missing/doubled comma, trailing comma, duplicate key) and non-fully-numeric value/timestamp tokens.
  - `scrape_util.h` — `scrapeUntilValue`, the polling helper that wires HTTP → parse → `timeSeries`.
- **Tests**: `telemetry_doca_nixl_test.cpp` — counter asserts cumulative == exact sum of distinct deltas; gauge asserts last-value-wins (`add_gauge` is absolute); two agents emit the same metric under different `agent_name` labels, kept apart by label set (a name-only lookup is ambiguous). `telemetry_doca_test.cpp` keeps the raw-DOCA-API counter test.

Validated locally: `meson test doca_test doca_nixl_test` (4 tests) green; `-Ddisable_plugins=TELEMETRY_DOCA` skips/unregisters them; clang-format clean; `-Werror` build. Confirmed green in CI (`nixl-ci-non-gpu` runs the DOCA tests).

## Summary by CodeRabbit

* **Tests**
  * Expanded DOCA telemetry end-to-end coverage for labeled counters and gauges, including per-agent isolation and verification that label-aware queries succeed while name-only lookups are rejected when ambiguous.
* **Refactor**
  * Reworked telemetry checks to use label-aware metric polling with a stricter OpenMetrics/Prometheus text parser and an in-memory time-series model for consistent “latest value” assertions.
* **Chores**
  * Improved standalone telemetry exporter test build/run gating by using build-time dependency discovery and conditionally executing only when the relevant test binaries are present and runnable.

### DIFF
--- a/test/doca-telemetry/open_metrics_text_parser.h
+++ b/test/doca-telemetry/open_metrics_text_parser.h
@@ -1,0 +1,156 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef NIXL_TEST_DOCA_TELEMETRY_OPEN_METRICS_TEXT_PARSER_H
+#define NIXL_TEST_DOCA_TELEMETRY_OPEN_METRICS_TEXT_PARSER_H
+
+#include <algorithm>
+#include <cstdint>
+#include <exception>
+#include <optional>
+#include <regex>
+#include <sstream>
+#include <string>
+#include <utility>
+
+#include "timeseries.h"
+
+namespace nixl::doca_test {
+
+// Parses a Prometheus / OpenMetrics text exposition body into time series.
+//
+// The grammar handled is the line format DOCA's endpoint serves:
+//   name[{key="value"(,key="value")*}] value [timestamp]
+// Comment (`#`) and blank lines are skipped. The parser is deliberately strict --
+// a malformed label block (missing/doubled/trailing comma, repeated key, stray
+// text) rejects the whole line rather than yielding partial labels -- so an
+// exporter-format regression surfaces as a missing series in these tests instead
+// of being silently absorbed.
+class OpenMetricsTextParser {
+public:
+    // Parse a whole /metrics body once. Each well-formed line contributes one
+    // sample to its (name+labels) series; the resulting map lets a caller query
+    // any number of series without rescanning the raw text.
+    [[nodiscard]] static seriesMap
+    parse(const std::string &body) {
+        seriesMap series;
+        std::istringstream stream(body);
+        std::string line;
+        while (std::getline(stream, line)) {
+            if (auto parsed = parseLine(line)) {
+                series[std::move(parsed->first)].push_back(parsed->second);
+            }
+        }
+        return series;
+    }
+
+private:
+    // Parse one exposition line into a series id + sample. Returns nullopt for
+    // comment/blank/malformed lines. Label values are assumed free of embedded
+    // quotes (true for these metrics).
+    [[nodiscard]] static std::optional<std::pair<seriesId, sample>>
+    parseLine(const std::string &line) {
+        // name, optional brace-delimited label block, value, optional timestamp.
+        static const std::regex lineRe(R"(^([^\s{]+)(?:\{([^}]*)\})?\s+(\S+)(?:\s+(\S+))?\s*$)");
+
+        std::smatch match;
+        if (line.empty() || line[0] == '#' || !std::regex_match(line, match, lineRe)) {
+            return std::nullopt;
+        }
+        const std::optional<sample> value = parseSample(
+            match[3].str(),
+            match[4].matched ? std::optional<std::string>(match[4].str()) : std::nullopt);
+        if (!value) {
+            return std::nullopt;
+        }
+        std::optional<labelSet> labels = parseLabels(match[2].str());
+        if (!labels) {
+            return std::nullopt;
+        }
+        return std::make_pair(seriesId{match[1].str(), std::move(*labels)}, *value);
+    }
+
+    // Build a sample from its value token (required) and the optional timestamp
+    // token (the exposition's trailing field). Returns nullopt if the value is
+    // non-numeric; a missing or non-numeric timestamp is left unset, since
+    // Prometheus timestamps are optional.
+    [[nodiscard]] static std::optional<sample>
+    parseSample(const std::string &valueToken, const std::optional<std::string> &timestampToken) {
+        sample s;
+        try {
+            s.value = std::stod(valueToken);
+        }
+        catch (const std::exception &) {
+            return std::nullopt;
+        }
+        if (timestampToken) {
+            try {
+                s.timestamp = static_cast<uint64_t>(std::stoull(*timestampToken));
+            }
+            catch (const std::exception &) {
+                // Leave the timestamp unset on a non-numeric token.
+            }
+        }
+        return s;
+    }
+
+    // The gap between two matched pairs must hold exactly one comma (pairs are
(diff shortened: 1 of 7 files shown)


test/doca-telemetry/scrape_util.h
```diff
@@ -19,67 +19,203 @@
...
+        // Custom raw-string delimiter so the pattern's )" does not close it early.
+        static const std::regex labelRe(R"re(([^=,\s]+)\s*=\s*"([^"]*)")re");
+        labelSet labels;
+        for (auto it = std::sregex_iterator(block.begin(), block.end(), labelRe);
+             it != std::sregex_iterator();
+             ++it) {
+            labels.emplace((*it)[1].str(), (*it)[2].str());
+        }
+        return labels;
     }
```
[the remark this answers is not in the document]
e-eygin: fixed.

test/doca-telemetry/scrape_util.h
```diff
@@ -19,67 +19,219 @@
...
+        for (auto it = std::sregex_iterator(block.begin(), block.end(), labelRe);
+             it != std::sregex_iterator();
+             ++it) {
+            const auto &match = *it;
+            // Only commas/whitespace may separate (and precede) the matched pairs.
+            if (block.find_first_not_of(", \t", cursor) < static_cast<size_t>(match.position())) {
+                return std::nullopt;
+            }
+            labels.emplace(match[1].str(), match[2].str());
+            cursor = static_cast<size_t>(match.position() + match.length());
```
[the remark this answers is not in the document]
e-eygin: fixed.

e-eygin:

test/doca-telemetry/timeseries.h
```diff
@@ -0,0 +1,121 @@
...
+#include <map>
+#include <optional>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+namespace nixl::doca_test {
+
+using labelSet = std::map<std::string, std::string>;
```
iyastreb, line 45: are you using tree map on purpose?
why not hash map?

test/doca-telemetry/timeseries.h
```diff
@@ -0,0 +1,121 @@
...
+};
+
+// Identity of a Prometheus time series: a metric name plus its label set. Every
+// sample of a series shares this identity, so it is the natural key -- the name
+// and labels live here once, not duplicated onto each sample.
+struct seriesId {
+    std::string name;
+    labelSet labels;
+
+    friend bool
```
iyastreb: this struct has public access to all fields, why method is a friend?

test/doca-telemetry/timeseries.h
e-eygin: Deliberate. labelSet is part of seriesId, which is the key of
seriesMap (std::map<seriesId, samples>), and seriesId::operator< orders by
std::tie(name, labels). std::map provides operator< (lexicographic);
std::unordered_map doesn't, so labels can't participate in the key ordering
without a custom hash for the whole key. Label sets here are tiny (hostname +
agent_name), so tree-vs-hash perf is irrelevant, and ordered iteration keeps the
parsed series deterministic.

test/doca-telemetry/timeseries.h
e-eygin: Good question, thanks -- and you're right it's not needed for access here (the
fields are public). I used the hidden-friend form out of habit: it's a common
way to define operator< as a non-member so both operands are treated
symmetrically. For seriesId it's functionally identical to a member operator<,
so it's really just a style preference -- happy to switch to a plain member
operator< if you'd find that clearer. Left it as-is for now since it's a small
test helper.

iyastreb: Yes, imo it's better to remove friend
Can you don't mark conversations as resolved? Normally reviewer marks them as resolved

e-eygin: dropped the friend as suggested.

test/doca-telemetry/timeseries.h
```diff
@@ -0,0 +1,121 @@
...
+};
+
+// Identity of a Prometheus time series: a metric name plus its label set. Every
+// sample of a series shares this identity, so it is the natural key -- the name
+// and labels live here once, not duplicated onto each sample.
+struct seriesId {
+    std::string name;
+    labelSet labels;
+
+    bool
```
ColinNV: `[[nodiscard]]` and `noexcept`, and can be a free function.

test/doca-telemetry/timeseries.h
```diff
@@ -0,0 +1,121 @@
...
+    [[nodiscard]] std::optional<double>
+    latestValue(const std::string &name, const labelSet &where = {}) const {
+        std::optional<double> result;
+        bool matched = false;
+        for (const auto &[id, seriesSamples] : series_) {
+            if (id.name != name || !labelsContain(id.labels, where)) {
+                continue;
+            }
+            if (matched) {
+                return std::nullopt; // ambiguous: same name, different label sets
```
ColinNV: Should this not be flagged as error?

test/doca-telemetry/open_metrics_text_parser.h
```diff
@@ -0,0 +1,167 @@
...
+// Parses a Prometheus / OpenMetrics text exposition body into time series.
+//
+// The grammar handled is the line format DOCA's endpoint serves:
+//   name[{key="value"(,key="value")*}] value [timestamp]
+// Comment (`#`) and blank lines are skipped. The parser is deliberately strict --
+// a malformed label block (missing/doubled/trailing comma, repeated key, stray
+// text) rejects the whole line rather than yielding partial labels -- so an
+// exporter-format regression surfaces as a missing series in these tests instead
+// of being silently absorbed.
+class OpenMetricsTextParser {
```
ColinNV: This does not need to be a class, it's just a bunch of `static` functions.

test/doca-telemetry/timeseries.h
e-eygin: Done — made it a free non-member operator<, marked [[nodiscard]] and noexcept.

test/doca-telemetry/open_metrics_text_parser.h
e-eygin: Done — replaced the class with an `open_metrics_text` namespace of free functions (helpers in a `detail` namespace). These are intentionally minimal, test-only helpers (not production-grade); I also added a header banner stating that scope.

test/doca-telemetry/timeseries.h
e-eygin: Good point — now throws `std::runtime_error` on an ambiguous (too-loose) lookup instead of silently returning `nullopt`.

ColinNV (approved):

PR #1750, opened 2026-06-10
feat: Add AMD accelerator detection and tracking in libfabric topology

mpashkovskii:
## What?
1. AMD GPU vendor ID detection (0x1002)
1. MI300X device ID range support (0x74a0-0x74af)
1. PCI class fallback detection for unknown AMD GPU models
1. `num_amd_accel` counter variable
1. `isAmdAccel()` helper function
1. `getNumAmdAccel()` public accessor
1. HMEM interface mapping returns `FI_HMEM_ROCR` for AMD GPUs
1. Topology-aware grouping includes AMD accelerators
1. Logging updated to report AMD GPUs separately

## Why?
To enable AMD GPU usage with libfabric plugin

## How?
Added hipification of libfabric plugins and tests

## Summary by CodeRabbit

* **New Features**
  * AMD (HIP/ROCm) GPU support added for accelerated communications alongside NVIDIA.

* **Improvements**
  * Runtime, memory registration, and context handling extended to treat AMD devices alongside CUDA where applicable.
  * Topology now reports NVIDIA/AMD/Neuron counts separately and includes AMD in NIC/EFA grouping.
  * Example Python script now detects missing optional bindings, logs a warning, and exits cleanly.

* **Tests**
  * Unit tests updated to exercise AMD/HIP GPU paths.

* **Chores**
  * Spell-check config updated to ignore AMD-related term.


examples/python/expanded_two_peers.py
```diff
@@ -11,10 +11,22 @@
...
+        from rixl.logging import get_logger
+
+    logger = get_logger(__name__)
+    NIXL_AVAILABLE = True
+except ImportError:
+    import logging
 
-logger = get_logger(__name__)
+    logger = logging.getLogger(__name__)
+    logger.error("NIXL API missing install NIXL.")
```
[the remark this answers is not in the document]
mpashkovskii: Fixed

test/unit/utils/libfabric/libfabric_topology_test.cpp
```diff
@@ -312,15 +316,15 @@ testBasicTopology() {
...
+        int num_gpus = topology.getNumNvidiaAccel() + topology.getNumAmdAccel();
         if (num_gpus > 0) {
             NIXL_INFO << "3. Testing GPU-specific queries (detected " << num_gpus << " GPUs)...";
             int test_gpus = std::min(num_gpus, 3); // Test up to 3 GPUs or all available
             for (int gpu_id = 0; gpu_id < test_gpus; ++gpu_id) {
 #ifdef CUDA_FOUND
                 // Get PCI bus ID for this GPU
                 cudaDeviceProp prop;
-                cudaGetDeviceProperties(&prop, gpu_id);
+                (void)cudaGetDeviceProperties(&prop, gpu_id);
```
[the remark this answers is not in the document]
mpashkovskii: fixed

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27408383539/job/81003143039)…

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27408408891/job/81003229411)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `38f40596`

I have all the information needed for a complete diagnosis. Here is the full report:

---

**Summary:** `clang-format-19` rejects a multi-line `NIXL_WARN` stream expression in `libfabric_topology_test.cpp` — the continuation is indented inconsistently with the project's `.clang-format` style.

**Root cause:** The PR introduced a new logging statement in `test/unit/utils/libfabric/libfabric_topology_test.cpp` around line 326 that breaks across two lines, but the second line was aligned with a 4-space indent relative to the first token rather than vertically aligned under the `<<` operator. `clang-format-19` requires the operator to be flush with the one above it. Specifically, the submitted code is:

```cpp
// SUBMITTED (fails):
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id
        << ": " << cudaGetErrorString(err);
```

`clang-format-diff-19` wants it reformatted to align the chained `<<` cleanly:

```cpp
// REQUIRED (passes):
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id << ": "
          << cudaGetErrorString(err);
```

The format check runs `clang-format-diff-19 -p1 -style=file` over the diff of all changed C/C++ files; when any hunk requires reformatting, it exits with code 1, failing the job.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1750, branch `feat/amd-detection-libfabric`)

**File:** `test/unit/utils/libfabric/libfabric_topology_test.cpp:326`

**Suggested fix:** Apply the exact reformatting that `clang-format-diff-19` requires. The simplest approach is to run the formatter locally before pushing:

```bash
# From the repo root, reformat only the changed lines:
git diff HEAD^1 HEAD -- test/unit/utils/libfabric/libfabric_topology_test.cpp \

# Then stage and amend/commit:
git add test/unit/utils/libfabric/libfabric_topology_test.cpp
git commit --amend --no-edit
```

Concretely, change lines 326-327 of `libfabric_topology_test.cpp` from:
```cpp
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id
        << ": " << cudaGetErrorString(err);
```
to:
```cpp
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id << ": "
          << cudaGetErrorString(err);
```

**Related:** none — this is a purely stylistic formatting issue introduced by the new AMD GPU-detection code in PR #1750.



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `5029689e-beeb-4f7e-b7f4-35ae580f3252` in the triage console for the audit trail.

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27408516982/job/81003599699)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `6d31e0a1`

The exact line numbers are slightly off due to the file view. The log diff is precise enough. The fix is clear and isolated. Here is the full diagnosis:

---

**Summary:** Clang Format Check failed — one line in `libfabric_topology_test.cpp` has incorrect stream-operator line-continuation alignment.

**Root cause:** `clang-format-19` found a formatting violation in `test/unit/utils/libfabric/libfabric_topology_test.cpp` around line 326–328. A multi-line `NIXL_WARN` stream expression has its continuation token (`<< ": "`) placed at the end of the first line instead of being aligned on the next line with the other stream operators. The diff from the CI check shows exactly what clang-format expects:

```diff
--- test/unit/utils/libfabric/libfabric_topology_test.cpp  (before formatting)
+++ test/unit/utils/libfabric/libfabric_topology_test.cpp  (after formatting)
@@ -326,8 +326,8 @@
-                    NIXL_WARN << "   Failed to get properties for GPU " << gpu_id
-                            << ": " << cudaGetErrorString(err);
+                    NIXL_WARN << "   Failed to get properties for GPU " << gpu_id << ": "
+                              << cudaGetErrorString(err);
```

The current code puts `<< ": "` on the second line with inconsistent indentation (12 spaces vs. the expected 14), whereas clang-format wants `<< ": "` kept on the first line and only the final `cudaGetErrorString` call moved to the second line, properly aligned (14 spaces / column-aligned with the opening `<<`).

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `feat/amd-detection-libfabric`, PR #1750)

**File:** `test/unit/utils/libfabric/libfabric_topology_test.cpp:326-327`

**Suggested fix:** Apply the exact reformatting shown in the diff above — move `<< ": "` to the end of the first line and realign the second line's `<< cudaGetErrorString(err)` to the proper column. The fastest way is to run clang-format locally and commit the result:

```bash
clang-format-19 -i -style=file test/unit/utils/libfabric/libfabric_topology_test.cpp
git add test/unit/utils/libfabric/libfabric_topology_test.cpp
git commit -m "style: fix clang-format alignment in libfabric_topology_test"
git push
```

**Related:** PR #1750 (`feat/amd-detection-libfabric`)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `f23c7511-b47c-4c81-abc7-66a9c9d608c8` in the triage console for the audit trail.

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `0561c241`

---

**Summary:** `clang-format` check failed on `test/unit/utils/libfabric/libfabric_topology_test.cpp` due to a line-break placement that doesn't conform to the project's `.clang-format` style rules.

**Root cause:** In commit `[REDACTED:Hex High Entropy String]`, the newly added test file `test/unit/utils/libfabric/libfabric_topology_test.cpp` contains a chained stream-insertion (`<<`) expression that was broken across lines in a way that `clang-format-19` disagrees with. Specifically, around line 326 of the file, the current code reads:

```cpp
// Current (violating) form:
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id
          << ": " << cudaGetErrorString(err);

// clang-format-19 requires:
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id << ": "
          << cudaGetErrorString(err);
```

The string literal `": "` must stay on the same line as `gpu_id` rather than starting a new continuation. The exit code was `1`, confirming a formatting violation was detected and the check failed.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `feat/amd-detection-libfabric`, PR #1750)

**File:** `test/unit/utils/libfabric/libfabric_topology_test.cpp` (around line 326–327)

**Suggested fix:** Apply the exact reformatting that `clang-format-19` requires. In your local checkout, run:

```bash
clang-format-19 -i -style=file test/unit/utils/libfabric/libfabric_topology_test.cpp
```

Or manually change the two offending lines from:
```cpp
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id
          << ": " << cudaGetErrorString(err);
```
to:
```cpp
NIXL_WARN << "   Failed to get properties for GPU " << gpu_id << ": "
          << cudaGetErrorString(err);
```
Then commit and push the fix. It is also advisable to run `clang-format-19 -i -style=file` across **all** files modified in this PR (the log lists 8 changed files) to catch any other violations before re-pushing.

> **Side note:** The workflow is using `actions/checkout@v3` which will be forcibly deprecated on June 16, 2026. Consider upgrading to `actions/checkout@v4` at the same time.

**Related:** [PR #1750 — feat: Add AMD accelerator detection and tracking in libfabric topology](https://github.com/ai-dynamo/nixl/pull/1750)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `352ab5d6-4f3c-4c15-954f-474cc3b1c5de` in the triage console for the audit trail.

mpashkovskii:
I fixed all the issues, I think. Could you @fengjica or @amitrad-aws review the PR?

examples/python/expanded_two_peers.py
```diff
@@ -11,10 +11,22 @@
...
 import torch
 
-from nixl._api import nixl_agent, nixl_agent_config
-from nixl.logging import get_logger
+try:
+    try:
+        from nixl._api import nixl_agent, nixl_agent_config
+        from nixl.logging import get_logger
+    except ImportError:
+        from rixl._api import nixl_agent, nixl_agent_config
```
edgargabriel: I don't think the rixl branch/part really required anymore.

mpashkovskii: You are right. Removed.

mpashkovskii:
@aranadive @brminich @ovidiusm would appreciate the PR review

edgargabriel:
I think there is some overlap between this PR and https://github.com/ai-dynamo/nixl/pull/1873

mpashkovskii:
@brminich, I made HIPifycation work only with `-Dwheel_variant=rocm` flag, any further comments?

ovidiusm:
This requires review from libfabric code owners @amitrad-aws @akkart-aws @fengjica @oa-aws

src/plugins/libfabric/libfabric_backend.cpp
fengjica: Is there any reason that this code path does not work for FI_HMEM_ROCR, since I see all other places make FI_HMEM_ROCR and FI_HMEM_CUDA behave the same?

src/utils/libfabric/meson.build
```diff
@@ -14,14 +14,65 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# AMD ROCm/HIP support for the libfabric plugin, selected with -Dwheel_variant=rocm
+# (the same switch that controls nixl_rocm wheel naming). Defined here in the first
```
fengjica: Is it a concern that the compilation depends on the order, which if changed may break it? Does it make sense to add it to both places, if it won't conflict, or to the one in the parent directory?

fengjica (commented):
Overall looks good. 

One question that I have is there may be some conflict with https://github.com/ai-dynamo/nixl/pull/1873. While the small conflict in the code can be merged later, I'm wondering if the two efforts are working toward the same goal or different specific ones. I'm not sure I fully understand the difference between ROCm and HIP, but it sounds like HIP targets a cuda-compatible approach while ROCm provides the underlying platform?

src/utils/libfabric/meson.build
mpashkovskii: Previously (in #1647 or another PR from @andyluo7) it was suggested to avoid putting this config into the root meson.build. However, I think the root meson.build is actually a good place for it — these duplicated chunks of code are already present in several places now, and centralizing it in the parent directory would reduce that duplication and eliminate the ordering concern altogether.

Shall I move it there?

mpashkovskii:
> One question that I have is there may be some conflict with https://github.com/ai-dynamo/nixl/pull/1873. While the small conflict in the code can be merged later, I'm wondering if the two efforts are working toward the same goal or different specific ones. I'm not sure I fully understand the difference between ROCm and HIP, but it sounds like HIP targets a cuda-compatible approach while ROCm provides the underlying platform?

@fengjica, yes, there is some overlap, but the two PRs target different hardware. #1873 adds support specifically for CXI accelerators, while this PR enables AMD GPUs. The code-level conflicts should be minor and easy to resolve at merge time.

src/plugins/libfabric/libfabric_backend.cpp
mpashkovskii: It is possible to adjust the rail manager to return FI_HMEM_CUDA for AMD devices, but, following [the existing manager's pattern](https://github.com/mpashkovskii/nixl/blob/main/src/utils/libfabric/libfabric_rail_manager.cpp#L205-L216) and symbols defined in ofiwg/libfabric, it seems cleaner to return a specific runtime per device:
- NVIDIA devices: FI_HMEM_CUDA
- AWS Trainium: FI_HMEM_NEURON
- AMD devices: FI_HMEM_ROCR

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -758,6 +777,22 @@ nixlLibfabricTopology::isNeuronAccel(hwloc_obj_t obj) const {
...
+    }
+    // AMD vendor ID is 0x1002
+    if (obj->attr->pcidev.vendor_id != 0x1002) {
+        return false;
+    }
+    // Only count devices with GPU class (0x300-0x3ff for display controllers)
+    // Class 0x302 is 3D controller (GPU), 0x680 is other devices (network, etc.)
+    // MI300X uses class 0x1200 (Processing accelerators), consumer GPUs use 0x300-0x3ff
+    uint16_t class_id = obj->attr->pcidev.class_id;
+    return (class_id >= 0x300 && class_id < 0x400) || (class_id == 0x1200);
```
ryanhankins: To ensure we catch all AMD Processing Accelerators like PR 1873 does, can we expand this to cover the 0x1200-0x12FF range?

`return (class_id >= 0x300 && class_id < 0x400) || (class_id >= 0x1200 && class_id < 0x1300);`

src/plugins/libfabric/meson.build
```diff
@@ -28,8 +42,13 @@ libfabric_plugin_deps = [
 
 compile_flags = ['-DHAVE_LIBFABRIC']
 
-# Add CUDA support if available
-if cuda_dep.found()
+# Add GPU support: CUDA normally, or HIP/ROCm when wheel_variant=rocm. The GPU
+# code paths are guarded by HAVE_CUDA; on ROCm the hipified sources compile those
+# paths against HIP via rocm_dep.
+if wheel_variant == 'rocm'
```
ryanhankins: PR 1873 introduced explicit use_rocm (bool) and rocm_path (string) options in the root meson_options.txt. Since that PR isn't merged yet, should we add those standard configuration options here rather
than solely relying on the $ROCM_PATH environment variable and wheel_variant?

```
    if wheel_variant == 'rocm'
        rocm_path = run_command('bash', '-c', 'echo -n "$ROCM_PATH"', check: false).stdout().strip()
        if rocm_path == ''
            rocm_path = '/opt/rocm'
        endif
```

ryanhankins:
> > One question that I have is there may be some conflict with #1873. While the small conflict in the code can be merged later, I'm wondering if the two efforts are working toward the same goal or different specific ones. I'm not sure I fully understand the difference between ROCm and HIP, but it sounds like HIP targets a cuda-compatible approach while ROCm provides the underlying platform?
> 
> @fengjica, yes, there is some overlap, but the two PRs target different hardware. #1873 adds support specifically for CXI accelerators, while this PR enables AMD GPUs. The code-level conflicts should be minor and easy to resolve at merge time.

I added a couple comments.  I would suggest we merge this PR, then I will see what remains and fix it up with a new PR with the remnants.

src/utils/libfabric/libfabric_topology.cpp
mpashkovskii: fixed

src/plugins/libfabric/meson.build
mpashkovskii: Frankly speaking I'm not sure. In both cases:
- if I bring common options from #1873 
- or, make `src/plugins/libfabric/meson.build` independent of `src/utils/libfabric/meson.build`

It would still require harmonisation of:
- `src/plugins/libfabric/meson.build`
- `src/utils/libfabric/meson.build`
- `benchmark/nixlbench/meson.build`
- `benchmark/nixlbench/meson_options.txt`
- with the root `meson.build` and `meson_options.txt`

Maybe I misunderstood your suggestion?

ryanhankins: I probably misunderstood, since meson isn't my area of expertise.  I'll try to test this PR and see if it works for us, but I suspect what's here already is fine.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -914,7 +915,7 @@ nixlLibfabricEngine::getSupportedMems() const {
     nixl_mem_list_t mems;
     mems.push_back(DRAM_SEG);
 #ifdef HAVE_CUDA
-    if (runtime_ == FI_HMEM_CUDA) {
+    if (runtime_ == FI_HMEM_CUDA || runtime_ == FI_HMEM_ROCR) {
         NIXL_DEBUG << "CUDA runtime detected, adding VRAM support";
```
aranadive: Suggest updating the debug message to be clear about the runtime

src/plugins/libfabric/meson.build
aranadive: @mpashkovskii I prefer the auto-detection versus adding more options to the meson files. But is this detection logic a bit fragile? How general is /opt/rocm as a path or the env var? And is it okay to break the build on non-standard paths?

src/utils/libfabric/meson.build
aranadive: This detection seems generic enough to add to src/meson.build or meson.build. Hmm I dont see the comment on #1647. To me it seems reasonable to have it there.

src/utils/libfabric/meson.build
```diff
@@ -14,14 +14,65 @@
...
+# compiler. Kept separate from cuda_dep. The ROCm toolchain path comes from
+# $ROCM_PATH, falling back to /opt/rocm.
+#
+# hipify injects <hip/hip_runtime.h> at the top of each translated source, ahead
+# of the nixl/toml++ headers. HIP defines __noinline__ as a macro, which breaks
+# toml++'s TOML_HAS_ATTR(__noinline__) preprocessor check; TOML_CUDA=1 routes
+# toml++ around that line. HIP also maps some CUDA driver calls onto deprecated
+# HIP APIs, so deprecation is downgraded from error (the tree builds with
+# werror=true) to a warning.
+if wheel_variant == 'rocm'
```
edgargabriel: are we using here the wheel_variant as the mechanism to detect whether rocm support has been requested? If yes, I would suggest to follow the pattern that was established with PR #1787 , which introduces in the top-level meson.build file an option `rocm_path` which has to be set for requesting rocm support.

test/unit/utils/libfabric/meson.build
```diff
@@ -14,6 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Hipify sources that include cuda_runtime.h for ROCm/HIP builds
```
edgargabriel: instead of using the hipify approach here, I would recommend that we mimick the pattern established in PR #1787  and #1993 for supporting AMD GPUs in the ucx plugin tests and gtests. The approach is base on having a gpu_utils.h header file that implements the requested functionality for both NVIDIA and AMD GPUs.

edgargabriel (changes requested):

test/unit/utils/libfabric/meson.build
mpashkovskii: Fixed. Added gpu_utils.h for tests

src/utils/libfabric/meson.build
mpashkovskii: Moved platform detection to the root level `meson.build`

src/plugins/libfabric/libfabric_backend.cpp
mpashkovskii: Updated

src/plugins/libfabric/meson.build
mpashkovskii: > How general is /opt/rocm as a path or the env var?

It is quite general and I think it is ok to break it if it is false. Moved that logic to the root `meson.build`

src/utils/libfabric/meson.build
mpashkovskii: Moved to the root `meson.build`

mpashkovskii:
All the comments are addressed. @edgargabriel @aranadive @ryanhankins @fengjica, LMK if I missed something

test/unit/utils/libfabric/meson.build
```diff
@@ -14,12 +14,18 @@
...
 # limitations under the License.
 
+# GPU support: CUDA normally, or HIP/ROCm when wheel_variant=rocm. The test uses
+# the gpu_utils.h abstraction (HAVE_CUDA / HAVE_ROCM), so no hipify translation is
+# needed here -- the HIP headers are consumed directly via rocm_dep.
 libfabric_utils_dep = [ libfabric_dep, nixl_common_deps, serdes_interface ]
 libfabric_test_cpp_args = []
 
-if cuda_dep.found()
+if wheel_variant == 'rocm'
```
edgargabriel: my understanding was that we are not using the wheel_variant as a discrimenator on whether we want rocm support or not. What is the library is compiled without setting wheel_variant=rocm, but rocm_path=/opt/rocm has been provided? 

The right approach is:
```
# ROCm detection (mutually exclusive with CUDA; default off)
rocm_path = get_option('rocm_path')
if rocm_path != ''
...
```
see e.g. https://github.com/ai-dynamo/nixl/blob/main/test/unit/plugins/ucx/meson.build#L16

meson.build
```diff
@@ -332,6 +332,50 @@ if wheel_variant != ''
...
+if wheel_variant == 'rocm'
+    rocm_path = run_command('bash', '-c', 'echo -n "$ROCM_PATH"', check: false).stdout().strip()
+    if rocm_path == ''
+        rocm_path = '/opt/rocm'
+    endif
+    rocm_dep = declare_dependency(
+        compile_args : [
+            '-I' + rocm_path + '/include',
+            '-D__HIP_PLATFORM_AMD__',
+            '-DTOML_CUDA=1',
```
edgargabriel, line 374: what is TOML_CUDA required for?

test/unit/utils/libfabric/meson.build
mpashkovskii: Switched to rocm_path

meson.build
mpashkovskii: Without it I got: `preprocessor.hpp:414: error: macro "__has_attribute" requires an identifier    #if TOML_GCC || TOML_CLANG || TOML_HAS_ATTR(__noinline__)`

Chain: libfabric_backend.cpp → common/configuration.h → toml++

hipify injects <hip/hip_runtime.h>, which `#defines __noinline__;` that macro-expands inside toml++'s `TOML_HAS_ATTR(__noinline__)` and breaks it. `TOML_CUDA=1` forces toml++ down its CUDA branch (TOML_ATTR(noinline)), avoiding the token.

edgargabriel (approved):
LGTM from the high level perspective.

mpashkovskii:
@fengjica @ryanhankins do you still see any issues?

ryanhankins (approved):

ryanhankins:
Looks ok.

src/utils/libfabric/libfabric_topology.h
```diff
@@ -145,6 +146,10 @@ class nixlLibfabricTopology {
     isNvidiaAccel(hwloc_obj_t obj) const;
     bool
     isNeuronAccel(hwloc_obj_t obj) const;
+    bool
+    isAmdAccel(hwloc_obj_t obj) const;
+    bool
+    isEfaDevice(hwloc_obj_t obj) const;
```
rongbingzhou: Nit: Seems this newly added function is not implemented and used. Recommend to remove it.

src/utils/libfabric/libfabric_topology.cpp
rongbingzhou: Thanks for fixing the `num_aws_accel` semantics above. With that change, we should probably change the `num_aws_accel` in this log to `num_nvidia_accel`.

src/utils/libfabric/libfabric_topology.h
mpashkovskii: Removed

src/utils/libfabric/libfabric_topology.cpp
mpashkovskii: Fixed

rongbingzhou (dismissed):

edgargabriel:
I wanted to check wether this PR is good to go / be merged?

mpashkovskii:
@fengjica is it ok to merge it?

rongbingzhou:
Hi @mpashkovskii , @fengjica is OOTO and can't review it for now. Our team agrees to merge the PR and I have approved it.

mpashkovskii:
@rongbingzhou for some reason I don't see merge button. What is the process?

rongbingzhou:
We need Nvidia team to manually trigger the CI. Once CI passes, you should be able to merge the PR.

@aranadive @brminich @ovidiusm Could you please help with the CI? Thanks!

mpashkovskii:
@aranadive @brminich @ovidiusm, could you help here?

brminich (dismissed):

aranadive:
@mpashkovskii Can you check the CI failure text posted above?

mpashkovskii:
Thanks @brminich and @aranadive for running CI. I'll check the CI failures

mpashkovskii:
Fixed the libfabric build failure flagged by the CI triage agent.

**Root cause:** the ROCr registration path set `mr_attr.device.rocr`, but the `rocr` member of `struct fi_mr_attr`'s `device` union was only introduced in libfabric **v2.3**. Every NIXL CI image builds against libfabric **v1.21.0**, where that member does not exist, so the compile aborted with `'union fi_mr_attr::<unnamed>' has no member named 'rocr'` across all build jobs. It compiled locally only because the dev environment used libfabric v2.4.x.

**Fix:** guarded the assignment with a compile-time version check:

```cpp
#if FI_VERSION(FI_MAJOR_VERSION, FI_MINOR_VERSION) >= FI_VERSION(2, 3)
    mr_attr.device.rocr = device_id;
#endif
```

On libfabric < 2.3 the assignment is compiled out and the zero-initialized union is used (the EFA provider ignores the ROCr device handle there); on v2.3+ it is set as before, so behavior is unchanged where ROCr is supported.

@aranadive @brminich could you check?

rongbingzhou (approved):

aranadive (approved):

PR #1751, opened 2026-06-10
nixl_ep: Optimize low-latency NVLink path

itayalroy:
This PR adds a set of optimizations for NIXL EP low-latency NVLink path.

Each optimization is in a separate commit, with the explanation included in the commit message.

On the TRT-LLM MoE communication benchmark ([bench_moe_comm.py](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/microbenchmarks/bench_moe_comm.py)) over single-node NVLink, 8xH100, these changes bring NIXL EP dispatch to parity with DeepEP low latency and reduce the remaining combine gaps.

<details>
<summary>Full dispatch benchmark table</summary>

| Batch | deepep low latency | nixl ep main | main vs deepep | nixl ep PR 1749 | PR vs deepep |
|---:|---:|---:|---:|---:|---:|
| 16  | 32.06 µs  | 38.34 µs  | 0.836x | 32.55 µs  | **0.985x** |
| 32  | 37.61 µs  | 45.49 µs  | 0.827x | 37.98 µs  | **0.990x** |
| 64  | 49.38 µs  | 61.12 µs  | 0.808x | 49.69 µs  | **0.994x** |
| 128 | 75.21 µs  | 92.18 µs  | 0.816x | 74.76 µs  | **1.006x** |
| 256 | 128.85 µs | 147.32 µs | 0.875x | 128.20 µs | **1.005x** |
| 512 | 229.28 µs | 248.60 µs | 0.922x | 228.14 µs | **1.005x** |

</details>

<details>
<summary>Combine active rank mask optimization</summary>

| Batch | disable mask / speed-of-light | current: load acq | current vs speed-of-light | this PR: plain load | PR vs speed-of-light |
|---:|---:|---:|---:|---:|---:|
| 16   | 25.03 µs  | 27.24 µs  | 0.919x | 26.26 µs  | **0.953x** |
| 32   | 31.74 µs  | 33.79 µs  | 0.939x | 33.06 µs  | **0.960x** |
| 64   | 45.65 µs  | 48.48 µs  | 0.942x | 46.94 µs  | **0.973x** |
| 128  | 73.58 µs  | 76.75 µs  | 0.959x | 74.37 µs  | **0.989x** |
| 256  | 137.18 µs | 141.01 µs | 0.973x | 137.34 µs | **0.999x** |
| 512  | 258.57 µs | 266.23 µs | 0.971x | 258.68 µs | **1.000x** |

</details>

**Note that there is no visible performance impact on NIXL EP’s [elastic.py](https://github.com/ai-dynamo/nixl/blob/main/examples/device/ep/tests/elastic/elastic.py).**

Follow-up PRs will continue optimizing NIXL EP combine, with the goal of matching and then exceeding DeepEP low-latency performance.

## Summary by CodeRabbit

* **New Features**
  * GPU-side P2P pointer caching to improve peer-to-peer GPU memory access performance.

* **Chores**
  * Reduced default index width from 64-bit to 32-bit.
  * Added a CUDA compile flag to alter PTX instruction selection for targeted memory operations.
  * Improved allocation, initialization, population, and cleanup of the GPU pointer cache to ensure consistent behavior.

### DIFF
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -1123,22 +1123,18 @@ void update_mask_buffer(int* mask_buffer_ptr, int rank, bool mask, cudaStream_t 
 }
 
 
-template <int kNumThreads> __launch_bounds__(kNumThreads, 1)
-__global__ void cache_p2p_ptrs_kernel(nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr) {
-    const auto thread_id = static_cast<int>(threadIdx.x);
+__global__ void cache_p2p_ptr_kernel(nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr,
+                                     int rank_id) {
     auto nixl_ctx = *nixl_ctx_ptr;
-
-    for (int rank_id = thread_id; rank_id < nixl_ctx.max_num_ranks; rank_id += kNumThreads) {
-        nixl_ctx.p2p_ptrs[rank_id] =
-            nixl_ctx.remote_mvh == nullptr ? nullptr : nixlGetPtr(nixl_ctx.remote_mvh, rank_id);
-    }
+    nixl_ctx.p2p_ptrs[rank_id] =
+        nixl_ctx.remote_mvh == nullptr ? nullptr : nixlGetPtr(nixl_ctx.remote_mvh, rank_id);
 }
 
-void cache_p2p_ptrs(gpu_nixl_ctx* nixl_ctx, cudaStream_t stream) {
+void cache_p2p_ptr(gpu_nixl_ctx* nixl_ctx, int rank_id, cudaStream_t stream) {
     constexpr int num_sms = 1;
-    constexpr int kNumThreads = 128;
+    constexpr int kNumThreads = 1;
     SETUP_LAUNCH_CONFIG(num_sms, kNumThreads, stream);
-    LAUNCH_KERNEL(&cfg, cache_p2p_ptrs_kernel<kNumThreads>, nixl_ctx);
+    LAUNCH_KERNEL(&cfg, cache_p2p_ptr_kernel, nixl_ctx, rank_id);
 }
 
 
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -43,15 +43,24 @@ __device__ inline void* p2p_ptr_get(gpu_nixl_ctx& ctx, uint64_t dst_ptr, int dst
 
 namespace ep_kernels {
 
-template<bool use_warp_sync = false>
+template<bool use_warp_sync = false, bool use_acquire = true>
 __forceinline__ __device__ bool is_rank_masked(int* mask_buffer_ptr, int rank) {
     if (mask_buffer_ptr == nullptr) {
         return false;
     }
     if constexpr (use_warp_sync) {
-        return __shfl_sync(0xffffffff, ld_acquire_global(mask_buffer_ptr + rank), 0) != 0;
+        if constexpr (use_acquire) {
+            return __shfl_sync(
+                       0xffffffff, ld_acquire_global(mask_buffer_ptr + rank), 0) != 0;
+        } else {
+            return __shfl_sync(0xffffffff, mask_buffer_ptr[rank], 0) != 0;
+        }
     } else {
-        return ld_acquire_global(mask_buffer_ptr + rank) != 0;
+        if constexpr (use_acquire) {
+            return ld_acquire_global(mask_buffer_ptr + rank) != 0;
+        } else {
+            return mask_buffer_ptr[rank] != 0;
+        }
     }
 }
 
@@ -918,7 +927,8 @@ COMBINE_RECV:
                     if (topk_idx_reg < 0)
                         continue;
                     EP_DEVICE_ASSERT(topk_idx_reg < active_expert_bound);
-                    if (is_rank_masked(mask_buffer_ptr, topk_idx_reg / num_local_experts))
+                    if (is_rank_masked<false, false>(
+                            mask_buffer_ptr, topk_idx_reg / num_local_experts))
                         continue;
 
                     mbarrier_wait<true>(empty_barriers[stage_idx], tma_phase, stage_idx);
@@ -959,7 +969,8 @@ COMBINE_RECV:
                     if (topk_idx_reg < 0)
                         continue;
                     EP_DEVICE_ASSERT(topk_idx_reg < active_expert_bound);
-                    if (is_rank_masked(mask_buffer_ptr, topk_idx_reg / num_local_experts))
+                    if (is_rank_masked<false, false>(
+                            mask_buffer_ptr, topk_idx_reg / num_local_experts))
                         continue;
                     const auto& topk_weight = __shfl_sync(0xffffffff, topk_weights_by_lane, i);
 
--- a/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
+++ b/examples/device/ep/csrc/kernels/nixl_ep_ll.cu
@@ -35,7 +35,7 @@ namespace nixl_ep {
 __device__ inline void* p2p_ptr_get(gpu_nixl_ctx& ctx, uint64_t dst_ptr, int dst_rank) {
     if (dst_rank == ctx.rank) return (void*) dst_ptr;
 
-    void *remote_ptr = nixlGetPtr(ctx.remote_mvh, dst_rank);
+    void *remote_ptr = ctx.p2p_ptrs[dst_rank];
     if (remote_ptr == nullptr) return nullptr;
 
     return (void*) ((uint64_t) remote_ptr + ctx.offset_get(dst_ptr));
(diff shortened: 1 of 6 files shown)


examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1370,6 +1378,10 @@ void Buffer::_nixl_ep_init(void) {
 
 void Buffer::_nixl_ep_destroy(void) {
     _nixl_ep_memory_views_destroy();
+    if (p2p_ptrs != nullptr) {
```
rakhmets: Not sure if it can be `nullptr` here. `_nixl_ep_destroy` is called by `destroy()` only if `destroyed` is `false`.

P.S. `gpu_ctx_ptr` seems to have the same extra check in this function.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -1112,6 +1112,29 @@ void update_mask_buffer(int* mask_buffer_ptr, int rank, bool mask, cudaStream_t
 }
 
 
+template <int kNumThreads> __launch_bounds__(kNumThreads, 1)
+__global__ void cache_p2p_ptrs_kernel(nixl_ep::gpu_nixl_ctx* nixl_ctx_ptr) {
+    const auto thread_id = static_cast<int>(threadIdx.x);
+    auto nixl_ctx = *nixl_ctx_ptr;
```
rakhmets, line 1128: Why is this copying needed?

examples/device/ep/csrc/nixl_ep.hpp
```diff
@@ -158,6 +158,7 @@ struct Buffer {
     NixlPeerInfo my_peer_info;
     nixl_ep::gpu_nixl_ctx gpu_ctx;
     nixl_ep::gpu_nixl_ctx* gpu_ctx_ptr = nullptr;
+    void** p2p_ptrs = nullptr;
```
rakhmets: Please remove this field, as having the field in `struct gpu_nixl_ctx` is enough, and `Buffer` class doesn't need this field.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: To avoid reading gpu_ctx_ptr from global memory multiple times on each iteration in the loop below, sometimes explicitly reading to a local variable helps with CUDA perf, I don't really see any reason to not do it

rakhmets: I see. If it improves performance, then it's needed.

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -1112,6 +1112,29 @@ void update_mask_buffer(int* mask_buffer_ptr, int rank, bool mask, cudaStream_t
...
+
+    for (int rank_id = thread_id; rank_id < nixl_ctx.max_num_ranks; rank_id += kNumThreads) {
+        if (rank_id == nixl_ctx.rank) {
+            nixl_ctx.p2p_ptrs[rank_id] = nixl_ctx.rdma_buffer_ptr;
```
rakhmets: It seems that this item is not intended to be used. Should it be `nullptr`?

examples/device/ep/meson.build
```diff
@@ -101,7 +101,7 @@ nixl_ep_cuda_args = [
     '-Xcompiler', '-Wno-attributes',
 ]
 
-topk_idx_bits = meson.get_external_property('topk_idx_bits', '64')
+topk_idx_bits = meson.get_external_property('topk_idx_bits', '32')
```
guy-ealey-morag, line 104: Should this be changed in `examples/device/ep/csrc/kernels/configs.cuh:80` as well?

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -918,7 +918,7 @@ COMBINE_RECV:
                     if (topk_idx_reg < 0)
                         continue;
                     EP_DEVICE_ASSERT(topk_idx_reg < active_expert_bound);
-                    if (is_rank_masked(mask_buffer_ptr, topk_idx_reg / num_local_experts))
+                    if (mask_buffer_ptr[topk_idx_reg / num_local_experts] != 0)
```
guy-ealey-morag: Can we put this in a function to improve readability?
You can either add a new function or use the same template with a new parameter like:

```cpp
template<bool use_warp_sync = false, bool use_acquire = true>
__forceinline__ __device__ bool is_rank_masked(int* mask_buffer_ptr, int rank) {
```

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
guy-ealey-morag: I think it's worth adding a comment in the code

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
ebarilan: why? also when the target expert is local there is still a copy

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
```diff
@@ -1112,6 +1112,29 @@ void update_mask_buffer(int* mask_buffer_ptr, int rank, bool mask, cudaStream_t
...
+                nixl_ctx.remote_mvh == nullptr ? nullptr : nixlGetPtr(nixl_ctx.remote_mvh, rank_id);
+        }
+    }
+}
+
+void cache_p2p_ptrs(gpu_nixl_ctx* nixl_ctx, cudaStream_t stream) {
+    constexpr int num_sms = 1;
+    constexpr int kNumThreads = 128;
+    SETUP_LAUNCH_CONFIG(num_sms, kNumThreads, stream);
+    LAUNCH_KERNEL(&cfg, cache_p2p_ptrs_kernel<kNumThreads>, nixl_ctx);
```
ebarilan: is there a value define a template? if it is const?

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
ebarilanM: the part that is not clear is how topk_idx_reg/num_local_experts -> global_rank

examples/device/ep/csrc/kernels/utils.cuh
```diff
@@ -179,7 +179,7 @@ __device__  __forceinline__ int64_t ld_volatile_global(const uint64_t *ptr) {
 #ifndef DISABLE_AGGRESSIVE_PTX_INSTRS
 #define LD_NC_FUNC "ld.global.nc.L1::no_allocate.L2::256B"
 #else
-#define LD_NC_FUNC "ld.volatile.global"
+#define LD_NC_FUNC "ld.volatile.global.L2::256B"
 #endif
```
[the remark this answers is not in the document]
ebarilanM: build target is sm90+

examples/device/ep/csrc/nixl_ep.hpp
ebarilanM: indeed seems redundant

ebarilanM (dismissed):

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: `topk_idx_reg` is not the best name (this is DeepEP code), it is not a idx, it is an actual topk entry (== global expert), and we divide it by `num_local_experts` to get the rank that owns it

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: It can if we create a buffer and immediately destroy it w/o calling `update_memory_buffers`

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: It's a common pattern in this file, all kernels work with configurable num_threads, I think it doesn't really add any complexity and allows us to tune num threads later if we see fit

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
guy-ealey-morag: @itayalroy Can you add a short comment explaining why the value is being copied?

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
rakhmets: `nixl_ctx.p2p_ptrs[rank_id]` replaces `nixlGetPtr(ctx.remote_mvh, dst_rank)` in `p2p_ptr_get`.
We don't use `nixl_ctx.p2p_ptrs[nixl_ctx.rank]` in `p2p_ptr_get`, as there is a condition in the beginning of the function.

itayalroy: Removed

examples/device/ep/csrc/kernels/nixl_ep_ll.cu
itayalroy: @guy-ealey-morag I think it's redundant, this is pretty standard in CUDA and we do it in many places (check out dispatch/combine, done there as well), also it might give a wrong impression that this control flow (connect_ranks/disconnect_ranks) is very optimized, which isn't really the case, for example for highly optimized control flow this p2p cache kernel should not be a separate kernel, we should unify it with the kernels that unmask the newly connected ranks, etc etc

itayalroy:
Changed P2P pointer caching to launch one small kernel per newly connected rank.

`nixlGetPtr` crashes on gap ranks, so this avoids calling it for unconnected ranks. Since this only runs on the connection path, and a kernel launch is ~2-5 us, this is fine for now and alignes with how we update other GPU state like the active mask.

The alternative is adding and populating a `gpu_connected_ranks` array, which is arguably worse, and also pretty messy.

I think we should consider making UCX return `nullptr` for gap ranks instead of crashing, but shouldn't block this PR

ebarilanM (approved):
LGTM

Issue #1754, opened 2026-06-10
UCX: PyTorch expandable_segments / CUDA VMM memory silently falls back to slow path (no cuda_ipc)


xyuzh:
## Summary

When CUDA memory registered with the UCX backend was allocated by PyTorch with `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` (i.e. CUDA VMM / `cuMemCreate` memory), registration **succeeds** but READ transfers silently fall back to UCX's software-emulated RMA over TCP at **~0.3 GB/s**, instead of the `cuda_ipc` zero-copy path at **~200 GB/s** (same-node H100/NVLink). A single env var produces a silent ~600x slowdown with no warning anywhere, and the data remains correct — which makes this brutal to diagnose in production.

We hit this through Ray's RDT integration (`ray.put(_tensor_transport="nixl")`, Ray 2.55.1 hardcodes the UCX backend) while implementing RL weight sync (trainer → inference engine) in [miles](https://github.com/radixark/miles). `expandable_segments:True` is a very common setting in large-model training (fragmentation mitigation), so this trap is broadly armed for the RL/LLM audience.

## Environment

- nixl 1.2.0 (`nixl-cu12` pip wheel, bundled UCX)
- Ray 2.55.1 (`ray.experimental` RDT, NIXL transport), torch 2.11, CUDA 12.9
- 8x H100 (NVLink), single node, AWS EKS pod

## Reproduction

Two single-GPU Ray actors, source registers a 512 MiB buffer, destination does a one-sided READ (mirrors Ray's `nixl_tensor_transport`). Identical code, one env var difference on the source actor:

| Source allocator | Steady-state READ | UCX proto (`UCX_PROTO_INFO=y`) |
|---|---|---|
| default (`cudaMalloc`) | 179-201 GB/s | `zero-copy \| cuda_ipc/cuda` |
| `expandable_segments:True` | **0.29-0.32 GB/s** | `software emulation \| tcp/eth0` |

Controls that did NOT matter (all 170-230 GB/s): `CUDA_VISIBLE_DEVICES` masking, `UCX_TLS`/`UCX_RNDV_THRESH` overrides, `CUDA_DEVICE_MAX_CONNECTIONS=1`, NCCL initialized in the source process. We can share the full repro scripts (self-contained, ~150 lines each) on request.

## Root cause (as far as we traced it)

[...]

PR #1755, opened 2026-06-10
ucx: warn when registered CUDA memory is not legacy IPC-capable

xyuzh:
## What?

Make the UCX plugin warn when it registers a CUDA region that cannot use the
`cuda_ipc` fast path, and document the limitation.

- `ucx_utils.cpp`: in `nixlUcxContext::memReg`, when registering `VRAM_SEG`,
  query `CU_POINTER_ATTRIBUTE_IS_LEGACY_CUDA_IPC_CAPABLE` via
  `cuPointerGetAttribute`. If the region is not legacy-CUDA-IPC-capable, emit a
  one-time `NIXL_WARN` naming the region and the consequence.
- `meson.build`: link `cuda_dep` into the UCX plugin and define `HAVE_CUDA`,
  gated on `cuda_dep.found()` so non-CUDA builds are unaffected. The new code is
  entirely under `#ifdef HAVE_CUDA`.
- `src/plugins/ucx/README.md` (new): document the limitation, the
  `UCX_PROTO_INFO=y` diagnostic, and the workaround. The UCX plugin was the only
  major plugin without a README.

## Why?

Fixes the silent-degradation half of #1754.

CUDA buffers allocated via the CUDA VMM API (`cuMemCreate`/`cuMemMap`) — what
PyTorch produces under `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` —
cannot be exported through the legacy `cuIpcGetMemHandle` mechanism that UCX's
`cuda_ipc` transport relies on. Registration still succeeds, but an intra-node
transfer that would otherwise use `cuda_ipc`, with no RDMA NIC available as an
alternative, silently falls back to UCX's software-emulated path over TCP. The
data stays correct while throughput collapses (we measured ~200 GB/s on
`cuda_ipc` vs ~0.3 GB/s on the emulated path, same-node H100/NVLink).
`expandable_segments:True` is a very common large-model-training setting, so
this trap is broadly armed — and with no log anywhere it cost us days to
diagnose. A one-line warning at registration time turns a silent ~600x
slowdown into something immediately diagnosable.

This is request #1 (loud failure) + request #3 (documentation) from #1754.
Request #2 (intra-node POSIX-FD VMM IPC support) is a larger feature I'd like to
discuss in the issue first, per CONTRIBUTING.

## How?

`CU_POINTER_ATTRIBUTE_IS_LEGACY_CUDA_IPC_CAPABLE` returns 0 for VMM/expandable
allocations and 1 for plain `cudaMalloc` allocations, so it precisely
distinguishes the two cases at registration time. The warning:

- fires only for `UCS_MEMORY_TYPE_CUDA` regions that report not-IPC-capable;
- is emitted at most once per process (`std::atomic<bool>` guard) to avoid spam;
- is phrased conditionally ("intra-node transfers with no RDMA NIC available
  fall back...") because VMM memory still registers and transfers fine over an
  RDMA NIC when one is present — the slow path only bites when `cuda_ipc` is the
  only fast lane;
- silently skips if `cuPointerGetAttribute` fails, so it can never break
  registration.

Detection example (UCX's `UCX_PROTO_INFO=y`): a healthy intra-node GPU READ
reports `zero-copy | cuda_ipc/cuda`; the degraded fallback reports
`software emulation | tcp/...`.

### Testing note

I don't have a CUDA/GPU machine to build or run on, so the `HAVE_CUDA` path has
not been compiled or exercised locally — I'm relying on CI for the build and
would appreciate a maintainer sanity-check on a GPU box. The change follows the
existing CUDA-driver-API usage pattern in the libfabric plugin
(`cuPointerGetAttributes`, `#ifdef HAVE_CUDA`, `cuda_dep` wiring). I'm also happy
to share the self-contained ~150-line repro scripts from #1754 for validating
the warning fires on `expandable_segments:True` memory and stays silent on
`cudaMalloc` memory.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive UCX plugin README covering supported memory types, transport selection, CUDA-related limitations, troubleshooting steps, diagnostic environment variables, expected log-path differences, and a recommended workaround for affected allocations.

* **Improvements**
  * Build now enables CUDA support for the UCX plugin when available.
  * Emits a one-time per-process warning when non-legacy CUDA allocations are registered to surface cases that may fall back to slower TCP paths.

### DIFF
--- a/src/plugins/ucx/meson.build
+++ b/src/plugins/ucx/meson.build
@@ -32,6 +32,14 @@ ucx_backend_dependencies = [asio_dep,
                             thread_dep,
                             ucx_dep]
 
+ucx_backend_cpp_args = []
+
+# CUDA driver API, used to warn when registered VRAM cannot take the cuda_ipc fast path.
+if cuda_dep.found()
+    ucx_backend_dependencies += [cuda_dep]
+    ucx_backend_cpp_args += ['-DHAVE_CUDA']
+endif
+
 ucx_backend_include_directories = [nixl_inc_dirs,
                                    utils_inc_dirs]
 
@@ -41,6 +49,7 @@ if 'UCX' in static_plugins
                dependencies: ucx_backend_dependencies,
                include_directories: ucx_backend_include_directories,
                install: false,
+               cpp_args : ucx_backend_cpp_args,
                name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
 else
     ucx_backend_lib = shared_library('UCX',
@@ -48,7 +57,7 @@ else
                dependencies: ucx_backend_dependencies,
                include_directories: ucx_backend_include_directories,
                install: true,
-               cpp_args : ['-fPIC'],
+               cpp_args : ucx_backend_cpp_args + ['-fPIC'],
                name_prefix: 'libplugin_',  # Custom prefix for plugin libraries
                install_dir: plugin_install_dir,
                build_rpath: get_option('ucx_path') != '' ? get_option('ucx_path') + '/lib' : '',
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -18,6 +18,7 @@
 #include "ucx_utils.h"
 
 #include <algorithm>
+#include <atomic>
 #include <cstring>
 #include <exception>
 #include <stdexcept>
@@ -30,6 +31,10 @@
 #include "common/nixl_log.h"
 #include "config.h"
 #include "serdes/serdes.h"
+
+#ifdef HAVE_CUDA
+#include <cuda.h>
+#endif
 
 [[nodiscard]] nixl_b_params_t
 get_ucx_backend_common_options() {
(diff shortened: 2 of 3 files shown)


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -543,6 +548,40 @@ nixlUcxWorker::connect(void *addr, std::size_t size) {
  * Memory management
  * =========================================== */
 
+#ifdef HAVE_CUDA
+namespace {
+// CUDA VMM allocations (cuMemCreate/cuMemMap, e.g. PyTorch's
+// expandable_segments:True) cannot be exported through the legacy
+// cuIpcGetMemHandle mechanism that UCX's cuda_ipc transport relies on, so
```
rakhmets, line 555: UCX's `cuda_ipc` transport support different types of memory allocation. It uses `cuIpc` API for memory allocated using `cuMemAlloc`/`cudaMalloc`, and `cuMemImport`/`cuMemExport` for memory allocated using VMM API (`cuMemCreate` + `CU_MEM_HANDLE_TYPE_FABRIC`).
The missing part (`cuMemCreate` + `CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR`) should be covered by https://github.com/openucx/ucx/pull/11183.

src/plugins/ucx/meson.build
```diff
@@ -32,6 +32,14 @@ ucx_backend_dependencies = [asio_dep,
                             thread_dep,
                             ucx_dep]
 
+ucx_backend_cpp_args = []
+
+# CUDA driver API, used to warn when registered VRAM cannot take the cuda_ipc fast path.
+if cuda_dep.found()
+    ucx_backend_dependencies += [cuda_dep]
+    ucx_backend_cpp_args += ['-DHAVE_CUDA']
+endif
```
rakhmets, line 41: NIXL UCX plugin doesn't use CUDA directly, and shouldn't depend on CUDA libraries. The only libraries that depend on CUDA in this stack are UCX internal libraries (e.g. `libuct_cuda`).

brminich (changes requested):
as @rakhmets mentioned the real fix is in openucx/ucx#11183
we do not want to warn for all types of non-legacy allocated memory

brminich:
@xyuzh can you pls close this PR?

PR #1756, opened 2026-06-11
Stride (compressed) descriptors for prep+make

iyastreb:
## What?
This PR introduces stride descriptors to keep prepared handles in a compressed form.

## Why?
Prepared handles are used in prep+make API, and such a handle is supposed to handle descriptors for the entire model. This may consume a lot of RAM.
For example, in DeepSeek-R1-Distill-Qwen-32B NIXL creates **1387520 descs of size 32B = ~44MB**
With compressed optimization in place this handle is represented by **128 strides of 56B = 7KB**

The aim is to reduce memory consumption and make lookup cache friendly
Another improvements:
- Join validation and merge loops
- Speed up merge algorithm 
- Reserve descs instead of resize

## Testing
**nixlbench:**
```
UCX_TLS=^cuda_ipc nixlbench --initiator_seg_type=VRAM --target_seg_type=VRAM --start_block_size=65536 --max_block_size=65536 --op_type=WRITE --start_batch_size=256 --max_batch_size=256 --prepared_xfer=1 --num_iter=10000
        B/W (GB/Sec)   Avg Lat. (us)  Avg Prep (us)  P99 Prep (us)  Avg Post (us)  P99 Post (us)  Avg Tx (us)    P99 Tx (us)
main:   18.143680      3.6            34.0           34.0           1.5            2.0            923.2          1123.0
stride: 18.149539      3.6            21.0           21.0           1.4            2.0            922.9          1125.0

UCX_TLS=^cuda_ipc nixlbench --initiator_seg_type=VRAM --target_seg_type=VRAM --start_block_size=65536 --max_block_size=65536 --op_type=WRITE --start_batch_size=256 --max_batch_size=256 --prepared_xfer=1 --num_iter=10000 --recreate_xfer=1
        B/W (GB/Sec)   Avg Lat. (us)  Avg Prep (us)  P99 Prep (us)  Avg Post (us)  P99 Post (us)  Avg Tx (us)    P99 Tx (us)
main:   17.938409      3.7            10.3           11.0           1.5            2.0            923.1          1129.0
stride: 17.994225      3.6            7.3            8.0            1.5            2.0            923.2          1127.0
```
**SGLANG**

Merge micro-benchmark on num_prompts=128
merge (before→after) | main avg µs | stride avg µs | speedup
-- | -- | -- | --
128 → 128 (no merge) | 3.3 | 2.7 | 1.2×
256 → 256 (no merge) | 7.4 | 5.2 | 1.4×
896 → 256 | 14.8 | 6.1 | 2.4×
1536 → 256 | 23.7 | 7.7 | 3.1×
1536 → 128 | 24.9 | 4.6 | 5.4×
8576 → 256 | 80.3 | 15.2 | 5.3×

Overall make_prepped_xfer call on num_prompts=128
make_prepped_xfer time (µs) | main | stride
-- | -- | --
median | 57.6 | 48.0
mean | 77.0 | 62.3
p99 | 325.4 | 255.9
max | 934.0 | 649.2

## Summary by CodeRabbit

* **New Features**
  * Added stride-based descriptors and response grouping to represent contiguous memory runs more compactly.

* **Refactor**
  * Descriptor preparation/management switched to stride-based lists.
  * Transfer-request assembly now merges compatible runs, lowering descriptor counts and recalculating totals.

* **Bug Fixes**
  * Earlier and stricter index validation; improved concurrency checks and safer metadata existence handling.


src/core/nixl_agent.cpp
```diff
@@ -674,13 +677,15 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+    if ((desc_count == 0) || (remote_indices.size() == 0) ||
+        (desc_count != (int)remote_indices.size())) {
+        NIXL_ERROR_FUNC << "different number of indices for local (" << desc_count << "), remote ("
+                        << remote_indices.size() << ")";
+        return NIXL_ERR_INVALID_PARAM;
+    }
+
+    if (extra_params && extra_params->skipDescMerge) {
+        NIXL_DEBUG << "skipDescMerge is deprecated and will be removed in a future release";
     }
```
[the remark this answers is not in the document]
iyastreb: re-enabled skipDescMerge

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+                            << local_indices[i];
+            return NIXL_ERR_INVALID_PARAM;
+        }
+        if (remote_idx >= remote_size) [[unlikely]] {
+            NIXL_ERROR_FUNC << "remote index out of range at index " << i << " with value "
+                            << remote_indices[i];
+            return NIXL_ERR_INVALID_PARAM;
+        }
+
+        const nixlStrideDesc local_stride = local_descs.find(local_idx, local_run_size);
```
ovidiusm, line 787: Could these be const references?

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+        const nixlStrideDesc remote_stride = remote_descs.find(remote_idx, remote_run_size);
```
ovidiusm, line 788: I'm also wondering if we could avoid calling find() every step of the iteration, is it expected to be the same as previous step? Then we can try caching it

iyastreb: Well, with current impl we don't call find on each iteration. The merge is mostly done by internal while loop, so that when we merge for example 4096 indices, this results in 128 find() and 128 resulting merges descs. I don't think we can optimize this more

src/core/nixl_agent.cpp
iyastreb: Yes, they were references in the beginning, but then it turned out that copying out is more efficient, because we advance index quite a lot in the while loop and reading back index[i] is heavier. But this is micro-optimisation anyway

src/core/nixl_agent.cpp
```diff
@@ -613,13 +613,18 @@ nixlAgent::prepXferDlist (const std::string &agent_name,
...
+        int stride_count;
+        nixl_stride_dlist_t dlist(descs.getType());
         if (section.populate(descs, backend, dlist) == NIXL_SUCCESS) {
-            dlists.try_emplace(backend, std::make_unique<nixl_meta_dlist_t>(std::move(dlist)));
+            stride_count = dlist.descCount();
+            dlists.try_emplace(backend, std::make_unique<nixl_stride_dlist_t>(std::move(dlist)));
         }
+
+        NIXL_DEBUG << "backend " << backend->getType() << ": prepared " << descs.descCount()
+                   << " descriptors into " << stride_count << " strides";
```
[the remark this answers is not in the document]
iyastreb: fixed

src/core/nixl_agent.cpp
```suggestion
: initiatorDescs(local_type),
targetDescs(remote_type),
remoteAgent(remote_agent),
backendOp(backend_op) {
  // reserve `initiatorDescs` and `targetDescs` instead of value-initializing
}
```

I found that here when giving `desc_count` as an argument, the constructor uses `resize` which value-initializes the entire vector, but the code is calling `resize(merged_size)` anyway, throwing away most of it.
So it might be better just to reserve here and use `addDesc` for appending.

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+        // Merge only dense strides
+        if (!skip_desc_merge && local_stride.stride == local_stride.len &&
+            local_stride.stride == remote_stride.stride) [[likely]] {
+            const size_t cap =
+                std::min(std::min(local_stride.start_idx + local_stride.count - local_idx,
+                                  remote_stride.start_idx + remote_stride.count - remote_idx),
+                         static_cast<size_t>(desc_count) - i);
+
+            auto local_indices_ptr = reinterpret_cast<const unsigned *>(&local_indices[i]);
+            auto remote_indices_ptr = reinterpret_cast<const unsigned *>(&remote_indices[i]);
```
```suggestion
            auto *local_indices_ptr = reinterpret_cast<const unsigned *>(&local_indices[i]);
            auto *remote_indices_ptr = reinterpret_cast<const unsigned *>(&remote_indices[i]);
```

Make the pointer explicit

src/core/nixl_agent.cpp
guy-ealey-morag: I would leave a comment here that explain that

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+                   remote_indices_ptr[seq_count] == remote_idx + seq_count) {
+                ++seq_count;
+            }
         }
-        NIXL_DEBUG << "reqH descList size down to " << j;
-        handle->initiatorDescs.resize(j);
-        handle->targetDescs.resize(j);
+
+        handle->initiatorDescs[merged_size] = local_stride.getMetaDesc(local_idx, seq_count);
+        handle->targetDescs[merged_size] = remote_stride.getMetaDesc(remote_idx, seq_count);
```
```suggestion
        handle->initiatorDescs.addDesc(local_stride.getMetaDesc(local_idx, seq_count));
        handle->targetDescs.addDesc(remote_stride.getMetaDesc(remote_idx, seq_count));
```

Relevant if you implement https://github.com/ai-dynamo/nixl/pull/1756/changes#r3394940873

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+        ++merged_size;
+        total_bytes += local_stride.len * seq_count;
     }
 
+    NIXL_DEBUG << "merged " << desc_count << " indices into " << merged_size << " descriptors";
+    handle->initiatorDescs.resize(merged_size);
+    handle->targetDescs.resize(merged_size);
```
```suggestion
```

These can be removed if you implement https://github.com/ai-dynamo/nixl/pull/1756/changes#r3394940873

src/core/nixl_agent.cpp
guy-ealey-morag, line 787: I think the `find` can be optimized to start from the previous result instead of starting from `mid` every time (assuming it's ascending between searches, or at least mostly ascending)

src/core/nixl_agent.cpp
```diff
@@ -756,55 +727,90 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
...
+            }
         }
-        NIXL_DEBUG << "reqH descList size down to " << j;
-        handle->initiatorDescs.resize(j);
-        handle->targetDescs.resize(j);
+
+        handle->initiatorDescs[merged_size] = local_stride.getMetaDesc(local_idx, seq_count);
+        handle->targetDescs[merged_size] = remote_stride.getMetaDesc(remote_idx, seq_count);
+        ++merged_size;
+        total_bytes += local_stride.len * seq_count;
```
```suggestion
```

This can be removed from the hot path and can be done retroactively only when telemetry is enabled (around line 818)

```c++
    if (data->telemetryEnabled) {
        size_t total_bytes = 0;
        for (const auto &desc : handle->initiatorDescs) {
            total_bytes += desc.len;
        }
        handle->telemetry.totalBytes = total_bytes;
        handle->telemetry.descCount = handle->initiatorDescs.descCount();
    }
```

src/core/nixl_agent.cpp
iyastreb: That's a very good point, I think you're right and the main problem here is that we **construct** all the elements in advance, and then reassign. I'll change that and remeasure perf

src/core/nixl_agent.cpp
iyastreb: will do

iyastreb: It does not start from the middle, it starts from educated guess which is normally (with uniform layout) always lands on the right stride. (Let me confirm that by running a test counting slow binary path - it must be zero normally).
I think educated guess is better than previous run

src/core/nixl_agent.cpp
iyastreb: ok

src/core/nixl_agent.cpp
iyastreb: I will measure the benefit and will adapt the code

src/core/nixl_agent.cpp
iyastreb: ok

src/core/nixl_agent.cpp
iyastreb: Here I hesitate, because I see that telemetry is always enabled in vLLM, and extra run over descs would add an overhead. I would keep it as is for now

src/core/nixl_agent.cpp
iyastreb: I can confirm that with current sglang implementation the educated guess lands to the right place in 100% of the cases, so that binary search is never triggered

src/core/nixl_agent.cpp
iyastreb: it made a lot of sense, I see now that performance improved because we skip this redundant construction

src/api/cpp/nixl_descriptors.h
```diff
@@ -394,6 +404,17 @@ class nixlDescList {
     virtual void
     addDesc(const T &desc);
 
+    /**
+     * @brief     Emplace a new descriptor into the descriptor list
```
guy-ealey-morag:
```suggestion
     * @brief Emplace a new descriptor into the descriptor list
```

src/api/cpp/nixl_descriptors.h
```diff
@@ -394,6 +404,17 @@ class nixlDescList {
...
+     *
+     * @param args Arguments to construct the descriptor
+     */
+    template<class... Args>
+    inline void
+    emplace(Args &&...args) {
+        descs.emplace_back(std::forward<Args>(args)...);
+    }
```
guy-ealey-morag:
```suggestion
    inline T &
    emplace(Args &&...args) {
        return descs.emplace_back(std::forward<Args>(args)...);
    }
```

Could be useful to return the reference

src/core/nixl_agent.cpp
guy-ealey-morag: Okay, so I suggest a minor simplification:

```c++
        const auto &local_desc =
            handle->initiatorDescs.emplace(local_stride.getMetaDesc(local_idx, seq_count));
        handle->targetDescs.emplace(remote_stride.getMetaDesc(remote_idx, seq_count));
        total_bytes += local_desc.len;
```

(using the returned reference from https://github.com/ai-dynamo/nixl/pull/1756/changes#r3396582741)

src/api/cpp/nixl_descriptors.h
iyastreb: ok

src/core/nixl_agent.cpp
iyastreb: ok

src/api/cpp/nixl_descriptors.h
```diff
@@ -394,6 +404,17 @@ class nixlDescList {
...
     addDesc(const T &desc);
 
+    /**
+     * @brief     Emplace a new descriptor into the descriptor list
+     *
+     * @param args Arguments to construct the descriptor
+     */
+    template<class... Args>
+    inline void
+    emplace(Args &&...args) {
```
guy-ealey-morag, line 415: I like the option to use `emplace` but it is a bit dangerous with inheriting classes.
For example `nixlSecDescList` keep the descriptors sorted and it's enforced in the implementation of `addDesc` (which is `virtual`).
Using `emplace` with a `nixlSecDescList` would break it.

From what I found we can add this to `nixlSecDescList` to prevent the usage of `emplace`:
```c++
    template<class... Args>
    nixlSectionDesc &
    emplace(Args &&...args) = delete;
```

dpressle:
👀 Investigating [`Run Pre-Commit Hooks`](https://github.com/ai-dynamo/nixl/actions/runs/27353782058/job/80822429661)…

dpressle:
🤖 **CI Triage Agent** — `Run Pre-Commit Hooks` · commit `823d26d5`

The diagnosis is clear. Here is the full report:

---

**Summary:** Pre-commit `mixed-line-ending` hook failed on `src/api/cpp/nixl_descriptors.h` due to CRLF (`\r\n`) line endings in lines 410–413 of the `emplace()` method.

**Root cause:** The PR #1756 (commit `[REDACTED:Hex High Entropy String]`) introduced Windows-style CRLF line endings (`\r\n`) into a 4-line block spanning the `emplace()` template method in `src/api/cpp/nixl_descriptors.h` (around lines 410–413). The rest of the repository uses Unix-style LF (`\n`) line endings, so the pre-commit `mixed-line-ending` hook detected the inconsistency and fixed — then failed — the check.

The diff in the CI log shows exactly which lines were affected:
```diff
-    inline T &
-    emplace(Args &&...args) {
-        return descs.emplace_back(std::forward<Args>(args)...);
-    }
+    inline T &       ← these 4 lines had CRLF endings
+    emplace(Args &&...args) {
+        return descs.emplace_back(std::forward<Args>(args)...);
+    }
```

This is almost certainly caused by editing the file on Windows (or in an editor configured with CRLF output) without a `.gitattributes` rule normalizing line endings.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1756 author)

**File:** `src/api/cpp/nixl_descriptors.h` lines 410–413

**Suggested fix:** The PR author should convert the CRLF line endings to LF on the affected lines before re-pushing. This can be done with any of:

```bash
# Option 1: using sed
sed -i 's/\r//' src/api/cpp/nixl_descriptors.h

# Option 2: using dos2unix
dos2unix src/api/cpp/nixl_descriptors.h

# Option 3: let pre-commit fix it locally, then commit
pre-commit run mixed-line-ending --files src/api/cpp/nixl_descriptors.h
git add src/api/cpp/nixl_descriptors.h
git commit --amend --no-edit   # or a new commit
```

To prevent recurrence, consider adding a `.gitattributes` entry to enforce LF normalization for C/C++ header files:
```gitattributes
*.h  text eol=lf
*.cpp text eol=lf
```

**Related:** none



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `83e430ac-b856-42c8-a820-dbb4826d0b81` in the triage console for the audit trail.

src/api/cpp/nixl_descriptors.h
```diff
@@ -394,6 +404,17 @@ class nixlDescList {
     virtual void
     addDesc(const T &desc);
 
+    /**
+     * @brief Emplace a new descriptor into the descriptor list
+     *
+     * @param args Arguments to construct the descriptor
```
guy-ealey-morag, line 410:
```suggestion
     * @param args Arguments to construct the descriptor
     * @return     Reference to the newly emplaced descriptor
```

src/api/cpp/nixl_descriptors.h
iyastreb: ok, makes sense

src/api/cpp/nixl_descriptors.h
iyastreb: done

guy-ealey-morag (dismissed):

src/api/cpp/backend/backend_aux.h
```diff
@@ -134,4 +134,70 @@ operator==(const nixlRemoteMetaDesc &lhs, const nixlRemoteMetaDesc &rhs) {
...
+                   uint64_t dev_id,
+                   nixlBackendMD *metadata,
+                   size_t stride_,
+                   size_t count_)
+        : nixlMetaDesc(addr, len, dev_id, metadata),
+          stride(stride_),
+          count(count_),
+          start_idx(0) {}
+
+    nixlMetaDesc
```
ColinNV: `[[nodiscard]]`

src/api/cpp/backend/backend_aux.h
```diff
@@ -134,4 +134,70 @@ operator==(const nixlRemoteMetaDesc &lhs, const nixlRemoteMetaDesc &rhs) {
...
+            addr + static_cast<uintptr_t>(idx - start_idx) * stride, len * size, devId, metadataP);
+    }
+};
+
+class nixlStrideDescList : public nixlDescList<nixlStrideDesc> {
+public:
+    using nixlDescList<nixlStrideDesc>::nixlDescList;
+
+    size_t
+    flatSize() const {
```
ColinNV, line 174: `[[nodiscard]]` and perhaps a comment to explain the intention.

src/api/cpp/backend/backend_aux.h
```diff
@@ -134,4 +134,70 @@ operator==(const nixlRemoteMetaDesc &lhs, const nixlRemoteMetaDesc &rhs) {
...
+            return 0;
+        }
+        const nixlStrideDesc &last = descs.back();
+        return last.start_idx + last.count;
+    }
+
+    const nixlStrideDesc &
+    find(size_t flat_idx, size_t run_size) const noexcept {
+        // Educated first probe: assume uniform runs
+        size_t mid = flat_idx / run_size;
```
ColinNV, line 188: Guard against `run_size == 0`?

src/api/cpp/nixl_descriptors.h
```diff
@@ -394,6 +404,18 @@ class nixlDescList {
...
+     * @return     Reference to the newly emplaced descriptor
+     */
+    template<class... Args>
+    inline T &
```
ColinNV: `inline` is not needed.

src/api/cpp/backend/backend_aux.h
iyastreb: done

src/api/cpp/backend/backend_aux.h
iyastreb: ok

src/api/cpp/nixl_descriptors.h
iyastreb: removed

src/api/cpp/backend/backend_aux.h
iyastreb: Well, by design it cannot be 0, because we don't allow 0 length desc list
But I can add an assertion

src/api/cpp/backend/backend_aux.h
```diff
@@ -134,4 +135,76 @@ operator==(const nixlRemoteMetaDesc &lhs, const nixlRemoteMetaDesc &rhs) {
...
+        const nixlStrideDesc &last = descs.back();
+        return last.start_idx + last.count;
+    }
+
+    // Returns the run covering the given flat block index, using an educated probe
+    // with a binary-search fallback.
+    const nixlStrideDesc &
+    find(size_t flat_idx, size_t run_size) const noexcept {
+        // Educated first probe: assume uniform runs
+        assert(run_size > 0);
```
guy-ealey-morag:
```suggestion
        NIXL_ASSERT(run_size > 0);
```

or `NIXL_ASSERT_ALWAYS`

src/api/cpp/backend/backend_aux.h
```diff
@@ -134,4 +135,76 @@ operator==(const nixlRemoteMetaDesc &lhs, const nixlRemoteMetaDesc &rhs) {
...
+        }
+        const nixlStrideDesc &last = descs.back();
+        return last.start_idx + last.count;
+    }
+
+    // Returns the run covering the given flat block index, using an educated probe
+    // with a binary-search fallback.
+    const nixlStrideDesc &
+    find(size_t flat_idx, size_t run_size) const noexcept {
+        // Educated first probe: assume uniform runs
```
guy-ealey-morag, line 187: I'd move the comment under the assert

iyastreb: ok

src/api/cpp/backend/backend_aux.h
iyastreb: yes, NIXL_ASSERT

guy-ealey-morag (approved):

brminich (changes requested):
@iyastreb 
is there an ABI break?

iyastreb:
> @iyastreb is there an ABI break?

the only change is the map value type:
transfer_request.h:82
`using descs_t = std::unordered_map<nixlBackendEngine *, std::unique_ptr<nixl_meta_dlist_t>>;`
→ `std::unique_ptr<nixl_stride_dlist_t>`. `sizeof(unique_ptr<T>)` and `sizeof(unordered_map<K, unique_ptr<T>>)` are independent of T, so `sizeof(nixlDlistH)` is unchanged. It's only ever constructed/destroyed inside libnixl, and users hold it as an opaque `nixlDlistH*` (created by `prepXferDlist`, freed by `releasedDlistH`). No layout or lifetime crosses the boundary

brminich (approved):

PR #1757, opened 2026-06-11
feat(kv): add shared KV backend interfaces under src/plugins/kv

gaowayne:
## What?
Introduce shared KV backend interfaces under src/plugins/kv/:

kv_store.h — iKVStore storage abstraction (put, get, exists)
kv_engine_impl.h — nixlKVEngineImpl abstract backend logic interface
kv_engine.h / kv_engine.cpp — nixlKVEngine thin nixlBackendEngine wrapper that delegates data-plane calls to impl_
Also updates meson.build to add src/plugins/kv to include paths and install the new public headers under include/nixl/plugins/kv/.

This is PR 1 of 3. It adds interfaces and the shared delegation layer only — no plugin behavior changes. INMEMKV and REDIS plugin work will follow in separate PRs.

## Why?
NIXL needs a consistent way to add KV-style backends (in-memory, Redis, etc.). These backends share the same NIXL lifecycle (registerMem, prepXfer, postXfer, …) but differ in how they store data.

Without shared interfaces, each plugin would duplicate the same nixlBackendEngine boilerplate (capability flags, connect/disconnect no-ops, delegation wiring). Centralizing that in src/plugins/kv/ gives future KV plugins a clear, reusable foundation.

## How?
Three-layer design:

nixlKVEngine              Shared thin wrapper (nixlBackendEngine)
    -> nixlKVEngineImpl   Plugin-specific logic (subclass per backend)
        -> iKVStore       Storage backend (in-memory, Redis, etc.)
nixlKVEngine owns common KV backend traits:

- local-only (supportsLocal() == true)

- no remote connection (supportsRemote() == false)

- no notifications (supportsNotif() == false)

All data-plane methods forward to a std::unique_ptr<nixlKVEngineImpl> impl_ supplied at construction time.

nixlKVEngineImpl defines the contract each KV plugin must implement: registerMem, deregisterMem, queryMem, prepXfer, postXfer, checkXfer, releaseReqH. prepXfer receives local_agent explicitly so impl classes do not depend on nixlBackendEngine protected members.

iKVStore isolates raw storage (put/get/exists) from NIXL descriptor-to-key mapping, which stays in the concrete nixlKVEngineImpl subclass.

Follow-up PRs:

PR	Scope
#2	Refactor INMEMKV example plugin onto these interfaces
#3	Add REDIS KV plugin (nixlRedisKVEngineImpl + hiredis iKVStore)

## Summary by CodeRabbit

* **New Features**
  * Added KV backend support: a KV engine wrapper, a plugin-facing KV engine interface, and a minimal synchronous KV store API to handle memory registration/deregistration, queries, and transfer lifecycle operations.
  * KV engine is declared local-only and provides standard engine entrypoints for plugins.

* **Chores**
  * KV plugin include directory exposed and public KV headers are installed into the include tree for plugin integration.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -439,7 +439,7 @@ if get_option('buildtype') == 'debug'
     run_command('truncate', '-s 0', plugfile, check: true)
 endif
 
-nixl_inc_dirs = include_directories('src/api/cpp', 'src/api/cpp/backend', 'src/infra', 'src/core', 'src/core/telemetry')
+nixl_inc_dirs = include_directories('src/api/cpp', 'src/api/cpp/backend', 'src/infra', 'src/core', 'src/core/telemetry', 'src/plugins/kv')
 nixl_gpu_inc_dirs = include_directories('src/api/gpu/ucx')
 plugins_inc_dirs = include_directories('src/plugins')
 utils_inc_dirs = include_directories('src/utils')
@@ -469,6 +469,9 @@ if get_option('install_headers')
   install_headers('src/core/transfer_request.h', install_dir: prefix_inc)
   install_headers('src/core/agent_data.h', install_dir: prefix_inc)
   install_headers('src/infra/mem_section.h', install_dir: prefix_inc)
+  install_headers('src/plugins/kv/kv_store.h', install_dir: prefix_inc + '/plugins/kv')
+  install_headers('src/plugins/kv/kv_engine_impl.h', install_dir: prefix_inc + '/plugins/kv')
+  install_headers('src/plugins/kv/kv_engine.h', install_dir: prefix_inc + '/plugins/kv')
   install_headers('src/core/telemetry/telemetry_event.h', install_dir: prefix_inc)
 
   if ucx_gpu_device_api_available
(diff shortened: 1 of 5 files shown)


ColinNV (changes requested):
The REDIS and INMEMKV PRs seem to both contain the KV infrastructure again. Please change them so that they don't duplicate everything from this PR. And please consider my comments about the general parts from those PRs, especially the one about what seems to be an overly complex structure.

aranadive:
> The REDIS and INMEMKV PRs seem to both contain the KV infrastructure again. Please change them so that they don't duplicate everything from this PR. And please consider my comments about the general parts from those PRs, especially the one about what seems to be an overly complex structure.

@ColinNV Im working with Harry and Wayne to simplify this. Both the REDIS and INMEMKV example plugin will rely on a base KV plugin infra (this one) and their own flavor engines to the base KV plugin (similar to how we do for S3).

gaowayne:
> The REDIS and INMEMKV PRs seem to both contain the KV infrastructure again. Please change them so that they don't duplicate everything from this PR. And please consider my comments about the general parts from those PRs, especially the one about what seems to be an overly complex structure.

@ColinNV 
yeah, after common nixl kv backend interface code defined. I will adjust redis plugin and inmemkv plugin to make sure no duplicated push. currently they are demo how it works for different type kv plugins.

PR #1759, opened 2026-06-11
Fix deregister backend status

### LINKED ISSUE #1758: [BUG]: deregisterMem ignores backend failures and reports false success
### Summary

`nixlLocalSection::remDescList()` currently calls `backend->deregisterMem(...)` but ignores its return value. As a result, `nixlAgent::deregisterMem()` may return `NIXL_SUCCESS` even when a backend failed to deregister one or more metadata objects.

This can leave NIXL and the backend in an inconsistent state, and the caller has no way to know that deregistration was only partially completed.

### Current behavior

In `src/infra/nixl_memory_section.cpp`, `remDescList()` does:

```cpp
for (size_t idx : indices) {
    backend->deregisterMem(target[idx].metadataP);
}

target.remDescs(std::move(indices));
return NIXL_SUCCESS;

```
The return value from `backend->deregisterMem(...)` is ignored.

Therefore:

1. Backend deregistration may fail.
2. NIXL still removes the descriptors from its internal base list.
3. `remDescList()` returns `NIXL_SUCCESS`.
4. `nixlAgent::deregisterMem()` may also return `NIXL_SUCCESS`.
5. The caller believes deregistration succeeded even though backend resources may still be registered.

### Expected behavior
`deregisterMem()` should not silently ignore backend deregistration failures.

At minimum:

- if any `backend->deregisterMem(...)` call fails, the failure should be propagated back to `nixlAgent::deregisterMem()`;

- descriptors whose backend deregistration succeeded should be removed from NIXL's internal descriptor list;

- descriptors whose backend deregistration failed should remain registered internally so their metadata is not lost.

flpanbin:
Fix https://github.com/ai-dynamo/nixl/issues/1758

## What?

Fix `nixlLocalSection::remDescList()` backend deregistration failures are propagated instead of being ignored.

The change removes descriptors from the internal section list only when their corresponding `backend->deregisterMem()` call succeeds. If any backend deregistration fails, the function returns that failure status.

## Why?
Before this change, `remDescList()` called `backend->deregisterMem()` but ignored the returned status:

```cpp
for (size_t idx : indices) {
    backend->deregisterMem(target[idx].metadataP);
}

target.remDescs(std::move(indices));
return NIXL_SUCCESS;
```
This allowed `nixlAgent::deregisterMem()` to report `NIXL_SUCCESS` even when the backend failed to deregister one or more metadata objects.

## How?
The fix records the result of each backend deregistration call:

For each descriptor:

- if backend deregistration succeeds, its index is added to a list of descriptors to remove;

- if backend deregistration fails, its index is not removed and the failure status is recorded.

After all attempts complete:

- successfully deregistered descriptors are removed from the section list;

- failed descriptors remain registered internally;

- the function returns the recorded failure status if any deregistration failed, otherwise `NIXL_SUCCESS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deregistration now logs a warning when a backend deregistration attempt fails but continues processing remaining descriptors and reports overall success to callers.

* **Tests**
  * Added a unit test verifying that a backend deregistration failure is ignored and that the deregistration call still returns success.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -21,6 +21,7 @@
 #include "nixl_descriptors.h"
 #include "mem_section.h"
 #include "backend/backend_engine.h"
+#include "common/nixl_log.h"
 #include "nixl_types.h"
 #include "serdes/serdes.h"
 
@@ -234,7 +235,10 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
     }
 
     for (size_t idx : indices) {
-        backend->deregisterMem(target[idx].metadataP);
+        const nixl_status_t ret = backend->deregisterMem(target[idx].metadataP);
+        if (ret != NIXL_SUCCESS) {
+            NIXL_WARN << "backend deregisterMem failed with status " << ret;
+        }
     }
 
     target.remDescs(std::move(indices));
--- a/test/gtest/unit/agent/agent.cpp
+++ b/test/gtest/unit/agent/agent.cpp
@@ -270,6 +270,26 @@ namespace agent {
         EXPECT_EQ(agent_->deregisterMem(reg_dlist, &extra_params), NIXL_SUCCESS);
     }
 
+    TEST_P(singleAgentWithMemParamFixture, DeregisterMemoryIgnoresBackendFailure) {
+        using testing::_;
+        using testing::Return;
+
+        nixl_b_params_t params;
+        nixlBackendH *backend;
+        EXPECT_EQ(agent_helper_->createBackendWithGMock(params, backend), NIXL_SUCCESS);
+
+        blob blob;
+        nixl_opt_args_t extra_params;
+        nixl_reg_dlist_t reg_dlist(GetParam());
+        EXPECT_EQ(agent_helper_->initAndRegisterMemory(blob, reg_dlist, extra_params, backend),
+                  NIXL_SUCCESS);
+
+        auto &engine = const_cast<mocks::GMockBackendEngine &>(agent_helper_->getGMockEngine());
+        EXPECT_CALL(engine, deregisterMem(_)).WillOnce(Return(NIXL_ERR_BACKEND));
+
+        EXPECT_EQ(agent_->deregisterMem(reg_dlist, &extra_params), NIXL_SUCCESS);
+    }
+
     TEST_F(singleAgentSessionFixture, RegisterDeregisterMemRepeatedTest) {
         constexpr int kWarmupIters = 3;
         constexpr int kTimedIters = 64;


src/infra/nixl_memory_section.cpp
```diff
@@ -233,11 +233,19 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
...
+        const nixl_status_t ret = backend->deregisterMem(target[idx].metadataP);
+        if (ret == NIXL_SUCCESS) {
+            deregistered_indices.push_back(idx);
+        } else {
+            bad_ret = ret;
+        }
     }
 
-    target.remDescs(std::move(indices));
+    target.remDescs(std::move(deregistered_indices));
```
guy-ealey-morag: Why do we need to filter out indices that failed to deregister?
Can't we call `remDescs` with the original `indices`?

flpanbin: Because `remDescs()` removes NIXL's internal handle to the backend metadata. If `backend->deregisterMem()` fails for an index and we still remove that index, we lose the metadata needed to retry cleanup while the backend may still have the resource registered. Filtering keeps failed deregistrations in `base` and removes only entries whose backend cleanup succeeded.

flpanbin: For example, if deregistering `[A, B, C]` succeeds for A/C but fails for B, removing the original `indices` would drop B from `base` even though B is still registered in the backend. Filtering keeps B in `base` and returns the backend error, while removing only A/C whose backend state was actually cleaned up.

guy-ealey-morag: Okay, I understand.

guy-ealey-morag (dismissed):

src/infra/nixl_memory_section.cpp
```diff
@@ -248,7 +256,7 @@ nixl_status_t nixlLocalSection::remDescList (const nixl_reg_dlist_t &mem_elms,
         memToBackend[nixl_mem].erase(backend);
     }
 
-    return NIXL_SUCCESS;
+    return bad_ret;
```
rakhmets: `nixlAgent::registerMem` and `nixlBackendEngine::registerMem `should throw an exception or return an error status (our case), and should issue an error message. Since making subsequent API calls is not possible.

In our case `nixlAgent::deregisterMem` and `nixlBackendEngine::deregisterMem` should return void and issue a warning message. Since making subsequent API calls is possible. However, these APIs already returns nixl_status_t, and it's not possible to change it at the moment. So, you can expect that they will always return OK status.

src/infra/nixl_memory_section.cpp
guy-ealey-morag: Given that `deregisterMem` is not expected to fail most of the time, I think it would be better for performance to keep a single `indices` vector and remove the failed elements from it.
In the happy path none are removed and it saves duplicating all of the entries into `deregistered_indices`.

One way to implement:
```c++
    nixl_status_t bad_ret = NIXL_SUCCESS;
    std::erase_if(indices, [&](size_t idx) {
        const nixl_status_t ret = backend->deregisterMem(target[idx].metadataP);
        if (ret != NIXL_SUCCESS) [[unlikely]] {
            bad_ret = ret;
            return true;
        }
        return false;
    });

    target.remDescs(std::move(indices));
```

wdyt?

dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27351850395/job/80815552795)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `dcd3e80e`

The log diff was crystal clear. I have everything needed for a precise diagnosis.

---

**Summary:** Clang Format Check failed on PR #1759 — a two-line `EXPECT_CALL` chain in `test/gtest/unit/agent/agent.cpp` must be collapsed to one line per `.clang-format` rules.

**Root cause:** The new test code introduced in commit `[REDACTED:Hex High Entropy String]` has a `EXPECT_CALL` + `.WillOnce` expression split across two lines with a line-continuation indent, which clang-format-19 requires to be on a single line. Specifically, the diff flagged around line 285 of `test/gtest/unit/agent/agent.cpp`:

```cpp
// Current (fails format check):
EXPECT_CALL(engine, deregisterMem(_))
    .WillOnce(Return(NIXL_ERR_BACKEND));

// Required by clang-format-19:
EXPECT_CALL(engine, deregisterMem(_)).WillOnce(Return(NIXL_ERR_BACKEND));
```

The formatter determined the chained call fits within the column limit and therefore must not be broken across lines.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (branch `fix-deregister-backend-status`, PR #1759)

**File:** `test/gtest/unit/agent/agent.cpp:285`

**Suggested fix:** Collapse the two-line `EXPECT_CALL` + `.WillOnce(...)` into a single line, exactly as clang-format prescribes:
```cpp
EXPECT_CALL(engine, deregisterMem(_)).WillOnce(Return(NIXL_ERR_BACKEND));
```
Then re-run `clang-format-19` locally to verify no other formatting issues remain before pushing:
```bash
clang-format-19 --style=file -i test/gtest/unit/agent/agent.cpp
```

**Related:** PR #1759 (`fix-deregister-backend-status`)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `dcbd7657-7cc7-42d0-b29c-405caef90fe3` in the triage console for the audit trail.

src/infra/nixl_memory_section.cpp
flpanbin: Got it. If the intended deregistration semantics are best-effort cleanup with warnings only, then returning `bad_ret` is not aligned with the API contract.
I can keep `deregisterMem()` returning `NIXL_SUCCESS`, and log a warning when a backend deregistration returns non-success.

src/infra/nixl_memory_section.cpp
flpanbin: Thanks for the suggestion. The `std::erase_if` approach makes sense if we decide to keep failed deregistration entries in the agent. I learned from it!

Based on the earlier [comments](https://github.com/ai-dynamo/nixl/pull/1759#discussion_r3395546697), my understanding is that `deregisterMem()` is intended to be best-effort cleanup, backend failures should be warning-only, and the API should still effectively complete the deregistration request from the agent side.

With that semantic, I think keeping the original `indices` is more consistent. I add a log to warn if `backend->deregisterMem()` fails, but still remove all requested descriptors. 

Does that sound right to you?

guy-ealey-morag: The new approach makes sense. @rakhmets can you verify?

rakhmets: I think it's more correct to log the error/warning in the place where it occurred.
Taking into account the fact that plugins (backends) also return `NIXL_SUCCESS` from `deregisterMem()`, this new warning message will not be printed.

Some backend implementations already issue a warning, some is missing that part, or issue an error instead.
E.g. in UCX plugin `nixlUcxContext::memDereg` is missing a warning in case of failure of `ucp_mem_unmap`; `nixlGusliEngine::deregisterMem` returns NIXL_SUCCESS, but issue an error.

The question is what was the motivation for this change? Did you notice some weird behavior during memory deregistration? if so, which backend(s) was(ere) used?

flpanbin: Thanks. The motivation came from an issue I observed while using an out-of-tree NIXL plugin that I developed.

The symptom was that a later transfer appeared to receive a stale/corrupted `metadataP`, which made me look into the core `registerMem` / `deregisterMem` / `prepXferDlist(populate)` path. During that code inspection, I noticed that `remDescList()` calls `backend->deregisterMem()` but ignores its return value, so I wanted to check whether backend cleanup failures could be hidden by the core layer.

I do not have evidence that this happens with any in-tree backend. Based on the discussion here, I agree that if `deregisterMem()` is intended to be best-effort cleanup, then backend-specific cleanup failures should probably be logged in the backend where the failure occurs, rather than as a generic warning in `remDescList()`.

flpanbin:
@guy-ealey-morag @rakhmets Thanks for the clarification. Based on the discussion, this PR is not the right fix, let us close it.

PR #1760, opened 2026-06-11
CI: revert DL partition to gb300 (revert #1698)

Alexey-Rivkin:
## What
Revert #1698, which switched `nixl-ci-dl-gpu` from `gb300nvl72_ci` (account `blackwell-ci`) to `gb200nvl72_ci` (account `oberon-gb-ci`) in `.ci/jenkins/lib/test-dl-matrix.yaml`.

## Why
The GB200 `ts2` fleet behind `gb200nvl72_ci` is failing the VRAM intra-agent CPP test with:

```
ibv_reg_dmabuf_mr(..., access=0x10000f) failed: Unknown error 524
failed to register ... dmabuf_fd N on md[6]=mlx5_16: Input/output error
```

`524` is `ENOTSUPP` - the GPU exports a valid dma-buf fd, but the mlx5 NIC kernel can't import it.
Looks like a host-side dma-buf regression on the GB200 nodes, not a NIXL/UCX code issue.

Reverting to `gb300nvl72_ci` gets CI green again while the GB200 host problem is chased with the cluster team separately.

## Trade-off
#1698 moved to GB200 because `gb300nvl72_ci` was short on resources, so jobs may queue longer here. Acceptable vs. a red CI on every DL run.

## Summary by CodeRabbit

* **Chores**
  * Updated the GPU test infrastructure configuration to use the newer SLURM partition and account settings.


guy-ealey-morag (approved):

dpressle:
Dont merge this PR we are not sure there are enough resources and we will fail on allocation timeout

PR #1762, opened 2026-06-11
UCX: provide an option to enable/disable UCX cache

ZhenlongMa:
## What?
Disable UCX registration caches and memtype caches by default in the UCX backend.

## Why?
NIXL always passes the local memory handle (`memh`) explicitly to UCX operations, so UCX does not need to rely on memory hooks for this path.

Installing memory hooks is intrusive and can lead to correctness or stability issues, especially for dynamic memory workloads. Disabling rcache and memtype cache by default avoids installing those hooks while preserving the existing explicit memory registration flow.

## How?
This PR adds a build-time `enable_cache` Meson option. When it is not set, NIXL configures UCX with:
- `RCACHE_ENABLE=n`
- `GDR_COPY_RCACHE=n`
- `ROCM_COPY_RCACHE=n`
- `MEMTYPE_CACHE=n`

so that all kinds of caches are disabled.

## Summary by CodeRabbit

* **New Features**
  * Added an optional build setting to enable UCX registration and memory-type caching, disabled by default.
* **Bug Fixes**
  * Improved UCX communication setup by explicitly identifying host memory and consistently initializing request options.
* **Chores**
  * Limited unreleased cached registrations and ensured caching remains disabled unless explicitly enabled.


ZhenlongMa:
Hi reviewers,
This PR has been waiting for review for about two weeks. Just wanted to check if there is any additional work or changes needed from my side.
Thanks.

iyastreb:
LGTM
Disabling GDR/rocm rcache may change registration cost, but I think it's not too risky

iyastreb (dismissed):

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -237,9 +237,10 @@ nixlUcxEp::sendAm(nixl::ucx::am_cb_op_t msg_id,
         return status;
     }
 
-    ucp_request_param_t param;
-    param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
+    ucp_request_param_t param = {0};
```
tomerg-nvidia: This restores a degradation that was resolved in https://github.com/ai-dynamo/nixl/pull/2017

iyastreb (approved):

iyastreb:
Does it bring perf improvement, if you test with nixlbench?

brminich (changes requested):
waiting for perf analysis

PR #1767, opened 2026-06-12
fix(libfabric): fix xfer_id wraparound dead code and stale pending notification eviction

ztxdcyy:
## What?
Two fixes to the libfabric backend's notification tracking lifecycle:

1. **Fix dead wraparound logic in `getNextXferId()`**: The `g_xfer_id_counter` is `std::atomic<uint16_t>`, so `xfer_id > 0xFFFF` is always false — the entire CAS reset branch is unreachable dead code. Replace with a simple skip of sentinel value 0.

2. **Evict stale `pending_notifications_` entries to prevent xfer_id collision**: 16-bit xfer_id wraps every ~65K `postXfer()` calls. If a `PendingNotification` entry never completes (e.g., lost RDMA Write CQ completion or missing notification fragment), it stays in the map indefinitely. On wraparound, `try_emplace` returns the stale entry, causing counter corruption between old and new transfers. Fix: add a 30-second TTL eviction in `checkPendingNotifications()`.

## Why?
- The dead wraparound code gives a false sense of safety while doing nothing
- Under sustained workloads (disaggregated inference with thousands of transfers/sec), xfer_id wraps in ~1 minute. Any stuck notification entry from a transient network error becomes a ticking time bomb for data corruption

## How?
- Commit 1: Remove dead CAS branch, rely on uint16_t natural overflow, skip sentinel 0
- Commit 2: Add `created_at` timestamp to `PendingNotification`, add `evictStalePendingNotifications()` called under existing `receiver_tracking_mutex_` lock

## Related
Partially addresses additional issues mentioned in https://github.com/ai-dynamo/nixl/issues/1722

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification tracking: pending receiver notifications are now automatically evicted after a 30-second timeout, reducing issues caused by stale entries and improving transfer reliability.
  * More robust transfer ID handling: the transfer ID generator now skips invalid/zero IDs to prevent rare ID reuse and enhance stability.


dpressle:
👀 Investigating [`Clang Format Check`](https://github.com/ai-dynamo/nixl/actions/runs/27404440162/job/80989950194)…

dpressle:
🤖 **CI Triage Agent** — `Clang Format Check` · commit `27746939`

I have everything needed for a confident diagnosis. Here is the full report:

---

**Summary:** Clang Format Check failed on `src/plugins/libfabric/libfabric_backend.cpp` due to a stream-operator line-continuation style violation introduced in PR #1767.

**Root cause:** The three modified files in commit `[REDACTED:Hex High Entropy String]` include a `NIXL_WARN` log statement at approximately line 1643 of `libfabric_backend.cpp` where `clang-format-19` disagrees with how the `<<` chained stream operators are split across lines. Specifically, the PR wrote:

```cpp
// PR version (fails clang-format):
<< "s, received_completions=" << it->second.received_completions
<< "/" << it->second.expected_completions
<< ", received_fragments=" << it->second.received_msg_fragments
<< "/" << it->second.expected_msg_fragments;
```

But `clang-format-19` (per the repo's `.clang-format` style file) expects the `/` separator to be appended to the *preceding* line rather than starting a new one:

```cpp
// clang-format expected version:
<< "s, received_completions=" << it->second.received_completions << "/"
<< it->second.expected_completions
<< ", received_fragments=" << it->second.received_msg_fragments << "/"
<< it->second.expected_msg_fragments;
```

The formatter exited with code 1 because of this diff, causing the CI job to fail.

**Implicated commit:** `[REDACTED:Hex High Entropy String]` (PR #1767, branch `bugfix/libfabric_xfer_id`)

**File:** `src/plugins/libfabric/libfabric_backend.cpp:1643`

**Suggested fix:** Run `clang-format-19` locally on the three changed files before pushing, or apply the exact reformatting shown in the diff above. The quickest one-liner to fix it in-place:

```bash
clang-format-19 -style=file -i \
  src/plugins/libfabric/libfabric_backend.cpp \
  src/plugins/libfabric/libfabric_backend.h \
  src/utils/libfabric/libfabric_common.cpp
```

Then `git add` and amend/push the commit. The only mandatory change is in `libfabric_backend.cpp` around line 1643 — move the `"/"` string literal to the end of the preceding `<<` chain line rather than starting a new line with it.

**Related:** PR #1767 (`bugfix/libfabric_xfer_id`)



> 🛡️ This comment had 1 potential secret(s) redacted (Hex High Entropy String). See request_id `df8f4683-9ec9-4cde-925d-4b043555e5f3` in the triage console for the audit trail.

ztxdcyy:
Hi @amitrad-aws, this PR addresses the xfer_id wraparound and stale pending notification issues mentioned in #1722 (the additional related issues section). It complements your PR #1747. Would appreciate your review when you get a chance. Thanks!

src/utils/libfabric/libfabric_common.cpp
```diff
@@ -178,19 +178,9 @@ uint16_t
...
 
-    // Handle wraparound: 16-bit field can hold 0 to 65,535
-    if (xfer_id > NIXL_XFER_ID_MASK) {
-        // Reset counter atomically and get a fresh ID
-        uint16_t expected = xfer_id;
-        while (expected > NIXL_XFER_ID_MASK &&
-               !g_xfer_id_counter.compare_exchange_weak(expected, 1)) {
-            expected = g_xfer_id_counter.load();
-        }
+    // uint16_t wraps naturally from 65535 back to 0; skip 0 since it is used as a sentinel value
```
fengjica, line 181: Why do need to keep 0 as a sentinel value to skip? I don't see any special meaning of 0 being xfer_id in the code today.

src/plugins/libfabric/libfabric_backend.h
```diff
@@ -221,6 +221,10 @@ class nixlLibfabricEngine : public nixlBackendEngine {
     std::unordered_set<uint32_t> received_remote_writes_; // All received XFER_IDs (global)
 
     // Notification Queuing
+    // Stale entry timeout: entries older than this are evicted to prevent xfer_id wraparound
+    // collision (16-bit xfer_id wraps every 65534 postXfer calls).
```
fengjica, line 231: Regarding xfer_id wraparound, a theoretic estimation:
If we use some data from a message size at smaller end to do some estimation, which I think will tend to have more transfers at the same time: 256 KB, end-to-end transfer time at around 100 us, in nixlbench we see 40 GB/sec per rail pair. That translates to 156,250 transfer/sec. Then times 100 us, according to Little's Law (capacity = throughput * latency), there are just ~16 messages at the same time. even if we consider the case that an agent is talking to multiple nodes, it looks still enough.

Then, if wraparound happens, because of those staying around for 30 seconds, as what you aim to fix here, it strong suggests we have some bugs, e.g. lost RDMA Write CQ or missing notification fragment, like you suggested. I feel having this stale eviction mechanism may hide the underlying bug, rather than fixing it.

brminich:
@ztxdcyy can you pls address the comments?

src/utils/libfabric/libfabric_common.cpp
ztxdcyy: Well, I'm not sure about the usage of 0, I just want to keep it same as it was.

src/plugins/libfabric/libfabric_backend.h
ztxdcyy: You’re right. This timeout cleanup mechanism resolves the warp loopback issue, yet operating without any notifications or logs obscures the root cause of failures and makes it harder to pinpoint the actual problem.

PR #1773, opened 2026-06-15
fix(core): forbid duplicate descriptor registration

lluki:
### Summary

This is a follow up for PR #1744 . Note the discussion there, at the end we agreed on a different approach ([comment](https://github.com/ai-dynamo/nixl/pull/1744#issuecomment-4678598319)) to prevent duplicate registrations instead of searching for duplicates. This PR implements this approach

### Existing client behavior

**Note**: This PR introduces user facing changes. Before it was allowed to perform a `register(A), register(A)` - after this PR the second register will return an error

### Problem

Backends  free metadata on deregister (e.g. POSIX `deregisterMem` closes the file descriptor). The section keys descriptors on the `nixlBasicDesc` sort key `(devId, addr, len)`. Registering the same descriptor twice collapses to one section entry while producing two distinct backend metadata objects, so `getIndex()` (first match) returns the same entry twice: deregistration frees it twice (double free) and the duplicate leaks.

### Fix

Two complementary, all-or-nothing rules: 

1. **Forbid duplicate descriptors** -- `nixlLocalSection::addDescList` rejects a descriptor that duplicates another in the same list or one already registered, before any `backend->registerMem`. Returns `NIXL_ERR_INVALID_PARAM`.
2. **Unique devId per path-mode file** -- in path-mode the file is named by the path while the section keys on `devId`, so distinct files must use distinct devIds. Each FILE_SEG backend (posix, cuda_gds, gds_mt, hf3fs) keeps a set of in-use path-mode devIds, rejects reuse, and releases on deregister. Serialized by the agent lock (`nixlAgent::registerMem`/`deregisterMem`), so no backend lock. Rule 2 also turns "different file, same devId" into a clear error instead of the confusing duplicate-descriptor rejection Rule 1 would give.

Because the section can never hold two entries with the same sort key, `getIndex()` always resolves uniquely -- so `remDescList` / `removeLocalData` stay as their original simple loops and no `resolveDuplicateIndices` helper is needed.

### Alternative designs considered

**Composition helper.** A small reusable `PathModeDevIdRegistry` class (living in `utils/file`, held as a member by each file engine) that centralizes the set, the `parsePathMeta` detection, the reject+log, and an RAII reservation that auto-rolls-back on error. Pro: single source of truth and the RAII guard removes the rollback footgun, with no engine-interface change. Con: each backend still has ~2 call sites (reserve in register, release in deregister), so it dedups the logic but not the wiring.

**Inline the registration logic**: instead of the helper we could also inline the reservations

**Boolean + NVI in nixlBackendEngine.** Add an "enforce unique devId" bool to the base ctor (subclasses opt in), make base `registerMem`/`deregisterMem` non-virtual wrappers that do the enforcement around new pure-virtual `registerMemImpl`/`deregisterMemImpl`, and store the devId on the `nixlBackendMD` base so deregister can release it. Pro: truly single enforcement point that removes all per-backend bookkeeping *and* the call sites. Con: it forces a mechanical `registerMem`->`registerMemImpl` rename in *every* engine (not just file ones) plus a field on the shared MD base and a path-mode-detection layering question - too much for this fix.

## Release Notes

* **Bug Fixes**
  * Reject duplicate memory registration descriptors during registration, with clear error logging and atomic all-or-nothing behavior.
  * Enforce unique `devId` for FILE_SEG path-mode registrations across backends, including safe reservation rollback on failure and correct `devId` release on deregistration.
* **Documentation**
  * Strengthened the path-mode contract: each path-mode file segment must use a unique `devId`.
* **Tests**
  * Added unit tests covering descriptor de-duplication and path-mode `devId` uniqueness, plus POSIX fd-mode `devId` reuse behavior.

### DIFF
--- a/src/infra/nixl_memory_section.cpp
+++ b/src/infra/nixl_memory_section.cpp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 #include <map>
+#include <set>
 #include <algorithm>
 #include <iostream>
 #include "nixl.h"
@@ -23,6 +24,7 @@
 #include "backend/backend_engine.h"
 #include "nixl_types.h"
 #include "serdes/serdes.h"
+#include "common/nixl_log.h"
 
 /*** Class nixlMemSection implementation ***/
 
@@ -141,6 +143,17 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
     const nixl_mem_t nixl_mem = mem_elms.getType();
 
     nixlSecDescList &target = emplace(nixl_mem, backend);
+
+    // Reject duplicates - within args (seen) and against registered entries (target)
+    std::set<nixlBasicDesc> seen;
+    for (const auto &mem : mem_elms) {
+        const nixlBasicDesc key = normalizeSecDesc(mem, nixl_mem);
+        if (!seen.insert(key).second || target.getIndex(key) >= 0) {
+            NIXL_ERROR << "duplicate descriptor registration (addr=" << key.addr
+                       << " len=" << key.len << " devId=" << key.devId << ")";
+            return NIXL_ERR_INVALID_PARAM;
+        }
+    }
 
     nixlSectionDesc local_sec, self_sec;
     nixlBasicDesc *lp = &local_sec;
--- a/src/plugins/gds_mt/gds_mt_backend.h
+++ b/src/plugins/gds_mt/gds_mt_backend.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -110,6 +110,7 @@ public:
 private:
     gdsMtUtil gds_mt_utils_;
     std::unordered_map<int, std::weak_ptr<gdsMtFileHandle>> gds_mt_file_map_;
+    nixl::PathModeDevIdRegistry path_mode_devids_;
     size_t thread_count_;
     std::unique_ptr<tf::Executor> executor_;
 };
--- a/test/gtest/unit/agent/agent.cpp
+++ b/test/gtest/unit/agent/agent.cpp
@@ -334,6 +334,68 @@ namespace agent {
                              << timed1_ns << " ns2=" << timed2_ns << " ratio=" << ratio << ")";
     }
 
+    // Registering a descriptor that is already registered must be rejected without
+    // re-registering it, so the section never holds two entries with the same sort key.
+    TEST_F(singleAgentSessionFixture, RegisterDuplicateDescriptorRejectedTest) {
+        nixl_b_params_t params;
+        nixlBackendH *backend;
+        ASSERT_EQ(agent_helper_->createBackendWithGMock(params, backend), NIXL_SUCCESS);
+
+        nixl_opt_args_t extra_params;
+        extra_params.backends.push_back(backend);
+
+        char local_md[1] = {};
+        char self_md[1] = {};
+        auto &engine = const_cast<mocks::GMockBackendEngine &>(agent_helper_->getGMockEngine());
+        // Only the first registration may reach the backend; the duplicate is rejected earlier.
+        EXPECT_CALL(engine, registerMem(testing::_, testing::_, testing::_))
+            .Times(1)
+            .WillRepeatedly([&](const nixlBlobDesc &, const nixl_mem_t &, nixlBackendMD *&out) {
+                out = reinterpret_cast<nixlBackendMD *>(&local_md[0]);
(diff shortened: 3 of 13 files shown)


test/gtest/unit/agent/agent.cpp
```diff
@@ -334,6 +334,68 @@ namespace agent {
...
+
+        auto &engine = const_cast<mocks::GMockBackendEngine &>(agent_helper_->getGMockEngine());
+        EXPECT_CALL(engine, registerMem(testing::_, testing::_, testing::_)).Times(0);
+        EXPECT_CALL(engine, deregisterMem(testing::_)).Times(0);
+
+        blob region;
+        nixl_reg_dlist_t reg(DRAM_SEG);
+        reg.addDesc(region.getDesc());
+        reg.addDesc(region.getDesc());
+        EXPECT_NE(agent_->registerMem(reg, &extra_params), NIXL_SUCCESS);
```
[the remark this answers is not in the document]
lluki: This is not actually true, the agent collapses this error into `NIXL_ERR_BACKEND` and thats all the client can learn (which is maybe a mistake, but thats another topic)

src/plugins/gds_mt/gds_mt_backend.h
```diff
@@ -110,6 +111,8 @@ class nixlGdsMtEngine : public nixlBackendEngine {
 private:
     gdsMtUtil gds_mt_utils_;
     std::unordered_map<int, std::weak_ptr<gdsMtFileHandle>> gds_mt_file_map_;
+    // in-use path-mode devIds
+    std::unordered_set<uint64_t> path_mode_devids_;
```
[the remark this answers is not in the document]
lluki: no other member in the backends im touching is using that style

src/infra/nixl_memory_section.cpp
```diff
@@ -142,6 +144,17 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
 
     nixlSecDescList &target = emplace(nixl_mem, backend);
 
+    // Reject duplicates - within args (seen) and against registered entries (target)
+    std::set<nixlBasicDesc> seen;
+    for (const auto &mem : mem_elms) {
```
brminich, line 149: why do we need this change?

brminich (changes requested):

src/infra/nixl_memory_section.cpp
lluki: The point of this PR is to avoid duplicate registrations. This can happen in two cases:

1) separate register calls `agent.registerMem([A])` followed by `agent.registerMem([A])` - covered by `target.getIndex()` below

2) `agent.registerMem([A,A])` - covered by the `seen` set.

src/infra/nixl_memory_section.cpp
```diff
@@ -142,6 +144,17 @@ nixlLocalSection::addDescList(const nixl_reg_dlist_t &mem_elms,
 
     nixlSecDescList &target = emplace(nixl_mem, backend);
 
+    // Reject duplicates - within args (seen) and against registered entries (target)
```
ovidiusm, line 147: This impacts all registrations, not just the path-based registrations from https://github.com/ai-dynamo/nixl/pull/1635

If we want to cherry pick this to a minor release (1.3.1) this makes it a no-go since it is a behavior change

Is it possible to have a simpler patch for 1.3.x?

lluki: Yes, i will make a path-mode focused patch and create bug report for the remaining issue. Then we can discuss the correct fix & timeline in the bug.

ovidiusm: I checked vLLM code and it seems that it may trigger double registrations.

Because of this I don't think we can forbid them.

Could we change the path-based code to accept and handle them gracefully?

lluki: We can also fix it on the deregistration path, see my first PR proposal: https://github.com/ai-dynamo/nixl/pull/1744

In particular these lines: 
https://github.com/lluki/nixl/blob/c030e5a2d0a39904c5510a508a1a6bb6aeab2a3c/src/infra/nixl_memory_section.cpp#L254-L257

I think both approaches are correct, this PR ( #1773 ) seems a bit more logical to me, but introduces more user visible changes. #1744 also fixes the problem, with less user visible changes. We had some discussions about the correctnes in #1744 - I personally think it is correct, but i might be wrong

lluki:
Patch for 1.3.x - only affecting path-mode  https://github.com/ai-dynamo/nixl/pull/1790

src/infra/nixl_memory_section.cpp
lluki: Patch for 1.3.x - only affecting path-mode  https://github.com/ai-dynamo/nixl/pull/1790Patch for 1.3.x - only affecting path-mode  https://github.com/ai-dynamo/nixl/pull/1790

ovidiusm: I think since double registrations are already ocuring in consumers of NIXL API we should allow it in general, for consiustency, even if is not necessarily desirable. And instead apply a smaller fix as in 1744 that just fixes the deregistration

However 1744 also seems to have expanded scope, and to me it is not clear why... you seem to block deregistration in case of duplicates, instead of just handling that gracefully. Let's discuss there

PR #1779, opened 2026-06-15
telemetry: fall back to NOP collector when requested without a sink

e-eygin:
## What?

Fixes the regression introduced by #1716 (NIX-1361) where telemetry requested
through the API but with no exporter/sink configured was silently disabled,
breaking in-process `get_xfer_telemetry()`.

- `getExporterName()` now falls back to the collect-only **NOP** exporter when
  telemetry is requested but neither `NIXL_TELEMETRY_EXPORTER` nor
  `NIXL_TELEMETRY_DIR` is set, instead of returning "no sink".
- `getExporterName()` returns `std::string` (it can no longer be empty) and
  `nixlTelemetry::create()` is simplified to always build an active instance.
- Tests and `docs/telemetry.md` updated.

## Why?

vLLM uses telemetry in-process: `nixl_agent_config(..., capture_telemetry=True)`
followed by `get_xfer_telemetry(handle)` at completion, with no
`NIXL_TELEMETRY_DIR`/`NIXL_TELEMETRY_EXPORTER`. After #1716, `create()` decided
enablement purely from the sink env vars and returned `nullptr`, so
`get_xfer_telemetry()` returned `NIXL_ERR_NO_TELEMETRY` and the transfer was
treated as failed.

`get_xfer_telemetry()` is an in-process consumer that needs no sink, so #1716's
assumption "no sink = nothing consumes telemetry" is wrong, and it contradicts
the documented `captureTelemetry` contract ("Capture telemetry info regardless
of environment variables"). API-only telemetry must keep working. This is
release-relevant and should be backported.

## How?

`create()` is only ever called on an explicit request (`NIXL_TELEMETRY_ENABLE`
or the agent's `captureTelemetry`), so the fix lives entirely in
`getExporterName()`. Resulting behavior:

| Telemetry requested? | Sink configured? | Result |
| --- | --- | --- |
| yes | `NIXL_TELEMETRY_EXPORTER` set | that exporter |
| yes | `NIXL_TELEMETRY_DIR` set | `BUFFER` |
| yes | neither | `NOP` (collect-only; `get_xfer_telemetry()` works, nothing written) |
| no | — | `telemetry_` stays null → `NIXL_ERR_NO_TELEMETRY` (unchanged) |

Note: the NOP static plugin is registered under the key `"NOP"` (the
`NIXL_REGISTER_STATIC_PLUGIN(Telemetry, NOP)` macro stringifies the name and the
lookup is case-sensitive), so the fallback constant is `"NOP"`. An explicit,
unloadable `NIXL_TELEMETRY_EXPORTER` still throws.

Tests (the C++ core is the source of truth, so coverage is added there rather
than duplicated per binding):
- `telemetry_test.cpp`: replaced `CreateReturnsNullWithoutSink` with
  `CreateFallsBackToNopWithoutSink`.
- `test_transfer.cpp`: flipped `GetXferTelemetryAPI` (`NIXL_TELEMETRY_ENABLE=y` +
  no sink) to `NIXL_SUCCESS`; added `GetXferTelemetryCaptureNoSink` (the vLLM
  scenario: `capture_telemetry=true` + no sink → `getXferTelemetry()` succeeds).
- `test_nixl_api.py`: fixed `test_get_xfer_telemetry_without_sink`, which had
  encoded the buggy "no telemetry without a sink" expectation.

Validation: `telemetryTest.*` (16/16) and `*TestTransferTelemetry*` (20/20
across the 4 UCX instantiations) pass; clang-format-19 clean on changed lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * When telemetry is requested but no sink/exporter is configured, telemetry remains active in-process and `getXferTelemetry()` / `get_xfer_telemetry()` succeed instead of failing.
  * Telemetry initialization now uses a collect-only (NOP) fallback when no output destination is available.
* **Documentation**
  * Updated telemetry docs to clarify collect-only fallback behavior, including sink/exporter fallback details and that the run interval has no observable effect without an output destination.
* **Tests**
  * Updated and added C++ and Python regression coverage for “no sink” initialization and capture-only telemetry behavior.

### DIFF
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -72,11 +72,15 @@ nixlTelemetryEventTypeForStatus(nixl_status_t s) {
 constexpr std::chrono::milliseconds DEFAULT_TELEMETRY_RUN_INTERVAL = 100ms;
 constexpr size_t DEFAULT_TELEMETRY_BUFFER_SIZE = 4096;
 constexpr const char *defaultTelemetryPlugin = "BUFFER";
+// Collect-only sink (registered as "NOP"): keeps telemetry active so the
+// datapath records events and getXferTelemetry() works, but writes nowhere.
+// Used when telemetry is explicitly requested but no output sink is configured.
+constexpr const char *collectOnlyTelemetryPlugin = "NOP";
 
 namespace {
 // Defined below; declared here for use by create() and the constructor's
 // member-initializer list.
-[[nodiscard]] std::optional<std::string>
+[[nodiscard]] std::string
 getExporterName();
 [[nodiscard]] std::string
 validateAgentName(const std::string &agent_name);
@@ -86,16 +90,11 @@ resolveMaxBufferedEvents();
 
 [[nodiscard]] std::unique_ptr<nixlTelemetry>
 nixlTelemetry::create(const std::string &agent_name) {
-    const std::optional<std::string> exporter_name = getExporterName();
-
-    // No exporter/sink configured: telemetry is intentionally disabled, so do
-    // not construct a nixlTelemetry at all (avoids spinning up its thread pool
-    // only to tear it down).
-    if (!exporter_name) {
-        return nullptr;
-    }
-
-    return std::make_unique<nixlTelemetry>(agent_name, *exporter_name);
+    // create() is only called when telemetry is explicitly requested (see
+    // nixlAgent's constructor), so it always produces an active instance:
+    // getExporterName() falls back to the collect-only NOP sink when no output
+    // sink is configured.
+    return std::make_unique<nixlTelemetry>(agent_name, getExporterName());
 }
 
 nixlTelemetry::nixlTelemetry(const std::string &agent_name, const std::string &exporter_name)
@@ -130,24 +129,28 @@ nixlTelemetry::~nixlTelemetry() {
 }
 
 namespace {
-[[nodiscard]] std::optional<std::string>
+[[nodiscard]] std::string
 getExporterName() {
     if (const auto name = nixl::config::getValueOptional<std::string>(telemetryExporterVar)) {
         if (!name->empty()) {
-            return name;
+            return *name;
         }
         NIXL_DEBUG << "Ignoring empty " << telemetryExporterVar << " environment variable";
     }
 
-    if (!nixl::config::checkExistence(telemetryDirVar)) {
-        NIXL_DEBUG << telemetryDirVar
-                   << " is not set and no exporter was specified; NIXL telemetry is disabled";
-        return std::nullopt;
+    if (nixl::config::checkExistence(telemetryDirVar)) {
+        NIXL_INFO << "No telemetry exporter was specified, using default: "
+                  << defaultTelemetryPlugin;
+        return defaultTelemetryPlugin;
     }
 
-    NIXL_INFO << "No telemetry exporter was specified, using default: " << defaultTelemetryPlugin;
-
-    return defaultTelemetryPlugin;
+    // Telemetry was explicitly requested but no output sink is configured. Fall
+    // back to the collect-only NOP sink so in-process consumers such as
+    // getXferTelemetry() keep working; nothing is written out.
+    NIXL_INFO << telemetryDirVar << " is not set and no exporter was specified; collecting "
+              << "in-process via " << collectOnlyTelemetryPlugin
+              << " so get_xfer_telemetry() works";
+    return collectOnlyTelemetryPlugin;
 }
 
 [[nodiscard]] std::string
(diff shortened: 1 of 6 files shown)


test/gtest/telemetry_test.cpp
```diff
@@ -138,12 +138,24 @@ TEST_F(telemetryTest, NonexistentExporterThrows) {
...
+    auto telemetry = nixlTelemetry::create(testFile_);
     envHelper_.addVar(TELEMETRY_DIR_VAR, testDir_.string()); // restore for TearDown symmetry
+
+    ASSERT_NE(telemetry, nullptr)
+        << "telemetry requested without a sink must collect in-process via NOP";
+    EXPECT_NO_THROW(telemetry->updateTxBytes(1024));
+
+    // Collect-only NOP fallback writes nothing to disk.
+    EXPECT_FALSE(fs::exists(testDir_.string() + "/" + testFile_));
 }
```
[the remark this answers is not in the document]
e-eygin: Addressed where an inherited env var could actually flip an assertion:
- `telemetry_test.cpp::CreateFallsBackToNopWithoutSink` neutralizes `NIXL_TELEMETRY_EXPORTER` (empty is ignored by `getExporterName()`), making the `EXPECT_FALSE(file exists)` check deterministic.
- `test_nixl_api.py::test_get_xfer_telemetry_without_sink` clears `NIXL_TELEMETRY_DIR`/`EXPORTER`.

`test_transfer.cpp` left as-is: `GetXferTelemetryAPI` sets `NIXL_TELEMETRY_ENABLE=y` and only asserts telemetry is active (true for any exporter, so inherited `DIR`/`EXPORTER` can't flip it); `GetXferTelemetryCaptureNoSink` would only break under an inherited `NIXL_TELEMETRY_ENABLE=n`, which CI doesn't set, and unsetting it would need a new `ScopedEnv` unset helper — out of scope for this regression fix.

src/core/telemetry/telemetry.cpp
```diff
@@ -130,24 +129,28 @@ nixlTelemetry::~nixlTelemetry() {
...
+        return defaultTelemetryPlugin;
     }
 
-    NIXL_INFO << "No telemetry exporter was specified, using default: " << defaultTelemetryPlugin;
-
-    return defaultTelemetryPlugin;
+    // Telemetry was explicitly requested but no output sink is configured. Fall
+    // back to the collect-only NOP sink so in-process consumers such as
+    // getXferTelemetry() keep working; nothing is written out.
+    NIXL_INFO << telemetryDirVar << " is not set and no exporter was specified; collecting "
```
iyastreb: IMO both comment and message is excessive, just enough to log that "in-process telemetry is used"
get_xfer_telemetry should not be mentioned, it's Py API

test/gtest/telemetry_test.cpp
```diff
@@ -138,12 +138,29 @@ TEST_F(telemetryTest, NonexistentExporterThrows) {
         { nixlTelemetry telemetry(testFile_, "nixl_nonexistent_exporter"); }, std::runtime_error);
 }
 
-TEST_F(telemetryTest, CreateReturnsNullWithoutSink) {
-    // With no exporter and no telemetry dir configured, create() resolves no
-    // sink and returns nullptr without constructing a telemetry instance.
-    envHelper_.popVar(); // drop NIXL_TELEMETRY_DIR set by SetUp
-    EXPECT_EQ(nixlTelemetry::create(testFile_), nullptr);
+TEST_F(telemetryTest, CreateFallsBackToNopWithoutSink) {
+    // Regression (#1716 / NIX-1361): when telemetry is explicitly requested
```
iyastreb: - I think we typically don't hardcode PR in the code
- This comment repeats what's already commented above
- Imo too verbose, better no comment at all

src/core/telemetry/telemetry.h
```diff
@@ -51,10 +51,15 @@ struct periodicTask {
 class nixlTelemetry {
 public:
     /**
-     * @brief Creates a telemetry instance if an exporter/sink is configured.
+     * @brief Creates an active telemetry instance.
```
ColinNV: IMO this either needs a definition of what 'active' means or removal of the term.

e-eygin: Done in `ddc30fa4` — removed "active" in favor of a concrete description (always returns non-null; NOP fallback records events in process so `getXferTelemetry()` returns data but writes nothing).

test/gtest/telemetry_test.cpp
e-eygin: Done in `4e121f81` — trimmed the log message/comment (dropped the `get_xfer_telemetry` mention) and removed the verbose/PR-hardcoded test comments.

ColinNV (approved):

iyastreb (approved):

ovidiusm (approved):

PR #1787, opened 2026-06-16
test/ucx: add AMD GPU support to plugins/ucx unit tests

edgargabriel:
## What?
Add support for running UCX plugin unit tests on AMD GPUs using ROCm/HIP

Changes:
- Add meson options 'use_rocm' and 'rocm_path' for ROCm configuration
- Implement ROCm detection in test/unit/plugins/ucx/meson.build
- Map CUDA API calls to HIP equivalents via preprocessor defines
- Add hipStreamSynchronize(0) after hipMemset calls (required for ROCm)
- Support both CUDA and ROCm builds (mutually exclusive)

## How?
Implementation details:
- All CUDA function calls remain unchanged in source code
- When HAVE_ROCM is defined, #define statements map CUDA→HIP
- ROCm detection is fully localized to UCX test directory
- Zero impact on other components (CUDA detection continues normally)

Building with ROCm:
  meson setup build -Duse_rocm=true -Drocm_path=/opt/rocm

Tested on an MI250X system with an MLX5 NIC and UCX 1.21.xRC2. All UCX plugin unit tests passed.

## Summary by CodeRabbit

* **New Features**
  * Added Meson options to enable ROCm/HIP (`use_rocm`) and configure the ROCm install location (`rocm_path`).
  * Extended UCX GPU/VRAM unit tests to support ROCm in addition to CUDA, including worker coverage.
* **Bug Fixes**
  * Improved backend-aware ROCm GPU memory and pointer handling for VRAM test paths, with correct device discovery and data movement.
* **Documentation**
  * Updated ROCm/HIP build guidance to be vendor-neutral, reorganized ROCm plugin notes, added a Meson ROCm command example, and refreshed remaining known gaps.


README.md
```diff
@@ -167,15 +167,20 @@ Common build options:
...
+$ meson setup build -Duse_rocm=true -Drocm_path=/opt/rocm
+```
 
 **Plugins on ROCm hosts (CUDA toolchain absent):**
 - `UCX` — primary transport for AMD GPU memory (requires UCX built with `--with-rocm`).
 - `POSIX`, `OBJ`, `AZURE_BLOB`, `HF3FS`, `MOONCAKE`, `GUSLI`, `UCCL` — vendor-neutral; build unchanged.
 - `GDS` / `GDS_MT`, `GPUNETIO`, `LIBFABRIC` (with `-DHAVE_CUDA`) — skip automatically because their CUDA / cuFile / DOCA dependencies are not found.
 
 **Known gaps (will be addressed in follow-up PRs):**
-- `nixlbench` (the NIXL benchmark tool) needs CUDA-driver-API → HIP translation work before it builds on ROCm. Use `examples/cpp/nixl_etcd_example` for transfer validation in the meantime.
```
brminich, line 178: why is it removed? nixlbench is not touched in this PR

test/unit/plugins/ucx/ucx_worker_test.cpp
```diff
@@ -42,6 +43,34 @@ checkCudaError(cudaError_t result, const char *message) {
...
+#define cudaMalloc hipMalloc
+#define cudaFree hipFree
+#define cudaMemset hipMemset
+#define cudaMemcpy hipMemcpy
+#define cudaMemcpyDeviceToHost hipMemcpyDeviceToHost
+#define cudaSuccess hipSuccess
+#define cudaGetErrorString hipGetErrorString
+#define cudaError_t hipError_t
+
+int gpu_id = 0;
```
brminich: why is it needed?

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -40,6 +40,37 @@ static void checkCudaError(cudaError_t result, const char *message) {
...
+#define cudaSuccess hipSuccess
+#define cudaGetErrorString hipGetErrorString
+#define cudaGetDeviceCount hipGetDeviceCount
+
+// Map CUDA types to HIP equivalents
+#define cudaError_t hipError_t
+#define CUdevice hipDevice_t
+#define CUcontext hipCtx_t
+
+int gpu_id = 0;
```
ColinNV: This is never used? (The `gpu_id` in the CUDA section appears unused, too.)

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -40,6 +40,37 @@ static void checkCudaError(cudaError_t result, const char *message) {
 }
 #endif
 
+#ifdef HAVE_ROCM
+
+#include <hip/hip_runtime.h>
+
+// Map CUDA calls to HIP equivalents
```
ColinNV: Doing this via `#define`s looks a bit strange, can the differences between the two be hidden behind a common interface?

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -147,11 +178,12 @@ std::string memType2Str(nixl_mem_t mem_type)
 }
 
 
-#ifdef HAVE_CUDA
+#if defined(HAVE_CUDA) || defined(HAVE_ROCM)
```
ColinNV: Perhaps we could introduce `HAVE_GPU` for when either is defined.

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -174,6 +206,18 @@ static int cudaQueryAddr(void *address, bool &is_dev,
     is_dev = (mem_type == CU_MEMORYTYPE_DEVICE);
 
     return (CUDA_SUCCESS != result);
+#elif defined(HAVE_ROCM)
+    hipPointerAttribute_t attrs;
+    hipError_t result = hipPointerGetAttributes(&attrs, address);
```
ColinNV: `const`

ColinNV (changes requested):

README.md
edgargabriel: The line here has probably been overlooked when nixlbench support for AMD GPU has been merged in PR https://github.com/ai-dynamo/nixl/pull/1647 has been merged.  It is technically incorrect now.

In addition, this line contradicts what we say in line 174 above, which gives explicit instructions how to compile nixlbench for AMD GPUs

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: I think some functions could be extracted out, and have a cuda / hip version (e.g. CudaGetDeviceCount, CudaSetDevice, cudaMalloc, cudeFree, CudaMemset, CudaMemcpy). 

Functions that take a cuda or hip specific argument would not be that easy, we might have to duplicate the function body and call the cuda or the hip version from a HAVE_CUDA or HAVE_ROCM block. See e.g. the interface of `cudaQueryAddr` which takes a  `CUdevice &dev, CUcontext &ctx` as arguments. Without the define statements we would have to define a separate `hipQueryAddr`

There is an alternative solution that would also get rid of the #define statements, namely to use the `hipify-perl` tool as part of the meson build process. This is what we had in RIXL, see e.g. https://github.com/ROCm/RIXL/blob/develop/test/unit/plugins/ucx/meson.build#L26

If you prefer that I use this approach, I am happy to adjust.

test/unit/plugins/ucx/ucx_worker_test.cpp
edgargabriel: In ucx_worker_test.c the gpu_id is actually being used, see line 89. Moved it to the local function, and I don't see any adverse effects.

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: removed gpu_id in this file.

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: @ColinNV  I implemented now the first solution (i.e. having functions that abstract out GPU handling), please have a look when you have a couple of min.

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: we have USE_VRAM that kind-off acts like a HAVE_GPU, since its set for both CUDA and ROCM

meson_options.txt
```diff
@@ -27,6 +27,8 @@ option('cudapath_inc', type: 'string', value: '', description: 'Include path for
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
 option('nixl_cuda_arch_list', type: 'string', value: 'auto', description: 'Comma-separated CUDA SM targets (e.g. 90,100), or auto to select defaults (sm_80+ normally, sm_90+ when build_nixl_ep is enabled). Use a single target like 100 for faster single-GPU builds.')
+option('use_rocm', type: 'boolean', value: false, description: 'Use ROCm/HIP instead of CUDA (mutually exclusive with CUDA)')
```
iyastreb: maybe this can be a single string option, empty by default (meaning rocm is disabled). When non-empty - means rocm is enabled

test/unit/plugins/ucx/meson.build
```diff
@@ -13,13 +13,38 @@
...
+        include_directories : include_directories(rocm_inc_path))
+    if rocm_dep.found()
+        rocm_available = true
+    endif
+endif
+
 ucx_test_dep = [ ucx_dep, nixl_common_deps, ucx_backend_interface]
 
-if cuda_dep.found()
+if rocm_available
```
tvegas1: use `if rocm_dev.found()` everywhere and remove `rocm_available` variable?

test/unit/plugins/ucx/meson.build
```diff
@@ -47,12 +72,12 @@ if get_option('buildtype') != 'release'
                include_directories: ucx_test_include_directories,
                install: true)
 
-    if cuda_dep.found()
+    if cuda_dep.found() or rocm_available
         ucx_worker_cuda_bin = executable('ucx_worker_test_cuda',
```
tvegas1: can we rename to `ucx_worker_test_gpu` and `ucx_worker_gpu_bin`?

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -185,17 +182,24 @@ void allocateBuffer(nixl_mem_t mem_type, int dev_id, size_t len, void* &addr)
     case DRAM_SEG:
         addr = calloc(1, len);
         break;
-#ifdef HAVE_CUDA
+#if defined(HAVE_CUDA) || defined(HAVE_ROCM)
```
tvegas1: maybe use everywhere, defined once, like `#if defined(HAVE_GPU)`?

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -185,17 +182,24 @@ void allocateBuffer(nixl_mem_t mem_type, int dev_id, size_t len, void* &addr)
...
+        gpuSetDevice(dev_id, "Failed to set device");
+        gpuMalloc(&addr, len, "Failed to allocate GPU buffer 0");
+
+#ifdef HAVE_CUDA
         CUdevice dev;
         CUcontext ctx;
-
-        checkCudaError(cudaSetDevice(dev_id), "Failed to set device");
-        checkCudaError(cudaMalloc(&addr, len), "Failed to allocate CUDA buffer 0");
         cudaQueryAddr(addr, is_dev, dev, ctx);
```
tvegas1: can it be gpuQueryAddr() and ctx is defined internally in CUDA implementation only?

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -245,10 +249,10 @@ void *getValidationPtr(nixl_mem_t mem_type, void *addr, size_t len)
     case DRAM_SEG:
         return addr;
         break;
-#ifdef HAVE_CUDA
+#if defined(HAVE_CUDA) || defined(HAVE_ROCM)
```
tvegas1: HAVE_GPU or HAVE_VRAM?

test/unit/plugins/ucx/ucx_worker_test.cpp
```diff
@@ -74,6 +55,9 @@ completeRequest(nixlUcxWorker w[2],
 
 int
 main() {
+#ifdef USE_VRAM
+    int gpu_id = 0;
```
tvegas1: move block below inside existing USE_VRAM

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,94 @@
...
+checkHipError(hipError_t result, const char *message) {
+    if (result != hipSuccess) {
+        std::cerr << message << " (Error code: " << result << " - " << hipGetErrorString(result)
+                  << ")" << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+#endif
+
+static inline void
```
ColinNV: The `static` is not needed on these `inline` functions.

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: done

meson_options.txt
edgargabriel: done, there is now a single rocm_path option, which - if empty- is equivalent of rocm support not being requested

test/unit/plugins/ucx/meson.build
edgargabriel: fixed.

test/unit/plugins/ucx/meson.build
edgargabriel: fixed

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: fixed, introduced a HAVE_GPU macro

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: @tvegas1  please have a look at the current version. I renamed `cudaQueryAddr` and `hipQueryAddr` both to `gpuQueryAddr`, but the interface is still different between the cuda and the hip version. Do you have another solution in mind?

test/unit/plugins/ucx/ucx_worker_test.cpp
edgargabriel: done, block has been moved.

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: removed the `static` keyword for these routines.

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -185,17 +181,24 @@ void allocateBuffer(nixl_mem_t mem_type, int dev_id, size_t len, void* &addr)
...
+#ifdef HAVE_GPU
     case VRAM_SEG:{
         bool is_dev;
+
+        gpuSetDevice(dev_id, "Failed to set device");
+        gpuMalloc(&addr, len, "Failed to allocate GPU buffer 0");
+
+#ifdef HAVE_CUDA
         CUdevice dev;
         CUcontext ctx;
```
tvegas1: move ctx and std::cout log inside gpuQueryAddr? same for ROCm log inside gpuQueryAddr?

tvegas1: and remove ifdef/elif/endif

test/unit/plugins/ucx/ucx_backend_test.cpp
```diff
@@ -211,10 +214,10 @@ void releaseBuffer(nixl_mem_t mem_type, int dev_id, void* &addr)
     case DRAM_SEG:
         free(addr);
         break;
-#ifdef HAVE_CUDA
+#ifdef HAVE_GPU
```
tvegas1: maybe remove HAVE_GPU here, always call gpu*() functions, and gpu*() function call nixl_exit_on_failure() when HAVE_CUDA/HAVE_ROCM are not defined? here and in other cases?

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: done

test/unit/plugins/ucx/ucx_backend_test.cpp
edgargabriel: done, I think I addressed all comments in the last round of reviews.

tvegas1 (dismissed):

ColinNV (dismissed):

edgargabriel:
Just want to point out that the rocm options added in `meson_options.txt` in this PR and in https://github.com/ai-dynamo/nixl/pull/1873 are not entirely aligned, one of the two PRs will have to change, I don't care which one.

ryanhankins:
> Just want to point out that the rocm options added in `meson_options.txt` in this PR and in #1873 are not entirely aligned, one of the two PRs will have to change, I don't care which one.

Happy to change #1873.

edgargabriel:
@tvegas1 @ColinNV gentle ping to see if this PR goood to be merged?

tvegas1 (dismissed):

edgargabriel:
@brminich could you please restart the test one more time? The error is not related to this PR, and it would be really good to get this PR merged this week.

ColinNV:
> @brminich could you please restart the test one more time? The error is not related to this PR, and it would be really good to get this PR merged this week.

Restarted the failed job.

meson_options.txt
```diff
@@ -27,6 +27,7 @@ option('cudapath_inc', type: 'string', value: '', description: 'Include path for
 option('cudapath_lib', type: 'string', value: '', description: 'Library path for CUDA')
 option('cudapath_stub', type: 'string', value: '', description: 'Extra Stub path for CUDA')
 option('nixl_cuda_arch_list', type: 'string', value: 'auto', description: 'Comma-separated CUDA SM targets (e.g. 90,100), or auto to select defaults (sm_80+ normally, sm_90+ when build_nixl_ep is enabled). Use a single target like 100 for faster single-GPU builds.')
+option('rocm_path', type: 'string', value: '', description: 'Path to ROCm installation (default: /opt/rocm if set to "default", empty = do not use ROCm)')
```
aranadive: Lets remove the "default" keyword here and rely explicitly on the path passed so its consistent across all options.

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,154 @@
...
+    attr_data[1] = &is_managed;
+    attr_type[2] = CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL;
+    attr_data[2] = &cu_dev;
+    attr_type[3] = CU_POINTER_ATTRIBUTE_CONTEXT;
+    attr_data[3] = &cu_ctx;
+
+    result = cuPointerGetAttributes(4, attr_type, attr_data, (CUdeviceptr)address);
+
+    is_dev = (mem_type == CU_MEMORYTYPE_DEVICE);
+    dev = cu_dev;
```
aranadive: Should check the result before assigning and printing

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,154 @@
...
+#elif defined(HAVE_ROCM)
+    checkHipError(hipMemset(ptr, value, size), message);
+    checkHipError(hipStreamSynchronize(0), "Failed to synchronize stream");
+#else
+    nixl_exit_on_failure(false, message);
+#endif
+}
+
+inline void
+gpuGetDeviceCount(int *count, const char *message) {
```
aranadive, line 84: message isnt consumed here. Maybe print the error if the count is 0 before returning

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,154 @@
...
+}
+
+inline int
+gpuQueryAddr(void *address, bool &is_dev, int &dev) {
+#ifdef HAVE_CUDA
+    CUdevice cu_dev;
+    CUcontext cu_ctx;
+    CUmemorytype mem_type = CU_MEMORYTYPE_HOST;
+    uint32_t is_managed = 0;
+#define NUM_ATTRS 4
```
aranadive: constexpr int num_attrs = 4;

ColinNV: It's a header, should be `inline constexpr`.

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,154 @@
...
+
+#include <iostream>
+#include <cstdlib>
+#include "test_utils.h"
+
+#ifdef HAVE_CUDA
+#include <cuda_runtime.h>
+#include <cuda.h>
+
+static void
```
ColinNV: Is there a reason these are `static` and not `inline` like the other functions?

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,154 @@
...
+        std::cerr << message << " (Error code: " << result << " - " << hipGetErrorString(result)
+                  << ")" << std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+#endif
+
+inline void
+gpuSetDevice(int device, const char *message) {
+#ifdef HAVE_CUDA
```
ColinNV: Another minor issue, can we use `#if defined( HAVE_CUDA )` to make it consistent with the `#elif`?

ColinNV (commented):
LGTM

meson_options.txt
edgargabriel: done

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: fixed

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: added a cout of the message if no GPUs are found, in all three compilation pathes

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: fixed

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: changed to inline

edgargabriel:
@ColinNV @aranadive I think I addressed all review comments in this round, and I have also run the tests on one of our systems to confirm that it still works. If you have a chance, can you please have a look again?

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,163 @@
...
+}
+
+inline int
+gpuQueryAddr(void *address, bool &is_dev, int &dev) {
+#if defined(HAVE_CUDA)
+    CUdevice cu_dev;
+    CUcontext cu_ctx;
+    CUmemorytype mem_type = CU_MEMORYTYPE_HOST;
+    uint32_t is_managed = 0;
+    inline constexpr int num_attrs = 4;
```
aranadive: Need to drop the inline

test/unit/plugins/ucx/gpu_utils.h
```diff
@@ -0,0 +1,163 @@
...
+
+    if (result == CUDA_SUCCESS) {
+        is_dev = (mem_type == CU_MEMORYTYPE_DEVICE);
+        dev = cu_dev;
+
+        std::cout << "CUDA addr: " << std::hex << address << " dev=" << std::dec << dev
+                  << " ctx=" << std::hex << cu_ctx << std::dec << std::endl;
+        return 0;
+    }
+    return 1;
```
aranadive, line 147: Can we return the error here and check/log it at point of invocation of gpuQueryAddr? Same for the hip side

test/unit/plugins/ucx/gpu_utils.h
ColinNV: @edgargabriel This is making the builds fail.

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: I would prefer to keep it this way, since the CUDA/HIP/no GPU supported error message is a bit more useful than a generic message about something not having worked.

test/unit/plugins/ucx/gpu_utils.h
edgargabriel: fixed

edgargabriel:
@ColinNV  the CI error is not related to this PR, can we maybe restart it one more time?

aranadive (approved):

PR #1790, opened 2026-06-17
fix(plugins): require a unique devId per path-mode file

### LINKED ISSUE #1766: [BUG] In file-path mode, a double free can occur
Path mode from PR https://github.com/ai-dynamo/nixl/pull/1635 introduces a path to a bug.  If two exact same regions get registered twice: they collapse to the same sort keys so the sequence: register(A), register(A), deregister(A), deregister(A) causes a double free on the second deregister.

The bug:

```
nixlBlobDesc file1, file2;

file1.addr = 0;
file1.size=4096;
file1.metaInfo = std::string("rw:file1.bin");
file2.addr = 0;
file2.size = 4096;
file2.metainfo = std::string("rw:file2.bin");

nixl_reg_dlist_t file_descs(FILE_SEG)
file_descs.addDesc(file1);
file_descs.addDesc(file2);

agent.registerMem(file_descs);

agent.deregisterMem(file_descs); <-- will cause a double free of file1
```

The workaround:
```
nixlBlobDesc file1, file2;

file1.devId = 1;
file1.addr = 0;
file1.size=4096;
file1.metaInfo = std::string("rw:file1.bin");
file2.devId = 2; <-- unique devId per file
file2.addr = 0;
file2.size = 4096;
file2.metainfo = std::string("rw:file2.bin");

nixl_reg_dlist_t file_descs(FILE_SEG)
file_descs.addDesc(file1);
file_descs.addDesc(file2);

agent.registerMem(file_descs);

agent.deregisterMem(file_descs);
```

The workaround is to use a unique devId per file.  This will keep the io descriptors unique so the behavior will not occur.

lluki:
## Summary

As of now, we dont require unique devIds for that path mode. That can lead to ambiguity:

```
reg([ (0,2000, devid=0, meta="/tmp/a"), (1000,3000, devid=0, meta="/tmp/b") ]) 
transfer( (1000,1000,devid=0), .... )
```
For that transfer it's unclear which file is read (`/tmp/a` or `/tmp/b` ?).

This PR enforces a **unique devId per path-mode file** in each FILE_SEG path-mode backend. It is a focused subset of the registration-time approach discussed in #1744 -- it does **not** touch the common section layer (`nixlLocalSection::addDescList`); the duplicate-descriptor rejection is intentionally out of scope here.

It fixes the double-free & leak problem for path mode, but leaves the problem present for other backends: Fixes https://github.com/ai-dynamo/nixl/issues/1766

## Fix

- New shared helper `nixl::PathModeDevIdRegistry` in `src/utils/file/file_path_mode.h`: a per-engine set of in-use path-mode devIds with an RAII `reserve()` that auto-rolls-back if registration fails before `commit()`. fd-in-devId mode (non-path `metaInfo`) is untracked and unaffected.
- All four FILE_SEG path-mode backends (`posix`, `cuda_gds`, `gds_mt`, `hf3fs`) reserve on register, reject a reused path-mode devId with `NIXL_ERR_INVALID_PARAM`, and release on deregister. The devId is stored on the file MD so deregister can release it.
- No backend lock added: `registerMem` / `deregisterMem` are already serialized by the agent lock.

## Changes

- `src/utils/file/file_path_mode.h`: `PathModeDevIdRegistry` helper; `nixlFilePathMD` gains a `(devId, metaInfo)` ctor and stores `devId`; path-mode unique-devId contract documented.
- `src/plugins/{posix,cuda_gds,gds_mt,hf3fs}`: per-engine path-mode devId registry; devId set via the MD ctor; release on deregister.

## Testing

`test/unit/plugins/posix/nixl_posix_test.cpp`:
- `runPathModeUniqueDevIdCheck`: within-list reuse rejected, cross-call reuse rejected, devId freed on deregister (reusable afterwards), distinct devIds register together. Addresses differ across the colliding descriptors so the rejection is driven by the devId rule, not by any duplicate-descriptor check.
- `runFdModeReuseCheck`: fd-mode (devId is a real fd) is **not** subject to the rule -- one fd may back several descriptors at different offsets.

Built and ran the POSIX unit suite (POSIX-only, `build_tests=true`): all checks pass, including the two above. `cuda_gds` + `gds_mt` compile-checked; `hf3fs` reviewed by inspection (no client dev libs locally).

## Summary

* **New Features**
  * Path-mode `FILE_SEG` registrations now reserve and enforce uniqueness of `(devId, metaInfo)` consistently across POSIX, CUDA GDS, GDS MT, and HF3FS, rejecting duplicates with `NIXL_ERR_INVALID_PARAM`.
  * Path-mode metadata now carries `devId`, improving deregistration so reserved mappings are released only for actual path-backed segments.

* **Tests**
  * Expanded POSIX unit coverage to validate unique `devId`/`metaInfo` enforcement within and across separate registrations, plus `fd`-mode reuse behavior.

### DIFF
--- a/src/plugins/gds_mt/gds_mt_backend.cpp
+++ b/src/plugins/gds_mt/gds_mt_backend.cpp
@@ -43,8 +43,11 @@ const size_t default_thread_count = std::max (1u, std::thread::hardware_concurre
 
 struct FileSegData {
     std::shared_ptr<gdsMtFileHandle> handle;
+    uint64_t devId;
 
-    FileSegData(std::shared_ptr<gdsMtFileHandle> h) : handle(std::move(h)) {}
+    FileSegData(std::shared_ptr<gdsMtFileHandle> h, uint64_t devid)
+        : handle(std::move(h)),
+          devId(devid) {}
 };
 
 struct MemSegData {
@@ -74,9 +77,9 @@ struct GdsMtTransferRequestH {
 
 class nixlGdsMtMetadata : public nixlBackendMD {
 public:
-    explicit nixlGdsMtMetadata (std::shared_ptr<gdsMtFileHandle> file_handle)
-        : nixlBackendMD (true),
-          data_ (FileSegData{std::move (file_handle)}) {}
+    nixlGdsMtMetadata(std::shared_ptr<gdsMtFileHandle> file_handle, uint64_t devid)
+        : nixlBackendMD(true),
+          data_(FileSegData{std::move(file_handle), devid}) {}
 
     explicit nixlGdsMtMetadata (void *addr, size_t size, int flags)
         : nixlBackendMD (true),
@@ -189,6 +192,12 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
                               nixlBackendMD *&out) {
     switch (nixl_mem) {
     case FILE_SEG: {
+        auto resv = path_mode_devids_.reserve(mem.devId, mem.metaInfo);
+        if (!resv.ok()) {
+            NIXL_ERROR << "GDS_MT: path-mode requires a unique devId per file (devId=" << mem.devId
+                       << " already registered)";
+            return NIXL_ERR_INVALID_PARAM;
+        }
         nixl::FileFd file_fd;
         try {
             file_fd = nixl::FileFd(mem.devId, mem.metaInfo);
@@ -217,7 +226,8 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
             }
             gds_mt_file_map_[fd] = handle;
         }
-        out = new nixlGdsMtMetadata(std::move(handle));
+        out = new nixlGdsMtMetadata(std::move(handle), mem.devId);
+        resv.commit();
         return NIXL_SUCCESS;
     }
 
@@ -254,6 +264,9 @@ nixlGdsMtEngine::deregisterMem (nixlBackendMD *meta) {
     if (auto *file_data = std::get_if<FileSegData> (&md->data_)) {
         if (file_data->handle) {
             int key = file_data->handle->file_fd.fd();
+            if (!file_data->handle->file_fd.path().empty()) {
+                path_mode_devids_.release(file_data->devId);
+            }
             md.reset(); // Release metadata first
 
             auto it = gds_mt_file_map_.find (key);
--- a/src/plugins/posix/posix_backend.cpp
+++ b/src/plugins/posix/posix_backend.cpp
@@ -232,13 +232,20 @@ nixlPosixEngine::registerMem(const nixlBlobDesc &mem,
     if (std::find(supported_mems.begin(), supported_mems.end(), nixl_mem) != supported_mems.end()) {
         out = nullptr;
         if (nixl_mem == FILE_SEG) {
+            auto resv = path_mode_devids_.reserve(mem.devId, mem.metaInfo);
+            if (!resv.ok()) {
+                NIXL_ERROR << "POSIX path-mode requires a unique devId per file (devId="
+                           << mem.devId << " already registered)";
+                return NIXL_ERR_INVALID_PARAM;
+            }
             try {
-                out = new nixlPosixFileMD(nixl::FileFd(mem.devId, mem.metaInfo));
+                out = new nixlPosixFileMD(mem.devId, mem.metaInfo);
             }
             catch (const std::system_error &e) {
                 NIXL_ERROR << "POSIX path-mode open failed: " << e.what();
                 return NIXL_ERR_BACKEND;
             }
+            resv.commit();
         }
         return NIXL_SUCCESS;
     }
@@ -248,6 +255,14 @@ nixlPosixEngine::registerMem(const nixlBlobDesc &mem,
 
 nixl_status_t
 nixlPosixEngine::deregisterMem(nixlBackendMD *meta) {
+    // non-null meta is always a file MD. Release the path-mode reservation (path() empty in
+    // fd-mode)
+    if (meta) {
+        auto *file_md = static_cast<nixlPosixFileMD *>(meta);
+        if (!file_md->file_fd.path().empty()) {
+            path_mode_devids_.release(file_md->devId);
+        }
+    }
     delete meta;
     return NIXL_SUCCESS;
 }
--- a/src/utils/file/file_path_mode.h
+++ b/src/utils/file/file_path_mode.h
@@ -14,13 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef __FILE_PATH_MODE_H
-#define __FILE_PATH_MODE_H
+#ifndef NIXL_SRC_UTILS_FILE_FILE_PATH_MODE_HPP
+#define NIXL_SRC_UTILS_FILE_FILE_PATH_MODE_HPP
 
 #include <sys/types.h>
 
+#include <cstdint>
 #include <optional>
 #include <string>
+#include <unordered_set>
 
 #include "backend/backend_engine.h"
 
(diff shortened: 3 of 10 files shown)


test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -843,6 +843,154 @@ runPathModeCreateCheck() {
...
+    {
+        nixl_reg_dlist_t d(FILE_SEG);
+        d.addDesc(pathDesc(kFileA, 0, 0));
+        d.addDesc(pathDesc(kFileB, 0, 4096));
+        if (agent.registerMem(d) == NIXL_SUCCESS) {
+            agent.deregisterMem(d);
+            std::cerr << "path-mode duplicate devId (within list) was not rejected" << std::endl;
+            cleanup();
+            return 1;
+        }
```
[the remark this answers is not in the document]
lluki: Not true, the agent turns it into a generic error code before returning

src/plugins/posix/posix_backend.cpp
```diff
@@ -248,6 +255,14 @@ nixlPosixEngine::registerMem(const nixlBlobDesc &mem,
 
 nixl_status_t
 nixlPosixEngine::deregisterMem(nixlBackendMD *meta) {
+    // non-null meta is always a file MD. Release the path-mode reservation (path() empty in
+    // fd-mode)
+    if (meta) {
+        auto *file_md = static_cast<nixlPosixFileMD *>(meta);
+        if (!file_md->file_fd.path().empty()) {
+            path_mode_devids_.release(file_md->devId);
```
ofer, line 263: why not pass in the file_md to the registry and release that? they all have the same if statement...

lluki: Hm, they have different types, for example in h3fs we have `nixlHf3fsFileMetadata` and not `nixlPosixFileMD`

ofer: but they both inherit the nixlFileMD

lluki: There are four cases, posix and gds inherit. but gds_mt and hf3fs do not... I can check how much change it would be to make them

lluki: I think especially GDS_MT is problematic, as it does FD sharing and it would conflict with the RAII logic of nixlPosixFileMD

src/plugins/gds_mt/gds_mt_backend.cpp
```diff
@@ -217,7 +226,8 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
             }
             gds_mt_file_map_[fd] = handle;
         }
-        out = new nixlGdsMtMetadata(std::move(handle));
+        resv.commit();
+        out = new nixlGdsMtMetadata(std::move(handle), mem.devId);
```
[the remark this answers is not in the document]
lluki: done

src/utils/file/file_path_mode.h
```diff
@@ -89,15 +94,89 @@ class FileFd {
...
     nixl::FileFd file_fd;
+    uint64_t devId = 0;
 
     nixlFilePathMD() : nixlBackendMD(true /*isPrivate*/) {}
 
-    explicit nixlFilePathMD(nixl::FileFd &&fd) : nixlBackendMD(true), file_fd(std::move(fd)) {}
+    nixlFilePathMD(uint64_t devid, const std::string &metaInfo)
+        : nixlBackendMD(true),
+          file_fd(devid, metaInfo),
+          devId(devid) {}
```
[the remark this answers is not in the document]
lluki: assertion is not correct, this would be a caller error. But there should probably an error on fd-mode registration with fd > INT_MAX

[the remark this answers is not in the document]
lluki: Let's ignore this for now, this is not new in this PR (nor is it new in the path-mode).

src/plugins/posix/posix_backend.cpp
lluki: Another good reason to keep it as is: We need to fix the same in the OBJ backend https://github.com/ai-dynamo/nixl/issues/1794 - What i would do to fix that bug too is promote `PathModeDevIdRegistry ` to `DevIdRegistry` and use it in the OBJ backends as well.

src/utils/file/file_path_mode.h
ColinNV: While you're at it, `NIXL_SRC_UTILS_FILE_FILE_PATH_MODE_HPP`.

src/utils/file/file_path_mode.h
```diff
@@ -89,15 +94,89 @@ class FileFd {
...
+        operator=(const Reservation &) = delete;
+
+        ~Reservation() {
+            if (reg_ && held_) {
+                reg_->release(devId_);
+            }
+        }
+
+        // false: this devId is already registered in path-mode -> caller must reject.
+        bool
```
ColinNV, line 125: `[[nodiscard]]`

src/utils/file/file_path_mode.h
```diff
@@ -89,15 +94,89 @@ class FileFd {
...
+        uint64_t devId_;
+        bool ok_;
+        bool held_;
+    };
+
+    // fd-mode (metaInfo not path-mode) is always ok and untracked.
+    Reservation
+    reserve(uint64_t devId, const std::string &metaInfo) {
+        if (!parsePathMeta(metaInfo)) {
+            return Reservation(this, devId, true, false);
```
ColinNV: Please add `enum class : bool` types to replace the `bool`s.

ColinNV (dismissed):

src/utils/file/file_path_mode.h
lluki: done

src/utils/file/file_path_mode.h
lluki: done

src/utils/file/file_path_mode.h
lluki: done

ovidiusm (approved):

src/plugins/gds_mt/gds_mt_backend.cpp
```diff
@@ -217,7 +226,8 @@ nixlGdsMtEngine::registerMem (const nixlBlobDesc &mem,
             }
             gds_mt_file_map_[fd] = handle;
         }
-        out = new nixlGdsMtMetadata(std::move(handle));
+        out = new nixlGdsMtMetadata(std::move(handle), mem.devId);
+        resv.commit();
```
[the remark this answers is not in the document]
ovidiusm: @lluki this seems relevant, could you please check?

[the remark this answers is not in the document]
lluki: My PR doesn't change the behavior in the case where the constructor throws an exception: The exception is forwarded to the caller. The newly introduced reservation mechanism would work correctly in that case.

The constructor is also trivial, the only way it could throw by failing the allocation in `new`. We probably have many of these cases in the codebase.

ofer (approved):

src/plugins/posix/posix_backend.cpp
ofer: We could have an interface for a DevIdMetadata of some sort so the RAII works for both cases

ofer: and have a case where it doesn't do the release like for GDS_MT

lluki: I think we should move the other way: The registration should be unaware of the surrounding types. It should be only concerned about the management of `devIds` and nothing else. The fact that it's used together with File Metadata is just a particularity of that backend. The object backend should also use it, and OBJ doesn't know anything about files. 

The best approach in my opinion is to have in the agent the ability to prevent double `devId` registration. So for example when `PosixBackend` gets instantiated, it could tell the agent somehow "I require unique devIds, do not allow duplicates". Then the agent would enforce it. Once for all backends. The agent already queries information of the backend (for example using `getSupportedMems()` ). 

~~Actually: Management in the agent is the only correct solution. Consider the case where i have File and Object backend active at the same time, they can register the same devId and then xfer using solely it~~ (this is actually not true, we have the mem type also in the xfer)

ovidiusm: I will merge it as is since we need to release the bugfix, in case you think there is still something to refactor please do that in a follow-up. Appreciate the review 👍

PR #1799, opened 2026-06-19
examples/ep: support configurable NVL group size

xtyao66:
## What?

This draft PR adds runtime-configurable NVLink-local group sizing to the NIXL EP high-throughput path.

Today the HT EP topology assumes an 8-rank local NVLink group:

```text
rdma_rank = rank / 8
nvl_rank  = rank % 8
```

This PR introduces `nvl_group_size`, defaulting to `8`, and uses it for rank mapping:

```text
rdma_rank = rank / nvl_group_size
nvl_rank  = rank % nvl_group_size
```

The main target is GB200/NVL72-style deployments where each worker or pod exposes 4 local GPUs. In that case, `nvl_group_size=4` keeps CUDA IPC local to the 4-GPU NVLink-visible group while routing inter-group traffic through the EP RDMA/fabric path.

The PR also moves CUDA VMM fabric-memory helpers into reusable NIXL utilities and exposes Python bindings/tests for support detection and allocation/free coverage.

## Why?

The current hard-coded 8-rank grouping is correct for deployments where each worker has 8 CUDA-IPC-local GPUs, but it is too rigid for GB200 deployments that expose fewer local GPUs per worker.

With 4 GPUs per worker, the old mapping incorrectly treats ranks `0..7` as one local NVLink group. That can cause the HT EP path to open CUDA IPC peer views across worker boundaries, even though those peers should be reached through RDMA/fabric buffers instead.

Making the local group size explicit lets the same EP HT implementation support both:

- existing 8-GPU-local deployments with the unchanged default `nvl_group_size=8`;
- 4-GPU-per-worker GB200 deployments with `nvl_group_size=4`.

## How?

- Adds `nvl_group_size` to `nixl_ep.Buffer`, defaulting to `8` for backward compatibility.
- Generalizes HT rank mapping and availability checks from the fixed `8`-rank local group to the runtime `nvl_group_size`.
- Keeps `NUM_MAX_NVL_PEERS=8` as the fixed kernel/scratch-layout maximum, while using runtime active-lane guards for smaller local groups.
- Opens CUDA IPC peer views only within the active NVLink-local group.
- Routes inter-group traffic through the EP RDMA/fabric path.
- Updates HT layout/config buffer-size hints to account for `nvl_group_size`.
- Adds `test_ht_gb200.py` to exercise GB200-style `nvl_group_size=4` layouts.
- Moves CUDA fabric VMM allocation helpers from the EP example tree into reusable `src/utils/cuda` utilities.
- Adds Python bindings and tests for CUDA fabric VMM support detection and allocation/free behavior.

## Compatibility

- The default remains `nvl_group_size=8`, so existing 8-GPU-local EP users should not need to change code.
- `NUM_MAX_NVL_PEERS` remains `8`; this PR changes the number of active local peers, not the maximum scratch layout.
- The new `nvl_group_size` must be positive, no larger than `8`, and compatible with the total rank count.
- This does not claim arbitrary topology support. It supports equal-size, rank-contiguous local groups such as `0..3`, `4..7`, `8..11`, etc.

## Non-goals / Limitations

- This is not a universal AWS or Kubernetes placement solution. Non-contiguous rank-to-GPU placement would need a future explicit rank-to-group mapping API.
- This does not replace the existing 8-GPU default.
- This does not make generic NIXL UCX `WRITE` transfers the production EP data path. The focus here is EP HT dispatch/combine topology.
- This does not claim performance parity across all GB200 deployment shapes; it adds the topology support and validates several `nvl_group_size=4` layouts.

## Validation

Local/static checks:

- `git diff --cached --check`: pass
- Python compile with redirected pycache: pass

GB200 EP HT validation with `nvl_group_size=4`:

| Layout | Result | FP8 Dispatch Median | BF16 Dispatch Median | Combine Median |
|---|---|---:|---:|---:|
| `1x4` | Expected boundary skip: no remote RDMA group | n/a | n/a | n/a |
| `2x4` | Pass | `2.56 ms` | `4.80 ms` | `3.98 ms` |
| `3x4` | Pass | `4.74 ms` | `9.02 ms` | `4.32 ms` |
| `4x4` | Pass | `6.63 ms` | `12.65 ms` | `4.62 ms` |
| `5x4` | Pass | `7.54 ms` | `14.29 ms` | `4.53 ms` |
| `6x4` | Pass | `8.04 ms` | `15.29 ms` | `4.13 ms` |

Additional observations:

- `3x4`, `5x4`, and `6x4` required rounding the test `nvl_buffer_size` so it is divisible by the runtime RDMA group count.
- The `1x4` case is intentionally treated as a boundary condition because `nvl_group_size=4` with one local group has no remote RDMA group to exercise in HT dispatch/combine.
- The table above is summarized from local GB200 run artifacts. Raw logs can be shared if maintainers want deeper timing or rank-level diagnostics.

## Reviewer Notes

This is intentionally opened as a draft because it combines two related but separable pieces:

1. EP HT topology support via `nvl_group_size`.
2. Reusable CUDA fabric VMM utilities and Python bindings/tests.

If maintainers prefer, I can split this into two PRs. The EP topology piece is the core functional change for GB200 `4`-GPU-local groups; the VMM utilities are included because they are useful for GB200-safe fabric-memory validation and remove EP-local duplication.

Areas where reviewer feedback would be especially useful:

- whether `nvl_group_size` is the right public API, or whether NIXL should move directly to an explicit rank-to-local-group mapping;
- whether the CUDA VMM utilities should live in this PR or be split out;
- whether additional CI coverage should be added for non-8 local group sizes without requiring GB200 hardware.

## Summary by CodeRabbit

* **New Features**
  * Added configurable NVL group sizing for flexible deployment on various GPU configurations, including specialized GB200 multi-node high-throughput setups.
  * Introduced debug diagnostics for barrier, atomic, and remote-put operations.
  * Expanded CUDA fabric memory management capabilities.

* **Documentation**
  * Updated guides for GB200 multi-node deployment with NVL grouping parameters.

* **Refactor**
  * Enhanced buffer initialization and rank mapping for dynamic NVL grouping support.


xtyao66:
Superseded by the clean replacement PR #1820. The new PR keeps the core scope to configurable rank-contiguous NVL groups and removes the old mixed debug/VMM scope.

PR #1811, opened 2026-06-23
Fixed race conditions with RW lock

iyastreb:
## What?
In some error handling cases NIXL may modify shared data while holding the reader lock.
This lead to race condition

## Release Notes

* **Bug Fixes**
  * Improved reliability when remote connections drop by invalidating remote cached data using generation-aware checks, including during transfer cleanup and transfer status handling.
  * Transfer requests now preserve remote generation metadata to ensure invalidation targets the correct cached state.

* **Chores**
  * Strengthened synchronization correctness with clearer shared-lock handling and added lock-held assertions.
  * Remote memory sections now track and expose a generation value to support precise cache invalidation when generations mismatch or are missing.


src/core/sync.h
```diff
@@ -69,16 +78,48 @@ class nixlLock {
...
+public:
+    explicit nixlLockUpgradeGuard(nixlLock &lock) : lock_(lock) {
+        lock_.assertHeld(false);
+        lock_.unlock_shared();
+        lock_.lock();
+    }
+
+    ~nixlLockUpgradeGuard() {
+        lock_.unlock();
+        lock_.lock_shared();
```
tomerg-nvidia: This can be dangerous, if you upgrade the lock mid-function and then the data changes when the inner scope changes

src/core/nixl_agent.cpp
```diff
@@ -1152,6 +1154,7 @@ nixlAgent::getXferStatus (nixlXferReqH *req_hndl) const {
         req_hndl->status = req_hndl->engine->checkXfer(req_hndl->backendHandle);
         if (req_hndl->status < 0) {
             if (req_hndl->status == NIXL_ERR_REMOTE_DISCONNECT) {
+                NIXL_UPGRADE_LOCK_GUARD(data->lock);
```
rakhmets: I think the problem here is not the wrong type of lock, but the implementation of this function.
I think the design is wrong here. The method that returns the status of the request shouldn't invalidate the data. It should be done only via `invalidateRemoteMD`.

rakhmets: Invalidating remote data in posting function also seems to be wrong.

src/core/nixl_agent.cpp
```diff
@@ -1109,6 +1110,7 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
 
     if (req_hndl->status < 0) {
         if (req_hndl->status == NIXL_ERR_REMOTE_DISCONNECT) {
+            NIXL_UPGRADE_LOCK_GUARD(data->lock);
```
ColinNV: After the log to reduce the scope of the exclusive lock?

src/core/sync.h
ColinNV: And perhaps "upgrade" is not the correct name because on a technical level it doesn't upgrade a shared lock to an exlusive lock because we don't have a true "upgrade" operation available.

.ci/jenkins/pipeline/proj-jjb.yaml
```diff
@@ -434,7 +434,7 @@
       - string:
```
ovidiusm, line 501: There were some rollbacks on main that left these stale files. @iyastreb Could you please remove the unrelated changes?

src/core/sync.h
nvjullin: I agree. If I were to see the name "upgrade", I'd assume it means atomically upgrading without releasing the lock.
On another note, I'm not sure how safe it is to lock in a destructor, which can block indefinitely, considering that destructors are called on stack unwinding.

iyastreb: Yes, it was misleading, I removed that RAII wrapper entirely

src/core/nixl_agent.cpp
iyastreb: Well, it was done automatically, but not relevant anymore

src/core/nixl_agent.cpp
iyastreb: Maybe it's not ideal, but I'm afraid to change current behavior
I think it's ok to drop stale connections in posting, because posting does not have to be READ-only.
Anyway, this would result in behavior change and clients code change. This sounds like a large refactoring

src/infra/mem_section.h
```diff
@@ -198,13 +198,19 @@ class nixlLocalSection : public nixlMemSection {
 class nixlRemoteSection : public nixlMemSection {
     private:
         std::string agentName;
+        uint64_t generation_;
 
         nixl_status_t addDescList (
                            const nixl_reg_dlist_t &mem_elms,
                            nixlBackendEngine *backend);
```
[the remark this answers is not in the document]
iyastreb: not relevant

src/core/agent_data.h
```diff
@@ -116,7 +116,7 @@ class nixlAgentData {
         nixl_status_t
         loadRemoteSections(const std::string &remote_name, nixlSerDes &sd);
         nixl_status_t
-        invalidateRemoteData(const std::string &remote_name);
+        invalidateRemoteData(const std::string &remote_name, uint64_t generation = 0);
```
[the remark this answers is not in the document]
iyastreb: ok

src/core/nixl_agent.cpp
rakhmets: I don't think that it will require clients code change. It is already assumed that the clients code should free up resources. I don't think that anyone relies on the fact that `getXferStatus` invalidates the data in some cases.
I would prefer to make a cleaner solution then workaround over this design.

rakhmets: And I'm not sure that posting should do this (invalidating remote md) as well. Moreover, `postXferReq` (and `getXferStatus`) is declared as `const` method.

iyastreb: Well, clients can't even handle this error right now, because NIXL python API does not even differentiate the erors:
```
    def check_xfer_state(self, handle: nixl_xfer_handle) -> str:
        status = self.agent.getXferStatus(handle._handle)
        if status == nixlBind.NIXL_SUCCESS:
            return "DONE"
        elif status == nixlBind.NIXL_IN_PROG:
            return "PROC"
        else:
            return "ERR"
```
For NIXL python clients they know that something bad happened, but they don't even know what exactly. And of course they never invalidate a remote connection.

What you proposed is maybe doable, but it needs to be properly propagated to the client, and properly handled by the client. This is a large change, and change of existing behavior

The aim of this PR is to fix the race condition that occurs with RW lock, I don't plan to redesign error handling part as of now

rakhmets: I meant that invalidation is already covered by `remove_remote_agent`/`invalidateRemoteMD`.
Why do we need to do it in `check_xfer_state`/`getXferStatus`?

iyastreb: TBH I can't answer why it was done like that.

`remove_remote_agent` exists, but it's invoked only if/when client application itself detects that communication is broken somehow (maybe by using keepalive mechanism).
How they can know that NIXL/UCX communication channel is broken?
They can't know this with current API and error handling

rakhmets: How urgent is this fix?
Is it possible, and does it make sense to do a hot fix for the release branch, and do a design review for the invalidation flow, and make a more stable solution for the main branch?

nvjullin: The current requirement of using STRICT locking affects **all** disagg runs, so this has an extremely large blast radius.
For context, SGLang recently fixed a bug where sync mode was NONE. 
Now that we're using STRICT, we're worried about regressions.
While there has been no perf regressions observed in SGLang yet, we'd be a lot less worried if there's a fix ready on NIXL side.
It would be great if a post release with quick fixes can be made (currently SGLang is on v1.2.0, upgrading to v1.3.0 is possible).

iyastreb: We need a fix now for what we provide to clients.
Redesign may be done later, if we find a clear plan how to implement it

src/core/nixl_listener.cpp
```diff
@@ -792,26 +792,30 @@ nixlAgentData::loadRemoteSections(const std::string &remote_name, nixlSerDes &sd
...
     if (remote_name == name_) {
         NIXL_ERROR << "Agent " << name_ << " cannot invalidate itself";
         return NIXL_ERR_INVALID_PARAM;
     }
 
-    nixl_status_t ret = NIXL_ERR_NOT_FOUND;
-    if (remoteSections_.erase(remote_name) > 0) {
-        ret = NIXL_SUCCESS;
+    const auto sec_it = remoteSections_.find(remote_name);
+    if (sec_it == remoteSections_.end() || generation == 0 ||
```
ColinNV: The `generation == 0` check is superfluous since `getGeneration()` will always be positive (generation starts at `1`; if we assume the `uint64_t` to never wrap).

src/core/nixl_agent.cpp
```diff
@@ -775,7 +775,8 @@ nixlAgent::makeXferReq (const nixl_xfer_op_t &operation,
 
     NIXL_SHARED_LOCK_GUARD(data->lock);
     // The remote was invalidated in between prepXferDlist and this call
-    if (data->remoteSections_.count(remote_side->remoteAgent) == 0) {
+    auto rem_sec_it = data->remoteSections_.find(remote_side->remoteAgent);
```
ColinNV: `const`

src/core/transfer_request.h
```diff
@@ -69,6 +69,8 @@ class nixlXferReqH {
     nixl_meta_dlist_t targetDescs;
 
     const std::string remoteAgent;
+    // Generation of the remote-connection
+    uint64_t remoteGeneration_ = 0;
```
ColinNV: Initialize in c'tor and make `const`?

src/core/nixl_agent.cpp
iyastreb: done

src/core/nixl_listener.cpp
iyastreb: right

src/core/transfer_request.h
iyastreb: good idea, done

src/infra/mem_section.h
```diff
@@ -198,13 +198,19 @@ class nixlLocalSection : public nixlMemSection {
 class nixlRemoteSection : public nixlMemSection {
     private:
         std::string agentName;
+        uint64_t generation_;
```
ColinNV, line 203: Perhaps a comment for "future me" to explain?

iyastreb: added

ColinNV (approved):

src/core/nixl_agent.cpp
rakhmets: I see that the quick fix is needed. However, I still don't understand why we need a workaround for this implementation of `getXferStatus` (and `postXferReq`) in the main branch. I still think that this type of lock is not needed at all in this method. Which makes the changes unnecessary.

PR #1824, opened 2026-06-24
telemetry: add last-operation byte gauges to exporters

e-eygin:
## What?

Adds two purely-additive last-operation gauges to both the Prometheus and DOCA
telemetry exporters:

- `agent_tx_last_bytes` — byte size of the most recent TX request
- `agent_rx_last_bytes` — byte size of the most recent RX request

They sit alongside the existing cumulative byte counters
(`agent_tx_bytes_total` / `agent_rx_bytes_total`): the counter reports the
lifetime running total, the gauge reports only the latest operation.

Naming follows Prometheus/OpenMetrics convention — the base unit stays the
**terminal** suffix, so the gauge is `agent_<subject>_last_<unit>` (the `_last`
qualifier precedes `_bytes`), not `_bytes_last`. The convention
(`agent_<subject>_<unit>_total` for counters, `agent_<subject>_last_<unit>` for
last-op gauges) is documented in `docs/telemetry.md`.

Implementation:
- **Prometheus** (`prometheus_exporter.{h,cpp}`): `registerGauge` separates the
  lookup key (event name) from the exposed metric name, registering
  `agent_tx_bytes → agent_tx_last_bytes` and `agent_rx_bytes → agent_rx_last_bytes`.
  `exportEvent()` is unchanged (it already sets the gauge keyed by event name).
- **DOCA** (`doca_exporter.cpp`): byte events now emit BOTH a cumulative counter
  and a last-operation gauge (dispatch was previously counter-xor-gauge);
  `appendGaugeSample` takes an explicit metric-name argument so the gauge is
  named `agent_<dir>_last_bytes` while keyed off the byte event.
- Tests + docs (Prometheus/DOCA READMEs, `docs/telemetry.md`) updated in the
  same PR.

## Why?

The byte events were routed through the counter path only, so there was no way
to see the size of the latest transfer — only the lifetime total. A PromQL
`idelta(*_total[...])` derivation answers a different question (bytes per scrape
interval) and is wrong when the op-rate exceeds the scrape-rate, since NIXL
batches/flushes events (~100 ms). An explicit last-operation gauge is the correct
primitive.

This is **zero breaking changes**: no new telemetry event type, no
`TELEMETRY_VERSION` bump, no core change. "Last" is derived exporter-side from
the per-op value NIXL already carries (`event.value_`), which is delta-only, so
the gauge is stateless. The existing memory gauges are intentionally left under
their current names (`agent_memory_registered` / `agent_memory_deregistered`);
renaming those to the `_last_<unit>` form is deferred to a follow-up to avoid a
metric-name break.

## How?

End-to-end tested through the real scrape endpoints (not internal state):

- **Prometheus** (`test/gtest/telemetry_prometheus_test.cpp`): injects TX
  `1000, 2000, 3500` and RX `500, 1500`, scrapes the live HTTP `/metrics`, and
  asserts `agent_tx_bytes_total == 6500` / `agent_tx_last_bytes == 3500` and
  `agent_rx_bytes_total == 2000` / `agent_rx_last_bytes == 1500`.
- **DOCA** (`test/doca-telemetry/telemetry_doca_nixl_test.cpp`): injects TX
  `10, 20, 35` and RX `5, 15`, flushes, scrapes the live DOCA Prometheus
  endpoint, and asserts the counters (`65`, `20`) and last-op gauges (`35`, `15`).
  Distinct TX/RX values also guard against cross-wiring.

Local validation: telemetry gtests + DOCA exporter tests pass; diff-scoped
clang-format clean.

## Summary by CodeRabbit

* **Documentation**
  * Updated telemetry docs to clarify per-operation byte metrics, cumulative `*_total` counters, and last-operation gauges `*_last_<unit>`, including naming conventions and exceptions.

* **New Features**
  * Prometheus and DOCA byte telemetry now expose both cumulative byte totals and last-operation byte gauges (`agent_tx_last_bytes` / `agent_rx_last_bytes`), alongside existing memory gauges.

* **Bug Fixes**
  * Ensured byte counters and last-operation gauges are emitted independently and reflect only the latest TX/RX values.

* **Tests**
  * Added end-to-end and Prometheus regressions validating totals vs last-operation gauges for TX/RX bytes.


e-eygin:

src/plugins/telemetry/doca/doca_exporter.h
```diff
@@ -53,8 +53,13 @@ class nixlTelemetryDocaExporter : public nixlTelemetryExporter {
     appendCounterSample(const nixlTelemetryEvent &event, const char *label_values[]);
 
     // Appends a gauge sample to the time-series (absolute last-operation value).
+    // metric_name is passed explicitly so an event can publish a gauge under a
+    // name that differs from the event name (e.g. AGENT_TX_BYTES drives both the
+    // "agent_tx_bytes" counter and the "agent_tx_bytes_last" gauge).
     [[nodiscard]] doca_error_t
-    appendGaugeSample(const nixlTelemetryEvent &event, const char *label_values[]);
+    appendGaugeSample(const nixlTelemetryEvent &event,
+                      const std::string &metric_name,
```
iyastreb: so on each sample we create a dynamic std::string from const char*
Is it desired behavior?
better pass const char * here

src/plugins/telemetry/doca/doca_exporter.h
```diff
@@ -53,8 +53,13 @@ class nixlTelemetryDocaExporter : public nixlTelemetryExporter {
     appendCounterSample(const nixlTelemetryEvent &event, const char *label_values[]);
 
     // Appends a gauge sample to the time-series (absolute last-operation value).
+    // metric_name is passed explicitly so an event can publish a gauge under a
```
iyastreb: imo this comment is excessive and trivial, better not to update existing one

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -306,14 +317,19 @@ nixlTelemetryDocaExporter::exportEvent(const nixlTelemetryEvent &event) {
         const std::lock_guard lock(g_metrics_mutex);
         const char *label_values[] = {agent_name_.c_str()};
 
+        // Counter and gauge are independent: the byte events publish BOTH a
+        // cumulative counter and a last-operation gauge from the same per-op
+        // value; every other metric is counter-only or gauge-only.
```
iyastreb: this comment documents current behavior, but if/when we add more "dual" counters, it won't be true anymore
Personally I would avoid commenting trivial code

src/plugins/telemetry/doca/doca_exporter.h
e-eygin: Good catch — fixed in `796f5dff`.

src/plugins/telemetry/doca/doca_exporter.h
e-eygin: Done — reverted to the original one-line comment in `796f5dff`.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Agreed, removed.

iyastreb (dismissed):

test/gtest/telemetry_prometheus_test.cpp
```diff
@@ -375,3 +381,56 @@ TEST_F(prometheusTelemetryTest, ExportEventIncrementReflectedInScrape) {
...
+        EXPECT_EQ(exporter->exportEvent(event), NIXL_SUCCESS);
+    }
+
+    const std::string body = waitForMetricsBody(port_);
+    ASSERT_FALSE(body.empty()) << "Got empty /metrics response on port " << port_;
+
+    PrometheusSample tx_total_sample;
+    ASSERT_TRUE(findAgentMetricSample(body, "agent_tx_bytes_total", agent_name, tx_total_sample))
+        << "agent_tx_bytes_total for this agent is not in scrape body";
+    EXPECT_EQ(tx_total_sample.value, 6500.0)
```
tomerg-nvidia, line 416: minor: would be better to compare against the actual sum vs a hardcoded number in here, same goes for the later checks

test/doca-telemetry/telemetry_doca_nixl_test.cpp
```diff
@@ -166,6 +166,52 @@ TEST_F(docaNixlExporterTest, DistinguishesSeriesByAgentLabel) {
...
+    }
+    ASSERT_EQ(exporter.flush(), NIXL_SUCCESS);
+
+    // RX is pushed last and each event appends its gauge sample before its
+    // counter sample, so the RX counter is the final sample; once agent_rx_bytes
+    // reads its total (20) every earlier sample (all TX, plus the RX gauge) has
+    // been served too -- read them all from that one settled scrape.
+    const auto metrics = scrapeUntilValue(port_, rxCounter, 20.0, std::chrono::seconds(12), labels);
+
+    EXPECT_EQ(metrics.latestValue(txCounter, labels), std::optional<double>(65.0))
```
tomerg-nvidia, line 205: Minor: Consider not to use EXPECT_EQ with doubles

iyastreb (approved):

PR #1826, opened 2026-06-24
telemetry: don't unlink the producer's file when attaching fails

EylonKrause:
## What?

`sharedRingBuffer<T>::openCyclicBuffer()` (`src/utils/common/cyclic_buffer.tpp`)
is the **reader/attach** path (`create == false`), but on two error branches it
`unlink()`s the shared-memory file it only *opened*:

- header `mmap` failure (line 211)
- version mismatch (line 223)

This PR removes those two `unlink()` calls.

## Why?

A reader is attaching to a file that a **producer** agent created and is actively
using. Deleting it on the reader's error path removes the producer's live
telemetry file from the filesystem. The most realistic trigger is a **version
mismatch**: a reader built from a different NIXL version (or attaching to an
exporter file from an older run) reads `version != TELEMETRY_VERSION` and unlinks
the producer's file.

The reader's *other* two error paths already do the right thing — "File too small
for buffer data" (~line 236) and the final whole-buffer `mmap` failure (~line 247)
both only `munmap` and `throw`, without unlinking. So this just makes all of
`openCyclicBuffer` consistent: **a reader never removes a file it did not create.**
The `unlink()`s in `createCyclicBuffer()` (the `create == true` path) are correct
and left unchanged — a creator may remove a file it just made. The `file_fd`
unique-ptr still closes the descriptor on the error path.

### Reproduction

A self-contained fs+mmap reproducer: a "producer" creates the file stamped with an
older version, and a "reader" attaches expecting a newer version.

```
BEFORE:  file present before reader: yes -> version mismatch -> present after: NO   (reader deleted it)
AFTER :  file present before reader: yes -> version mismatch -> present after: yes  (intact)
```

## How (verification)

- Confirmed the before/after with the reproducer above.
- Compiled the telemetry consumers (`buffer_exporter.cpp`, `telemetry.cpp` — which
  instantiate `sharedRingBuffer` / include `cyclic_buffer.tpp`) **in-tree** with
  `-Dsanitizer=address,undefined` (exit 0).

Happy to add a GoogleTest regression (construct the reader with a mismatched
version under `EXPECT_THROW`, then assert the file still exists) in `test/gtest/`
if you'd like one.

## Related Issues

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when opening cyclic buffers: on header mapping failures or on-disk version mismatches, the backing file is now preserved instead of being removed.
  * Errors are still logged and reported, but the existing file remains available for inspection and recovery.


iyastreb (approved):

EylonKrause:

ColinNV (approved):

e-eygin (approved):

PR #1827, opened 2026-06-24
posix: build the io pool from the clamped size, not the raw request

EylonKrause:
## What?

`nixlPosixIOQueueImpl` (`src/plugins/posix/io_queue.h`) sized its `ios_` pool and
`free_ios_` free-list from the **raw** `ios_pool_size` constructor argument. The base
class `nixlPosixIOQueue` clamps that value to `[MIN_IOS_POOL_SIZE (64),
MAX_IOS_POOL_SIZE]` and stores the clamped result in `ios_pool_size_`. This PR builds
the pool from `ios_pool_size_` so the pool and the rest of the class agree.

```diff
-          ios_(ios_pool_size) {
-        for (uint32_t i = 0; i < ios_pool_size; i++) {
+          ios_(ios_pool_size_) {
+        for (uint32_t i = 0; i < ios_pool_size_; i++) {
```

## Why?

The io_uring and Linux-AIO completion handlers detect "all I/O finished" with
`if (free_ios_.size() == ios_pool_size_) return NIXL_SUCCESS;`
(`io_uring_io_queue.cpp:137`, `linux_aio_io_queue.cpp:149,179`). `ios_pool_size_` is
the **clamped** value, but the pool was built with the **raw** count. For a requested
`ios_pool_size` of 1..63 (settable via the `ios_pool_size` backend param) the pool
holds fewer than 64 entries, so `free_ios_.size()` can never reach `ios_pool_size_` —
`doCheckCompleted` never returns `NIXL_SUCCESS` and the transfer never completes (hang).

### Reproduction

A self-contained model of the base clamp + pool construction + the "all free" check:

```
buggy:  requested=1  -> pool_entries=1,  ios_pool_size_=64  -> all-free reachable: NO (hang)
        requested=32 -> pool_entries=32, ios_pool_size_=64  -> all-free reachable: NO (hang)
fixed:  requested=1  -> pool_entries=64, ios_pool_size_=64  -> all-free reachable: YES
        requested=32 -> pool_entries=64, ios_pool_size_=64  -> all-free reachable: YES
```

## How (verification)

- Confirmed the before/after with the model above.
- Compiled the io-queue consumers (`posix_aio_io_queue.cpp`, `io_uring_io_queue.cpp`,
  which instantiate `nixlPosixIOQueueImpl`) **in-tree** with
  `-Dsanitizer=address,undefined` (exit 0; liburing builds via the meson subproject).

## Related Issues

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an internal sizing inconsistency in the POSIX I/O queue so its storage now matches the effective pool size used by the rest of the component.
  * Improved reliability when initializing queue resources, reducing the chance of mismatched allocation behavior.


EylonKrause:

iyastreb (approved):

iyastreb (changes requested):
CI triage agent feedback seems relevant

EylonKrause:
Thanks @iyastreb. The CI triage agent's non-gpu diagnosis actually points at exactly this PR's fix — it concludes with *"Apply the change proposed in PR #1827 — build the pool from the clamped size"* and notes the failing build (#2228) ran on a pre-fix commit (`621c7401`) where `io_queue.h` still had the raw `ios_pool_size`. After merging `main`, CI re-ran on the current head (`25bf52ba`) and all checks pass now, including **nixl-ci-non-gpu #2229** (the Python test that previously hung) plus the gpu/sanitizer jobs. The other triage comment (nixl-ci-gpu enroot-export hang) is an unrelated infra timeout it flags as transient. So the fix is present and CI is green — could you take another look? Thanks!

iyastreb (approved):

ofer (approved):

ovidiusm:
@lluki Would it make sense to add a unit test for the failure case, even if it has been resolved? To make sure we do not regress later

PR #1828, opened 2026-06-24
mooncake: fix request array leak on postXfer error path

EylonKrause:
## What?

`nixlMooncakeEngine::postXfer` (`src/plugins/mooncake/mooncake_backend.cpp`) allocated
the `transfer_request_t` array with `new[]` and only `delete[]`'d it once, after the
submit loop:

```diff
-    transfer_request_t *request = new transfer_request_t[request_count];
+    std::vector<transfer_request_t> request(request_count);
     for (size_t index = 0; index < request_count; ++index) {
         if (local[index].len != remote[index].len) return NIXL_ERR_INVALID_PARAM; // leaks request
         ...
     }
     ...
-    delete[] request;
```

It now uses `std::vector` and passes `request.data()` to `submitTransfer` /
`submitTransferWithNotify`.

## Why?

The per-descriptor length check inside the fill loop returns
`NIXL_ERR_INVALID_PARAM` **before** the lone `delete[]`, leaking the whole array on
every length-mismatched request. The `descCount()` equality is pre-checked, but the
per-descriptor `len` is not — so that early return is reachable from caller input.
`std::vector` frees on every exit path and removes the manual `delete[]`.

### Reproduction

The Mooncake plugin requires the Mooncake Transfer Engine library, which isn't present
in my environment, so I verified the leak with an extracted reproducer of the
`new[]` / loop-with-early-return / `delete[]` shape (2 descriptors, the second with a
length mismatch), built with `-fsanitize=address,leak`:

```
buggy (new[]):       SUMMARY: AddressSanitizer: 80 byte(s) leaked in 1 allocation(s)
fixed (std::vector): no leak
```

`std::vector` is already used elsewhere in this file, so the change needs no new include.

## Related Issues

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transfer request handling during data submissions with safer, size-managed buffering.
  * Reduced the risk of memory-management issues across notification and non-notification transfer operations.
  * Transfer behavior and supported capabilities remain unchanged.


EylonKrause:

iyastreb (approved):

alogfans (approved):
LGTM

EylonKrause:
The Blossom-CI authorization gate declined the auto-trigger on the rebased head (commit is not GPG-signed, so it wants a manual trigger). @iyastreb could you `/build` this one again when you get a chance? Nothing else changed since your last run besides the rebase.

PR #1832, opened 2026-06-24
build(manylinux): build wheels on public PyPA base + bundle INFINIA p…

nv-nmailhot:
…lugin

Rebuild contrib/Dockerfile.manylinux on the public quay.io/pypa/manylinux_2_28 base (+ nvcr.io CUDA stage, gcc-toolset-14) so the wheel builds on the AWS CI runners instead of the internal GitLab base image, and enable the INFINIA (DDN "red") plugin:

- Add an `infinia` stage from the ECR-mirrored DDN libs image and COPY the per-arch libs to /opt/ddn/red, where meson auto-detects libred_client and builds the INFINIA plugin into the wheel. libred_* are DDN-proprietary and stay auditwheel --excluded (already in build-wheel.sh); only the INFINIA plugin .so ships, loading libred_* at runtime from the customer's install.
- Install the libred link-time transitive deps (libuuid, jansson, libjose, xxhash, liburing 2.x, OpenLDAP 2.6).

build-container.sh: thread the manylinux/CUDA base + --cuda-version through, and disable buildx default attestations via BUILDX_NO_DEFAULT_ATTESTATIONS instead of the --provenance/--sbom flags (which the classic docker build frontend rejects as unknown).

## What?

## Why?

## How?
architecture, approach, etc._

## Summary by CodeRabbit

* **New Features**
  * Added `--cuda-version` for container-based wheel builds, with `MAJOR.MINOR` validation and automatic passing into the image build.
* **Bug Fixes**
  * Improved manylinux wheel builds by switching to public PyPA manylinux bases, bringing in CUDA tooling, and adding INFINIA and native client libraries for more reliable compilation and packaging.
  * Enhanced build-time library discovery to improve native linking during wheel creation.
* **Documentation**
  * Documented wheel-building steps (including CUDA examples) and how to build without INFINIA; added instructions to retrieve built `.whl` artifacts.
* **Chores**
  * Disabled Buildx’s default provenance/SBOM attestations for the container build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,40 @@ pip install nixl
 
 This installs both CUDA 12 and CUDA 13 backends. At runtime, the correct backend is selected automatically based on the CUDA version reported by PyTorch.
 
+### Building the wheels
+
+The release wheels are built inside a manylinux container by
+[`contrib/build-container.sh`](contrib/build-container.sh). Contributors and customers who
+need to build wheels manually can reproduce the release build with the same parameters used
+by the release pipeline (`.github/workflows/ci.yml`, the `build` job). Keep the commands
+below in sync with that workflow.
+
+```bash
+# CUDA 12, x86_64 — produces the cu12 manylinux_2_28 wheels
+./contrib/build-container.sh \
+  --base-image nvcr.io/nvidia/cuda \
+  --base-image-tag 12.9.1-devel-ubi8 \
+  --cuda-version 12.9 \
+  --wheel-base manylinux_2_28 \
+  --python-versions "3.10,3.11,3.12,3.13,3.14" \
+  --os ubuntu24 \
+  --arch x86_64 \
+  --dockerfile contrib/Dockerfile.manylinux \
+  --tag nixl-wheel-build:cu12-x86_64
+
+# CUDA 13:  --base-image-tag 13.0.1-devel-ubi8  --cuda-version 13.0
+# aarch64:  --arch aarch64   (build on an arm64 host)
+```
+
+The wheels are written to `/workspace/nixl/dist` inside the image; extract them with:
+
+```bash
+cid=$(docker create nixl-wheel-build:cu12-x86_64)
+docker cp "$cid:/workspace/nixl/dist" ./dist
+docker rm "$cid"
+ls dist/*.whl
+```
+
 ## Prerequisites for source build (Linux)
 
 NIXL requires a C++20 compatible compiler (GCC >= 11 or Clang >= 14).
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ The release wheels are built inside a manylinux container by
 the release pipeline (`.github/workflows/ci.yml`, the `build` job). Keep the commands below in
 sync with that workflow.
 
-> **Note:** the `contrib/Dockerfile.manylinux` build below pulls an **NVIDIA-internal** INFINIA
-> (DDN) libs image, so it currently only builds inside NVIDIA. To build wheels **without** the
-> INFINIA plugin — works anywhere, no internal image required — see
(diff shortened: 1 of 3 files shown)


contrib/Dockerfile.manylinux
```diff
@@ -14,12 +14,98 @@
 # limitations under the License.
```
ovidiusm, line 14: Could we add a short section to the NIXL readme with the canonical commands that can be used to build the wheels? E.g. ./contrib/build-container.... 

It's quite common for either contributors and customers to sometimes have to build wheels manually so it's useful to know the procedure

This should be kept in sync with the commands/parameters used in the release pipeline step

ovidiusm (approved):

README.md
```diff
@@ -45,6 +45,40 @@ pip install nixl
...
+
+The release wheels are built inside a manylinux container by
+[`contrib/build-container.sh`](contrib/build-container.sh). Contributors and customers who
+need to build wheels manually can reproduce the release build with the same parameters used
+by the release pipeline (`.github/workflows/ci.yml`, the `build` job). Keep the commands
+below in sync with that workflow.
+
+```bash
+# CUDA 12, x86_64 — produces the cu12 manylinux_2_28 wheels
+./contrib/build-container.sh \
```
ovidiusm, line 63: I'm not sure these commands will work for anyone if the infinia image is private :(

contrib/Dockerfile.manylinux
```diff
@@ -97,6 +188,43 @@ ENV OPENSSL_ROOT_DIR="/usr/local/openssl3"
 ENV OPENSSL_LIBRARIES="/usr/local/openssl3/lib64:/usr/local/openssl3/lib"
 ENV OPENSSL_INCLUDE_DIR="/usr/local/openssl3/include"
 
+# --- INFINIA (libred_client) transitive deps not packaged at the right soname on EL8 ---
+# liburing 2.x: EL8 only ships liburing.so.1 (1.0.7); libred_client needs liburing.so.2.
+RUN cd /tmp && \
+    LIBURING_VERSION=2.6 && \
```
aranadive: Lets use version 2.14. We also build liburing as a meson wrap as part of NIXL build. Can we inherit that instead?

contrib/Dockerfile.manylinux
```diff
@@ -14,12 +14,98 @@
...
+# puts libred_client.so / libred_async.so + headers where meson looks, so it auto-detects
+# and builds the INFINIA plugin into the wheel (default -Dinfinia_path=/opt/ddn/red).
+# NOTE: libred_client/libred_async are DDN proprietary and are intentionally NOT
+# redistributed — contrib/build-wheel.sh already auditwheel --excludes them, so the wheel
+# ships only the INFINIA plugin (libplugin_INFINIA.so), which loads libred_* at runtime
+# from a customer's DDN install via its RUNPATH (/opt/ddn/red/lib). The plugin's other
+# transitive deps (liburing.so.2, libldap.so.2, etc.) are why this Dockerfile installs/
+# builds them above: they must resolve at link time even though libred itself is excluded.
+ARG INFINIA_DEST=/opt/ddn/red
+COPY --from=infinia /infinia/${ARCH}/ ${INFINIA_DEST}/
```
aranadive, line 106: We should skip this copy if the infinia image is not accessible.

ovidiusm (changes requested):
There should be an option to skip the inifinia private image in case it is not available. Or to specify a different URL for it

ovidiusm (approved):

contrib/Dockerfile.manylinux
nv-nmailhot: added skip option

PR #1833, opened 2026-06-24
nixlbench: change to range lookup in Neuron allocation tracking

fengjica:
Today in nixlbench, when data is copied from Neuron device to host, it checks the address of an IOV within the existing Neuron memory allocation. The latter is done once with a single large buffer, which has the base address different from the IOV's base address. Not found neuron allocation leads to nixlbench failure.

## What?
Change to range lookup when searching in Neuron allocation tracking.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed device copy and memset to operate on the correct region within an existing allocation, using the full VA range (not just the base).
  * Improved device allocation tracking to reliably resolve tensors for any VA within a mapped range.
  * Added stronger bounds checking for copy and fill operations to prevent out-of-range reads and writes.
  * Enhanced diagnostics with clearer messages when allocations are missing, access is out of bounds, or device read/write calls fail.


benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -121,20 +121,35 @@ struct NrtTensorDeleter {
...
 
-    auto it = allocation_tracker.find(va);
-    if (it == allocation_tracker.end()) {
-        return nullptr;
+    // Find the first entry with base > addr, then step back one.
+    auto it = allocation_tracker.upper_bound(addr);
+    if (it == allocation_tracker.begin()) {
+        return {nullptr, 0};
+    }
+    --it;
```
oa-aws, line 152: Note that if the map is empty, decrementing an end iterator to a std::map is not safe. Maybe we can add a check in the beginning of the function (after taking lock) that the map is not empty, (and if empty return null)?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -161,14 +176,15 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
     *addr = nrt_tensor_get_va(tensor);
     if (*addr == nullptr) {
+        nrt_tensor_free(&tensor);
         return -1;
     }
 
     std::lock_guard lock{allocation_tracker_mutex};
-    allocation_tracker.emplace(*addr, std::move(ptr));
+    uintptr_t base = reinterpret_cast<uintptr_t>(*addr);
+    allocation_tracker.emplace(base, NrtAllocation{tensor, buffer_size});
```
oa-aws: Missing return value check of call to emplace (second item in returned pair). Also how do we handle error? Should the tensor be freed? Does the caller address errors?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -178,38 +194,50 @@ neuronFree(void *addr) {
     if (!addr) return 0;
 
     std::lock_guard lock{allocation_tracker_mutex};
-    return allocation_tracker.erase(addr) - 1;
+    return allocation_tracker.erase(reinterpret_cast<uintptr_t>(addr)) - 1;
```
oa-aws: This looks a bit risky. Although return type of erase() is size_type (unsigned, 32 bits), and if the mapping was erased 1 will be returned from ease (so the function returns 0 for success), otherwise it returns 0 (no mapping was find by erase key), and we will return UINT32_MAX, which converted to int is -1. So this works, but is not so intuitive or readable. Can we check return value fro ease, and return explicit result to caller instead?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -178,38 +194,50 @@ neuronFree(void *addr) {
...
+        std::cerr << "neuronMemcpy: no allocation found for VA " << device_addr
+                  << " (kind=" << (kind == neuronMemcpyHostToDevice ? "H2D" : "D2H")
+                  << ", count=" << count << ")" << std::endl;
         return -1;
     }
 
+    int status;
     if (kind == neuronMemcpyHostToDevice) {
-        return nrt_tensor_write(tensor, src, 0, count);
+        status = nrt_tensor_write(tensor, src, offset, count);
```
oa-aws, line 233: Missing size check. What if we cross tensor memory boundary? This can be resolved if we return also size from findTensorForVA(), and then check that offset + count does not exceed tensor memory region size. The same is true also for read case. Conversely, do we have guarantee that nrt_tensor_read/write does a range check? I still think it is better for us to make the check if we can, instead of relying on external library to do so.

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -178,38 +194,50 @@ neuronFree(void *addr) {
...
 
     constexpr size_t kMaxChunkSize = 1UL << 21; // 2MB
     std::vector<unsigned char> buf(kMaxChunkSize, static_cast<unsigned char>(val));
     int status = 0;
-    size_t offset = 0;
-    while (offset < count && status == 0) {
-        const size_t write_len = std::min(kMaxChunkSize, count - offset);
-        status = nrt_tensor_write(tensor, buf.data(), offset, write_len);
-        offset += write_len;
+    size_t pos = 0;
```
oa-aws, line 264: Should we check that offset + count does not exceed tensor memory region (just as in comment above)?

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: If the maps is empty, end() is the same to begin(). Then the next if, comparing to allocation_tracker.begin(), will be true, and so it returns {nullptr, 0}. Right?

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed in the next revision.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed in the next revision.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed in the next revision.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed.

benchmark/nixlbench/src/utils/neuron.cpp
oa-aws: That is correct, but I think that is not so intuitive, and could be hard to maintain. Do you think this can improve code readability or is it ok like this?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -174,39 +175,54 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
     const void *device_addr = (kind == neuronMemcpyHostToDevice) ? dest : src;
-    auto [tensor, offset] = findTensorForVA(device_addr);
+    auto [tensor, offset, alloc_size] = findTensorForVA(device_addr);
     if (tensor == nullptr) {
         std::cerr << "neuronMemcpy: no allocation found for VA " << device_addr
                   << " (kind=" << (kind == neuronMemcpyHostToDevice ? "H2D" : "D2H")
                   << ", count=" << count << ")" << std::endl;
         return -1;
     }
+    if (offset + count > alloc_size) {
```
oa-aws: Maybe overflow check here is required (and a warning)?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -225,10 +241,16 @@ neuronMemcpy(void *dest, const void *src, size_t count, neuronMemcpyKind kind) {
 
 int
 neuronMemset(void *addr, int val, size_t count) {
-    auto [tensor, offset] = findTensorForVA(addr);
+    std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
+    auto [tensor, offset, alloc_size] = findTensorForVA(addr);
     if (tensor == nullptr) {
         return -1;
     }
+    if (offset + count > alloc_size) {
```
oa-aws: Same as above (overflow check).

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -174,39 +175,54 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
         return -1;
     }
 
-    std::lock_guard lock{allocation_tracker_mutex};
+    std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
     uintptr_t base = reinterpret_cast<uintptr_t>(*addr);
-    allocation_tracker.emplace(base, NrtAllocation{tensor, buffer_size});
+    auto [it, inserted] = allocation_tracker.emplace(base, NrtAllocation{tensor, buffer_size});
+    if (!inserted) {
+        *addr = nullptr;
```
oa-aws, line 196: Is a calll to nrt_tensor_free() missing here?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -134,22 +135,22 @@ struct NrtAllocation {
 std::map<uintptr_t, NrtAllocation> allocation_tracker;
 std::mutex allocation_tracker_mutex;
 
-// Find the allocation containing the given VA. Returns {tensor, offset} or {nullptr, 0}.
-std::pair<nrt_tensor *, size_t>
+// Find the allocation containing the given VA. Caller must hold allocation_tracker_mutex.
+// Returns {tensor, offset, alloc_size} or {nullptr, 0, 0}.
+std::tuple<nrt_tensor *, size_t, size_t>
 findTensorForVA(const void *va) {
-    std::lock_guard lock{allocation_tracker_mutex};
```
oa-aws, line 131: Is there a reason for removing the lock here? Is it still safe without a lock? I see that you added a lock in a few places before calling findTensorForVA(). Is that for making sure the memory does not get concurrently deleted while accessing it? If that is the case, I have 2 questions:

1. Are all calls to findTensorForVA() made after locking?
2. Could it cause severe performance degradation, since now we hold the lock on the entire map during I/O over DMA/PCI? If that is true, I suggest using at least ref-count per memory region. That is, after finding the memory region in the map, but before getting out of lock scope, we increment ref-count on region (maybe can use shared ptr that calls nrt_tensor_free() when done). This way we can bring back the locked scope only for the period of search in the map, and we can still safely access the memory region. This is also more maintainable making sure we lock before each call.

fengjica: Yes, the old way allowed memory being set/modified (in neuronMemcpy/set) because of no lock there, and neuronFree() could delete it. Now they are all serialized with locks added.

With the current interface, I don't think we can completely remove the problem of use-after-free:
 - neuronMalloc(void **addr,...) gives the caller direct access to the pointer. Unless this is changed to shared_ptr as well, which will need more code change outside, I guess implementing ref count only address any race of within this file.

On performance, I don't expect it's the case that multiple threads are doing operations (memset/cpy/free) on a single region in a performance critical code path. So here I'm taking a minimum approach to protect against a bug of racing neuronFree with neruoMemcpy/set here.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: I see your point.

I tried with a few ways, but that will always requires two different function calls on the allocation, whereas comparing to begin() is one function call.

Updated with adding comments there, in the next revision.

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -121,20 +122,35 @@ struct NrtTensorDeleter {
...
-        return nullptr;
+    // Find the first entry with base > addr, then step back one.
+    auto it = allocation_tracker.upper_bound(addr);
+    if (it == allocation_tracker.begin()) {
+        return {nullptr, 0, 0};
     }
-    return it->second.get();
+    --it;
+    if (addr < it->first + it->second.size) {
+        return {it->second.tensor.get(), addr - it->first, it->second.size};
```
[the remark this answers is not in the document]
fengjica: Fixed.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: NrtAllocation's deleter (NrtTensorDeleter) will take care of that, when the temporary object is deleted if the insertion fails.

Added a comment on NrtAllocation.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed.

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Fixed.

fengjica:
@oa-aws The new code change is in the 3rd commit "[fix to address reviews: overflow, readability](https://github.com/ai-dynamo/nixl/pull/1833/commits/e2a2ccb574a31589511aa7270bd1525349b20ded)"

benchmark/nixlbench/src/utils/neuron.cpp
oa-aws: Ok, let's consider addressing this in a subsequent PR if this turns out as causing performance degradation in some use case.

benchmark/nixlbench/src/utils/neuron.cpp
oa-aws: Ok, a comment in place is good enough. Resolving.

oa-aws (dismissed):

fengjica:
Thanks @oa-aws for the review!

As the next step, I'll merge the current commits, and then seek approval from the code owner of nixlbench dir.

fengjica:
Hi @aranadive @ovidiusm @brminich Could you help take a look at this PR and trigger CI? Thanks!

After oa-aws's approval above, I squashed the commits into one.

aranadive:
@fengjica can you address the CR comments?

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -159,57 +179,91 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
 }
 
 int
 neuronMemset(void *addr, int val, size_t count) {
-    nrt_tensor *tensor = getTensorFromVA(addr);
+    std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
+    auto [tensor, offset, alloc_size] = findTensorForVA(addr);
     if (tensor == nullptr) {
         return -1;
     }
```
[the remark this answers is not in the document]
fengjica: Fixed.

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -159,57 +179,91 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
+    std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
+    uintptr_t base = reinterpret_cast<uintptr_t>(*addr);
+    auto [it, inserted] = allocation_tracker.emplace(base, NrtAllocation{tensor, buffer_size});
+    if (!inserted) {
+        *addr = nullptr;
+        return -1;
+    }
 
     return 0;
 }
```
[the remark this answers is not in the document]
fengjica: Fixed.

fengjica:
@aranadive Could you look again? I add a second commit and resolved the two CR comments.
[address CR comment: add const to lock_guard and add logging](https://github.com/ai-dynamo/nixl/pull/1833/commits/6681b61db1402d1e84e3a94b06f7c6738bc46a0f)

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -159,57 +179,93 @@ neuronMalloc(void **addr, size_t buffer_size, int devid) {
...
-    return allocation_tracker.erase(addr) - 1;
+    const std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
+    const auto erased = allocation_tracker.erase(reinterpret_cast<uintptr_t>(addr));
+    return erased == 1 ? 0 : -1;
 }
 
 int
 neuronMemcpy(void *dest, const void *src, size_t count, neuronMemcpyKind kind) {
-    nrt_tensor *tensor = getTensorFromVA(kind == neuronMemcpyHostToDevice ? dest : src);
+    const std::lock_guard<std::mutex> lock{allocation_tracker_mutex};
```
aranadive, line 216: this now hold allocation_tracker_mutex across the DMA, not just the lookup — memset keeps it for a full loop of 2 MB writes. Wouldnt this serializes all Neuron transfers, which could skew throughput on the concurrent benchmarks or is it acceptable here?

fengjica: I see your point.

However I don't think this is a concern in nixlbench, because this memcpy is only used in xferBenchUtils::validateTransfer()

like in https://github.com/ai-dynamo/nixl/blob/main/benchmark/nixlbench/src/main.cpp#L147

I don't think that is a hot path, i.e. outside that code path that's timed, right? It's also out of the multi-threading path.

aranadive (approved):

PR #1835, opened 2026-06-25
PLUGINS/UCX: Add scatter-gather list (SGL) put path

michal-shalev:
## What?
Add a UCX scatter-gather (SGL) put path.

## Why?
Posting one UCX request per descriptor is wasteful for multi-element writes. An SGL put hands UCX the whole descriptor list in one call, reducing per-element overhead.

## How?
Off by default, gated at two layers:
- **Build-time**: `meson.build` probes for `ucp_dt_local_sgl_t` and defines `HAVE_UCX_SGL_API` only when UCX exposes the SGL API. All SGL code is `#ifdef`-guarded.
- **Run-time**: `NIXL_UCX_SGL_ENABLE` (default `false`) toggles it; if set on a build without SGL support, it warns and falls back to the per-descriptor path.

For `NIXL_WRITE`, a new `sglXfer` gathers the descriptor range into parallel arrays (addrs, lengths, memhs, rkeys) kept on the request handle. `sendXferRange` short-circuits to `sendXferSgl`, issuing one `postSgl` (`ucp_put_nbx` + `ucp_dt_make_sgl()`) and a single `flushEp` per batch.

## Summary by CodeRabbit

* **New Features**
  * Added optional UCX scatter-gather list (SGL) offload for write transfers.
  * When supported and enabled, write operations are submitted using SGL for more efficient handling across multiple segments.
* **Compatibility**
  * UCX SGL support is auto-detected at build time; builds without it won’t include the feature.
  * Runtime enablement is controlled by `NIXL_UCX_SGL_ENABLE`; if SGL isn’t compiled in, it is forced off with a warning.
  * When SGL can’t be used, the system falls back to the existing transfer behavior.


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -826,6 +850,17 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params)
     const auto engine_config =
         nixl::getBackendParamDefaulted(custom_params, "engine_config", std::string());
 
+    sglEnabled_ = nixl::config::getValueOptional<bool>("NIXL_UCX_SGL_ENABLE").value_or(false);
```
ColinNV: `...::getValueDefaulted("NIXL_UCX_SGL_ENABLE", false)`

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -85,6 +90,86 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
 
     std::optional<Notif> notif;
 
+#ifdef HAVE_UCX_SGL_API
```
ColinNV, line 106: Separate header for this class, please.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -85,6 +90,86 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
...
+    class sglXfer {
+    public:
+        sglXfer(const nixl_meta_dlist_t &local,
+                const nixl_meta_dlist_t &remote,
+                size_t worker_id,
+                size_t start_idx,
+                size_t end_idx) {
+            const size_t count = end_idx - start_idx;
+            resize(count);
+            conn_ = static_cast<nixlUcxPublicMetadata *>(remote[start_idx].metadataP)->conn;
```
ColinNV: Member initializer list and `const`.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -85,6 +90,86 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
...
+#ifdef HAVE_UCX_SGL_API
+    class sglXfer {
+    public:
+        sglXfer(const nixl_meta_dlist_t &local,
+                const nixl_meta_dlist_t &remote,
+                size_t worker_id,
+                size_t start_idx,
+                size_t end_idx) {
+            const size_t count = end_idx - start_idx;
+            resize(count);
```
ColinNV: Should this validate the indices as in `assert( end_idx > start_idx )`?

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -85,6 +90,86 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
...
+                .field_mask = UCP_DT_REMOTE_SGL_FIELD_REMOTE_ADDRS |
+                    UCP_DT_REMOTE_SGL_FIELD_LENGTHS | UCP_DT_REMOTE_SGL_FIELD_RKEYS,
+                .remote_addrs = remoteAddrs_.data(),
+                .lengths = lengths_.data(),
+                .rkeys = rkeys_.data(),
+            };
+        }
+
+    private:
+        void
```
ColinNV: Could be worth optimizing this because (a) each vector allocates memory and (b) the resize initializes the elements which is discarded work. Switching to `reserve()` and then using `emplace_back()` in the construct loop probably doesn't save much, but doing a single memory allocation for all the data and managing it manually might.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -85,6 +90,86 @@ class nixlUcxBackendReqH : public nixlBackendReqH {
...
+                localAddrs_[out] = reinterpret_cast<void *>(local[i].addr);
+                remoteAddrs_[out] = static_cast<uint64_t>(remote[i].addr);
+                lengths_[out] = local[i].len;
+                memhs_[out] = lmd->getMem().getMemh();
+                rkeys_[out] = rmd->getRkey(worker_id).get();
+            }
+        }
+
+        [[nodiscard]] size_t
+        count() const noexcept {
```
ColinNV: `size()`?

src/plugins/ucx/ucx_backend.h
```diff
@@ -292,6 +302,7 @@ class nixlUcxEngine : public nixlBackendEngine {
     std::vector<std::unique_ptr<nixlUcxWorker>> uws;
     std::string workerAddr;
     mutable std::atomic<size_t> sharedWorkerIndex_;
+    bool sglEnabled_ = false;
```
ColinNV: `const` and member initializer list with helper function.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1227,9 +1371,16 @@ nixlUcxEngine::sendXferRange(const nixl_xfer_op_t &operation,
         return NIXL_ERR_INVALID_PARAM;
     }
 
-    /* Assuming we have a single EP, we need 3 requests: one pending request,
-     * one flush request, and one notification request */
-    int_handle->reserve(3);
+#ifdef HAVE_UCX_SGL_API
+    if (sglEnabled_ && operation == NIXL_WRITE) {
+        if (!int_handle->sgl) {
+            int_handle->sgl.emplace(local, remote, worker_id, start_idx, end_idx);
```
ColinNV, line 1265: Why do we need both this auto-fill _and_ `prepXferSgl()`?

ColinNV (commented):
Test coming separately?

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Added const, need to validate start_idx before init

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Renamed

src/plugins/ucx/ucx_backend.h
michal-shalev: Done

src/plugins/ucx/ucx_backend.cpp
michal-shalev: They serve two different paths: `prepXferSgl()` builds the SGL at prep time for the normal engine path (keeping it out of the post hot path). The `sendXferRange()` auto-fill is for the thread-pool composite path, where per-chunk handles only exist at send time and each chunk builds its own sub-range SGL. The `!int_handle->sgl` guard skips the rebuild when prep already populated it.

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Done

src/plugins/ucx/ucx_sgl.cpp
```diff
@@ -0,0 +1,67 @@
...
+    constexpr size_t field_bytes = sizeof(uint64_t);
+    const size_t array_bytes = size_ * field_bytes;
+
+    storage_ = std::make_unique_for_overwrite<std::byte[]>(num_fields * array_bytes);
+    std::byte *const base = storage_.get();
+    localAddrs_ = reinterpret_cast<void **>(base + 0 * array_bytes);
+    remoteAddrs_ = reinterpret_cast<uint64_t *>(base + 1 * array_bytes);
+    lengths_ = reinterpret_cast<size_t *>(base + 2 * array_bytes);
+    memhs_ = reinterpret_cast<ucp_mem_h *>(base + 3 * array_bytes);
+    rkeys_ = reinterpret_cast<ucp_rkey_h *>(base + 4 * array_bytes);
```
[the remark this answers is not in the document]
iyastreb: alignment concern seems valid

src/plugins/ucx/ucx_sgl.h
```diff
@@ -0,0 +1,91 @@
...
+            .lengths = lengths_,
+            .rkeys = rkeys_,
+        };
+    }
+
+private:
+    static constexpr size_t num_fields = 5;
+
+    std::unique_ptr<std::byte[]> storage_;
+    void **localAddrs_ = nullptr;
```
iyastreb: all fields are initialized in ctor, so defaults are not needed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1212,6 +1243,48 @@ nixlUcxEngine::sendXferRangeBatch(nixlUcxEp &ep,
...
+    const ucp_dt_local_sgl_t local_sgl = sgl.localView();
+    const ucp_dt_remote_sgl_t remote_sgl = sgl.remoteView();
+    const ucx_connection_ptr_t &conn = sgl.conn();
+
+    auto &ep = conn->getEp(int_handle->getWorkerId());
+
+    int_handle->reserve(single_ep_request_count);
+
+    nixlUcxReq req;
+    const nixl_status_t post_ret = ep->postSgl(local_sgl, remote_sgl, sgl.size(), req);
```
iyastreb: why not to pass just sgl there?

src/plugins/ucx/ucx_sgl.cpp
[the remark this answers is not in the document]
michal-shalev: We're using C++20 which includes P0593R6.
@coderabbitai

src/plugins/ucx/ucx_backend.cpp
michal-shalev: What do you mean by "pass sgl there"?

ovidiusm (dismissed):

tomerg-nvidia (dismissed):

src/plugins/ucx/ucx_backend.cpp
iyastreb: why not `postSgl(const nixl::ucx::sglXfer &sgl)`?

I think it is error prone to split sgl into 4 params and post like that. Because there is no case when we can mix local_sgl from one sglXfer with remote_sgl from another. Then why API allows that?

Can just pass sglXfer object, which is properly constructed

iyastreb (commented):
Please consider `postSgl(const sglXfer& sgl)`

ColinNV (dismissed):

ColinNV (dismissed):

ovidiusm (dismissed):

tomerg-nvidia (dismissed):

src/plugins/ucx/ucx_backend.cpp
michal-shalev: While the original `ep->postSgl(local, remote, count)` matched the `write/read` pattern (backend unpacks, EP takes UCX-level args), your suggestion of `postSgl(const sglXfer&)` on the EP inverts the dependency: today it’s `sglXfer -> backend -> EP`, and that would add `EP -> sglXfer`.

However, I agree it would be better to pair `local/remote/count`, so I pushed a cleaner alternative: `sglXfer::post(ep, req)` which keeps UCX views private on the owner, and preserves `sglXfer -> backend -> EP`.

iyastreb: IMO it was not done the right way for EP, and it should not be the reference.

NIXL ucx_utils is supposed to wrap UCX primitives like worker, ep etc.
Ideally outside of ucx_utils we should not directly reference UCX primitives, that's the whole point of wrapping.  Let's do it properly at least in the new code

iyastreb (dismissed):

ColinNV (approved):

iyastreb:
I have just submitted PR for perf degradation caused by previous SGL changes
https://github.com/ai-dynamo/nixl/pull/2017
Please check that we don't have the similar issue in this PR

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1063,6 +1088,12 @@ nixl_status_t nixlUcxEngine::prepXfer (const nixl_xfer_op_t &operation,
     /* TODO: try to get from a pool first */
     handle = new nixlUcxBackendReqH(getSharedWorker(worker_id).get());
 
+#ifdef HAVE_UCX_SGL_API
+    if (sglEnabled_ && operation == NIXL_WRITE) {
```
iyastreb, line 1092: Wait, just because we _can_ post SGL doesn't mean we _should_.
Verification team checks for degradations on all batch dimensions, starting from 1. For example they check nixlbench with 4K message and batch size 1.
Could you please measure overhead of SGL at this dimension?
Most likely for small batches (below 8-16) it's not practical at all
We need to know the threshold and use SGL only when it's beneficial, never for size 1

iyastreb (changes requested):

src/plugins/ucx/ucx_backend.cpp
michal-shalev: Thanks for bringing this up, it's a valid concern.

From the current `ucx_perftest` results, the only case where we don't see a benefit is `count == 1`, and even there the overhead is small.

At the moment I'd prefer not to introduce a threshold in NIXL because:
1. The remaining UCX changes may affect these measurements.
2. The SGL path is disabled by default.
3. The validation team will run a full performance evaluation.
4. Alternatively, if we eventually decide that a threshold is needed, it could be implemented in UCX rather than NIXL, avoiding the need to keep both APIs in NIXL.

Once the remaining UCX work is merged, I'll re-run the measurements and, if the data still shows a clear threshold, we can introduce it then.

iyastreb: Thanks a lot for checking!

I would propose to do the following things, but up to you

1. The most relevant result would be `nixlbench` test, because NIXL side adds a bit of overhead too

2. May you please attach to the PR results of the perf test (with enabled and disabled SGL in release mode):
```
UCX_TLS=^cuda_ipc nixlbench --initiator_seg_type=VRAM --target_seg_type=VRAM --start_block_size=4096 --max_block_size=4096 --start_batch_size=1--max_batch_size=128 --num_iter=100000
```
It costs nothing to run it, but we know 100% the current state

3. If we know in advance that for 1 batch size plain call is faster, I would propose to at least:
```
if (sglEnabled_ && operation == NIXL_WRITE && size > 1)
```
This would completely eliminate the risk of regression

iyastreb (approved):

src/plugins/ucx/ucx_backend.cpp
michal-shalev: 1. I agree that `nixlbench` is a relevant benchmark. Let's have the verification team measure it once the remaining PRs in this series are merged so we get full coverage.
2. Regarding attaching benchmark results to the PR, the guidance I received is to avoid publishing performance results in public PRs. The verification process for this feature will be done together with the verification team, and we'll address any issues or follow-up changes based on their feedback before considering making this feature the default.
3. I considered the `size > 1` condition before replying to your previous comment. For the reasons I explained there, I decided not to add it at this stage.

If these address your concerns, could you please resolve the review comments so they no longer block this PR from merging?

michal-shalev:
> I have just submitted PR for perf degradation caused by previous SGL changes #2017 Please check that we don't have the similar issue in this PR

Good catch on the NIXL issue, and thanks for fixing it in #2017.

PR #2017 addresses a pre-existing issue in the default NIXL path that became more noticeable after openucx/ucx#11258 increased the size of `ucp_request_param_t`, so describing it as a performance degradation caused by previous SGL changes is somewhat misleading.

PR #1835 introduces a separate, optional SGL path behind an environment variable and does not modify the default execution path. If you believe there is a similar issue in this PR, please point to the specific code path or demonstrate the regression here.

tomerg-nvidia (approved):

src/plugins/ucx/ucx_backend.cpp
iyastreb: 2. It's ok if you share results in the common group

michal-shalev:
/ok to test [514c03e](https://github.com/ai-dynamo/nixl/pull/1835/commits/514c03ed0b37883a5e709f7f6644e4a4524e4015)

brminich (approved):

PR #1840, opened 2026-06-26
utils: add shared S3-over-RDMA transport and protocol helpers

harshavardhana:
## What

Adds the shared GPU-direct S3-over-RDMA infrastructure as reusable utilities under `src/utils/object/`, so object engines compose it rather than each embedding its own copy. This keeps the transport/protocol layer in one place and easy to maintain as more engines adopt it. Additive only — no existing code paths change.

## Contents

- **`rdma_protocol.h`** — dependency-free wire helpers (token formatting, reply parsing, constants), unit-tested.
- **`rdma.{h,cpp}`** — `SharedCuObjClient` (process-wide cuObject RDMA plumbing) and `S3RdmaControlPlane` (SigV4 `UNSIGNED-PAYLOAD` control plane carrying `x-amz-rdma-token`, reading `x-amz-rdma-reply` / `x-amz-rdma-bytes-transferred`) plus retry wrappers. Guarded by `HAVE_CUOBJ_CLIENT`; compiles to nothing without cuObjClient.
- Built into the `obj_utils` library; `RdmaProtocol` unit test under `test/gtest/unit/obj`.

The header names and token layout follow NVIDIA's published [cuObject docs](https://docs.nvidia.com/gpudirect-storage/cuobject/index.html) and the protocol proposed to AWS ([aws-c-s3 `RDMA_PROTOCOL_SPEC.md`](https://github.com/KiranModukuri/aws-c-s3/blob/nvidia_rdma/RDMA_PROTOCOL_SPEC.md)).

## Notes

- `formatRdmaToken` uses `PRIx64`; `rdmaGet` rejects `size==0`; credentials are resolved once at construction (documented limitation: rotating IAM session tokens need a fresh backend).
- `SharedCuObjClient` has no S3 dependency, so it can move under `src/utils/cuobj/` if preferred — happy to do that.

## Test

Builds into `obj_utils`; `RdmaProtocol.*` unit tests pass (7/7).

## Types of changes
- [x] New feature (non-breaking — additive utilities)

## Summary by CodeRabbit

* **New Features**
  * Added accelerated S3 transfers over RDMA for supported hardware and configurations.
  * Added RDMA support for uploads, downloads, ranged reads, checksums, and multipart operations.
  * Added reliable retry handling and clear fallback behavior when RDMA is unavailable.
  * Added strict validation for RDMA transfer responses and protocol tokens.

* **Bug Fixes**
  * Improved S3 configuration handling by treating backend parameters as read-only.
  * Aligned build and linkage requirements for RDMA-enabled S3 functionality.

* **Tests**
  * Added comprehensive coverage for RDMA token formatting, response parsing, and protocol constants.


harshavardhana:
PTAL @aranadive let me know if there is anything specific that needs to change in the PR

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+        impl_->scheme = (config.scheme == Aws::Http::Scheme::HTTP) ? "http" : "https";
+        impl_->virtual_addressing = nixl_s3_utils::getUseVirtualAddressing(custom_params);
+
+        // Endpoint authority: explicit override (AIStor / S3-compatible) or the
+        // default AWS S3 regional host.
+        if (!config.endpointOverride.empty()) {
+            Aws::Http::URI ep(config.endpointOverride);
+            // Honor the override's scheme for both http and https (a bare http
+            // config.scheme must not stick when the override is https).
+            impl_->scheme = (ep.GetScheme() == Aws::Http::Scheme::HTTP) ? "http" : "https";
```
jgoldsch12: Maybe consider a warning if this causes an overwrite.

src/utils/object/rdma/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+        auto req =
+            Aws::Http::CreateHttpRequest(uri,
+                                         Aws::Http::HttpMethod::HTTP_PUT,
+                                         Aws::Utils::Stream::DefaultResponseStreamFactoryMethod);
+        req->SetHeaderValue("x-amz-content-sha256", unsigned_payload);
+        req->SetHeaderValue(amz_rdma_token, formatRdmaToken(token, buf_addr, size).c_str());
+        req->SetHeaderValue("content-type", "application/octet-stream");
+        req->SetContentLength("0");
+        if (!ctx.checksumCrc64nvme.empty()) {
```
jgoldsch12: Does this ignore the req_checksum backend setting?  Or is that maybe handled in other util code.  Apologies for not being sure.

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+S3RdmaControlPlane::~S3RdmaControlPlane() = default;
+
+ssize_t
+S3RdmaControlPlane::rdmaPut(S3RdmaClientCtx &ctx,
+                            const char *token,
+                            uint64_t buf_addr,
+                            uint64_t size) {
```
jgoldsch12: Is it legitimate to have a size of 0?

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+    try {
+        Aws::Http::URI uri = impl_->buildUri(ctx.bucket, ctx.object);
+        auto req =
+            Aws::Http::CreateHttpRequest(uri,
+                                         Aws::Http::HttpMethod::HTTP_GET,
+                                         Aws::Utils::Stream::DefaultResponseStreamFactoryMethod);
+        req->SetHeaderValue("x-amz-content-sha256", unsigned_payload);
+        req->SetHeaderValue(amz_rdma_token, formatRdmaToken(token, buf_addr, size).c_str());
+        // Byte-range fetch when reading a slice of the object (server replies 206).
+        if (size != 0) {
```
jgoldsch12: This can be removed to improve readability

src/utils/object/rdma_protocol.h
```diff
@@ -0,0 +1,109 @@
...
+
+// Aggressive control-plane timeouts (seconds) so a transport stall surfaces fast
+// and the retry path can take over.
+inline constexpr long rdma_connect_timeout_secs = 5;
+inline constexpr long rdma_timeout_secs = 10;
+
+/**
+ * Format the value of the x-amz-rdma-token header.
+ *
+ * Wire format (see AIStor server eos/internal/rdma/rdma.go Request.String):
```
jgoldsch12: Continuing my review comment here that this is a vendor specific wire protocol format (even furthermore with the comment referring to vendor specific code.)  The only fields detailed for the write protocol is descriptor itself which is produced by the Get/Put APIs for token generation from the cuObject libraries.  
An alternative approach, while still attempting to satisfy producing a "generic" accelerated engine, would be to include a vendor specific parameter to include these trailing values to match the AiStor expected fields.

jgoldsch12 (changes requested):

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,545 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
aranadive, line 2: Lets move the rdma files into src/utils/object/rdma to give a clean separation. ANd update meson.build

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,545 @@
...
+        Aws::Utils::DateTime now = Aws::Utils::DateTime::Now();
+        const Aws::String amz_date = now.ToGmtString("%Y%m%dT%H%M%SZ");
+        const Aws::String date_stamp = now.ToGmtString("%Y%m%d");
+
+        // Host header (with port) must be signed and match what is sent.
+        Aws::String host = req.GetUri().GetAuthority();
+        const unsigned p = req.GetUri().GetPort();
+        if (p != 0 && p != 80 && p != 443) {
+            host += ":" + std::to_string(p);
+        }
```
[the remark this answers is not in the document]
aranadive: @harshavardhana is the check correct? It skips setting the port if 80/443 ports are specified currently

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,545 @@
...
+        return rdma_error;
+    }
+    catch (const std::exception &e) {
+        NIXL_ERROR << "rdmaPut failed: " << e.what();
+        return rdma_error;
+    }
+}
+
+ssize_t
+S3RdmaControlPlane::rdmaGet(S3RdmaClientCtx &ctx,
```
aranadive, line 362: Please add [[nodiscard]] to these public functions inline with C++20 standards - instance(), registerBuffer(), isDeviceMemory(), getToken(), valid(), rdmaPut(), rdmaGet(), rdmaPutWithRetry(), rdmaGetWithRetry(), formatRdmaToken(), parseRdmaReply()

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+        const std::string etag =
+            resp->HasHeader("etag") ? stripQuotes(resp->GetHeader("etag").c_str()) : "";
+
+        // Trust the server's reported transferred byte count (can be < requested
+        // for ranged/partial GETs). Publish ctx.etag only once the response is
+        // fully accepted, so a malformed byte count doesn't leave a stale ETag.
+        ssize_t transferred = static_cast<ssize_t>(size);
+        if (resp->HasHeader(amz_rdma_bytes_transferred)) {
+            try {
+                const long long n = std::stoll(resp->GetHeader(amz_rdma_bytes_transferred).c_str());
```
aranadive: use std::from_chars

src/utils/object/rdma_protocol.h
```diff
@@ -0,0 +1,109 @@
...
+ *   "<descriptor>:<start_addr_hex>:<size_hex>"
+ * where the two trailing fields are 16-digit zero-padded lowercase hex. The
+ * server parses the address/size with a base-16 parser, so zero-padding is
+ * accepted and keeps the field widths fixed.
+ */
+inline std::string
+formatRdmaToken(const char *descriptor, uint64_t buf_addr, uint64_t size) {
```
aranadive: descriptor should be std::string_view instead

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+        ctx.etag = etag;
+        return transferred;
+    }
+    catch (const std::exception &e) {
+        NIXL_ERROR << "rdmaGet failed: " << e.what();
+        return rdma_error;
+    }
+}
+
+// ---------------------------------------------------------------------------
```
aranadive: Comments should be in /** ... */ format for proper doc generation

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+        // GetAuthority() drops the port, so set it explicitly — otherwise the
+        // request goes to the scheme default (80/443) and fails to connect.
+        if (port != 0) {
+            uri.SetPort(static_cast<uint16_t>(port));
+        }
+        return uri;
+    }
+};
+
+S3RdmaControlPlane::S3RdmaControlPlane(nixl_b_params_t *custom_params) : impl_(new Impl()) {
```
aranadive: Should be const nixl_b_params_t *

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+    }
+    return ret;
+}
+
+ssize_t
+rdmaGetWithRetry(SharedCuObjClient &rdma,
+                 S3RdmaControlPlane &cp,
+                 S3RdmaClientCtx &ctx,
+                 void *buf,
+                 size_t size,
```
aranadive: Size and offset should be uint64_t to match rdmaGet signature

src/utils/object/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+// is itself transient (cuObject NIC selection / registration hiccup), so it is
+// retried rather than aborting on the first attempt.
+// ---------------------------------------------------------------------------
+
+ssize_t
+rdmaPutWithRetry(SharedCuObjClient &rdma,
+                 S3RdmaControlPlane &cp,
+                 S3RdmaClientCtx &ctx,
+                 void *buf,
+                 size_t size) {
```
aranadive: uint64_t

src/utils/object/rdma/rdma.cpp
```diff
@@ -0,0 +1,545 @@
...
+SharedCuObjClient::deregisterBuffer(void *ptr) {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    if (client_->cuMemObjPutDescriptor(ptr) != CU_OBJ_SUCCESS) {
+        NIXL_WARN << "cuMemObjPutDescriptor failed for ptr " << ptr;
+    }
+}
+
+bool
+SharedCuObjClient::isDeviceMemory(const void *ptr) const {
+    return cuObjClient::getMemoryType(ptr) == CUOBJ_MEMORY_CUDA_DEVICE;
```
aranadive: Add a comment here why the lock isnt required

src/utils/object/rdma/s3_control_plane_http.cpp
aranadive: @harshavardhana lets add a guard here

src/utils/object/rdma.cpp
aranadive: Agreed

src/utils/object/rdma_protocol.h
aranadive: Isnt this what the CUObject/RDMA specification defines to be used as the token?
Maybe call it the CUObject Format?

aranadive (changes requested):
@harshavardhana left comments for the PR. Can you look at the CR and Jason's comments and fix up.

harshavardhana:
Changes are done PTAL

src/utils/object/rdma/rdma_protocol.h
```diff
@@ -0,0 +1,117 @@
...
+ * S3 RDMA endpoint parses them with a base-16 parser, so the fixed field width
+ * is preserved. The descriptor is opaque and of arbitrary length.
+ */
+[[nodiscard]] inline std::string
+formatRdmaToken(std::string_view descriptor, uint64_t buf_addr, uint64_t size) {
+    // Build via std::string so an opaque descriptor of any length is never
+    // truncated; only the two fixed-width hex fields use a small buffer.
+    char suffix[40];
+    std::snprintf(suffix, sizeof(suffix), ":%016" PRIx64 ":%016" PRIx64, buf_addr, size);
+    return std::string(descriptor) + suffix;
```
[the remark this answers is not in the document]
harshavardhana: Done in fc35214c — added @brief/@param/@return to all public APIs in object/rdma (formatRdmaToken, parseRdmaReply, SharedCuObjClient, S3RdmaControlPlane, retry helpers) and documented S3RdmaClientCtx members with ///<.

src/utils/object/rdma/rdma.cpp
```diff
@@ -0,0 +1,568 @@
...
+    // path is meaningless for it. Reject up front — a legitimate zero-byte object
+    // PUT must go through the ordinary HTTP path in the caller.
+    if (size == 0) {
+        NIXL_ERROR << "rdmaPut: zero-size request for key=" << ctx.object;
+        return rdma_error;
+    }
+    try {
+        Aws::Http::URI uri = impl_->buildUri(ctx.bucket, ctx.object);
+        if (!ctx.uploadId.empty()) {
+            if (ctx.partNumber == 0 || ctx.partNumber > 10000) {
```
aranadive: Add a named constant for this

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,568 @@
...
+#include <aws/core/utils/HashingUtils.h>
+#include <aws/core/utils/StringUtils.h>
+
+#include "object/s3/utils.h"
+#include "object/s3/aws_sdk_init.h"
+#include "common/nixl_log.h"
+
+namespace nixl_obj_rdma {
+
+namespace {
```
aranadive, line 35: This file is lifting a lot of things. Maybe split the file into cuobj_client.h/cpp (that handles the cuobj interaction), 
s3_control_plane_http.h/cpp - for the s3rdma control plane class
rdmaRetry.h for the retry logic

src/utils/object/rdma/rdma.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+    // SigV4-sign the request with payload hash "UNSIGNED-PAYLOAD". We sign
+    // manually (rather than via the SDK's AWSAuthV4Signer) because that signer
+    // hashes the empty body over plain HTTP — the S3 RDMA server only skips
+    // content-sha256 validation when the header is exactly UNSIGNED-PAYLOAD, and
+    // the data here travels out-of-band over RDMA. Mirrors minio-cpp/rs SignV4S3.
+    // All non-signed headers (host, x-amz-rdma-token, content-*, checksum) must
+    // already be set on the request before calling this.
+    void
+    signV4(Aws::Http::HttpRequest &req) const {
```
aranadive: This can be move to an anonymous space doesnt belong in Impl

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,568 @@
...
+}
+
+S3RdmaControlPlane::~S3RdmaControlPlane() = default;
+
+ssize_t
+S3RdmaControlPlane::rdmaPut(S3RdmaClientCtx &ctx,
+                            const char *token,
+                            uint64_t buf_addr,
+                            uint64_t size) {
+    // A 0-byte PUT would mint a 0-length RDMA region and move no data; the RDMA
```
aranadive, line 287: Some of the initial parts of rdmaPut and rdmaGet can refactored out as creating the httpRequest and getting a response. You can use a lambda function for put/get response parsing.
std::shared_ptr<Aws::Http::HttpResponse> sendRdmaRequest ...

src/utils/object/rdma/rdma.cpp
```diff
@@ -0,0 +1,568 @@
...
+        // real failure — return rdma_error so the retry path still runs, rather
+        // than misclassifying it as a decline.
+        const std::string reply = resp->HasHeader(amz_rdma_reply) ?
+            std::string(resp->GetHeader(amz_rdma_reply).c_str()) :
+            "";
+        std::ostringstream body;
+        body << resp->GetResponseBody().rdbuf();
+        if (reply == "501") {
+            NIXL_ERROR << "rdmaPut declined: http=" << http_status << " x-amz-rdma-reply='" << reply
+                       << "' url=" << uri.GetURIString() << " body=" << body.str().substr(0, 400)
```
aranadive: named constant for 400 as well

harshavardhana:
Done in b704a46 (file split) + 264c467 (the four control-plane nits). PTAL @aranadive

src/utils/object/rdma/cuobj_client.cpp
```diff
@@ -0,0 +1,93 @@
...
+SharedCuObjClient::SharedCuObjClient() {
+    try {
+        // The token-based flow does not use the get/put callbacks, so empty ops
+        // suffice (matches the reference SDKs' availability probe).
+        client_ = std::make_unique<cuObjClient>(ops_, CUOBJ_PROTO_RDMA_DC_V1);
+        connected_ = client_ && client_->isConnected();
+        if (connected_) {
+            NIXL_INFO << "S3 RDMA fabric connected (cuObject)";
+        } else {
+            NIXL_INFO << "S3 RDMA fabric not connected; transfers use HTTP";
```
jgoldsch12: I like this, but are we ready to do HTTP fail back?

src/utils/object/rdma/rdma_protocol.h
```diff
@@ -0,0 +1,121 @@
...
+// and the retry path can take over.
+inline constexpr long rdma_connect_timeout_secs = 5;
+inline constexpr long rdma_timeout_secs = 10;
+
+/**
+ * @brief Format the value of the x-amz-rdma-token header.
+ *
+ * The token is the cuObject RDMA descriptor (as produced by
+ * cuMemObjGetRDMAToken) followed by the buffer's start address and size:
+ *   "<descriptor>:<start_addr_hex>:<size_hex>"
```
jgoldsch12: start_addr_hex:size_hex must be moved into a different header as it is overloading the descriptor by adding vendor specific fields to the string.

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,460 @@
...
+        const int http_status = static_cast<int>(resp->GetResponseCode());
+        const std::string etag =
+            resp->HasHeader("etag") ? stripQuotes(resp->GetHeader("etag").c_str()) : "";
+
+        // Success: the server completed the RDMA_READ and returns a standard
+        // HTTP 200 + ETag (the object payload moved out-of-band, so the HTTP body
+        // is empty). Matches minio-cpp/minio-rs rdmaPut.
+        if (http_status == 200 && !etag.empty()) {
+            ctx.etag = etag;
+            if (resp->HasHeader("x-amz-checksum-crc64nvme")) {
```
jgoldsch12, line 328: what if another checksum alg is used?

jgoldsch12 (commented):
Thanks for continuing to make updates.  Reposting the token + vendor fields comment

src/utils/object/rdma/cuobj_client.cpp
harshavardhana: There's no HTTP fallback on the accelerated path. When the fabric isn't available, `SharedCuObjClient::instance()` returns `nullptr` and the caller treats it as a hard error; `rdmaPut`/`rdmaGet` return negative sentinels (`rdma_not_supported` / `rdma_error`) that surface as an error rather than silently retrying over HTTP. The earlier log line implying "transfers use HTTP" was misleading and was removed in the latest commit. Resolving as addressed.

src/utils/object/rdma/rdma_protocol.h
harshavardhana: The `start_addr_hex:size_hex` suffix pre-dates cuObject 1.0 — the descriptor already carries addr/size in its leading fields, so it's redundant and now removed; the token is the unmodified cuObject descriptor.

src/utils/object/rdma/s3_control_plane_http.cpp
harshavardhana: `x-amz-checksum-crc64nvme` is the best-available full-object checksum here — it's ideal for validating full-object integrity even across multipart uploads. Other algorithms exist but don't strike the same balance of speed and checksum strength; crc64nvme is the most balanced of the lot. Since NIXL is an application (not a general-purpose SDK), it can choose what it uses — AWS S3-compatible crc64nvme — and the server implements the corresponding checksum algorithms accordingly.

harshavardhana:
@aranadive PTAL

src/utils/object/rdma/rdma.h
```diff
@@ -0,0 +1,37 @@
...
+//
+// RDMA is NOT a separate engine or a vendor plugin. It is an optimization of
+// the normal S3 GET/PUT path on the standard client, enabled per backend via
+// `accelerated=true` (generic S3-over-RDMA): the client issues an out-of-band
+// RDMA transfer over the published `x-amz-rdma-*` protocol. Under
+// `accelerated=true` an RDMA decline/failure is a hard
+// error — there is no silent HTTP fallback today, because a server that ignores
+// the token (instead of returning `x-amz-rdma-reply: 501`) would accept a
+// body-less PUT as a 0-byte object (see s3/client.cpp). The protocol is an AWS
+// S3 convention (not vendor-specific), so the same code works against MinIO
```
jgoldsch12: Please remove AiStor reference.

src/utils/object/rdma/rdma_protocol.h
```diff
@@ -0,0 +1,98 @@
...
+
+#ifndef NIXL_SRC_UTILS_OBJECT_RDMA_RDMA_PROTOCOL_H
+#define NIXL_SRC_UTILS_OBJECT_RDMA_RDMA_PROTOCOL_H
+
+// Generic S3-over-RDMA wire protocol helpers.
+//
+// This header is intentionally free of any AWS SDK or cuObject dependency so the
+// protocol logic can be unit-tested on its own. It encodes the published S3 RDMA
+// convention (the `x-amz-rdma-*` headers) used by NVIDIA cuObject and implemented
+// by any compliant S3 endpoint (MinIO AIStor today; transparently usable against
```
jgoldsch12: Please remove AIStor reference

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    // SigV4-sign the request with payload hash "UNSIGNED-PAYLOAD". A free
+    // function (not a member): it derives everything from the request plus the
+    // resolved credentials/region, so it has no dependency on the control plane's
+    // internals. We sign manually (rather than via the SDK's AWSAuthV4Signer)
+    // because that signer hashes the empty body over plain HTTP — the S3 RDMA
+    // server only skips content-sha256 validation when the header is exactly
+    // UNSIGNED-PAYLOAD, and the data here travels out-of-band over RDMA. Mirrors
+    // minio-cpp/rs SignV4S3. All non-signed headers (host, x-amz-rdma-token,
```
jgoldsch12: Please remove minio reference.

jgoldsch12 (approved):
Latest changes look good - thank you for making them.

Some feedback about removing minio references in the comments.  I'm not certain how these help the reader.

aranadive:
Thanks @harshavardhana, @jgoldsch12. Can we remove the specific MinIO/AIStor references?

harshavardhana:
@aranadive latest review feedback is addressed and the branch is up to date with `main`:

- Removed the MinIO/AIStor references from the S3-over-RDMA comments per @jgoldsch12's note (`7d96d1f3`) — the wire convention is generic, so the comments are now vendor-neutral.
- The earlier token/descriptor and HTTP-fallback threads are resolved (the `start_addr:size` append is gone; under `accelerated=true` a decline/failure is a hard error, no silent HTTP fallback).

@jgoldsch12 has approved. This is the only PR gating the next release together with the generic S3 RDMA follow-up — could you re-review to clear the earlier changes-requested so we can get it merged? Thanks!

src/utils/object/rdma/rdmaRetry.cpp
```diff
@@ -0,0 +1,72 @@
...
+    ssize_t ret = -1;
+    for (int attempt = 0; attempt < rdma_max_attempts; ++attempt) {
+        char *token = rdma.getToken(buf, size, 0, CUOBJ_GET);
+        if (token == nullptr) {
+            ret = -1;
+            continue; // transient mint failure: retry
+        }
+        ret = cp.rdmaGet(ctx, token, size, offset);
+        rdma.putToken(token);
+        if (ret > 0 || ret == rdma_not_supported) {
```
vvenkates27: Could `ret > 0` turn a legitimate 0-byte GET into a retry/failure?

`rdmaGet` already rejects `size == 0`, so this only matters if the server answers 200/206 with `x-amz-rdma-bytes-transferred: 0` (empty object, or a range that transferred nothing). I do not know if your target endpoints ever do that versus 416.

If that is in scope, would `ret >= 0` be safer here? No need to change PUT if it never returns 0 today.

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,449 @@
...
+        }
+
+        const int http_status = static_cast<int>(resp->GetResponseCode());
+        const std::string etag =
+            resp->HasHeader("etag") ? stripQuotes(resp->GetHeader("etag").c_str()) : "";
+
+        // Success: the server completed the RDMA_READ and returns a standard
+        // HTTP 200 + ETag (the object payload moved out-of-band, so the HTTP body
+        // is empty). Matches the reference client implementations.
+        if (http_status == 200 && !etag.empty()) {
```
vvenkates27, line 326: PUT success is HTTP 200 + ETag, and we return the requested `size` rather than a verified byte count. That matches a server that only 200s after a full RDMA_READ.

Two things I was not 100% sure about, so flagging as questions:

1. If an endpoint ever 200+ETags a short RDMA write, we would still report a full PUT (e.g. a multipart part). Do you consider that out of scope?
2. `rdma_reply_no_content` (204) is documented as PUT success in `rdma_protocol.h`, but this path does not accept 204. If none of the target stores use 204 for PutObject/UploadPart, the comment/constant might just need a tweak.

src/utils/object/rdma/s3_control_plane_http.cpp
```diff
@@ -0,0 +1,449 @@
...
+            return rdma_error;
+        }
+
+        const std::string etag =
+            resp->HasHeader("etag") ? stripQuotes(resp->GetHeader("etag").c_str()) : "";
+
+        // Trust the server's reported transferred byte count (can be < requested
+        // for ranged/partial GETs). Publish ctx.etag only once the response is
+        // fully accepted, so a malformed byte count doesn't leave a stale ETag.
+        ssize_t transferred = static_cast<ssize_t>(size);
```
vvenkates27: Would you mind confirming the contract when `x-amz-rdma-bytes-transferred` is absent?

Right now transferred defaults to the requested `size`. That is fine if every GET passes the exact object/part length from HEAD and the server always writes that many bytes. It is less fine if a caller passes an oversized buffer (last KV/checkpoint chunk) and a shorter object, and the endpoint omits the header — we would report a full fill.

If the protocol requires that header on GET, could we fail closed when it is missing instead of assuming `size`? Completely fine if you have already verified the servers you care about always send it.

src/utils/object/rdma/rdmaRetry.cpp
harshavardhana: Good catch — fixed. A GET can legitimately transfer 0 bytes (an empty object or an empty range), so the success check is now `ret >= 0`; only a negative `rdma_error` is transient and retried, a decline is terminal. Pairs with the GET byte-count fix below.

src/utils/object/rdma/s3_control_plane_http.cpp
harshavardhana: On both points: (1) A PUT `200 + ETag` is only returned after the full RDMA_READ (and multipart complete) succeeds — the endpoint never `200`s a short/partial write, so reporting the requested `size` is correct, including for a multipart part; a partial write surfaces as a decline/failure. (2) You're right the `204` comment was misleading — 204 is No Content (S3 DELETE semantics), not a PUT-success code, and this path only does GET/PUT. Corrected the comment; PUT success is `200`, ranged GET `206`.

src/utils/object/rdma/s3_control_plane_http.cpp
harshavardhana: Confirmed the contract and fixed it. The server reports the actual transferred count in `x-amz-rdma-bytes-transferred`, already clamped to the servable object/range length — so an oversized buffer can't over-report: the header carries the true (smaller) count. The header is present iff that count is > 0, so its absence on an accepted 200/206 means a zero-byte transfer (empty object/range). The client now reports `0` in that case instead of defaulting to the requested `size`. Not fail-closed, since a 0-byte transfer is a legitimate success rather than a missing-header error.

aranadive (approved):

harshavardhana:
@aranadive the two failing checks look non-code — blossom `Authorization` is the #2219 auto-trigger regression on unsigned commits, and `nixl-ci-dl-gpu-ep` is a flake that's green on every other current PR — is there anything I need to do to get the builds green, or just a `/build` re-trigger?

vvenkates27 (approved):

PR #1848, opened 2026-06-29
Skip nixl_ep meta module install when option is disabled

jbelloncastro:
## What?
Disables nixl_ep python module from nixl-meta wheel when the corresponding option is disabled

## Why?
Nixl EP bindings module is only built for the CUDA platform. Otherwise, its build is skipped but the nixl-meta wheel package still installs the nixl_ep module unconditionally.

Some projects such as vLLM decide to import the module or not based on the visibility of the module, not whether it loads correctly or not (see https://github.com/vllm-project/vllm/blob/0ca39c4f1fc450339f57ceca6bddc2af1abe84a5/vllm/utils/import_utils.py#L485-L487).

## How?
The simplest solution is to leverage the current option `build_nixl_ep` such that the meta module is not installed when the option is disabled.

After the change, a NIXL installation with disabled `build_nixl_ep` will trigger a _module not found_ error when running `python -c "import nixl_ep"`, instead of a module import due to neither `nixl_ep_cu12` nor `nixl_ep_13`  being found.

## Summary by CodeRabbit

* **Refactor**
  * Improved the Python meta package wheel build setup by consolidating and structuring build inputs.
  * Ensured optional artifacts are included consistently when the related build option is enabled, leading to more reliable and repeatable package builds.


src/bindings/python/nixl-meta/meson.build
```diff
@@ -38,21 +38,27 @@ subdir('nixl')
...
       nixl_init_copy,
       nixl_api_copy,
       nixl_logging_copy,
-      nixl_ep_init_copy,
-    ],
+    ]
+
+if get_option('build_nixl_ep')
+  meta_inputs +=  [nixl_ep_init_copy]
+endif
```
rakhmets: Please move `meta_inputs` under `if uv.found()`.

src/bindings/python/nixl-meta/meson.build
```diff
@@ -38,21 +38,27 @@ subdir('nixl')
...
       license_path,
       nixl_init_copy,
       nixl_api_copy,
       nixl_logging_copy,
-      nixl_ep_init_copy,
-    ],
+    ]
+
+if get_option('build_nixl_ep')
+  meta_inputs +=  [nixl_ep_init_copy]
```
rakhmets: It seems that `subdir('nixl_ep')` should be moved here.

roiedanino:
/build

@jbelloncastro - 
image

src/bindings/python/nixl-meta/meson.build
```diff
@@ -35,7 +35,10 @@ root_license_path = join_paths(source_root, 'LICENSE')
 license_path = fs.copyfile(root_license_path)
 
 subdir('nixl')
-subdir('nixl_ep')
+
+if get_option('build_nixl_ep')
```
ovidiusm, line 40: While this is the right way to fix it at this level, our wheel build scripts have the flag enabled, so the wheel pulled by vllm will contain EP

Are you building the wheel separately for AMD? Could you please provide some info on how the packaging works as I do not know the details

jbelloncastro: I am personally building the wheel locally. For AMD builds, I believe a new flag is being added [in PR #1873](https://github.com/ai-dynamo/nixl/pull/1873/changes#diff-76ea9078099f2ee8cc0e760a874dd153f225ba16ad10375c24ab6cd53518f3c4).

That flag would enable EP to be built for ROCm. Once that happens, I imagine the only change necessary would be to make sure that the inferface module can not only detect Nixl EP for cu12 or cu13 but also for rocm.

In my particular case I am building a Nixl backend wheel with UCX backend enabled, using different name via meson option `-Dwheel_variant=rocm`. Then, I install it and the nixl-meta wheel too.

ovidiusm (dismissed):

PR #1852, opened 2026-06-30
tracing: NVTX completeness + cross-thread correlation

e-eygin:
## What?

Complete the NVTX trace backend (NIX-1540 / PR7) and add cross-thread request
correlation (NIX-1538 / PR5b): typed attribute payloads on ranges,
registered-string span labels, broader agent call sites, and a shared
correlation id linking a transfer's post to its completion across threads.

## Why?

PR1/PR2 (#1765 + #1845) made NVTX functional but incomplete: attributes were
lossy `key=value` marks, span names were allocated per `beginSpan`, and
`loadRemoteMD` / `prepMemView` were missing from the timeline. Separately, an
async transfer showed up as two unrelated events — `postXferReq` on the caller
thread and completion on the polling thread — with no way to tell they belong to
the same request. This PR closes both gaps so a single nsys capture shows the
full NIXL agent lifecycle with inspectable typed metadata and post↔complete
correlation.

## How?

**NVTX completeness**
* **Structured payloads:** register static NVTX payload schemas (int64/double/string
  key/value); accumulate attrs on `NvtxSpan` and attach via `nvtxRangePopPayload`.
* **Registered strings:** `nvtxDomainRegisterStringA` for all fixed `nixl::*` names.
* **Plugin modules:** `nvtx_payload_schemas`, `nvtx_events`, `nvtx_span`,
  `nvtx_trace_backend` (+ thin `createNvtxBackend` factory).
* **Call sites:** `loadRemoteMD`, `fetchRemoteMD`, `prepMemView`, `releaseMemView`
  in `nixl_agent.cpp`.

**Cross-thread correlation** (tracing-only; no core-struct change)
* `CorrelationScope` RAII guard + `NIXL_TRACE_CORRELATION_SCOPE` on the facade;
  `Tracer::push/popCorrelationId` fan out to backends (previously no-ops).
* NVTX keeps a per-thread correlation-id stack and stamps the active id as the
  event's `uint64` payload on both ranges and marks.
* Correlation id is the transfer-request handle address, so `postXferReq` and the
  `xfer.complete` mark share it even when posted and polled from different threads.

**Out of scope:** NIX-1576 (nsys auto-enable, merged separately in #1867); UCX
backend-internal sub-spans (NIX-1538 / PR5a, de-scoped); Chakra.

## Test plan

* `ninja -C build/trace src/plugins/tracing/nvtx/libtrace_backend_nvtx.so`
* `./build/trace/test/gtest/unit/unit --gtest_filter='Tracing.*'` (11 pass, incl.
  3 correlation tests)
* `./build/trace/test/gtest/gtest --tests_plugin_dirs=... --gtest_filter='*TestTransferTracing*'` (8 pass)
* `meson test -C build/trace tracing_nsys`

## Summary by CodeRabbit

* **New Features**
  * Enhanced NVTX tracing with richer, typed span attributes and dedicated timeline markers for remote metadata exchange and the memory-view lifecycle.
  * Added correlation scopes to automatically propagate tracing correlation IDs for more consistent cross-thread linking.
* **Bug Fixes**
  * Improved NVTX correlation-id propagation to better link related events across threads.
* **Tests**
  * Strengthened NVTX capture validation to ensure expected ranges are present.
  * Added an NVTX transfer test covering metadata exchange via the direct blob exchange path.
* **Build/Chores**
  * Updated NVTX backend detection to require both headers needed for payload support.

### DIFF
--- a/src/core/tracing/trace.h
+++ b/src/core/tracing/trace.h
@@ -185,6 +185,38 @@ private:
     std::vector<std::unique_ptr<TraceBackend>> backends_;
 };
 
+/**
+ * @brief RAII guard that keeps a correlation id active on the tracer for its
+ *        lifetime, so every span/mark created while it is in scope is tagged
+ *        with that id. Null-tracer-safe. Used to link spans emitted on different
+ *        threads (e.g. postXferReq on the caller thread vs. the completion on
+ *        the polling thread) to the same logical request.
+ */
+class CorrelationScope {
+public:
+    CorrelationScope(Tracer *tracer, std::uint64_t id) noexcept : tracer_(tracer) {
+        if (tracer_ != nullptr) {
+            tracer_->pushCorrelationId(id);
+        }
+    }
+
+    ~CorrelationScope() {
+        if (tracer_ != nullptr) {
+            tracer_->popCorrelationId();
+        }
+    }
+
+    CorrelationScope(const CorrelationScope &) = delete;
+    CorrelationScope &
+    operator=(const CorrelationScope &) = delete;
+    CorrelationScope(CorrelationScope &&) = delete;
+    CorrelationScope &
+    operator=(CorrelationScope &&) = delete;
+
+private:
+    Tracer *tracer_;
+};
+
 /** @brief Inputs used to build the composite tracer from enabled backends. */
 struct TracerConfig {
     /** @brief Agent name; used e.g. as the NVTX domain name. */
--- a/src/core/tracing/trace.h
+++ b/src/core/tracing/trace.h
@@ -194,7 +194,7 @@ private:
  */
 class CorrelationScope {
 public:
-    CorrelationScope(Tracer *tracer, std::uint64_t id) noexcept : tracer_(tracer) {
+    [[nodiscard]] CorrelationScope(Tracer *tracer, std::uint64_t id) : tracer_(tracer) {
         if (tracer_ != nullptr) {
             tracer_->pushCorrelationId(id);
         }
--- a/src/plugins/tracing/nvtx/nvtx_events.cpp
+++ b/src/plugins/tracing/nvtx/nvtx_events.cpp
@@ -81,14 +81,6 @@ namespace {
         return ev;
     }
 
-    [[nodiscard]] nvtxEventAttributes_t
-    makeEvent(const char *ascii_name, const Kind kind) {
-        nvtxEventAttributes_t ev = makeEventBase(kind);
-        ev.messageType = NVTX_MESSAGE_TYPE_ASCII;
-        ev.message.ascii = ascii_name;
-        return ev;
-    }
-
     [[nodiscard]] nvtxStringHandle_t
     lookupRegistered(std::string_view name,
                      const std::vector<nvtxStringHandle_t> &handles) noexcept {
@@ -115,13 +107,17 @@ registerSpanNames(const nvtxDomainHandle_t domain) {
 nvtxEventAttributes_t
 eventForName(const std::string_view name,
              const Kind kind,
-             const std::vector<nvtxStringHandle_t> &registered_handles,
-             std::string &fallback_storage) {
-    if (const nvtxStringHandle_t registered = lookupRegistered(name, registered_handles)) {
-        return makeEvent(registered, kind);
+             const nvtxDomainHandle_t domain,
+             const std::vector<nvtxStringHandle_t> &registered_handles) {
+    nvtxStringHandle_t handle = lookupRegistered(name, registered_handles);
+    if (handle == nullptr) {
+        // Not one of the fixed nixl::* names: register it now so the event still
+        // carries the real label. Registered strings need no caller-owned storage
+        // (unlike an ASCII message), keeping the call sites simple.
+        const std::string owned(name);
+        handle = nvtxDomainRegisterStringA(domain, owned.c_str());
     }
-    fallback_storage.assign(name);
-    return makeEvent(fallback_storage.c_str(), kind);
+    return makeEvent(handle, kind);
 }
 
 } // namespace nixl::trace::nvtx_internal
--- a/src/plugins/tracing/nvtx/nvtx_events.cpp
+++ b/src/plugins/tracing/nvtx/nvtx_events.cpp
@@ -1,0 +1,127 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "nvtx_events.h"
+
+#include <cstdint>
+#include <iterator>
+
+namespace nixl::trace::nvtx_internal {
+namespace {
+
+    constexpr const char *kRegisteredSpanNames[] = {
+        "nixl::registerMem",
+        "nixl::deregisterMem",
+        "nixl::makeConnection",
+        "nixl::makeXferReq",
+        "nixl::createXferReq",
+        "nixl::postXferReq.write",
+        "nixl::postXferReq.read",
+        "nixl::genNotif",
+        "nixl::getNotifs",
+        "nixl::xfer.complete",
+        "nixl::loadRemoteMD",
+        "nixl::fetchRemoteMD",
+        "nixl::prepMemView",
+        "nixl::releaseMemView",
+    };
+
+    [[nodiscard]] constexpr std::uint32_t
+    colorFor(const Kind kind) noexcept {
+        switch (kind) {
+        case Kind::Generic:
+            return 0xFF455A64u; // blue-gray
+        case Kind::Compute:
+            return 0xFF00838Fu; // teal
+        case Kind::MemoryR:
+            return 0xFFEF6C00u; // orange
+        case Kind::MemoryW:
+            return 0xFFF9A825u; // amber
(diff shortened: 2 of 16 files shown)


test/gtest/run_nsys_tracing.sh
```diff
@@ -51,5 +51,16 @@ if [ ! -f "${OUT}.nsys-rep" ]; then
     exit 1
 fi
 
+# Sanity-check expected NVTX ranges and the new metadata-exchange span.
+NVTX_RANGES="$(nsys stats --force-export=true --report nvtx_sum --format csv "${OUT}.nsys-rep" 2>/dev/null \
+    | grep 'nixl::' | sed 's/.*,//' | sort -u)"
+for expected in nixl::loadRemoteMD nixl::registerMem nixl::postXferReq.write; do
+    if ! echo "${NVTX_RANGES}" | grep -q "${expected}"; then
```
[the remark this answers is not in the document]
e-eygin: Fixed in `29b3b953`.

e-eygin:

e-eygin:
@coderabbitai review

src/plugins/tracing/nvtx/nvtx_payload_schemas.cpp
```diff
@@ -0,0 +1,146 @@
...
+                             sizeof(Int64Attr));
+
+    const nvtxPayloadSchemaAttr_t kDoubleAttrSchemaAttr =
+        makeStaticSchemaAttr("nixl.trace.double_attr",
+                             kDoubleAttrSchemaEntries,
+                             std::size(kDoubleAttrSchemaEntries),
+                             sizeof(DoubleAttr));
+
+    const nvtxPayloadSchemaAttr_t kStrAttrSchemaAttr =
+        makeStaticSchemaAttr("nixl.trace.str_attr",
```
ColinNV: Why does `str_attr` use a shortened form of `string` while `double` is not shortened?

src/plugins/tracing/nvtx/nvtx_payload_schemas.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <nvtx3/nvToolsExt.h>
+
+namespace nixl::trace::nvtx_internal {
+
+/** @brief Domain-registered schema IDs for typed span attribute payloads. */
+struct PayloadSchemaIds {
+    std::uint64_t int64_attr{};
+    std::uint64_t double_attr{};
+    std::uint64_t str_attr{};
```
ColinNV: Same here regarding the `str` naming.

src/plugins/tracing/nvtx/nvtx_span.h
```diff
@@ -0,0 +1,75 @@
...
+#include <vector>
+
+#include <nvtx3/nvToolsExt.h>
+
+#include "nvtx_payload_schemas.h"
+#include "tracing/trace.h"
+
+namespace nixl::trace::nvtx_internal {
+
+/** @brief One active NVTX range; typed attributes are attached on pop. */
```
ColinNV: Not quite sure how to parse this comment.

src/plugins/tracing/nvtx/nvtx_span.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+void
+NvtxSpan::emitStrAttr(const std::string_view key, const std::string_view value) {
+    auto stored = std::make_unique<StoredPayload>();
+    stored->key_storage.assign(key);
+    stored->value_storage.assign(value);
+    stored->str.key = stored->key_storage.c_str();
+    stored->str.value = stored->value_storage.c_str();
+    stored->data = {schemaIds_.str_attr, sizeof(StrAttr), &stored->str};
+    payloads_.push_back(std::move(stored));
```
ColinNV: This can invalidate `stored->str.key/value` due to the small string optimization. Suggest switching `payloads_` to a `std::list<StoredPayload>` to prevent them from being moved, doesn't increase the number of allocations and `emplace()`ing directly also eliminates the move.

src/plugins/tracing/nvtx/nvtx_span.cpp
```diff
@@ -0,0 +1,118 @@
...
+    refs.reserve(payloads_.size());
+    for (const auto &stored : payloads_) {
+        refs.push_back(stored->data);
+    }
+    nvtxRangePopPayload(domain_, refs.data(), refs.size());
+}
+
+void
+NvtxSpan::addAttribute(const std::string_view key, const std::string_view value) {
+    emitStrAttr(key, value);
```
ColinNV: Why this function indirection?

src/plugins/tracing/nvtx/nvtx_span.cpp
```diff
@@ -0,0 +1,118 @@
...
+        const char *value;
+    };
+
+} // namespace
+
+struct NvtxSpan::StoredPayload {
+    nvtxPayloadData_t data{};
+    std::string key_storage;
+    std::string value_storage;
+    Int64Attr int64{};
```
ColinNV: Since only one attribute is ever used at a time (correct?), can we use a `std::variant`, or at least only a single `key` for all three?

src/plugins/tracing/nvtx/nvtx_events.cpp
```diff
@@ -0,0 +1,127 @@
...
+        nvtxEventAttributes_t ev = makeEventBase(kind);
+        ev.messageType = NVTX_MESSAGE_TYPE_ASCII;
+        ev.message.ascii = ascii_name;
+        return ev;
+    }
+
+    [[nodiscard]] nvtxStringHandle_t
+    lookupRegistered(std::string_view name,
+                     const std::vector<nvtxStringHandle_t> &handles) noexcept {
+        for (std::size_t i = 0; i < std::size(kRegisteredSpanNames); ++i) {
```
ColinNV, line 87: Probably not worth it, but one could in theory sort `kRegisteredSpanNames` manually once, add a `static_assert` that it is sorted, and do a binary search here.

ColinNV (changes requested):

src/plugins/tracing/nvtx/nvtx_span.cpp
e-eygin: Went with the single-key option. `std::variant `can't back an NVTX payload — the schema needs a standard-layout struct with a fixed `offsetof(value)/sizeof` (NVTX reads raw bytes there), and variant is neither offset-stable nor standard-layout and carries a discriminant NVTX can't parse. Collapsed the three structs into one shared `AttrPayload` (single `key` + `union` value); the schema id already discriminates the type. `32867f02`.

src/plugins/tracing/nvtx/nvtx_span.cpp
e-eygin: Done in `32867f02` — `payloads_` is now `std::list<StoredPayload>` with in-place `emplace_back`, so nodes never move and the `key`/`string_value` pointers stay valid until pop.

src/plugins/tracing/nvtx/nvtx_span.cpp
e-eygin: Removed — the `emit*` helpers are inlined into the `addAttribute` overloads (only a small `addPayload(key)` remains to share the emplace+key step). `32867f02`.

src/plugins/tracing/nvtx/nvtx_span.h
e-eygin: Reworded: "One active NVTX range. Attributes are buffered as typed payloads and attached to the range when it is popped in the destructor." `32867f02`.

src/plugins/tracing/nvtx/nvtx_payload_schemas.h
e-eygin: Renamed to `string_attr` / `"nixl.trace.string_attr"` for consistency with `int64`/`double`. `32867f02`.

src/plugins/tracing/nvtx/nvtx_payload_schemas.cpp
e-eygin: Done in `32867f02`.

src/plugins/tracing/nvtx/nvtx_events.cpp
e-eygin: Left as a linear scan — the table is ~14 fixed names scanned once per span begin; a sorted invariant + `static_assert` + binary search adds maintenance for no measurable gain at this size. Happy to revisit if the list grows.

src/core/tracing/trace.h
```diff
@@ -185,6 +185,38 @@ class Tracer {
...
+/**
+ * @brief RAII guard that keeps a correlation id active on the tracer for its
+ *        lifetime, so every span/mark created while it is in scope is tagged
+ *        with that id. Null-tracer-safe. Used to link spans emitted on different
+ *        threads (e.g. postXferReq on the caller thread vs. the completion on
+ *        the polling thread) to the same logical request.
+ */
+class CorrelationScope {
+public:
+    CorrelationScope(Tracer *tracer, std::uint64_t id) noexcept : tracer_(tracer) {
```
ColinNV: `pushCorrelationId()` is not `noexcept`...

ColinNV: Could also be `[[nodiscard]]` in case somebody uses it outside of the macros?

src/plugins/tracing/nvtx/nvtx_events.cpp
```diff
@@ -0,0 +1,127 @@
...
+        handles.push_back(nvtxDomainRegisterStringA(domain, name));
+    }
+    return handles;
+}
+
+nvtxEventAttributes_t
+eventForName(const std::string_view name,
+             const Kind kind,
+             const std::vector<nvtxStringHandle_t> &registered_handles,
+             std::string &fallback_storage) {
```
ColinNV: Why is this an argument, it doesn't seem like the callers are using it?

ColinNV (approved):

src/core/tracing/trace.h
e-eygin: Dropped `noexcept`. `5e186eeb`.

e-eygin: Added on the ctor. `5e186eeb`.

src/plugins/tracing/nvtx/nvtx_events.cpp
e-eygin: Removed the out-param; `eventForName` now registers an unknown name on the domain on demand and returns a registered-string event. `5e186eeb`.

Issue #1781, opened 2026-06-15
[FEATURE]: Add support for AMD Infinity Storage for DMA to/from AMD GPU memory for storage IO


sbates130272:
NIXL was recently updated to support AMD GPUs. We extend the functionality by adding support for AMD Infinity Storage which is comparable to NVIDIA's GPU Direct Storage.

AMD Infinity Storage is enabled via the [ROCm hipFile][ref-hipfile] library and we integrate NIXL support for this for both local NVMe based storage and shared RDMA-based filesystem storage.

I plan to work on a PR for this. So opening an issue so we can track that.

[ref-hipfile]: https://github.com/ROCm/hipFile

sbates130272:
Note that I plan to break this out and to start with a PR for just a new AIS_MT plugin which is aligned with the existing GDS_MT plugin. And updates to both nixl and nixlbench to support this. After that I will work up something for rocm_ais which will be aligned with cuda_gds. Stay tuned!

PR #1856, opened 2026-06-30
Consolidate GDS and GDS_MT under cuda_gds

### LINKED ISSUE #1855: Consolidate GDS and GDS_MT under cuda_gds
## What I’d like to change

We currently have `GDS` and `GDS_MT` in separate plugin directories. They intentionally submit I/O differently, but most of the surrounding cuFile code is duplicated: driver setup, file and buffer registration, metadata, `queryMem`, validation, and request preparation.

I’d like to move the standalone `gds_mt` implementation under `src/plugins/cuda_gds` and share that common plumbing. Both backend names would remain available:

- `GDS` would continue to use the cuFile batch API.
- `GDS_MT` would continue to use its Taskflow executor and one `cuFileRead`/`cuFileWrite` task per request.

This should make it less likely that registration or lifetime fixes land in one backend but not the other. As part of the merge, I also want the shared-fd file handles and pooled GDS batch handles to have clear ownership, since the current GDS paths can otherwise invalidate or reclaim objects too early.

The existing backend names, memory types, configuration defaults, path-mode behavior, and static/dynamic plugin support should stay compatible. The initialization options should also be visible through `get_backend_options`.

## Related follow-up work

While working on the consolidation, I found a few areas that may be worth improving later. They are not required to merge the plugins, and each one changes runtime behavior enough to deserve its own hardware testing and review:

- **Spreading a GDS batch across submitting CPUs:** Today `GDS` submits a cuFile batch from one CPU. A future performance change could divide a large batch across several worker threads/CPUs so the storage stack has a chance to use more NVMe hardware queues. That introduces new scheduling, CPU-affinity, concurrency, and tuning decisions, so I don’t want to mix it into a source-code consolidation.

- **How requests larger than `max_request_size` are represented:** With the default 16 MiB limit, a 32 or 64 MiB transfer is submitted as multiple chunks. The current GDS code advances the memory pointer for each chunk, while cuFile also provides `devPtr_offset` for expressing an offset from a base pointer. It may be worth testing separately whether the choice of representation affects registration reuse or performance across cuFile versions and direct/compatibility paths. This consolidation does not change that behavior or assume that one representation is better.

- **Existing completion and release behavior:** This consolidation preserves behavior that predates it. `postXfer()` is asynchronous in both backends: `GDS` submits the cuFile batch and returns while I/O is in progress, and `GDS_MT` schedules its Taskflow work and returns `NIXL_IN_PROG`. Their status and release behavior differs today:
  - `GDS::checkXfer()` asks cuFile for all remaining batch events and can wait. NIXL's `releaseXferReq()` calls `checkXfer()` before attempting backend cancellation, so releasing an active GDS request can wait there.
  - `GDS_MT::checkXfer()` uses a zero-duration future poll and does not wait, but `GDS_MT::releaseReqH()` destroys the request, whose destructor waits for the Taskflow future when work is still active.

  The [NIXL Backend Guide](https://github.com/ai-dynamo/nixl/blob/main/docs/BackendGuide.md#transfer-operations) says that `postXfer()` should not wait for transfer completion and that `releaseXferReq()` should be nonblocking and asynchronous. It does not separately state that `checkXfer()` must be nonblocking, but a wait there can make release block because the agent calls it first. A separate change can evaluate whether and how to make these existing completion and active-release paths fully nonblocking, including safe cancellation and request/buffer lifetimes.

Before opening the upstream PR, I’ll run the focused GDS/GDS_MT tests, static and dynamic builds, 64 MiB data-consistency checks, and a baseline-versus-change benchmark on GDS hardware.

I have a draft implementation and initial GH200 results in my fork for review: https://github.com/maheshrbapatu/nixl/pull/1

maheshrbapatu:
## What?

Consolidate the standalone `GDS_MT` sources into `src/plugins/cuda_gds` while preserving both public backend names and exposing both transfer strategies through the primary `GDS` name.

* `GDS` uses the cuFile batch engine by default and accepts `mode=batch` or `mode=mt` to select the batch or multi-threaded engine.
* `GDS_MT` continues to expose the multi-threaded engine as a compatibility backend name while downstream users migrate to `GDS` with `mode=mt`.
* Both backends share the cuFile driver lifecycle, file and buffer registration, metadata, `queryMem`, validation, and request preparation.
* Shared file descriptors use refcounted file-handle ownership, and the GDS engine owns its pooled batch handles for their full lifetime.
* Existing backend options and their defaults are exposed through `get_backend_options`.
* Build wiring, CODEOWNERS, documentation, and focused GDS tests are updated for the consolidated source layout.

### Backend selection

| Backend | Parameters | Transfer engine |
| -- | -- | -- |
| `GDS` | none or `mode=batch` | cuFile batch |
| `GDS` | `mode=mt` | Multi-threaded Taskflow |
| `GDS_MT` | none | Multi-threaded Taskflow compatibility entry point |

## Why?

The two backends intentionally submit I/O differently, but the surrounding cuFile integration was duplicated across separate plugin directories. Keeping that code in sync made registration and lifetime fixes easy to apply to one backend but miss in the other.

This change gives the common behavior one implementation, exposes both strategies through `GDS`, and retains `GDS_MT` as a compatibility path.

Closes ai-dynamo/nixl#1855.

## How?

`nixlGdsEngine` is now the abstract shared base. `nixlGdsBatchEngine` and `nixlGdsMtEngine` inherit from it and implement only their backend-specific preparation, posting, completion, and release behavior.

Meson builds a shared internal GDS engine archive and two plugin entry points. `GDS` links both concrete engines and selects one from `mode`; `GDS_MT` is a thin MT-only compatibility entry point. Dynamic and static discovery continue to expose both backend names. Because `GDS` now supports `mode=mt`, Taskflow is required when either entry point is built.

The class and runtime flow diagrams are in the [issue design comment](<https://github.com/ai-dynamo/nixl/issues/1855#issuecomment-4845868169>).

### Compatibility and scope

This PR preserves the backend names, plugin versions, supported memory types, path-mode behavior, GDS/GDS_MT mutual exclusion, and existing completion/release behavior. Existing `GDS` calls without `mode` still select the batch strategy, and existing `GDS_MT` calls still select the Taskflow MT strategy. `GDS` adds the optional `mode=batch|mt` selector and advertises the parameters for both strategies.

`GDS_MT` is not removed in this PR. Consumers can migrate to `GDS` with `mode=mt`; the standalone compatibility name can be removed in a later release after downstream users have migrated.

Multi-CPU batch submission, DRAM-path changes, split-request `devPtr_offset` handling, and nonblocking release/cancellation redesign are intentionally left for follow-up changes.

### Validation

* GCC 13/Meson build with warnings as errors using CUDA 13.3, cuFile 1.18, and Taskflow 3.10.
* Dynamic focused GDS/GDS_MT suite: 18 passed; one unrelated plugin-manager fixture skipped.
* Static build containing both plugin names: mode, discovery, and default tests passed 7/7.
* Optimized dynamic build of both plugin entry points passed.
* Both legacy path-mode round-trip smoke tests passed.
* Eight 64 MiB direct-I/O consistency cases passed across base/PR, GDS/GDS_MT, and READ/WRITE. GDS used `batch_limit=1` to force four sub-batches.
* Independent GB200 validation on 8 NVMe and 4 GPUs covered 512 KiB, 1 MiB, and 16 MiB sequential direct-VRAM I/O at depth 16. The corrected matrices completed 8/8 consistency cases and 704/704 performance processes. PR1 tracked its exact base at every size, with one follow-up signal: 512 KiB GDS_MT WRITE was `-0.461%` across three pairs and needs more rounds before calling it a regression. An earlier same-file six-pair 32/64 MiB control also found no reproducible slowdown.
* `git diff --check` passed.

One baseline GDS READ sweep hit the existing NVFS `nvfs_bio:209` assertion; its immediate baseline retry passed, and no PR run failed.

#### GB200 performance by I/O size

**Test system:** dual-socket NVIDIA Grace (144 Neoverse-V2 cores), 4 x GB200 GPUs, and 8 x Samsung 3.5 TB NVMe drives using ext4, with two drives mapped to each GPU. The software stack was Linux 6.14 aarch64 with 64 KiB pages, NVIDIA driver 580.105.08, CUDA 13.0, cuFile 1.19.0.76, and `nvidia_fs` 2.26.

Sequential direct-VRAM I/O across 8 NVMe drives and 4 GB200 GPUs, with 16 entries/workers and a 16 GiB working set per drive. Throughput is aggregate GiB/s. The 1 MiB point used five paired base/PR1 rounds; 512 KiB and 16 MiB used three. GDSIO is a matched cross-run reference; baseline and PR1 were measured in paired, alternating-order rounds.

Only x6 and x0 are shown because they match the two NIXL execution models: GDS uses the cuFile batch API and is compared with x6 `GPU_BATCH`; GDS_MT issues synchronous `cuFileRead`/`cuFileWrite` operations from multiple workers and is compared with x0 `GPUD`. The other GDSIO modes use CPU/page-cache staging, asynchronous stream APIs, or vectored APIs that these NIXL backends do not use, so they are not like-for-like references.

| I/O size | NIXL path / GDSIO mode | READ GDSIO (GiB/s) | READ baseline (GiB/s) | READ this PR (GiB/s) | WRITE GDSIO (GiB/s) | WRITE baseline (GiB/s) | WRITE this PR (GiB/s) |
| -- | -- | -- | -- | -- | -- | -- | -- |
| 512 KiB | GDS batch / x6 | 49.549 | 35.749 | 35.722 | 30.672 | 29.802 | 29.873 |
| 512 KiB | GDS_MT / x0 | 49.049 | 39.896 | 39.818 | 30.664 | 30.044 | 29.906 |
| 1 MiB | GDS batch / x6 | 49.624 | 43.379 | 43.395 | 30.536 | 30.405 | 30.387 |
| 1 MiB | GDS_MT / x0 | 49.368 | 44.125 | 44.094 | 30.806 | 30.282 | 30.275 |
| 16 MiB | GDS batch / x6 | 49.511 | 47.680 | 47.596 | 30.593 | 30.730 | 30.723 |
| 16 MiB | GDS_MT / x0 | 49.404 | 48.931 | 48.978 | 30.776 | 30.769 | 30.732 |

The read gap versus GDSIO shrinks as I/O size grows, while writes remain close to the matched GDSIO reference. PR1 follows the same size curve as baseline. All 8 consistency cases and 704 corrected performance processes passed. Full methodology, base values, pair ranges, CPU results, and retained-artifact details are in the [independent GB200 report](<https://github.com/maheshrbapatu/nixl/pull/1#issuecomment-4861853943>).

## Summary by CodeRabbit

* **New Features**
  * Added batch/MT selection to **GDS** through `mode=batch|mt`, with batch remaining the default.
  * Preserved **GDS_MT** as a multi-threaded compatibility backend and exposed mode-specific defaults, including the MT thread count derived from available CPU concurrency.
* **Bug Fixes**
  * Enforced mutual exclusivity: **GDS** and **GDS_MT** cannot both be created in the same agent.
  * Improved correctness for file-backed transfers, including repeated `FILE_SEG` registrations against the same underlying file.
* **Documentation**
  * Expanded the CUDA GDS README with configuration tables for both backends.
* **Tests**
  * Added/expanded hardware-gated end-to-end and validation tests for round-trips and mutual exclusivity.

## Summary by CodeRabbit

* **New Features**
  * Added unified CUDA GDS support with selectable batch and multithreaded transfer modes.
  * Added configurable batch limits, request sizing, and worker-thread counts.
  * Improved file and memory registration handling for reliable transfers.
  * Added backend validation for supported transfer combinations and modes.

* **Documentation**
  * Expanded GDS documentation with mode selection, configuration, and usage guidance.

* **Tests**
  * Added end-to-end coverage for plugin creation, transfers, configuration, validation, and cleanup.


sbates130272:
@maheshrbapatu I am tracking this because if it gets merged it will impact the AIS_MT and hip_ais plugins that enable DGS/AIS for AMD GPUs that I am working on a PR for. See issue #1781.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -14,395 +14,217 @@
...
-            custom_params, "batch_pool_size", DEFAULT_BATCH_POOL_SIZE);
-        batch_limit =
-            nixl::getBackendParamDefaulted(custom_params, "batch_limit", DEFAULT_BATCH_LIMIT);
-        max_request_size = nixl::getBackendParamDefaulted(
-            custom_params, "max_request_size", DEFAULT_MAX_REQUEST_SIZE);
+        driver_ = std::make_unique<gdsDriverHandle>();
     }
     catch (const std::exception &e) {
         NIXL_ERROR << e.what();
         this->initErr = true;
```
ofer, line 72: Please use traling _ as per coding style

sbates130272:
@maheshrbapatu I have some patches coming to enable AMD AIS_MT plugin. I would prefer to do that after this work is merged if it is going to get merged. So I do not cause you more refactor work. Can you comment on your ETA for addressing @ofer review comments? That way I can determine if I need to move forward based on main or your PR. Thanks!

maheshrbapatu:
> @maheshrbapatu I have some patches coming to enable AMD AIS_MT plugin. I would prefer to do that after this work is merged if it is going to get merged. So I do not cause you more refactor work. Can you comment on your ETA for addressing @ofer review comments? That way I can determine if I need to move forward based on main or your PR. Thanks!

Hi @sbates130272  appreciate you flagging this!

I'm currently running performance benchmarks to validate there are no regressions before addressing @ofer's review comments. Hardware availability is the main constraint, so realistically I'm looking at **~2 weeks** before this is ready to merge.

If you're able to share your target timeline, I'd love to sync up and make sure we're not stepping on each other's toes.

sbates130272:
@maheshrbapatu I am probably at about the same timeline. Let me work up my PR on top of yours and then my testing helps you. Stay tuned.

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
```diff
@@ -326,6 +326,10 @@ int main(int argc, char *argv[])
         if (int rc = runPathModeSmoke(); rc != 0) {
             return rc;
         }
+        // Meson registers the no-argument invocation as the path-mode smoke test.
```
aranadive, line 409: I think we can also keep only one unit test file? So the GDS_MT file can be merged into this.

maheshrbapatu:
### H100 performance results

| I/O size | NIXL path | READ baseline GB/s | READ baseline CPU | READ PR GB/s | READ PR CPU | WRITE baseline GB/s | WRITE baseline CPU | WRITE PR GB/s | WRITE PR CPU |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 MiB | GDS batch | 10.631076 | 121.77% | 10.639171 | 121.83% | 6.049401 | 111.40% | 6.053428 | 111.23% |
| 1 MiB | GDS_MT (32) | 11.702507 | 191.59% | 11.696630 | 192.98% | 5.929846 | 150.28% | 5.972819 | 150.29% |
| 4 MiB | GDS batch | 11.349297 | 83.96% | 11.360970 | 83.90% | 6.036027 | 87.57% | 6.034320 | 87.10% |
| 4 MiB | GDS_MT (32) | 11.874692 | 140.44% | 11.879912 | 140.42% | 5.797791 | 121.43% | 5.930050 | 121.19% |

test/unit/plugins/cuda_gds/nixl_gds_test.cpp
maheshrbapatu: Addressed in bba30e2c. Both nixl_gds_test and nixl_gds_mt_test now build from the single cuda_gds/nixl_gds_test.cpp source, while preserving both executable names. The unified source covers default GDS batch, GDS mode=mt, and the standalone GDS_MT compatibility entry. The duplicate 799-line GDS_MT source was removed. Dynamic and static focused validation both passed: 3/3 standalone path-mode tests and 18 GDS GTests, with 1 unrelated lifecycle skip.

src/plugins/cuda_gds/gds_backend.cpp
maheshrbapatu: Thanks. `initErr` is the existing protected member inherited from `nixlBackendEngine` (`src/api/cpp/backend/backend_engine.h:43`); this PR only assigns that inherited member. Renaming it here would require a repository-wide source-level rename across the base engine and all backend users, which is outside this consolidation. I am keeping the inherited spelling in this PR; it can be handled separately as a consistent codebase-wide style cleanup.

src/plugins/cuda_gds/gds_batch_engine.cpp
```diff
@@ -0,0 +1,473 @@
...
+    auto keep = batch_list.begin();
+
+    for (nixlGdsIOBatch *batch : batch_list) {
+        if (batch == nullptr) {
+            continue;
+        }
+        if (batch->cancelBatch() == NIXL_SUCCESS) {
+            // TODO: Establish and test cancel-to-resubmit semantics for every
+            // cuFile I/O path before immediately reusing a batch handle.
+            returnBatchToPool(batch);
```
vvenkates27, line 371: I'd skip putting this back in the pool after cancel. If I'm reading it right, abort-while-IN_PROG now does checkXfer → releaseReqH → here, so the next transfer can submit on the same CUfileBatchHandle_t. Today we don't recycle on abort — we delete the req and ~nixlGdsIOBatch bails if the batch is still in flight. Not pretty, but we also don't resubmit on a cancelled handle.

cancel + returnBatchToPool on post/check errors is already there, so I'm not trying to unwind that. Using it from releaseReqH is the new bit, and the TODO above sounds like cancel → reset → submit isn't locked down yet.  Until we have a test for that, destroying and doing a fresh cuFileBatchIOSetUp (or just leaving the slot unused) seems safer. Same idea for the checkXfer error path that calls this helper.

src/plugins/cuda_gds/gds_batch_engine.cpp
```diff
@@ -0,0 +1,473 @@
...
+        return gds_handle->overall_status;
+    }
+
+    return gds_handle->batch_io_list.empty() ? NIXL_SUCCESS : NIXL_IN_PROG;
+}
+
+nixl_status_t
+nixlGdsBatchEngine::releaseReqH(nixlBackendReqH *handle) const {
+    auto *gds_handle = static_cast<nixlGdsBatchReqH *>(handle);
+    if (cancelAndReclaimBatches(gds_handle->batch_io_list) != NIXL_SUCCESS) {
```
vvenkates27: One more thing on cancel failure: if cancelBatch() fails we keep the pointer on the handle, then still delete the req. Those batches live in batch_storage_; batch_io_list is only a checkout, so I don't think they ever get back to the pool. Later posts could hit "pool exhausted". Today checkXfer always returnBatchToPool after a failed batch, even if cancel fails, so we don't strand slots that way.

I'd rather not delete the req while anything is still checked out — destroy/replace those objects first, or fail release and leave the handle around.

vvenkates27 (commented):
Thanks for this consolidation work— shared engine and keeping the GDS_MT name is a good call until we are able to modify all integrations to use GDS with appropriate mode. I think from my experience the default for GDS should be mt as most people prefer that. But given we have mt separately we should do that when mt is removed.

releaseXferReq() isn't in this diff. What's new is GDS releaseReqH cancelling and pooling batch handles on abort-while-IN_PROG; today that path deletes/leaks instead of reuse. Cancel-on-post/check-error reuse I think we already had.

The one I'd like eyes on is recycling cancelled cuFile batches from releaseReqH — either we test cancel → submit on the same handle, or we keep the old "don't reuse on abort" behavior for now so abort stays compatible with current GDS.

src/plugins/cuda_gds/gds_batch_engine.cpp
maheshrbapatu: Good catch — you're right that canceling and returning these batches to the pool from releaseReqH is new behavior in this refactor. I'll keep this PR behavior-compatible with the pre-consolidation path: release will no longer cancel or return in-flight batches to the pool. The existing post/check error handling will remain unchanged. We'll handle safe destroy-and-replacement, together with the relevant cancel/reuse tests, as a separate follow-up.

src/plugins/cuda_gds/gds_batch_engine.cpp
maheshrbapatu: Agreed — leaving slots unavailable is not a good long-term solution, and repeated aborts can reduce usable pool capacity. For this consolidation, I'll preserve the old release behavior rather than introduce unverified reuse of canceled handles. The cancel-failure case in releaseReqH goes away because release will no longer call cancelBatch. I'll address replenishing retired slots through safe destroy-and-replacement in a follow-up.

vvenkates27 (dismissed):
Overall this is ready once CI passes. 
@maheshrbapatu  need a follow up PR for documentation after this merges.

Docs: Fern still treats GDS and GDS_MT as two separate backends and never mentions mode. Please match src/plugins/cuda_gds/README.md in fern/docs/pages/user-guide/backends/gds.md and gds-mt.md.

GDS defaults to batch; mode=mt selects the MT engine. thread_count applies only with mode=mt; batch knobs apply only with mode=batch.
GDS_MT stays a compatibility name for the MT engine (thread_count only). createBackend("GDS") / createBackend("GDS_MT") defaults are unchanged.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -14,395 +14,219 @@
...
+    std::shared_ptr<gdsFileHandle> handle;
+    uint64_t dev_id;
+
+    fileSegData(std::shared_ptr<gdsFileHandle> h, uint64_t id) : handle(std::move(h)), dev_id(id) {}
+};
+
+struct memSegData {
+    gdsMemBuf buf;
+
+    memSegData(void *addr, size_t size, int flags) : buf(addr, size, flags) {}
```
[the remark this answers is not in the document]
aranadive: @maheshrbapatu ?

test/gtest/unit/gds/gds.cpp
```diff
@@ -0,0 +1,519 @@
+/*
```
aranadive, line 1: This needs to be in test/gtest/unit/gds/

src/plugins/cuda_gds/README.md
```diff
@@ -18,7 +18,43 @@ limitations under the License.
...
+simultaneously within a single agent.
+
+## Backend selection
+
+| Backend | Parameters | Transfer engine |
+| --- | --- | --- |
+| `GDS` | none | Batch |
+| `GDS` | `mode=batch` | Batch |
+| `GDS` | `mode=mt` | Multi-threaded |
+| `GDS_MT` | none | Multi-threaded |
```
aranadive, line 42: Can you add a note here that in a future update GDS_MT option will be removed

src/plugins/cuda_gds/meson.build
```diff
@@ -26,34 +30,99 @@ else
...
+    include_directories: gds_incs)
+
+# "GDS" backend name entry point (batch by default, MT when mode=mt).
+if enabled_plugins.get('GDS')
+    gds_sources = ['gds_plugin.cpp']
+    gds_deps = [gds_engine_dep]
+    if 'GDS' in static_plugins
+        gds_backend_lib = static_library('GDS',
+            gds_sources,
+            dependencies: gds_deps,
```
aranadive: Just gds_engine_dep?

src/plugins/cuda_gds/meson.build
```diff
@@ -26,34 +30,99 @@ else
...
+    gds_backend_interface = declare_dependency(
+        link_with: gds_backend_lib,
+        dependencies: [taskflow_proj, thread_dep])
 endif
 
-gds_backend_interface = declare_dependency(link_with: gds_backend_lib)
+# "GDS_MT" backend name entry point (per-request TaskFlow strategy).
+if enabled_plugins.get('GDS_MT')
+    gds_mt_sources = ['gds_mt_plugin.cpp']
+    gds_mt_deps = [gds_engine_dep]
```
aranadive: This means that if GDS and GDS_MT are enabled, GDS_MT plugin will also have the GDS engine in it.

src/plugins/cuda_gds/gds_backend.cpp
ColinNV: FYI, completely removing `initErr` is on my "todo" list, it is no longer needed.

src/plugins/cuda_gds/gds_backend.cpp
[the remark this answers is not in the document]
maheshrbapatu: Fixed in b16bbed9. Both constructors now break before the initializer colon, and the changed lines pass clang-format 19.

test/gtest/unit/gds/gds.cpp
maheshrbapatu: Moved unchanged in b16bbed9 to test/gtest/unit/gds/gds.cpp. It remains in the existing gtest target; GPU-only CI selection can be handled in the follow-up as discussed.

src/plugins/cuda_gds/README.md
maheshrbapatu: Added in b16bbed9. The README now says GDS_MT is temporary, will be removed in a future update, and new integrations should use GDS with mode=mt. I updated both Fern backend pages as well.

src/plugins/cuda_gds/meson.build
maheshrbapatu: Yes. Simplified both plugin targets to use gds_engine_dep directly in b16bbed9.

src/plugins/cuda_gds/meson.build
maheshrbapatu: Fixed in b16bbed9. gds_engine_lib now contains only the common and MT sources, while gds_batch_engine is compiled only into GDS. I built both shared plugins and verified that GDS_MT contains no nixlGdsBatchEngine or nixlGdsIOBatch symbols.

test/gtest/meson.build
```diff
@@ -108,6 +108,7 @@ gtest_sources = [
     'metadata_worker.cpp',
     'common.cpp',
     'query_mem.cpp',
+    'unit/gds/gds.cpp',
```
aranadive: This file inclusion should go into test/gtest/unit/meson.build. See the azure_blob test inclusion there.

src/plugins/cuda_gds/gds_mt_engine.cpp
```diff
@@ -0,0 +1,152 @@
...
+        return;
+    }
+
+    if ((size_t)nbytes != req->size) {
+        NIXL_ERROR << "GDS_MT: error: short " << ((req->op == CUFILE_READ) ? "read: " : "write: ")
+                   << nbytes << " out of " << req->size << " bytes - address=" << req->addr;
+        overall_status->store(NIXL_ERR_BACKEND);
+        return;
+    }
+}
```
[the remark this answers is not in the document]
aranadive: @maheshrbapatu?

test/gtest/meson.build
maheshrbapatu: Fixed in 03222adc. The GDS suite is now declared in `test/gtest/unit/gds/meson.build` and included from `test/gtest/unit/meson.build`, following the `azure_blob` pattern. It is no longer part of the top-level `gtest_sources`. I rebuilt the relocated unit target and all 142 unit tests passed, including all 17 GDS/GDS_MT tests.

src/plugins/cuda_gds/gds_mt_engine.cpp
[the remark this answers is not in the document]
maheshrbapatu: Fixed in 03222adc. Both `cuFileRead` and `cuFileWrite` error paths now use the thread-safe `nixl_strerror(errno)` helper, and the unused `<cstring>` include was removed. Both plugins rebuilt successfully and the GDS/GDS_MT suite passed 17/17.

src/plugins/cuda_gds/gds_backend.cpp
```diff
@@ -14,399 +14,226 @@
...
 
-        if (status < 0) {
-            batch->cancelBatch();
-        }
-        returnBatchToPool(batch);
+        reqs.push_back(GdsXferReq{base_addr,
+                                  mem_desc.len,
+                                  (size_t)file_desc.addr,
+                                  file_data->handle->cu_fhandle,
+                                  (operation == NIXL_READ) ? CUFILE_READ : CUFILE_WRITE});
```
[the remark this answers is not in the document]
maheshrbapatu: Verified against the current call paths; no backend change is needed here. `nixlAgent::createXferReq` rejects mismatched descriptor lengths before calling backend `prepXfer`, and `nixlAgent::makeXferReq` performs the same check on the selected strides. The GDS backend therefore cannot receive a mismatched pair through the public agent API.

src/plugins/cuda_gds/gds_mt_engine.cpp
```diff
@@ -0,0 +1,153 @@
...
+nixlGdsMtEngine::postXfer(const nixl_xfer_op_t &operation,
+                          const nixl_meta_dlist_t &local,
+                          const nixl_meta_dlist_t &remote,
+                          const std::string &remote_agent,
+                          nixlBackendReqH *&handle,
+                          const nixl_opt_b_args_t *opt_args) const {
+    auto *gds_handle = static_cast<nixlGdsMtReqH *>(handle);
+
+    gds_handle->overall_status.store(NIXL_SUCCESS);
+    gds_handle->running_transfer = executor_->run(gds_handle->taskflow);
```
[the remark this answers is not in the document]
maheshrbapatu: Verified against the current repost path; no engine change is needed. `nixlAgent::postXferReq` checks an in-progress request with `checkXfer` and returns `NIXL_ERR_REPOST_ACTIVE` while it is still running. The backend `postXfer` is reached only after the prior future is ready and consumed, so `running_transfer` is not overwritten while active.

src/plugins/cuda_gds/gds_mt_engine.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+nixl_status_t
+nixlGdsMtEngine::checkXfer(nixlBackendReqH *handle) const {
+    auto *gds_handle = static_cast<nixlGdsMtReqH *>(handle);
+    if (gds_handle->running_transfer.wait_for(std::chrono::seconds(0)) !=
+        std::future_status::ready) {
+        return NIXL_IN_PROG;
+    }
+    gds_handle->running_transfer.get();
```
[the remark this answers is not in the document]
maheshrbapatu: Verified against the public polling path; no engine change is needed. `nixlAgent::getXferStatus` calls backend `checkXfer` only while the outer request status is `NIXL_IN_PROG`. Once `checkXfer` consumes the ready future, that result is cached on the outer request and later status calls return it without polling the future again. This also preserves the pre-consolidation GDS_MT behavior.

src/plugins/meson.build
```diff
@@ -41,33 +41,23 @@ if enabled_plugins.get('LIBFABRIC')
...
         if disable_gds_backend
-            error('GDS plugin requested but GDS backend is disabled')
+            error('GDS/GDS_MT plugin requested but GDS backend is disabled')
         else
-            error('GDS plugin requested but CUDA dependency not found')
+            error('GDS/GDS_MT plugin requested but CUDA dependency not found')
         endif
     elif not disable_gds_backend and cuda_dep.found()
         subdir('cuda_gds')
     endif
```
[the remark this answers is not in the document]
maheshrbapatu: This is intentionally unchanged from both pre-consolidation GDS implementations: their plugin subdirectories also used `subdir_done()` when `gds_path` was empty, including for an explicit plugin selection. The default remains `/usr/local/cuda/`. Making an empty override fatal would be a separate build-contract change, so this PR keeps the existing behavior.

src/plugins/meson.build
```diff
@@ -41,30 +41,21 @@ if enabled_plugins.get('LIBFABRIC')
...
+            error('GDS/GDS_MT plugin requested but GDS backend is disabled')
+        elif not taskflow_proj.found()
+            error('GDS/GDS_MT plugin requested but Taskflow dependency not found')
         else
-            error('GDS_MT plugin requested but Taskflow dependency not found')
+            error('GDS/GDS_MT plugin requested but CUDA dependency not found')
         endif
     elif not disable_gds_backend and cuda_dep.found() and taskflow_proj.found()
-        subdir('gds_mt')
+        subdir('cuda_gds')
```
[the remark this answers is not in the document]
maheshrbapatu: The Taskflow dependency is intentional for the consolidated contract: the `GDS` backend now exposes both `mode=batch` and `mode=mt`, with batch remaining the runtime default. Building `GDS` without the MT engine would make its documented `mode=mt` unavailable. The compatibility `GDS_MT` target remains MT-only and, as verified from the built shared object, contains no batch-engine symbols.

src/plugins/cuda_gds/gds_backend.cpp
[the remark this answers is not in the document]
aranadive: Core already enforces this — nixlAgent::createXferReq (nixl_agent.cpp:944-948) and nixlAgent::makeXferReq (:859-863) both reject per-descriptor length mismatches before any backend sees the request. mem_desc.len == file_desc.len is an invariant at this point, so reading one and ignoring the other is correct. Not adding a third check.

vvenkates27 (approved):

aranadive (approved):

sbates130272:
@maheshrbapatu dont slow down any merged but I'll try to confirm this works under the PR from @riley-dixon for AMD GPUs. We assumed as much so it should work fine. Thanks for doing this!

PR #1867, opened 2026-07-01
tracing: auto-enable NVTX backend when running under Nsight Systems

e-eygin:
## What?

When `NIXL_TRACE_BACKENDS` is **unset**, NIXL now auto-enables the NVTX tracing
backend if the process is running under Nsight Systems, so a profiled run
produces an NVTX timeline without setting the variable by hand. Outside an nsys
run the agent stays inert (no plugin load, no per-agent NVTX domain).

Detection keys off `NVTX_INJECTION64_PATH`, which nsys injects into the
environment of the process it profiles — its presence means "running under nsys",
not merely that nsys is installed.

Selection precedence (`resolveTraceBackends`):
- explicit non-empty `NIXL_TRACE_BACKENDS` → honored, **plus** `nvtx` added under
  nsys (deduplicated) — auto-enable is additive and never suppresses a requested
  backend;
- set-but-empty (`NIXL_TRACE_BACKENDS=`) → tracing off (hard override, beats
  auto-enable);
- unset → `nvtx` under nsys, else off.

Resolves NIX-1576.

## Why?

Running a NIXL workload under `nsys profile` should "just work" and yield the
NVTX timeline without requiring users to also remember `NIXL_TRACE_BACKENDS=nvtx`.
Conversely, when not profiling we must pay zero cost, and users must retain an
explicit way to force tracing off.

## How?

- New `src/core/tracing/trace_selection.h` holds the agent-wiring selection
  policy — `runningUnderNsys()` and the pure `resolveTraceBackends()` — so the
  nsys/NVTX-specific signal never leaks into the backend-agnostic tracer facade
  (`trace.h` / `tracer.cpp`). `makeAgentTracer()` reads `NIXL_TRACE_BACKENDS`
  (`std::optional`, distinguishing unset from set-but-empty) and feeds the
  existing `makeTracer()` path.
- No public ABI change: selection remains env-only (no `nixlAgentConfig` field).

### Testing

- Facade unit tests (deterministic, no plugin/profiler): explicit-list honored,
  nsys adds NVTX to an explicit list (union + dedup), set-but-empty = off,
  unset+nsys = NVTX, and `runningUnderNsys` detection.
- Real-agent test `TestTransferTracingNsysAuto`: `NIXL_TRACE_BACKENDS` unset +
  injection var set → NVTX loads via auto-enable and a UCX transfer succeeds.
- Verified an isolated `nsys` capture of only the auto-enable test (env
  confirmed unset) produces real per-agent `nixl::*` ranges — proving the
  backend is genuinely loaded via auto-enable alone.
- All via CTest; `docs/tracing.md` documents the behavior with a precedence table.

## Summary by CodeRabbit

* **New Features**
  * Tracing can now auto-enable the NVTX backend when running under Nsight Systems, without requiring an explicit backend setting.

* **Bug Fixes**
  * An explicit “turn tracing off” setting remains off, even in Nsight Systems environments.
  * Backend selection is more consistent: whitespace is ignored and duplicates are removed.

* **Documentation**
  * Updated tracing documentation and profiling examples to explain the Nsight Systems auto-enable behavior and overrides.

* **Tests**
  * Added coverage for Nsight Systems NVTX auto-enable and backend resolution edge cases.


e-eygin:
@ovidiusm @ColinNV please review

src/core/nixl_agent.cpp
```diff
@@ -152,6 +151,43 @@ makeAgentTracer(const std::string &name) {
...
+    }
+
+    // Running under nsys auto-enables NVTX *in addition to* any explicitly
+    // requested backends (so specifying e.g. "chakra" still yields the NVTX
+    // timeline). Dedup so an explicit "nvtx" is not loaded twice.
+    if (under_nsys && !explicit_off &&
+        std::find(backends.begin(), backends.end(), "nvtx") == backends.end()) {
+        backends.emplace_back("nvtx");
+    }
+    return backends;
```
ColinNV: Make `backends` a `std::set` to deduplicate everything and simplify the NVTX auto-enable, then `return { backends.begin(), backends.end() };`?

src/core/tracing/trace_selection.h
```diff
@@ -0,0 +1,64 @@
+/*
```
ColinNV: It looks like this header only exists to call the functions from the test? In that case I'd get rid of it and just declare them in the test, but that's just my personal style -- don't add to the code tree "only" for testing unless absolutely necessary.

src/core/nixl_agent.cpp
e-eygin: Done in `afffb5e3` — `std::set `+ `return {begin, end}`.

src/core/tracing/trace_selection.h
e-eygin: Agreed, removed in `afffb5e3`.

src/core/nixl_agent.cpp
```diff
@@ -15,11 +15,15 @@
  * limitations under the License.
  */
 
+#include <algorithm>
```
ColinNV: Not needed? The previous call to `std::find()` is gone.

src/core/nixl_agent.cpp
```diff
@@ -106,6 +110,46 @@ nixlDlistH::nixlDlistH(const std::string &remote_agent, descs_t &&descs)
 
 /*** nixlAgentData constructor/destructor, as part of nixlAgent's ***/
 
+namespace nixl::trace {
+
+// True when the process runs under Nsight Systems: nsys injects
+// NVTX_INJECTION64_PATH into the profiled process's environment (its presence
+// means "running under nsys", not merely that nsys is installed).
+bool
```
ColinNV: `[[nodiscard]]`, also below.

src/core/nixl_agent.cpp
e-eygin: Removed in `92bda3fb`.

src/core/nixl_agent.cpp
e-eygin: Added on both in `92bda3fb`.

ColinNV (approved):

PR #1874, opened 2026-07-01
ci: skip ucx_backend_multi under TSan and lower sanitizer stage timeout to 40m

NirWolfer:
## What?
ucx_backend_multi rarely hangs to the stage timeout under ThreadSanitizer. It is a multi-threaded UCX connect/disconnect stress test whose two threads own separate engines and coordinate a one-sided connect/disconnect via busy-wait spin loops with asymmetric progress pumping; TSan's scheduling perturbation occasionally leaves a UCX wireup step un-progressed and the peers deadlock. TSan also cannot see UCX's own synchronization (all of libucp/uct/ucs is already suppressed in tsan.supp), so the test adds little TSan value. Skip it under TSan only, keyed on SAN_LABEL; it still runs under ASan/UBSan, where it is reliable.

Also lower TEST_TIMEOUT from 90 to 40 minutes: sanitizer runs average 15-30 min, so 90 let a hung stage burn far longer than needed before being killed. 40m keeps comfortable headroom over the normal runtime while failing hangs faster.

## Why?
test-sanitizers pipeline on rare occasions hangs until timeout, this restores stability

## How?
architecture, approach, etc._

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced sanitizer test job timeouts to shorten waiting time in CI.
  * Updated sanitizer execution so one backend stage is skipped under ThreadSanitizer, preventing failures in that configuration.
  * Improved reliability of sanitizer-related test runs without changing normal build behavior.

### DIFF
--- a/.ci/jenkins/lib/test-sanitizer-matrix.yaml
+++ b/.ci/jenkins/lib/test-sanitizer-matrix.yaml
@@ -39,7 +39,7 @@ kubernetes:
 env:
   NIXL_CI_NON_GPU: 1
   NIXL_INSTALL_DIR: /opt/nixl
-  TEST_TIMEOUT: 90
+  TEST_TIMEOUT: 40
   UCX_TLS: "^shm"
   STORAGE_DRIVER: 'overlay'
   CI_IMAGE_TAG: "20260625-1"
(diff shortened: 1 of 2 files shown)


ColinNV (approved):
LGTM

dpressle (approved):

PR #1851, opened 2026-06-30
telemetry: expose error counts in exporters

e-eygin:
## What?

Expose existing `AGENT_ERR_*` telemetry events from both telemetry exporters as a bounded labeled counter:

`agent_errors_total{status="..."}`

This adds error-count export support to the native Prometheus exporter and the DOCA telemetry exporter. Core telemetry event types, event names, and the shared-memory buffer format are unchanged.

## Why?

The agent already emits per-error telemetry events via `updateErrorCount`, but the active exporter paths did not expose them. This makes error counts visible through the exporter metrics surface without adding new core events or changing `TELEMETRY_VERSION`.

## How?

Prometheus registers one `agent_errors_total` counter family with `hostname`, `agent_name`, and `status` labels, then maps each existing `AGENT_ERR_*` event to the corresponding bounded `status` value.

DOCA adds a second label set for `agent_name` + `status` and routes `AGENT_ERR_*` events through a dedicated `agent_errors_total` counter path.

Tests cover both exporters:
- Prometheus scrape test verifies `invalid_param` and `backend` accumulate under separate `status` labels.
- DOCA scrape test verifies the same labeled counter behavior after flush.

Validation:
- `ninja -C build install`
- `build/test/gtest/gtest --gtest_filter='telemetryTest.*:prometheusTelemetryTest.*'`
- `meson test -C build doca_test doca_nixl_test`
- diff-scoped `clang-format`

## Summary by CodeRabbit

* **Documentation**
  * Updated telemetry docs and “Metrics & Events” guidance to describe agent error telemetry as a single labeled counter: `agent_errors_total{status="..."}`, with `status` constrained to the predefined `AGENT_ERR_*` set.
* **New Features**
  * Prometheus and DOCA exporters now aggregate agent error events into `agent_errors_total` using the bounded `status` label (and include `hostname`/`agent_name`), removing legacy per-error-type `agent_err_*` counter families.
* **Tests**
  * Added/updated end-to-end and scrape-parsing tests to validate per-`status` totals and confirm legacy `agent_err_*` series are not present.

PR #1878, opened 2026-07-02
telemetry: key Prometheus exporter maps by event-type enum

e-eygin:
## What?

Re-key the Prometheus exporter's `counters_` / `gauges_` maps on the
`nixl_telemetry_event_type_t` enum instead of `std::string`. `exportEvent()`
now looks up metrics directly by `event.eventType_` with no per-event string
construction on the hot path; series names are still derived from
`telemetryEventTypeStr()` at registration time only. Removes the
`// TODO(C++20)` in `exportEvent()`.

## Why?

Addresses the review question on #1851 ("why construct a string per event?",
`prometheus_exporter.cpp`). `exportEvent()` allocated a `std::string` from the
`string_view` returned by `telemetryEventTypeStr()` on every event just to
index the maps.

The originally-ticketed fix (transparent `string_view` hash) was rejected: it
relies on heterogeneous `unordered_map` lookup, which is C++20-only
(`#if __cplusplus > 201703L`), and the project still builds under C++17 (the
`cpp_std` meson option is overridable and ASan/CI configs use `c++17`). Keying
by the enum is allocation-free for every event, uses a cheaper integer hash,
and compiles cleanly under both the C++17 and C++20 build configs.

Refs NIX-1588.

## How?

- `counters_` / `gauges_` become
  `std::unordered_map<nixl_telemetry_event_type_t, ...>` (`std::hash` supports
  enums in C++17).
- `registerCounter` takes the enum and derives the `*_total` family name via
  `telemetryEventTypeStr(event_type)`; `registerGauge` / `registerErrorCounters`
  key by the enum.
- Behavior-preserving: identical series names, labels, and values. Verified by
  the existing `prometheusTelemetryTest` suite (4 tests) under both C++17 and
  C++20 builds; no new tests needed since observable output is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry metric registration and export reliability by using consistent event-type handling.
  * Reduced the chance of missing or mismatched Prometheus counters and gauges during metric export.
  * Kept exported metric names and behavior unchanged while making metric lookup more robust.

### DIFF
--- a/src/plugins/telemetry/prometheus/prometheus_exporter.cpp
+++ b/src/plugins/telemetry/prometheus/prometheus_exporter.cpp
@@ -117,31 +117,39 @@ nixlTelemetryPrometheusExporter::~nixlTelemetryPrometheusExporter() {
 // Events are defined in the telemetry.cpp file
 void
 nixlTelemetryPrometheusExporter::initializeMetrics() {
-    registerCounter("agent_tx_bytes", "Number of bytes sent by the agent");
-    registerCounter("agent_rx_bytes", "Number of bytes received by the agent");
-    registerCounter("agent_tx_requests_num", "Number of requests sent by the agent");
-    registerCounter("agent_rx_requests_num", "Number of requests received by the agent");
-    registerCounter("agent_memory_registered", "Cumulative memory registered");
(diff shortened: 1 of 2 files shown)


e-eygin:

ColinNV (approved):

PR #1829, opened 2026-06-24
bindings: free the dlist handle wrapper in release_xfer_dlist_handle

EylonKrause:
## What?

`nixl_capi_prep_xfer_dlist` (`src/bindings/rust/wrapper.cpp`) allocates the handle
wrapper with `new nixl_capi_xfer_dlist_handle_s`, but
`nixl_capi_release_xfer_dlist_handle` only released the **inner** handle and never
freed the wrapper:

```diff
     try {
         nixl_status_t ret = agent->inner->releasedDlistH(dlist_handle->handle);
+        delete dlist_handle;
         return ret == NIXL_SUCCESS ? NIXL_CAPI_SUCCESS : NIXL_CAPI_ERROR_BACKEND;
     }
```

## Why?

`releasedDlistH()` deletes the inner `nixlDlistH`; the `nixl_capi_xfer_dlist_handle_s`
wrapper that `prep` `new`'d is a **separate** heap allocation that was never freed, so
each prep/release cycle leaks one wrapper. Rust's `XferDlistHandle::drop` calls release
exactly once and never retries, so an unconditional `delete` here is the sole, correct
teardown (no double-free).

### Reproduction

Extracted LeakSanitizer reproducer of the `new wrapper` / release-frees-only-inner
shape (`-fsanitize=address,leak`):

```
buggy:  SUMMARY: AddressSanitizer: 8 byte(s) leaked in 1 allocation(s)
fixed:  no leak
```

## How (verification)

- Confirmed the before/after with the reproducer above.
- Compiled the modified `wrapper.cpp` **in-tree** with `-Dsanitizer=address,undefined`
  (exit 0).

## Related Issues

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a handle cleanup issue so released transfer handles are now properly freed after use, helping prevent memory leaks.

PR #1879, opened 2026-07-02
bindings: free the dlist handle wrapper when prepXferDlist fails

EylonKrause:
## Summary

`nixl_capi_prep_xfer_dlist` (`src/bindings/rust/wrapper.cpp`) leaks the dlist handle wrapper on its error path.

The function allocates the wrapper directly into the out-param, then calls `prepXferDlist`:

```cpp
*dlist_handle = new nixl_capi_xfer_dlist_handle_s;
nixl_status_t ret = agent->inner->prepXferDlist(..., (*dlist_handle)->handle, ...);
return ret == NIXL_SUCCESS ? NIXL_CAPI_SUCCESS : NIXL_CAPI_ERROR_BACKEND;
```

When `prepXferDlist` returns a non-success status (e.g. backend/agent not found, unregistered descriptors) — or throws into the `catch (...)` — the function returns an error **without deleting the wrapper**. On that error path the Rust binding (`agent.rs`) returns `Err` and never constructs an `XferDlistHandle`, so `nixl_capi_release_xfer_dlist_handle` is never called, and the wrapper `delete` added in #1829 never runs. The wrapper is leaked once per failed prep.

This is complementary to #1829: that PR fixes the normal success/release lifecycle, but the release path cannot free a wrapper for a prep that never handed a handle back to Rust.

## Fix

Allocate into a local, delete it on the failure path, and publish it to `*dlist_handle` only on success — the delete-on-error idiom already used by the sibling `nixl_capi_make_xfer_req`.

## Testing

C-API binding change confined to the error path; the fix mirrors the established `make_xfer_req` delete-on-error idiom, so a correct prep still round-trips through `release`. No functional NIXL run is required to observe the leaked allocation on the error return.

### DIFF
--- a/src/bindings/rust/wrapper.cpp
+++ b/src/bindings/rust/wrapper.cpp
@@ -1408,12 +1408,17 @@ nixl_capi_prep_xfer_dlist(nixl_capi_agent_t agent,
     }
 
     try {
-        *dlist_handle = new nixl_capi_xfer_dlist_handle_s;
+        auto h = new nixl_capi_xfer_dlist_handle_s;
         nixl_status_t ret = agent->inner->prepXferDlist(std::string(agent_name),
                                                         *descs->dlist,
-                                                        (*dlist_handle)->handle,
+                                                        h->handle,
                                                         opt_args ? &opt_args->args : nullptr);
-        return ret == NIXL_SUCCESS ? NIXL_CAPI_SUCCESS : NIXL_CAPI_ERROR_BACKEND;
+        if (ret != NIXL_SUCCESS) {
+            delete h;
+            return NIXL_CAPI_ERROR_BACKEND;
+        }
+        *dlist_handle = h;
+        return NIXL_CAPI_SUCCESS;
     }
     catch (...) {
         return NIXL_CAPI_ERROR_BACKEND;


EylonKrause:

src/bindings/rust/wrapper.cpp
```diff
@@ -1408,12 +1408,17 @@ nixl_capi_prep_xfer_dlist(nixl_capi_agent_t agent,
     }
 
     try {
-        *dlist_handle = new nixl_capi_xfer_dlist_handle_s;
+        auto h = new nixl_capi_xfer_dlist_handle_s;
```
iyastreb, line 1411: why not using std::unique_ptr to automate memory management and make it exception safe?

EylonKrause:
Closing: `main` now handles this exactly the way @iyastreb suggested — `nixl_capi_prep_xfer_dlist` builds the handle with `std::make_unique<nixl_capi_xfer_dlist_handle_s>()` and only `release()`s it into `*dlist_handle` on success, so the failure path no longer leaks the wrapper and the exception path is covered too. Thanks for the pointer — this PR is superseded.

PR #1880, opened 2026-07-02
ucx: fix infinite loop dropping pending requests on dedicated-thread exit

EylonKrause:
## Summary

The dedicated UCX worker thread (`nixlUcxDedicatedThread::run()`, `src/plugins/ucx/ucx_backend.cpp`) enters an infinite loop on shutdown whenever it exits with requests still queued — hanging thread teardown.

The "dropping requests on exit" loop iterates `requests_` with a manual iterator but never advances or erases it:

```cpp
for (auto it = requests_.begin(); it != requests_.end();) {
    NIXL_INFO << "dropping " << *(*it);
    (*it)->complete(NIXL_ERR_BACKEND);
}
requests_.clear();
```

`complete()` only operates on the request's shared state (`sharedState_`); it does **not** remove the entry from `requests_`. This is confirmed by the sibling drain loop directly above, which erases explicitly after `complete()`:

```cpp
(*it)->complete(status);
it = requests_.erase(it);
```

So in the drop loop `it` never changes, `it != requests_.end()` stays true forever, and the thread spins indefinitely. It is reached whenever the thread exits with `!requests_.empty()` (any request still pending at shutdown).

## Fix

Use a range-for to complete every queued request exactly once, then `clear()` the vector.

## Testing

Localized shutdown-path fix. The non-terminating loop is evident from the sibling drain loop's explicit `erase` (proving `complete()` does not modify `requests_`); the range-for advances unconditionally and preserves the original ordering and the post-loop `clear()`.


ColinNV (approved):

EylonKrause:

iyastreb (approved):
Thanks for fixing it!

PR #1887, opened 2026-07-04
telemetry: add dropped-events counter for metrics loss detection (NIX-454)

e-eygin:
## Summary

Makes silent producer-side telemetry loss observable via a new cumulative
`agent_telemetry_events_dropped` counter on both exporters and in the raw BUFFER
stream. Prerequisite for NIX-1205.

- **Core**: new `AGENT_TELEMETRY_EVENTS_DROPPED` event type (`= 20`). `updateData`
  and the all-or-none `addXferStats` batch now count drops when the staging queue
  is full (`droppedEvents_`). On each flush, `exportDroppedEvents()` emits the
  delta since the last flush as a synthetic event — it bypasses the staging queue
  (so it can never itself be staging-dropped) and is emitted only when the delta
  is positive, so no-overflow paths keep their exact event counts.
- **Exporters**: registered through the shared metric descriptor (from #1894), so
  both Prometheus and DOCA serve `agent_telemetry_events_dropped_total` — our own
  counter through the standard counter path, no DOCA-native "dropped" feature
  involved.
- **Scope**: counts only staging-queue loss — events that never reach an exporter.
  It does **not** count BUFFER cyclic-ring loss from a slow downstream reader (a
  separate, uncounted condition).
- **Compatibility**: new event type is a stream schema change, so
  `TELEMETRY_VERSION` is bumped **3 → 4** and `examples/python/telemetry_reader.py`
  updated to match.
- **Docs**: `docs/telemetry.md` (new metric + version note), Prometheus/DOCA/plugin
  READMEs.

Design/plan: `docs/proposals/metrics-loss-detection-plan.md`.

## Tests

Deterministic and hardware-independent (no concurrent-reader ring races):

- **Core** (`telemetry_test.cpp`): `DroppedEventsAppearInBufferStream` stages
  `capacity-2` accepted events + forces `addXferStats` to reject whole batches, so
  the single flush pushes exactly `capacity-1` into the ring — provably lossless,
  single-threaded. Verifies the synthetic reaches the BUFFER stream with the exact
  drop count and `produced == exported + dropped`.
- **Prometheus** (`telemetry_prometheus_test.cpp`): drives `nixlTelemetry` end-to-end
  through the (lossless, ring-free) Prometheus exporter for both `updateData` and
  `addXferStats` overflow. Conservation `accepted + dropped == produced` holds
  regardless of flush timing; plus a direct cumulative-counter check. Both tests
  also assert `dropped > 0`, so a zero-drop or drop-but-don't-count implementation
  cannot pass.
- **DOCA** (`telemetry_doca_nixl_test.cpp`, `TELEMETRY_DOCA`-gated): scrapes
  `agent_telemetry_events_dropped_total` == sum of emitted deltas.

Run via CTest / the gtest + doca_nixl_test binaries: telemetry 26/26, DOCA 7/7,
stable across repeated runs.

## Notes

- Rebased onto #1894 (unified exporter descriptor contract).
- NIX-1205 (extended/stress telemetry tests) is now unblocked.
- Separately noticed (not addressed here, out of scope): `nixlTelemetry::buffer_`
  is never assigned, so the destructor's `if (buffer_) flushPendingEvents()` block
  is dead code — the only flush path is the periodic task. Worth a follow-up ticket.

## Summary by CodeRabbit

* **New Features**
  * Added a new telemetry metric for dropped events, exposed in Prometheus and DOCA exports.
  * Telemetry now reports producer-side queue drops as a synthetic event during flushes.

* **Bug Fixes**
  * Improved telemetry accounting so dropped events are captured instead of silently lost.
  * Updated telemetry version handling to require the newer format for readers and exporters.

* **Documentation**
  * Documented the new dropped-events metric and updated telemetry version notes in the telemetry guides.


src/core/telemetry/telemetry.cpp
```diff
@@ -293,13 +311,17 @@ nixlTelemetry::addXferStats(std::chrono::microseconds xfer_time,
...
+
     const auto bytes_type = is_write ? nixl_telemetry_event_type_t::AGENT_TX_BYTES :
                                        nixl_telemetry_event_type_t::AGENT_RX_BYTES;
     const auto requests_type = is_write ? nixl_telemetry_event_type_t::AGENT_TX_REQUESTS_NUM :
                                           nixl_telemetry_event_type_t::AGENT_RX_REQUESTS_NUM;
 
     const std::lock_guard lock(mutex_);
-    if (events_.size() + 4 > maxBufferedEvents_) {
+    if (events_.size() + xfer_stats_events > maxBufferedEvents_) {
+        droppedEvents_.fetch_add(xfer_stats_events, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
e-eygin: Skipping: the repo `.clang-tidy` sets `readability-identifier-naming.LocalVariableCase = lower_case`, so `xfer_stats_events` (snake_case) is the correct convention and matches the adjacent locals (`bytes_type`, `total_dropped`). Renaming to camelCase would violate the configured lint.

e-eygin:

src/core/telemetry/telemetry.cpp
```diff
@@ -219,9 +219,26 @@ nixlTelemetry::flushPendingEvents() {
...
 }
 
+void
+nixlTelemetry::exportDroppedEvents() {
+    // Emit the drops accumulated since the last flush as a synthetic
+    // AGENT_EVENTS_DROPPED event; bypassing the staging queue, it can never
+    // itself be staging-dropped. Only when the counter advanced: keeps the
+    // no-overflow path event-free (preserving exact-count contracts) and avoids
+    // unsigned underflow if the counter ever regresses.
+    const uint64_t total_dropped = droppedEvents_.load(std::memory_order_relaxed);
```
ColinNV: Both x86_64 and Aarch64 have atomic swap operations, this could use `const uint64_t dropped = droppedEvents_.exchange( 0, std::memory_order_relaxed );` and remove `lastReportedDropped_`?

examples/python/telemetry_reader.py
```diff
@@ -56,6 +56,7 @@
 AGENT_ERR_REMOTE_DISCONNECT = 17
 AGENT_ERR_CANCELED = 18
 AGENT_ERR_NO_TELEMETRY = 19
+AGENT_EVENTS_DROPPED = 20
```
ColinNV: If backend events are re-introduced again in the future, would they use the same counter on overflow, a different counter, or are they on a different path and the question doesn't make sense?

src/core/telemetry/telemetry.cpp
e-eygin: Done in `4e2b2734` — switched to `droppedEvents_.exchange(0, std::memory_order_relaxed)` and removed `lastReportedDropped_`. Simpler and drops the subtraction entirely. Thanks!

examples/python/telemetry_reader.py
e-eygin: Same counter, same path. All event types share the single staging queue, and `agent_events_dropped` counts staging drops in aggregate (not per-type), so re-introduced backend events would be counted by it on overflow. Per-type breakdown was deferred and could be added later as a labeled drop if needed.

ColinNV: I was just wondering whether we can prevent a future rename...

e-eygin: No rename needed. `agent_` is the shared prefix on every series this agent emits (`agent_tx_bytes`, `agent_errors_total`, …), so it isn't a per-type tag, and `events_dropped` is generic — re-introduced backend events would just increment this same aggregate, keeping the name accurate. If per-type visibility is ever wanted, that's additive (a `type=` label or sibling series), which never renames `agent_events_dropped`.

ColinNV: Thinking long-term, "events" is quite generic, have you thought about `AGENT_TELEMETRY_DISCARD` or `AGENT_TELEMETRY_OVERFLOWS`? Of course it's fine if you already discarded such ideas :-)

ColinNV (commented):
LGTM but have you thought about `agent_telemetry_events_dropped`

e-eygin:
@ColinNV Renamed to `agent_telemetry_events_dropped` and rebased onto #1894 — thanks.

ColinNV (approved):

PR #69, opened 2025-03-20
ucx_mo_backend_test: add distribution function

artpol84:
To More thoroughly test block-to-device distribution Make Source and destination block-to-device distribution uneven:
 * For source blocks use round-robin
 * For destination - block-wise

PR #1888, opened 2026-07-05
Build: skip the DOCA SDK install when the base ships only the DOCA runtime

Alexey-Rivkin:
## What?
When the base image ships the DOCA runtime (`doca-sdk-common`) without the GPUNETIO SDK, skip the SDK install instead of failing. Full-SDK bases still skip as before; DOCA-less bases still install from the pinned 3.3 repo.

## Why?
The daily DLFW pytorch base moved to the DOCA 3.4 runtime while our repo pin is 3.3. Installing the 3.3 `-dev` packages against the 3.4 runtime is an apt dependency conflict, so every DLFW container build fails (nightly [#69](https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/job/nixl-ci-build-container/69/)).

## How?
GPUNETIO is an optional meson plugin and disables cleanly when the SDK is absent (`doca-sdk-common` is the runtime only - it does not provide the GPUNETIO SDK). Installing a version-matched SDK from the online DOCA repo is possible but adds moving parts (repo URL scheme, keyring, version derivation) for one optional plugin; if GPUNETIO on DLFW bases turns out to be needed, that can be a follow-up.

Earlier revisions of this PR carried a torch fallback in `build-wheel.sh` for bases without public torch wheels - dropped, superseded by #1896 (the image no longer builds wheels).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container build behavior for DOCA-based images by skipping a pinned SDK install when the base image already includes a compatible runtime package. This helps avoid version conflicts and build failures.
  * Expanded the build’s detection logic to handle more base-image scenarios cleanly, reducing unnecessary installation attempts.

### DIFF
--- a/benchmark/nixlbench/contrib/Dockerfile
+++ b/benchmark/nixlbench/contrib/Dockerfile
@@ -50,7 +50,9 @@ RUN apt-get update -y && \
     build-essential
 
 # Add DOCA repo + upgrade always so the RDMA reinstall below resolves MLNX/DOCA versions.
-# Skip only the DOCA SDK install when the base image already ships it.
+# Skip the DOCA SDK install when the base image already ships the SDK, and also when it
+# ships only the DOCA runtime at another version: the pinned repo's SDK can't install
+# against it (apt refuses the downgrade), and the GPUNETIO plugin is optional in meson.
 RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi) && \
     MELLANOX_OS="$(. /etc/lsb-release; echo ${DISTRIB_ID}${DISTRIB_RELEASE} | tr A-Z a-z | tr -d .)" && \
     wget --tries=3 --waitretry=5 --no-verbose https://www.mellanox.com/downloads/DOCA/DOCA_v3.3.0/host/doca-host_3.3.0-088000-26.01-${MELLANOX_OS}_${ARCH_SUFFIX}.deb -O doca-host.deb && \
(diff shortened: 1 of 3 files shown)


benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -59,6 +61,8 @@ RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64"; else echo "am
     apt-get upgrade -y && \
     if dpkg -s doca-sdk-gpunetio >/dev/null 2>&1; then \
         echo "DOCA SDK already installed in base image, skipping SDK install"; \
+    elif dpkg -s doca-sdk-common >/dev/null 2>&1; then \
```
dpressle, line 64: this is not clear why it fixes the issue.
if you dont have gpunetio package but you have common package does this means we dont need gpunetio or that it is part of common

dpressle (changes requested):

benchmark/nixlbench/contrib/Dockerfile
Alexey-Rivkin: gpunetio is not part of common

benchmark/nixlbench/contrib/Dockerfile
```diff
@@ -59,6 +61,8 @@ RUN ARCH_SUFFIX=$(if [ "${ARCH}" = "aarch64" ]; then echo "arm64"; else echo "am
...
+        echo "Base image ships DOCA runtime $(dpkg-query -W -f='${Version}' doca-sdk-common) without the GPUNETIO SDK, skipping SDK install to avoid a version mismatch"; \
     else \
         apt-get install -y --no-install-recommends doca-sdk-gpunetio libdoca-sdk-gpunetio-dev libdoca-sdk-verbs-dev libdoca-sdk-telemetry-exporter-dev collectx-clxapidev; \
     fi
```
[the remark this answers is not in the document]
Alexey-Rivkin: Intentional - unconditional skip by design. GPUNETIO is optional and version-compare adds complexity for a base that doesn't exist

benchmark/nixlbench/contrib/Dockerfile
dpressle: So it does not make any sense the current logic say if gpunetio is not install then install it - your change causes it not being installed when its not installed but we have common installed which does not have gpunetio in it as you said.

Alexey-Rivkin: The change causes the installation to skip if either gounetio runtime or common is already present (separated cases for clear logging). Only install on a clean slate.
This fixes the current failure when trying to install a mismatched gounetio version.

Alexey-Rivkin:
The original issue is now resolved on the DLFW side.

PR #1894, opened 2026-07-06
telemetry: unify native Prometheus & DOCA exporter metric output (parity)

e-eygin:
## What?

Unifies the metric output of the two Prometheus-family telemetry exporters — the
native (prometheus-cpp) exporter and the DOCA/CollectX exporter — so both emit
**identical** `agent_*` series (same names, types, and coverage).

A single shared `constexpr` descriptor (`telemetryMetricDescriptor` +
`telemetry_metric_event_types` in `src/core/telemetry/telemetry_event.h`) now
drives both exporters, replacing the two hand-written counter/gauge policies that
had drifted (`initializeMetrics()`/`registerCounter()`/`registerGauge()` on the
native side vs `isCounterEvent()`/`gaugeMetricName()` on the DOCA side).

Net changes:
- DOCA counters gain the OpenMetrics `_total` suffix and the previously-missing
  memory and xfer/post-time counters.
- The native exporter gains the `agent_xfer_time` / `agent_xfer_post_time`
  last-op gauges.
- `docs/telemetry.md` and both plugin READMEs updated to identical metric tables.

Exporter-side only: no event-type or buffer-format change, and **no
`TELEMETRY_VERSION` bump**.

## Why?

The two exporters diverged for the same events — e.g. native exposed
`agent_xfer_time_total` (counter) while DOCA exposed `agent_xfer_time` (gauge);
DOCA counters lacked the `_total` suffix; DOCA emitted no memory counters. A user
switching telemetry backends got different metric names/types and different
dashboards. Consolidating onto one descriptor makes parity structural (a `switch`
with no `default`, so `-Wswitch` turns any future missing metric into a compile
error).

Tracking issue: NIX-1598. Also a prerequisite for the DOCA IPC→DTS work
(NIX-1599), which runs the DOCA exporter on a different CollectX backend and
therefore inherits its series shape.

Not a breaking change: the DOCA exporter has never shipped in a released NIXL
version, so there are no DOCA scrape consumers — the `_total` rename just gets the
OpenMetrics shape right before first release.

## How?

- Add the shared descriptor + metric-event list to `telemetry_event.h`; both
  exporters and the tests read from it.
- Native `initializeMetrics()` iterates the descriptor and stops appending
  `_total` locally; DOCA `exportEvent()` drives off it, and
  `isCounterEvent()`/`gaugeMetricName()` are deleted (this also removes DOCA's
  per-event `std::string` allocation, superseding #1886). Error counters keep the
  `status`-labelled path; `agent_errors_total` comes from the shared definition.
- Tests (GoogleTest, run via CTest):
  - `telemetryMetricContract` pins the descriptor's exact name/type contract.
  - `prometheusTelemetryTest.ScrapeEmitsExactlyTheDescriptorSeries` and
    `docaNixlExporterTest.ScrapeEmitsExactlyTheDescriptorSeries` each scrape the
    live endpoint and assert the emitted `agent_*` series set equals the shared
    descriptor oracle — both using the **same** OpenMetrics parser. (The two
    exporters can't be co-linked — DOCA gRPC vs nixl-core gRPC — so each is
    validated in its own binary against the common oracle.)
  - Existing byte counter/last-gauge and error-status tests updated to `_total`.
- `ninja -C build` green; telemetry gtests (43) + DOCA runtime tests
  (`doca_test`, `doca_nixl_test`, real CollectX) pass; clang-format clean.

## Summary by CodeRabbit

- **New Features**
  - Telemetry metrics are now driven by a unified descriptor contract, aligning Prometheus and DOCA series names and semantics.
  - Transfer-time and byte metrics now expose both cumulative `_total` counters and corresponding last-value gauges (when applicable).

- **Bug Fixes**
  - Fixed inconsistent counter/gauge support and tx/rx byte-series naming between exporters.

- **Documentation**
  - Updated telemetry documentation and exporter README tables to reflect the unified metric series and naming.

- **Tests**
  - Added contract validation and scrape-coverage tests to confirm exporters emit exactly the expected series.


e-eygin:

ColinNV (approved):

PR #1895, opened 2026-07-07
[Python] Reduce latency of NIXL import

ovidiusm:
## What?
Speedup NIXL wheel import time

## How?

Slowest dependency import was profiled to be PyTorch.

Changes:

* Import `torch` only as needed. Not all projects using `nixl` use `torch`.
* Lazy import for torch tensor type.
* Rely on `TYPE_CHECKING` to import `torch` in IDEs and static analysis tools.
* Use a faster way to detect CUDA major version, relying on alternative CUDA wrappers (imported PyTorch, CUDA bindings, CuPy, non-imported PyTorch)
* Refactored duplicate code between `nixl` and `nixl.ep` modules to remove repetition

## Summary by CodeRabbit

* **New Features**
  * Improved automatic selection of CUDA-enabled Python backends based on the detected CUDA version.

* **Bug Fixes**
  * Reduced startup issues by avoiding unnecessary GPU library imports during module loading.
  * Made backend selection more reliable when CUDA-related packages are already available or installed in different combinations.
  * Updated error messages to better reflect the detected CUDA version when a matching backend is missing.


galletas1712 (dismissed):

src/bindings/python/nixl-meta/nixl_meta_utils.py
```diff
@@ -0,0 +1,97 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
iyastreb, line 1: maybe name it nixl_meta_utils to avoid collisions with existing nixl_utils?

src/bindings/python/nixl-meta/nixl_meta_utils.py
```diff
@@ -0,0 +1,97 @@
...
+# limitations under the License.
+
+"""Shared helpers for the NIXL meta packages (``nixl`` and ``nixl_ep``)."""
+
+import importlib.util
+import pathlib
+import sys
+
+
+def detect_cuda_major() -> int | None:
```
iyastreb, line 23: Do we need for nixl_ep to prefer Torch’s CUDA version even if Torch is not imported yet, because nixl_ep is Torch-dependent? For plain nixl, the current “use already-loaded CUDA stack” behavior is more defensible.

```
def detect_cuda_major(*, prefer_torch: bool = False) -> int | None:
 
...
    if prefer_torch:
        found = major(_torch_cuda_version_from_disk())
        if found is not None:
            return found

    cuda_bindings = sys.modules.get("cuda.bindings")
```

ovidiusm: I think it's better to reuse the same code to detect the CUDA version instead of diverging. The result should be the same.

src/bindings/python/nixl-meta/nixl_meta_utils.py
ovidiusm: Good point, changed

README.md
```diff
@@ -411,3 +411,4 @@ For contribution guidelines, see [CONTRIBUTING.md](https://github.com/ai-dynamo/
 ## Third-Party Components
 
 This project will download and install additional third-party open source software projects. Review the license terms of these open source projects before use.
+
```
iyastreb: minor: redundant change

iyastreb (dismissed):

README.md
ovidiusm: Right... It's a missing line at EOF, let's add it

guy-ealey-morag: remove empty line

src/bindings/python/nixl-meta/nixl_meta_utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def major(version: str | None) -> int | None:
+        return int(version.split(".")[0]) if version else None
+
+    # torch already imported: use its official API.
+    torch = sys.modules.get("torch")
+    if torch is not None:
+        found = major(getattr(getattr(torch, "version", None), "cuda", None))
+        if found is not None:
+            return found
```
guy-ealey-morag:
```suggestion
        version = major(getattr(getattr(torch, "version", None), "cuda", None))
        if version is not None:
            return version
```

`found` made me think it's boolean, I suggest renaming (also in other places in this func)

src/api/python/_api.py
```diff
@@ -13,19 +13,37 @@
...
 from .logging import get_logger
 
+if TYPE_CHECKING:
+    import torch
+
 # Get logger using centralized configuration
 logger = get_logger(__name__)
 
+
+def _torch_tensor_types() -> tuple:
```
guy-ealey-morag:
```suggestion
def _torch_tensor_types() -> tuple[type, ...]:
```

Indicate it's a tuple of types

README.md
ovidiusm: done

src/bindings/python/nixl-meta/nixl_meta_utils.py
ovidiusm: Good point. I renamed all occurences

src/api/python/_api.py
ovidiusm: I think this helper is too complicated, I pushed a simpler verion

guy-ealey-morag (approved):

src/bindings/python/nixl-meta/nixl_meta_utils.py
```diff
@@ -0,0 +1,97 @@
...
+        version_spec = importlib.util.spec_from_file_location(
+            "_nixl_torch_version", version_py
+        )
+        if version_spec is None or version_spec.loader is None:
+            return None
+        module = importlib.util.module_from_spec(version_spec)
+        version_spec.loader.exec_module(module)
+    except Exception:
+        return None
+    return getattr(module, "cuda", None)
```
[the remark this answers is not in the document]
ovidiusm: PyTorch works on CPU-only installs. Also we do not need to optimize perf there. This comment is nonsense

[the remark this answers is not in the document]
ovidiusm: This is covered in integration tests

mkhazraee (approved):

PR #1897, opened 2026-07-07
telemetry/doca: add IPC->DTS delivery backend (NIX-1599)

e-eygin:
## What?
Add an IPC->DTS delivery mode to the DOCA/CollectX telemetry exporter. Delivery backends
are now selected by `NIXL_TELEMETRY_DOCA_BACKENDS` — a comma-separated **set** (default
`scrape`; e.g. `scrape`, `ipc`, `scrape,ipc`), parsed with `absl::StrSplit` +
`absl::StripAsciiWhitespace` (the same idiom `nixl::trace` uses for `NIXL_TRACE_BACKENDS`),
plus `NIXL_TELEMETRY_DOCA_IPC_SOCKETS_DIR` for the shared socket folder:
- `scrape` — opens the local Prometheus HTTP endpoint (existing behavior).
- `ipc` — pushes metrics over IPC to the DOCA Telemetry Service (DTS); no HTTP endpoint of
  its own. If DTS is unreachable the exporter logs a WARN and continues (no crash).
- The two compose; unknown tokens warn+ignore; an empty/unknown set falls back to `scrape`.

Default is unchanged (`scrape`), so this is backward-compatible. Exporter-side only — no
`TELEMETRY_VERSION` bump. Remote-write / OTLP remain DTS-side onward backends (reached via
`ipc`), not exporter flags.

## Why?
Under multi-process runs (TP/DP/PP/EP) every NIXL process opens its own scrape endpoint and
they collide on the shared port, so only one rank exports (NIX-1557). Pushing over IPC to a
single per-node DTS aggregates all processes behind one Prometheus endpoint with no
per-process listener — the strategic DOCA-path fix. NIX-1590 (per-rank port offset) stays the
near-term mitigation; DTS deployment (K8s DaemonSet/sidecar, multi-process aggregation,
per-rank identity) is the follow-up NIX-1600. Tracking: NIX-1599. Builds on the unified
exporter descriptor from #1894 (NIX-1598).

## How?
All wiring is in the `DocaSharedContext` ctor. In `ipc` mode it skips
`setenv(PROMETHEUS_ENDPOINT)`, calls `doca_telemetry_exporter_schema_set_ipc_enabled`
(+ `set_ipc_sockets_dir` when configured) between `schema_init` and `schema_start`, and after
`source_start` calls `check_ipc_status`, logging WARN-and-continue when not `CONNECTED`. Links
`absl_strings` into the plugin + test. Adds DOCA-gated unit tests (backend gating,
multi-backend, graceful degradation without DTS) and an opt-in `DISABLED_` end-to-end test
that pushes over IPC and scrapes a live DTS. Validated end-to-end (client DOCA 3.3 → DTS 3.5):
`agent_tx_bytes_total=7000` + `agent_tx_last_bytes=4000` observed at the DTS endpoint. Docs:
DOCA plugin README + `docs/telemetry.md`.

## Summary by CodeRabbit

* **New Features**
  * Added experimental DOCA/CollectX telemetry delivery backends with environment-driven selection for `scrape` and/or `ipc`, plus optional IPC socket directory control.
* **Bug Fixes**
  * If IPC can’t reach the telemetry service, the process continues running and metrics exporting via other enabled backends still works.
  * If no known backend is selected, the exporter defaults to `scrape`; unknown backend tokens now fail fast.
* **Documentation**
  * Updated telemetry documentation and the DOCA exporter README with backend configuration and runtime behavior details.
* **Tests**
  * Added end-to-end coverage for backend selection, IPC-only endpoint behavior, scrape+IPC mixed operation, and a disabled live IPC integration test.

### DIFF
--- a/src/plugins/telemetry/doca/doca_exporter.cpp
+++ b/src/plugins/telemetry/doca/doca_exporter.cpp
@@ -21,6 +21,8 @@
 
 #include <doca_telemetry_exporter.h>
 #include <doca_error.h>
+#include <absl/strings/ascii.h>
+#include <absl/strings/str_split.h>
 #include <array>
 #include <cstdint>
 #include <cstdlib>
@@ -47,6 +49,10 @@ constexpr uint64_t histogramFlushIntervalMs = 1000;
 
 const char docaPrometheusPortVar[] = "NIXL_TELEMETRY_DOCA_PROMETHEUS_PORT";
 const char docaPrometheusLocalVar[] = "NIXL_TELEMETRY_DOCA_PROMETHEUS_LOCAL";
+const char docaBackendsVar[] = "NIXL_TELEMETRY_DOCA_BACKENDS";
+const char docaIpcSocketsDirVar[] = "NIXL_TELEMETRY_DOCA_IPC_SOCKETS_DIR";
+const char docaBackendScrape[] = "scrape";
+const char docaBackendIpc[] = "ipc";
 
 const std::string docaExporterLocalAddress = "http://127.0.0.1";
 const std::string docaExporterPublicAddress = "http://0.0.0.0";
@@ -58,6 +64,42 @@ getBindAddress() {
         nixl::config::getValueDefaulted(docaPrometheusPortVar, docaPrometheusExporterDefaultPort);
     return (local ? docaExporterLocalAddress : docaExporterPublicAddress) + ":" +
         std::to_string(port);
+}
+
+struct DocaBackends {
+    bool scrape = false;
+    bool ipc = false;
+};
+
+[[nodiscard]] DocaBackends
+parseBackends() {
+    const std::string spec =
+        nixl::config::getValueDefaulted<std::string>(docaBackendsVar, docaBackendScrape);
+    DocaBackends backends;
+    for (const absl::string_view raw : absl::StrSplit(spec, ',')) {
+        const absl::string_view token = absl::StripAsciiWhitespace(raw);
+        if (token.empty()) {
+            continue;
+        }
+        if (token == docaBackendScrape) {
+            backends.scrape = true;
+        } else if (token == docaBackendIpc) {
+            backends.ipc = true;
+        } else {
+            throw std::runtime_error(std::string(docaBackendsVar) + ": unknown backend '" +
+                                     std::string(token) + "'; valid backends are '" +
+                                     docaBackendScrape + "', '" + docaBackendIpc + "'");
+        }
+    }
+    if (!backends.scrape && !backends.ipc) {
+        backends.scrape = true;
+    }
+    return backends;
+}
+
+[[nodiscard]] std::string
+getIpcSocketsDir() {
+    return nixl::config::getValueDefaulted<std::string>(docaIpcSocketsDirVar, std::string());
 }
 
 [[nodiscard]] std::string
@@ -88,10 +130,13 @@ std::mutex g_metrics_mutex;
 /**
  * @brief Process-wide shared DOCA context
  *
- * DOCA only supports one metrics context per process, so all agents share
- * this context. The underlying CLX Metrics API is not thread-safe, so all
- * metric recording calls (metrics_add_counter / metrics_add_gauge) are
- * serialised by a dedicated mutex (g_metrics_mutex).
+ * All agents in a process share one context. The exporter drives process-global
+ * CollectX state -- the Prometheus endpoint via the PROMETHEUS_ENDPOINT env var
+ * (one HTTP server) and a fixed schema name -- so independently-configured
+ * per-agent contexts would collide; sharing one avoids that. The underlying CLX
+ * Metrics API is not thread-safe, so all metric recording calls
+ * (metrics_add_counter / metrics_add_gauge) are serialised by a dedicated mutex
+ * (g_metrics_mutex).
  */
 struct DocaSharedContext {
     doca_telemetry_exporter_schema *schema = nullptr;
@@ -100,6 +145,9 @@ struct DocaSharedContext {
     doca_telemetry_exporter_label_set_id_t error_label_set_id = 0;
     bool source_started = false;
     bool metrics_context_created = false;
+    const bool scrape_enabled;
+    const bool ipc_enabled;
+    std::string backend_desc;
 
     // Base histograms are templates created once on the shared source; observing
     // against one with concrete label values yields a concrete histogram id. The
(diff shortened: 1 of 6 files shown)


e-eygin:

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -46,6 +52,42 @@ getBindAddress() {
...
+    bool scrape = false;
+    bool ipc = false;
+};
+
+[[nodiscard]] DocaBackends
+parseBackends() {
+    const std::string spec =
+        nixl::config::getValueDefaulted<std::string>(docaBackendsVar, docaBackendScrape);
+    DocaBackends backends;
+    for (const absl::string_view raw : absl::StrSplit(spec, ',')) {
```
ColinNV: Would a "split `std::string(_view)` into `std::vector< std::string >` of non-empty whitespace-stripped tokens" function that can be re-used in multiple similar places be worth it? Possibly in a separate PR.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -198,17 +282,27 @@ nixlTelemetryDocaExporter::nixlTelemetryDocaExporter(
...
     ctx_ = g_ctx_weak.lock();
     if (!ctx_) {
-        ctx_ = std::make_shared<DocaSharedContext>(bind_address);
+        ctx_ = std::make_shared<DocaSharedContext>(
+            backends.scrape, backends.ipc, bind_address, ipc_sockets_dir);
         g_ctx_weak = ctx_;
-        NIXL_INFO << "DOCA Telemetry exporter initialized on " << bind_address;
+        NIXL_INFO << "DOCA Telemetry exporter initialized (" << ctx_->backend_desc << ")";
     } else {
+        if (backends.scrape != ctx_->scrape_enabled || backends.ipc != ctx_->ipc_enabled) {
```
ColinNV: How can this be triggered, only if the environment is changed between multiple agent constructions?

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Agreed — this same `StrSplit(',') + StripAsciiWhitespace + skip-empty` loop is
also open-coded in `nixl::trace`'s `resolveTraceBackends` (`nixl_agent.cpp`), so a
shared helper is worth it. Since unifying it touches core (outside this DOCA
exporter PR), I've kept it out of here per your "separate PR" suggestion and filed
NIX-1621 to add a `nixl::str::splitStripped(...)` helper and migrate both sites.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Correct — it only triggers when `NIXL_TELEMETRY_DOCA_BACKENDS` differs between
agent constructions in the same process. NIXL shares one DOCA context per process
(`g_ctx_weak`) because the exporter drives process-global CollectX state — the
Prometheus endpoint via `PROMETHEUS_ENDPOINT` (one HTTP server) and a fixed schema
name. `parseBackends()` re-reads the env each construction, so a later agent
requesting different backends reuses the first context; the WARN surfaces that its
request was ignored. Normal single-config runs never hit it.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -46,6 +52,42 @@ getBindAddress() {
...
+        const absl::string_view token = absl::StripAsciiWhitespace(raw);
+        if (token.empty()) {
+            continue;
+        }
+        if (token == docaBackendScrape) {
+            backends.scrape = true;
+        } else if (token == docaBackendIpc) {
+            backends.ipc = true;
+        } else {
+            NIXL_WARN << docaBackendsVar << ": ignoring unknown backend '" << token << "'";
```
ColinNV: I am generally pushing towards making any config input that can't be used/parsed a hard error, the rationale being that ignoring such values probably means that (a) the user had an intention when setting up the value, but (b) we are not following the intention because of some mistake. Not making this a hard error means that we are continuing with a value that the user probably didn't intend...

ColinNV: https://github.com/ai-dynamo/nixl/pull/1914/changes#r3558649736

e-eygin: Done in 79cbd175 — parseBackends now throws on an unrecognized token (listing the
valid values), so e.g. `scrape,bogus` fails telemetry init loudly rather than
silently degrading. Unset/empty still defaults to `scrape`.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -60,6 +65,38 @@ getBindAddress() {
...
+        nixl::config::getValueDefaulted<std::string>(docaBackendsVar, docaBackendScrape);
+    DocaBackends backends;
+    for (const std::string &token : nixl::str::splitStrippedSet(spec)) {
+        if (token == docaBackendScrape) {
+            backends.scrape = true;
+        } else if (token == docaBackendIpc) {
+            backends.ipc = true;
+        } else {
+            throw std::runtime_error(std::string(docaBackendsVar) + ": unknown backend '" +
+                                     std::string(token) + "'; valid backends are '" +
```
ColinNV: `token` already is a `std::string`, also `src/utils/common/exception.h`.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -241,17 +323,28 @@ nixlTelemetryDocaExporter::nixlTelemetryDocaExporter(
...
+    const DocaBackends backends = parseBackends();
     const std::string bind_address = getBindAddress();
+    const std::string ipc_sockets_dir = getIpcSocketsDir();
 
     const std::lock_guard lock(g_ctx_mutex);
     ctx_ = g_ctx_weak.lock();
     if (!ctx_) {
-        ctx_ = std::make_shared<DocaSharedContext>(bind_address);
+        ctx_ = std::make_shared<DocaSharedContext>(
+            backends.scrape, backends.ipc, bind_address, ipc_sockets_dir);
```
ColinNV: Why not pass `backends` all the way down?

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Done — dropped the redundant `std::string(token)` and switched to `nixl::throwRuntimeError` (`common/exception.h`).

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Done — folded `bind_address`/`ipc_sockets_dir` into `DocaExporterConfig` (renamed from `DocaBackends`); the ctor now takes the single config.

ColinNV (approved):

Issue #1838, opened 2026-06-25
Prometheus exporter issues: Failed to create exporter: null context when constructing CivetServer. Possible problem binding to port


minhminh2k:
### Summary

I am deploying NVIDIA Dynamo on Kubernetes to serve a GLM-4.7 model using Prefill/Decode disaggregation across two 8xH200 nodes, with tensor parallelism = 8 and InfiniBand-enabled networking.

When enabling NIXL Prometheus telemetry using the following environment variables and security context, the model starts successfully, but the NIXL Prometheus exporter fails on both the Prefill and Decode components.

Deployment configuration:
```
securityContext:
  capabilities:
    add:
      - IPC_LOCK
      - SYS_RESOURCE

env:
  - name: NIXL_TELEMETRY_ENABLE
    value: "y"
  - name: NIXL_TELEMETRY_EXPORTER
    value: prometheus
  - name: NIXL_TELEMETRY_PROMETHEUS_PORT
    value: "19090"
  - name: UCX_IB_ADDR_TYPE
    value: eth
  - name: UCX_NET_DEVICES
    value: mlx5_0:1,mlx5_3:1,mlx5_4:1,mlx5_5:1,mlx5_6:1,mlx5_7:1,mlx5_8:1,mlx5_9:1
  - name: DYN_SYSTEM_ENABLED
    value: "true"
  - name: DYN_SYSTEM_PORT
    value: "9090"
  - ...

ports:
  - containerPort: 9090
    name: system
    protocol: TCP
  - containerPort: 19090
    name: nixl
    protocol: TCP

resources:
  limits:
    cpu: '64'
    nvidia.com/gpu: '8'
    rdma/mlnx_rdma: '8'
  requests:
    cpu: '64'
    nvidia.com/gpu: '8'
    rdma/mlnx_rdma: '8'

args:
    - '--model'
    - /models/zai-org/GLM-4.7-FP8
    - '--served-model-name'
    - zai-org/GLM-4.7-FP8
    - '--tensor-parallel-size'
    - '8'
    - '--disaggregation-mode'
    - decode
    - '--disaggregation-transfer-backend'
    - nixl
    - ...
  command:
    - python3
    - '-m'
    - dynamo.sglang


 ```

#### Actual behavior
The DynamoGraphDeployment starts successfully and the model is served as expected.
However, the NIXL Prometheus exporter throws an error on both the Prefill and Decode components.
```
[...]

PR #1898, opened 2026-07-07
telemetry: downgrade Prometheus bind-collision from ERROR to WARN

e-eygin:
## What?

Make a NIXL Prometheus telemetry scrape-port bind collision **benign and non-fatal**:
the losing process logs a single `WARNING` and continues **without** a telemetry sink,
instead of logging a per-rank `ERROR` and throwing out of the `nixlAgent` constructor.

- New dedicated exception `nixlTelemetryBindFailed` (in `telemetry_exporter.h`).
- The Prometheus exporter translates **only** the civetweb port-bind failure
  (`"Failed to setup server ports"`) into it; every other startup failure
  (thread creation, ACL, OOM, ...) is rethrown unchanged and stays a genuine `ERROR`.
- The plugin creator rethrows `nixlTelemetryBindFailed`; `nixlTelemetry::create()`
  catches it, logs one `NIXL_WARN`, and returns `nullptr`, so the agent runs with
  telemetry disabled.
- gtest coverage for both the thrown type and the non-fatal WARN path.
- Docs: multi-process note in `src/plugins/telemetry/prometheus/README.md` and
  `docs/telemetry.md`.

Scope: native Prometheus only. DOCA is out of scope (its bind happens inside CollectX
after `setenv`, and is harder to catch cleanly).

## Why?

Reported in [ai-dynamo/nixl#1838](https://github.com/ai-dynamo/nixl/issues/1838) (NIX-1557):
under multi-process runs (TP/DP/…) every rank binds the same
`NIXL_TELEMETRY_PROMETHEUS_PORT`; only one wins. On `main` this was worse than mere log
spam — the losing ranks **threw `std::runtime_error` out of the agent constructor**
(`makeExporter` throws on the null exporter and the `create()` call site is unguarded),
a fatal regression from 1.2.0, which continued without a telemetry sink. This restores
non-fatal behavior and removes the alarming ERROR spam.

Tracked as NIX-1604 (the graceful-bind half carved out of NIX-1590; the disputed per-rank
port offset is separate and on hold).

## How?

The benign case is distinguished from real failures by a **dedicated exception type**
rather than a blanket ERROR→WARN downgrade (which would also silence genuine
misconfiguration). Net user-visible effect: at most one `WARN` per losing rank, no `ERROR`,
no crash; the process that won the port still exports normally.

Example log on a losing rank:

    W ... telemetry.cpp:98] Prometheus telemetry endpoint '0.0.0.0:9090' could not be bound (likely already in use by another process); this process will run without a telemetry sink

### Test plan

- `ninja -C build && meson test -C build gtest` (telemetry suites)
- New: `prometheusTelemetryTest.BindCollisionThrowsBindFailed`,
  `prometheusTelemetryTest.BindCollisionCreateIsNonFatalWarn`
- Verified green locally: core telemetry (21), Prometheus (10, incl. the 2 new),
  descriptor contract + UCX transfer-telemetry (21).

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry now safely handles Prometheus scrape-port binding conflicts in multi-process runs without crashing.
  * If the Prometheus exporter can’t bind its scrape port, telemetry startup becomes non-fatal (telemetry creation returns `nullptr`) and execution continues.
  * Logs a single, binding-related warning when the collision occurs.

* **Documentation**
  * Updated telemetry and Prometheus exporter docs with expected multi-process port contention behavior.

* **Tests**
  * Added/extended regression tests to deterministically simulate scrape-port bind collisions and verify warning counts and non-fatal behavior.

### DIFF
--- a/test/gtest/telemetry_prometheus_test.cpp
+++ b/test/gtest/telemetry_prometheus_test.cpp
@@ -296,6 +296,9 @@ public:
     ScopedFd(const ScopedFd &) = delete;
     ScopedFd &
     operator=(const ScopedFd &) = delete;
+    ScopedFd(ScopedFd &&) = delete;
+    ScopedFd &
+    operator=(ScopedFd &&) = delete;
 
     int
     get() const {
--- a/test/gtest/telemetry_prometheus_test.cpp
+++ b/test/gtest/telemetry_prometheus_test.cpp
@@ -283,6 +283,29 @@ scrapeCoreOverflow(uint16_t port,
     return result; // ok stays false: full accounting was not observed in time
 }
 
+class ScopedFd {
+public:
+    explicit ScopedFd(int fd) : fd_(fd) {}
+
+    ~ScopedFd() {
+        if (fd_ >= 0) {
+            ::close(fd_);
+        }
+    }
+
+    ScopedFd(const ScopedFd &) = delete;
+    ScopedFd &
+    operator=(const ScopedFd &) = delete;
+
+    int
+    get() const {
+        return fd_;
+    }
+
+private:
+    int fd_ = -1;
+};
+
 int
 occupyLocalPort(uint16_t port) {
     const int fd = ::socket(AF_INET, SOCK_STREAM, 0);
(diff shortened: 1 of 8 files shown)


e-eygin:

AlexNvd:
LGTM. 
As we discussed, it is a temporary patch. After we define a full solution, to match Dynamo's approach, the problem will be resolved at its root. Then the telemetry loss of the other ranks (those who came after the port was taken by the first, initialized rank) will be avoided.

test/gtest/telemetry_prometheus_test.cpp
```diff
@@ -281,6 +283,90 @@ scrapeCoreOverflow(uint16_t port,
...
+    }
+
+    int
+    otherProblems() const {
+        const std::lock_guard<std::mutex> lock(mutex_);
+        return otherProblems_;
+    }
+
+private:
+    mutable std::mutex mutex_;
```
ColinNV: Minor: `std::atomic< std::size_t >` for the two counters?

ColinNV (dismissed):

test/gtest/telemetry_prometheus_test.cpp
e-eygin: Done — counters are now `std::atomic<std::size_t>`, mutex dropped.

ColinNV (approved):

dpressle:
@ci-triage-agent confirmed

PR #1899, opened 2026-07-07
deps: patch pytest, bytes, tracing-subscriber advisories

yafshar:
## What?

Update dependencies to address three security advisories and fix the Rust
example workspace configuration so Dependabot can resolve the crate
correctly:

- **pytest** `7.4.4` → `9.0.3` (`benchmark/kvbench`): predictable
  `/tmp/pytest-of-<user>` directory handling could allow local denial of
  service or privilege-escalation scenarios; affected versions include
  releases through 9.0.2.
- **bytes** `1.10.1` → `1.11.1` (RUSTSEC-2026-0007): an integer overflow
  in `BytesMut::reserve` could corrupt capacity tracking and lead to
  out-of-bounds slices.
- **tracing-subscriber** `0.3.19` → `0.3.20` (RUSTSEC-2025-0055,
  CVE-2025-58160): ANSI escape sequence injection when logging untrusted
  input.
- Add an empty `[workspace]` table to `examples/rust/Cargo.toml`.

## Why?

All three dependencies are flagged by Dependabot/RUSTSEC.

In addition, Dependabot currently fails to update the Rust example with:

> current package believes it's in a workspace when it's not

This occurs because `examples/rust` resides within the repository tree but
is not a member of the root workspace
(`members = ["src/bindings/rust/"]`). Adding an empty `[workspace]` table
causes Cargo to treat the example as a standalone package, resolving the
error and enabling future automated dependency updates.

## How?

Only the vulnerable dependencies are updated. In
`examples/rust/Cargo.lock`, only the crates required to address the two
Rust advisories are changed; all other lockfile entries are intentionally
left untouched to keep the fix minimal and easy to review.

The existing Rust lockfile was already partially out of sync with the
manifests independent of this change. Once the workspace fix is merged,
Dependabot will be able to refresh the remaining dependencies separately.

## Summary by CodeRabbit

* **Chores**
  * Updated the licensing header in benchmark and Rust example manifests.
  * Pinned the test framework dependency to a newer version for the benchmark project.
  * Adjusted the Rust example manifest to explicitly declare it as separate from the root workspace (via a standalone workspace section).


yafshar:
Hi @aranadive, @ovidiusm, @brminich, could you please take a look at this PR when you have a chance? It only contains dependency updates to address security advisories and avoid CI/security check failures.

yafshar:
Hi @aranadive, @ovidiusm, @brminich, could you please take a look at this PR when you have a chance?

PR #1900, opened 2026-07-07
nixlbench: fix HIP/toml++ include ordering on GCC

sbates130272:
## What?

Remove the unnecessary and error-inducing conditional includes in nixlbench source files.

## Why?

ROCm's host_defines.h defines an empty __noinline__ macro on GCC, which breaks toml++'s TOML_HAS_ATTR(__noinline__) check when hip/hip_runtime.h is included before toml++/toml.hpp. utils.h already includes toml++ first and then conditionally includes the GPU headers, so the early redundant includes in utils.cpp and nixl_worker.cpp are both the source of the breakage and unnecessary.

This enables nixlbench to be build on ROCm-based systems.

## Summary by CodeRabbit

* **Chores**
  * Cleaned up platform-specific runtime header handling in benchmark components.
  * This change has no expected impact on end-user functionality.

### DIFF
--- a/benchmark/nixlbench/src/utils/utils.cpp
+++ b/benchmark/nixlbench/src/utils/utils.cpp
@@ -27,11 +27,6 @@
 #include <omp.h>
 #include <set>
 
-#if HAVE_CUDA
-#include <cuda_runtime.h>
-#elif HAVE_ROCM
-#include <hip/hip_runtime.h>
-#endif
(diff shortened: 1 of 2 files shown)


sbates130272:
@ColinNV can we get someone to review this as it gates some incoming PRs. Thanks!!

aranadive (approved):

PR #1906, opened 2026-07-08
Refactoring: derive nixlUcxThreadPoolEngine from nixlUcxThreadEngine

iyastreb:
## What?
To unify the code and reuse the same functions
It's pre-requisite for completions API PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UCX backend thread lifecycle so the shared progress thread is created and joined only when enabled and shared workers are available.
  * Notification retrieval now properly runs progress when the shared progress thread is not active.

* **Refactor**
  * Updated the UCX thread-pool engine to inherit from the shared thread-engine, centralizing shared-worker sizing and notification behavior and removing pool-specific notification overrides.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -363,7 +363,11 @@ nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_param
     : nixlUcxEngine(init_params) {
     const size_t num_threads =
         nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
-    numSharedWorkers_ = getWorkers().size() - num_threads;
+    const size_t num_workers = getWorkers().size();
+    if (num_threads > num_workers) {
+        throw std::invalid_argument("num_threads exceeds available UCX workers");
+    }
+    numSharedWorkers_ = num_workers - num_threads;
 
     const size_t shared_count = init_params.enableProgTh ? numSharedWorkers_ : 0;
     if (shared_count == 0) {
@@ -650,13 +654,15 @@ private:
 
 nixlUcxThreadPoolEngine::nixlUcxThreadPoolEngine(const nixlBackendInitParams &init_params)
     : nixlUcxThreadEngine(init_params) {
-    const size_t num_threads =
-        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
-    NIXL_ASSERT(getSharedWorkersSize() > 0);
+    if (getSharedWorkersSize() == 0) {
+        throw std::invalid_argument("thread-pool engine requires at least one shared worker");
+    }
 
     splitBatchSize_ =
         nixl::getBackendParamDefaulted(init_params.customParams, "split_batch_size", 1024u);
 
+    const size_t num_threads =
+        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
     if (num_threads > 0) {
         io_.reset(new asio::io_context());
         dedicatedThreads_.reserve(num_threads);
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -769,8 +769,7 @@ nixlUcxEngine::create(const nixlBackendInitParams &init_params) {
     return std::unique_ptr<nixlUcxEngine>(engine);
 }
 
-nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params,
-                             size_t num_dedicated_workers)
+nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params, size_t num_dedicated_workers)
     : nixlBackendEngine(&init_params),
       sharedWorkerIndex_(1) {
     std::vector<std::string> devs; /* Empty vector */
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -779,8 +779,11 @@ nixlUcxEngine::nixlUcxEngine(const nixlBackendInitParams &init_params, size_t nu
         devs = absl::StrSplit(*opt, ", ");
     }
 
-    numSharedWorkers_ = nixl::getBackendParamDefaulted(custom_params, "num_workers", 1u);
-    const size_t num_workers = numSharedWorkers_ + num_dedicated_workers;
+    size_t num_workers = nixl::getBackendParamDefaulted(custom_params, "num_workers", 1u);
+    if (num_workers <= num_dedicated_workers) {
+        num_workers = num_dedicated_workers + 1;
+    }
+    numSharedWorkers_ = num_workers - num_dedicated_workers;
 
     const size_t num_device_channels =
         nixl::getBackendParamDefaulted(custom_params, "ucx_num_device_channels", 4u);
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -361,20 +361,28 @@ private:
 
 nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
     : nixlUcxEngine(init_params) {
+    const size_t num_threads =
+        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
+    const size_t shared_count = init_params.enableProgTh ? getWorkers().size() - num_threads : 0;
+    if (shared_count == 0) {
+        return;
+    }
+
     if (!nixlUcxMtLevelIsSupported(nixl::ucx::mt_mode_t::WORKER)) {
         throw std::invalid_argument("UCX library does not support multi-threading");
     }
 
-    size_t num_workers = getWorkers().size();
-    thread_ = std::make_unique<nixlUcxSharedThread>(this, num_workers, init_params.pthrDelay);
-    for (size_t i = 0; i < num_workers; i++) {
+    thread_ = std::make_unique<nixlUcxSharedThread>(this, shared_count, init_params.pthrDelay);
+    for (size_t i = 0; i < shared_count; i++) {
         thread_->addWorker(getWorkers()[i].get(), i);
     }
     thread_->start();
 }
 
 nixlUcxThreadEngine::~nixlUcxThreadEngine() {
-    thread_->join();
+    if (thread_) {
+        thread_->join();
+    }
 }
 
 void
@@ -387,6 +395,10 @@ nixl_status_t
 nixlUcxThreadEngine::getNotifs(notif_list_t &notif_list) {
     if (!notif_list.empty()) {
         return NIXL_ERR_INVALID_PARAM;
+    }
+
+    if (!thread_) {
+        progressLoop();
     }
 
     const std::lock_guard lock(notifMutex_);
@@ -635,7 +647,7 @@ private:
 };
 
 nixlUcxThreadPoolEngine::nixlUcxThreadPoolEngine(const nixlBackendInitParams &init_params)
-    : nixlUcxEngine(init_params) {
+    : nixlUcxThreadEngine(init_params) {
     const size_t num_threads =
         nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
     numSharedWorkers_ = getWorkers().size() - num_threads;
@@ -643,15 +655,6 @@ nixlUcxThreadPoolEngine::nixlUcxThreadPoolEngine(const nixlBackendInitParams &in
 
     splitBatchSize_ =
         nixl::getBackendParamDefaulted(init_params.customParams, "split_batch_size", 1024u);
-
-    if (init_params.enableProgTh) {
-        sharedThread_ =
-            std::make_unique<nixlUcxSharedThread>(this, numSharedWorkers_, init_params.pthrDelay);
-        for (size_t i = 0; i < numSharedWorkers_; i++) {
-            sharedThread_->addWorker(getWorkers()[i].get(), i);
-        }
-        sharedThread_->start();
-    }
 
     if (num_threads > 0) {
         io_.reset(new asio::io_context());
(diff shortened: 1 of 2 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -361,20 +361,28 @@ class nixlUcxSharedThread : public nixlUcxThread {
 
 nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
     : nixlUcxEngine(init_params) {
+    const size_t num_threads =
+        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
+    const size_t shared_count = init_params.enableProgTh ? getWorkers().size() - num_threads : 0;
+    if (shared_count == 0) {
+        return;
+    }
```
rakhmets: It seems now that this logic contradicts the comment in `.h` file, that says: `nixlUcxThreadEngine` is an engine with a single progress thread.

src/plugins/ucx/ucx_backend.h
```diff
@@ -323,7 +323,7 @@ namespace asio {
 class io_context;
 }
 
-class nixlUcxThreadPoolEngine : public nixlUcxEngine {
+class nixlUcxThreadPoolEngine : public nixlUcxThreadEngine {
```
rakhmets, line 336: Please consider moving the class to a separate file.
This will help to better identify redundant interfaces in classes.
It may seem like déjà vu (https://github.com/ai-dynamo/nixl/pull/606#discussion_r2222857340, https://github.com/ai-dynamo/nixl/pull/573#discussion_r2276764847), but I really believe it will help make `ucx_backend.[h|cpp]` cleaner.

iyastreb: This would mean that we need to move all the common part from ucx_backend.cpp to a separate h+cpp file.
I'm not against this at all, but it's out of scope of this PR, which aims to do one single thing: reuse common functionality

ColinNV: Right, moving things should be done in a separate PR that does not change the code.

src/plugins/ucx/ucx_backend.cpp
iyastreb: updated comment

src/plugins/ucx/ucx_backend.h
```diff
@@ -300,7 +300,9 @@ class nixlUcxEngine : public nixlBackendEngine {
 class nixlUcxThread;
 
 /**
- * Represents an engine with a single progress thread for all shared workers
+ * Engine with an optional single progress thread that progresses all shared
+ * workers. The thread is started only when there are shared workers; with none
```
rakhmets, line 312: Should we move `numSharedWorkers_` field to `nixlUcxThreadEngine` class?

src/plugins/ucx/ucx_backend.h
```diff
@@ -300,7 +300,9 @@ class nixlUcxEngine : public nixlBackendEngine {
...
+ * (shared_count == 0) no progress thread is created and progress is synchronous.
```
rakhmets: `shared_count` is a local variable of `nixlUcxThreadEngine::nixlUcxThreadEngine`. Please remove it from the comment, describing the class.

src/plugins/ucx/ucx_backend.h
rakhmets: Agreed, this shouldn't be in this PR.
I wanted to highlight this early while it’s still relatively easy to refactor.

src/plugins/ucx/ucx_backend.h
rakhmets: Because a code duplication is introduced by the PR. The same value is calculated in ctors of `nixlUcxThreadEngine` and `nixlUcxThreadPoolEngine`.
Probably it should be a member of `nixlUcxThreadEngine`. The only question is the implementation of `getSharedWorkersSize`, because `nixlUcxThreadEngine` uses `nixlUcxEngine::getSharedWorkersSize` at the moment. Maybe should be updated.

src/plugins/ucx/ucx_backend.h
iyastreb: done

src/plugins/ucx/ucx_backend.h
iyastreb: makes sense, refactored too

src/plugins/ucx/ucx_backend.h
```diff
@@ -310,20 +312,26 @@ class nixlUcxThreadEngine : public nixlUcxEngine {
...
     getNotifs(notif_list_t &notif_list) override;
 
+    size_t
+    getSharedWorkersSize() const override {
+        return numSharedWorkers_;
+    }
+
 protected:
     void
     appendNotif(std::string &&remote_name, std::string &&msg) override;
```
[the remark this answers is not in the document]
iyastreb: fixed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -361,20 +361,30 @@ class nixlUcxSharedThread : public nixlUcxThread {
 
 nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
     : nixlUcxEngine(init_params) {
+    const size_t num_threads =
+        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
+    numSharedWorkers_ = getWorkers().size() - num_threads;
+
+    const size_t shared_count = init_params.enableProgTh ? numSharedWorkers_ : 0;
```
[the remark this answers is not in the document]
iyastreb: fixed

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -361,20 +361,34 @@ class nixlUcxSharedThread : public nixlUcxThread {
 
 nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
     : nixlUcxEngine(init_params) {
+    const size_t num_threads =
+        nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
+    const size_t num_workers = getWorkers().size();
+    if (num_threads > num_workers) {
+        throw std::invalid_argument("num_threads exceeds available UCX workers");
+    }
```
rakhmets: `nixlUcxEngine::nixlUcxEngine` creates at least `num_threads + 1` workers. This check is redundant in the current state of the code.

Something seems to be incorrect here.
We use `num_threads` to decide which UCX engine to create. If it's non-negative we create `nixlUcxThreadPoolEngine`. Else if the progress thread is enabled we create `nixlUcxThreadEngine`, otherwise `nixlUcxEngine`.

I'm not sure how to fix without touching `nixlUcxEngine` code. But it seems that `nixlUcxEngine` should not get `num_threads` parameter. And `nixlUcxEngine` should only create `num_workers` UCP workers.
Probably, `nixlUcxThreadPoolEngine` should modify parameters that are passed to `nixlUcxThreadEngine`( and then passed to `nixlUcxEngine`) ctor.

rakhmets: ~~Maybe it's easier to remove `nixlUcxThreadEngine`?~~
(no, that will require more changes)

iyastreb: I would keep this check as a safety net

I don't see an issue here. nixlUcxEngine checks both num_threads and num_workers and initializes shared workers, this is its responsibility. Then a subclass initializes threads and/or dedicated workers.
I don't see a need to complicate the logic here, looks good imo

iyastreb: I would not remove nixlUcxThreadEngine, because in the future I believe we better remove nixlUcxThreadPoolEngine. Because with SGL in place it should be redundant

rakhmets: > I would keep this check as a safety net
> 
> I don't see an issue here. nixlUcxEngine checks both num_threads and num_workers and initializes shared workers, this is its responsibility. Then a subclass initializes threads and/or dedicated workers. I don't see a need to complicate the logic here, looks good imo

The issue is that the logic processing the input parameter ("num_threads") is scattered in different places.
I'm suggesting to simplify the code and the logic.

At the moment the logic is complicate and fragile. `nixlUcxEngine` updates `num_workers` depending on `num_threads` value. `nixlUcxThreadEngine` also obtains `num_threads`, and validates it. And `nixlUcxThreadPoolEngine` obtains `num_threads`, and checks that it is greater than zero before creating dedicated threads.
And the only class that should use and check this parameter, according to fabric method, is `nixlUcxThreadPoolEngine`.

The goal is to avoid checking indirect values, and handling impossible cases. These complicate further code changes.

At the moment the only thing that should be checked is: `num_threads > 0` in `nixlUcxThreadPoolEngine`.

iyastreb: Ok, I see your point, agree it's a bit messy.
But the only real root cause of this is that `uws` is used both for shared and dedicated workers.
If we clearly split the workers into shared and dedicated ones, it would solve all the issues.
Trying that...

iyastreb: @rakhmets I have split workers into shared and dedicated, now it should be cleaner

src/plugins/ucx/ucx_backend.h
```diff
@@ -208,22 +209,32 @@ class nixlUcxEngine : public nixlBackendEngine {
...
+        return {workers_.data(), numSharedWorkers_};
     }
 
-    const std::unique_ptr<nixlUcxWorker> &
-    getWorker(size_t worker_id) const {
-        return uws[worker_id];
+    [[nodiscard]] worker_span_t
+    getDedicatedWorkers() const {
+        return {workers_.data() + numSharedWorkers_, workers_.size() - numSharedWorkers_};
+    }
```
rakhmets, line 218: This doesn't look correct. It seems that `nixlUcxEngine` shouldn't know about the existence of shared and dedicated workers.
1. `nixlUcxEngine` has `N` workers.
2. `nixlUcxThreadEngine` has `N` workers, and may have a thread shared between `N` workers.
3. `nixlUcxThreadPoolEngine` has `N+M` workers, may have a thread shared between workers `{0,1,...,N-1}` , and has `M` threads dedicated to workers `{N,N+1,...,N+M-1}`.
 
So, `nixlUcxEngine` doesn't really need `getDedicatedWorkers` method, and `numSharedWorkers_` field.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -359,22 +359,29 @@ class nixlUcxSharedThread : public nixlUcxThread {
     std::vector<pollfd> pollFds_;
 };
 
-nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params)
-    : nixlUcxEngine(init_params) {
+nixlUcxThreadEngine::nixlUcxThreadEngine(const nixlBackendInitParams &init_params,
+                                         size_t num_dedicated_workers)
+    : nixlUcxEngine(init_params, num_dedicated_workers) {
+    if (!init_params.enableProgTh) {
```
rakhmets, line 368: Since it's optional now, shouldn't it be wrapper/decorator instead?

src/plugins/ucx/ucx_backend.h
iyastreb: Yes, correct, that's logical and it was my initial idea too

But it turned out that moving dedicated workers to nixlUcxThreadPoolEngine is a very hard task.
Because then we base class knows nothing about them, and we need 3 new virtual functions overridden by TPengine:
- createWorker (because we need a common context and error handling)
- connectWorkers (to connect dedicated workers on load MD)
- disconnectWorkers (because now dedicated workers are destroyed before base class dtor is called)
So after all we end up with more complex code and even more dependencies between these classes..
We can do that, but honestly I find current code easier to understand then having 3 more virtual calls.

Then I made a prototype with dependency injection of new `class nixlUcxWorkers`. But again, this comes with its own limitations and dependencies.. I don't see ideal solution for now.

So I would not this refactoring further, because after SGL is introduced we can remove TPengine altogether.
Or we can leave this refactoring for the future.
This PR already improves the behavior by unifying common logic and splitting shared/dedicated workers (and modifying progressLoop so that we don't poll dedicated workers anymore)

src/plugins/ucx/ucx_backend.cpp
iyastreb: I don't see how it can improve, so I prefer not to refactor for nothing

src/plugins/ucx/ucx_backend.h
rakhmets: I don't think that we need to move dedicate workers to `nixlUcxThreadPoolEngine`.
`nixlUcxEngine` should create the required amount of workers. Since we need to create/connect/disconnect all the workers, the code should stay in `nixlUcxEngine`, processing all the workers.
Since we shouldn't poll dedicated workers, that's seems to be the virtual method, e.g.: poll first (workers.size() - num_dedicated_workers) workers.

src/plugins/ucx/ucx_backend.cpp
rakhmets: I meant, such inheritance does not look like the best solution. Because in the absence of this option, this class is redundant in the inheritance chain. It looks more like an optional wrapper.
```C++
engine = new Engine();
if (progress_thread) {
    engine = new ProgressThreadWrapper(engine);
}
```

src/plugins/ucx/ucx_backend.h
iyastreb: But this is how it's done now
We do need to distinguish between shared and dedicated workers - this is also deducted from your comment. Let's discuss in person, this is getting too long

src/plugins/ucx/ucx_backend.cpp
iyastreb: It can be done like that, like many other things too
It will lead to massive renaming, but I don't see a clear benefit.
It is just out of scope of my PR. The goal is not to redesign the entire UCX backend, but to solve a concrete issue: remove code duplication.
Again, I don't mind if we do it one day, but it can be done in a different PR

src/plugins/ucx/ucx_backend.h
iyastreb: As agreed in person, we can keep it as is for now

brminich (dismissed):

brminich (approved):

PR #1909, opened 2026-07-08
TEST/GTEST: Fail on unexpectedly skipped tests.

ColinNV:
## What?

 - Make the gtests fail when a test is unexpectedly skipped.

Similar to how the gtests now fail when tests succeeds but prints unexpected warnings or errors.

## Why?
Some gtests that were expected to run were being skipped with no diagnostics due to `gtest-parallel` treating skipped tests as successful.

## How?

 - Add a mechanism to detect skipped tests and compare then to an "expected to skip" list.
 - Use said mechanism to fail a gtest run when a test unexpectely skips.

## Todo

 - Double-check `test/gtest/main.cpp` for correct skip conditions and test sets.
 - Move choice of defining `NIXL_CI_ALLOW_NVTX_SKIP` from `test_cpp.sh` to CI config.
 - Generally find more streamlined and homogenized way of defining test "scenarios"?


.gitlab/test_cpp.sh
```diff
@@ -105,7 +105,12 @@ kill -s INT $telePID
 # fi
 
 # shellcheck disable=SC2154
-gtest-parallel --workers=1 --serialize_test_cases ./bin/gtest -- --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port"
+gtest-parallel --output_dir=/tmp --workers=1 --serialize_test_cases ./bin/gtest -- --min-tcp-port="$min_gtest_port" --max-tcp-port="$max_gtest_port"
+echo $NIXL_CI_NON_GPU
+echo $TEST_LIBFABRIC
+echo $NIXL_CI_ALLOW_NVTX_SKIP
+grep -r Skipped /tmp/gtest-parallel-logs || true
```
ovidiusm, line 114: I assume this will be changed to print some clear message

test/gtest/meson.build
```diff
@@ -59,9 +59,11 @@ endif
 plugin_dirs_arg = '--tests_plugin_dirs=' + mocks_dep.get_variable('path')
 
 cpp_flags = []
+cuda_flags = []
 
 if cuda_dep.found()
     cpp_flags += '-DHAVE_CUDA'
+    cuda_flags += '-DHAVE_CUDA'
```
ovidiusm, line 89: This explains gtest but not the nixlbench skip on DL cluster

test/gtest/main.cpp
```diff
@@ -67,6 +72,72 @@ void ParseArguments(int argc, char **argv) {
   }
 }
 
+namespace {
```
ovidiusm, line 75: Maybe add a comment to explain for users and/or agents? Optional, up to you

ovidiusm:
LGTM except it does not cover nixlbench, but if you want it can be done separately

.gitlab/test_cpp.sh
ColinNV: That's just temporary until everything stabilizes.

PR #1913, opened 2026-07-08
nixlbench: limit Neuron core allocation in SG mode for multi-processing

fengjica:
On Neuron instances, nrt_init() claims all visible Neuron cores by default. When testing multiple nixlbench processes in SG pairwise mode (one process per device), each process tries to claim all cores, causing "Logical Neuron Core(s) not available" failures for all but the first process.

Set NEURON_RT_NUM_CORES=1 before nrt_init() in SG mode so each process only reserves one core.

MG mode is unaffected (single process per host, claims all cores).

## What?
Fix nixlbench multiprocessing with neruon runtime.

## Why?
Without the fix, nixlbench multiprocessing (SG mode) would fail.

## How?
Set NEURON_RT_NUM_CORES = 1, before nrt_init().
Tested with  trn2.48xlarge instance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Neuron runtime initialization in SG mode by automatically setting the core count when core-related environment variables aren’t provided.
  * Corrected device memory allocation to use the appropriate accelerator/VNC index for SG vs non-SG execution, improving consistency for single-GPU/SG runs.

### DIFF
--- a/benchmark/nixlbench/src/utils/neuron.cpp
+++ b/benchmark/nixlbench/src/utils/neuron.cpp
@@ -17,14 +17,20 @@
  */
 
 #include "neuron.h"
+#include "utils.h"
 
+#include <cstdlib>
 #include <dlfcn.h>
 
+#include <charconv>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
+#include <iostream>
 #include <map>
 #include <memory>
 #include <mutex>
+#include <system_error>
 #include <tuple>
 #include <vector>
 
@@ -136,6 +142,18 @@ struct NrtAllocation {
 std::map<uintptr_t, NrtAllocation> allocation_tracker;
 std::mutex allocation_tracker_mutex;
 
+// Parse NEURON_RT_NUM_CORES; return -1 if the value is not a plain integer.
+int
+parseCoreCount(const char *value) {
+    int count = 0;
+    const char *end = value + std::strlen(value);
+    auto [ptr, ec] = std::from_chars(value, end, count);
+    if (ec != std::errc{} || ptr != end) {
+        return -1;
+    }
+    return count;
+}
+
 // Find the allocation containing the given VA. Caller must hold allocation_tracker_mutex.
 // Returns {tensor, offset, alloc_size} or {nullptr, 0, 0}.
 std::tuple<nrt_tensor *, size_t, size_t>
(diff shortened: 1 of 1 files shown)


benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -142,6 +144,16 @@ getTensorFromVA(const void *va) {
 int
 neuronCoreCount() {
     static const int core_count = []() {
+        // Scope this process to the minimum number of NeuronCores it needs,
+        // analogous to cudaSetDevice() scoping a CUDA process to one GPU.
+        // Without this, nrt_init() claims ALL visible cores by default,
+        // preventing sibling processes on the same host from initializing.
+        if (!getenv("NEURON_RT_NUM_CORES") && !getenv("NEURON_RT_VISIBLE_CORES")) {
```
snadampal: since SG mode is meant for only one device per process, check `NEURON_RT_VISIBLE_CORES` set to a single core or a range of cores if it was explicitly set by user.

Even a range of visible cores would work because you are handling only the first core (index=0) in SG mode, but checking here and logging a warning will avoid any surprises.

snadampal (changes requested):

benchmark/nixlbench/src/utils/neuron.cpp
fengjica: Thanks for the review! Fixed in the next revision.

snadampal (approved):

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -142,6 +145,27 @@ getTensorFromVA(const void *va) {
...
+            const char *num_cores = getenv("NEURON_RT_NUM_CORES");
+            const char *visible_cores = getenv("NEURON_RT_VISIBLE_CORES");
+            if (!num_cores && !visible_cores) {
+                if (setenv("NEURON_RT_NUM_CORES", "1", 0) != 0) {
+                    std::cerr << "nixlbench: failed to set NEURON_RT_NUM_CORES" << std::endl;
+                    return -1;
+                }
+                std::cerr << "nixlbench: SG mode — set NEURON_RT_NUM_CORES=1" << std::endl;
+            } else {
+                std::cerr << "nixlbench: SG mode uses only the first core (vnc=0)"
```
oa-aws: This is a bit unclear. Is this an error? What if user sets env NEURON_RT_NUM_CORES=1 as needed, without setting NEURON_RT_VISIBLE_CORES (as in the if-success case)? Should a message be printed in that case as well?

fengjica: Fixed in the second commit.

fengjica:
@oa-aws Could you look again? The diff from the last revision is in the second commit 
[address comment: improved warning message condition and message](https://github.com/ai-dynamo/nixl/pull/1913/changes/16d5ac4f0023542f31a677d7031aa9595251293f)
Thanks!

oa-aws (dismissed):

benchmark/nixlbench/src/utils/neuron.cpp
```diff
@@ -142,6 +145,25 @@ getTensorFromVA(const void *va) {
...
+                setenv("NEURON_RT_NUM_CORES", "1", 0);
+                std::cerr << "nixlbench: SG mode — set NEURON_RT_NUM_CORES=1"
+                          << std::endl;
+            } else if (visible_cores) {
+                std::cerr << "nixlbench: NEURON_RT_VISIBLE_CORES is set to \""
+                          << visible_cores
+                          << "\" but SG mode uses only the first core (vnc=0)"
+                          << std::endl;
+            }
+        }
```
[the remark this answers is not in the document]
fengjica: It doesn't matter because 1) we do not run any non-pairwise SG test configuration; and 2) if we run in the future, we still want SG (single process per accelerator) to limit to one Neuron core, i.e. NEURON_RT_NUM_CORES=1 and vnc=0.

fengjica:
Squashed commits into one, and then rebased to head as there will be merge conflict.

fengjica:
Hi @aranadive @brminich @ovidiusm would you mind taking a look at this PR? The delta from the last revision where  @oa-aws approved is only that of rebasing to head and merge.
Thanks!

aranadive (approved):

PR #1914, opened 2026-07-08
telemetry: add transfer-duration histograms to Prometheus and DOCA exporters

e-eygin:
## What?

This PR adds latency histograms for transfer-time telemetry events:

- `agent_xfer_time_us` — transfer duration from start to completion.
- `agent_xfer_post_time_us` — time from start to backend post.

Both Prometheus and DOCA/CollectX exporters publish standard `_bucket`, `_sum`, and `_count` series. Existing counters and last-operation gauges remain unchanged.

## Why?

NIX-1364 requires visibility into transfer-latency distributions. Existing cumulative counters and last-value gauges cannot expose tail latency, regressions within latency bands, or the number of transfers within a range.

The histograms provide this information without adding telemetry event types or changing the telemetry buffer format.

## How?

Histogram names and help text are defined in the shared `nixlTelemetryMetricDescriptor`, keeping both exporters and parity tests consistent.

Each exporter uses its native histogram implementation:

- Prometheus: `prometheus::BuildHistogram` and `Observe`.
- DOCA/CollectX: `add_base_histogram`, `base_histogram_observe`, and `histogram_flush`.

Default buckets cover approximately 10 microseconds to 10 seconds. `NIXL_TELEMETRY_HISTOGRAM_BUCKETS_US` can override them with a comma-separated list of strictly increasing positive microsecond bounds. An absent or empty value uses the defaults; a malformed non-empty value is rejected and causes exporter initialization to fail.

DOCA observations remain cached until each concrete histogram is flushed. The exporter therefore calls `histogram_flush` for every histogram ID before `metrics_flush`.

A cross-exporter compliance test sends identical samples through both exporters, scrapes their endpoints, and compares all histogram buckets, sums, and counts after normalizing timestamps and numeric label formatting.

## Tests

- `ninja -C build test/gtest/gtest`
- `ctest --test-dir build -R gtest --output-on-failure`
- `ninja -C build test/doca-telemetry/doca_nixl_test`
- `ctest --test-dir build -R doca_nixl_test --output-on-failure`
- `ninja -C build test/doca-telemetry/histogram_parity_test`
- `ctest --test-dir build -R histogram_parity_test --output-on-failure`

## Summary by CodeRabbit

* **New Features**
  * Added histogram support for transfer-time telemetry in both Prometheus and DOCA exporters.
  * Exposed latency-distribution series for `agent_xfer_time` and `agent_xfer_post_time` as `*_bucket`, `*_sum`, and `*_count` (with `*_us` metric variants).
  * Added `NIXL_TELEMETRY_HISTOGRAM_BUCKETS_US` to configure histogram bucket upper bounds, using built-in defaults and rejecting malformed non-empty overrides.
* **Documentation**
  * Updated telemetry, Prometheus, and DOCA docs to reflect histogram outputs and bucket configuration.
* **Tests**
  * Added histogram validation (including bucket boundary checks) and cross-exporter histogram parity tests.

### DIFF
--- a/src/core/telemetry/telemetry_event.h
+++ b/src/core/telemetry/telemetry_event.h
@@ -91,6 +91,8 @@ struct nixlTelemetryMetricDescriptor {
     const char *counterHelp;
     const char *gaugeName;
     const char *gaugeHelp;
+    const char *histogramName;
+    const char *histogramHelp;
 };
 
 namespace nixlEnumStrings {
@@ -188,10 +190,10 @@ telemetryErrorStatusLabel(const nixl_telemetry_event_type_t type) noexcept {
  * @brief Exporter-side Prometheus series descriptor for a telemetry event.
  *
  * Both the native Prometheus and DOCA/CollectX exporters derive their series
- * from this single mapping, so they emit identical output. A null @c counterName
- * or @c gaugeName means the event has no cumulative counter or no last-operation
- * gauge, respectively. Error events (@c AGENT_ERR_*) and any unmapped value
- * return an all-null descriptor.
+ * from this single mapping, so they emit identical output. A null @c counterName,
+ * @c gaugeName, or @c histogramName means the event has no cumulative counter, no
+ * last-operation gauge, or no distribution histogram, respectively. Error events
+ * (@c AGENT_ERR_*) and any unmapped value return an all-null descriptor.
  *
  * @param type Telemetry event type.
  * @return Counter/gauge series names and HELP strings for @p type.
@@ -203,45 +205,63 @@ telemetryMetricDescriptor(const nixl_telemetry_event_type_t type) noexcept {
         return {"agent_tx_bytes_total",
                 "Number of bytes sent by the agent",
                 "agent_tx_last_bytes",
-                "Bytes sent by the last request"};
+                "Bytes sent by the last request",
+                nullptr,
+                nullptr};
     case nixl_telemetry_event_type_t::AGENT_RX_BYTES:
         return {"agent_rx_bytes_total",
                 "Number of bytes received by the agent",
                 "agent_rx_last_bytes",
-                "Bytes received by the last request"};
+                "Bytes received by the last request",
+                nullptr,
+                nullptr};
     case nixl_telemetry_event_type_t::AGENT_TX_REQUESTS_NUM:
         return {"agent_tx_requests_num_total",
                 "Number of requests sent by the agent",
+                nullptr,
+                nullptr,
                 nullptr,
                 nullptr};
     case nixl_telemetry_event_type_t::AGENT_RX_REQUESTS_NUM:
         return {"agent_rx_requests_num_total",
                 "Number of requests received by the agent",
                 nullptr,
+                nullptr,
+                nullptr,
                 nullptr};
     case nixl_telemetry_event_type_t::AGENT_MEMORY_REGISTERED:
         return {"agent_memory_registered_total",
                 "Cumulative memory registered",
                 "agent_memory_registered_last_bytes",
-                "Memory registered by the last operation"};
+                "Memory registered by the last operation",
+                nullptr,
+                nullptr};
     case nixl_telemetry_event_type_t::AGENT_MEMORY_DEREGISTERED:
         return {"agent_memory_deregistered_total",
                 "Cumulative memory deregistered",
                 "agent_memory_deregistered_last_bytes",
-                "Memory deregistered by the last operation"};
+                "Memory deregistered by the last operation",
+                nullptr,
+                nullptr};
     case nixl_telemetry_event_type_t::AGENT_XFER_TIME:
         return {"agent_xfer_time_total",
                 "Cumulative sum of transfer time from start to completion",
                 "agent_xfer_time",
-                "Transfer time of the last request"};
+                "Transfer time of the last request",
+                "agent_xfer_time_us",
+                "Distribution of transfer time from start to completion, in microseconds"};
     case nixl_telemetry_event_type_t::AGENT_XFER_POST_TIME:
         return {"agent_xfer_post_time_total",
                 "Cumulative sum of time from start to posting to the back-end",
                 "agent_xfer_post_time",
-                "Post time of the last request"};
+                "Post time of the last request",
+                "agent_xfer_post_time_us",
+                "Distribution of time from start to posting to the back-end, in microseconds"};
     case nixl_telemetry_event_type_t::AGENT_TELEMETRY_EVENTS_DROPPED:
         return {"agent_telemetry_events_dropped_total",
                 "Cumulative telemetry events dropped at the producer-side staging queue",
+                nullptr,
+                nullptr,
                 nullptr,
                 nullptr};
     case nixl_telemetry_event_type_t::AGENT_ERR_NOT_POSTED:
@@ -256,9 +276,9 @@ telemetryMetricDescriptor(const nixl_telemetry_event_type_t type) noexcept {
     case nixl_telemetry_event_type_t::AGENT_ERR_REMOTE_DISCONNECT:
     case nixl_telemetry_event_type_t::AGENT_ERR_CANCELED:
     case nixl_telemetry_event_type_t::AGENT_ERR_NO_TELEMETRY:
-        return {nullptr, nullptr, nullptr, nullptr};
+        return {nullptr, nullptr, nullptr, nullptr, nullptr, nullptr};
     }
-    return {nullptr, nullptr, nullptr, nullptr};
+    return {nullptr, nullptr, nullptr, nullptr, nullptr, nullptr};
 }
 } // namespace nixlEnumStrings
 
--- a/src/core/telemetry/telemetry_event.h
+++ b/src/core/telemetry/telemetry_event.h
@@ -276,7 +276,7 @@ telemetryMetricDescriptor(const nixl_telemetry_event_type_t type) noexcept {
     case nixl_telemetry_event_type_t::AGENT_ERR_REMOTE_DISCONNECT:
     case nixl_telemetry_event_type_t::AGENT_ERR_CANCELED:
     case nixl_telemetry_event_type_t::AGENT_ERR_NO_TELEMETRY:
-        return {nullptr, nullptr, nullptr, nullptr, nullptr, nullptr};
+        break;
     }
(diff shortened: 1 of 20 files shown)


src/plugins/telemetry/common/histogram_buckets.h
```diff
@@ -0,0 +1,102 @@
...
+#include <string>
+#include <vector>
+
+#include <absl/strings/ascii.h>
+#include <absl/strings/numbers.h>
+#include <absl/strings/str_split.h>
+
+namespace nixl::telemetry {
+
+inline constexpr char histogramBucketsUsVar[] = "NIXL_TELEMETRY_HISTOGRAM_BUCKETS_US";
```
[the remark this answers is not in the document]
e-eygin: camelCase matches the sibling constants (`docaPrometheusPortVar`, `prometheusPortVar`). Intentional.

e-eygin:

src/core/telemetry/telemetry_event.h
```diff
@@ -256,9 +276,9 @@ telemetryMetricDescriptor(const nixl_telemetry_event_type_t type) noexcept {
     case nixl_telemetry_event_type_t::AGENT_ERR_REMOTE_DISCONNECT:
     case nixl_telemetry_event_type_t::AGENT_ERR_CANCELED:
     case nixl_telemetry_event_type_t::AGENT_ERR_NO_TELEMETRY:
-        return {nullptr, nullptr, nullptr, nullptr};
+        return {nullptr, nullptr, nullptr, nullptr, nullptr, nullptr};
```
ColinNV: Just `break`?

src/plugins/telemetry/common/histogram_buckets.h
```diff
@@ -0,0 +1,110 @@
...
+[[nodiscard]] inline std::vector<double>
+parseHistogramBucketsUs(const std::string &spec) {
+    std::vector<double> buckets;
+    double previous = 0.0;
+    for (const absl::string_view raw : absl::StrSplit(spec, ',')) {
+        const absl::string_view token = absl::StripAsciiWhitespace(raw);
+        double value = 0.0;
+        if (!absl::SimpleAtod(token, &value) || !std::isfinite(value) || value <= 0.0 ||
+            value <= previous) {
+            return {};
```
ColinNV, line 77: Should this be a hard error? If we reach this point the user thinks they have specified some buckts, but we aren't using that specification because it's broken.

That's in line with how I generally set up the NIXL config handling, if there is no value and there is a fallback, use the fallback, but if there is a value and we can't use it, make the user notice that we can't follow their presumed intention.

src/plugins/telemetry/common/histogram_buckets.h
```diff
@@ -0,0 +1,110 @@
...
+ *         Emits a WARN only when a non-empty override was provided but is invalid.
+ */
+[[nodiscard]] inline std::vector<double>
+resolveHistogramBucketsUs() {
+    const auto spec = nixl::config::getValueOptional<std::string>(histogramBucketsUsVar);
+    if (!spec || spec->empty()) {
+        return defaultHistogramBucketsUs();
+    }
+
+    std::vector<double> buckets = parseHistogramBucketsUs(*spec);
```
ColinNV: `const`

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -162,6 +174,26 @@ DocaSharedContext::DocaSharedContext(const std::string &bind_address) {
...
+            if (descriptor.histogramName == nullptr) {
+                continue;
+            }
+            doca_telemetry_exporter_base_histogram_id_t base_id = 0;
+            result = doca_telemetry_exporter_metrics_add_base_histogram(source,
+                                                                        descriptor.histogramName,
+                                                                        histogram_buckets.data(),
+                                                                        histogram_buckets.size(),
+                                                                        label_set_id,
+                                                                        1000,
```
ColinNV: This is `// Flush interval in ms`?

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -87,6 +91,14 @@ struct DocaSharedContext {
     bool source_started = false;
     bool metrics_context_created = false;
 
+    // Base histograms are templates created once on the shared source; observing
+    // against one with concrete label values yields a concrete histogram id. The
+    // general metrics flush does not push histograms, so track those concrete ids
+    // to flush them explicitly.
+    std::unordered_map<nixl_telemetry_event_type_t, doca_telemetry_exporter_base_histogram_id_t>
```
ColinNV: Since lookup should be fast, we could exploit `nixl_telemtry_event_type_t` being small integers and use a `std::vector` for direct access instead of hashing? (Is there a good invalid value for the id, or would that need an extra `bool` or `optional`?) Since we can know the size at compile-time it might even be a `std::array` to remove all indirections.

src/plugins/telemetry/common/histogram_buckets.h
```diff
@@ -0,0 +1,110 @@
...
+ *         can fall back to the defaults (validation kept local to the telemetry plugins).
+ */
+[[nodiscard]] inline std::vector<double>
+parseHistogramBucketsUs(const std::string &spec) {
+    std::vector<double> buckets;
+    double previous = 0.0;
+    for (const absl::string_view raw : absl::StrSplit(spec, ',')) {
+        const absl::string_view token = absl::StripAsciiWhitespace(raw);
+        double value = 0.0;
+        if (!absl::SimpleAtod(token, &value) || !std::isfinite(value) || value <= 0.0 ||
```
ColinNV, line 75: Since this is parsing `double`, should the documentation clarify that non-integer values are allowed? And IIRC `absl::SimpleAtod()` ignores leading and trailing whitespace.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -261,6 +293,25 @@ nixlTelemetryDocaExporter::appendGaugeSample(const nixlTelemetryEvent &event,
...
+    if (base == ctx_->base_histograms.end()) {
+        return DOCA_SUCCESS;
+    }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    doca_telemetry_exporter_histogram_id_t histogram_id = 0;
+    const doca_error_t result = doca_telemetry_exporter_metrics_base_histogram_observe(
+        ctx_->source, base->second, label_values, static_cast<double>(event.value_), &histogram_id);
+    if (result == DOCA_SUCCESS) {
+        ctx_->histogram_ids.insert(histogram_id);
```
ColinNV, line 322: Just to make sure I understand correctly, when `doca_telemetry_exporter_metrics_base_histogram_observe()` is called with the same labels and histogram base id it will return the same histogram id, and we don't want duplicates in `histogram_ids` because otherwise we would flush histograms multiple times?

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
```diff
@@ -201,6 +224,11 @@ nixlTelemetryPrometheusExporter::exportEvent(const nixlTelemetryEvent &event) {
             gauge->second.metric->Set(static_cast<double>(event.value_));
         }
 
+        const auto histogram = histograms_.find(event.eventType_);
```
ColinNV: `const auto iter`

src/plugins/telemetry/prometheus/prometheus_exporter.h
```diff
@@ -93,13 +94,37 @@ class nixlTelemetryPrometheusExporter : public nixlTelemetryExporter {
         prometheus::Gauge *metric = nullptr;
     };
 
+    struct HistogramEntry {
+        HistogramEntry(prometheus::Family<prometheus::Histogram> *family,
```
ColinNV, line 101: `noexcept`, and can we `assert` that both are not `nullptr`, then omit the check in the d'tor? Otherwise `nullptr` looks like a legitimate use case, which it isn't.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Correct — `base_histogram_observe` returns the same concrete id for the same base+labels, and the `histogram_ids` set de-dups so `flush()` flushes each concrete histogram exactly once. Added a comment to make that explicit.

src/core/telemetry/telemetry_event.h
e-eygin: Done.

src/plugins/telemetry/common/histogram_buckets.h
e-eygin: Agreed, now throws `std::invalid_argument`; fallback only when absent/empty. Tests updated.

src/plugins/telemetry/common/histogram_buckets.h
e-eygin: Done.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Yes; named `histogramFlushIntervalMs` now.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Done — `std::array` indexed by event type.

src/plugins/telemetry/prometheus/prometheus_exporter.h
e-eygin: Done for all three entries.

src/plugins/telemetry/prometheus/prometheus_exporter.cpp
e-eygin: Renamed to `*_it`.

src/plugins/telemetry/common/histogram_buckets.h
e-eygin: Doc now says fractional allowed; dropped the redundant strip since `SimpleAtod` ignores whitespace.

src/plugins/telemetry/doca/doca_exporter.cpp
ColinNV: And we are using an `unordered_set` as opposed to a bitmask because we have no control over the ids. Got it.

ColinNV (approved):

PR #1919, opened 2026-07-09
telemetry: add per-metric activation (NIX-453)

e-eygin:
## What?

Adds **per-metric activation** for telemetry via `NIXL_TELEMETRY_METRICS` — a
comma-separated glob allowlist (POSIX `fnmatch`, e.g. `agent_tx_bytes`,
`agent_err_*`) selecting which metrics are exported.

- Resolved **once at construction** into an enum-indexed mask.
- **Gated at the producer**: `updateData()` / `addXferStats()` skip deactivated
  metrics *before the staging queue*, so a deactivated metric costs no
  mutex/append on the transfer hot path and is never counted as a drop. The
  check is a single lock-free read of the immutable mask (`isMetricEnabled`).
- **Unset/empty = all metrics active** (backward compatible); a token matching
  nothing is ignored with a warning.
- `getXferTelemetry()` is unaffected (reads the request handle, not the queue).

Docs (`docs/telemetry.md`, prometheus/doca/plugin READMEs) and tests updated in
the same PR.

## Why?

Tracked by NIX-453 (parent NIX-1423): operators need to enable/disable
individual telemetry metrics without turning telemetry off. Gating at the source
(rather than at drain) means deactivated metrics incur zero hot-path cost — per
review feedback from @AlexNvd.

This is **Phase 1** — event-type granularity, core-only, collision-free with the
in-flight histogram PR #1914. **Phase 2** (NIX-1612: per-series /
histogram-group control, `*_us`/`*_total` globs) lands after #1914.

## How?

- `telemetry_event.h`: `nixl_telemetry_event_type_count` + `nixl_telemetry_metric_mask_t`
  alias (append-only; no descriptor changes).
- `telemetry.cpp`: `resolveEnabledMetrics()` parses `NIXL_TELEMETRY_METRICS`
  (`absl::StrSplit` + `fnmatch`) into the mask; `updateData()`/`addXferStats()`
  gate on `isMetricEnabled` before staging; `flushPendingEvents()` no longer filters.
- Tests (gtest, 33/33): producer-gating cases (subset, family glob, unknown-token)
  on the BUFFER stream + an end-to-end Prometheus scrape case.

## Summary by CodeRabbit

* **New Features**
  * Added `NIXL_TELEMETRY_ENABLED_METRICS` to restrict exported telemetry using a comma-separated POSIX-glob allowlist (selects whole metric families).
* **Bug Fixes**
  * Deactivated metrics are filtered out before buffering/staging, and the “telemetry dropped” event is emitted only when enabled.
* **Tests**
  * Added Prometheus and buffer-export allowlist tests (subset, family globs, and unknown tokens). Added a disabled performance benchmark for fully deactivated per-transfer stats.
* **Documentation**
  * Updated telemetry and exporter docs to cover matching rules, warnings for unmatched tokens, and default “export all” behavior.


src/core/telemetry/telemetry_event.h
```diff
@@ -25,6 +25,7 @@
 
 constexpr char TELEMETRY_BUFFER_SIZE_VAR[] = "NIXL_TELEMETRY_BUFFER_SIZE";
 constexpr char TELEMETRY_RUN_INTERVAL_VAR[] = "NIXL_TELEMETRY_RUN_INTERVAL";
+constexpr char TELEMETRY_METRICS_VAR[] = "NIXL_TELEMETRY_METRICS";
```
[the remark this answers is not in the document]
e-eygin: Done — renamed to `nixl_telemetry_metric_mask_t` (lower_case + _t) in `9eaa5974`.

test/gtest/telemetry_prometheus_test.cpp
```diff
@@ -718,6 +718,38 @@ TEST_F(prometheusTelemetryTest, DroppedEventsCounterAccumulates) {
...
 }
 
+// End-to-end through the core (NIX-453): a per-event allowlist filters at the
+// drain, so an enabled metric advances while a disabled one stays at its
+// pre-registered 0. Families are always registered, so this asserts values, not
+// series presence (Phase 1 is event-type granularity, not per-series).
+TEST_F(prometheusTelemetryTest, MetricAllowlistFiltersAtDrain) {
+    gtest::ScopedEnv telemetry_env;
+    telemetry_env.addVar(TELEMETRY_METRICS_VAR, "agent_tx_bytes");
+    telemetry_env.addVar(TELEMETRY_RUN_INTERVAL_VAR, "5");
```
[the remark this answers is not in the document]
e-eygin: Done — run interval 5→1 ms, well under the 5 s deadline.

docs/telemetry.md
```diff
@@ -126,8 +126,10 @@ Telemetry is configured by environment variables:
...
 | `NIXL_TELEMETRY_RUN_INTERVAL` | Flush interval (ms) | `100` |
 | `NIXL_TELEMETRY_EXPORTER` | Name of the exporter plugin to use | - |
+| `NIXL_TELEMETRY_METRICS` | Comma-separated allowlist of metric names to export (glob) | all |
 
 - `NIXL_TELEMETRY_ENABLE` can be set to `y`/`yes`/`on`/`true`/`enable`/`1` to be enabled, and `n`/`no`/`off`/`false`/`disable`/`0` (or not set) to be disabled. Matching is case insensitive.
+- `NIXL_TELEMETRY_METRICS` restricts which metrics are exported. It is a comma-separated allowlist of metric names, each matched as a POSIX glob (`*`, `?`) against the base event names (`agent_tx_bytes`, `agent_rx_bytes`, `agent_tx_requests_num`, `agent_rx_requests_num`, `agent_memory_registered`, `agent_memory_deregistered`, `agent_xfer_time`, `agent_xfer_post_time`, `agent_telemetry_events_dropped`, and the `agent_err_*` errors). Unset (or empty) exports everything; a token that matches nothing is ignored with a warning. A metric name selects its whole family (counter, gauge, and histogram). Filtering happens where events are drained to the exporter, so disabled metrics are still collected in-process and remain visible via `getXferTelemetry()`.
```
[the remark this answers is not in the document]
e-eygin: Fixed — reworded to "counter and gauge"; histograms land with #1914.

e-eygin:

src/core/telemetry/telemetry_event.h
```diff
@@ -56,6 +57,12 @@ enum class nixl_telemetry_event_type_t : uint32_t {
     AGENT_TELEMETRY_EVENTS_DROPPED = 20,
 };
 
+inline constexpr std::size_t nixl_telemetry_event_type_count =
+    static_cast<std::size_t>(nixl_telemetry_event_type_t::AGENT_TELEMETRY_EVENTS_DROPPED) + 1;
+
+// Per-event-type flag mask indexed by nixl_telemetry_event_type_t.
+using nixl_telemetry_metric_mask_t = std::array<bool, nixl_telemetry_event_type_count>;
```
ColinNV, line 64: Not saying necessary to change, but did you consider `std::bitset` for a smaller representation?

src/core/telemetry/telemetry.cpp
```diff
@@ -175,6 +183,49 @@ resolveMaxBufferedEvents() {
...
+        }
+    }
+
+    for (size_t t = 0; t < tokens.size(); ++t) {
+        if (!token_matched[t]) {
+            NIXL_WARN << "Ignoring " << TELEMETRY_METRICS_VAR << " entry '" << tokens[t]
+                      << "': no telemetry metric matches";
+        }
+    }
+
```
ColinNV, line 232: Some debug log about which metrics are enabled and which are disabled?

src/core/telemetry/telemetry.cpp
```diff
@@ -260,6 +313,11 @@ nixlTelemetry::registerPeriodicTask(periodicTask &task) {
 
 void
 nixlTelemetry::updateData(nixl_telemetry_event_type_t event_type, uint64_t value) {
+    // Skip deactivated metrics before taking the lock: no staging cost, and a
+    // deactivated metric is not a drop.
+    if (!isMetricEnabled(event_type)) {
```
ColinNV, line 329: Since this is a fast path, should the check be in the `.h` so that it can be inlined at call sites? Might also be exaggerated...

src/core/telemetry/telemetry.cpp
e-eygin: Added in `68c3aabd` — logs active/inactive metrics at DEBUG when the allowlist is set.

src/core/telemetry/telemetry_event.h
e-eygin: Considered — kept` std::array<bool>` so the check is a plain byte load (bitset adds shift/mask per read); ~13 B/agent isn't worth it on the hot path.

src/core/telemetry/telemetry.cpp
e-eygin: `isMetricEnabled` is already an inline member in `telemetry.h`, so it inlines into `updateData`/`addXferStats`; the public entry points stay out-of-line since they take the mutex.

src/core/telemetry/telemetry_event.h
ColinNV: `NIXL_TELEMETRY_ENABLED_METRICS`?

Makes the intent clearer, especially if we decide to add a `DISABLED` switch in the future to selectively disable some metrics.

e-eygin: Renamed to `NIXL_TELEMETRY_ENABLED_METRICS`. Thanks!

test/gtest/telemetry_test.cpp
```diff
@@ -331,6 +331,82 @@ TEST(telemetryMetricContract, DescriptorIsUnifiedExporterSeriesContract) {
     }
 }
 
+// A subset allowlist exports only the listed event; deactivated metrics are
+// skipped at the source and never enter the staging queue (BUFFER here).
+TEST_F(telemetryTest, MetricAllowlistSubset) {
+    envHelper_.addVar(TELEMETRY_RUN_INTERVAL_VAR, "1");
+    envHelper_.addVar(TELEMETRY_ENABLED_METRICS_VAR, "agent_tx_bytes");
```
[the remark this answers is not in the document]
e-eygin: `ScopedEnv` is already RAII — `Variable::~Variable` restores (or unsets) each override, and `m_vars` unwinds when the per-test `envHelper_` is destroyed. So even if an `ASSERT` skips a manual `popVar`, the fixture teardown restores everything; no stale config leaks to later tests. The manual pops just keep the intra-test stack balanced, matching the pattern used throughout this file. Leaving as-is.

ColinNV (approved):

PR #1920, opened 2026-07-09
telemetry: native multi-process Prometheus aggregation (prometheus_mp)

e-eygin:
## What?

Adds `prometheus_mp`, a native telemetry exporter plug-in that exposes the telemetry of **all** processes of a multi-process NIXL run (TP/DP/PP) behind a **single** Prometheus scrape endpoint — with **no external dependency** (no DOCA/DTS, not built on Python's `prometheus_client`).

Enable with `NIXL_TELEMETRY_EXPORTER=prometheus_mp` and point every rank at one shared directory via `NIXL_TELEMETRY_MULTIPROC_DIR`.

## Why?

The existing native Prometheus exporter binds one scrape port per process. Under multi-process parallelism every rank creates its own agent and tries to bind the **same** port; only one wins, so every other rank's telemetry is lost by design (reported for a Dynamo + SGLang TP=8 deployment, issue ai-dynamo/nixl#1838). The existing exporter can only make that collision *benign* (WARN and continue) — it can't recover the lost ranks.

This makes **every** rank's telemetry available behind one endpoint, natively — no external aggregator (the users run the native exporter specifically to avoid extra infrastructure) and no parent process to coordinate (NIXL is loaded independently per rank, so ownership is self-elected at runtime). It complements, and does not replace, the DOCA/CollectX exporter (which aggregates via an external service).

Tracking: [#1838](<https://github.com/ai-dynamo/nixl/issues/1838>) (ai-dynamo/nixl#1838 / [NIX-1614](https://linear.app/nvidia/issue/NIX-1614/native-multi-process-prometheus-aggregation-doca-free-single-endpoint)).

## How?

Every process writes its own metric state to a fixed-slot, memory-mapped file in the shared dir. One of them is elected to serve: the processes race for an exclusive `flock` on a lock file named after the address they were configured to serve, and only the winner (**owner**) binds the scrape port, running the `Exposer` plus a custom `prometheus::Collectable` that, on each scrape, reads a snapshot of every live peer and republishes it. The losers run **writer-only** and never bind, so losing is benign and no rank is dropped. Each process is its own series (no cross-process summing, so counters stay monotonic).

The owner is not a single point of failure: the kernel frees the lock when it dies, and a writer re-running the election takes the endpoint — and the reaping — over.

The exported metric set is at full parity with the single-process `prometheus` exporter: counters, last-operation gauges, error counters, and the transfer-duration histograms.

<details><summary>Design detail (election, crash-safety, histograms, labels, cleanup, config, limitations, tests)</summary>

### Owner election & failover

The lock, not the bind, is what elects: two ranks binding concurrently cannot tell which of them got there first, so gating the bind on an exclusive lock is what makes exactly one process serve. That guarantee holds as long as the lock is usable — if the lock file cannot be opened, is not a regular file owned by the run's user, or sits on a filesystem without `flock`, every process warns and falls back to the port bind deciding.

Naming the lock file after the address keeps it contentless and scopes the election to the ranks that would actually collide, which turns two otherwise silent misconfigurations into warnings. An owner that cannot bind reports the port as held from outside the run (a foreign service, or a rank pointed at a different directory) and concedes the election rather than holding it, so the next rank to win takes the address over once the port frees. A directory served on more than one address means the ranks disagree on the port; each serves what it was configured with, but every one of them exports every rank, so a Prometheus scraping more than one target sees the same series twice. Owners find each other by trying the directory's other lock files: one that can be locked is a leftover from an earlier run, one that cannot is a live second owner.

A process that is not serving re-runs the election as it exports, throttled to a few times a second, and binds if it now wins. The endpoint is therefore unreachable for that gap plus up to one scrape interval when the owner dies, rather than for the rest of the run. Two consequences: a rank that exports nothing never re-elects, so a run that goes fully idle at the wrong moment stays down until any rank produces telemetry again; and when the port is held from outside the run, the retry backs off to every few seconds instead of hammering a bind that cannot succeed. Alert on the scrape target's `up` metric rather than on absent series.

### Crash- & race-safety

The store is a raw mmap of a fixed POD layout, so it must survive concurrent readers, processes appearing mid-scrape, and processes killed mid-update:

* **Hot path writes only the numeric value**, as a single 8-byte aligned `__atomic` op — `SIGKILL` can't tear it; a process killed mid-batch just leaves the store a few increments behind (metrics are snapshots, not a ledger).
* **Identity is fixed & positional** — metric names are never stored (the collector supplies them); slots are indexed by event type. No variable-length key/label on the hot path.
* **Per-process labels are written once** at startup, and the store is not linked into the shared directory until it is initialized *and* locked — so a reader never finds a half-built store in the first place. The single atomic `magic`, published last, still guards the staging path used on filesystems without `O_TMPFILE`.
* **A store that can't be read is not assumed abandoned.** Only genuinely invalid content (bad/zero magic, wrong schema, truncated) *whose lock nobody holds* is reapable; a transient `open`/`mmap` failure leaves a live peer's store untouched.
* `Collect()` **cannot take down the process.** prometheus-cpp calls it on its HTTP handler thread, so directory iteration is non-throwing and the body is wrapped in a catch-all that degrades to an empty scrape.
* **The on-disk layout is a contract**: it lives in its own header with a `static_assert` on its size, so a reordered field or a changed cap fails the build instead of shifting offsets under a peer that still validates the header.

### Duration histograms

`agent_xfer_time_us` / `agent_xfer_post_time_us` are exported per process as the usual `_bucket{le="..."}` / `_sum` / `_count` series, with the same names and default bounds as the single-process exporter. Keeping them in a fixed mmap layout has two consequences:

* **Bucket counts are stored non-cumulative** and accumulated at scrape time. Cumulative slots would let a reader racing a writer observe a non-monotonic histogram, which Prometheus treats as malformed; deriving both the cumulative sequence and `sample_count` from one read makes consistency structural rather than timing-dependent.
* **Bucket bounds travel in each store file**, because every process resolves `NIXL_TELEMETRY_HISTOGRAM_BUCKETS_US` from its own environment and the collector cannot know what a peer was configured with. Give every rank the same value, or the family ends up with series carrying different `le` sets. A fixed layout also has to cap the list: an override longer than 32 bounds is rejected at construction rather than silently truncated — the only behavioural difference from the single-process exporter.

### Series & labels

Labeled `hostname`, `agent_name`, `pid` (cross-process uniqueness; deliberately not the reserved `instance`), `agent_instance` (distinguishes multiple same-name agents in one process; `0` in the common case), and optional `local_rank` (local/per-GPU rank, only when the rank env is set). `agent_errors_total` additionally carries the bounded `status` label. Names/types/semantics are identical to the single-process `prometheus` exporter.

### Cleanup & lifecycle

A departing process leaves its store behind, cleanly or not: its last values are usually unscraped, and unlinking on exit would drop everything recorded since the previous scrape. The **owner** reaps lazily during `Collect()`, and liveness is a lock rather than a pid — each process `flock`s its store before the file has a name (created nameless with `O_TMPFILE`, linked into the directory once initialized) and holds it for its lifetime, so a store the owner can lock has no writer left and will never change again. Such a store is published until its last update ages past the TTL, then reaped: the same path for a clean exit and a `SIGKILL`, with no pids, no `/proc`, and no shared PID namespace. Whole-run cleanup is left to the deployment (e.g. a per-pod `emptyDir`).

### Configuration

| Variable | Description | Default |
| -- | -- | -- |
| `NIXL_TELEMETRY_EXPORTER=prometheus_mp` | Select this exporter |  |
| `NIXL_TELEMETRY_MULTIPROC_DIR` | Shared **local** folder, same for all ranks (required; use tmpfs/`/dev/shm` to stay in RAM) |  |
| `NIXL_TELEMETRY_PROMETHEUS_PORT` / `_LOCAL` | Scrape port / bind scope | 9090 / public |
| `NIXL_TELEMETRY_RANK_ENV` | Env var holding the rank for the optional `local_rank` label | `LOCAL_RANK` |
| `NIXL_TELEMETRY_MP_STALE_TTL` | Seconds after a departed process's last update before its store is reaped | 30 |
| `NIXL_TELEMETRY_HISTOGRAM_BUCKETS_US` | Histogram bucket bounds, shared with the other exporters; capped at 32 bounds here | built-in µs defaults |

Must be a local filesystem (mmap `MAP_SHARED` coherence) — **not** NFS. Ranks must also agree on a host-wide `CLOCK_MONOTONIC` (a shared time namespace); there is no PID namespace requirement, since liveness is the store's own lock. A missing directory is created `0700`; an existing one is left alone, with a warning when it is group- or world-writable or owned by another user, and store files planted by another user are ignored rather than read. Unlike a bind collision, a *configuration* error is fatal: a missing `NIXL_TELEMETRY_MULTIPROC_DIR`, or a bucket override longer than 32 bounds, fails `nixlAgent` construction rather than degrading.

### Scope / limitations (intentional)

Purpose-built for NIXL's fixed, low-cardinality, per-process label model — it can't represent a metric with a dynamic/high-cardinality label that varies per observation (none exist today; that would need a keyed store). The `pid`/`agent_instance` labels that keep per-process counters monotonic also make a restarted rank a new series, so a crash-looping deployment grows TSDB cardinality at the restart rate (scrape size still follows the live process count).

### Testing

Runs under CTest/meson: 46 unit tests over the store, the collector and the exporter — including histogram bucket-edge semantics, cumulative/`+Inf` emission, rejection of an over-long bucket override, each election outcome, a writer promoting itself when the owner exits, the two invariants that keep failover safe (re-electing while already serving cannot drop the held lock, and a conceded election really does free the endpoint), and the lock-liveness contract (a writer holds its own store and releases it on destruction, a live writer outlives a TTL that expires everything, and a departed writer's final values are published once more before its file is reaped) — plus a forking multi-process e2e that scrapes the owner over HTTP and asserts all ranks aggregate, the histogram `_bucket`/`_sum`/`_count` series are served, and a killed rank is dropped and reaped. All `prometheus_mp` tests pass; changed lines are clang-format-19 clean; every commit is DCO-signed.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added experimental `prometheus_mp` multiprocess Prometheus exporter that exposes a single `/metrics` endpoint and aggregates per-process telemetry.
  * Added multi-process scrape ownership election, per-process store sharing, and labels to disambiguate series (`pid`, `agent_instance`, optional `local_rank`).
* **Documentation**
  * Expanded guidance for multi-process aggregation, stale TTL/reaping behavior, and histogram parity/limits (including a 32-bucket cap).
* **Bug Fixes**
  * Standardized the shared `agent_errors_total` metric family naming/help across Prometheus exporters.
  * Standardized hostname handling across telemetry exporters.
* **Tests**
  * Added unit and end-to-end tests covering the multiprocess store, collector semantics (including histograms), and scrape ownership behavior.
* **Chores**
  * Improved test setup to avoid repeated plugin registration issues.


e-eygin:
@coderabbitai help

e-eygin:
@coderabbitai resume

src/plugins/telemetry/prometheus_mp/mp_collector.cpp
```diff
@@ -0,0 +1,237 @@
...
+            auto labels = baseLabels(snap);
+            labels.push_back({"status", nixlEnumStrings::telemetryErrorStatusLabel(type)});
+            errors.metric.push_back(
+                counterMetric(std::move(labels), snap.counters[static_cast<std::size_t>(type)]));
+        }
+    }
+    families.push_back(std::move(errors));
+
+    return families;
+}
```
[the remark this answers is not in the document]
e-eygin: Compile-time-safe as of `3bf21390`: `mp_store.h` `static_asserts maxTelemetrySlot() < MP_STORE_SLOT_COUNT` over the same event arrays, so `operator[]` can't go out of bounds. No runtime `.at()` needed.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,163 @@
...
+constexpr char defaultRankEnvName[] = "LOCAL_RANK";
+
+constexpr char prometheusPortVar[] = "NIXL_TELEMETRY_PROMETHEUS_PORT";
+constexpr char prometheusLocalVar[] = "NIXL_TELEMETRY_PROMETHEUS_LOCAL";
+constexpr char multiprocDirVar[] = "NIXL_TELEMETRY_MULTIPROC_DIR";
+constexpr char rankEnvVar[] = "NIXL_TELEMETRY_RANK_ENV";
+constexpr char staleTtlVar[] = "NIXL_TELEMETRY_MP_STALE_TTL";
+
+const std::string localAddress = "127.0.0.1";
+const std::string publicAddress = "0.0.0.0";
```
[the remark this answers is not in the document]
e-eygin: Intentional: `char[]` converts implicitly for `getValueDefaulted(const std::string&)` / `getenv(const char*)`; `string_view` would force casts. Keeping as-is.

e-eygin:

src/plugins/telemetry/prometheus_mp/mp_store.h
```diff
@@ -0,0 +1,195 @@
...
+              "keep AGENT_TELEMETRY_EVENTS_DROPPED last or update MP_STORE_SLOT_COUNT");
+
+/**
+ * @brief A point-in-time copy of one process's metric-state store file.
+ *
+ * Produced by readStoreSnapshot(). Values are plain (already loaded from the
+ * shared file); @c counters are cumulative running totals and @c gauges hold the
+ * last-operation value, both indexed by nixl_telemetry_event_type_t.
+ */
+struct mpStoreSnapshot {
```
ColinNV: The `mp` prefix feels superfluous in namespace `nixl::telemetry::mp`.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+        // Zeroed header: either a store still being initialized by a live process,
+        // or an orphan left by a process that died mid-creation. Skip quietly (no
+        // WARN); the collector reaps stale orphans by file age.
+        ::munmap(mapping, sizeof(mpStoreLayout));
+        return std::nullopt;
+    }
+    if (magic != MP_STORE_MAGIC) {
+        NIXL_WARN << "prometheus_mp: ignoring telemetry store '" << path.string()
+                  << "' with bad magic";
+        ::munmap(mapping, sizeof(mpStoreLayout));
```
ColinNV: The multiple calls to `munmap` are hinting at RAII-style management.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+        mapping_ = nullptr;
+    }
+    std::error_code ec;
+    std::filesystem::remove(path_, ec);
+}
+
+void
+mpStoreWriter::touch() noexcept {
+    auto *layout = static_cast<mpStoreLayout *>(mapping_);
+    __atomic_store_n(&layout->lastUpdateNs, nixlTime::getNs(), __ATOMIC_RELEASE);
```
ColinNV: What does the `release` refer to, are there any value that can have changed without their own respective atomic operation?

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+    ::close(fd);
+    if (mapping == MAP_FAILED) {
+        NIXL_WARN << "prometheus_mp: cannot map telemetry store '" << path.string()
+                  << "': " << std::strerror(errno);
+        return std::nullopt;
+    }
+
+    const auto *layout = static_cast<const mpStoreLayout *>(mapping);
+
+    const uint64_t magic = __atomic_load_n(&layout->magic, __ATOMIC_ACQUIRE);
```
ColinNV, line 285: `relaxed`

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+    std::memset(layout, 0, mappingSize_);
+    layout->schemaVersion = MP_STORE_SCHEMA_VERSION;
+    layout->slotCount = static_cast<uint32_t>(MP_STORE_SLOT_COUNT);
+    layout->pid = static_cast<int64_t>(::getpid());
+    layout->startTime = readProcessStartTime(layout->pid);
+    layout->instance = instance;
+    copyField(layout->agentName, MP_MAX_AGENT_NAME, agent_name, "agent name");
+    copyField(layout->hostname, MP_MAX_HOSTNAME, hostname, "hostname");
+    copyField(layout->localRank, MP_MAX_LOCAL_RANK, local_rank, "local_rank");
+    __atomic_store_n(&layout->lastUpdateNs, nixlTime::getNs(), __ATOMIC_RELEASE);
```
ColinNV: Only one release is needed.

src/plugins/telemetry/prometheus_mp/mp_collector.h
```diff
@@ -0,0 +1,99 @@
...
+     */
+    explicit nixlMultiprocessCollector(std::filesystem::path dir,
+                                       std::chrono::nanoseconds stale_ttl = MP_DEFAULT_STALE_TTL,
+                                       bool reap_stale = true);
+
+    [[nodiscard]] std::vector<prometheus::MetricFamily>
+    Collect() const override;
+
+private:
+    std::filesystem::path dir_;
```
ColinNV: All `const`?

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+    constexpr uint64_t MP_STORE_MAGIC = 0x3153504d4c58494eULL;
+
+    constexpr std::size_t MP_MAX_AGENT_NAME = 256;
+    constexpr std::size_t MP_MAX_HOSTNAME = 128;
+    constexpr std::size_t MP_MAX_LOCAL_RANK = 64;
+
+    // Fixed on-disk layout. Plain trivially-copyable POD operated on with __atomic
+    // builtins (not std::atomic) so it is safe to memset/reinterpret over an mmap'd
+    // region shared between processes. Field order keeps every uint64 8-byte aligned.
+    struct mpStoreLayout {
```
ColinNV: Perhaps something to indicate whether a writer has shut down in orderly fashion? Might be useful for diagnostics... (Just like a "number updates"/generation counter, but that has an overhead.)

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,275 @@
...
+    __atomic_store_n(&layout->magic, MP_STORE_MAGIC, __ATOMIC_RELEASE);
+}
+
+mpStoreWriter::~mpStoreWriter() {
+    if (mapping_ != nullptr) {
+        ::munmap(mapping_, mappingSize_);
+        mapping_ = nullptr;
+    }
+    std::error_code ec;
+    std::filesystem::remove(path_, ec);
```
ColinNV: It's not a problem that we are losing the final state?

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+// civetweb reports a failed port bind with this exact text (as used by the
+// single-process prometheus exporter). Only this case is treated as a benign
+// bind collision; any other Exposer failure is a genuine error.
+constexpr char bindFailureMarker[] = "Failed to setup server ports";
+
+[[nodiscard]] std::string
+getHostname() {
```
ColinNV: That must be the third implementation of this function in telemetry. Perhaps one central one that returns an optional?

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,164 @@
...
+// the rank. Empty when the named env var is unset -- rank is a best-effort label
+// only. This is the local/per-GPU (TP) rank, distinct from Dynamo's dp_rank.
+[[nodiscard]] std::string
+resolveLocalRank() {
+    const std::string rank_source =
+        nixl::config::getValueDefaulted<std::string>(rankEnvVar, defaultRankEnvName);
+    if (rank_source.empty()) {
+        return {};
+    }
+    const char *value = std::getenv(rank_source.c_str());
```
ColinNV: Why the double lookup, and why does the inner lookup not also use `nixl::config` to keep things homogeneous, environment variables win against config files.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.h
```diff
@@ -0,0 +1,64 @@
...
+    explicit nixlTelemetryPrometheusMpExporter(const nixlTelemetryExporterInitParams &init_params);
+    ~nixlTelemetryPrometheusMpExporter() override;
+
+    nixl_status_t
+    exportEvent(const nixlTelemetryEvent &event) override;
+
+    // True if this process won the bind race and serves the scrape endpoint.
+    [[nodiscard]] bool
+    isExporter() const noexcept {
+        return owner_;
```
ColinNV: `bool( exposer_ )` and remove `owner_`?

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.h
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+private:
+    // Declared so destruction is exposer_ -> collector_ -> store_: stop serving
+    // before dropping the collector it weak-references, then unmap the store.
+    std::unique_ptr<nixl::telemetry::mp::mpStoreWriter> store_;
+    std::shared_ptr<nixl::telemetry::mp::nixlMultiprocessCollector> collector_;
```
ColinNV: Looks like `collector_` and `exposer_` can be `std::shared_ptr< void >`.

src/plugins/telemetry/prometheus_mp/mp_collector.h
e-eygin: Done — `dir_`/`staleTtl_`/`reapStale_` are now `const`.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Clean shutdown already self-signals: the writer removes its own store file, so its absence is the "shut down cleanly" marker. A generation counter would add hot-path cost for little extra signal — left out for now.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Fixed — dropped to `RELAXED`; the `magic` store is the sole publish barrier.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
e-eygin: Done — added `nixl::getHostname()` in `common/hostname.h` returning `std::optional<std::string>`; prometheus, prometheus_mp, and DOCA all use it now.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.h
e-eygin: Done — dropped `owner_`; `isExporter() `returns `static_cast<bool>(exposer_)`.

src/plugins/telemetry/prometheus_mp/mp_store.h
e-eygin: Done — renamed `mpStoreSnapshot`/`mpStoreWriter`/`mpStoreLayout` → `storeSnapshot`/`storeWriter`/`storeLayout`.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Done — replaced the per-return `munmap` with a `unique_ptr` guard.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Intentional: on clean shutdown the producer is gone and removes its store; its values were already scraped while alive and the series goes stale. A crash instead leaves the file for the owner's liveness/TTL reaping.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
e-eygin: Inner lookup now goes through `nixl::config`. The two-step is intentional indirection — `NIXL_TELEMETRY_RANK_ENV` names _which_ env var holds the rank — not a duplicate read.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.h
e-eygin: Kept them typed: they're only kept-alive, so `shared_ptr<void>` would just trade a header include for lost intent. Can change if header decoupling is a priority.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Kept `ACQUIRE` here on purpose: it pairs with the ctor's `magic` `RELEASE` to make the once-written header fields visible before we read them just below; relaxing would race those reads. (162/180 are now `RELAXED`.)

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Right — nothing to publish (counters/gauges are `RELAXED` both sides). Changed to `RELAXED`.

e-eygin:
@coderabbitai resume

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    const int lock_fd = ::open((dir / ownerLockName).c_str(), O_CREAT | O_RDWR | O_CLOEXEC, 0600);
+    const bool holds_owner_lock = lock_fd >= 0 && ::flock(lock_fd, LOCK_EX | LOCK_NB) == 0;
+
+    try {
+        auto exposer = std::make_shared<prometheus::Exposer>(bind_address);
+        collector_ = std::make_shared<nixlMultiprocessCollector>(dir, resolveStaleTtl());
+        exposer->RegisterCollectable(collector_);
+        exposer_ = std::move(exposer);
+        ownerLockFd_ = lock_fd;
```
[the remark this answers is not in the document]
e-eygin: Fixed in b7dbc34e. Correct diagnosis: lock winner and bind winner are
independent, so both warnings misfire on simultaneous startup.

The lock is now the election. Only its winner binds, losers never touch the
port, so several owners are impossible rather than merely reported. The winner
records its bound endpoint in the lock file, so a loser configured for a
different one still warns, and a winner that cannot bind knows the port is held
outside the run.

Shutdown ordering is structural: the lock lives in an `ownerElection` member
declared before `store_`/`collector_`/`exposer_`, so it is released last and the
destructor is back to `= default`. The descriptor is RAII (`scopedFd`, promoted
out of `mp_store.cpp`), so it is also closed if the constructor throws.

No concurrent-election test: a loser returns before reaching `Exposer`, so there
is no interleaving left to exercise. `WriterModeWhenPortTaken`,
`WriterBehindSiblingOwnerIsQuiet` and the forked `MpE2ETest` cover the
observable contract.

src/plugins/telemetry/prometheus_mp/mp_store_layout.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace nixl::telemetry::mp {
+
+// "NIXLMPS1" as a little-endian tag; changing the layout must change either this
+// or MP_STORE_SCHEMA_VERSION so stale-format files are rejected.
+inline constexpr uint64_t MP_STORE_MAGIC = 0x3153504d4c58494eULL;
+
+inline constexpr std::size_t MP_MAX_AGENT_NAME = 256;
+inline constexpr std::size_t MP_MAX_HOSTNAME = 128;
+inline constexpr std::size_t MP_MAX_LOCAL_RANK = 64;
```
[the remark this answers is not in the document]
e-eygin: Not taken. `docs/CodeStyle.md` does say constants are snake_case, but `.clang-tidy` — the enforced gate — constrains only enum constants, and the sibling single-process exporter uses `prometheusPortVar` / `prometheusExporterDefaultPort`. Renaming only the constants this PR happens to touch would leave the plugin inconsistent with its sibling and with the `MP_STORE_*` constants in `mp_store.h`, which moved into `mp_store_layout.h` verbatim. Worth a repo-wide pass, not a partial one here.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return true;
+    }
+    catch (const std::exception &e) {
+        if (std::string(e.what()).find(bindFailureMarker) == std::string::npos) {
+            throw;
+        }
+        // Hold the election and the port stays unreachable for the whole run,
+        // since no other rank can then be elected to try it.
+        election_.release();
+        return false;
```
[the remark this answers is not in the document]
e-eygin: Reworded in `33251dec`. It was a rationale for why we concede, written as if it described the code.

test/gtest/telemetry_mp_collector_test.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+    auto dead_fresh = makeSnap("b", "");
+    dead_fresh.pid = 0x7fffffff;
+    dead_fresh.lastUpdateNs = monotonicNs();
+    EXPECT_TRUE(isSnapshotLive(dead_fresh, ttl));
+
+    auto dead_stale = makeSnap("c", "");
+    dead_stale.pid = 0x7fffffff;
+    dead_stale.lastUpdateNs =
+        monotonicNs() - std::chrono::nanoseconds(std::chrono::seconds(60)).count();
```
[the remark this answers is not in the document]
e-eygin: Fixed in `33251dec`: the snapshot now uses a zero heartbeat against a zero TTL, which is stale on any host. As written it passed on a freshly booted one for the wrong reason — the wrapped timestamp read as being in the future, not as expired.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // outside the run rather than by a rank of it, which is a condition that
+    // lasts. Retrying an Exposer five times a second against it is the one costly
+    // case, so back off; a transient conflict then clears in seconds instead of
+    // milliseconds.
+    constexpr uint64_t backoffNs = 5000000000ULL;
+
+    // civetweb reports a failed port bind with this exact text (as used by the
+    // single-process prometheus exporter). Only this case is treated as a benign
+    // bind collision; any other Exposer failure is a genuine error.
+    constexpr char bindFailureMarker[] = "Failed to setup server ports";
```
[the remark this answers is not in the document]
e-eygin: Not taken. `docs/CodeStyle.md` does say constants are snake_case, but `.clang-tidy` — the enforced gate — constrains only enum constants, and the sibling single-process exporter uses `prometheusPortVar` / `prometheusExporterDefaultPort`. Renaming only the constants this PR happens to touch would leave the plugin inconsistent with its sibling and with the `MP_STORE_*` constants in `mp_store.h`, which moved into `mp_store_layout.h` verbatim. Worth a repo-wide pass, not a partial one here.

e-eygin:
@coderabbitai resume

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
ColinNV: Yes, I was wondering why the inner environment variable name itself needs to be configured?

src/plugins/telemetry/prometheus_mp/mp_store.cpp
ColinNV: How is it ensured that the final values were scraped at the time the file is removed?

e-eygin: It isn't — the store can't know a scrape happened. A clean shutdown loses whatever came after the last scrape, exactly as the single-process exporter does when its endpoint dies with the process.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
e-eygin: Launchers name it differently: `LOCAL_RANK` (torchrun, our default), `OMPI_COMM_WORLD_LOCAL_RANK`, `MPI_LOCALRANKID`, `SLURM_LOCALID`. Hardcoding one drops the label everywhere else, and probing a list guesses wrong when two are set and disagree (torchrun under Slurm). One indirection var covers any launcher without a NIXL change.

src/utils/common/hostname.h
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+#ifndef NIXL_SRC_UTILS_COMMON_HOSTNAME_H
+#define NIXL_SRC_UTILS_COMMON_HOSTNAME_H
+
+#include <array>
+#include <optional>
+#include <string>
+
+#include <limits.h>
```
ColinNV, line 24: Not needed?

src/utils/common/scoped_fd.h
```diff
@@ -0,0 +1,82 @@
...
+namespace nixl::telemetry::mp {
+
+/**
+ * @class scopedFd
+ * @brief Owning wrapper over a file descriptor, closed on destruction.
+ *
+ * Held as a member, it also closes the descriptor when a constructor throws
+ * after acquiring it.
+ */
+class scopedFd {
```
ColinNV, line 34: Perhaps this should also go into `utils`, #1930 has a near identical implementation.

src/plugins/telemetry/prometheus_mp/owner_election.h
```diff
@@ -0,0 +1,206 @@
...
+ * @brief Picks the single process of a telemetry directory allowed to serve the
+ *        scrape endpoint.
+ *
+ * The election is an flock rather than the port bind itself: two processes
+ * binding concurrently cannot tell which of them got there first, whereas the
+ * lock admits exactly one, so only the winner ever binds. The winner then
+ * publishes the endpoint it bound, which is how a loser can tell that it was
+ * configured for a different one. The kernel releases the lock when the holder
+ * dies, so it needs no cleanup -- and a writer that re-runs the election then
+ * wins it, which is how the endpoint survives the owner's death.
```
ColinNV: It doesn't survive as much as being resurrected by the new winner?

src/plugins/telemetry/prometheus_mp/owner_election.h
```diff
@@ -0,0 +1,206 @@
...
+     * since flock contends between two open file descriptions of the same
+     * process, so moving it in would close the descriptor that holds the lock
+     * and leave a process serving while holding nothing. Release first if
+     * giving the lock up is what was meant. Only the lock is protected: a loser
+     * keeps its descriptor open too, and re-running its election is the whole
+     * point of the retry.
+     */
+    ownerElection &
+    operator=(ownerElection &&other) noexcept {
+        if (holdsLock_) {
```
ColinNV: This is not intuitive, and `scrapeEndpoint::reclaim()` should never trigger this. Perhaps the move-constructor and move-assignment-operator should be removed and `scrapeEndpoint` should hold a `std::optional< ownerElection >`, then `reclaim()`, if not serving, can clear the optional and perhaps be partly implemented in terms of `claim()` with suppressed warnings. Or `claim()` leaves the optional empty when not serving? Would also remove the default c'tor.

src/plugins/telemetry/prometheus_mp/owner_election.h
```diff
@@ -0,0 +1,206 @@
...
+     * @param endpoint The bound "address:port".
+     *
+     * Best-effort: it only decides whether a loser warns.
+     */
+    void
+    publishEndpoint(const std::string &endpoint) const {
+        if (!fd_.valid()) {
+            return;
+        }
+        if (::ftruncate(fd_.get(), 0) != 0 ||
```
ColinNV: Is this ever needed, beyond the truncate in the c'tor?

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
```diff
@@ -0,0 +1,121 @@
...
+    scrapeEndpoint(std::filesystem::path dir,
+                   std::string bind_address,
+                   std::chrono::nanoseconds stale_ttl);
+
+    /**
+     * @brief Runs the election once and serves the endpoint if it wins.
+     * @throw std::exception if the exposer fails for any reason other than the
+     *        port being taken -- a genuine error, unlike a bind collision.
+     */
+    status
```
ColinNV: `[[nodiscard]]`?

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
```diff
@@ -0,0 +1,121 @@
...
+private:
+    // Binds and starts aggregating; false when the port is held elsewhere, in
+    // which case the election is conceded rather than held for the whole run.
+    [[nodiscard]] bool
+    serve();
+
+    std::filesystem::path dir_;
+    std::string bindAddress_;
+    std::chrono::nanoseconds staleTtl_;
+    uint64_t lastAttemptNs_ = 0;
```
ColinNV: Can we not consistently use `std::chrono` types, and nanoseconds resolution is not needed here...?

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
```diff
@@ -0,0 +1,137 @@
...
+        election_.publishEndpoint(bindAddress_);
+        return true;
+    }
+    catch (const std::exception &e) {
+        if (std::string(e.what()).find(bindFailureMarker) == std::string::npos) {
+            throw;
+        }
+        // Concede: holding an election this process cannot act on would keep
+        // every other rank from trying the port for the rest of the run.
+        election_.release();
```
ColinNV: Does `e.what()` need logging here?

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,212 @@
...
+        store_->setGauge(type, event.value_);
+    }
+    if (descriptor.histogramName != nullptr) {
+        store_->observeHistogram(type, event.value_);
+    }
+    // Once per event, not once per slot updated: a duration event touches three
+    // slots, and the clock read costs several times the atomics it would follow.
+    const uint64_t now = store_->refreshHeartbeat();
+    if (!endpoint_.serving()) {
+        endpoint_.reclaim(now);
```
ColinNV: Is this not still too often?

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
```diff
@@ -0,0 +1,212 @@
...
+    case scrapeEndpoint::status::SERVING:
+        NIXL_INFO << "prometheus_mp exporter (owner) serving " << bind_address
+                  << ", aggregating telemetry dir " << dir_.string();
+        return;
+
+    case scrapeEndpoint::status::PORT_TAKEN:
+        // Elected, so no sibling can be serving: the port belongs to something
+        // outside this run and nothing will aggregate this directory. The
+        // election is conceded rather than held, so once the port frees -- a
+        // conflict as short as a previous run still shutting down -- the next
```
ColinNV: Would a previous run still shutting down not be detected as `SIBLING_OWNS`?

src/utils/common/hostname.h
e-eygin: Needed — HOST_NAME_MAX comes from <limits.h>. It's often pulled in transitively via <unistd.h>, but that isn't guaranteed.

src/utils/common/scoped_fd.h
e-eygin: Agreed. #1930 is still open, so there's nothing to converge on yet — whichever of us lands second should move it to utils and drop the duplicate. Happy to take that as the follow-up if this merges first.

src/plugins/telemetry/prometheus_mp/owner_election.h
e-eygin: Fair — reworded to "served again after the owner's death".

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
e-eygin: Right, that example is wrong: teardown frees the port before releasing the election, so a previous run still shutting down holds the lock with the port already free — `SIBLING_OWNS`. Dropped the clause.

src/plugins/telemetry/prometheus_mp/prometheus_mp_exporter.cpp
e-eygin: The call is skipped entirely while serving, and inside `reclaim()` the throttle is one integer compare, so the flock runs at most 5x/s per non-owner. Happy to lengthen the interval if you'd rather — it's the takeover latency after the owner dies.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
e-eygin: Added a `NIXL_DEBUG` with it before conceding. The exporter warns about the outcome, but the civetweb message itself was being dropped.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
e-eygin: Switched to `std::chrono::nanoseconds` for the parameter and both members, with the constants written as 200ms and 5s. Kept nanoseconds rather than coarser: the value is the store heartbeat, which is ns on disk, so anything else adds a division per exported event to save nothing.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
e-eygin: Added.

src/plugins/telemetry/prometheus_mp/owner_election.h
e-eygin: Yes, for the degraded path: the c'tor only truncates when it actually took the lock. When the lock is unusable (flock fails with anything but `EWOULDBLOCK`) every rank believes it won and nothing truncates, so a shorter endpoint written over a longer stale one would read back as the two spliced together. Added a comment saying so.

src/plugins/telemetry/prometheus_mp/owner_election.h
e-eygin: Done, your first variant. Dropped the default c'tor and both move ops; `scrapeEndpoint` holds a `std::optional<ownerElection>`, so giving it up is `reset()` and re-running it `emplace()` — the two open file descriptions never coexist. The `serving()` guard in `reclaim()` stays, since `emplace()` would destroy the election being served on.

src/plugins/telemetry/prometheus_mp/owner_election.h
ColinNV: That sounds like there is a race condition if multiple losers do a "truncate then write", e.g. if two processes both first truncate and then both write.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
```diff
@@ -0,0 +1,139 @@
...
+    : dir_(std::move(dir)),
+      bindAddress_(std::move(bind_address)),
+      staleTtl_(stale_ttl),
+      retryInterval_(retryInterval) {}
+
+scrapeEndpoint::status
+scrapeEndpoint::claim() {
+    election_.emplace(dir_);
+    if (!election_->won()) {
+        return status::SIBLING_OWNS;
```
ColinNV, line 67: In `reclaim()` there is always an `election_.reset()` at this point, why not here? Perhaps the owner election should be a function, not a class, that returns a `std::optional< scopedFd >`? Then there would be only one layer to it.

src/utils/common/scoped_fd.h
e-eygin: Done now rather than as a follow-up, since CI re-runs anyway: it's at src/utils/common/scoped_fd.h. I took #1930's copy verbatim — it's ours with the guard, namespace and one doc line changed, class body identical — so both PRs now add the same file with the same content and whoever lands second keeps either side.

src/plugins/telemetry/prometheus_mp/owner_election.h
e-eygin: Right — and only when the lock is unusable, which is the very path the truncate was there for. Now one fixed-size `pwrite`: nothing to interleave, no splicing, reader stops at the first `NUL`.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.cpp
e-eygin: Deliberate, now commented:` claim()`'s loser keeps the fd because the exporter reads the owner's endpoint off it; `reclaim()`'s loser has no reader.

A function can't say it in two states — won-and-holding, lost, and won-without-a-lock (flock failing with anything but `EWOULDBLOCK`) — and nullopt for "lost" drops the fd the loser still reads from.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
ColinNV: Right, I was thinking we might want to close the gap.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
```diff
@@ -0,0 +1,123 @@
...
+    reclaim(std::chrono::nanoseconds now) noexcept;
+
+    [[nodiscard]] bool
+    serving() const noexcept {
+        return static_cast<bool>(exposer_);
+    }
+
+    /// The endpoint the current owner published, empty if there is none.
+    [[nodiscard]] std::string
+    ownerEndpoint() const {
```
ColinNV: Perhaps standardize on "address" for `ip:port` strings and not call them endpoint?

src/plugins/telemetry/prometheus_mp/owner_election.h
```diff
@@ -0,0 +1,177 @@
...
+        struct stat st{};
+        if (::fstat(fd_.get(), &st) != 0 || !S_ISREG(st.st_mode) || st.st_uid != ::geteuid()) {
+            fd_.reset();
+            won_ = true;
+            warnUnusable("not a regular file owned by this user", warn_if_unusable);
+            return;
+        }
+        if (::flock(fd_.get(), LOCK_EX | LOCK_NB) == 0) {
+            won_ = true;
+            // A previous owner's endpoint outlives it in the file. Drop it now,
```
ColinNV: There's still a race because the flock and ftruncate are separate operations.

ColinNV: Perhaps the two things "choose a winner between all instances using the same address" and "warn if it looks like another address is used in the same directory" should be separated, make the address (or just the port) part of the lock file name, use the lock file for owner election, and separately warn if there are multiple lock files in the same directory? This would allow the same data to be exported by multiple winners if they are configured to use different ports, but only if that's what the user configured and launched.

ColinNV (changes requested):

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Closed in 6a394758. The writer no longer unlinks its store when it shuts down, so the file stays behind holding its final values and the collector keeps publishing them until the process is gone and its last update is older than the stale TTL — the same path a crashed rank already took. The values now survive at least one more scrape instead of dying with the process.

src/plugins/telemetry/prometheus_mp/owner_election.h
e-eygin: Done as proposed, in 22c64d4c.

src/plugins/telemetry/prometheus_mp/scrape_endpoint.h
e-eygin: Renamed in f2f65867.

src/plugins/telemetry/prometheus_mp/mp_collector.cpp
```diff
@@ -0,0 +1,307 @@
...
+            static_cast<long>(std::chrono::duration_cast<std::chrono::seconds>(ttl).count());
+        const long grace = std::max<long>(ttl_seconds, kInvalidFileFloorSeconds);
+        const long age = static_cast<long>(::time(nullptr) - st.st_mtime);
+        return age > grace;
+    }
+
+    [[nodiscard]] bool
+    nameMatchesStore(const std::string &name) {
+        return name.size() > MP_STORE_FILE_PREFIX.size() + MP_STORE_FILE_SUFFIX.size() &&
+            name.compare(0, MP_STORE_FILE_PREFIX.size(), MP_STORE_FILE_PREFIX) == 0 &&
```
ColinNV: We're using C++20 now, can use `std::string::starts_with()` and `...::ends_with()`.

src/plugins/telemetry/prometheus_mp/mp_collector.cpp
```diff
@@ -0,0 +1,307 @@
...
+        }
+        m.histogram.sample_count = cumulative;
+        m.histogram.sample_sum = static_cast<double>(s.histSums[slot]);
+        return m;
+    }
+
+    // Minimum age before an unparsable file is reaped, even when the TTL is 0.
+    // Protects a store a live process is actively creating (a sub-second window)
+    // from being deleted out from under it.
+    constexpr long kInvalidFileFloorSeconds = 2;
```
ColinNV: The file handling and liveness checks seem to be quite complicated and with a lot of border cases. Why not stick to a simple "one process one file" model, i.e. the file published by a rank is created with O_TMPFILE, then flock-ed, then given a name (or created with a temporary name and then renamed into place after being flock-ed) with the name including the pid, the startup time and for extra paranoia a random string. Then the entire "is the writer still alive" can be reduced to an attempted flock, and any file that can be flock-ed can be removed (after getting the final values if they are not too old) because the only writer it will ever have is gone.

ColinNV (changes requested):

src/plugins/telemetry/prometheus_mp/mp_collector.cpp
e-eygin: Done in 5f927354, in both the lock-file and store-name checks.

src/plugins/telemetry/prometheus_mp/mp_collector.cpp
e-eygin: Done as proposed, in b9ccedb1.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+std::string
+makeStoreFileName(int64_t pid, uint64_t start_time, uint64_t instance) {
+    return std::string(MP_STORE_FILE_PREFIX) + std::to_string(pid) + "." +
+        std::to_string(start_time) + "." + std::to_string(instance) +
+        std::string(MP_STORE_FILE_SUFFIX);
+}
+
+uint64_t
+readProcessStartTime(int64_t pid) {
```
ColinNV: Why is this needed beyond a global variable that initializes to the current timestamp?

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,370 @@
...
+        ::munmap(mapping_, mappingSize_);
+        mapping_ = nullptr;
+    }
+}
+
+bool
+storeWriterAlive(const std::filesystem::path &path) {
+    const scopedFd fd(::open(path.c_str(), O_RDONLY | O_CLOEXEC | O_NOFOLLOW));
+    if (!fd.valid()) {
+        return true;
```
ColinNV, line 191: The `return true` is slightly surprising at first sight, perhaps a comment that error handling is left to the next operation on `path`.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
```diff
@@ -0,0 +1,370 @@
...
+        ::munmap(mapping_, mappingSize_);
+        mapping_ = nullptr;
+        throw;
+    }
+    fd_ = std::move(staged.fd);
+}
+
+storeWriter::~storeWriter() {
+    if (mapping_ != nullptr) {
+        ::munmap(mapping_, mappingSize_);
```
ColinNV: RAII helper?

ColinNV (commented):
I can't find any more race conditions.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Added `nixl::mappedRegion` in `src/utils/common`, next to `scopedFd` and shaped like it.

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: Commented as suggested: nothing was probed, so nothing is concluded, and whatever stopped the open is for the next operation on the path to report. 8f259807

src/plugins/telemetry/prometheus_mp/mp_store.cpp
e-eygin: It isn't, once liveness became the lock: the name only has to differ between two runs that share a pid, and nothing reads the value. `processRunMarker()` is now a function-local `static const` taken from `CLOCK_REALTIME` at first call, `makeStoreFileName`'s parameter is `run_marker`, and the `/proc/<pid>/stat` field-22 parser is deleted. 8f259807

ColinNV (approved):

PR #1925, opened 2026-07-10
build: fix ucx: add UCS_BIT_GET fallback for UCX < 1.19

RuixiangMa:
## What?
Add `#ifndef UCS_BIT_GET` fallback in `ucx_utils.cpp` for UCX < 1.19.
## Why?
`UCS_BIT_GET` is used unconditionally (no version guard), but the macro only exists in UCX 1.22+. Ubuntu 24.04 et al apt provides UCX 1.16, causing compile failure. meson.build does not enforce a minimum UCX version.

## How?
     #ifndef UCS_BIT_GET
     #define UCS_BIT_GET(_mask, _bit) (!!((_mask) & UCS_BIT(_bit)))
     #endif

     Matches UCX 1.22 upstream definition. #ifndef guard ensures no conflict with newer UCX.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility of UCX feature/flag checks by adding a guarded fallback for bit-get support when it isn’t provided by external UCX headers.
  * Ensures bitmask-based flag evaluation continues to work reliably across environments.

cc @brminich

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -32,7 +32,7 @@
 #include "serdes/serdes.h"
 
 #ifndef UCS_BIT_GET
-#define UCS_BIT_GET(_mask, _bit) (!!((_mask) & UCS_BIT(_bit)))
+#define UCS_BIT_GET(_value, _i)  (!!((_value) & UCS_BIT(_i)))
 #endif
 
 [[nodiscard]] nixl_b_params_t
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -30,6 +30,7 @@
 #include "common/nixl_log.h"
 #include "config.h"
 #include "serdes/serdes.h"
+
 #ifndef UCS_BIT_GET
 #define UCS_BIT_GET(_mask, _bit) (!!((_mask) & UCS_BIT(_bit)))
 #endif
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -32,7 +32,7 @@
 #include "serdes/serdes.h"
 
 #ifndef UCS_BIT_GET
-#define UCS_BIT_GET(_value, _i)  (!!((_value) & UCS_BIT(_i)))
+#define UCS_BIT_GET(_value, _i) (!!((_value) & UCS_BIT(_i)))
 #endif
 
 [[nodiscard]] nixl_b_params_t
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -30,6 +30,9 @@
 #include "common/nixl_log.h"
 #include "config.h"
 #include "serdes/serdes.h"
+#ifndef UCS_BIT_GET
+#define UCS_BIT_GET(_mask, _bit) (!!((_mask) & UCS_BIT(_bit)))
+#endif
 
 [[nodiscard]] nixl_b_params_t
 get_ucx_backend_common_options() {


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -30,6 +30,9 @@
 #include "common/nixl_log.h"
 #include "config.h"
 #include "serdes/serdes.h"
+#ifndef UCS_BIT_GET
```
guy-ealey-morag, line 34:
```suggestion

#ifndef UCS_BIT_GET
```

add empty line

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -30,6 +30,9 @@
...
+#define UCS_BIT_GET(_mask, _bit) (!!((_mask) & UCS_BIT(_bit)))
```
guy-ealey-morag:
```suggestion
#define UCS_BIT_GET(_value, _i)  (!!((_value) & UCS_BIT(_i)))
```

use the same arg names as in UCX

guy-ealey-morag:
It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?

rakhmets:
> It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?

Right, it was added in UCX 1.19.
And NIXL's UCX plugin does not imply use with a version older than 1.19.
it is better to avoid adding this macro to the UCX plugin code, and add a version limitation to the meson file.

RuixiangMa:
> > It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?

Thanks for raising this. I double-checked the UCX, and my earlier version analysis was not correct

RuixiangMa:
> > It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?
> 
> Right, it was added in UCX 1.19. And NIXL's UCX plugin does not imply use with a version older than 1.19. it is better to avoid adding this macro to the UCX plugin code, and add a version limitation to the meson file.

 Thanks for the suggestion. After taking another look, it seems the UCX plugin still accommodates UCX < 1.19 rather than clearly treating it as unsupported:

  - src/plugins/ucx/ucx_utils.cpp has a ucpVersion_ < UCP_VERSION(1, 19) branch that emits a warning about limited CUDA support, rather than rejecting the version.

I agree that, if we decide to raise the minimum supported UCX version, enforcing that in Meson would be cleaner than adding a local fallback.

rakhmets:
> > > It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?
> > 
> > 
> > Right, it was added in UCX 1.19. And NIXL's UCX plugin does not imply use with a version older than 1.19. it is better to avoid adding this macro to the UCX plugin code, and add a version limitation to the meson file.
> 
> Thanks for the suggestion. After taking another look, it seems the UCX plugin still accommodates UCX < 1.19 rather than clearly treating it as unsupported:
> 
> * src/plugins/ucx/ucx_utils.cpp has a ucpVersion_ < UCP_VERSION(1, 19) branch that emits a warning about limited CUDA support, rather than rejecting the version.
> 
> I agree that, if we decide to raise the minimum supported UCX version, enforcing that in Meson would be cleaner than adding a local fallback.

These are two related, but not exactly the same topics.
We have a runtime UCX version check. We issue a warning, if version older than 1.19 is loaded. And warning is enough here. As, by itself, the use of older versions is not prohibited, but it has never been fully tested. Therefore, full functionality support is not guaranteed.
The topic of the current PR is a compile time check. And from the compilation point of view, there is no great benefit from supporting older versions of UCX. So, we can add a restriction to the meson file.

RuixiangMa:
> > > > It looks like `UCS_BIT_GET` was first added to UCX `v1.19.0-rc1`, are you sure your analysis is correct?
> > > 
> > > 
> > > Right, it was added in UCX 1.19. And NIXL's UCX plugin does not imply use with a version older than 1.19. it is better to avoid adding this macro to the UCX plugin code, and add a version limitation to the meson file.
> > 
> > 
> > Thanks for the suggestion. After taking another look, it seems the UCX plugin still accommodates UCX < 1.19 rather than clearly treating it as unsupported:
> > 
> > * src/plugins/ucx/ucx_utils.cpp has a ucpVersion_ < UCP_VERSION(1, 19) branch that emits a warning about limited CUDA support, rather than rejecting the version.
> > 
> > I agree that, if we decide to raise the minimum supported UCX version, enforcing that in Meson would be cleaner than adding a local fallback.
> 
> These are two related, but not exactly the same topics. We have a runtime UCX version check. We issue a warning, if version older than 1.19 is loaded. And warning is enough here. As, by itself, the use of older versions is not prohibited, but it has never been fully tested. Therefore, full functionality support is not guaranteed. The topic of the current PR is a compile time check. And from the compilation point of view, there is no great benefit from supporting older versions of UCX. So, we can add a restriction to the meson file.

I am not fully sure about is whether the current AWS EFA workflow comment is stale. It still pins UCX to v1.18.0 with a comment saying this is a temporary workaround until EFA support is vailable in UCX 1.19.

If EFA is already supported and validated with UCX 1.19, then I agree that adding the Meson version restriction would be the right direction. In that case, we should also update the AWS EFA workflow accordingly.

PR #1927, opened 2026-07-11
util: add reusable comma-separated env-list parser (NIX-1621)

e-eygin:
## What?
Adds a small, header-only helper `nixl::str::splitStripped(absl::string_view, char delim = ',')`
in `src/utils/common/str_util.h` that splits a delimited string into non-empty,
whitespace-stripped tokens (order-preserving, duplicates kept). Refactors
`resolveTraceBackends()` in `src/core/nixl_agent.cpp` (`NIXL_TRACE_BACKENDS`) to use it,
replacing the open-coded `absl::StrSplit(',')` + `absl::StripAsciiWhitespace` + skip-empty
loop with a range-insert into the backend set. Adds unit tests
(`test/gtest/unit/util/str_util_test.cpp`).

## Why?
Follow-up from the PR #1897 review (NIX-1621): the same comma-separated env-list parsing
idiom is open-coded in several places (tracing backends, DOCA exporter backends, histogram
buckets, per-metric activation). This lands the shared helper and converts the one call site
that exists on `main` today (`resolveTraceBackends`). Kept intentionally minimal and
behavior-preserving so the remaining call sites can adopt it independently as their PRs land;
if this merges first the helper is already available to cherry-pick on rebase, avoiding a
separate follow-up PR.

Ref: https://linear.app/nvidia/issue/NIX-1621

## How?
`splitStripped` centralizes the `StrSplit` + `StripAsciiWhitespace` + skip-empty logic and
returns a `std::vector<std::string>`; deduplication stays a caller concern (`resolveTraceBackends`
still collects into a `std::set`, so the set-but-empty "off" semantics and nsys auto-enable
precedence are unchanged). No `TELEMETRY_VERSION`/API change. Existing tracing tests
(`Tracing.ResolveBackends*`) plus 4 new `StrSplitStripped` cases pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trace backend configuration parsing to trim ASCII whitespace, ignore empty entries, and correctly handle repeated delimiters while keeping the prior NVTX enable/disable behavior.
- **New Features**
  - Added shared string-splitting utilities to split/trim and optionally return a deduplicated set.
- **Tests**
  - Added unit tests for the new split behaviors (ordering vs. deduplication, whitespace-only handling, and custom delimiters).
  - Updated the unit-test build setup to include the new utility tests.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -22,9 +22,6 @@
 #include <optional>
 #include <set>
 
-#include <absl/strings/ascii.h>
-#include <absl/strings/str_split.h>
-
 #include "nixl.h"
 #include "serdes/serdes.h"
 #include "backend/backend_engine.h"
@@ -35,6 +32,7 @@
 #include "common/nixl_log.h"
 #include "common/operators.h"
 #include "common/hw_info.h"
+#include "common/str_util.h"
 #include "telemetry.h"
 #include "telemetry_event.h"
 #include "tracing/trace.h"
@@ -129,12 +127,8 @@ resolveTraceBackends(const std::optional<std::string> &explicit_spec, bool under
     if (explicit_spec) {
         // Trim entries so a padded value like "chakra, nvtx" matches backend
         // names; the set dedups them.
-        for (const absl::string_view raw : absl::StrSplit(*explicit_spec, ',')) {
-            const absl::string_view name = absl::StripAsciiWhitespace(raw);
-            if (!name.empty()) {
-                backends.emplace(name);
-            }
-        }
+        const auto names = nixl::str::splitStripped(*explicit_spec);
+        backends.insert(names.begin(), names.end());
         // Set-but-empty (or all-blank) is an explicit "off" that must beat the
         // nsys auto-enable below.
         explicit_off = backends.empty();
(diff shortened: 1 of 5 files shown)


e-eygin:

src/core/nixl_agent.cpp
```diff
@@ -129,12 +127,8 @@ resolveTraceBackends(const std::optional<std::string> &explicit_spec, bool under
...
         // Trim entries so a padded value like "chakra, nvtx" matches backend
         // names; the set dedups them.
-        for (const absl::string_view raw : absl::StrSplit(*explicit_spec, ',')) {
-            const absl::string_view name = absl::StripAsciiWhitespace(raw);
-            if (!name.empty()) {
-                backends.emplace(name);
-            }
-        }
+        const auto names = nixl::str::splitStripped(*explicit_spec);
+        backends.insert(names.begin(), names.end());
```
ColinNV: IIIRC the dedup is used multiple times, too; second version that returns a `std::set`?

e-eygin: Done in `286bc4df` — added `splitStrippedSet` (deduped + sorted) and switched `resolveTraceBackends` to it. Vector `splitStripped` stays for order-sensitive callers (e.g. histogram buckets).

ColinNV (approved):

PR #1931, opened 2026-07-13
Cpu proxy ucx backend

tomerdav:
## What?

Add a CPU-proxy backend for the NIXL GPU Device API, allowing GPU kernels to initiate transfers (PUT, atomic add) on transports without native device-initiated support by delegating submission to CPU worker threads.

- **GPU Device API restructure**: split `src/api/gpu/` into a backend-agnostic common layer with `ucx/` and `proxy/` implementations, selected via the `NIXL_GPU_DEVICE_BACKEND_PROXY` macro and a new `gpu_device_api_backend` meson option (`auto`/`ucx`/`proxy`/`none`).
- **Proxy runtime** (`src/core/device_proxy/`): `ProxyRuntime` manages per-channel GPU-to-CPU submission rings and a memview registry; `ProxyWorker` threads drain the rings and dispatch through a `DeviceProxyBackendAdapter` interface.
- **UCX adapter**: `nixlUcxEngine` implements the backend-provider interface; `nixlUcxProxyBackend` translates proxy submissions into direct UCX RMA puts and atomic adds.
- **Agent API**: new `nixlAgentConfig` fields (`enableDeviceProxy`, `proxyWorkerCount`, `proxyChannelCount`) and `nixlAgent::getProxyDeviceContext()` for publishing the device context to kernels; remote connection info is propagated to the proxy runtime after metadata exchange.
- **Tests**: unit tests for the runtime and memview registry (stub backend, no GPU required), plus an end-to-end `proxy_write_test.cu` and proxy-mode coverage in `single_write_test.cu`.

## Why?

The GPU Device API currently requires UCX device-side support (GDAKI-capable transports). A CPU proxy makes device-initiated transfers available on any backend that can serve as a proxy adapter, with UCX as the first provider.

## How?

```
      GPU                      pinned host memory                CPU                    transport
┌────────────────┐   submission ring (64B records)   ┌──────────────────────┐   ┌──────────────────┐
│ CUDA kernel    │ ────────────────────────────────► │ ProxyWorker thread   │──►│ DeviceProxy-     │
│ device API     │      write payload, commit          acquire-poll op_idx,  │  │ BackendAdapter   │──► remote
│ proxy enqueue  │      op_idx last (release)          dispatch via adapter  │  │ (UCX RMA put /   │    peer
│                │                                   │                      │   │  atomic add)     │
│ poll completion│ ◄──────────────────────────────── │ publish completions  │   └──────────────────┘
└────────────────┘        completion slot            └──────────────────────┘
```

- Each channel is a fixed-depth ring of 64-byte submission records in mapped pinned host memory. 
- The GPU enqueue path writes the record payload first and commits `op_idx` last with a GPU-scope release; the CPU worker acquire-polls `op_idx` as the readiness word, dispatches through the backend adapter, and publishes completions back through a pinned completion slot that the GPU polls. 
- To keep the device fast path cheap, the producer index lives in device memory, the consumer index is cached on device and refreshed from host only when the ring appears full, and the proxy context is published via constant memory with force-inlined device helpers.

### DIFF
--- a/examples/device/ep/meson.build
+++ b/examples/device/ep/meson.build
@@ -18,6 +18,10 @@ if not cuda_dep.found()
     subdir_done()
 endif
 
+if gpu_device_api_backend_resolved != 'ucx'
+    error('nixl_ep requires gpu_device_api_backend=ucx because it currently uses the UCX GPU Device API')
+endif
+
 py = import('python').find_installation('python3', pure: false)
 
 # Check if PyTorch is available
@@ -72,7 +76,7 @@ nixl_ep_sources = [
 
 nixl_ep_inc_dirs = [
     nixl_inc_dirs,
-    nixl_gpu_inc_dirs,
+    nixl_gpu_device_api_inc_dirs,
     include_directories('csrc'),
     include_directories('csrc/kernels'),
     torch_inc_dirs,
@@ -96,7 +100,7 @@ nixl_ep_cpp_args = [
     '-Wno-sign-compare',
     '-Wno-reorder',
     '-Wno-attributes',
-]
+] + gpu_device_api_cpp_args
 
 nixl_ep_cuda_args = [
     '-DHAVE_CUDA',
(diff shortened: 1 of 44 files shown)


tomerdav:
Replaced by a stack of PRs, starting with https://github.com/ai-dynamo/nixl/pull/2111

PR #1932, opened 2026-07-13
ci: right-size run-step and job timeouts

NirWolfer:
## What?
The on-failure enroot export (~5 min for a multi-GB container) runs inside each run step's timeout(), so the tight Python/Rust steps could SIGTERM the export mid-mksquashfs. Meanwhile CPP (120) and Nixlbench (50) were far larger than their real runtimes, and the four timeouts summed to 190 min — more than the Slurm allocation's --time (SLURM_JOB_TIMEOUT = 140 min), so a slow run could lose the allocation before the Jenkins timeouts even fired.

Right-size run steps to observed durations plus export headroom:
  CPP        120 -> 60   (~35 min observed)
  Python      10 -> 15   (~7 min observed; UCX createBackend can be slow)
  Rust        10 -> 10   (~1.5 min observed; already has headroom)
  Nixlbench   50 -> 25   (~10 min observed)

Sum is now 110 min, under the 140-min allocation. Axes run in parallel, so the job cap only needs to cover the longest single axis (docker ~11 + allocation wait up to 60 + steps 110 + stop ~5 ~= 186), so drop timeout_minutes 240 -> 200. Durations taken from three nixl-ci-gpu runs.

## Why?
The export capability would sometime fail on timeout if tests took all the test allocated time

## How?
architecture, approach, etc._

## Summary by CodeRabbit

* **Chores**
  * Updated GPU CI test time limits to improve overall execution efficiency.
  * Adjusted time allocations across C++, Python, Rust, and Nixlbench test phases to better fit workload expectations.

### DIFF
--- a/.ci/jenkins/lib/test-matrix.yaml
+++ b/.ci/jenkins/lib/test-matrix.yaml
@@ -20,7 +20,9 @@ job: nixl-ci-gpu
 # Fail job if one of the steps fails or continue
 failFast: false
 
-timeout_minutes: 240
+# Job cap. Axes run in parallel, so this bounds the longest single axis:
+# docker build (~11) + allocation wait (up to immediateTimeout 60) + run steps (110) + stop (~5).
+timeout_minutes: 200
 
 registry_host: artifactory.nvidia.com
 registry_auth: svc-nixl-new-artifactory-token
@@ -133,7 +135,7 @@ steps:
 
   - name: Run CPP tests
     containerSelector: "{name: 'build_helper'}"
-    timeout: 120
+    timeout: 60
     parallel: false
     shell: action
     module: slurmCI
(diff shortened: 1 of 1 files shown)


Alexey-Rivkin (changes requested):
Just to be on the safe side:
The export headroom on the two short steps is thinner than the export itself.
Export takes ~5 min, but after the test runs there's only ~3 min left before the timeout on Python (15) and Rust (10).

Could we bump Python 15>18 and Rust 10>12? Still well under the 140m allocation, just gives both steps a full export's worth of margin. wdyt?

NirWolfer:
> Just to be on the safe side: The export headroom on the two short steps is thinner than the export itself. Export takes ~5 min, but after the test runs there's only ~3 min left before the timeout on Python (15) and Rust (10).
> 
> Could we bump Python 15>18 and Rust 10>12? Still well under the 140m allocation, just gives both steps a full export's worth of margin. wdyt?

Updated Python to 20 min and Rust to 15 min

Alexey-Rivkin (approved):

PR #1934, opened 2026-07-13
nixlbench: add scoped raw POSIX CLI

razorback4417:
## Summary

This PR adds the first opt-in CLI11 command path for NIXLBench:

```text
nixlbench raw posix ...
```

When NIXLBench is configured with `-Dbuild_raw_cli=true`, the command provides scoped help, 1024-based human-readable sizes, validation for NIXLBench-owned options, a deterministic resolved plan, and `--dry-run`. Every non-`raw` invocation continues to use the existing gflags path.

## Ownership and execution

The command has three ownership layers:

| Owner | Responsibility |
| --- | --- |
| NIXLBench raw benchmark | Operation, sizes, iterations, warmups, threads, pipeline depth, consistency, and dry-run |
| NIXLBench `FILE_SEG` resources | Path, filenames, file count, and direct file opening |
| Selected plugin | Advertised initialization keys, defaults, value semantics, validation, and relationships |

```text
CLI11 RawOptions + FileOptions -> bounded compatibility adapter -> existing worker
plugin metadata/overrides -----> exact string map -------------> backend

legacy gflags -----------------> existing xferBenchConfig ------> existing worker
```

NIXLBench discovers the installed POSIX plugin's advertised parameters and accepts overrides only for published keys. It preserves key spelling, value spelling, and case, then forwards the resolved string map to backend creation without copying plugin aliases, defaults, ranges, or enum translations into the CLI. Invalid values therefore fail at the plugin/backend boundary.

The compatibility adapter carries only runner, benchmark, and `FILE_SEG` settings into the existing execution core. Plugin parameters travel separately, and the legacy POSIX parameter assembly remains unchanged for flags-only invocations.

## Build and example

The raw CLI is experimental and disabled by default:

```bash
meson setup build -Dbuild_raw_cli=true
```

Sizes accept `KB`, `MB`, `GB`, and `TB` with 1024-based multipliers.

```bash
mkdir -p /tmp/nixlbench-data

nixlbench raw posix \
  --path /tmp/nixlbench-data \
  --operation write \
  --total-buffer-size 64MB \
  --start-block-size 4KB \
  --max-block-size 1MB \
  --iterations 16 \
  --warmup-iterations 0 \
  --check-consistency \
  --plugin-param ios_pool_size 4096
```

`nixlbench raw posix --help` lists the initialization parameters and defaults advertised by the installed POSIX build.

## Validation

The merged source adds 20 focused GoogleTest tests across six suites. Coverage includes dispatch, parsing and validation, metadata-derived help and defaults, exact plugin overrides, deterministic plans, dry-run behavior, POSIX write/read consistency, failure cleanup, and legacy/raw equivalence.

Development validation also covered full NIXL and NIXLBench builds, the existing POSIX plugin test, representative legacy POSIX and two-process legacy UCX paths, and raw POSIX runtime checks on a GH200. All final checks listed on the PR passed.

## Scope

This PR does not add other raw plugins, modeled scenarios, configuration files, shell completion, gflags removal, or a full execution-configuration rewrite. Those remain separate review boundaries.

One help-generation cleanup remains for a follow-up: use CLI11's generated help output instead of maintaining the higher-level help summary manually.


src/plugins/posix/posix_plugin.cpp
```diff
@@ -22,17 +22,46 @@
...
+nixl_b_params_t
+makePosixPluginParams() {
+    nixl_b_params_t params = {{"ios_pool_size", "65536"}, {"kernel_queue_size", "256"}};
+#ifdef HAVE_LINUXAIO
+    params["use_aio"] = "true";
+#endif
+#ifdef HAVE_LIBURING
+    params["use_uring"] =
+#ifdef HAVE_LINUXAIO
+        "false";
```
ofer: They can't all be true, the default should either be all falses or follow the default selection code (std::string_view
nixlPosixIOQueue::getDefaultIoQueueType(void) {
#ifdef HAVE_LINUXAIO
    return "AIO";
#elif defined(HAVE_LIBURING)
    return "URING";
#elif defined(HAVE_POSIXAIO)
    return "POSIXAIO";
#else
    // Should never reach here. At least one of the queues should be available.
    NIXL_ASSERT(false);
    return nullptr;
#endif)

ofer: Unfortunately, the POSIX plugin has the boolean on one of the use_ parameters instead of a use_io_backend = aio, uring, or posixaio but the idea is that only one of the use_* are true

benchmark/nixlbench/src/utils/raw_cli.h
```diff
@@ -0,0 +1,87 @@
...
+    size_t max_batch_size = 1;
+    int iterations = 1000;
+    int warmup_iterations = 100;
+    int threads = 1;
+    int pipeline_depth = 1;
+    bool check_consistency = false;
+    bool dry_run = false;
+};
+
+struct PosixOptions {
```
ofer: Although this is technically ok for Posix, you should understand what is a FILE_SEG memory type backend option and what is a specific POSIX option (when you extend this to GDS, some of the stuff is common, some is not)

ofer: Why did you modify the posix plugin if you placed the settings hardcoded in nixlbench?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,488 @@
...
+    posix->add_option("--path", request.posix.path, "Directory for automatically named files");
+    posix->add_option(
+        "--filenames", request.posix.filenames, "Comma-separated explicit file names");
+    posix->add_option("--num-files", request.posix.num_files, "Number of backing files");
+    posix->add_flag("--direct", request.posix.direct, "Use direct I/O");
+    auto *api_option =
+        posix->add_option("--api", request.posix.api, "POSIX I/O API exposed by plugin metadata");
+    api_option->check(CLI::IsMember(apis, CLI::ignore_case));
+    api_option->default_str(request.posix.api);
+    if (metadata.parameters.contains("ios_pool_size")) {
```
ofer: This is very confusing, your both hardcoding ios_pooL_size and checking whether it exists in the metadata, like i wrote before, either change posix_plugin and add the defaults or have it all with magic numbers making nixlbench know things about the plugins.  Not both.  The idea behind this update though is to start having the plugin supply defaults via the custom params and nixlbench pick them up without knowing anything about them so that's the direction I suggest.  It means that ios_pool_size should not be known by nixlbench, the posix plugin should supply the default ios_pool_size and nixlbench needs to allow to change it because it exists as a default.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,488 @@
...
+        help_requested = true;
+        return app.exit(exception, out, err);
+    }
+    catch (const CLI::ParseError &exception) {
+        return app.exit(exception, out, err);
+    }
+
+    request.raw.operation = upper(request.raw.operation);
+    request.posix.api = upper(request.posix.api);
+    if (request.posix.api == "POSIX-AIO") {
```
ofer: I would suggest removing this and the option needs to be POSIXAIO without the dash and no correction should be applied.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,488 @@
...
+        }
+        if (request.raw.start_block_size == 0 ||
+            request.raw.max_block_size < request.raw.start_block_size) {
+            return fail("block sizes must be positive and max must be at least start");
+        }
+        if (request.raw.start_batch_size == 0 ||
+            request.raw.max_batch_size < request.raw.start_batch_size) {
+            return fail("batch sizes must be positive and max must be at least start");
+        }
+        if (request.posix.io_pool_size < POSIX_MIN_IO_POOL_SIZE ||
```
ofer: The posix stuff should probably no longer be validated by nixlbench as it doesn't know the ranges allowed, it should only know what's exposed by the custom params.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,488 @@
...
+            << formatSize(xferBenchConfig::max_block_size)
+            << "\n  batch sizes: " << xferBenchConfig::start_batch_size << " .. "
+            << xferBenchConfig::max_batch_size << "\n  iterations: " << xferBenchConfig::num_iter
+            << " (warmup " << xferBenchConfig::warmup_iter << ")"
+            << "\n  threads: " << xferBenchConfig::num_threads
+            << "\n  pipeline depth: " << xferBenchConfig::pipeline_depth
+            << "\n  consistency check: "
+            << (xferBenchConfig::check_consistency ? "enabled" : "disabled") << "\n  direct I/O: "
+            << (xferBenchConfig::storage_enable_direct ? "enabled" : "disabled")
+            << "\n  POSIX API: " << xferBenchConfig::posix_api_type
```
ofer: the xferBenchConfig is where the idea of basing the settings on the custom fields given by the plugin starts falling apart because xferBenchConfig also knows everything about the plugins.  How do you think it should be solved?

src/plugins/posix/posix_plugin.cpp
razorback4417: I set all the advertised use_*  defaults to false. when none is selected, the plugin’s existing fallback chooses the compiled default, so nixlbench no longer duplicates that selection logic

benchmark/nixlbench/src/utils/raw_cli.h
razorback4417: removed the nixlbench copies, so the plug metadata is now the single source of truth (previously was trying to expose the settings from the plugin while still using nixlbench's existing execution path)

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: ack! options are now generated mechanically from `metadata.parameters`, including params nixlbench hasn't seen before

(also removed all the `ios_pool_size` logic from the raw CLI)

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: see update commit-

in general, plug params now travel separately through `main`, `runBenchmark`, and `createWorker`. The worker applies them over the plugin's published defaults and skips the legacy posix reconstruction.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,428 @@
...
+int
+parseRawPosixCommand(int argc,
+                     char *argv[],
+                     const PluginMetadata &metadata,
+                     RawPosixRequest &request,
+                     bool &help_requested,
+                     std::ostream &out,
+                     std::ostream &err) {
+    help_requested = false;
+    if (metadata.name != "POSIX" || !hasMemoryType(metadata, DRAM_SEG) ||
```
ofer: This is confusing, nixlbench is not supposed to know about plugins, it's supposed to read information from backend plugins.  We are in sort of a middle ground where we know we're talking about POSIX but we should pick up as much information from NIXL and know the least amount about it in nixlbench itself.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,428 @@
...
+        ->group("Raw benchmark options");
+    raw->add_option("--pipeline-depth", request.raw.pipeline_depth, "Transfer requests in flight")
+        ->group("Raw benchmark options");
+    raw->add_flag(
+           "--check-consistency", request.raw.check_consistency, "Validate transferred bytes")
+        ->group("Raw benchmark options");
+    raw->add_flag("--dry-run", request.raw.dry_run, "Print the resolved plan without executing")
+        ->group("Raw benchmark options");
+
+    posix->add_option("--path", request.file.path, "Directory for automatically named files")
```
ofer: just to start making this generic, check that the memory types that the plugin that was chosen has fileseg, if it does, then you add path, filenames, and num files.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,428 @@
...
+    posix->add_option("--num-files", request.file.num_files, "Number of backing files")
+        ->group("FILE_SEG resource options");
+    posix->add_flag("--direct", request.file.direct, "Use direct file opening")
+        ->group("FILE_SEG resource options");
+    for (const auto &key : sortedParameterKeys(metadata.parameters)) {
+        posix
+            ->add_option(
+                "--" + key, request.plugin_parameters.at(key), "Plugin initialization parameter")
+            ->default_str(metadata.parameters.at(key))
+            ->group("Plugin initialization parameters");
```
ofer: Is the key necessarily a valid string for cli?

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Addressed - I removed the gate so the parser now uses the returned metadata to decide what it can expose (we shouldnt treat POSIX metadata as a fixed schema)

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Makes sense - new commit derives file support from whether the plugin advertises fileseg, then uses it consistently for path, filenames, filecount

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Addressed in `2bc4091`

razorback4417:
CodeRabbit comment above is already addressed by the integration path - `smallRawCommand()` places the raw options after posix and the dry-run tests execute it through the real binary

The unit test intentionally covers the opposite ordering

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,443 @@
...
+            ->group("FILE_SEG resource options");
+        posix->add_option("--num-files", request.file.num_files, "Number of backing files")
+            ->group("FILE_SEG resource options");
+        posix->add_flag("--direct", request.file.direct, "Use direct file opening")
+            ->group("FILE_SEG resource options");
+    }
+    if (!metadata.parameters.empty()) {
+        posix
+            ->add_option("--plugin-param",
+                         plugin_parameter_overrides,
```
ofer: This is better but causes problems when you run helpRequested() to print out the available options as it won't display the available parameters if I'm not mistaken

razorback4417: `--plugin-param` still appeared, but once I hid cli11's generated validator text, help no longer showed which plugin keys were available. I fixed this by just listing the advertised keys/defaults explicitly

Separately, after rechecking the dry-run path, I caught that the plan was printing the requested iteration counts even after the exiting config path normalized execution (so now it shows requested/normalized counts separately without changing exec behavior)

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,443 @@
...
+std::vector<std::string>
+benchmarkFileArguments(const RawPosixRequest &request, const std::string &program_name) {
+    const auto boolean = [](bool value) { return value ? "true" : "false"; };
+    std::vector<std::string> arguments = {
+        program_name,
+        // Fixed values select the existing NIXL/POSIX runner.
+        "--worker_type=nixl",
+        "--backend=POSIX",
+        "--initiator_seg_type=DRAM",
+        "--target_seg_type=DRAM",
```
ofer: In a storage plugin (like POSIX), the target seg isn't used.  If the backend is GDS, the initiator seg can be DRAM or VRAM, whether that's allowed you can query from the plugins memory types.

razorback4417: I removed that argument. POSIX still uses DRAM locally, but we now check that against the plugin’s advertised memory types (GDS/VRAM selection stays out of this PR).

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,477 @@
...
+    catch (const CLI::CallForHelp &exception) {
+        help_requested = true;
+        return app.exit(exception, out, err);
+    }
+    catch (const CLI::ParseError &exception) {
+        return app.exit(exception, out, err);
+    }
+
+    if (!hasMemoryType(metadata, DRAM_SEG)) {
+        err << "Error: POSIX plugin must advertise DRAM_SEG for local memory\n";
```
ofer: use the metadata.Name, remember you want to keep the generic case in mind

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,477 @@
...
+        return 2;
```
ofer: Don't use a magic number on the return, define it in a #def or something else

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: addressed - both `DRAM_SEG` and `FILE_SEG` capability errors now use `metadata.name`

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: addressed - replaced the with `kInvalidArgumentsExitCode` and normalized remaining raw-command statuses to `EXIT_SUCCESS` / `EXIT_FAILURE`

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,479 @@
...
+        arguments.push_back("--num_files=" + std::to_string(request.file.num_files));
+        arguments.push_back("--storage_enable_direct=" + std::string(boolean(request.file.direct)));
+    }
+    return arguments;
+}
+
+RawCommandResult
+prepareRawCommand(int argc, char *argv[], std::ostream &out, std::ostream &err) {
+    std::string discovery_error;
+    const auto metadata = discoverPluginMetadata(XFERBENCH_BACKEND_POSIX, discovery_error);
```
ofer, line 423: This is kind of weird in relation to making this change more general, remember to plan for having multiple backends, you will need to eventually discover all plugins that support FILE_SEG then push that list into the parseRawPosixCommand

benchmark/nixlbench/README.md
```diff
@@ -384,11 +384,90 @@ export LD_LIBRARY_PATH=/usr/local/nixlbench/lib:$LD_LIBRARY_PATH
 - `etcd_lib_path`: Path to ETCD C++ client library
 - `nvshmem_inc_path`: Path to NVSHMEM include directory
 - `nvshmem_lib_path`: Path to NVSHMEM library directory
+- `build_tests`: Build NIXLBench tests for non-release builds (default: true)
 - `buildtype`: Build type: `debug`, `release`, `debugoptimized` (default: release)
 - `prefix`: Installation prefix (default: /usr/local)
 
 ## Usage
 
+### Raw POSIX command
```
ofer: The original readme does not go into gflags, this should just go into how you can use a verb based interface as well with the first available feature being raw etc.

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Addressed in `146dc991` -  `prepareRawCommand` now discovers all installed plugins advertising FILE_SEG and passes collection into a more generic raw command dispatcher

(PR 1 still scoped to just raw posix)

benchmark/nixlbench/README.md
razorback4417: Addressed in `146dc991`

ofer (dismissed):

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+    if (parsed.ec != std::errc() || number == 0) {
+        error = "size must be a positive integer";
+        return std::nullopt;
+    }
+
+    const std::string suffix = value.substr(digit_count);
+    uint64_t multiplier = 0;
+    if (suffix.empty() || suffix == "B") {
+        multiplier = 1;
+    } else if (suffix == "K" || suffix == "KB" || suffix == "KIB") {
```
aranadive: Having the Kibibyte vs Kilobyte support can lead to confusion especially if you're doing the 1024 multiplier. I would suggest to drop KiB/MiB support and keep the well understood (though incorrect) KB, MB, etc.

benchmark/nixlbench/meson.build
```diff
@@ -142,6 +142,10 @@ etcd_dep = dependency('etcd-cpp-api', required : false)
 # GFlags
 gflags_dep = dependency('gflags', required: true)
 
+# CLI11 is used only by the explicit `raw` command hierarchy. Legacy invocations
+# continue to use gflags unchanged.
+cli11_dep = dependency('CLI11', required: true, fallback: ['cli11', 'CLI11_dep'])
```
aranadive: This should be required: false. When we will have CLI11-based as the only option then we can trun this to be true.

benchmark/nixlbench/src/utils/meson.build
```diff
@@ -18,12 +18,16 @@ utils_sources = [
   'neuron.h',
   'utils.cpp',
   'utils.h',
+  'raw_cli.cpp',
```
aranadive: These should only be added if the cli11_dep is found.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+        std::ostringstream output;
+        output << std::fixed << std::setprecision(value == static_cast<size_t>(value) ? 0 : 2)
+               << value << ' ' << units[unit] << " (" << bytes << " bytes)";
+        return output.str();
+    }
+
+} // namespace
+
+void
+printRawPosixPlan(const RawPosixRequest &request,
```
aranadive: How are these related to posix as they seem basic nixlbench options?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+bool
+isRawCommand(int argc, char *argv[]) {
+    return argc > 1 && std::string_view(argv[1]) == "raw";
+}
+
+std::optional<size_t>
+parseHumanSize(const std::string &input, std::string &error) {
+    std::string value;
+    value.reserve(input.size());
+    for (unsigned char ch : input) {
```
aranadive: Doesnt CLI11 do any of these checks?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+    PluginMetadata metadata;
+    metadata.name = name;
+    const auto status = agent.getPluginParams(name, metadata.memory_types, metadata.parameters);
+    if (status != NIXL_SUCCESS) {
+        error =
+            "failed to query " + name + " plugin metadata: " + nixlEnumStrings::statusStr(status);
+        return std::nullopt;
+    }
+    return metadata;
```
aranadive: Does the agent leak outside this function?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+        const auto status = agent.getPluginParams(name, metadata.memory_types, metadata.parameters);
+        if (status != NIXL_SUCCESS) {
+            error = "failed to query " + name +
+                " plugin metadata: " + nixlEnumStrings::statusStr(status);
+            return std::nullopt;
+        }
+        if (hasMemoryType(metadata, memory_type)) {
+            matches.push_back(std::move(metadata));
+        }
+    }
```
aranadive: Same question here. Why not create one helper agent to let you do all the discovery?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+    return matches;
+}
+
+int
+parseRawPosixCommand(int argc,
```
aranadive, line 327: Some of these options will be repeated for other plugins as well that support FILE_SEG. I feel like this should be broken up into a generic nixlbench part, seg_type part and plugin part to build the final raw command.

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,530 @@
...
+    raw->add_option("--pipeline-depth", request.raw.pipeline_depth, "Transfer requests in flight")
+        ->group("Raw benchmark options");
+    raw->add_flag(
+           "--check-consistency", request.raw.check_consistency, "Validate transferred bytes")
+        ->group("Raw benchmark options");
+    raw->add_flag("--dry-run", request.raw.dry_run, "Print the resolved plan without executing")
+        ->group("Raw benchmark options");
+
+    if (request.has_file_options) {
+        posix->add_option("--path", request.file.path, "Directory for automatically named files")
```
aranadive: Shouldnt these options be part of the raw commands?

benchmark/nixlbench/src/main.cpp
```diff
@@ -236,3 +237,20 @@ int main(int argc, char *argv[]) {
 
     return worker_ptr->signaled() ? EXIT_FAILURE : EXIT_SUCCESS;
 }
+
+int
+main(int argc, char *argv[]) {
+    if (nixlbench::isRawCommand(argc, argv)) {
+        const auto result = nixlbench::prepareRawCommand(argc, argv, std::cout, std::cerr);
```
aranadive, line 246: Cant you store the plugin_params within nixlbenchconfig if raw mode is supported and used?

benchmark/nixlbench/subprojects/gtest.wrap
```diff
@@ -0,0 +1,16 @@
+[wrap-file]
```
aranadive: Why do we need the gtest wrap? We install libgtest-dev as an additional package in the dockerfile and is a requirement for nixl also. Lets rely on that.
Do you need anything specific from here thats not in the package?

benchmark/nixlbench/tests/meson.build
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+gtest_dep = dependency('gtest',
+                       main: true,
+                       required: true,
```
aranadive: Should be false and nixlbench_tests should be skipped if gtest is missing.

src/plugins/posix/posix_plugin.cpp
```diff
@@ -16,23 +16,46 @@
...
 #include "posix_backend.h"
+#include "io_queue.h"
 #include "backend/backend_plugin.h"
 
 // Plugin type alias for convenience
 using posix_plugin_t = nixlBackendPluginCreator<nixlPosixEngine>;
 
+namespace {
+nixl_b_params_t
+makePosixPluginParams() {
```
aranadive: These can be moved to posix_backend.h?

benchmark/nixlbench/meson_options.txt
```diff
@@ -26,3 +26,4 @@ option('nvshmem_inc_path', type: 'string', value: '', description: 'Path to NVSH
 option('nvshmem_lib_path', type: 'string', value: '', description: 'Path to NVSHMEM library directory')
 option('rocm_path', type: 'string', value: '', description: 'Path to the ROCm installation directory (used only when use_rocm=true). Empty falls back to /opt/rocm.')
 option('use_rocm', type: 'boolean', value: false, description: 'Build nixlbench against AMD ROCm/HIP instead of CUDA. Default false; the NVIDIA path is unchanged.')
+option('build_tests', type: 'boolean', value: true, description: 'Build NIXLBench tests for non-release builds')
```
aranadive, line 30: We also need an option here to build raw path, disabled by default.

aranadive (changes requested):
Thanks for the work! Some comments on clarifying certain aspects and some things can be simplified. The dependencies for libraries should be fixed.

benchmark/nixlbench/subprojects/cli11.wrap
```diff
@@ -0,0 +1,8 @@
+[wrap-git]
```
aranadive: Lets make sure we align the cli11 wrap version with a released one. See here:
[2.7.2-1](https://wrapdb.mesonbuild.com/v2/cli11_2.7.2-1/cli11.wrap)

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    return queryPluginMetadata(agent, name, error);
+}
+
+std::optional<std::vector<PluginMetadata>>
+discoverPluginsWithMemoryType(nixl_mem_t memory_type, std::string &error) {
+    nixlAgent agent("nixlbench-cli", nixlAgentConfig{});
+    std::vector<nixl_backend_t> plugins;
+    if (!listAvailablePlugins(agent, plugins, error)) {
```
aranadive: SHould this check for all plugins or only POSIX?

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,501 @@
...
+        return std::nullopt;
+    }
+
+    std::vector<PluginMetadata> matches;
+    for (const auto &name : plugins) {
+        auto metadata = queryPluginMetadata(agent, name, error);
```
aranadive: ALso here? Since not being able to query a plugin will cause the raw cli to stop nixlbench.

benchmark/nixlbench/tests/posix_integration_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+namespace nixlbench {
+namespace {
+
+    class TemporaryDirectory {
+    public:
+        TemporaryDirectory() {
+            const auto base = std::filesystem::temp_directory_path() / "nixlbench-pr1-XXXXXX";
+            std::string pattern = base.string();
+            pattern.push_back('\0');
```
aranadive: This isnt required

src/plugins/posix/io_queue.h
```diff
@@ -57,6 +57,9 @@ class nixlPosixIOQueue {
     static std::string_view
     getDefaultIoQueueType(void);
 
+    static const uint32_t DEF_IOS_POOL_SIZE;
```
aranadive: Can we define all of these as static constexpr uint32_t? I would move the ones from io_queue.cpp and set them here:
static constexpr uint32_t MIN_IOS_POOL_SIZE     = 64;
static constexpr uint32_t MAX_IOS_POOL_SIZE     = 1024 * 64;
static constexpr uint32_t DEF_IOS_POOL_SIZE     = MAX_IOS_POOL_SIZE;
static constexpr uint32_t MIN_KERNEL_QUEUE_SIZE = 16;
static constexpr uint32_t MAX_KERNEL_QUEUE_SIZE = 1024;
static constexpr uint32_t DEF_KERNEL_QUEUE_SIZE = 256;

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -205,25 +205,29 @@ xferBenchNixlWorker::xferBenchNixlWorker(const std::vector<std::string> &devices
...
-        }
-        std::cout << "POSIX backend with API type: " << xferBenchConfig::posix_api_type
-                  << std::endl;
-        backend_params["ios_pool_size"] = std::to_string(xferBenchConfig::posix_ios_pool_size);
-        backend_params["kernel_queue_size"] =
-            std::to_string(xferBenchConfig::posix_kernel_queue_size);
+        if (!xferBenchConfig::plugin_parameters) {
+            // Preserve the existing flags-only POSIX parameter assembly.
+            if (xferBenchConfig::posix_api_type == XFERBENCH_POSIX_API_AIO) {
+                backend_params["use_aio"] = "true";
```
aranadive, line 211: Can the default value be just false for a backend param?

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Changed this to POSIX-only discovery. When another raw backend is added, dispatch can select that backend first and then query it.

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Addressed by the same change

benchmark/nixlbench/tests/posix_integration_test.cpp
razorback4417: Removed the explicit null append

src/plugins/posix/io_queue.h
razorback4417: All six bounds/defaults are now `static constexpr uint32_t` definitions in `io_queue.h`

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
razorback4417: Yes- legacy path now sends only the selected queue parameter as `true`; unselected selectors are omitted. The POSIX backend reads missing selector parameters with a `false` default

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,456 @@
...
+#include <iomanip>
+#include <map>
+#include <sstream>
+#include <string_view>
+#include <utility>
+
+namespace nixlbench {
+namespace {
+
+    constexpr int kInvalidArgumentsExitCode = 2;
```
aranadive: should be inval_args_exit_code to match codestyle

aranadive (approved):

benchmark/nixlbench/src/utils/raw_cli.cpp
```diff
@@ -0,0 +1,456 @@
...
+    std::vector<std::string> arguments = {
+        program_name,
+        // Fixed values select the existing NIXL/POSIX runner.
+        std::string("--worker_type=") + XFERBENCH_WORKER_NIXL,
+        std::string("--backend=") + XFERBENCH_BACKEND_POSIX,
+        std::string("--initiator_seg_type=") + XFERBENCH_SEG_TYPE_DRAM,
+        // Backend-neutral raw benchmark configuration.
+        "--op_type=" + request.raw.operation,
+        "--check_consistency=" + std::string(boolean(request.raw.check_consistency)),
+        "--total_buffer_size=" + std::to_string(request.raw.total_buffer_size),
```
ofer, line 402: Probably for the next revision but the CLI:App has a help() method that should print this out without having to update this ad-hoc for every change (prone to missing changes)

ofer (approved):

benchmark/nixlbench/src/utils/raw_cli.cpp
razorback4417: Agreed- will be included in the following change.

PR #1942, opened 2026-07-14
fix(posix): recover io_uring after I/O failures

### LINKED ISSUE #1696: POSIX backend (libaio & io_uring): ENOSPC short writes silently reported as DONE (data loss)

lluki:
## What?

Fix `io_uring` submission failures so partial submissions remain in progress and are resubmitted. Permanent ring failures cancel all batches.

Make partial failures cancel and drain all outstanding requests of that batch.

This is a uring focused version of PR #1697, I'll send follow up PR's (#1947 - for aio, already based on this PR) for the other POSIX modes.

  ## Why?

A single failed I/O or a single failed `io_uring_submit()` could leave requests queued indefinitely and disallow the queue to be reused. 

## How?

- CQE user_data now points to a tagged internal object: either the stable `nixlPosixIoUringIO` entry for a normal I/O, or `nixlPosixIoUringCancel` entry for a cancel request. The queue distinguishes normal and cancel CQEs from this tag.
 - Normal I/O entries remain in the fixed `ios_ pool`. Each has `in_flight_` and `cancel_pending_`. It returns to `free_ios_` only when both its normal I/O and, if requested, cancel CQE have been reaped.
 - Cancel work is held in `cancels_to_submit_` until SQEs are available - mirroring the code structure of the normal I/O requests. `prepareSQEs()` prioritizes cancellation work.
 - The request handle tracks expected and observed cancel CQEs. Completion requires all normal I/Os plus all expected cancel CQEs, preventing late CQEs from referencing a released request.
 - `driveSubmissions()` centralizes SQE preparation and `io_uring_submit()` error handling. `poll()` drives submissions before reaping CQEs, while `cancel()` fails unsent work for only the target request and queues cancellation for its prepared/in-flight work.

### Behavior on failure

- `EAGAIN`, `EBUSY`, and `EINTR` from `io_uring_submit()` keep the request `NIXL_IN_PROG`. Ready SQEs remain queued and are retried during later progress.
  - Terminal submission failures abort the process via `NIXL_FATAL`. Returning an error is unsafe because SQEs submitted before the failure may still access user buffers.
  - A failed or short I/O marks only its owning transfer failed. Unprepared I/Os for that transfer fail immediately. prepared/in-flight I/Os receive cancel SQEs.
  - Cancellation is request-scoped: unrelated requests continue progressing on the shared ring.
  - The backend returns the transfer error only after every normal I/O is accounted for and every expected cancel CQE has been reaped, keeping the request handle alive until no later CQE can reference it.

### Tests

- Added io_uring submission-path coverage using a wrapped io_uring_submit():
      - partial submission: remaining SQEs are retained and later submitted;
      - recoverable EAGAIN: progress remains NIXL_IN_PROG and submission is retried;
      - terminal EINVAL: the child process must abort via NIXL_FATAL.

- Added request-scoped cancellation coverage: cancel one request on a shared ring, then post and complete an unrelated request while cancellation cleanup is pending.
- Added an integration case that forces an io_uring short write (RLIMIT_FSIZE) and verifies it is returned as a transfer error.
- Enabled the POSIX io_uring test path in CI. Previously CI exercised the POSIX backend without explicitly running its io_uring mode; it now runs the unit test with -U.

### State Diagram of the I/O object

image

# Scope of this PR

It is deliberately focused on `io_uring`. AIO will be treated in a follow up PR, also user invoked cancellation is out of scope (#1955)


test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -34,6 +34,193 @@
...
+}
+
+int
+runUringSubmissionTests() {
+    struct io_uring probe_ring = {};
+    struct io_uring_params probe_params = {};
+    const int probe_status = io_uring_queue_init_params(kRingEntries, &probe_ring, &probe_params);
+    if (probe_status < 0) {
+        std::cerr << "io_uring runtime probe failed: " << std::strerror(-probe_status) << std::endl;
+        return 1;
```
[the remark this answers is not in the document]
lluki: No this is intended - if uring is requested i want it to fail, not fail back to anything

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -715,6 +902,34 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+            return 1;
+        }
+        struct rlimit limit{page_size, saved.rlim_max};
+        if (setrlimit(RLIMIT_FSIZE, &limit) != 0) {
+            return 1;
+        }
+        status = agent.postXferReq(treq_write);
+        while (status == NIXL_IN_PROG) {
+            status = agent.getXferStatus(treq_write);
+        }
```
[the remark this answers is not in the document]
lluki: CI can be quite slow. It's unclear how that should be done. I want to just rely on CI timeout

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -51,20 +54,40 @@ class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
             void *ctx) override;
     virtual nixl_status_t
     poll(void) override;
+    virtual nixl_status_t
+    cancel(void *ctx) override;
+
+    virtual bool
+    hasPendingCleanup(void) const override {
+        return submission_failure_draining_;
+    }
```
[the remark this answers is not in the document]
lluki: added

src/plugins/posix/io_queue.h
```diff
@@ -52,6 +52,17 @@ class nixlPosixIOQueue {
...
+    // Abort the ios owned by ctx without waiting on the kernel when supported.
+    virtual nixl_status_t
+    cancel(void *ctx) {
+        return NIXL_ERR_NOT_SUPPORTED;
+    }
+
+    virtual bool
+    hasPendingCleanup(void) const {
+        return false;
+    }
```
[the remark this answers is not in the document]
lluki: added

src/plugins/posix/posix_backend.cpp
```diff
@@ -148,22 +154,42 @@ nixlPosixBackendReqH::ioDoneClb(void *ctx, uint32_t data_size, int error) {
...
-    if (status < 0) {
-        return status;
+    if (requestCancellationOnce() == NIXL_ERR_NOT_SUPPORTED) {
+        return queue_result < 0 ? queue_result : NIXL_ERR_BACKEND;
+    }
+
+    if (needsQueueProgress()) {
+        return NIXL_IN_PROG;
     }
+    return transfer_failed_ ? NIXL_ERR_BACKEND : NIXL_SUCCESS;
```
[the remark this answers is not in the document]
lluki: the finding is not valid for current code.

  NIXL_ERR_NOT_SUPPORTED is returned only by queues without cancellation support; io_uring overrides cancel(). Linux AIO and POSIX AIO can return fatal submission/completion errors without generating every remaining callback, so needsQueueProgress() would remain true indefinitely. Returning the terminal error
  immediately preserves their existing behavior until those backends receive dedicated cleanup implementations.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -715,6 +913,34 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+        setrlimit(RLIMIT_FSIZE, &saved);
+        if (status >= 0) {
+            std::cerr << "io_uring short write was not reported" << std::endl;
+            return 1;
+        }
+    }
+#endif
+
```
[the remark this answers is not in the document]
lluki: its a small test program, there are no signal handlers

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -73,58 +97,145 @@ nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t ke
...
-        NIXL_ERROR << "io_uring_submit failed: " << nixl_strerror(-ret);
-        return NIXL_ERR_BACKEND;
+    if (ret >= 0 || ret == -EAGAIN || ret == -EBUSY || ret == -EINTR) {
+        return NIXL_IN_PROG;
     }
 
-    return NIXL_IN_PROG;
+    NIXL_ERROR << "io_uring_submit failed: " << nixl_strerror(-ret);
+    beginSubmissionFailureDrain();
+    return NIXL_ERR_BACKEND;
```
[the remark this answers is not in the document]
lluki: That's something for another PR

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -34,6 +34,204 @@
...
+        fd = mkstemp(path);
+        if (fd < 0) {
+            throw std::runtime_error("mkstemp failed");
+        }
+        unlink(path);
+        for (size_t i = 0; i < buffers.size(); i++) {
+            std::memset(buffers[i].data(), static_cast<int>(i + 1), buffers[i].size());
+        }
+        queue = nixlPosixIOQueue::instantiate("URING", 64, kRingEntries);
+    }
```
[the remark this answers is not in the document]
lluki: it's ok this is just a test script

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -51,20 +55,40 @@ class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
...
+    void
+    beginSubmissionFailureDrain(void);
+    void
+    prepareIOSQEs(void);
+    bool
+    prepareCancelSQEs(void *ctx);
+
     struct io_uring uring; // The io_uring instance for async I/O operations
+    bool submission_failure_draining_;
+    unsigned cancel_sqes_outstanding_;
```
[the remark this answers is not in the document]
lluki: thats what the doc says, but everything in this file is snake case... let's ignore it

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -73,58 +97,147 @@ nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t ke
...
+nixlPosixIOQueueUring::prepareCancelSQEs(void *ctx) {
+    bool prepared = false;
+    for (auto &io : ios_) {
+        if (!io.in_flight_ || io.cancel_queued_ || (ctx && io.ctx_ != ctx)) {
+            continue;
+        }
+
+        struct io_uring_sqe *sqe = io_uring_get_sqe(&uring);
+        if (!sqe) {
+            break;
```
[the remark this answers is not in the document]
lluki: Your interpretation is correct, cancel SQEs are only submitted on a best effort basis. This is not a correctness problem, if cancel SQEs can't be submitted, we will simply wait until the transfer itself has completed.

We could do it, but given that the cancellation is mostly invoked in a degraded ring state, there is a good chance the cancellations also can't be submitted.

yanok (dismissed):
Can we make a targeted test for it? Like: force `prepareCancelSQEs(ctx)` to run out of SQEs, verify the transfer remains `IN_PROG`, then let the uncancelled I/Os complete naturally and verify terminal `NIXL_ERR_BACKEND` only after all callbacks are accounted.

src/plugins/posix/posix_backend.cpp
```diff
@@ -172,8 +198,14 @@ nixlPosixBackendReqH::postXfer() {
         NIXL_ERROR << "POSIX I/O queue is not initialized";
         return NIXL_ERR_BACKEND;
     }
+    if (io_queue_->hasPendingCleanup()) {
```
benlwalker: This isn't right. hasPendingCleanup() returns true if any cancel sqes are outstanding, but I don't think a cancel pending for an entirely different request should cause this request to fail.

src/plugins/posix/io_uring_io_queue.cpp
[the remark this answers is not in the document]
benlwalker: I actually agree with the AI here - this is a real bug introduced by the logic in this patch. For every error that isn't one of the transient errors, we need to special case handle the SQEs that never successfully submitted, rather than attempting to cancel them. Because the cancel won't work. Once a submit fails, we probably need to destroy the entire ring and re-create it.

benlwalker (changes requested):

lluki:
Hi @yanok @benlwalker   I updated the PR quite heavily (see the updated PR description for an overview). In particular we have scoped cancels that only block that request and terminate the process with panic (instead of remaining IN_PROG forever).  It also has some focused tests on that (see PR desc).

src/plugins/posix/io_uring_io_queue.cpp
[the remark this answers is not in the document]
lluki: fixed in latest version

src/plugins/posix/posix_backend.cpp
lluki: fixed in latest version

lluki:
I had to make a separate binary for the tests that instrument `io_uring_submit`. This hack seems incompatible with the asan tests so it must not run there...

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -60,11 +83,27 @@ class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
...
+
     struct io_uring uring; // The io_uring instance for async I/O operations
+    std::list<nixlPosixIoUringIO *> cancels_to_submit_;
+    std::vector<nixlPosixIoUringCancel> cancels_;
 };
 
 nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t kernel_queue_size)
-    : nixlPosixIOQueueImpl<nixlPosixIoUringIO>(ios_pool_size, kernel_queue_size) {
+    : nixlPosixIOQueueImpl<nixlPosixIoUringIO>(ios_pool_size, kernel_queue_size),
+      cancels_(ios_pool_size_) {
```
benlwalker: `cancels_` isn't necessarily the same size as `ios_`, because the base template sizes `ios_` from the raw `ios_pool_size` while this uses the clamped `ios_pool_size_`. This can result in a heap overflow in the loop below.

src/plugins/posix/posix_backend.cpp
```diff
@@ -194,7 +218,7 @@ nixlPosixBackendReqH::postXfer() {
         }
     }
 
-    return io_queue_->post();
+    return queueResult(io_queue_->post());
```
benlwalker, line 238: I/Os already enqueued for this transfer are left in the shared `ios_to_submit_` with `ctx_` still pointing at this handle when `enqueue()` fails partway through the loop above. The caller gets an error and can release the handle, so a later `poll()` driven by an unrelated request calls `ioDoneClb` on freed memory.

src/plugins/posix/posix_backend.h
```diff
@@ -73,6 +62,34 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
             return code_;
         }
     };
+
+private:
+    friend class nixlPosixIOQueueUring;
+
+    bool
+    isComplete() const {
+        return num_confirmed_ios_ == queue_depth_ && cancel_cqes_expected_ == cancel_cqes_seen_;
```
benlwalker: `cancel_cqes_expected_ == cancel_cqes_seen_` is an exact match, and `postXfer()` resets both counters without checking `isComplete()` first. Reposting a handle that still has an outstanding cancel CQE can leave the request in `NIXL_IN_PROG` forever, and the `NIXL_ASSERT` that would catch it is a `DCHECK` that compiles out in release builds.

src/plugins/posix/posix_backend.h
```diff
@@ -73,6 +62,34 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
             return code_;
         }
     };
+
+private:
+    friend class nixlPosixIOQueueUring;
```
benlwalker: `cancel(void *ctx)` declares `ctx` opaque on the generic queue interface, but the io_uring implementation casts it to `nixlPosixBackendReqH *` and writes private counters on it. That inverts the dependency and ties the queue abstraction to one client, which matters now because #1947 copies this pattern for AIO.

.gitlab/test_cpp.sh
```diff
@@ -93,9 +93,9 @@ telePID=$!
 sleep 15
 kill -s INT $telePID
 
-# POSIX test disabled until we solve io_uring and Docker compatibility
-
-./bin/nixl_posix_test -n 128 -s 1048576
+# Forward optional test arguments selected by the CI runner.
+./bin/nixl_posix_test -n 128 -s 1048576 "${@:2}"
```
benlwalker: `-U` switches the entire run to io_uring rather than adding a mode, and this is the only invocation. `getDefaultIoQueueType()` returns `AIO` whenever `HAVE_LINUXAIO` is set, so these matrices no longer cover the default queue type.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -60,11 +83,27 @@ class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
     doCheckCompleted(void);
```
benlwalker, line 81: `getBufInfo()` is declared but never defined, and takes a libaio `struct iocb *` in the io_uring queue. Delete it.

src/plugins/posix/posix_backend.h
lluki: `postXfer` of the backend only gets invoked by if we have completed the previous request. The agent ensures this: https://github.com/ai-dynamo/nixl/blob/main/src/core/nixl_agent.cpp#L1160-L1166

Given that doing this would be a caller error, i can add a `NIXL_ASSERT(isComplete())` at the beginning of posix_backend's `postXfer`.

src/plugins/posix/posix_backend.h
lluki: Right, this is not good. I think the best solution is to mirror the `ioDone` mechanism:

Have the backend invoke a callback (`cancelDone`, `cancelDoneClb`, ...) when a cancellation completes.

Rename the `cancel_cqes_{seen,expected}_` - drop the `cqe` to make them io uring naming independent.
`requestCancellation` will be changed to return how many cancellations have been submitted, the `nixlPosixBackendReqH` remembers this in `cancels_expected`. Backends that do not support cancellation return `0` and never invoke a `cancelDone` callback.

Then we should be backend independent and support cancellation requests.

src/plugins/posix/posix_backend.cpp
lluki: I know, i made some separate issues for this a while back see #1957 and #1955 . I wanted to keep this PR uring focused. Do you think it should be fixed in this PR?

src/plugins/posix/io_uring_io_queue.cpp
lluki: Good catch, will fix

lluki: Why is `ios_` actually not using the clamped size?

.gitlab/test_cpp.sh
lluki: Will fix thanks

src/plugins/posix/io_uring_io_queue.cpp
lluki: will do

src/plugins/posix/io_uring_io_queue.cpp
lluki: There is a PR that changes the size of ios_ to the clamped value: https://github.com/ai-dynamo/nixl/pull/1827

It just hasnt been merged

src/plugins/posix/posix_backend.h
lluki: Implemented in last force push. [diff](https://github.com/ai-dynamo/nixl/compare/17905ce211b741b159548db045029e92dc3321a4..ef4be79894a2d7b40772917194564fea83e2690b)

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -51,20 +73,36 @@ class nixlPosixIOQueueUring : public nixlPosixIOQueueImpl<nixlPosixIoUringIO> {
...
+
     struct io_uring uring; // The io_uring instance for async I/O operations
+    std::list<nixlPosixIoUringIO *> cancels_to_submit_;
+    std::vector<nixlPosixIoUringCancel> cancels_;
 };
 
 nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t kernel_queue_size)
-    : nixlPosixIOQueueImpl<nixlPosixIoUringIO>(ios_pool_size, kernel_queue_size) {
+    : nixlPosixIOQueueImpl<nixlPosixIoUringIO>(ios_pool_size, kernel_queue_size),
+      cancels_(ios_.size()) {
```
lluki, line 101: This makes it work whether #1827 is merged or not

src/plugins/posix/io_uring_io_queue.cpp
lluki: I'm now using `ios_.size()` for `cancels_` so with or without #1827 it is correct

.gitlab/test_cpp.sh
lluki: done

src/plugins/posix/io_uring_io_queue.cpp
lluki: done

src/plugins/posix/io_uring_io_queue.cpp
lluki: Update: i just got #1827 through CI and merged it (it already had 2 approvals)

src/plugins/posix/posix_backend.cpp
benlwalker: No, keep it out. #1957 covers it and the failure predates this patch.

src/plugins/posix/posix_backend.h
benlwalker: Agreed, `postXferReq()` returns `NIXL_ERR_REPOST_ACTIVE` while the handle is in progress, so the counters can't be reset with a cancel outstanding. Dropping this.

benlwalker (dismissed):

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -73,40 +110,83 @@ nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t ke
...
     int ret = io_uring_submit(&uring);
-    if (ret < 0) {
-        NIXL_ERROR << "io_uring_submit failed: " << nixl_strerror(-ret);
-        return NIXL_ERR_BACKEND;
+    if (ret >= 0 || ret == -EAGAIN || ret == -EBUSY || ret == -EINTR) {
+        return NIXL_IN_PROG;
     }
 
-    return NIXL_IN_PROG;
+    NIXL_FATAL << "io_uring_submit failed: " << nixl_strerror(-ret);
```
vvenkates27: NIXL_FATAL is LOG(FATAL) and aborts the process. A submit failure is a backend error, not an unrecoverable invariant. Linux AIO uses NIXL_ERROR and returns NIXL_ERR_BACKEND. The return after NIXL_FATAL is also dead.

Please keep NIXL_ERROR and return NIXL_ERR_BACKEND so the new queueResult / cancel path can drain in-flight I/Os instead of killing the agent.

src/plugins/posix/io_uring_io_queue.cpp
```diff
@@ -73,40 +110,83 @@ nixlPosixIOQueueUring::nixlPosixIOQueueUring(uint32_t ios_pool_size, uint32_t ke
...
-            io_uring_prep_write(sqe, io->fd, io->buf_, io->len_, io->offset_);
+            nixlPosixIoUringIO *io = ios_to_submit_.front();
+            ios_to_submit_.pop_front();
+            if (io->read_) {
+                io_uring_prep_read(sqe, io->fd, io->buf_, io->len_, io->offset_);
+            } else {
+                io_uring_prep_write(sqe, io->fd, io->buf_, io->len_, io->offset_);
+            }
+            io_uring_sqe_set_data(sqe, static_cast<nixlPosixIoUringCQEData *>(io));
+            io->in_flight_ = true;
```
vvenkates27, line 145: Unlike Linux AIO, which pushes unsubmitted I/Os back onto ios_to_submit_, this pops the I/O and sets in_flight_ as soon as an SQE is filled.

If io_uring_submit then fails with a hard error (or you drop NIXL_FATAL as above), that I/O is no longer queued, cancel() will io_uring_prep_cancel a request the kernel may never have seen (-ENOENT), and no I/O CQE will arrive. in_flight_ stays true, releaseIOIfIdle never frees the slot, and isComplete() never becomes true because ioDone never runs. The request can hang in NIXL_IN_PROG.

Suggested fix: only set in_flight_ for SQEs actually submitted, or on submit failure fail those I/Os through the completion callback (same as failQueuedIOs) so the handle can reach a terminal status?

.ci/jenkins/lib/test-matrix.yaml
```diff
@@ -146,7 +146,7 @@ steps:
     run: run
     args:
       jobIdFile: "${JOB_ID_FILE_ROOT}/job_id_${BUILD_NUMBER}.txt"
-      testScript: ".gitlab/test_cpp.sh ${NIXL_INSTALL_DIR}"
+      testScript: ".gitlab/test_cpp.sh ${NIXL_INSTALL_DIR} run_uring"
```
vvenkates27, line 153: Running Uring on Docker was a problem in CI as uring has a priviledged mode requirement. If this is fixed then allowing this is not a problem cc: @aranadive

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -715,6 +717,37 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+            if (ftruncate(file.fd, 0) != 0) {
+                return 1;
+            }
+        }
+        struct rlimit saved{};
+        if (getrlimit(RLIMIT_FSIZE, &saved) != 0) {
+            return 1;
+        }
+        struct rlimit limit{page_size, saved.rlim_max};
+        if (setrlimit(RLIMIT_FSIZE, &limit) != 0) {
```
vvenkates27, line 732: This is process-global, has no timeout, and sits in the middle of the happy-path repost test. If the wait never leaves NIXL_IN_PROG, the process keeps a tiny RLIMIT_FSIZE and SIGXFSZ ignored; later nixl_gusli_test / UCX tests in test_cpp.sh can fail for unrelated reasons?

test/unit/plugins/posix/nixl_posix_uring_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+int
+main() {
+    io_uring probe_ring{};
+    io_uring_params probe_params{};
+    int probe_ret = io_uring_queue_init_params(ring_entries, &probe_ring, &probe_params);
+    if (probe_ret < 0) {
+        std::cerr << "io_uring backend test requires a usable ring: " << std::strerror(-probe_ret)
+                  << " (" << probe_ret << ")" << std::endl;
+        return 1;
+    }
```
vvenkates27, line 170: If io_uring_queue_init_params fails (Docker/seccomp, missing permission), this returns 1 and fails meson/CI. Please skip (exit 0) when a usable ring is not available?

vvenkates27 (commented):
Few comments mostly minor, overall its a good design to address handle cancelation and completions.

src/plugins/posix/io_uring_io_queue.cpp
lluki: In the first version of this PR reviewed by Ben, he requested this behavior (and I think its fair). The problem is that if submit fails with a permanent failure, we lost any guarantees. So reads (modifications of our memory) might happen and we cant be quite sure when that happens. The correct alternative is to keep the pending requests always `IN_PROG`, essentially blocking the process forever. But we thought that it is less favorable than terminating the process.

Note, I think this permanent failures are exceptionally rare. They happen essentially in cases when the kernel OOMs. And i think NIXL is not gracefully handling allocation failures either...

src/plugins/posix/io_uring_io_queue.cpp
lluki: I think the code is correct, because the `sqe` is already in the ring, we are modifying it in place in the ring (see the [origin of sqe](https://github.com/lluki/nixl/blob/eacbe49676ca4b383458ccc55057e3cb33658d01/src/plugins/posix/io_uring_io_queue.cpp#L122)). So from our perspective, the sqe is "in the ring" when we set `in_flight` - we just havent notified the kernel of ring updates yet (what `io_uring_submit` is doing).

Btw, there is also a mode with a kernel side poller thread (I will add this and few other uring modes in a future PR). As far as i know, if this poller thread is enabled, we do not have to call `io_uring_submit` at all, this memory write here is enough for enqueuing the SQE.

.ci/jenkins/lib/test-matrix.yaml
lluki: I think i managed to use uring on the GPU nodes, but not on the cpu-only test runners, maybe something changed there on the infrastructure? I would merge only if these tests pass of course.

test/unit/plugins/posix/nixl_posix_test.cpp
lluki: This only modifies the current process' RLIMIT. That will cause a failure reliably, so we should not sit in NIXL_IN_PROG for long. We restore the RLIMIT and the signal handler in line 741 (and following). It should be ok, no?

test/unit/plugins/posix/nixl_posix_uring_test.cpp
lluki: If this happens, this is a CI misconfiguration, I would like to catch that instead of just not running the tests. In the later case I'm sure in no time the tests wouldnt run anymore at all

src/plugins/posix/io_uring_io_queue.cpp
lluki: @benlwalker

lluki:
Btw, CI is green at this state - the uring tests do work. They just dont work in the CPU-only node ( @vvenkates27 @aranadive ) - but if scoped to the GPU node they do work

src/plugins/posix/io_uring_io_queue.cpp
vvenkates27: I still think we should return `NIXL_ERR_BACKEND` here rather than `NIXL_FATAL`.

`NIXL_FATAL` in a plugin should be reserved for truly unrecoverable cases (e.g. init/connect where the backend cannot operate at all). A failed `io_uring_submit` is a storage I/O error: rare, yes, but from an inference standpoint it is not always fatal. NIXL should report the failure and let the framework decide whether to retry, fail the request, or abort.

Keeping the request stuck in `NIXL_IN_PROG` forever is worse, agreed — but returning `NIXL_ERR_BACKEND` (and letting `queueResult`/cancel drain what it can) is the better middle ground than killing the process.

.ci/jenkins/lib/test-matrix.yaml
vvenkates27: Ok it appears Docker in CI runs in Priviledged mode - so this is addressed.

src/plugins/posix/io_uring_io_queue.cpp
vvenkates27: Thanks — that clarifies it. I was treating `in_flight_` as “kernel-accepted,” but “owned by the ring / not yet reaped” is the right model, including for SQPOLL.

Resolving this. One non-blocking follow-up: if cancel hits -ENOENT on a prepared-but-not-yet-submitted SQE, make sure the slot can’t stay permanently `in_flight_`.

test/unit/plugins/posix/nixl_posix_uring_test.cpp
vvenkates27: Fair point!

test/unit/plugins/posix/nixl_posix_test.cpp
vvenkates27: Got it its local to this process. I think its fine given the failure should be prompt and you restore the limit/handler afterward.

src/plugins/posix/io_uring_io_queue.cpp
lluki: Well the question is when exactly to return the errors. It will be a heuristic and there will be unavoidable corner cases where we violate the buffer ownership - ie. there may be stale memory writes to buffers we have signaled as safe to be reused. Or, depending on the heuristic, also a stuck forever in IN_PROG.  Imo this is the worst of all the options, it may introduce (very) hard to debug errors, if randomly a delayed read overwrites some process memory...


But ok, if we really want it here's a suggestion for a heuristic: 

```
On terminal error of io_uring_submit: mark the ring as terminal error with a flag, return IN_PROG for now.

On poll (of each request):
 * (Already exists): handle CQEs, check if completed, if completed return success.
 * (NEW) If ring terminal error flag is set, return NIXL_ERR_BACKEND
 * (Already exists): Return IN_PROG
```

Btw, NIXL should have an error code `NIXL_ERR_BACKEND_FATAL` that tells the client that it went wrong and that it is not free to reuse the buffer. I was thinking about this in other contexts too

## Currently unhandled errors
Here's a list of errors we treat as fatal in the current version:

| error code | class | desc |
|----|----|----|
| `EBADF` | terminal | The ring file descriptor is no longer valid. This indicates broken ring lifetime or corrupted queue state. |
| `EBADR` | terminal | At least one CQE was dropped under critical resource pressure. Request completion accounting can no longer be trusted. |
| `ENOMEM` | terminal | The kernel failed an internal allocation, such as attaching a new submitting task to the ring. Retrying immediately provides no reliable progress mechanism. |
| `ENXIO` | terminal | The io_uring instance is being torn down and cannot accept further submissions. |
| `EOPNOTSUPP` | terminal | The stored descriptor does not refer to an io_uring instance, indicating invalid or corrupted ring state. |

I think it's fair to say that these are either code bugs or out of memory errors. In both cases the system is in serious trouble.


## Correct IO URING cleanup

I was investigating if  it even is possible to correctly cleanup the ring after EBADR (lost CQE):

On kernels < 6:
> On kernels without IORING_REGISTER_SYNC_CANCEL, including the current 5.15 test environment, there is no general correctness-preserving in-process recovery for caller-owned buffers. Closing and recreating the ring is insufficient because teardown is asynchronous. In that case, treating EBADR as process-fatal is the safe behavior.

On kernels >6: There is  `io_uring_register_sync_cancel` which helps to do correct cleanup, but the `io_uring_register_sync_cancel` can itself fail, at which point we'd have to panic again. So there is not "fatal free" path either...

lluki: I pushed a commit with the suggested heuristic (as separate commit), but I prefer the panic variant

lluki:
I verified the current state of the branch f914f1bc3d814afd5a1c38db0429b663ffa95746 integrated in SGLang: Testsuite passes, a quick e2e test works, and i prototyped the hard-limit reclaimer i want to do (if disk full, reclaim file, retry). All of this works correctly.

Patched SGLang: https://github.com/lluki/sglang/pull/new/fix/hicache-nixl-write-retry

vvenkates27 (approved):

dpressle (approved):

PR #1945, opened 2026-07-15
telemetry: extract bounded staging queue

e-eygin:
## What?

Extracts the bounded, mutex-backed telemetry staging queue out of
`nixlTelemetry` into a focused internal class,
`nixlTelemetryStagingQueue` (`src/core/telemetry/telemetry_staging_queue.{h,cpp}`),
with no change in behavior.

The new class owns all queue mechanics:
- event storage and its capacity reserve,
- the producer/consumer mutex,
- capacity enforcement,
- all-or-none single (`tryPush`) and batch (`tryPushBatch`) insertion,
- the swap-drain (`takePending`),
- the staging-drop counter (`exchangeDropped`).

`nixlTelemetry` keeps everything metric-specific: per-metric activation,
event selection, periodic scheduling, exporter dispatch, and synthetic
dropped-event construction. `updateData()`, `addXferStats()`,
`flushPendingEvents()`, and `exportDroppedEvents()` are rewired onto the new
interface.

Adds focused and concurrent-producer unit tests under
`test/gtest/unit/telemetry/`, wired into the existing CTest `unit` target.

No public API, metric, exporter, or BUFFER schema change.

## Why?

`nixlTelemetry` mixed two independent responsibilities: mapping telemetry
operations to events/exporting them, and implementing a bounded
multi-producer/single-consumer queue. Isolating the queue behind a stable
interface makes the queue mechanics independently testable and provides the
seam for the follow-up low-lock implementation (NIX-1544) to replace the queue
without touching metric logic.

Tracking: NIX-1542 (Perf-3, parent NIX-1359).

## How?

Behavior is preserved exactly:
- Metric activation stays at the producer — deactivated event types never enter
  the queue and are not counted as drops.
- Single-event and variable-size batch insertion keep all-or-none semantics
  under one mutex acquisition; a rejected batch increments drops by its length,
  and a zero-length batch is a no-op with no lock and no drop-accounting change.
- The consumer swaps the live vector for a fresh capacity-reserved one under the
  mutex and exports outside the lock; drained order is preserved.
- The synthetic `AGENT_TELEMETRY_EVENTS_DROPPED` event still bypasses the queue.
- Shutdown is unchanged (no final flush added).

The interface uses pointer + count for batch insertion so it stays C++17-clean
(no `std::span` dependency); `addXferStats()` builds its activated subset in a
fixed-size stack array and submits the populated prefix with no allocation.

Validated: staging-queue unit tests plus existing core-telemetry, Prometheus,
and DOCA telemetry suites pass on the C++17 and C++20 builds; the extracted
queue (including the concurrent-producer test) passes under ThreadSanitizer and
AddressSanitizer.

## Summary by CodeRabbit

- **Improvements**
  - Upgraded telemetry event buffering to a bounded staging queue with improved concurrent producers handling.
  - Flush now drains queued events consistently while applying a “drop-newest” policy when capacity is exceeded.
  - Transfer statistics are queued in all-or-nothing batches to avoid partial updates.
- **Reliability**
  - More accurate dropped-event accounting derived from queue behavior, including proper reset after reporting.
- **Tests**
  - Added unit coverage for capacity, ordering, batching, draining, drop-counter reset, and multi-producer concurrency invariants.
- **Chores**
  - Updated build configuration to compile the staging queue and include related unit tests.

### DIFF
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -16,6 +16,7 @@
  */
 #include <array>
 #include <chrono>
+#include <span>
 #include <sstream>
 #include <thread>
 #include <filesystem>
@@ -292,7 +293,7 @@ nixlTelemetry::exportDroppedEvents() {
     // staging queue, it can never itself be staging-dropped; emitting only a
     // positive count keeps the no-overflow path event-free (preserving
     // exact-count contracts).
-    const uint64_t dropped = stagingQueue_.exchangeDropped();
+    const uint64_t dropped = stagingQueue_.takeNumDropped();
     if (dropped > 0 &&
         isMetricEnabled(nixl_telemetry_event_type_t::AGENT_TELEMETRY_EVENTS_DROPPED)) {
         exporter_->exportEvent(
@@ -324,8 +325,9 @@ nixlTelemetry::updateData(nixl_telemetry_event_type_t event_type, uint64_t value
     if (!isMetricEnabled(event_type)) {
         return;
     }
-    // agent can be multi-threaded; the queue performs its own drop accounting.
-    stagingQueue_.tryPush({event_type, value});
+    // agent can be multi-threaded; the queue performs its own drop accounting,
+    // so the accepted/dropped result is intentionally not consumed here.
+    (void)stagingQueue_.tryPush({event_type, value});
 }
 
 // TODO: the next 4 update* methods may be removable -- addXferStats covers them.
@@ -408,5 +410,7 @@ nixlTelemetry::addXferStats(std::chrono::microseconds xfer_time,
         batch[count++] = {requests_type, 1};
     }
 
-    stagingQueue_.tryPushBatch(batch.data(), count);
+    // The queue performs its own drop accounting, so the accepted/dropped
+    // result is intentionally not consumed here.
+    (void)stagingQueue_.tryPushBatch(std::span{batch}.first(count));
 }
--- a/src/core/telemetry/telemetry.cpp
+++ b/src/core/telemetry/telemetry.cpp
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <array>
 #include <chrono>
 #include <sstream>
 #include <thread>
@@ -111,12 +112,11 @@ nixlTelemetry::nixlTelemetry(const std::string &agent_name, const std::string &e
       maxBufferedEvents_(resolveMaxBufferedEvents()),
       exporter_(makeExporter(exporter_name)),
       metricEnabled_(resolveEnabledMetrics()),
+      stagingQueue_(maxBufferedEvents_),
       pool_(1),
       writeTask_(pool_.get_executor(), DEFAULT_TELEMETRY_RUN_INTERVAL, false) {
     // A constructed nixlTelemetry always has an exporter (makeExporter throws
-    // otherwise), so reserve the event buffer up front and start the periodic
-    // export task.
-    events_.reserve(maxBufferedEvents_);
+    // otherwise), so start the periodic export task.
     startExportTask();
 
     if (!nixlTime::fastClockUsesHwCounter()) {
@@ -272,14 +272,9 @@ nixlTelemetry::startExportTask() {
 
 bool
 nixlTelemetry::flushPendingEvents() {
-    std::vector<nixlTelemetryEvent> next_queue;
-    next_queue.reserve(maxBufferedEvents_);
-    {
-        std::lock_guard<std::mutex> lock(mutex_);
-        events_.swap(next_queue);
-    }
+    auto pending = stagingQueue_.takePending();
 
-    for (auto &event : next_queue) {
+    for (auto &event : pending) {
         // Deactivated metrics never enter the queue (filtered at the source), so
         // everything here is already exportable.
         exporter_->exportEvent(event);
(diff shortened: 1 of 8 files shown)


e-eygin:

src/core/telemetry/telemetry_staging_queue.cpp
```diff
@@ -0,0 +1,67 @@
...
+    return true;
+}
+
+bool
+nixlTelemetryStagingQueue::tryPushBatch(const nixlTelemetryEvent *events, size_t count) {
+    if (count == 0) {
+        return true;
+    }
+    const std::lock_guard<std::mutex> lock(mutex_);
+    if (events_.size() + count > capacity_) {
```
guy-ealey-morag:
```suggestion
    if (count > capacity_ - events_.size()) {
```

Technically safer to avoid overflow

src/core/telemetry/telemetry.h
guy-ealey-morag:
```suggestion
```

I think these are not needed anymore

src/core/telemetry/telemetry.cpp
```diff
@@ -398,24 +388,25 @@ nixlTelemetry::addXferStats(std::chrono::microseconds xfer_time,
...
-    // Atomic per-transfer batch: all activated events land together or none do.
-    const std::lock_guard lock(mutex_);
-    if (events_.size() + batch_size > maxBufferedEvents_) {
-        droppedEvents_.fetch_add(batch_size, std::memory_order_relaxed);
-        return;
-    }
+    // Build the activated subset in a fixed-size stack array (post time, transfer
+    // time, bytes, request count) and submit it as one all-or-none batch; the
+    // queue accepts every event or drops the whole batch.
+    std::array<nixlTelemetryEvent, 4> batch;
```
guy-ealey-morag, line 396: Notice that now the events array is built even if they are going to be dropped, is it intentional?

e-eygin: Intentional — the batch must be materialized to submit it atomically (capacity is only known under the lock). It's a 4-slot stack array of trivial `{enum, uint64}` values, so a dropped batch just costs a few plain value copies; no heap, no hot-path impact.

src/core/telemetry/telemetry_staging_queue.cpp
e-eygin: Good catch, applied in `b923ec4f`.

src/core/telemetry/telemetry.h
e-eygin: Removed in `b923ec4f`.

src/core/telemetry/telemetry.cpp
```diff
@@ -297,7 +292,7 @@ nixlTelemetry::exportDroppedEvents() {
     // staging queue, it can never itself be staging-dropped; emitting only a
     // positive count keeps the no-overflow path event-free (preserving
     // exact-count contracts).
-    const uint64_t dropped = droppedEvents_.exchange(0, std::memory_order_relaxed);
+    const uint64_t dropped = stagingQueue_.exchangeDropped();
```
guy-ealey-morag: If we already touch the code here, I suggest changing the terminology from `dropped` to `num_dropped / exchangeNumDropped / numDroppedEvents_`  to better indicate that it's the number and not the events themselves. 
wdyt?

src/core/telemetry/telemetry.cpp
```diff
@@ -398,24 +388,25 @@ nixlTelemetry::addXferStats(std::chrono::microseconds xfer_time,
...
     if (bytes_on) {
-        events_.emplace_back(bytes_type, bytes);
+        batch[count++] = {bytes_type, bytes};
     }
     if (requests_on) {
-        events_.emplace_back(requests_type, 1);
+        batch[count++] = {requests_type, 1};
     }
+
+    stagingQueue_.tryPushBatch(batch.data(), count);
```
guy-ealey-morag: I suggest refactoring this function to be more C++, it could be defined as:
`bool tryPushBatch(std::span<const nixlTelemetryEvent> events)`
and then the call would be:
`stagingQueue_.tryPushBatch(std::span{batch}.first(count));`

src/core/telemetry/telemetry.cpp
e-eygin: Done in `f3cd5f96`.

src/core/telemetry/telemetry.cpp
e-eygin: Done in `f3cd5f96`.

guy-ealey-morag (approved):

src/core/telemetry/telemetry_staging_queue.cpp
```diff
@@ -0,0 +1,67 @@
...
+        return false;
+    }
+    events_.insert(events_.end(), events.begin(), events.end());
+    return true;
+}
+
+std::vector<nixlTelemetryEvent>
+nixlTelemetryStagingQueue::takePending() {
+    std::vector<nixlTelemetryEvent> pending;
+    pending.reserve(capacity_);
```
ColinNV, line 51: This could be further optimized by switching between two pre-allocated buffers that are not `std::vector` (`nixlTelemetryEvent` owns nothing, clearing a buffer would be just setting a `used` to zero). Not sure whether it's worth it though.

ColinNV (approved):

src/core/telemetry/telemetry_staging_queue.cpp
e-eygin: Good idea — filed as NIX-1641 to do with its own review/TSan pass, keeping this PR the behavior-preserving extraction. Thanks!

PR #1949, opened 2026-07-15
libfabric: PT-owns-endpoint with MPSC lock-free ring

oa-aws:
libfabric: Move fi_writemsg/fi_readmsg posting into the Progress Thread (when progress thread is enabled) via a lock-free MPSC ring buffer, eliminating ep_mutex_ contention on the data path, resulting in performance improvement with larger batch sizes. Main thread enqueues to per-rail MpscRing; PT dequeues, posts, and polls completions in a single thread.

- MpscRing<T,2048>: CAS multi-producer, single-consumer ring
- drainPostQueue: PT posts with FI_MORE batching, interleaves CQ
- cudaSetDevice in PT for multi-GPU VRAM buffer validation
- Graceful error: completion callback on failure, continue draining

## Summary by CodeRabbit

- **Performance**
  - Added a deferred, progress-thread-aware per-rail post queue that batches requests and interleaves completion draining with retry on temporary backpressure.
  - Introduced configurable post-queue sizing (default: 2048) to improve transfer throughput under progress-thread mode.

- **Reliability**
  - Strengthened CUDA context validation and selection for CUDA VRAM transfers, with more defensive handling of mismatches and query failures.
  - Improved handling of `-FI_EAGAIN` so progress continues via CQ drain/retry instead of stopping early.
  - Enabled per-transfer device targeting for multi-GPU CUDA operations.

- **Chores**
  - Added clearer error reporting when per-rail post queue initialization fails in progress-thread mode.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -260,16 +260,28 @@ nixlLibfabricCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &wa
 
     was_updated = false;
 
-    if (expected_dev == -1) return -1;
-    if (myDevId_ != -1 && expected_dev != myDevId_) return -1;
+    if (expected_dev == -1) {
+        return -1;
+    }
+    if (myDevId_ != -1 && expected_dev != myDevId_) {
+        return -1;
+    }
 
     ret = cudaQueryAddr(address, is_dev, dev, ctx, pci_bus_id);
-    if (ret) return ret;
-    if (!is_dev) return 0;
-    if (dev != expected_dev) return -1;
+    if (ret) {
+        return ret;
+    }
+    if (!is_dev) {
+        return 0;
+    }
+    if (dev != expected_dev) {
+        return -1;
+    }
 
     if (pthrCudaCtx_) {
-        if (pthrCudaCtx_ != ctx) return -1;
+        if (pthrCudaCtx_ != ctx) {
+            return -1;
+        }
         return 0;
     }
 
@@ -283,15 +295,41 @@ nixlLibfabricCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &wa
 int
 nixlLibfabricCudaCtx::cudaSetCtx() {
     CUresult result;
-    if (NULL == pthrCudaCtx_) return 0;
+    if (NULL == pthrCudaCtx_) {
+        return 0;
+    }
 
     result = cuCtxSetCurrent(pthrCudaCtx_);
     return (CUDA_SUCCESS == result);
 }
 
+class nixlLibfaricCudaCtxEngineMediator : public LibfabricUtils::nixlLibfaricCudaCtxMediator {
+public:
+    nixlLibfaricCudaCtxEngineMediator(nixlLibfabricEngine *engine) : engine_(engine) {}
+
+    ~nixlLibfaricCudaCtxEngineMediator() override {}
+
+    nixl_status_t
+    cudaSetCtx(bool &use_cuda_addr_wa) override {
+        if (engine_ == nullptr) {
+            return NIXL_ERR_INVALID_PARAM;
+        }
+        return engine_->vramApplyCtxEx(use_cuda_addr_wa);
+    }
+
+private:
+    nixlLibfabricEngine *engine_;
+};
+
 void
 nixlLibfabricEngine::vramInitCtx() {
     cudaCtx_ = std::make_unique<nixlLibfabricCudaCtx>();
+
+    // install a mediator so that the progress thread can also use this
+    // NOTE: the mediator is stateless and therefore it cannot be involved in a race
+    std::unique_ptr<LibfabricUtils::nixlLibfaricCudaCtxMediator> mediator;
+    mediator.reset(new (std::nothrow) nixlLibfaricCudaCtxEngineMediator(this));
+    setCudaCtxMediator(std::move(mediator));
 }
 
 int
@@ -328,6 +366,18 @@ void
 nixlLibfabricEngine::vramFiniCtx() {
     const std::lock_guard<std::mutex> lock(cuda_ctx_mutex_);
     cudaCtx_.reset();
+    LibfabricUtils::clearCudaCtxMediator();
+}
+
+nixl_status_t
+nixlLibfabricEngine::vramApplyCtxEx(bool &use_cuda_addr_wa) const {
+    const std::lock_guard<std::mutex> lock(cuda_ctx_mutex_);
+    use_cuda_addr_wa = cuda_addr_wa_;
+    if (use_cuda_addr_wa && cudaCtx_ && !cudaCtx_->cudaSetCtx()) {
+        NIXL_ERROR << "Failed to set CUDA context before posting descriptors";
+        return NIXL_ERR_BACKEND;
+    }
+    return NIXL_SUCCESS;
 }
 #endif
 
@@ -500,8 +550,18 @@ nixlLibfabricEngine::nixlLibfabricEngine(const nixlBackendInitParams *init_param
 
         // Start Progress thread for rail completion processing
         if (progress_thread_enabled_) {
+            // in case of PT=1 we need to allocate post ring buffer per rail
+            size_t post_queue_size = NIXL_LIBFABRC_DEFAULT_POST_QUEUE_SIZE;
+            LibfabricUtils::getCustomIntParam(
+                getCustomParams(), "post_queue_size", post_queue_size);
+
             for (size_t i = 0; i < rail_manager_.getNumRails(); ++i) {
                 rail_manager_.getRail(i).setProgressThreadEnabled(true);
+                if (!rail_manager_.getRail(i).initPostQueue(post_queue_size)) {
+                    NIXL_ERROR << "Failed to initialize post-queue for rail " << i;
+                    throw std::runtime_error("Failed to initialize the rail manager: unable to "
+                                             "initialize rail post queue");
+                }
             }
 
             NIXL_INFO << "Starting Progress thread for rails with delay: "
@@ -1211,31 +1271,19 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
     submitted_count = 0;
 
 #ifdef HAVE_CUDA
+    // NOTE: when progress thread is enabled and the call is deferred via ring-buffer, this should
+    // take place in the context of the progress thread
     const bool is_cuda_vram = local.getType() == VRAM_SEG && runtime_ == FI_HMEM_CUDA;
     bool use_cuda_addr_wa = false;
     int current_cuda_device = -1;
-    if (is_cuda_vram) {
-        const std::lock_guard<std::mutex> lock(cuda_ctx_mutex_);
-        use_cuda_addr_wa = cuda_addr_wa_;
-        if (use_cuda_addr_wa && cudaCtx_ && !cudaCtx_->cudaSetCtx()) {
-            NIXL_ERROR << "Failed to set CUDA context before posting descriptors";
-            return NIXL_ERR_BACKEND;
+    if (!progress_thread_enabled_ && is_cuda_vram) {
+        nixl_status_t status = vramApplyCtxEx(use_cuda_addr_wa);
+        if (status != NIXL_SUCCESS) {
+            return status;
         }
     }
-
-    auto prepare_cuda_descriptor_post = [&](int device_id, int desc_idx) -> nixl_status_t {
-        if (!is_cuda_vram || use_cuda_addr_wa || device_id == current_cuda_device) {
-            return NIXL_SUCCESS;
-        }
-        cudaError_t cuda_ret = cudaSetDevice(device_id);
-        if (cuda_ret != cudaSuccess) {
-            NIXL_ERROR << "Failed to set CUDA device " << device_id << " while posting descriptor "
-                       << desc_idx << ": " << cudaGetErrorString(cuda_ret);
-            return NIXL_ERR_BACKEND;
-        }
-        current_cuda_device = device_id;
-        return NIXL_SUCCESS;
-    };
+#else
+    const bool is_cuda_vram = false;
 #endif
 
     // A FI_MORE post rings no doorbell; a rail's queued batch is only submitted by a later
@@ -1270,9 +1318,18 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
         int device_id = local[desc_idx].devId;
 
 #ifdef HAVE_CUDA
-        if (nixl_status_t status = prepare_cuda_descriptor_post(device_id, desc_idx);
-            status != NIXL_SUCCESS) {
-            return status;
+        // NOTE: when progress thread is enabled and the call is deferred via ring-buffer, this
+        // should take place in the context of the progress thread
+        if (!progress_thread_enabled_ && is_cuda_vram && !use_cuda_addr_wa &&
+            device_id != current_cuda_device) {
+            cudaError_t cuda_ret = cudaSetDevice(device_id);
+            if (cuda_ret != cudaSuccess) {
+                NIXL_ERROR << "Failed to set CUDA device " << device_id
+                           << " while posting descriptor " << desc_idx << ": "
+                           << cudaGetErrorString(cuda_ret);
+                return NIXL_ERR_BACKEND;
+            }
+            current_cuda_device = device_id;
         }
 #endif
 
@@ -1300,7 +1357,9 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             desc_submitted_count,
             desc_idx,
             xfer_base_offset,
-            apply_fi_more);
+            apply_fi_more,
+            local[desc_idx].devId,
+            is_cuda_vram);
 
         if (status != NIXL_SUCCESS) {
             NIXL_ERROR << "prepareAndSubmitTransfer failed for descriptor " << desc_idx
--- a/src/plugins/libfabric/libfabric_backend.h
+++ b/src/plugins/libfabric/libfabric_backend.h
@@ -356,6 +356,11 @@ private:
     vramApplyCtx();
     void
     vramFiniCtx();
+
+    // same as vramApplyCtx, but with additional output parameter
+    nixl_status_t
+    vramApplyCtxEx(bool &use_cuda_addr_wa) const;
+    friend class nixlLibfaricCudaCtxEngineMediator;
 #endif
 
 public:
--- a/src/utils/libfabric/libfabric_common.cpp
+++ b/src/utils/libfabric/libfabric_common.cpp
@@ -298,4 +298,24 @@ getCustomIntParam(const nixl_b_params_t &custom_params, const std::string &key, 
     return NIXL_SUCCESS;
 }
 
+static std::unique_ptr<nixlLibfaricCudaCtxMediator> cuda_ctx_mediator;
+
+void
+setCudaCtxMediator(std::unique_ptr<nixlLibfaricCudaCtxMediator> &&mediator) {
+    cuda_ctx_mediator = std::move(mediator);
+}
+
+void
+clearCudaCtxMediator() {
+    cuda_ctx_mediator.reset(nullptr);
+}
+
+nixl_status_t
+cudaSetCtx(bool &use_cuda_addr_wa) {
+    if (cuda_ctx_mediator.get() == nullptr) {
+        return NIXL_ERR_INVALID_PARAM;
+    }
+    return cuda_ctx_mediator->cudaSetCtx(use_cuda_addr_wa);
+}
+
 } // namespace LibfabricUtils
--- a/src/utils/libfabric/libfabric_common.h
+++ b/src/utils/libfabric/libfabric_common.h
@@ -41,6 +41,7 @@
 #define NIXL_LIBFABRIC_CQ_SREAD_TIMEOUT_MS 10
 #define NIXL_LIBFABRIC_DEFAULT_STRIPING_THRESHOLD (128 * 1024) // 128KB
 #define LF_EP_NAME_MAX_LEN 56
+#define NIXL_LIBFABRC_DEFAULT_POST_QUEUE_SIZE (32 * 1024) // 32K MPSC entries
 
 // Number of consecutive WRITE descriptors batched onto one rail with FI_MORE before flushing.
 #define NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE 16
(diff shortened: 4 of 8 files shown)


src/utils/libfabric/libfabric_rail_manager.h
```diff
@@ -219,7 +219,8 @@ class nixlLibfabricRailManager {
                              size_t &submitted_count_out,
                              int desc_idx,
                              int desc_count,
-                             size_t base_offset);
+                             size_t base_offset,
+                             int device_id = -1);
```
fengjica: Update comment above?

src/plugins/libfabric/libfabric_backend.cpp
fengjica, line 1261: postXferDescriptors() is called in two code paths today. One of them is one where these post tasks are done in a thread pool: https://github.com/ai-dynamo/nixl/blob/main/src/plugins/libfabric/libfabric_backend.cpp#L1318-L1323.

I guess if we post the tasks to the MPSC ring, it does not make sense to go through that thread pool. Right?

I don't think any immediate code change needs to be done, as that thread pool is not enabled by default today. Just want to point that out and hear your thoughts.

src/utils/libfabric/libfabric_rail_manager.cpp
fengjica, line 407: This change now is for non-striping case. Is there a reason that we should apply them different for striping case below?

src/utils/libfabric/libfabric_rail.h
```diff
@@ -380,6 +473,11 @@ class nixlLibfabricRail {
     // EP mutex to protect endpoint and CQ operations
     mutable std::mutex ep_mutex_;
 
+    // Lock-free MPSC ring: main thread pushes, PT pops and posts
+    static constexpr size_t POST_RING_CAPACITY = 2048;
```
fengjica: Is this big enough to hold the buffer if the draining thread (1 PT today across multiple rails) cannot catch up with the speed of the posting thread? And there can be multiple posting threads (e.g. https://github.com/ai-dynamo/nixl/blob/main/benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp#L1685).

Some of our workloads uses very large batch size, which translates to large post request sizes. Maybe worth running one perf test, or just nixlbench with a big batch size.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -21,6 +21,10 @@
 #include "serdes/serdes.h"
 #include "libfabric_common.h"
 
+#ifdef HAVE_CUDA
+#include <cuda_runtime.h>
```
fengjica, line 25: Is this included here because some new cuda runtimes functions are added?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1063,6 +1072,7 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
         }
 
         if (ret == -FI_EAGAIN) {
+            // SQ full - do inline CQ progress to free slots, then retry
```
fengjica: This comment is probably not accurate, as CQ progress is conditioned on !progress_thread_enabled_, right?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
 }
 
+/****************************************
+ * PT-owns-endpoint: post queue methods
+ ****************************************/
+
+void
+nixlLibfabricRail::enqueuePost(const nixlLibfabricPostRequest &req) {
+    while (!post_ring_.push(req)) {
+        std::this_thread::yield();
```
fengjica: Is it better to yield and retry, or abort to let let users know that something is full (and let user retry)?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+    }
+}
+
+nixl_status_t
+nixlLibfabricRail::drainPostQueue() {
+    nixlLibfabricPostRequest pr;
+    int posted = 0;
+    [[maybe_unused]] int current_cuda_device = -1;
```
fengjica: not used anywhere?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+
+                ret = fi_readmsg(endpoint, &msg, pr.fi_flags);
+            }
+
+            if (ret == -FI_EAGAIN) {
+                struct fi_cq_data_entry cq_buf[16];
+                int cq_ret = fi_cq_read(cq, cq_buf, 16);
+                if (cq_ret > 0) {
+                    for (int c = 0; c < cq_ret; c++) {
+                        processCompletionQueueEntry(&cq_buf[c]);
```
fengjica: Could we defer processing them so that post can be drained faster? Just a thought, not necessary for this PR.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+        if (ret != 0) {
+            NIXL_ERROR << "drainPostQueue: post failed ret=" << ret << " (" << fi_strerror(-ret)
+                       << ") on rail " << rail_id;
+            if (pr.req && pr.req->completion_callback) {
+                pr.req->completion_callback();
+            }
+            continue;
+        }
+        posted++;
+        if (posted % 32 == 0) {
```
fengjica: I guess this 32 is something that can be tuned later to find an optimal ratio of draining vs reading cq, is that the case?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+    }
+}
+
+nixl_status_t
+nixlLibfabricRail::drainPostQueue() {
+    nixlLibfabricPostRequest pr;
+    int posted = 0;
+    [[maybe_unused]] int current_cuda_device = -1;
+
+    while (post_ring_.pop(pr)) {
```
fengjica: This while loop will not break until the post_ring is empty. With one PT, it's possible that it's stuck here, and do not get a chance to progress other rails, right?

fengjica (commented):
Reviewed the change except for MPSC ring implementation.

The major question is still that when one PT drains all rings (of multiple rails), the unbounded while loop in draining can block it from progressing other rails.

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: The case of a thread-pool may resemble multiple user threads, so we are expected to see meow pressure on the MPSC ring buffer, which currently does not scale well with multiple threads. There a re a few solutions for that (SPSC per-thread, separate ring buffer and endpoint per user thread, to name a few). Right now I think this is out of scope, as this PR mainly addresses the performance issue with one user thread.

src/utils/libfabric/libfabric_rail_manager.cpp
oa-aws: Yes, we can add that too for non-striping. Actually I think the performance issue should be more acute in this case. Fixing.

src/utils/libfabric/libfabric_rail.h
oa-aws: I agree. I actually don't like it being compile time constant. Maybe it is better to have it configurable with some large default according to perf tests?

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I just took it form the original PR. I think it was in preparation for setting cuda context per deferred request in the progress thread (whereas now this is done in nixlLibfabricEngine::postXferDescriptors before calling rail_manager_.prepareAndSubmitTransfer, which assumes request execution in the current thread). This change is also required (i.e. when PT=1 do not set CUDA context in the engine, but rather let the PT do that).

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Right, this is from the original PR. I guess it was some comment during work (or some copy mistake on my side). Removing.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: It should be used, see comments above.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: This eventually increments an atomic integer. Unless there is other heavy computation, I don't think we can gain much. The question is, is there a use case where a completion callback would be heavy (i.e. someone developing on top of the libfabric plugin)? If so, then this should have a design, because I don't think it is trivial (e.g. does someone rely on counters to catch up to tell request is finished? If so, would deferring cause unwanted delays? Maybe user should declare whether the callback should be deferred? etc.).

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I think so. This is coming from the original PR, and I was wondering when to benchmark fine-tuning of it. We should also (as you mentioned above) enforce some batch-size limit so that post requests from other rails are executed in a timely manner. This budgeting is not trivial though, so I suggest for this PR we put some constants that work reasonably, and for next PRs we improve on that based on perf test results.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Correct. I was worried about that too. This is how it was done in the original PR, and we need to think about a solution. See previous comment. Currently adding some compile time constant as a batch limit.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I changed the implementation due to code rabbit comments, but the question remains the same: what should we do in case of a full ring buffer? I think the best option for now is to retry a few times until some request timeout is reached and let user know we failed, so the ring buffer size can be configured to larger value for the workload in question.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+            }
+
+            if (ret == -FI_EAGAIN) {
+                struct fi_cq_data_entry cq_buf[16];
+                int cq_ret = fi_cq_read(cq, cq_buf, 16);
+                if (cq_ret > 0) {
+                    for (int c = 0; c < cq_ret; c++) {
+                        processCompletionQueueEntry(&cq_buf[c]);
+                    }
+                }
```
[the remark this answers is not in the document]
oa-aws: I think this is incorrect. In case of other errors, the inner while loop exits, next an ERROR is issued after the loop, and the completion callback is called for the request (and the request is dropped). Error cannot be returned, since subsequent requests must be served. The only fix required here, is to add error message when fi_cq_read() fails, as done in other places.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1099,6 +1109,111 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+            }
+        }
+
+        if (ret != 0) {
+            NIXL_ERROR << "drainPostQueue: post failed ret=" << ret << " (" << fi_strerror(-ret)
+                       << ") on rail " << rail_id;
+            if (pr.req && pr.req->completion_callback) {
+                pr.req->completion_callback();
+            }
+            continue;
```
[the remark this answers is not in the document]
oa-aws: I think this is incorrect, otherwise the call to check whether a multi-part/segment request finished will never declare success, as the counters will never match up if a completion is missing, even if failed. In addition to that, maybe a fix should be added to tell the total request status, in case some sub-segments posting failed.

src/utils/libfabric/libfabric_rail.cpp
fengjica: You are right.

Upon checking the callbacks today in the code, I also notice a bug. Because I was thinking a notif needs to be sent out after all completions are done, for READ. I only found one place in the code in checkXfer() for that. It means, if the user (initiator) does not poll status through checkXfer() into our code, we do not send the notif to the target. https://github.com/ai-dynamo/nixl/blob/main/src/plugins/libfabric/libfabric_backend.cpp#L1452-L1454. I was expecting when PT=1, this notif should be sent to target earlier. If you also think this is the expected behavior, I'll open a bug.

No need to address it in this PR, too.

src/utils/libfabric/libfabric_rail.cpp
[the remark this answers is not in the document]
oa-aws: This will be addressed in a separate PR, when revising error handling.

src/utils/libfabric/libfabric_rail.h
```diff
@@ -232,6 +234,83 @@ operator<<(std::ostream &os, const ConnectionState &state) {
...
+    T ring_[Capacity];
+    std::atomic<bool> committed_[Capacity] = {};
+};
+
+/**
+ * @brief Queued post request for PT-owns-endpoint model
+ * Main thread enqueues; progress thread dequeues and posts.
+ */
+struct nixlLibfabricPostRequest {
+    enum Type { WRITE, READ, SEND } type;
```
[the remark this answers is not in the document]
oa-aws: Fixing

src/utils/libfabric/libfabric_rail.h
oa-aws: I am making this configurable, with default 2048.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: The ring buffer implementation now is based on optimistic fetch-add (instead of weak CAS in loop), so there is no option for failing after N retries, but we can still report that the call encountered full ring buffer. In a future PR we can report these as statistics so we don't flood the log.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I added the missing code for setting the CUDA device.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -282,15 +294,40 @@ nixlLibfabricCudaCtx::cudaUpdateCtxPtr(void *address, int expected_dev, bool &wa
...
 nixlLibfabricCudaCtx::cudaSetCtx() {
     CUresult result;
-    if (NULL == pthrCudaCtx_) return 0;
+    if (NULL == pthrCudaCtx_) {
+        return 0;
+    }
 
     result = cuCtxSetCurrent(pthrCudaCtx_);
     return (CUDA_SUCCESS == result);
 }
```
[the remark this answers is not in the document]
oa-aws: This is in line with existing implementation. A fix here is out of scope, and should be given in a separate PR.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -328,6 +365,18 @@ void
 nixlLibfabricEngine::vramFiniCtx() {
     const std::lock_guard<std::mutex> lock(cuda_ctx_mutex_);
     cudaCtx_.reset();
+    LibfabricUtils::clearCudaCtxMediator();
+}
+
+nixl_status_t
+nixlLibfabricEngine::vramApplyCtxEx(bool &use_cuda_addr_wa) const {
```
fengjica, line 373: This function seems to be related to vramApplyCtx() above. Could we merge them, or move them closer, and call that in here?

src/plugins/libfabric/libfabric_backend.h
```diff
@@ -342,6 +342,10 @@ class nixlLibfabricEngine : public nixlBackendEngine {
     vramApplyCtx();
     void
     vramFiniCtx();
+
+    nixl_status_t
+    vramApplyCtxEx(bool &use_cuda_addr_wa) const;
```
fengjica, line 362: Mind adding a comment, because this looks very similar to vramApplyCtx() above, and it's hard to tell the semantics difference just from their names?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -755,7 +764,8 @@ nixlLibfabricRail::progressCompletionQueue() const {
     // CQ lock released here - completion is now local data
 
     if (ret == -FI_EAGAIN) {
-        return NIXL_IN_PROG; // No completions available
+        // SQ full - do inline CQ progress to free slots, then retry
```
fengjica: This comment line should be removed?

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -770,10 +780,14 @@ nixlLibfabricRail::progressCompletionQueue() const {
         }
 
         NIXL_DEBUG << "Processed " << ret << " completions on rail " << rail_id;
-        return NIXL_SUCCESS;
     }
 
-    return NIXL_ERR_BACKEND; // Unexpected case
+    // PT-owns-endpoint: drain pending posts AFTER processing completions
+    if (progress_thread_enabled_) {
+        drainPostQueue();
```
fengjica: check its return value?

src/utils/libfabric/libfabric_common.h
```diff
@@ -41,6 +41,7 @@
 #define NIXL_LIBFABRIC_CQ_SREAD_TIMEOUT_MS 10
 #define NIXL_LIBFABRIC_DEFAULT_STRIPING_THRESHOLD (128 * 1024) // 128KB
 #define LF_EP_NAME_MAX_LEN 56
+#define NIXL_LIBFABRC_DEFAULT_POST_QUEUE_SIZE 2048
```
fengjica: Is this big enough to run one of our perf test with a large batch size under mg mode? 

The more efficient push() can do, the more likely it can be full, esp. considering one PT today works across multiple NICs.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1118,6 +1130,163 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+            }
+        }
+
+        if (ret != 0) {
+            NIXL_ERROR << "drainPostQueue: post failed ret=" << ret << " (" << fi_strerror(-ret)
+                       << ") on rail " << rail_id;
+            if (pr.req && pr.req->completion_callback) {
+                // completion is notified also for failed requests
+                // (otherwise counters would never match)
+                pr.req->completion_callback();
```
fengjica, line 1274: There is no error status passing back to the caller today. Right?

https://github.com/ai-dynamo/nixl/pull/1747 has added something, but that PR is not merged.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1118,6 +1130,163 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+                               << " while posting deferred descriptor: "
+                               << cudaGetErrorString(cuda_ret);
+                    continue;
+                }
+                current_cuda_device = pr.device_id;
+            }
+        }
+#endif
+
+        while (ret == -FI_EAGAIN) {
```
fengjica, line 1217: Does it make sense to break this while loop after a number of iterations, similar to POST_MAX_BATCH_SIZE in the outer loop?

With one PT, it can be stuck here, and has no chance to drive another ring.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1118,6 +1130,163 @@ nixlLibfabricRail::postSend(uint64_t immediate_data,
...
+                msg.addr = pr.dest_addr;
+                msg.rma_iov = &rma_iov;
+                msg.rma_iov_count = 1;
+                msg.context = &pr.req->ctx;
+
+                ret = fi_readmsg(endpoint, &msg, pr.fi_flags);
+            }
+
+            if (ret == -FI_EAGAIN) {
+                struct fi_cq_data_entry cq_buf[16];
```
fengjica: Can these few lines of code be put into a function to eliminate duplicated code?

src/utils/libfabric/libfabric_common.cpp
```diff
@@ -298,4 +298,24 @@ getCustomIntParam(const nixl_b_params_t &custom_params, const std::string &key,
     return NIXL_SUCCESS;
 }
 
+static std::unique_ptr<nixlLibfaricCudaCtxMediator> cuda_ctx_mediator;
```
fengjica, line 301: Could we avoid a global variable by passing it down to where it's needed?

And, will there be a race, if a user thread and PT both tries to update it?

fengjica (commented):
I feel I have not fully digested the PR, but do not want to hold sending out a few questions that I have in mind.

src/plugins/libfabric/libfabric_backend.cpp
oa-aws: This function is different only in allowing for output parameter. How do you prefer doing that? Pointer to optional bool (with default value nullptr) does not seem right, so I added another method. I am open for suggestions.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Currently drainPostQueue() only returns success, and any error is just logged. In the future we can consider managing some error counters, to avoid log flooding, but still reporting errors. For the sake of completeness (and in case drainPostQueue will return error in the future), I am adding code to check result.

src/utils/libfabric/libfabric_common.h
oa-aws: I am in favor of big ring buffers. Host memory is usually available and writers should not be blocked. Do you have a number to suggest?

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Current completion only report success. Now that post requests are deferred, there is no way to tell caller something went wrong. Without this, when error occurs, the counters for the request will never add up, and CheckXfer() will never report the request as completed. I think the right way to do this is to add another counter for failed requests, and make sure the completion callback returns success/fail code. CheckXfer() will have to address that, so I am not sure of north/south-bound API is ready for that. I opened a ticket for that, so we can look into that issue. For now, keeping this is as (i.e. caller will see request completed even when some errors occurred). I suggest addressing this in a separate PR.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I already opened a ticket for this. We need to reconsider budgeting of this loop. Other rails are waiting as well. I have an idea how to make that much more efficient. From testing, I see that it get's stuck a bit and then it can send more. This is inefficient, but not blocking forever. I suggest dealing with this in a separate PR, I already opened a ticket for that (same one as checking SPSC per thread solution for better scalability when user threads increase).

src/utils/libfabric/libfabric_rail.cpp
oa-aws: I was actually preferring not to add refactoring to this PR, as it is big enough. What do you think? I am fine either way.

src/utils/libfabric/libfabric_common.cpp
oa-aws: setCudaCtxMediator() is called only once during backed initialization, so there is no race here. This is a minimal solution to avoid much bigger refactoring. I needed to decouple the backend/engine from the rail (otherwise rail code needs to call directly backend code, which introduces a circular dependency, a clear design anti-pattern). The correct solution is to manage CUDA context outside of the backend (e.g. some singleton class), but this PR is already big enough.

src/utils/libfabric/libfabric_common.h
oa-aws: Coming to think of it, this may require further consideration. Can we tolerate writer blocked until it can push item on the queue? Or should we fail if queue is full?

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Refactoring

fengjica (approved):

src/utils/libfabric/libfabric_rail.cpp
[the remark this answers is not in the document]
oa-aws: This is tracked in an internal ticket.

src/utils/libfabric/libfabric_rail.h
oa-aws: Default changed to 32K entries.

src/utils/libfabric/libfabric_rail.cpp
oa-aws: Change added (PT also uses CUDA context workaround code).

src/utils/libfabric/libfabric_common.h
oa-aws: Decided on 32K entries by default.

PR #1950, opened 2026-07-15
PLUGINS/UCX: Added option to forcibly close UCP endpoints. (#1846)

ovidiusm:
## What?
Added option `ucx_ep_close_force` to forcibly close UCP endpoints.

## Why?
At the moment `ucp_nb_close_nbx` is called with zeroed flags, meaning UCP endpoints (connected remotely) are destroyed only when the corresponding UCP worker is destroyed. Passing `UCP_EP_CLOSE_FLAG_FORCE` forces the destruction of a UCP endpoint and corresponding UCT endpoints, releasing UCX transport layer resources, e.g. UCT CUDA IPC cache https://github.com/openucx/ucx/pull/11335.
The issue was observed using nixl_ep examples in case of repeated connections and disconnections between ranks.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -1407,7 +1407,8 @@ void Buffer::_nixl_agent_init() {
 
     const char* num_channels_env = std::getenv("NIXL_EP_NUM_CHANNELS");
     init_params["ucx_num_device_channels"] = num_channels_env ? num_channels_env : "4";
-    init_params["ucx_error_handling_mode"] = "none";
+    init_params["ucx_error_handling_mode"] = "peer";
+    init_params["ucx_ep_close_force"] = "yes";
     init_params["num_workers"] = std::to_string(1);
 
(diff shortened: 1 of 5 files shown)


ofirfarjun7 (approved):

rakhmets (approved):

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1407,7 +1407,8 @@ void Buffer::_nixl_agent_init() {
 
     const char* num_channels_env = std::getenv("NIXL_EP_NUM_CHANNELS");
     init_params["ucx_num_device_channels"] = num_channels_env ? num_channels_env : "4";
-    init_params["ucx_error_handling_mode"] = "none";
+    init_params["ucx_error_handling_mode"] = "peer";
+    init_params["ucx_ep_close_force"] = "yes";
```
rakhmets, line 1411: This was reverted from the main branch. https://github.com/ai-dynamo/nixl/pull/1996

Please do not merge.

rakhmets (changes requested):

PR #1951, opened 2026-07-15
telemetry: add full-pipeline stress tests

e-eygin:
## What?

Test-only, full-pipeline stress coverage for telemetry: concurrent producers through the staging queue → periodic flush → the BUFFER / Prometheus / DOCA exporters. No product, queue, exporter, schema, or public-API changes.

- `test/gtest/telemetry_stress_test.cpp` — value parity, `addXferStats` batch atomicity, lifecycle shake-out, safe shutdown (BUFFER, end-to-end).
- `test/gtest/telemetry_prometheus_stress_test.cpp` — concurrent staging-overflow conservation and mixed-workload aggregation.
- `test/doca-telemetry/telemetry_doca_stress_test.cpp` — high-volume DOCA exporter accumulation (standalone `doca_nixl_stress_test`).

## Why?

Closes [NIX-1205](https://linear.app/nvidia/issue/NIX-1205). Existing tests cover queue mechanics and single-threaded exporters, but not the concurrent observable pipeline. This pins down that contract — conservation, all-or-none batching, per-value fidelity, shutdown safety — so a regression anywhere in the path fails an assertion. It builds only on already-merged behavior (NIX-454 #1887, NIX-453 #1919), so it is independent of NIX-1542 and mergeable on its own, while serving as the validation gate for the NIX-1542/NIX-1544 queue changes.

## How?

Deterministic and timing-independent: conservation identities and value multisets rather than fixed accepted/dropped splits, bounded polling instead of sleeps, and sanitizer-scaled iteration counts.

<details>
<summary>Design & validation details</summary>

- Under-capacity BUFFER tests size the ring above the total so nothing is dropped; the overflow case drives the lossless, ring-free Prometheus exporter, so `accepted + dropped == produced` is exact and hardware-independent.
- `addXferStats` atomicity is proven by per-call identity + contiguity in the drained ring (a torn or interleaved batch fails), not just aggregate counts.
- Safe-shutdown destroys on a worker thread bounded by an external timeout, so a teardown deadlock fails as an assertion instead of hanging; producers are released together via a start latch to maximize real concurrency.
- Validation: normal C++17, TSan, UBSan, and DOCA-enabled builds pass; focused telemetry suites run warning-clean. ASan can't run in some environments due to a prebuilt mixed-Abseil ABI incompatibility (documented in the plan).

</details>

## Tests

- Added high-volume telemetry stress coverage, including concurrent Prometheus metrics scraping and validation of counters, last-operation gauges, and dropped-event accounting.
- Added concurrency tests for overflow handling, metric aggregation, and error/status labeling.
- Added end-to-end checks for batching integrity, repeated telemetry lifecycles, and safe shutdown during active flushing.
- Added a dedicated DOCA telemetry stress test with an extended timeout.

### DIFF
--- a/test/doca-telemetry/meson.build
+++ b/test/doca-telemetry/meson.build
@@ -63,6 +63,26 @@ if build_doca_tests
     )
 
     test('doca_nixl_test', doca_nixl_test_exe, is_parallel: false)
+
+    # Exporter volume/stress case, kept in its own source/binary so the sustained
+    # high-volume run stays isolated from the functional exporter tests.
+    doca_nixl_stress_test_exe = executable('doca_nixl_stress_test',
+        sources: [
+            'telemetry_doca_stress_test.cpp',
+            doca_exporter_dir / 'doca_exporter.cpp',
+        ],
+        include_directories: [
+            nixl_inc_dirs,
+            utils_inc_dirs,
+            include_directories(doca_exporter_dir),
+            include_directories('../../src/plugins/telemetry/common'),
+        ],
+        dependencies: [gtest_dep, doca_telemetry_dep, doca_common_dep, nixl_infra, nixl_common_dep, absl_log_dep, absl_strings_dep],
+        cpp_args: ['-Wno-deprecated-declarations'],
+        install: true,
+    )
+
+    test('doca_nixl_stress_test', doca_nixl_stress_test_exe, is_parallel: false)
     message('Building standalone DOCA telemetry exporter tests')
 
     cmake = import('cmake')
--- a/test/doca-telemetry/meson.build
+++ b/test/doca-telemetry/meson.build
@@ -82,7 +82,9 @@ if build_doca_tests
         install: true,
     )
 
-    test('doca_nixl_stress_test', doca_nixl_stress_test_exe, is_parallel: false)
+    # scrapeUntilValue polls up to 15s; give generous headroom over Meson's 30s
+    # default for the high-volume run plus setup/flush on loaded or sanitizer CI.
+    test('doca_nixl_stress_test', doca_nixl_stress_test_exe, is_parallel: false, timeout: 120)
     message('Building standalone DOCA telemetry exporter tests')
 
     cmake = import('cmake')
--- a/test/doca-telemetry/meson.build
+++ b/test/doca-telemetry/meson.build
@@ -49,25 +49,6 @@ if build_doca_tests
     doca_nixl_test_exe = executable('doca_nixl_test',
         sources: [
             'telemetry_doca_nixl_test.cpp',
-            doca_exporter_dir / 'doca_exporter.cpp',
-        ],
-        include_directories: [
-            nixl_inc_dirs,
-            utils_inc_dirs,
-            include_directories(doca_exporter_dir),
-            include_directories('../../src/plugins/telemetry/common'),
-        ],
-        dependencies: [gtest_dep, doca_telemetry_dep, doca_common_dep, nixl_infra, nixl_common_dep, absl_log_dep, absl_strings_dep],
-        cpp_args: ['-Wno-deprecated-declarations'],
-        install: true,
-    )
-
-    test('doca_nixl_test', doca_nixl_test_exe, is_parallel: false)
-
-    # Exporter volume/stress case, kept in its own source/binary so the sustained
-    # high-volume run stays isolated from the functional exporter tests.
-    doca_nixl_stress_test_exe = executable('doca_nixl_stress_test',
-        sources: [
             'telemetry_doca_stress_test.cpp',
             doca_exporter_dir / 'doca_exporter.cpp',
         ],
@@ -82,9 +63,11 @@ if build_doca_tests
         install: true,
     )
 
-    # scrapeUntilValue polls up to 15s; give generous headroom over Meson's 30s
-    # default for the high-volume run plus setup/flush on loaded or sanitizer CI.
-    test('doca_nixl_stress_test', doca_nixl_stress_test_exe, is_parallel: false, timeout: 120)
+    # The suite normally runs in well under a second, but each case may spend its
+    # scrape settle budget (12-15s) waiting on a lagging DOCA endpoint. Bound the
+    # test by the sum of those budgets so a slow CI machine reports the failing
+    # assertion instead of a bare timeout.
+    test('doca_nixl_test', doca_nixl_test_exe, is_parallel: false, timeout: 180)
     message('Building standalone DOCA telemetry exporter tests')
 
     cmake = import('cmake')
--- a/test/doca-telemetry/telemetry_doca_stress_test.cpp
+++ b/test/doca-telemetry/telemetry_doca_stress_test.cpp
@@ -1,0 +1,131 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// DOCA exporter volume within the standalone test architecture.
+//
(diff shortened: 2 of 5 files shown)


test/gtest/telemetry_stress_test.cpp
```diff
@@ -0,0 +1,570 @@
...
+eventIndex(nixl_telemetry_event_type_t type) noexcept {
+    return static_cast<std::size_t>(type);
+}
+
+using EventCounts = std::array<uint64_t, nixl_telemetry_event_type_count>;
+// Per event type, a histogram of received event values -> occurrence count. Used
+// to prove value parity: the multiset of values drained from the ring must equal
+// the multiset produced, so no value is corrupted, duplicated, or lost.
+using ValueHistograms =
+    std::array<std::unordered_map<uint64_t, uint64_t>, nixl_telemetry_event_type_count>;
```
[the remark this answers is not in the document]
e-eygin: Renamed `EventCounts`→`event_counts_t`, `ValueHistograms`→`value_histograms_t`, `et`→`event_type_t`

test/gtest/telemetry_stress_test.cpp
```diff
@@ -0,0 +1,570 @@
...
+        gate.release();
+        for (auto &producer : producers) {
+            producer.join();
+        }
+
+        const auto start = std::chrono::steady_clock::now();
+        telemetry.reset(); // destructor races an active periodic flush
+        const auto elapsed = std::chrono::steady_clock::now() - start;
+        EXPECT_LT(elapsed, shutdown_budget)
+            << "telemetry destruction must complete promptly, not deadlock on the flush timer";
```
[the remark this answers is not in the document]
e-eygin: External timeout: done in `1344c937` — the destructor now runs on a worker thread bounded by `future::wait_for(shutdown_budget)`, so a `reset()` deadlock fails via `ADD_FAILURE` instead of being unreachable.

Deterministic flush-entry overlap: intentionally not added. This is a load test — 12 iterations × 4 producers × 2000 calls at a 1 ms interval means `reset()` repeatedly lands mid-flush, and TSan/ASan runs are what actually detect a teardown race. The NIX-1205 plan specifies this case may "arrange for or allow the periodic consumer to be active" with only a bounded-completion assertion. A blocking exporter/hook would test an artificially parked flush rather than the natural under-load teardown, so it'd make the stress test less realistic for no added coverage. Keeping it test-only (no production seam) as intended.

e-eygin:

test/gtest/telemetry_prometheus_stress_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Concurrent Prometheus stress coverage. Kept separate from the functional
+// Prometheus suite (telemetry_prometheus_test.cpp) but sharing its fixture via
+// prometheus_telemetry_fixture.h and scraping through the same OpenMetrics
+// parser/time-series helpers the DOCA and histogram tests use. The Prometheus
```
ColinNV, line 21: > so concurrent staging overflow and mixed aggregation resolve to exact, hardware-independent conservation identities.

What does this mean?

test/gtest/telemetry_stress_test.cpp
```diff
@@ -0,0 +1,662 @@
...
+    while (size <= n) {
+        size <<= 1;
+    }
+    return size;
+}
+
+// A start latch so every producer thread is released together, maximizing the
+// concurrency the staging queue actually sees (per the plan: coordinate with
+// atomics/barriers, do not rely on sleeps).
+class StartGate {
```
ColinNV: Can this be a `std::latch` or `std::barrier` instead?

test/doca-telemetry/meson.build
```diff
@@ -63,6 +63,28 @@ if build_doca_tests
     )
 
     test('doca_nixl_test', doca_nixl_test_exe, is_parallel: false)
+
+    # Exporter volume/stress case, kept in its own source/binary so the sustained
+    # high-volume run stays isolated from the functional exporter tests.
```
ColinNV: Why is that necessary? We are moving towards running all gtests within a single invocation, but still sequentially.

test/gtest/telemetry_prometheus_stress_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+using nixl::doca_test::labelSet;
+using nixl::doca_test::loopbackConnection;
+using nixl::doca_test::timeSeries;
+
+namespace {
+
+// Sanitizer builds run the same concurrent aggregation with fewer producer
+// iterations so TSan/ASan stay within a CI-friendly runtime. Compile-time
+// detection in test code only.
```
ColinNV: > Compile-time detection in test code only.

This is in `test/gtest`...?

test/gtest/telemetry_prometheus_stress_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+constexpr bool kSanitizerBuild = false;
+#endif
+#else
+constexpr bool kSanitizerBuild = false;
+#endif
+
+// A start latch so every producer thread is released together, after all threads
+// exist -- otherwise early threads can finish before later ones are created and
+// the workload runs largely serially, hiding concurrency defects.
+class StartGate {
```
ColinNV: Can this use `std::latch` or `std::barrier`?

test/doca-telemetry/telemetry_doca_stress_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+        const nixlTelemetryEvent event(nixl_telemetry_event_type_t::AGENT_TELEMETRY_EVENTS_DROPPED,
+                                       kDropPerDelta);
+        ASSERT_EQ(exporter.exportEvent(event), NIXL_SUCCESS);
+        drop_total += kDropPerDelta;
+    }
+
+    ASSERT_EQ(exporter.flush(), NIXL_SUCCESS);
+
+    const auto metrics = scrapeUntilValue(
+        port_, txCounter, static_cast<double>(counter_sum), std::chrono::seconds(15), labels);
```
ColinNV: Does the "scrape until" need to wait for the dropped counter with its later events to ensure that everything has arrived?

test/gtest/telemetry_prometheus_stress_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+                               }
+                               gate.release();
+                               for (auto &producer : producers) {
+                                   producer.join();
+                               }
+                           });
+
+    ASSERT_TRUE(scrape.ok) << "accepted*4 + dropped must reach produced events (" << kProducedEvents
+                           << ") under concurrent overflow -- no silent loss";
+    EXPECT_GT(scrape.dropped, 0.0)
```
ColinNV, line 178: Is this timing dependent or is there enough margin to "guarantee" no false positives?

test/gtest/telemetry_prometheus_stress_test.cpp
e-eygin: Rewrote it: the Prometheus exporter can't drop on export and has no ring behind
it, so the staging queue is the only loss point — hence exact totals instead of
tolerances.

test/gtest/telemetry_stress_test.cpp
e-eygin: Done — `std::latch start_gate(1)`, StartGate deleted.

test/gtest/telemetry_prometheus_stress_test.cpp
e-eygin: Done.

test/doca-telemetry/meson.build
e-eygin: Agreed — folded into doca_nixl_test. Separate source, one binary, one invocation;
extra target, main() and test() entry gone.

test/gtest/telemetry_prometheus_stress_test.cpp
e-eygin: Not timing-dependent now: with a 256-slot queue flushed every 50ms, pushing 640000
events without a single drop would require the producers to run 2+ minutes. They
finish in ~270ms, so drops are guaranteed.

test/gtest/telemetry_prometheus_stress_test.cpp
e-eygin: Right, removed (and its twin in telemetry_stress_test.cpp).

test/doca-telemetry/telemetry_doca_stress_test.cpp
e-eygin: Yes — now gates on the drop counter, the last events pushed before the flush.

ColinNV (approved):

PR #1952, opened 2026-07-16
telemetry/doca: share one timestamp per export batch (NIX-1625)

e-eygin:
## What?
The DOCA telemetry exporter now stamps every sample in one drain with a single
timestamp, instead of reading the DOCA clock 2–3× per event.

## Why?
Each append helper read the clock independently, so one flush produced ~2N–3N
slightly-different timestamps for samples that belong to the same instant. One
read per drain is cheaper and semantically consistent. (NIX-1625)

## How?
`nixlTelemetryExporter::withBatch()` wraps the drain via a private RAII guard and
no-op `onBatchBegin`/`onBatchEnd` hooks; the DOCA exporter caches one timestamp
per batch. No metric names, labels, or buffer-format changes — scrape output is
unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added batch-scoped telemetry exporting via `withBatch`, including begin/end lifecycle hooks for exporters (no-op by default).
* **Bug Fixes**
  * Ensured telemetry metrics written during a single batch export share one consistent timestamp.
  * Ensured events exported outside a batch receive fresh, per-event timestamps.
  * Exported queued events and dropped-event metrics within the same batch context.
  * Documented that batch scopes are not intended to be nested.
* **Tests**
  * Added end-to-end tests for shared timestamps within a batch and distinct timestamps for unbatched exports.


src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -356,32 +356,53 @@ nixlTelemetryDocaExporter::~nixlTelemetryDocaExporter() {
...
+void
+nixlTelemetryDocaExporter::onBatchBegin() noexcept {
+    inBatch_ = true;
+    batchTimestamp_.reset();
+}
+
+void
+nixlTelemetryDocaExporter::onBatchEnd() noexcept {
+    inBatch_ = false;
+    batchTimestamp_.reset();
```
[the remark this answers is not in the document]
e-eygin: Fixed in `b18b7ae8` — in the base `nixlTelemetryExporter` rather than the DOCA hooks: `withBatch` now tracks a depth counter and fires `onBatchBegin`/`onBatchEnd` only on the outermost `0→1`/`1→0` transitions, so a nested scope's exit can't reset the outer batch's timestamp (and every exporter gets it for free). Added regression test `NestedBatchScopesShareOuterTimestamp`.

e-eygin:

src/api/cpp/telemetry/telemetry_exporter.h
```diff
@@ -75,7 +76,58 @@ class nixlTelemetryExporter {
...
+        nixlTelemetryExporter &exporter_;
+    };
+
+    virtual void
+    onBatchBegin() noexcept {}
+
+    virtual void
+    onBatchEnd() noexcept {}
+
+    unsigned batchDepth_ = 0;
```
ColinNV: Is there any conceivable case for nesting batch scopes? If not this could be simplified a lot.

e-eygin: Dropped nesting support in `d7297e65` — `withBatch` is only called from `flushPendingEvents()`, never nested.

src/api/cpp/telemetry/telemetry_exporter.h
```diff
@@ -75,7 +76,52 @@ class nixlTelemetryExporter {
...
+     * exception). Batches are not nested.
+     *
+     * @tparam Body Callable invocable with no arguments.
+     * @param body Callable run once while the batch is open.
+     */
+    template<typename Body>
+    void
+    withBatch(Body &&body) {
+        const batchScope scope{*this};
+        std::forward<Body>(body)();
```
ColinNV: The `forward` should not be needed.

src/plugins/telemetry/doca/doca_exporter.cpp
```diff
@@ -356,32 +356,53 @@ nixlTelemetryDocaExporter::~nixlTelemetryDocaExporter() {
...
+    }
+    if (!batchTimestamp_.has_value()) {
+        batchTimestamp_ = docaTimestamp();
+    }
+    return *batchTimestamp_;
+}
+
+void
+nixlTelemetryDocaExporter::onBatchBegin() noexcept {
+    inBatch_ = true;
```
ColinNV: If nesting batches is not needed, can this not set `batchTimestamp_` to `docaTimestamp()` and the `bool` be removed, using the `batchTimestamp_.has_value()` instead?

src/api/cpp/telemetry/telemetry_exporter.h
e-eygin: Removed.

src/plugins/telemetry/doca/doca_exporter.cpp
e-eygin: Done — stamped at batch open, so the optional signals the batch.

ColinNV (approved):

PR #1953, opened 2026-07-16
PLUGINS/UCX: Log UCX version with 'less than 1.19' version.

ColinNV:
## What?
Add the used UCX version to the existing `less than 1.19` warning.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UCX version compatibility warnings by highlighting the detected runtime version (major/minor/release number) when it’s below 1.19.
  * Refined the warning wording to better clarify current CUDA limitations and multi-GPU behavior within a single process.
  * Updated the associated test expectation to align with the new warning text.


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -455,8 +455,11 @@ nixlUcxContext::nixlUcxContext(const std::vector<std::string> &devs,
     }
 
     if (ucpVersion_ < UCP_VERSION(1, 19)) {
-        NIXL_WARN << "UCX version is less than 1.19, CUDA support is limited, "
-                  << "including the lack of support for multi-GPU within a single process.";
+        unsigned major, minor, patch;
```
guy-ealey-morag: Maybe have these as class members and populate them along with `ucpVersion_` in the constructor?
We already call `ucp_get_version` so it makes sense

ColinNV: This change will be temporary in the sense that a version below 1.19 might not be supported for *too* much longer, i.e. this is "throwaway" code.

guy-ealey-morag: This warning might stay for longer because in the future it might warn about the next unsupported UCX version (e.g. 1.20 and so on)

ColinNV: The case we have now is unsupported features on supported versions, and we probably don't need to store the three individual version fields, but we can call `ucp_get_version` only once.

test/gtest/main.cpp
```diff
@@ -70,8 +70,9 @@ void ParseArguments(int argc, char **argv) {
 namespace {
     const std::regex
         ib_regex("IB device\\(s\\) were detected, but accelerated IB support was not found");
-    const std::regex aws_regex("UCX version is less than 1.19, CUDA support is limited, including"
-                               " the lack of support for multi-GPU within a single process.");
+    const std::regex aws_regex("UCX version .+ is less than 1\\.19, CUDA support is limited, "
+                               "including the lack of support for multi-GPU within a single "
+                               "process.");
```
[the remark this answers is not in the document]
guy-ealey-morag: @ColinNV I agree with the proposed fix here, `.+` seems too broad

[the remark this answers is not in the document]
ColinNV: Done, sans the rabbit's mistake.

guy-ealey-morag (approved):

iyastreb (approved):

PR #1960, opened 2026-07-16
fix(python): release deferred transfer handles through agent

lluki:
This method doesnt exist, but because we swallow all exceptions we dont notice

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of transfer resources when an agent is destroyed, reducing the risk of leaked handles.


ebarilanM (approved):
LGTM

ColinNV (approved):

iyastreb (approved):

PR #1966, opened 2026-07-20
libfabric: flush FI_MORE per rail, not per group

akkart-aws:
## What?
The positional flush then rings only one rail's doorbell; the other rail's batch is never submitted and the transfer hangs. nixlbench doesn't hit this: one buffer per device = one rail list.

## Why?
FI_MORE WRITE batching (#1626) flushes the doorbell positionally (last of each 16-group / last of the transfer), which is only correct when every descriptor in a group is on the same rail.  But each descriptor's rail comes from its own buffer's selected_rails, so a transfer spanning buffers with different rail lists (e.g. dynamo's many DRAM registrations) alternates physical rails within a group. 

## How?
postXfer precomputes the flush decision per physical rail (walking descriptors backwards, the first rail seen is its last post), and prepareAndSubmitTransfer takes a bool apply_fi_more instead of deciding positionally. Batching is opt-in, so unmarked descriptors (READ, striped, rail-less) flush.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved Libfabric WRITE batching by precomputing per-descriptor `FI_MORE` decisions, with tighter rail-aligned round-robin selection for better endpoint targeting.
* **Behavior Changes**
  * WRITE posts now respect controlled “flush” boundaries more consistently; behavior differs between single-path posting and chunked/thread-pool posting to avoid incorrect “last post” assumptions.
  * READ behavior remains unchanged.
* **Configuration**
  * Added `NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE` (default: 16) to limit consecutive WRITE descriptors eligible for `FI_MORE` batching before flushing.

### DIFF
--- a/src/plugins/libfabric/libfabric_backend.cpp
+++ b/src/plugins/libfabric/libfabric_backend.cpp
@@ -1106,6 +1106,40 @@ nixlLibfabricEngine::estimateXferCost(const nixl_xfer_op_t &operation,
     return NIXL_SUCCESS;
 }
 
+int
+nixlLibfabricEngine::batchingRail(const nixl_meta_dlist_t &local,
+                                  int desc_idx,
+                                  size_t xfer_base_offset) const {
+    auto *md = static_cast<nixlLibfabricPrivateMetadata *>(local[desc_idx].metadataP);
+    if (!md || md->selected_rails_.empty()) {
+        return -1;
+    }
+    // Striped descriptors are split across their rails and posted without FI_MORE, so they
+    // are not part of the single-rail batching tracked here.
+    if (rail_manager_.usesStriping(local[desc_idx].len, md->selected_rails_.size())) {
+        return -1;
+    }
+    return (int)md->selected_rails_[nixlLibfabricRailManager::railSelectionIndex(
+        xfer_base_offset, desc_idx, /*batch_write=*/true, md->selected_rails_.size())];
+}
+
+bool
+nixlLibfabricEngine::useFiMore(int desc_idx,
+                               int rail_id,
+                               const std::vector<int> &last_desc_idx_per_rail,
+                               std::vector<int> &posts_since_flush) const {
+    if (rail_id < 0) {
+        return false;
+    }
+    if (desc_idx == last_desc_idx_per_rail[rail_id] ||
+        posts_since_flush[rail_id] == NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE - 1) {
+        posts_since_flush[rail_id] = 0;
+        return false;
+    }
+    ++posts_since_flush[rail_id];
+    return true;
+}
+
 nixl_status_t
 nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
                                          const nixl_meta_dlist_t &local,
@@ -1114,8 +1148,8 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
                                          nixlLibfabricBackendH *backend_handle,
                                          int start_idx,
                                          int end_idx,
-                                         int desc_count,
                                          size_t xfer_base_offset,
+                                         bool allow_fi_more,
                                          size_t &submitted_count) const {
     submitted_count = 0;
 
@@ -1147,7 +1181,30 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
     };
 #endif
 
+    // A FI_MORE post rings no doorbell; a rail's queued batch is only submitted by a later
+    // non-FI_MORE post on the same rail. Rails are resolved per-buffer, so descriptors of one
+    // round-robin group can land on different rails: every rail this chunk touches must have its
+    // last post flushed, or its batch would never be submitted.
+    // allow_fi_more is false on the thread-pool path: chunks on other threads can interleave
+    // posts on the same rail, so a per-chunk walk cannot know a rail's true last post there.
+    const bool batch_writes = allow_fi_more && op_type == nixlLibfabricReq::WRITE;
+
+    std::vector<int> last_desc_idx_per_rail(rail_manager_.getNumRails(), -1);
+    std::vector<int> posts_since_flush(rail_manager_.getNumRails(), 0);
+    if (batch_writes) {
+        for (int i = start_idx; i < end_idx; ++i) {
+            const int rail_id = batchingRail(local, i, xfer_base_offset);
+            if (rail_id >= 0) {
+                last_desc_idx_per_rail[rail_id] = i;
+            }
+        }
+    }
+
     for (int desc_idx = start_idx; desc_idx < end_idx; ++desc_idx) {
+        const int rail_id = batch_writes ? batchingRail(local, desc_idx, xfer_base_offset) : -1;
+        const bool apply_fi_more =
+            useFiMore(desc_idx, rail_id, last_desc_idx_per_rail, posts_since_flush);
+
         auto *local_md = static_cast<nixlLibfabricPrivateMetadata *>(local[desc_idx].metadataP);
         auto *remote_md = static_cast<nixlLibfabricPublicMetadata *>(remote[desc_idx].metadataP);
 
@@ -1182,8 +1239,8 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             [backend_handle]() { backend_handle->increment_completed_requests(); },
             desc_submitted_count,
             desc_idx,
-            desc_count,
-            xfer_base_offset);
+            xfer_base_offset,
+            apply_fi_more);
 
         if (status != NIXL_SUCCESS) {
             NIXL_ERROR << "prepareAndSubmitTransfer failed for descriptor " << desc_idx
@@ -1294,8 +1351,8 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
                                                    backend_handle,
                                                    0,
                                                    desc_count,
-                                                   desc_count,
                                                    xfer_base_offset,
+                                                   /*allow_fi_more=*/true,
                                                    total_submitted);
         if (status != NIXL_SUCCESS) {
             return status;
@@ -1327,8 +1384,8 @@ nixlLibfabricEngine::postXfer(const nixl_xfer_op_t &operation,
                                                      backend_handle,
                                                      start_idx,
                                                      end_idx,
-                                                     desc_count,
                                                      xfer_base_offset,
+                                                     /*allow_fi_more=*/false,
                                                      chunk_submitted);
                     }
                     catch (const std::exception &e) {
--- a/src/utils/libfabric/libfabric_rail_manager.cpp
+++ b/src/utils/libfabric/libfabric_rail_manager.cpp
@@ -367,11 +367,6 @@ nixlLibfabricRailManager::createRails(const std::vector<std::string> &efa_device
     return NIXL_SUCCESS;
 }
 
-bool
-nixlLibfabricRailManager::shouldUseStriping(size_t transfer_size) const {
-    return transfer_size >= striping_threshold_;
-}
-
 size_t
 nixlLibfabricRailManager::reserveBaseOffset() {
     return round_robin_counter.fetch_add(1);
@@ -394,8 +389,8 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
     std::function<void()> completion_callback,
     size_t &submitted_count_out,
     int desc_idx,
-    int desc_count,
-    size_t base_offset) {
+    size_t base_offset,
+    bool apply_fi_more) {
     // Initialize output parameter
     submitted_count_out = 0;
 
@@ -405,22 +400,22 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
     }
 
     // Determine striping strategy
-    bool use_striping = shouldUseStriping(transfer_size) && selected_rails.size() > 1;
+    bool use_striping = usesStriping(transfer_size, selected_rails.size());
     NIXL_DEBUG << "use_striping=" << use_striping;
     if (!use_striping) {
-        // WRITE: group 16 consecutive descs to same rail for FI_MORE batching.
-        // READ: per-descriptor round-robin (FI_MORE has no benefit for reads).
-        constexpr int FI_MORE_BATCH_SIZE = 16;
+        // WRITE: group NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE consecutive descriptors on one rail
+        // for FI_MORE batching. READ: per-descriptor round-robin (FI_MORE has no benefit).
+        // apply_fi_more is precomputed by the caller, which leaves it false for the last post
+        // on each rail so that every rail's FI_MORE batch is flushed.
         const bool batch_write = (op_type == nixlLibfabricReq::WRITE);
-        const size_t rr_idx =
-            batch_write ? (base_offset + desc_idx / FI_MORE_BATCH_SIZE) : (base_offset + desc_idx);
-        const size_t rail_id = selected_rails[rr_idx % selected_rails.size()];
-        const size_t remote_ep_id =
-            remote_selected_endpoints[rr_idx % remote_selected_endpoints.size()];
-        const int pos_in_group = desc_idx % FI_MORE_BATCH_SIZE;
-        const bool is_last_in_group =
-            (pos_in_group == FI_MORE_BATCH_SIZE - 1) || (desc_idx == desc_count - 1);
-        const uint64_t fi_flags = (batch_write && !is_last_in_group) ? FI_MORE : 0;
+        const size_t rail_sel_idx =
+            railSelectionIndex(base_offset, desc_idx, batch_write, selected_rails.size());
+        const size_t rail_id = selected_rails[rail_sel_idx];
+        // The remote endpoint round-robins over its own count, which can differ from the
+        // local rail count.
+        const size_t remote_ep_id = remote_selected_endpoints[railSelectionIndex(
+            base_offset, desc_idx, batch_write, remote_selected_endpoints.size())];
+        const uint64_t fi_flags = (batch_write && apply_fi_more) ? FI_MORE : 0;
         NIXL_DEBUG << "rail " << rail_id << ", remote_ep_id " << remote_ep_id;
         // Allocate request
         nixlLibfabricReq *req = rails_[rail_id]->allocateDataRequest(op_type, xfer_id);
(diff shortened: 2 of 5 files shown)


src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1182,8 +1219,8 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
             [backend_handle]() { backend_handle->increment_completed_requests(); },
             desc_submitted_count,
             desc_idx,
-            desc_count,
-            xfer_base_offset);
+            xfer_base_offset,
+            (bool)apply_fi_more[(size_t)desc_idx]);
```
amitrad-aws: Why not just check to make sure the "next" localMd is the same as the current one to decide this flag? 
Why do we need to go over all the descs beforehand?

akkart-aws: I looked at doing it that way and it doesn't quite work, here's why.                                                                                   
Checking "is local[desc_idx+1] the same registration/rail as the current one" is correct (it never leaves a rail unflushed, so no hang), but it silently defeats the FI_MORE batching in exactly the case this PR is about.                                                                                                                   

The batch groups by rr_idx = base_offset + desc_idx / FI_MORE_BATCH_SIZE, but the physical rail is md->selected_rails_[rr_idx % nrails] from each descriptor's own buffer. When a transfer spans buffers with different rail lists, consecutive descriptors alternate rails — e.g.:
```
desc: 0   1   2   3   4   5                                                                                                                                              
rail: R9  R11 R9  R11 R9  R11                                                                                                                                            
```                                                                                                                                                                      
With a "next descriptor == same rail?" test, every descriptor's neighbor is a different rail, so every post flushes → we'd send zero FI_MORE and lose the doorbell-batching PR #1626 added, precisely in the multi-rail workload where it matters.
                                                                                                                                                                           
The property we actually need is "is this the last post on this rail within the group," which is a look-ahead over the remaining descriptors, not just the immediate neighbor (the rail can recur non-adjacently). 

Doing that per-descriptor would be an O(n²) forward scan; the single backward pass computes the same thing in O(n) — first time a rail is seen scanning backwards is its last post → flush, everything earlier on that rail → FI_MORE. So the precompute is really just the O(n) way to answer your question without the quadratic scan.                                                                                                                                     
                                                                                                                                                                           
Net: peeking the next descriptor is safe but drops the optimization; the backward pass keeps both correctness and batching. Happy to switch if you'd prefer simplicity over the batching in the alternating case, but I think we want to keep the batching.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1147,6 +1147,43 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
     };
 #endif
 
+    // Precompute, per descriptor in this chunk, whether it should carry FI_MORE (be batched).
```
fengjica: I feel this comment block can be simplified for readability.

It repeats what's covered in the commit description. And what the code is going to implement does not need to be described in the comment, because when one updates code in the future, they will need to make sure two things are consistent.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1147,6 +1147,43 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+        // not yet seen in this backward walk. Indexed by rail id (< getNumRails()).
+        std::vector<int> posts_since_flush(rail_manager_.getNumRails(), -1);
+        for (int i = end_idx - 1; i >= start_idx; --i) {
+            auto *md = static_cast<nixlLibfabricPrivateMetadata *>(local[i].metadataP);
+            if (!md || md->selected_rails_.empty()) {
+                continue;
+            }
+            if (rail_manager_.shouldUseStriping(local[i].len) && md->selected_rails_.size() > 1) {
+                continue;
+            }
```
fengjica: Could you add a comment why it should continue here?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1147,6 +1147,43 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+    // buffers with different rail lists can alternate rails within one round-robin group. A rail
+    // whose batch is never followed by a non-FI_MORE post is never submitted, so every rail this
+    // chunk touches must have its last post flushed. Walk this chunk's descriptors backwards,
+    // resolving each one's rail exactly as prepareAndSubmitTransfer does: the first time a rail is
+    // seen (going backwards) is its last post in this chunk -> flush (leave false); earlier posts
+    // on that rail -> FI_MORE (true), also flushing once per batch to bound size. The scope is
+    // [start_idx, end_idx): chunks post independently (possibly on different threads), so each
+    // chunk flushes the rails it owns. Default false (flush) -> READ / striped / rail-less
+    // descriptors are never batched.
+    std::vector<uint8_t> apply_fi_more((size_t)desc_count, 0); // indexed by desc_idx; default flush
```
fengjica: Does the following work:
std::vector<int> last_descriptor_index_per_rail(rail_manager_.getNumRails(), -1);
for(;;) {. 
  // find the last descriptor index per rail
}

for (;;) {
 // if (idx != last_descriptor_index_per_rail[rail]) && (pos_in_group != FI_MORE_BATCH_SIZE - 1)), apply FI_MORE.
}

This way, the first for loop is simpler. And we don't need apply_fi_more, which can grow with desc_count.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -408,19 +408,28 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+        // for FI_MORE batching. READ: per-descriptor round-robin (FI_MORE has no benefit).
+        //
+        // FI_MORE flush invariant: the provider rings no doorbell for a FI_MORE post; a
+        // rail's queued batch is only submitted by a later non-FI_MORE post on that same
+        // rail. selected_rails is per-buffer, so a topology-aware policy can resolve the
+        // same rr_idx to different rails across adjacent descriptors. The caller therefore
+        // precomputes, per physical rail, which post is the rail's last and passes
+        // apply_fi_more = true only for the earlier (batchable) posts. Batching is opt-in:
+        // the unsafe failure is a never-flushed rail (permanent hang), the safe one is an
+        // extra doorbell, so any descriptor the caller didn't mark flushes.
```
fengjica: The same as above. I don't think we need all these details in the comment. Some of them are hard to follow too.

I'd also avoid using 'opt-in', as that reads to me as if there were a parameter for users of the code to control whether to use it, which is not the case.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -408,19 +408,28 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
...
+        const size_t rail_id = selected_rails[rail_sel_idx];
+        // Select the remote endpoint at the same position as the rail: rail and endpoint
+        // are parallel per-buffer arrays, so independent modulos could mismatch them.
         const size_t remote_ep_id =
-            remote_selected_endpoints[rr_idx % remote_selected_endpoints.size()];
-        const int pos_in_group = desc_idx % FI_MORE_BATCH_SIZE;
-        const bool is_last_in_group =
-            (pos_in_group == FI_MORE_BATCH_SIZE - 1) || (desc_idx == desc_count - 1);
-        const uint64_t fi_flags = (batch_write && !is_last_in_group) ? FI_MORE : 0;
+            remote_selected_endpoints[rail_sel_idx % remote_selected_endpoints.size()];
```
fengjica: I don't think this is correct. Imagine selected_rails.size == 1, and remote_selected_endpoints.size() == 2. 
rr_idx can round robin on remote side, rail_sel_idx cannot.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1147,6 +1147,40 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+        // not yet seen in this backward walk. Indexed by rail id (< getNumRails()).
+        std::vector<int> posts_since_flush(rail_manager_.getNumRails(), -1);
+        for (int i = end_idx - 1; i >= start_idx; --i) {
+            auto *md = static_cast<nixlLibfabricPrivateMetadata *>(local[i].metadataP);
+            if (!md || md->selected_rails_.empty()) continue;
+            if (rail_manager_.shouldUseStriping(local[i].len) && md->selected_rails_.size() > 1)
+                continue;
+            const size_t rr_idx =
+                xfer_base_offset + (size_t)i / NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE;
+            const size_t rail_id = md->selected_rails_[rr_idx % md->selected_rails_.size()];
```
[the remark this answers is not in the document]
akkart-aws: Keeping the macro for local consistency: every configuration constant in this header is an UPPER_SNAKE #define (NIXL_LIBFABRIC_CQ_SREAD_TIMEOUT_MS,  NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL, etc.), and the file's one existing constexpr (NIXL_LIBFABRIC_GIGA) also uses UPPER_SNAKE. Switching this single constant to snake_case inline constexpr would break the file's convention. A wholesale macro→constexpr migration of this header would be reasonable, but belongs in a separate cleanup PR, not this bug fix.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Trimmed the comments.

src/utils/libfabric/libfabric_rail_manager.cpp
akkart-aws: Trimmed to two sentences and removed the "opt-in" phrasing

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1147,6 +1147,33 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+            const size_t rr_idx = xfer_base_offset + (size_t)i / NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE;
+            const size_t rail_id = md->selected_rails_[rr_idx % md->selected_rails_.size()];
+            int &seen = posts_since_flush[rail_id];
+            if (seen < 0 || seen >= NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE - 1) {
+                seen = 0;
+            } else {
+                apply_fi_more[(size_t)i] = 1;
+                ++seen;
+            }
+        }
```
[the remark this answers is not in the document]
akkart-aws: Done — postXferDescriptors now takes allow_fi_more; the thread-pool path passes false, so pooled chunks post every write with an immediate doorbell (no FI_MORE), while the default single-threaded path keeps batching. Serializing per-rail state across workers would put a shared lock in the parallel post path, defeating the pool's purpose, so disabling batching there is the safer trade.

src/utils/libfabric/libfabric_rail_manager.cpp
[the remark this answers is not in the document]
akkart-aws: Fixed — same issue fengjica flagged above.

akkart-aws: You're right — that was a regression I introduced in this PR; upstream used rr_idx % remote_selected_endpoints.size(). fixed it.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Done

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Done — adopted this structure: a last_desc_idx_per_rail array sized by rail count, with the FI_MORE decision made inline in the submit loop, so the per-descriptor apply_fi_more array is gone (and with it the now-unused desc_count parameter).

One adjustment to the sketch: pos_in_group != FI_MORE_BATCH_SIZE - 1 doesn't bound the per-rail batch when descriptors interleave across rails — every 16th position can land on the same rail, letting another rail's deferred batch grow with desc_count. Used a per-rail posts_since_flush counter (also rail-count-sized) for the cap instead.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1204,7 +1147,60 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+    const bool batch_writes = allow_fi_more && op_type == nixlLibfabricReq::WRITE;
+
+    // Rail of a single-rail WRITE descriptor, or -1 if it does not participate in batching.
+    auto batching_rail = [&](int i) -> int {
+        auto *md = static_cast<nixlLibfabricPrivateMetadata *>(local[i].metadataP);
+        if (!md || md->selected_rails_.empty()) {
+            return -1;
+        }
+        // Striped descriptors (same condition as use_striping in prepareAndSubmitTransfer)
+        // are split across their rails and posted without FI_MORE, so they are not part of
```
fengjica: Is it possible to avoid duplicating this logic?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1204,7 +1147,60 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+
+    // True if descriptor i should carry FI_MORE: not its rail's last post in this chunk and the
+    // rail's running batch is below NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE; otherwise it flushes.
+    auto use_fi_more = [&](int i) -> bool {
+        const int rail_id = batch_writes ? batching_rail(i) : -1;
+        if (rail_id < 0) {
+            return false;
+        }
+        if (i == last_desc_idx_per_rail[rail_id] ||
+            posts_since_flush[rail_id] >= NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE - 1) {
```
fengjica: Will "==" work? I don't think it will exceed that.

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1204,7 +1147,60 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+            const int rail_id = batching_rail(i);
+            if (rail_id >= 0) {
+                last_desc_idx_per_rail[rail_id] = i;
+            }
+        }
+    }
+
+    // True if descriptor i should carry FI_MORE: not its rail's last post in this chunk and the
+    // rail's running batch is below NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE; otherwise it flushes.
+    auto use_fi_more = [&](int i) -> bool {
```
fengjica: can we use explicit capture?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1204,7 +1147,60 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+    // A FI_MORE post rings no doorbell; a rail's queued batch is only submitted by a later
+    // non-FI_MORE post on the same rail. Rails are resolved per-buffer, so descriptors of one
+    // round-robin group can land on different rails: every rail this chunk touches must have its
+    // last post flushed, or its batch would never be submitted.
+    // allow_fi_more is false on the thread-pool path: chunks on other threads can interleave
+    // posts on the same rail, so a per-chunk walk cannot know a rail's true last post there.
+    const bool batch_writes = allow_fi_more && op_type == nixlLibfabricReq::WRITE;
+
+    // Rail of a single-rail WRITE descriptor, or -1 if it does not participate in batching.
+    auto batching_rail = [&](int i) -> int {
```
fengjica: can we do explicit capture?

src/plugins/libfabric/libfabric_backend.cpp
```diff
@@ -1204,7 +1147,60 @@ nixlLibfabricEngine::postXferDescriptors(nixlLibfabricReq::OpType op_type,
...
+            if (rail_id >= 0) {
+                last_desc_idx_per_rail[rail_id] = i;
+            }
+        }
+    }
+
+    // True if descriptor i should carry FI_MORE: not its rail's last post in this chunk and the
+    // rail's running batch is below NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE; otherwise it flushes.
+    auto use_fi_more = [&](int i) -> bool {
+        const int rail_id = batch_writes ? batching_rail(i) : -1;
```
fengjica: Is it possible to avoid nested lambda? 

In general, do the lambdas make it easier or harder compared to functions? I feel it is hard to follow.

ovidiusm (dismissed):

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Done — merged the condition into a single nixlLibfabricRailManager::usesStriping(transfer_size, rail_count); shouldUseStriping is removed and both call sites use it.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: Converted to a private member function batchingRail() — no captures; all inputs are parameters.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: You're right — the counter resets on every flush and increments by one otherwise, so it never exceeds NIXL_LIBFABRIC_FI_MORE_BATCH_SIZE - 1. Changed to ==.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: now a private member function useFiMore() with explicit parameters.

src/plugins/libfabric/libfabric_backend.cpp
akkart-aws: replaced both lambdas with private member functions (batchingRail, useFiMore), so there's no nesting and the data flow is visible in the signatures

ovidiusm:
Seems that it needs a newline after `reserveBaseOffset();` in  libfabric_rail_manager.h

akkart-aws:
> Seems that it needs a newline after `reserveBaseOffset();` in libfabric_rail_manager.h

Yes Fixed it.

ovidiusm (approved):

fengjica (approved):

PR #1978, opened 2026-07-22
NIXL stride desc API

iyastreb:
## What?
When using 2 step NIXL API (prep+make), inference frameworks have to allocate a huge array of descriptors (6.5M in my case) which are later compressed by NIXL into a strided array internally. This internal representation is tiny, because for each layer we store just a single stride desc.

## Why?
The problem is initial registration: due to the huge size of desc array it can take up to several seconds to create and populate the desc array on Python side, which is later merged and compressed into a stride array.

## How?
The solution is ot expose NIXL stride API directly to clients, so that we eliminate completely the need of creating huge desc arrays.

```
Initial TTFT: 632  ->  397   ~50% improvement

Breakdown (vLLM):
_build_fa_remote: 112ms -> 0.1ms  1000x
get_xfer_descs:   75ms  -> 0.1ms  1000x
prep_xfer_dlist:  20ms  -> 0.1ms  200x
```

## Summary by CodeRabbit

* **New Features**
  * Added compressed strided transfer descriptors for repeated, equally sized blocks with fixed spacing.
  * Added C++ and Python APIs for preparing transfers from strided descriptor lists.
  * Python users can provide contiguous Nx5 NumPy arrays with an explicit memory type.
  * Preserved flattened block indexing while avoiding per-block expansion.
  * Added support for local and remote transfer preparation.

* **Bug Fixes**
  * Added validation for descriptor shape, data type, layout, coverage, memory registration, and invalid or empty transfer runs.


src/bindings/python/nixl_bindings.cpp
```diff
@@ -138,6 +138,34 @@ throw_nixl_exception(const nixl_status_t &status) {
     }
 }
 
+// Builds a compressed (strided) descriptor list from an Nx5 numpy array, where each row is a run
+// of `count` blocks of `len` bytes with consecutive block starts spaced `stride` bytes apart:
+// (addr, len, dev_id, stride, count). A dense run has stride == len.
+nixl_stride_dlist_t
+to_stride_dlist(nixl_mem_t mem, const py::array &descs) {
```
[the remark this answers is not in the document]
iyastreb: fixed

src/infra/nixl_memory_section.cpp
```diff
@@ -183,6 +183,53 @@ nixlMemSection::populate(const nixl_xfer_dlist_t &query,
...
+        const int s_index = base.getCoveringIndex(nixlBasicDesc(run.addr, span_len, run.devId));
+        if (s_index < 0) [[unlikely]] {
+            resp.clear();
+            return NIXL_ERR_UNKNOWN;
+        }
+
+        nixlMetaStrideDesc &md = resp.emplace(
+            run.addr, run.len, run.devId, base[s_index].metadataP, run.stride, run.count);
+        md.start_idx = start_idx;
+        start_idx += run.count;
```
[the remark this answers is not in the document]
iyastreb: done

src/api/cpp/nixl_descriptors.h
```diff
@@ -137,6 +137,43 @@ class nixlBasicDesc {
...
+ *        equal-length blocks, with consecutive block starts spaced `stride`
+ *        bytes apart (stride == len means a dense/contiguous run). Lets callers
+ *        express a large, regular access pattern without materializing every
+ *        block as a separate nixlBasicDesc.
+ */
+struct nixlStrideDesc : public nixlBasicDesc {
+    /** @var Byte distance between consecutive block starts (stride == len => dense) */
+    size_t stride = 0;
+    /** @var Number of equal-length blocks in this run */
+    size_t count = 0;
```
brminich, line 152: should you also change operator== or something else?

src/infra/nixl_memory_section.cpp
```diff
@@ -183,6 +183,53 @@ nixlMemSection::populate(const nixl_xfer_dlist_t &query,
...
+
+    size_t start_idx = 0;
+    for (int i = 0; i < n; ++i) {
+        const nixlStrideDesc &run = query[i];
+        if (run.count == 0 || run.len == 0 || run.stride < run.len) [[unlikely]] {
+            resp.clear();
+            return NIXL_ERR_INVALID_PARAM;
+        }
+
+        // Full byte extent of the run: from the first block start to the last block end.
```
brminich, line 215: Could we document in the public prepXferDlist(nixl_stride_dlist_t) API that each stride run’s full byte span must be covered by a single registered memory region? vLLM satisfies this by registering the whole backing allocation, but this requirement is not currently obvious to callers.

src/api/cpp/nixl_descriptors.h
iyastreb: implemented == and != like in the base class

src/infra/nixl_memory_section.cpp
iyastreb: added

brminich (dismissed):

src/api/python/_api.py
```diff
@@ -509,13 +509,22 @@ def prep_xfer_dlist(
             agent_name: Name of the agent. It can be "NIXL_INIT_AGENT", local agent name, or remote agent name
             xfer_list: List of transfer descriptors, can be list of memory region tuples, tensors,
                 Nx3 numpy array, or nixlXferDList. See get_xfer_descs for more details on the structure.
-            mem_type: Optional memory type necessary for list of memory regions.
+                An Nx5 numpy array is taken as a compressed (strided) list, where each row is
+                (address, len, device ID, stride, count): `count` blocks of `len` bytes spaced
+                `stride` bytes apart (stride == len is dense). This avoids materializing every
```
mkhazraee: Too much details about KV cache or so. Let's just describe what is the 5 to show a strided region. Adress being the first one, and len being each element len or so.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -654,21 +683,30 @@ PYBIND11_MODULE(_bindings, m) {
         .def(
             "prepXferDlist",
```
mkhazraee, line 699: Are we adding overload or replacing? The _api sounds overload, but here it feels we are replacing. I thought there would be 4 definitions. Or somehow py::array takes care of it?

src/core/agent_data.h
```diff
@@ -102,6 +102,13 @@ class nixlAgentData final : public nixlMetadataContext {
         void
         warnAboutEfaHardwareMismatch();
 
+        template<class dlist_t>
```
mkhazraee, line 105: I'm not sure to keep it just as a helper method inside nixl_agent, or move it to agent_data. Modularity wise do we have anything similar like this?

src/api/python/_api.py
iyastreb: updated, pls let me know what you think

src/core/agent_data.h
iyastreb: Well, nixl_agent is an API class, and it does not contain any private helpers, which is a good thing.
All the private helpers are in the agent_data, and so this helper too.
An identical pattern is used for many other functions like `loadRemoteMD`, `invalidateRemoteMD` where agent delegates handling to the data

src/bindings/python/nixl_bindings.cpp
iyastreb: We add an overload.
But at the same time I removed one of the 2 overloads of prepXferDlist(without agent name), so the git diff shows like I replaced it.
I thought this overload without agent name is not needed, because I assumed that bindings are only used directly from NIXL _api.py. But if some clients are using binding explicitly, then I can restore this removed overload

mkhazraee: I see, technically the bindings are our C++ API exposed, but at the same time there are corner cases we don't expose, like not all descriptor capabilities. How about we add it to the binding as one that just calls this one with the agent name included, to keep it backward compatible, and not add a lot of code. Very unlikely use case, just to be sure if someone was using it.

src/api/python/_api.py
mkhazraee: I guess if you just add ", with start address of `addr`, in the form (address, len, device ID, stride, count)" is enough. Main issue is we're redefining the other fields which were already available, and the two new ones, you already explained well, so those 5 lines explaining each field is a tad extra.

iyastreb: ok, done

src/bindings/python/nixl_bindings.cpp
iyastreb: ok, I added common template function and reuse it from all places

mkhazraee (dismissed):

ovidiusm (approved):

PR #1979, opened 2026-07-22
enabling -G flag, with meson arg -Dnvcc_device_debug=true

NitayAmiel:
## What?

This PR enables building and running NIXL (including nixl_ep) with CUDA **device debug** (`nvcc -G`), and adds a verification -G compilation to CI for future support.

It adds:

- A Meson option `-Dnvcc_device_debug=true` that passes `-G` (and host debug symbols where needed for nixl_ep).
- Targeted `__forceinline__` on selected UCX device helpers so debug builds still link.
- A compile-time workaround for a cooperative grid sync hang under `-G` (`NIXL_EP_GRID_SYNC_PRESYNC()` → `__syncthreads()` only when device debug is enabled).
-  A -G build command in CI (we added an arch detection too for minimizing the time of compilation which is longer that usual under -G). 

## Why?

Device-level debugging with **cuda-gdb** and similar tools helps NIXL developers investigate GPU-side behavior in nixl_ep and the UCX device API path.

Today, enabling `-G` is impractical: builds can fail (register / nvlink limits) or runs can hang at `cg::this_grid().sync()`. This PR removes those blockers so debug builds are usable while keeping release behavior unchanged.

## How?

Two main issues had to be addressed when adding `-G`:

### 1. Register pressure and inlining

With `-G`, the compiler often **does not inline** functions that are inlined under normal optimization. That increased register usage and led to **nvlink** failures (e.g. entry kernels with a low register limit calling `nixlGetPtr` and similar helpers as separate functions with much higher register counts).

We added __forceinline__ on a small set of hot device functions — nixlPut, nixlAtomicAdd, nixlGetPtr (UCX device API) and p2p_ptr_get, combine_token (nixl_ep) — after confirming they are inlined in non--G builds, so release performance should be unaffected.

This trades slightly less convenient single-step debugging in those few call sites for a build that actually completes and can be debugged at all.

### 2. Hang in `cg::this_grid().sync()` under `-G`

The second issue is an **inherent problem** the cooperative-groups call `cg::this_grid().sync()` with `-G` enabled. In the send/recv path in `nixl_ep_ll.cu`, we observed hangs around this grid sync(more precisely, CTA threads got stuck on this sync op, while the other completed it).
By our analysis, there appears to be a bug in that sync path (not resolved in the newer CUDA versions we tested). In our view, it is related to the fact that, under `-G`, barriers inside this function can end up associated with the **same instruction addresses** in the generated code, which can cause "confusion" between CTA-level participation and warp-level threads.
This class of issue has been reported several times in NVBUGS to the relevant teams; so far we have not seen a satisfactory fix in CUDA. We also explored a solution that would require changes in the cooperative_groups library itself, which is out of scope for the NIXL project.
We found that calling another sync primitive—**`__syncthreads()`**—immediately before `cg::this_grid().sync()` avoids the hang. To avoid affecting performance in normal builds, that extra sync is only compiled in when `-Dnvcc_device_debug=true` is passed to Meson (via `NIXL_EP_GRID_SYNC_PRESYNC()`). We consider this a **temporary** workaround until the CUDA team provides a permanent fix; we are publishing this PR now so others can use device debugging in the meantime.
## Build and run

**Meson option:** `nvcc_device_debug` (boolean, default: `false`)

**Release build (unchanged):**

```bash
meson setup build -Dbuild_nixl_ep=true -Dbuildtype=release <other required options, e.g. -Ducx_path=...>
ninja -C build
ninja -C build install
```
**Device debug build(-G):**
```bash
meson setup build -Dbuild_nixl_ep=true -Dnvcc_device_debug=true <other required options, e.g. -Ducx_path=...>
ninja -C build
ninja -C build install
```

## Summary by CodeRabbit

* **Performance**
  * Improved GPU kernel inlining for more efficient device-side execution.
* **Build Improvements**
  * Release CUDA builds now consistently apply release-mode settings.
  * Debug builds retain their configured optimization and debugging options.
* **Reliability**
  * Added debug synchronization checks to improve validation of active data-transfer phases.

### DIFF
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -435,6 +435,23 @@ else
     ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} && ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} install
     mkdir -p dist && cp ${NIXL_BUILD_DIR}/src/bindings/python/nixl-meta/nixl-*.whl dist/
 
+    # Verify the nixl_ep device-debug (-G) build compiles, so future changes
+    # keep the nvcc_device_debug feature working. Compile-only (no install/run).
+    # nixl_ep requires sm_90 or newer, so detect the GPU compute capability and
+    # only run the check when it qualifies, building for the detected arch.
+    NIXL_EP_GPU_COMPUTE_CAP=""
+    if test -d "$CUDA_HOME" && command -v nvidia-smi >/dev/null 2>&1; then
+        NIXL_EP_GPU_COMPUTE_CAP=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d '. ')
+    fi
+    if [ -n "${NIXL_EP_GPU_COMPUTE_CAP}" ] && [ "${NIXL_EP_GPU_COMPUTE_CAP}" -ge 90 ] 2>/dev/null; then
+        NIXL_DEVICE_DEBUG_BUILD_DIR=${NIXL_DEVICE_DEBUG_BUILD_DIR:-nixl_device_debug_build}
+        # shellcheck disable=SC2086
+        meson setup ${NIXL_DEVICE_DEBUG_BUILD_DIR} --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Drust=false ${EXTRA_BUILD_ARGS} -Dlibfabric_path="${LIBFABRIC_INSTALL_DIR}" -Dbuild_nixl_ep=true -Dnvcc_device_debug=true -Dnixl_cuda_arch_list=${NIXL_EP_GPU_COMPUTE_CAP}
+        ninja -j"$NPROC" -C ${NIXL_DEVICE_DEBUG_BUILD_DIR}
+    else
+        echo "nixl_ep -G (nvcc_device_debug) check skipped: requires an sm_90+ GPU (detected compute_cap='${NIXL_EP_GPU_COMPUTE_CAP:-none}')."
+    fi
+
     # TODO(kapila): Copy the nixl.pc file to the install directory if needed.
     # cp ${BUILD_DIR}/nixl.pc ${INSTALL_DIR}/lib/pkgconfig/nixl.pc
 
--- a/.gitlab/build.sh
+++ b/.gitlab/build.sh
@@ -442,23 +442,6 @@ else
     ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} && ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} install
     mkdir -p dist && cp ${NIXL_BUILD_DIR}/src/bindings/python/nixl-meta/nixl-*.whl dist/
 
-    # Verify the nixl_ep device-debug (-G) build compiles, so future changes
-    # keep the nvcc_device_debug feature working. Compile-only (no install/run).
-    # nixl_ep requires sm_90 or newer, so detect the GPU compute capability and
-    # only run the check when it qualifies, building for the detected arch.
-    NIXL_EP_GPU_COMPUTE_CAP=""
-    if test -d "$CUDA_HOME" && command -v nvidia-smi >/dev/null 2>&1; then
-        NIXL_EP_GPU_COMPUTE_CAP=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d '. ')
-    fi
-    if [ -n "${NIXL_EP_GPU_COMPUTE_CAP}" ] && [ "${NIXL_EP_GPU_COMPUTE_CAP}" -ge 90 ] 2>/dev/null; then
-        NIXL_DEVICE_DEBUG_BUILD_DIR=${NIXL_DEVICE_DEBUG_BUILD_DIR:-nixl_device_debug_build}
-        # shellcheck disable=SC2086
-        meson setup ${NIXL_DEVICE_DEBUG_BUILD_DIR} --prefix=${INSTALL_DIR} -Ducx_path=${UCX_INSTALL_DIR} -Drust=false ${EXTRA_BUILD_ARGS} -Dlibfabric_path="${LIBFABRIC_INSTALL_DIR}" -Dbuild_nixl_ep=true -Dnvcc_device_debug=true -Dnixl_cuda_arch_list=${NIXL_EP_GPU_COMPUTE_CAP}
-        ninja -j"$NPROC" -C ${NIXL_DEVICE_DEBUG_BUILD_DIR}
-    else
-        echo "nixl_ep -G (nvcc_device_debug) check skipped: requires an sm_90+ GPU (detected compute_cap='${NIXL_EP_GPU_COMPUTE_CAP:-none}')."
-    fi
-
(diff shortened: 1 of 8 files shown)


meson_options.txt
```diff
@@ -43,6 +43,7 @@ option('sanitizer', type: 'combo',
 # Tracing (nixl::trace)
 option('with_trace', type: 'boolean', value: true, description: 'Build the nixl::trace tracing facade')
 option('trace_backends', type: 'string', value: 'nvtx', description: 'Comma-separated tracing backends to compile in (e.g. nvtx, chakra)')
+option('nvcc_device_debug', type: 'boolean', value: false, description: 'Add -G flag to nvcc for CUDA device debug info')
```
ebarilanM: i dont think we should add a dedicated option why not based on buildtype == debug?

examples/device/ep/meson.build
```diff
@@ -128,13 +128,18 @@ nixl_ep_rpath += ':' + nixl_lib_dir
...
+# and debug so a global buildtype=debug cannot leak --device-debug (-G) onto
+# the EP nvcc line.
 nixl_ep_override_options = []
-if get_option('buildtype') == 'debug'
+if get_option('nvcc_device_debug')
+    nixl_ep_override_options = ['optimization=3']
+    nixl_ep_cuda_args += ['-G', '-DNIXL_EP_NVCC_DEVICE_DEBUG']
+    nixl_ep_cpp_args += ['-g']
+    nixl_ep_cuda_args += ['-Xcompiler', '-g']
+elif get_option('buildtype') == 'debug'
```
ebarilanM: seems like now we can remove this limitation not by adding a specific flags

.gitlab/build.sh
```diff
@@ -435,6 +435,23 @@ else
     ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} && ninja -j"$NPROC" -C ${NIXL_BUILD_DIR} install
     mkdir -p dist && cp ${NIXL_BUILD_DIR}/src/bindings/python/nixl-meta/nixl-*.whl dist/
 
+    # Verify the nixl_ep device-debug (-G) build compiles, so future changes
```
ebarilanM: does this the right place to add a CI testing that this building is works? @rakhmets

.gitlab/build.sh
```diff
@@ -435,6 +435,23 @@ else
...
+    # keep the nvcc_device_debug feature working. Compile-only (no install/run).
+    # nixl_ep requires sm_90 or newer, so detect the GPU compute capability and
+    # only run the check when it qualifies, building for the detected arch.
+    NIXL_EP_GPU_COMPUTE_CAP=""
+    if test -d "$CUDA_HOME" && command -v nvidia-smi >/dev/null 2>&1; then
+        NIXL_EP_GPU_COMPUTE_CAP=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d '. ')
+    fi
+    if [ -n "${NIXL_EP_GPU_COMPUTE_CAP}" ] && [ "${NIXL_EP_GPU_COMPUTE_CAP}" -ge 90 ] 2>/dev/null; then
```
ebarilanM: why not based on BUILD_NIXL_EP?

examples/device/ep/meson.build
```diff
@@ -128,13 +128,18 @@ nixl_ep_rpath += ':' + nixl_lib_dir
...
+# By default nixl ep is not built with -G due to register usage limits;
+# build the EP target as release even in a global debug build unless
+# nvcc_device_debug is explicitly requested. Override buildtype, optimization,
+# and debug so a global buildtype=debug cannot leak --device-debug (-G) onto
+# the EP nvcc line.
 nixl_ep_override_options = []
-if get_option('buildtype') == 'debug'
+if get_option('nvcc_device_debug')
+    nixl_ep_override_options = ['optimization=3']
+    nixl_ep_cuda_args += ['-G', '-DNIXL_EP_NVCC_DEVICE_DEBUG']
```
ebarilanM: Seems like there is a standard in CUDA code #ifndef NDEBUG

ebarilanM (approved):
LGTM @itayalroy  could you confirm all benchmark run in release mode?

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -117,7 +117,7 @@ nixlGpuGetXferStatus(nixlGpuXferStatusH &xfer_status) {
  * @return NIXL_ERR_BACKEND An error occurred in UCX backend.
  */
 template<nixl_gpu_level_t level = nixl_gpu_level_t::THREAD>
-__device__ nixl_status_t
+__device__ nixl_status_t __forceinline__
```
rakhmets, line 120:
```suggestion
__device__ __forceinline__ nixl_status_t
```

src/api/gpu/ucx/nixl_device.cuh
```diff
@@ -157,7 +157,7 @@ nixlPut(const nixlMemViewElem &src,
...
```
rakhmets, line 160:
```suggestion
__device__ __forceinline__ nixl_status_t
```

examples/device/ep/meson.build
```diff
@@ -128,16 +128,9 @@ nixl_ep_rpath += ':' + nixl_lib_dir
...
 nixl_ep_rpath += ':' + join_paths(nixl_lib_dir, 'plugins')
 
-# For now, nixl ep cannot be built with -G due to register usage limits;
-# build the EP target as release even in a global debug build (same flags
-# we ship in release wheels), and add -g back so we keep debug symbols.
-# Override buildtype, optimization, and debug explicitly so a global
-# buildtype=debug cannot leak --device-debug (-G) onto the EP nvcc line.
+# nixl_ep follows the global buildtype: a debug build gets device debug (-G)
+# and host -g from Meson's CUDA debug args, while a release build gets neither.
 nixl_ep_override_options = []
```
itayalroy, line 133: This is unused, remove it

meson.build
```diff
@@ -293,6 +293,13 @@ if cuda_dep.found()
     # Build all CUDA targets with C++20 to match the core's standard and ABI.
     add_project_arguments('-std=c++20', language: 'cuda')
 
+    # Debug builds get device debug (-G) from Meson's CUDA debug args; release
+    # builds define NDEBUG (matching the C/C++ flags) so device-debug-only code
+    # (e.g. the cg::this_grid().sync() __syncthreads() workaround) compiles out.
+    if get_option('buildtype') == 'release'
+        add_project_arguments('-DNDEBUG', language: 'cuda')
```
itayalroy, line 300: What about `debugoptimized`, `custom`, etc? are you sure that meson sets -G for these? to me it seems odd, I would assume -G is only set for "debug".

Also, having "NDEBUG" flag for release mode and then checking "if not ndebug" in code is extremely confusing. The C flag should be DEBUG and we should set it whenever meson sets -G, or better, use a built in C flag that indicates whether device debug is enabled if there is any

src/api/gpu/ucx/nixl_device.cuh
itayalroy: If these `__forceinline__`s are a debug-only hack, why are we adding them for non debug builds as well? we should not introduce any performance risks on non-debug builds just to pass debug compilation

itayalroy (changes requested):

PR #1982, opened 2026-07-22
[DO NOT MERGE] CI: test held Slurm allocations (liveness-coupled salloc)

NirWolfer:
## What

CI-only experiment: loads `swx-jenkins-lib@slurm-hold-allocation` in every slurmCI step of the **dl-gpu** job via the existing `ref:` override. That branch changes `slurm.allocation()` (SSH mode) to hold the allocation open with a background `ssh -tt ... salloc ... sleep infinity` holder instead of `salloc --no-shell`, so Slurm revokes the allocation as soon as the build's agent dies — however it dies.

## Why

When a new commit is pushed to a PR, the dispatcher hard-kills (`doKill`) the stale builds, `pipeline_stop` never runs, and Slurm allocations hang until their `--time` limit. This tests coupling the allocation's lifetime to the build agent instead.

## Test plan

1. Run the dl pipeline for this PR; wait for `Slurm allocation job ID: <N>` in the console.
2. Kill the build (or push a second commit so the dispatcher does it).
3. Verify job `<N>` disappears from `squeue` on dlcluster within seconds (`dl-gpu-ep` intentionally stays on the default lib as the control group).
4. Let a run finish untouched: all test steps pass against the held allocation and `pipeline_stop` cleans up normally.

**Not for merge** — test scaffolding only; `ci-overview.md` deliberately not synced.

### DIFF
--- a/.ci/jenkins/lib/test-dl-matrix.yaml
+++ b/.ci/jenkins/lib/test-dl-matrix.yaml
@@ -119,6 +119,7 @@ steps:
     module: slurmCI
     run: allocation
     args:
+      ref: 'slurm-hold-allocation'
       partition: "${SLURM_PARTITION}"
       headNode: "${SLURM_HEAD_NODE}"
       headUser: "${SLURM_HEAD_USER}"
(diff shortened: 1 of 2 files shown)


NirWolfer:
Validation complete - test scaffolding no longer needed. Results: hard-killed build freed its Slurm allocation in ~4s via the salloc SIGHUP path (job 1682328, COMPLETED at 14:36:54 vs pod kill at ~14:36:50), and a full untouched run behaved identically to the current flow with pipeline_stop cleaning up normally. The real change ships in swx-jenkins-lib.

PR #1987, opened 2026-07-23
Fixes a race in the UCX backend between synchronous rkey unpack and endpoint teardown.

### LINKED ISSUE #1986: [UCX] Race between endpoint close and rkey unpack can crash loadRemoteMD after peer restart
## Summary

  A production failure reported in
  [vllm-project/vllm#49238](https://github.com/vllm-project/vllm/issues/49238)
  shows that a surviving Decode process can segfault inside
  `ucp_ep_rkey_unpack()` after a Prefill process is restarted in a
  Prefill/Decode-disaggregated deployment.

  Instead of returning a recoverable error such as
  `NIXL_ERR_REMOTE_DISCONNECT`, one Decode TP worker crashes inside NIXL/UCX,
  causing the entire vLLM Decode instance to terminate and restart.

  The NIXL UCX backend currently allows an endpoint to be force-closed by the
  asynchronous UCX error callback while, or immediately before, another thread
  uses the same endpoint to unpack a remote key.

  ## Upstream vLLM issue

  The production report, complete logs, environment, and crash stack are
  available here:

  - https://github.com/vllm-project/vllm/issues/49238

  The failure was reproduced with:

  - vLLM 0.25.0 and 0.25.1
  - NIXL 1.3.1
  - Kubernetes Prefill/Decode disaggregation
  - `NixlConnector` / `NixlPullConnector`
  - Prefill TP=1
  - Decode TP=4
  - NVIDIA H200
  - UCX over Mellanox InfiniBand/RoCE (`mlx5`)

  ## Production sequence

  1. Start one Prefill instance and one Decode instance.
  2. Send request 1.
  3. Decode loads the Prefill metadata and completes the KV transfer.
  4. Restart the Prefill Pod without restarting Decode.
  5. The Decode-side UCX endpoints to the old Prefill eventually time out and
     transition from `CONNECTED` to `FAILED`.
  6. Start a replacement Prefill with a new NIXL agent ID.
  7. Send request 2.
  8. Decode calls `loadRemoteMD()` for the replacement Prefill.
  9. While the new connection and remote metadata are being loaded, UCX reports
     an RDMA I/O error and force-closes an endpoint.
  10. Three Decode TP workers return `NIXL_ERR_REMOTE_DISCONNECT`.
  11. One Decode TP worker segfaults in `ucp_ep_rkey_unpack()`.

  The failure of one worker shuts down the entire vLLM Decode executor, causing
  a full Decode-instance outage instead of a request-level retryable failure.

  ## Relevant log sequence

  The endpoint used by the crashing TP worker is connected and then fails while
  `loadRemoteMD()` is still in progress:

  ```text
  20:15:24.486575  Loading remote metadata for the replacement Prefill agent

  20:15:24.499522  ep 0x72c3150610c0:
                   UNINITIALIZED -> CONNECTED

  20:15:24.527907  UCX error handling callback:
                   Input/output error

  20:15:24.527921  ep 0x72c3150610c0:
                   CONNECTED -> FAILED

  20:15:24.531680  The loadRemoteMD thread is still creating endpoints

  ... subsequently ...

  Segfault in ucp_ep_rkey_unpack

  This ordering indicates that loadRemoteMD() may try to unpack an rkey using
  an endpoint that has already failed and been force-closed. A concurrent overlap
  between rkey unpack and endpoint close is also possible.

  ## Crash stack

  ucp_ep_rkey_unpack
  nixl::ucx::rkey::unpackUcpRkey
  nixl::ucx::rkey::rkey
  nixlUcxEngine::internalMDHelper
  nixlRemoteSection::addDescList
  nixlRemoteSection::loadRemoteData
  nixlAgentData::loadRemoteSections
  nixlAgent::loadRemoteMD
  pybind11 loadRemoteMD binding

chaunceyjiang:
Fix https://github.com/ai-dynamo/nixl/issues/1986

## What?
  Fixes a race in the UCX backend between synchronous rkey unpack and endpoint teardown.

  After a remote peer disconnects or restarts, the UCX error callback may force-close an endpoint while, or immediately
  before, loadRemoteMD() calls ucp_ep_rkey_unpack() on the same endpoint. This can cause a native segfault instead of
  returning a recoverable NIXL error.

  This PR:

  - Serializes rkey unpack with both error-driven and normal endpoint close.
  - Checks endpoint state before calling ucp_ep_rkey_unpack().
  - Returns NIXL_ERR_REMOTE_DISCONNECT when unpack is attempted on a failed endpoint.
  - Preserves the status through the rkey constructor and internalMDHelper().
  - Adds focused tests for close-during-unpack and unpack-after-close.

## Why?
In vLLM P/D disaggregation, restarting a Prefill Pod can leave the surviving Decode process handling asynchronous UCX
  endpoint errors while loading metadata for the replacement Prefill.

  The reported crash occurs in:

  ucp_ep_rkey_unpack
  → nixl::ucx::rkey::unpackUcpRkey
  → nixlUcxEngine::internalMDHelper
  → nixlAgent::loadRemoteMD

  One Decode TP worker segfaults while the other workers return NIXL_ERR_REMOTE_DISCONNECT. The crashed worker brings down
  the entire Decode instance.

  NIXL previously called ucp_ep_rkey_unpack() without synchronizing against endpoint close or checking whether the endpoint
  had already failed.
## How?
  An endpoint-level mutex now protects:

  - Endpoint state validation followed by rkey unpack.
  - Force-close from the UCX error callback.
  - Normal endpoint close during teardown.

  The endpoint state remains atomic so existing lock-free transfer-path state checks are preserved.

  Since the rkey constructor cannot return a NIXL status directly, a typed rkey_error carries the failure status to
  internalMDHelper(). This allows a remote disconnect to remain NIXL_ERR_REMOTE_DISCONNECT instead of being converted to the
  generic NIXL_ERR_BACKEND.

  The tests cover both relevant orderings:

  rkey unpack starts → endpoint close waits
  endpoint closes first → unpack returns remote disconnect without calling UCX

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UCX connection error and disconnect handling for more reliable endpoint state transitions.
  * Prevented metadata creation when a remote endpoint is unavailable or in a failed transmit state.
  * Ensured failed connections are closed consistently and reported with the appropriate remote-disconnect status.

* **Tests**
  * Added automated coverage for UCX endpoint connection-reset and failure scenarios.
  * Updated optional UCX test integration for supported builds.


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -103,30 +105,45 @@ nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
         return;
     case nixl::ucx::ep_state_t::CONNECTED:
         setState(nixl::ucx::ep_state_t::FAILED);
-        request = ucp_ep_close_nb(ucp_ep, UCP_EP_CLOSE_MODE_FORCE);
+        request = ops_.closeNb(ops_.context, ucp_ep, UCP_EP_CLOSE_MODE_FORCE);
```
rakhmets: I think it would be better to close UCP endpoint in `nixlUcxEp::~nixlUcxEp` (which calls `nixlUcxEp::closeImpl`). And `nixlUcxEp::err_cb` only changes state. So, it will be easier to control the closing of the UCP endpoint.

P.S. It seems that the atomic would be enough in this case, and mutex can be avoided.

src/plugins/ucx/ucx_utils.h
```diff
@@ -65,13 +84,17 @@ class nixlUcxEp {
 
     [[nodiscard]] nixl_status_t
     checkTxState() const noexcept {
-        return nixl::ucx::toNixlStatus(state);
+        return nixl::ucx::toNixlStatus(state.load(std::memory_order_acquire));
```
rakhmets: I think the change in this line is not needed.

src/plugins/ucx/ucx_utils.h
```diff
@@ -41,11 +43,28 @@ using nixlUcxReq = void *;
...
+
+    void *context = nullptr;
+    rkey_unpack_fn_t rkeyUnpack =
+        [](void *, ucp_ep_h ep, const void *buffer, ucp_rkey_h *rkey) {
+            return ucp_ep_rkey_unpack(ep, buffer, rkey);
+        };
+    close_nb_fn_t closeNb = [](void *, ucp_ep_h ep, unsigned mode) {
+        return ucp_ep_close_nb(ep, mode);
+    };
+};
```
rakhmets: Not sure I understand the need of this structure. Could you please explain the purpose of the structure.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -971,6 +971,10 @@ nixlUcxEngine::internalMDHelper (const nixl_blob_t &blob,
             it->second, makePublicMetadataRkeys(it->second, workers_.size(), blob.data()));
         return NIXL_SUCCESS;
     }
+    catch (const nixl::ucx::rkey_error &e) {
+        NIXL_ERROR << e.what();
+        return e.status();
+    }
```
rakhmets: I'm not sure that we need a new kind of exception. At the moment std errors are used.

src/plugins/ucx/rkey.cpp
```diff
@@ -21,16 +21,15 @@
 #include "ucx_utils.h"
 
 namespace nixl::ucx {
```
iyastreb, line 23: I think we can implement this fix much simpler, without new types etc
1. We can make the state atomic as @rakhmets proposed
`std::atomic<nixl::ucx::ep_state_t> state{nixl::ucx::ep_state_t::UNINITIALIZED};`

2. err_cb only marks state FAILED — never closes.
3. closeImpl force-closes a FAILED ep (since err_cb no longer does) and sets a terminal state on the CONNECTED path too
4. Guard the rkey path with the same check the data path already uses, returning NIXL_ERR_REMOTE_DISCONNECT — placed in internalMDHelper so no new exception type is needed:
```
const auto it = remoteConnMap.find(agent);
if (it == remoteConnMap.end())
    return NIXL_ERR_NOT_FOUND;
for (size_t i = 0; i < workers_.size(); ++i) {
    const nixl_status_t st = it->second->getEp(i)->checkTxState();
    if (st != NIXL_SUCCESS)
        return st;                 // clean REMOTE_DISCONNECT, matches the "other workers"
}

File: src/plugins/ucx/ucx_utils.h (hunk 3 above)
Comment (chaunceyjiang) [reply src/plugins/ucx/ucx_utils.h]: This was code I used for local testing and I forgot to remove it. Sorry about that.

File: src/plugins/ucx/rkey.cpp (hunk 5 above)
Comment (chaunceyjiang) [reply src/plugins/ucx/rkey.cpp:23]: @iyastreb Thanks! This approach is indeed much simpler.

File: test/gtest/unit/ucx/ucx_ep_test.cpp [hunk 6]
```diff
@@ -0,0 +1,33 @@
+/*
```
iyastreb: maybe better name is ucx_ep_test

chaunceyjiang: Done.

iyastreb (dismissed):

chaunceyjiang:
> [blossom-ci](https://github.com/ai-dynamo/nixl/actions/runs/30074114435) 


Hi @iyastreb, it looks like this CI failure is unrelated to my code.

src/plugins/ucx/ucx_utils.h
```diff
@@ -65,7 +66,7 @@ class nixlUcxEp {
 
     [[nodiscard]] nixl_status_t
     checkTxState() const noexcept {
-        return nixl::ucx::toNixlStatus(state);
+        return nixl::ucx::toNixlStatus(state_.load(std::memory_order_acquire));
```
rakhmets:
```suggestion
        return nixl::ucx::toNixlStatus(state_);
```

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -103,57 +103,70 @@ nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
...
 
 void
 nixlUcxEp::setState(nixl::ucx::ep_state_t new_state) {
-    NIXL_ASSERT(new_state != state);
-    NIXL_DEBUG << "ep " << eph << ": state " << state << " -> " << new_state;
-    state = new_state;
+    const auto old_state = state_.load(std::memory_order_relaxed);
+    NIXL_ASSERT(new_state != old_state);
+    NIXL_DEBUG << "ep " << eph << ": state " << old_state << " -> " << new_state;
+    state_.store(new_state, std::memory_order_release);
```
rakhmets:
```suggestion
    state_ = new_state;
```

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -103,57 +103,70 @@ nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
...
 }
 
 nixl_status_t
 nixlUcxEp::closeImpl() {
     ucs_status_ptr_t request = nullptr;
+    const auto current_state = state_.load(std::memory_order_acquire);
```
rakhmets:
```suggestion
    const nixl::ucx::ep_state_t current_state = state_;
```

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -86,15 +86,15 @@ err_cb_wrapper(void *arg, ucp_ep_h ucp_ep, ucs_status_t status) {
 
 void
 nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
-    ucs_status_ptr_t request;
+    const auto current_state = state_.load(std::memory_order_relaxed);
```
rakhmets, line 89: Could you please explain why the memory order is relaxed here?

chaunceyjiang: I initially used memory_order_relaxed because this load is only used to inspect the current endpoint state and select the corresponding state-machine branch. It does not publish or consume any associated data, so I thought atomicity alone was sufficient here

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -103,57 +103,70 @@ nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
...
         }
 
         if (UCS_PTR_IS_ERR(request)) {
             eph = nullptr;
+            setState(nixl::ucx::ep_state_t::DISCONNECTED);
             return nixl::ucx::ucsToNixlStatus(UCS_PTR_STATUS(request));
         }
 
         ucp_request_free(request);
+        setState(nixl::ucx::ep_state_t::DISCONNECTED);
```
rakhmets: Is this really necessary to setState in `nixlUcxEp::closeImpl`/`nixlUcxEp::~nixlUcxEp`? IMO, it's not needed.

chaunceyjiang: Done.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -103,40 +103,49 @@ nixlUcxEp::err_cb(ucp_ep_h ucp_ep, ucs_status_t status) {
...
         // The EP has not been connected, or already disconnected.
         // Nothing to do.
         NIXL_ASSERT(eph == nullptr);
         return NIXL_SUCCESS;
-    case nixl::ucx::ep_state_t::FAILED:
-        // The EP was closed in error callback, just return error.
+    case nixl::ucx::ep_state_t::FAILED: {
+        if (eph == nullptr) {
+            return NIXL_ERR_REMOTE_DISCONNECT;
+        }
```
rakhmets: I think this block can be removed, since this is not possible in the current code.

rakhmets (dismissed):

src/plugins/ucx/ucx_utils.cpp
chaunceyjiang: Removed

chaunceyjiang:
Hi @iyastreb, @ColinNV it looks like this CI failure is unrelated to my code.

iyastreb (dismissed):

test/gtest/unit/ucx/ucx_ep_test.cpp
iyastreb: Was this test executed by CI?
I didn't find confirming logs

Maybe instead of creating new file you could just extend existing test error_handling, so that:
1. We don't have just another test binary
2. It's automatically picked up by CI
3. It runs on different configurations (plain, threadpool etc)

iyastreb (commented):
pls check if test is in CI

test/gtest/unit/ucx/ucx_ep_test.cpp
chaunceyjiang: > Maybe instead of creating new file you could just extend existing test error_handling,

OK.

chaunceyjiang:
> pls check if test is in CI

@iyastreb This test is now included in CI.

chaunceyjiang:
> [nixl-ci-dl-gpu-ep](http://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-dl-gpu-ep/detail/nixl-ci-dl-gpu-ep/518/pipeline/)

Hi @iyastreb, it looks like this CI failure is unrelated to my code.

iyastreb (approved):

ovidiusm:
@iyastreb We need to think about how to add test coverage for this kind of issues

PR #1258, opened 2026-01-28
Ofi neuron support

akkart-aws:
## What?
This PR adds AWS Neuron accelerator support to NIXL, enabling the library to work with both NVIDIA GPUs and AWS Neuron devices (INF2, TRN1, TRN2, TRN3).

## Why?
NIXL previously only supported NVIDIA GPUs. AWS Neuron accelerators are specialized ML accelerators that require different device detection and memory registration mechanisms. This PR extends NIXL to support both accelerator types.

#### Neuron Device Detection:
- Added runtime loading of libnrt.so to query Neuron device PCI addresses via nrt_get_attached_efa_bdf()
- Implemented system accelerator type detection (CUDA/NEURON/SYSTEM) at initialization
- Extended topology discovery to detect both NVIDIA (vendor 0x10de) and AWS Neuron (vendor 0x1d0f) devices
- Added Neuron device ID detection for INF2 (0x7264), TRN1 (0x7164), TRN2 (0x7364), and TRN3 (0x7564, 0x7565) families
#### Memory Registration:
- Implemented FI_HMEM_NEURON interface support in libfabric memory registration
- Added conditional CUDA context management (only for CUDA systems)
- Unified PCI-to-EFA device mapping to support both GPU and Neuron device queries
#### Build System:
- Added dl (dynamic linking) dependency for runtime libnrt.so loading

PR #1416, opened 2026-03-09
Fix for PCIe to libfabric mapping for CXI provider

joshishr:
## Summary by CodeRabbit

* **New Features**
  * Device topology discovery and mapping now include CXI devices alongside EFA.

* **Improvements**
  * Validation logic generalized to run provider-aware checks for multiple providers.
  * Provider initialization tuned for better detection and compatibility.
  * Mapping initialization now retries when advanced memory features are unavailable to improve robustness.

* **Behavior**
  * EFA-specific validation runs only for EFA; other providers skip that step with a logged notice.
  * CXI device discovery now reports device counts in logs.

PR #1994, opened 2026-07-26
libfabric: preflight-check Neuron accelerator vs EFA NIC count

jimburtoft:
## What?

Adds a preflight check in `nixlLibfabricTopology::discoverTopology()` that emits an actionable log line at plugin init when the discovered Neuron accelerator count doesn't match the discovered EFA NIC count. The check distinguishes two failure modes:

1. **Wrong instance type** — e.g. `trn2.3xlarge` (4 Neuron devices, 1 host-level EFA, no per-Neuron-device EFA). Not supported for `FI_HMEM_NEURON` memory registration in hardware.
2. **Instance launched without EFA NICs** — e.g. default `aws ec2 run-instances` on `trn2.48xlarge` attaches only 1 ENA NIC, not the 16 EFA-typed NICs the plugin needs. User must attach EFA NICs explicitly via `--network-interfaces InterfaceType=efa` per `NetworkCardIndex`.

Also adds an "AWS Neuron (Trainium) requirements" section to `src/plugins/libfabric/README.md` with an instance-type table and an example `--network-interfaces` launch spec.

## Why?

The libfabric plugin's Neuron support (PR #1258) requires one EFA NIC per Neuron device to route `FI_HMEM_NEURON` memory registrations via peer-direct RDMA. Currently, when the Neuron/EFA topology doesn't satisfy this, both failure modes above surface as a mysterious `fi_mr_reg -EINVAL` deep inside `registerMem()`, which is hard to trace back to root cause without reading NIXL/EFA/libfabric source.

The Neuron support works cleanly on `trn2.48xlarge` with all 16 EFA NICs attached, but users on other Neuron instance types (or who forgot the launch-time EFA config) currently have no early warning that their configuration cannot support VRAM_SEG transfers.

This patch turns the mysterious deep-stack failure into a clear log line at plugin init time with actionable guidance ("wrong instance type" vs "instance launched without EFA NICs").

## How?

Adds ~48 lines in `nixlLibfabricTopology::discoverTopology()`, after `buildNicInfoMap()` populates `nic_info_map`. The check is gated on `num_aws_accel > 0 && num_nvidia_accel == 0` so it only fires on Neuron-only systems (no impact on CUDA or CUDA+Neuron hybrids).

Four cases:

- `efa_nic_count == 0` → `NIXL_WARN` recommending `--network-interfaces InterfaceType=efa` or noting unsupported instance types
- `0 < efa_nic_count < num_aws_accel` → `NIXL_WARN` distinguishing `trn2.3xlarge`-style (no per-device EFA in hardware) from partial-EFA-launch case
- `efa_nic_count > num_aws_accel && efa_nic_count % num_aws_accel != 0` → `NIXL_WARN` "not multiple, sub-optimal rail selection"
- Otherwise → `NIXL_INFO` one-line topology summary

Uses `NIXL_WARN` for mismatches (never `NIXL_ERROR` or `throw`, matching the existing sanity-check style for NVIDIA GPUs on P5en and for non-uniform NUMA / NIC bandwidth in the same file). No new API surface, no behavior change — mismatched configurations still fail at `fi_mr_reg` as before; this patch only adds a clearer diagnostic upstream of that failure.

## Testing

Tested on `trn2.48xlarge` (us-east-2) launched with 16 EFA NICs via `--network-interfaces InterfaceType=efa` for each `NetworkCardIndex` 0-15.

Built NIXL from source (`meson setup build -Denable_plugins=LIBFABRIC -Dlibfabric_path=/opt/amazon/efa` + `ninja -C build`) and ran the shipped `examples/cpp/nixl_example LIBFABRIC`, capturing NIXL logs at INFO level.

**Positive test (real 16 EFA / 16 Neuron):**

```
$ ls /sys/class/infiniband/ | wc -l   # 16
$ lspci | grep -c 'Elastic Fabric'    # 16
$ neuron-ls -j | jq length            # 16

I libfabric_topology.cpp:188] Discovered 16 EFA devices
I libfabric_topology.cpp:145] Neuron/EFA preflight: 16 Neuron device(s), 16 EFA NIC(s) (1 NIC(s) per Neuron device)
I libfabric_topology.cpp:308] Topology: 2 NUMA nodes, 16 NICs, 0 NVIDIA GPUs, 16 AWS accelerators
```

**Negative branches** — validated via a temporary test-only override honoring `NIXL_TEST_FORCE_AWS_ACCEL` env var, rebuilt, ran all cases, then removed the override before committing:

| Case | num_aws_accel (forced) | efa_nic_count | Expected branch | Result |
|------|------------------------|---------------|-----------------|--------|
| Positive (no override) | 16 | 16 | `NIXL_INFO` "1 NIC(s) per Neuron device" | ✅ PASS |
| Fewer EFA than Neuron | 32 | 16 | `NIXL_WARN` "trn2.3xlarge / launch config" | ✅ PASS |
| No Neurons (gated out) | 0 | 16 | correctly silent | ✅ PASS |
| Even multiple (2:1) | 8 | 16 | `NIXL_INFO` "2 NIC(s) per Neuron device" | ✅ PASS |
| Non-integer ratio | 5 | 16 | `NIXL_WARN` "not multiple, sub-optimal" | ✅ PASS |
| Much fewer EFA | 100 | 16 | `NIXL_WARN` "trn2.3xlarge / launch config" | ✅ PASS |

Cannot include automated CI tests: the topology-discovery path requires real hwloc + libfabric + EFA driver interaction that the existing `libfabric_topology_test.cpp` (a plain `main()` smoke test) doesn't mock. This matches the precedent from PR #1258, which merged the original Neuron support without a Neuron-specific gtest.

## Pre-submission checks

- [x] Code follows style guidelines (`.clang-format` applied; verified with `clang-format-19` matching CI)
- [x] Follows conventions in `docs/CodeStyle.md`
- [x] Existing tests pass (no test changes)
- [ ] New tests added for new functionality — see note above; no clean gtest possible without hwloc/libfabric mocking. Happy to add one if reviewers point at a suitable mocking pattern.
- [x] Documentation updated (`src/plugins/libfabric/README.md`)
- [x] PR template filled out completely
- [x] Commits are signed with DCO

## Summary by CodeRabbit

* **Documentation**
  * Added AWS Neuron (Trainium) requirements and troubleshooting guidance for Libfabric `FI_HMEM_NEURON`, including EFA NIC compatibility, launch configuration examples, and instance verification steps.
* **Bug Fixes**
  * Improved Neuron/EFA startup diagnostics with preflight warnings for missing, insufficient, or ratio-mismatched EFA NICs.
  * Added clearer warnings when Neuron is present but a non-EFA provider is selected (no `FI_HMEM_NEURON` memory registration).
  * Accelerator discovery now occurs earlier to improve reliability during provider fallback.
* **Tests**
  * Expanded Trainium topology unit tests with new hwloc fixtures for EFA and no-EFA scenarios and Neuron-specific validation.


oa-aws:
Starting review. A general question about testing. Is it possible to reuse the hwloc mocking capabilities in libfabric_topology_test.cpp? It requires an input topology file (by running "lstopo-no-graphics --of xml" on the target instance type), and it can mock any instance type. This required though adding test cases for Neuron in the main function. You can add your own topology array and ingest it, as it is done for NUMA-aware rail selection topology tests. The actual way to really test it, is by adding a wrapper script that runs the test with varying instance type parameter, and verifying stdout shows the expected info/warn message.

jimburtoft:
Thanks for the review pointer -- reusing the existing hwloc-mocking framework was straightforward once I found it. Pushed [`06a06679`](https://github.com/jimburtoft/nixl/commit/06a06679):

1. **Neuron topology test coverage**. Added `test/unit/utils/libfabric/topo/trn2.48xl-topo.xml` (captured on a real trn2.48xlarge in us-east-2 via `lstopo-no-graphics --whole-io --of xml` -- the `--whole-io` flag turned out to be required to expose the `1d0f:7364` Neuron PCI devices, which hwloc filters out by default). Added a `NeuronTopologyInfo` array and `testNeuronTopology()` / `testNeuronTopologies()` functions patterned on the existing `TopologyInfo` / `testNumaDramTopology()`. Wired into `main()`:
   - Default suite now runs `testNeuronTopologies()` in addition to the existing DRAM_SEG tests.
   - Added a `neuron:<instance>` CLI arg pattern so a wrapper script can invoke a single-topology run without exercising the DRAM_SEG rail-selection tests (which don't apply to Neuron-only topologies).

2. **Coverage.** The new test verifies `getNumAwsAccel() == 16`, `getTotalNicCount() == 16`, `getNumNvidiaAccel() == 0` when loading the trn2.48xl XML. The preflight fires as part of `nixlLibfabricTopology()` construction; log output is verifiable externally (e.g. `libfabric_topology_test neuron:trn2.48xl 2>&1 | grep "Neuron/EFA preflight"`) which matches your "wrapper script grepping stdout" suggestion.

3. **All existing tests still pass.** Ran the full default suite locally on the trn2.48xlarge -- basic topology, new Neuron topology, and NUMA-aware DRAM_SEG rail selection across all 8 existing GPU instance-type XMLs. Clang-format-19 clean; copyright check clean.

Two things I did NOT do that might warrant a follow-up:

- **Negative-case XML** (Neuron devices present but 0 or fewer-than-expected EFA NICs -- e.g. trn2.3xlarge, or trn2.48xlarge launched without the 16 EFA NICs). Would need to capture the XML on those hardware configs, which is straightforward but I skipped for scope. Happy to add if you'd like coverage of the WARN branches specifically.
- **Wrapper script that greps stdout for expected messages.** The current test asserts the counts; the WARN/INFO message contents are still verified by hand. Could add a bash wrapper that runs the binary and greps stderr, though this shifts test infrastructure. Let me know if you'd like this in this PR or as a follow-up.

src/plugins/libfabric/README.md
```diff
@@ -35,6 +35,45 @@ Validated compatibility with:
...
+    {"DeviceIndex": 0, "NetworkCardIndex": 0, "InterfaceType": "efa",
+     "SubnetId": "...", "Groups": ["..."]},
+    {"DeviceIndex": 1, "NetworkCardIndex": 1, "InterfaceType": "efa",
+     "SubnetId": "...", "Groups": ["..."]},
+    ...
+    {"DeviceIndex": 1, "NetworkCardIndex": 15, "InterfaceType": "efa",
+     "SubnetId": "...", "Groups": ["..."]}
+  ]
+}
+```
```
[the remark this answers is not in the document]
jimburtoft: Fixed in [`06a06679`](https://github.com/jimburtoft/nixl/commit/06a06679). Restructured the README example so it is no longer labelled as valid `--cli-input-json` while containing an ellipsis: showed a per-entry template with the shape, then a short Python helper that generates the full 16-entry `NetworkInterfaces` array programmatically. No ellipsis, no invalid JSON.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -100,6 +100,54 @@ nixlLibfabricTopology::discoverTopology() {
...
+        // Preflight: warn on Neuron-accelerator vs EFA-NIC count mismatches. AWS Trainium
+        // instances that support DI need one EFA NIC per Neuron device (e.g. 16
+        // per-device EFA NICs on trn2.48xlarge). Some smaller Neuron instance types (e.g.
+        // trn2.3xlarge) expose only a single host-level EFA NIC and are not sufficient for
+        // FI_HMEM_NEURON memory registration on all Neuron devices. Also, EFA-typed NICs must
+        // be attached explicitly at instance launch time via
+        // `--network-interfaces InterfaceType=efa`; the AWS default only attaches an ENA NIC.
+        // Warn here so users see a clear diagnostic at plugin init rather than a mysterious
+        // fi_mr_reg -EINVAL later during memory registration.
+        if (num_aws_accel > 0 && num_nvidia_accel == 0) {
```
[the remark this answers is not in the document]
jimburtoft: Fixed in [`06a06679`](https://github.com/jimburtoft/nixl/commit/06a06679). Restructured `discoverTopology()` to call `discoverAccelWithHwloc()` **early** (before `discoverProviderWithDevices()`), so `num_aws_accel` is populated for both provider paths. Added a mirror preflight WARN in the TCP/sockets fallback branch, before `num_aws_accel` is zeroed at line ~200. The message tailored for the TCP case ("libfabric provider selected `tcp`, no EFA") is distinct from the EFA-branch WARNs so users can tell which path triggered.

`discoverAccelWithHwloc()` is now called twice on the EFA path (once early, once inside `discoverHwlocTopology()`), but it's idempotent (resets counters at the top) so this is safe -- just one extra hwloc PCI walk. If you prefer eliminating the duplicate call I can gate the inner call on `num_aws_accel == 0` or similar, but figured being minimally invasive was preferable.

oa-aws:
I think negative XML test is important here. It would best if we can capture lstopo for all instances/cases added in code. The bash wrapper can be done separately. Also I think topo files should be sanitized (see Code Rabbit comment, and compare to existing topo files). Thanks.

jimburtoft:
Two follow-up test/analysis items I wanted to flag proactively:

**GPU (NVIDIA-only) instances -- code review + integrated test pass:** The preflight is gated on `num_aws_accel > 0 && num_nvidia_accel == 0`, so it stays silent on any pure NVIDIA host. On the full test suite run I did before pushing 06a06679, all 8 existing GPU topology XMLs (p3dn/p4d/p5/p5en/p6-b200/g5/g6/c5n) continued to pass with no new output from our patch. Confirmed by tracing:
- On p5.48xl: `discoverAccelWithHwloc()` produces `num_aws_accel=8` AND `num_nvidia_accel=8` (the counter increments for both vendor types -- likely a pre-existing naming quirk, not something my patch introduced). The gate `num_nvidia_accel == 0` is false, preflight is skipped.
- The rest of `discoverTopology()` proceeds to `buildAccelToEfaMapping()` for NVIDIA as before.

**Trn3 -- code review only (no hardware access in any region I have):** Trn3 isn't offered in the us-east/us-west/sa-east/ap-southeast-4 regions I can launch in, so I can't capture a Trn3 lstopo XML the way I did for trn2.48xl. Walking through the logic against Trn3's likely topology:
- `isNeuronAccel()` already covers TRN3 device IDs (`0x7564`, `0x7565`, per PR #1258).
- Assuming Trn3 exposes N Neuron devices through hwloc paired with N EFA NICs at launch, the preflight fires the `NIXL_INFO` "1 NIC(s) per Neuron device" summary.
- Assuming Trn3.48xlarge default `run-instances` also attaches only 1 ENA and requires explicit `InterfaceType=efa` per NIC (same as trn2.48xlarge), our WARN would fire with a message that already references "On Trn3 attach the count matching the instance size" -- actionable.

Two concerns for Trn3 specifically that a Trn3-having reviewer could confirm:

1. **`0x7564` vs `0x7565` PCI enumeration.** If Trn3 exposes each Neuron chip as **two** PCI functions (one at `0x7564` and one at `0x7565`), `discoverAccelWithHwloc()` would count `num_aws_accel = 2*N` while `efa_nic_count = N`, and our WARN would misleadingly say "fewer EFA than Neuron". If Trn3 exposes one PCI function per chip (either `0x7564` OR `0x7565`), no false positive. I don't have visibility into which. Would appreciate a check on real Trn3 hardware, or a comment from whoever added those device IDs on the intended semantics.

2. **Trn3-specific WARN wording.** The "fewer EFA than Neuron" WARN currently names trn2.3xlarge as the example of "smaller instance without per-device EFA". If there's a "trn3.3xlarge"-style small instance with a similar limitation, the message would still be informative but not maximally specific. Easy to update if Trn3 has such an instance -- just needs the name.

If either of those is a concern for the PR, happy to update the wording or the counter logic. Otherwise, I'd propose merging as-is and following up with Trn3-specific test data in a subsequent PR once someone with Trn3 access can capture the lstopo XML.

jimburtoft:
Pushed [`a55d20c`](https://github.com/jimburtoft/nixl/commit/a55d20c) with the remaining review items addressed:

**1. Sanitize the trn2.48xl-topo.xml fixture** ([oa-aws](https://github.com/ai-dynamo/nixl/pull/1994#issuecomment-5093365279) + [coderabbitai](https://github.com/ai-dynamo/nixl/pull/1994#discussion_r3658028893)). Removed all non-topology metadata to match the conventions of the existing `p5.48xl-topo.xml` and `p5en.48xl-topo.xml` sanitized fixtures. Specifically stripped:
- `HostName` (contained private IP)
- 16 `Address` fields (MAC addresses -- one per EFA NIC)
- 5 storage `SerialNumber` fields
- 16 `NodeGUID` fields
- `DMIBoardAssetTag`, `DMIProductName`, `DMISysVendor`, etc.
- `OSName`, `OSRelease`, `ProcessName`, `LinuxCgroup`

Verified: no EC2 instance IDs, no AWS account IDs, no volume IDs, no AMI/subnet/VPC/SG/ENI IDs, no ARNs, no private IPs, no MAC addresses, no serial numbers, no internal hostnames. The only 12-digit numbers remaining are legitimate NUMA `local_memory` byte counts.

**2. Negative-case XMLs** ([oa-aws](https://github.com/ai-dynamo/nixl/pull/1994#issuecomment-5093365279)):
- **`trn2.48xl-no-efa-topo.xml`**: captured on a trn2.48xlarge launched with the AWS default `run-instances` invocation (single ENA NIC, no `--network-interfaces InterfaceType=efa`). Reveals 16 Neuron devices, 0 EFA NICs via hwloc. Exercises the "0 EFA NIC(s)" WARN branch of the preflight for the "instance launched wrong" case.
- **`trn2.3xl-topo.xml`**: captured on a trn2.3xlarge (also default-launched, no EFA NIC attached). Reveals 1 Neuron device, 0 EFA NICs via hwloc. Same WARN branch but with a single Neuron device, closer to the "smaller instance without per-device EFA" wording.
- Both sanitized to the same standard as trn2.48xl-topo.xml.

**3. Test coverage for the 0-EFA case:**

The plugin's `nixlLibfabricTopology` constructor **throws** when no libfabric provider can be selected (which is what happens when hwloc discovers 0 EFA NICs). This is a fatal end-state -- the preflight WARN would fire upstream of it in production, before construction throws.

Extended `testNeuronTopology()` to expect this throw when `expected_nic_count == 0`, and to fail if the plugin silently accepts an unusable topology. Both new negative-case topologies now covered:

```
1. Testing Neuron topology for instance type trn2.48xl
   SUCCESS: Neuron topology trn2.48xl loaded (16 Neuron device(s), 16 EFA NIC(s))
2. Testing Neuron topology for instance type trn2.48xl-no-efa
   SUCCESS: Neuron topology trn2.48xl-no-efa correctly threw at topology-discovery
            time (0 EFA NICs); in production the preflight WARN would fire before
            the throw.
3. Testing Neuron topology for instance type trn2.3xl
   SUCCESS: Neuron topology trn2.3xl correctly threw at topology-discovery time
            (0 EFA NICs); in production the preflight WARN would fire before the throw.
```

All 8 existing GPU DRAM_SEG rail-selection tests still pass. Clang-format-19 clean; copyright check clean; hwloc parses all 3 XMLs without warnings.

**Follow-up not in this PR:**
- Bash wrapper that runs the binary once per instance-type XML and greps stderr for the specific preflight message contents. oa-aws mentioned this could be a separate change ("The bash wrapper can be done separately"). Happy to open a follow-up PR for that whenever you want it.
- Positive-case negative-XML: a "trn2.3xlarge WITH `InterfaceType=efa` attached" fixture (would trigger the INFO summary path on a small instance). Skipped because the capacity block slot was exhausted; not strictly needed for coverage since the 16 EFA / 16 Neuron case already exercises that branch.

test/unit/utils/libfabric/topo/trn2.48xl-topo.xml
```diff
@@ -0,0 +1,2630 @@
...
+    <info name="DMIBIOSVendor" value="Amazon EC2"/>
+    <info name="DMIBIOSVersion" value="1.0"/>
+    <info name="DMIBIOSDate" value="10/16/2017"/>
+    <info name="DMISysVendor" value="Amazon EC2"/>
+    <info name="Backend" value="Linux"/>
+    <info name="LinuxCgroup" value="/user.slice"/>
+    <info name="OSName" value="Linux"/>
+    <info name="OSRelease" value="6.17.0-1019-aws"/>
+    <info name="OSVersion" value="#19~24.04.1-Ubuntu SMP Tue Jun 23 18:53:06 UTC 2026"/>
+    <info name="HostName" value="ip-172-31-32-8"/>
```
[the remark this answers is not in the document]
jimburtoft: Fixed in [`a55d20c`](https://github.com/jimburtoft/nixl/commit/a55d20c). Sanitized to match the conventions of the existing p5.48xl-topo.xml / p5en.48xl-topo.xml fixtures. Removed HostName (private IP), 16 MAC addresses, 5 storage SerialNumbers, DMI board/chassis/product tags, OS name/release/hostname, NodeGUIDs, and process info. Kept only PCI-topology-relevant `info` fields.

Verified: no EC2 instance IDs, no AWS account IDs, no volume/AMI/subnet/VPC/SG/ENI IDs, no ARNs, no private IPs, no MAC addresses, no serial numbers. The only 12-digit numbers remaining are legitimate NUMA `local_memory` byte counts.

test/unit/utils/libfabric/libfabric_topology_test.cpp
```diff
@@ -787,6 +863,147 @@ testNumaDramTopology(const TopologyInfo &topology_info) {
...
+            NIXL_ERROR << "Expected " << topology_info.instance_type
+                       << " with 0 EFA NICs to throw at topology-discovery time, but no "
+                          "exception was raised; the plugin silently accepted an unusable "
+                          "topology, which contradicts the preflight design.";
+            rc = 1;
+        } else {
+            NIXL_INFO << "   SUCCESS: Neuron topology " << topology_info.instance_type
+                      << " correctly threw at topology-discovery time (0 EFA NICs); "
+                         "in production the preflight WARN would fire before the throw.";
+        }
```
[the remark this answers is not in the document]
jimburtoft: Fixed in [`13cace3`](https://github.com/jimburtoft/nixl/commit/13cace3). Documented that `expected_num_aws_accel` is not verified for zero-NIC rows (the constructor throws before accelerator-count getters can be called). Added a comment to `testNeuronTopology`'s docstring noting that the field is retained purely as documentation of the underlying hardware topology in those rows. Kept the field populated with real values (16 for `trn2.48xl-no-efa`, 1 for `trn2.3xl`) rather than removing it, so a future refactor that added a standalone `discoverAccelWithHwloc()`-equivalent code path could assert against those numbers.

test/unit/utils/libfabric/libfabric_topology_test.cpp
```diff
@@ -787,6 +863,147 @@ testNumaDramTopology(const TopologyInfo &topology_info) {
...
+
+        if (rc == 0) {
+            NIXL_INFO << "   SUCCESS: Neuron topology " << topology_info.instance_type
+                      << " loaded (" << topology.getNumAwsAccel() << " Neuron device(s), "
+                      << topology.getTotalNicCount() << " EFA NIC(s))";
+        }
+    }
+
+    clearTesting();
+    unsetenv("HWLOC_XMLFILE");
```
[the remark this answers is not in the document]
jimburtoft: Fixed in [`13cace3`](https://github.com/jimburtoft/nixl/commit/13cace3). Adopted your `TestEnvGuard` RAII scope-guard suggestion and wrapped the positive-branch constructor in `try`/`catch` via `std::optional<nixlLibfabricTopology>` to defer construction. Now:

- `clearTesting()` and `unsetenv("HWLOC_XMLFILE")` run on every exit path, including if the positive branch throws unexpectedly (via the guard's destructor).
- Unexpected throws in the positive branch are caught and reported as a test failure (`rc=4`) rather than propagating out and aborting the binary.
- Verified on a trn2.48xlarge that all 3 Neuron topologies pass AND all 8 subsequent GPU DRAM_SEG tests still pass (i.e., no env-var contamination).

Also included the two committable-suggestion changes verbatim except for two minor consolidations: moved the guard right after the `rc` declaration (so it's active from the earliest possible point after `setTesting()` / `setenv()`), and kept the outer `rc` declared once at the top of the function rather than duplicating it inside both branches.

jimburtoft:
Pushed [`b0c6479`](https://github.com/jimburtoft/nixl/commit/b0c6479) addressing the last outside-diff CodeRabbit finding: dropped `static` from the `dummy` `TopologyInfo` in `testNeuronTopology()` so it's freshly constructed per call. As you noted, `fi_getinfo` only reads `nic_line_speed` today, but the stale `instance_type` / `topo_file` / `nic_count` fields were a real correctness landmine.

Extended the `TestEnvGuard` destructor to reset `curr_topology = nullptr` alongside `clearTesting()` and `unsetenv("HWLOC_XMLFILE")`. C++ destructor order runs the guard before the stack-local `dummy` goes out of scope, so `curr_topology` never dangles.

Verified locally on trn2.48xlarge: full suite passes -- basic topology, all 3 Neuron topologies, all 8 GPU DRAM_SEG topologies. Clang-format-19 clean.

If there's anything else, let me know. Otherwise the PR is ready for another look from @oa-aws.

jimburtoft:
Trimmed the XML fixtures in [`82ef396`](https://github.com/jimburtoft/nixl/commit/82ef396) to match the sanitization level of the existing `p5.48xl-topo.xml` / `p5en.48xl-topo.xml` reference fixtures, which strip all CPU-hierarchy elements (`Core`, `PU`, `L1Cache`, `L2Cache`, `L3Cache`) and retain only the PCIe topology + NUMA node structure that the NIXL topology discovery actually reads.

Size delta:

| Fixture | Before | After | Change |
|---------|--------|-------|--------|
| `trn2.48xl-topo.xml` | 2491 lines | 1237 lines | **-50%** |
| `trn2.48xl-no-efa-topo.xml` | 2275 lines | 1021 lines | **-55%** |
| `trn2.3xl-topo.xml` | 373 lines | 292 lines | -22% |

Confirmed on a trn2.48xlarge that the full test suite continues to pass with the trimmed fixtures (basic topology, all 3 Neuron topologies, all 8 GPU DRAM_SEG rail-selection tests).

**On the `size/XXL` label**: the label comes from an external `pull-request-size` GitHub App with a `XXL >= 1000` default threshold. After trimming, the PR is 2929 additions total, which is still above 1000 -- so `size/XXL` is expected to stick. The XML fixtures are inherently the bulk of the diff; the actual code changes are:

- `src/utils/libfabric/libfabric_topology.cpp`: +78 lines (preflight logic)
- `src/plugins/libfabric/README.md`: +46 lines (AWS Neuron docs)
- `test/unit/utils/libfabric/libfabric_topology_test.cpp`: +255 lines (Neuron test harness + 3 test cases)

For NIXL's `check-pr-size` CI gate (which counts only modified files, excluding new fixture files), we're at 438 lines vs the 500-line threshold, so that gate passes.

I re-read `CONTRIBUTING.md` and don't see any additional requirements triggered by size beyond what the CI gate already enforces. Happy to split into smaller PRs (e.g., a fixture-only PR followed by a code+test PR) if you'd prefer that for reviewability, but each of the three fixture files is used exclusively by the tests added in this PR, so I think it makes sense to keep them together for review.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -72,6 +72,18 @@ nixlLibfabricTopology::discoverTopology() {
         return status;
     }
 
+    // Discover accelerators from hwloc early, before provider selection.
+    // This lets the Neuron/EFA preflight (below) fire even when the libfabric
+    // provider falls back to tcp/sockets (see PR discussion on ai-dynamo/nixl#1994).
+    // If EFA discovery runs later, `discoverAccelWithHwloc()` is called again
+    // from within `discoverHwlocTopology()`; both calls are idempotent, as they
+    // reset counters at the start.
+    status = discoverAccelWithHwloc();
```
oa-aws, line 81: I think it is ok moving the call to here, but it is unnecessary to make the call twice. You can remove the next call in discoverHwlocTopology(), but add a comment there that the call to discoverAccelWithHwloc() was already made earlier.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -100,6 +112,54 @@ nixlLibfabricTopology::discoverTopology() {
         // build nic info map regardless of accelerator to EFA mapping
         buildNicInfoMap();
 
+        // Preflight: warn on Neuron-accelerator vs EFA-NIC count mismatches. AWS Trainium
```
oa-aws: Let's put this if-statement (with preceding comment) in its own function, to avoid too large function and increase readability.

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -113,6 +173,24 @@ nixlLibfabricTopology::discoverTopology() {
         NIXL_INFO << "Using simplified topology for " << provider_name
                   << " devices (no topology mapping needed)";
 
+        // Preflight for TCP fallback with Neuron accelerators present: this typically
```
oa-aws: You can put this as well in a separate function for TCP preflight check.

oa-aws (dismissed):

src/utils/libfabric/libfabric_topology.cpp
jimburtoft: Done in [`77341fe`](https://github.com/jimburtoft/nixl/commit/77341fe). Removed the `discoverAccelWithHwloc()` call from `discoverHwlocTopology()` and replaced it with a comment noting that the earlier call from `discoverTopology()` has already populated `num_aws_accel` / `num_nvidia_accel`. Also updated the comment on the early call site to explain the ordering (early call is required for the preflight in both the EFA and TCP-fallback paths).

src/utils/libfabric/libfabric_topology.cpp
jimburtoft: Done in [`77341fe`](https://github.com/jimburtoft/nixl/commit/77341fe). Extracted into a new private method `neuronEfaPreflightForEfaProvider()`. `discoverTopology()` now just calls it after `buildNicInfoMap()`, so the EFA branch of `discoverTopology()` is a linear ~15-line sequence instead of a ~50-line inline chain. Body is unchanged.

src/utils/libfabric/libfabric_topology.cpp
jimburtoft: Done in [`77341fe`](https://github.com/jimburtoft/nixl/commit/77341fe). Extracted the TCP-fallback preflight into `neuronEfaPreflightForNonEfaProvider()`. Method name intentionally mirrors `neuronEfaPreflightForEfaProvider()` (only difference is the provider path). Called from the TCP branch of `discoverTopology()` before `num_aws_accel` is cleared.

Net effect on `discoverTopology()`: dropped from ~140 lines to ~75 lines, reads at a single level of abstraction. Both helpers are private, `void`-returning, and scoped to preflight logic.

jimburtoft:
Force-pushed [`536c61b`](https://github.com/jimburtoft/nixl/commit/536c61b) to amend the extraction commit with a clang-format-19 reflow. The refactor moved the `NIXL_WARN` statements into new helper methods, which changed the indentation context enough that clang-format-19 wanted them re-broken across lines. Amended rather than added a follow-up commit since it's cosmetic and belongs with the extraction.

All 9 CI checks green:
- clang-format ✅
- copyright-checks ✅
- check-pr-size ✅
- pre-commit ✅
- python-checks (all 4) ✅
- CodeRabbit ✅

Mergeable, blocked only on maintainer approval.

jimburtoft:
Merged `origin/main` into the branch in [`189ba3a`](https://github.com/jimburtoft/nixl/commit/189ba3a) to resolve the conflict introduced by the CXI provider work (#1416) and the RDMA-device rename in libfabric_topology. Two conflict spots, both trivial:

1. In `discoverTopology()`, the EFA branch condition became `provider_name == "efa" || provider_name == "cxi"`. Kept our `neuronEfaPreflightForEfaProvider()` call in place. Since the preflight is already gated on `num_aws_accel > 0 && num_nvidia_accel == 0` AND Neuron devices don't coexist with CXI (Slingshot) interconnects, the preflight is effectively a no-op on CXI hosts. Also added an explicit `provider_name != "efa"` early-return inside the helper for defence-in-depth, and updated the docstring to note the CXI case.

2. In `discoverHwlocTopology()`, `discoverEfaDevicesWithHwloc()` was renamed to `discoverRDMADevicesWithHwloc()` on main. Took the rename and preserved my "discoverAccelWithHwloc() is intentionally NOT called here" comment.

Verified locally on trn2.48xlarge: full test suite passes (basic topology, all 3 Neuron topologies, all 8 GPU DRAM_SEG topologies). Clang-format-19 clean. All 9 CI checks green.

The 8-commit branch history is: 7 review-response commits + 1 merge from main. NIXL uses squash-merge, so this all collapses into a single commit on `main` at merge time. Let me know if you'd like me to squash locally into fewer commits before that (e.g., squash all commits into one) or if the current shape is fine.

jimburtoft:
Thanks for the trn1.32xlarge flag — you're right, that's a supported topology (16 Neuron / 8 EFA by design, 2 Neurons sharing each EFA) and our WARN would have misleadingly told users to switch to trn2.48xlarge.

Pushed [`8a053ca`](https://github.com/jimburtoft/nixl/commit/8a053ca):

1. **Rewrote the WARN body** to enumerate the three cases users can actually be in:
   - (1) trn1.32xlarge — 16 Neuron / 8 EFA is expected, plugin works, ignore the warning
   - (2) trn2.3xlarge (or a similar smaller Trn2 with only host-level EFA) — unsupported, use trn2.48xlarge
   - (3) trn2.48xlarge / Trn3 launched without all EFA NICs — misconfiguration, verify `--network-interfaces InterfaceType=efa` per NetworkCardIndex

2. **Added two new topology fixtures** captured on real hardware:
   - `trn1.32xl-topo.xml` — 16 Neuron / 8 EFA (the scenario you flagged)
   - `trn1.2xl-topo.xml` — 1 Neuron / 0 EFA (smallest Trn1, which does not support EFA per `describe-instance-types`)

3. **Verified the test suite runs on generic hardware** — built libfabric 2.7 from `ofiwg/libfabric` main on a plain c5.4xlarge (no Neuron / no EFA), ran the full suite: all 5 Neuron topologies (trn2.48xl, trn2.48xl-no-efa, trn2.3xl, trn1.32xl, trn1.2xl) and all 8 GPU DRAM_SEG topologies pass. The test infrastructure is fully portable and doesn't require Neuron hardware.

Left `NIXL_WARN` as the severity for the "fewer EFA than Neuron" case rather than `NIXL_INFO`. Rationale: trn2.3xlarge and partial-launch trn2.48xlarge are real misconfigurations that need attention, and downgrading the message would silence them. The trn1.32xlarge callout is prominent (case #1 in the message), so users on that instance can quickly identify and dismiss.

Test output on trn1.32xl fixture:
```
[TIMING] Testing Neuron topology for instance type trn1.32xl
[WARN]   Discovered 16 Neuron accelerator(s) but only 8 EFA NIC(s) with usable PCIe topology.
         Three cases: (1) trn1.32xlarge has 16 Neuron devices and 8 EFA NICs by design
         (2 Neurons share each EFA); if this is trn1.32xlarge, the plugin works and this
         warning can be ignored. (2) trn2.3xlarge...
[INFO]   SUCCESS: Neuron topology trn1.32xl loaded (16 Neuron device(s), 8 EFA NIC(s))
```

oa-aws:
Hi Jim, after some delay, I am getting back to this. Will try to have another person join the review.

rongbingzhou (dismissed):

src/plugins/libfabric/README.md
```diff
@@ -35,6 +35,52 @@ Validated compatibility with:
...
+### AWS Neuron (Trainium) requirements
+
+For `FI_HMEM_NEURON` (VRAM_SEG registration of Neuron device memory) the plugin requires **one EFA network interface per Neuron device** attached to the instance at launch time, so that each Neuron device has a paired EFA NIC for peer-direct RDMA.
+
+Not all Neuron instance sizes meet this requirement. In particular:
+
+| Instance type | Neuron devices | EFA interfaces | `FI_HMEM_NEURON` supported |
+|--|--|--|--|
+| `trn2.48xlarge` | 16 | 16 (one per Neuron device) | Yes, when launched with all 16 EFA NICs attached |
+| `trn2.3xlarge` | 4 | 1 (host-level EFA only) | **No** -- single host EFA cannot be routed to individual Neuron devices |
```
rongbingzhou: According to public doc (https://aws.amazon.com/ec2/instance-types/trn2/), `trn2.3xlarge` only has 1 Neuron device instead of 4. Could you double check?

jimburtoft: Good catch, thank you -- you're right. trn2.3xlarge has 1 Trainium2 chip (per AWS docs), not 4. The chip presents 4 logical cores at LNC=2 (the default), which is what I originally conflated with "device count" when drafting the table.

Fixed in [`6758b07`](https://github.com/jimburtoft/nixl/commit/6758b07). Only the README table needed the change:

```diff
- | `trn2.3xlarge` | 4 | 1 (host-level EFA only) | ...
+ | `trn2.3xlarge` | 1 | 1 (host-level EFA only) | ...
```

The test fixture (`trn2.3xl-topo.xml`) already reflects the correct count -- captured on a real trn2.3xlarge, it has exactly 1 Neuron PCI device (`1d0f:7364`), and the `NeuronTopologyInfo` entry correctly declares `expected_num_aws_accel = 1`. The source-code WARN message doesn't state a device count for trn2.3xlarge (it just describes it as "smaller Trn2 with only a single host-level EFA and no per-device EFA"), so no source change was needed.

jimburtoft:
Thanks @oa-aws, and thanks @rongbingzhou for the review and approval. Fixed the trn2.3xlarge device-count typo in [`6758b07`](https://github.com/jimburtoft/nixl/commit/6758b07) (was 4, should have been 1 -- the confusion was between the single Trainium2 chip and its 4 logical cores at LNC=2, the default). Ready for another look whenever convenient.

jimburtoft:
While fixing the `trn2.3xlarge` device count, @jburtoft asked whether the preflight would behave differently under `NEURON_LOGICAL_NC_CONFIG=1` (LNC=1), since a Trainium2 chip presents 8 logical cores at LNC=1 versus 4 at LNC=2 (the default). Verified empirically on a real trn2.3xlarge (ap-southeast-4c) -- **the preflight is LNC-invariant**, because it counts PCI devices rather than cores.

**1. `neuron-ls` core count changes, device count does not:**

| `NEURON_LOGICAL_NC_CONFIG` | NEURON DEVICE | NEURON CORES | CORE IDS | PCI BDF |
|--|--|--|--|--|
| `1` | 0 | **8** | 0-7 | `0000:33:00.0` |
| `2` (default) | 0 | **4** | 0-3 | `0000:33:00.0` |

Same single device, same BDF; only the logical-core partitioning differs.

**2. hwloc PCI enumeration is byte-identical:**

```
LNC=1: Neuron PCI devices = 1, EFA PCI devices = 0
LNC=2: Neuron PCI devices = 1, EFA PCI devices = 0

$ diff <(NEURON_LOGICAL_NC_CONFIG=1 lstopo-no-graphics --whole-io --of xml) \
       <(NEURON_LOGICAL_NC_CONFIG=2 lstopo-no-graphics --whole-io --of xml)
(no output -- byte-for-byte identical)
```

This is expected: hwloc reads `/sys/bus/pci/devices/`, populated by the kernel from PCI config space at boot. `NEURON_LOGICAL_NC_CONFIG` is read by `libnrt` at `nrt_init()` and only affects how the Neuron runtime partitions cores *within* each chip -- it never re-enumerates PCI.

**3. The preflight itself produces identical output under both settings.** Built this branch on the trn2.3xlarge and ran the topology test with the trn2.48xl fixture under both LNC values:

```
NEURON_LOGICAL_NC_CONFIG=1 → Neuron/EFA preflight: 16 Neuron device(s), 16 EFA NIC(s) (1 NIC(s) per Neuron device)
NEURON_LOGICAL_NC_CONFIG=2 → Neuron/EFA preflight: 16 Neuron device(s), 16 EFA NIC(s) (1 NIC(s) per Neuron device)
```

Full suite under both: 148 SUCCESS, 0 ERROR/FAIL, exit 0.

So `num_aws_accel` is always the physical Neuron **chip** count, and the ratio the preflight compares against the EFA NIC count is unaffected by LNC mode. No LNC-specific handling is needed in this patch.

rongbingzhou (approved):

oa-aws (dismissed):

src/utils/libfabric/libfabric_topology.cpp
```diff
@@ -100,6 +112,8 @@ nixlLibfabricTopology::discoverTopology() {
         // build nic info map regardless of accelerator to EFA / CXI mapping
         buildNicInfoMap();
 
+        neuronEfaPreflightForEfaProvider();
```
oa-aws, line 115: Is this correct also when provider is not efa (i.e. "cxi")?

jimburtoft: Yes -- it's a no-op on CXI. The helper early-returns when the provider isn't EFA:

```cpp
void
nixlLibfabricTopology::neuronEfaPreflightForEfaProvider() {
    if (provider_name != "efa") {
        return;
    }
    if (!(num_aws_accel > 0 && num_nvidia_accel == 0)) {
        return;
    }
    ...
```

I added that guard when resolving the merge conflict with #1416 (the CXI provider work), since that PR widened this branch from `provider_name == "efa"` to `provider_name == "efa" || provider_name == "cxi"`. The docstring notes it too:

> Called from the EFA / CXI branch of discoverTopology(); returns without any log message on CXI hosts, since Neuron devices don't coexist with CXI (HPE Slingshot) interconnects.

So on a CXI host the call returns before reading `nic_info_map` or emitting anything.

Worth noting the guard is belt-and-suspenders rather than strictly required: the second condition (`num_aws_accel > 0 && num_nvidia_accel == 0`) would also short-circuit on any realistic CXI system, because `num_aws_accel` counts Neuron PCI devices (vendor `0x1d0f` + the Trainium/Inferentia device IDs) and those don't appear on Slingshot hardware. I kept the explicit provider check anyway so the intent is obvious at the call site and it can't silently start firing if the branch condition widens again.

One thing I should flag for symmetry, since you're looking at this: `neuronEfaPreflightForNonEfaProvider()` has only the accelerator-count guard, not a provider check. That's deliberate -- it's called from the `else` branch, so by construction the provider is neither efa nor cxi, and its message interpolates `provider_name` to report whatever was actually selected. But if you'd prefer both helpers guard explicitly for consistency, that's a one-line change and I'm happy to make it.

oa-aws (approved):

PR #2004, opened 2026-07-29
EP/TESTS: Skip Kineto reporting when CUDA profiling is unavailable

amastbaum:
## What?
Skip optional Kineto reporting when CUDA profiling data is unavailable.

## Why?
CUPTI 12 is incompatible with gb300 and fails to initialize, leaving Kineto without CUDA timing. NIXL previously reported this as a misleading kernel-dispatch assertion.

## How?
Detect missing CUDA profiling data, print a clear skip message, and preserve kernel validation when profiling is available.

## Summary by CodeRabbit

- **Tests**
  - Improved profiling test reliability by detecting when Kineto/CUDA timing activities are unavailable and skipping Kineto-based benchmarking instead of failing.
  - Synchronized device-support checks across distributed test processes so unsupported environments are handled consistently.
  - Enhanced diagnostics for kernel-name mismatches by including the full profiling table in error output.
  - Simplified elastic profiling execution to gracefully handle unsupported or unavailable Kineto scenarios.

### DIFF
--- a/examples/device/ep/tests/utils.py
+++ b/examples/device/ep/tests/utils.py
@@ -31,6 +31,10 @@ import torch
 import torch.distributed as dist
 
 
+class KinetoUnavailableError(RuntimeError):
+    """Raised when the profiler records no CUDA timing activities."""
+
+
 def init_dist(local_rank: int, num_local_ranks: int):
     # NOTES: you may rewrite this function with your own cluster settings
     ip = os.getenv("MASTER_ADDR", "127.0.0.1")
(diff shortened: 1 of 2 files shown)


examples/device/ep/tests/utils.py
```diff
@@ -221,17 +225,23 @@ def bench_kineto(
...
     is_tuple = isinstance(kernel_names, tuple)
-    prof_lines = (
-        prof.key_averages()
-        .table(sort_by="cuda_time_total", max_name_column_width=100)
-        .split("\n")
+    prof_table = prof.key_averages().table(
+        sort_by="cuda_time_total", max_name_column_width=100
     )
+
+    if "CUDA" not in prof_table:
```
itayalroy: This detection logic is pretty hacky, if kineto records no CUDA timings for some other reason (which might suggest a NIXL EP bug) we might accidentally ignore it. Looks like CUPTI added support for Blackwell in CUDA 12.8 and Blackwell Ultra in CUDA 12.9, can we skip kineto tests with a warning when `--kineto` is used on one of these systems with an older CUPTI version?

ofirfarjun7: @itayalroy  are you sure they added support for GB200/GB300?
[here](https://developer.nvidia.com/cupti-ctk12_8) I can see only B100/GB10x
Regarding Blackwell support, for cu12.8 it is mentioned that "It is supported on Linux x86_64 and Windows x86_64 platforms only."
AFAIR lyris is arm based arch

We will try to add proper detection based on Cupti API, not on versioning.

itayalroy: Main issue is that we don't use CUPTI directly, we use `torch.profiler` which internally uses CUPTI, and torch does not expose any API to query for device support for `torch.profiler`. CUPTI itself has a CPP API to query for device support (unfortunately no equivalent API in their Python API), we can probably use that instead of versioning.

amastbaum: I am working on a fix that calls CUPTI CPP API directly and checks if `cuptiDeviceSupported`. Will upload soon.

ofirfarjun7:
IMO using ctypes adds a lot of code for a very simple thing and it is kind of hacky.
Maybe we can use cupti-python?

PR #2006, opened 2026-07-29
EP/TESTS: Skip Kineto reporting when CUDA profiling is unavailable - 1.4.0

amastbaum:
## What?
Skip optional Kineto reporting when CUDA profiling data is unavailable.

## Why?
CUPTI 12 is incompatible with gb300 and fails to initialize, leaving Kineto without CUDA timing. NIXL previously reported this as a misleading kernel-dispatch assertion.

## How?
Detect missing CUDA profiling data, print a clear skip message, and preserve kernel validation when profiling is available.

## Summary by CodeRabbit

* **New Features**
  * Added build verification for container images across supported architectures and targets.
  * Improved wheel builds with cached base images for faster CI execution.
  * Added optional support for Infinia libraries and the UCX SPcX plugin during builds.

* **Bug Fixes**
  * Improved profiling diagnostics when CUDA timing data is unavailable.

* **Documentation**
  * Updated CI documentation to describe the expanded validation workflow and image caching.

* **Chores**
  * Updated the project version to 1.4.0 and refreshed attribution metadata.

### DIFF
--- a/examples/device/ep/tests/elastic/elastic.py
+++ b/examples/device/ep/tests/elastic/elastic.py
@@ -39,6 +39,7 @@ from plan import Plan
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 from utils import (  # noqa: E402
+    KinetoUnavailableError,
     bench,
     bench_kineto,
     calc_diff,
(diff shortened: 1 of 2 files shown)


examples/device/ep/tests/utils.py
```diff
@@ -221,17 +225,23 @@ def bench_kineto(
...
     is_tuple = isinstance(kernel_names, tuple)
-    prof_lines = (
-        prof.key_averages()
-        .table(sort_by="cuda_time_total", max_name_column_width=100)
-        .split("\n")
+    prof_table = prof.key_averages().table(
+        sort_by="cuda_time_total", max_name_column_width=100
     )
+
+    if "CUDA" not in prof_table:
```
ofirfarjun7, line 232: The intent makes sense, but please don’t detect CUDA profiling availability by searching the rendered profiler table for "CUDA". That table is presentation text and may change across PyTorch versions. Can we maybe inspect prof.key_averages() structured device timing instead, and raise KinetoUnavailableError only when no CUDA/device timing was recorded?

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -435,14 +436,18 @@ def test_barrier():
...
-        dispatch_t, combine_t = bench_kineto(
-            partial(test_func, return_recv_hook=return_recv_hook),
-            kernel_names=("dispatch", "combine"),
-            barrier_comm_profiling=True,
-            suppress_kineto_output=False,
-            num_kernels_per_period=2 if return_recv_hook else 1,
-            barrier_fn=test_barrier,
-        )
+        try:
+            dispatch_t, combine_t = bench_kineto(
```
ofirfarjun7, line 440: This is the only place we use bench_kineto?

PR #2010, opened 2026-07-30
telemetry: expose selected backend with transfer stats

catyans:
## What?
Expose the selected backend name alongside per-transfer telemetry.

- add an ABI-safe three-output C++ `getXferTelemetry` overload
- expose `backendName` on Python `nixlXferTelemetry`
- cover the backend name in C++ and Python transfer tests
- document the new field/access pattern

## Why?
Per-transfer telemetry currently contains timing and size data but not the NIXL backend that carried the request. Consumers building request-level traces have to call `queryXferBackend` separately and, in Python, map the returned handle back to a name.

This addresses the coarse backend-attribution part of #1975. Backend-internal path details such as UCX RDMA-vs-TCP selection or the specific NIC/HCA still require a backend-facing path-info hook and are intentionally left for a follow-up design.

## How?
The existing two-output C++ API and `nixlXferTelemetry` layout remain unchanged for ABI compatibility. A new overload returns the backend name as a third output and reuses the existing telemetry validation path. The Python binding wraps the numeric telemetry and backend name in the existing Python-visible `nixlXferTelemetry` shape.

## Testing
- `pre-commit run --files <changed files>`
- `git clang-format --diff`
- C++20 syntax check for the public overload
- Python `compileall` for changed API/tests

A full Meson build was not run locally because the development host is macOS and NIXL's build currently requires Linux `dl`/`rt` libraries and transport dependencies. CI should exercise the Linux C++/Python integration tests.

## Summary by CodeRabbit

* **New Features**
  * Transfer telemetry now reports the backend selected for each completed transfer.
  * Python telemetry results include a read-only `backendName` field, such as `UCX`, `GDS`, or `POSIX`.
* **Documentation**
  * Clarified telemetry behavior, including descriptor counts after merging and backend name availability.
* **Bug Fixes**
  * Improved telemetry validation for transfer and post-processing durations.


docs/telemetry.md
```diff
@@ -144,6 +144,10 @@ Telemetry is configured by environment variables:
...
 - Telemetry is requested either via `NIXL_TELEMETRY_ENABLE` or via the agent config flag `captureTelemetry` (`capture_telemetry=True` in Python). It is fully off only when it is not requested.
+- The three-output C++ overload `getXferTelemetry(request, telemetry, backend_name)` returns the
```
ColinNV: The PR looks good, but this does not feel like the right place for documenting the new function -- it's in the middle of the run-time configuration...

test/python/test_nixl_api.py
```diff
@@ -251,7 +251,9 @@ def test_incorrect_plugin_env(monkeypatch):
         nixl_agent("bad env agent")
 
 
-def _run_xfer_telemetry_check(agent1, agent2, expect_telemetry: bool = True) -> None:
+def _run_xfer_telemetry_check(
+    agent1, agent2, backend_name: str, expect_telemetry: bool = True
+) -> None:
```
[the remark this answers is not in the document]
catyans: Fixed in 614a4ea5: `expect_telemetry` is now keyword-only.

docs/telemetry.md
catyans: Agreed. Moved the API documentation out of Runtime Configuration into a dedicated “Per-transfer request telemetry” subsection in 614a4ea5.

src/api/cpp/nixl.h
```diff
@@ -310,6 +310,19 @@ class nixlAgent {
...
+        /**
+         * @brief  Get telemetry data and the selected backend name for `req_hndl`.
+         *
+         * @param  req_hndl          Transfer request handle from makeXferReq/createXferReq
+         * @param  telemetry [out]   Output telemetry information
+         * @param  backend_name [out] Name of the backend selected for the transfer
+         * @return nixl_status_t     Error code if call was not successful
+         */
+        nixl_status_t
+        getXferTelemetry(const nixlXferReqH *req_hndl,
```
iyastreb, line 322: Instead of introducing new API function, maybe you could just extend `nixl_xfer_telem_t` structure with backend information?
There it could be just a `const char *` pointing to a static backend name, no even need to allocate a string actually

e-eygin: +1 on this, and I don't think the ABI argument in the PR description blocks it.

`nixl_xfer_telem_t` is a plain C++ header struct that the caller allocates; there is no C++ ABI guarantee in this repo, and `CONTRIBUTING.md` promises "backward compatibility and stable APIs", which is source-level — adding a field to it is source-compatible. The struct that *is* a frozen ABI is a different one: `nixl_capi_xfer_telemetry_s` in `src/bindings/rust/wrapper.h`, which is what the C/Rust binding marshals into, and it is untouched either way.

So extending the struct costs nothing that the description claims it costs, and it removes the second overload, the `nixlPyXferTelemetry` wrapper in the Python bindings, and the C++/Python/Rust divergence in one go.

One caveat on `const char *`: the name is not static. `getType()` returns a reference to a `std::string` member of `nixlBackendEngine` (`src/api/cpp/backend/backend_engine.h`), so the pointer is only valid while the agent lives. That outlives the request handle, so it is safe in practice, but the contract would have to be documented — otherwise a caller holding a copy of the telemetry struct past agent teardown gets a dangling pointer. A `std::string` field matching `nixl_backend_t` sidesteps the question.

src/api/cpp/nixl.h
```diff
@@ -310,6 +310,19 @@ class nixlAgent {
...
+                         nixl_xfer_telem_t &telemetry,
+                         nixl_backend_t &backend_name) const;
```
e-eygin, line 324: The contract for `backend_name` isn't documented: it is written only when the inner call returns `NIXL_SUCCESS` and left untouched otherwise. That matters more than usual here because the two-arg overload can return `req_hndl->status == NIXL_IN_PROG`, a non-negative, non-error status — so "the call did not fail" and "`backend_name` was written" are not the same condition, and a caller checking `status >= 0` will read whatever it passed in.

Please either state "unmodified unless `NIXL_SUCCESS` is returned" in the Doxygen block, or clear it unconditionally on entry. Nothing in the added tests pins this down either — `backend_name` is only asserted on the success path.

test/gtest/test_transfer.cpp
```diff
@@ -436,18 +436,20 @@ class TestTransfer : public nixl_test_t {
...
                 nixl_xfer_telem_t telemetry;
+                nixl_backend_t backend_name;
                 if (expected_telem_status == NIXL_ERR_NO_TELEMETRY) {
                     const LogIgnoreGuard lig("cannot return values when telemetry is not enabled");
-                    status = from.getXferTelemetry(xfer_req, telemetry);
+                    status = from.getXferTelemetry(xfer_req, telemetry, backend_name);
                     EXPECT_EQ(status, expected_telem_status);
                 } else {
-                    status = from.getXferTelemetry(xfer_req, telemetry);
+                    status = from.getXferTelemetry(xfer_req, telemetry, backend_name);
```
e-eygin, line 445: Both call sites move to the three-arg overload, which leaves the two-arg `getXferTelemetry()` with no C++ test coverage at all — after this PR only the Rust wrapper (`nixl_capi_get_xfer_telemetry`) still calls it.

Keeping that overload source- and ABI-compatible is the stated reason for adding an overload rather than extending the struct, so it should keep at least one exercised call site. Leaving one of these two on the two-arg form would cover both.

src/bindings/python/nixl_bindings.cpp
```diff
@@ -187,20 +192,25 @@ PYBIND11_MODULE(_bindings, m) {
         .value("NIXL_ERR_NOT_SUPPORTED", NIXL_ERR_NOT_SUPPORTED)
         .export_values();
 
-    py::class_<nixl_xfer_telem_t>(m, "nixlXferTelemetry")
+    py::class_<nixlPyXferTelemetry>(m, "nixlXferTelemetry")
```
e-eygin, line 195: Re-registering the `nixlXferTelemetry` Python name onto the wrapper means `nixl_xfer_telem_t` is no longer a pybind-registered type. That is fine today, since this is the only site that produces one, but it turns any future binding that returns or accepts a `nixl_xfer_telem_t` into a *runtime* caster error rather than a compile error.

The forced move of `totalBytes`/`descCount` from `def_readonly` member pointers to lambdas is also churn caused purely by the nesting. Both of these disappear if the backend name lands in `nixl_xfer_telem_t` instead of a binding-local wrapper.

e-eygin (commented):
The change is internally coherent and the lint-level checks are green, but note the NVIDIA build/test CI has not run yet (copy-pr-bot vetting), so nothing here has been exercised against gtest or the Python suite.

My main concern is the shape of the API rather than the implementation, and I don't think the justification given for it holds — details in my reply to @iyastreb's thread on `src/api/cpp/nixl.h`. Three more specific notes are inline.

Two points that don't map to a changed line:

**The C/Rust surface is left behind.** `nixl_capi_get_xfer_telemetry()` (`src/bindings/rust/wrapper.cpp`) and `XferTelemetry` (`src/bindings/rust/src/lib.rs`) do not gain the backend name, so C++ and Python have it and Rust does not. If the value moves into `nixl_xfer_telem_t`, extending `nixl_capi_xfer_telemetry_s` and the Rust struct is a few lines and keeps all three surfaces aligned.

**The "Why?" overstates the Python gap.** `nixl_agent.query_xfer_backend()` (`src/api/python/_api.py`) already returns the backend *name*, not a handle, so consumers are not mapping handles back to names today. The real benefits are narrower but still worth stating accurately: one call instead of two on the completion path, and it works when the backend was not created through the Python `nixl_agent` wrapper — that reverse lookup over `self.backends` raises `StopIteration` otherwise.

Nits, not worth a round trip on their own: the new docs section leads with "without changing the ABI of `nixlXferTelemetry`", which is reviewer-facing rationale rather than user documentation; and `backend_name` is a required positional in `_run_xfer_telemetry_check()` that is never used on the `expect_telemetry=False` path.

PR #2016, opened 2026-07-30
build: discover cuobjclient across CUDA Toolkit versions

### LINKED ISSUE #1763: cuobjclient dependency pinned to version 13.1
The dependency check for cuobjclient to support RDMA accelerated object storage engines in the OBJ plugin is pinned to version 13.1.

cuobj_dep = dependency('cuobjclient-13.1', required: false)

The release of 13.3 and an updated cuobjclient library (v1.2.0) addresses some issues in the previous library version.  However, since the version is pinned, there is no way to pick up the new library.  13-1 should be a minimum package version, accepting later versions also.

jgoldsch12:
The OBJ plugin's accelerated S3 engines looked for the pkg-config module 'cuobjclient-13.1' only. That module name encodes the CUDA Toolkit release the library ships with, so newer toolkits (e.g. 13.3, carrying cuObject client v1.2.0) install 'cuobjclient-13.3.pc' and were never picked up, leaving the accelerated engines silently disabled.

Enumerate the installed cuobjclient pkg-config modules and select the newest one that is at least 13.1, treating 13.1 as a minimum rather than an exact requirement.

Fixes ai-dynamo/nixl#1763

## Summary by CodeRabbit

* **Build Improvements**
  * Automatically detects the newest compatible CUObjClient library version (CUDA Toolkit 13.1 or later) instead of requiring one fixed version.
  * S3 accelerated engines now build when a suitable installed version is available and are skipped otherwise.
* **Documentation**
  * Updated dependency and configuration guidance for S3 accelerated engines.
  * Build messages now report the detected or required CUObjClient version.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -147,8 +147,40 @@ has_posix_aio = cpp.has_function('aio_cancel', prefix: '#include <aio.h>', depen
 # Check whether the POSIX plugin is enabled
 has_posix_plugin = has_linux_aio or has_io_uring or has_posix_aio
 
-# Check for CUObject Client library
-cuobj_dep = dependency('cuobjclient-13.1', required: false)
+# Check for CUObject Client library.
+# The library ships with the CUDA Toolkit and its pkg-config file is named after
+# the toolkit release it belongs to (cuobjclient-13.1.pc, cuobjclient-13.3.pc, ...),
+# so a single module name cannot be used. Probe the pkg-config modules that are
+# installed and pick the newest one that satisfies the minimum supported version.
+cuobj_min_version = '13.1'
+cuobj_dep = dependency('cuobjclient', required: false)
+
+if not cuobj_dep.found()
+    pkgconfig_prog = find_program('pkg-config', 'pkgconf', required: false)
+    cuobj_module = ''
+    cuobj_version = ''
+    if pkgconfig_prog.found()
+        pkgconfig_modules = run_command(pkgconfig_prog, '--list-all', check: false)
+        if pkgconfig_modules.returncode() == 0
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -147,8 +147,40 @@ has_posix_aio = cpp.has_function('aio_cancel', prefix: '#include <aio.h>', depen
...
+    cuobj_version = ''
+    if pkgconfig_prog.found()
+        pkgconfig_modules = run_command(pkgconfig_prog, '--list-all', check: false)
+        if pkgconfig_modules.returncode() == 0
+            foreach line : pkgconfig_modules.stdout().strip().split('\n')
+                module = line.strip().split(' ')[0]
+                if module.startswith('cuobjclient-')
+                    version = module.split('cuobjclient-')[1]
+                    if version.version_compare('>=' + cuobj_min_version)
+                        if cuobj_version == '' or version.version_compare('>' + cuobj_version)
```
ofer, line 170: pkg-config does not guarantee an order, shouldn't it take the latest version?

jgoldsch12: Thanks for pointing out that `--list-all` order isn't stable. The loop doesn't rely on it though: rather than taking the first or last match, it keeps a running maximum via `version_compare('>' + cuobj_version)`, so it always lands on the newest module regardless of emission order. I verified this by testing with a shuffled `--list-all` output; it consistently selects the highest version, and because `version_compare` is semantic rather than lexicographic it also correctly prefers `13.10` over `13.3`.

If you'd rather, I could rename `cuobj_module`/`cuobj_version` to `cuobj_best_module`/`cuobj_best_version` and add a one-line comment if that would help.

ofer (approved):

PR #2017, opened 2026-07-30
PLUGIN/UCX: Init only needed fields in UCP

iyastreb:
## What?
There is a visible 2% perf degradation in nixlbench (512Bx64000 8 threads) on EOS and DFW

## Why?
The default NIXL path zero-initializes the entire ucp_request_param_t on every operation. After openucx/ucx#11258 increased the size of the structure, the cost of this existing initialization became more noticeable.

## How?
We need to explicitly set only fields that we populate, no need to zero-init the rest

## Summary by CodeRabbit

* **Performance**
  * Improved handling of low-probability transfer failures.
  * Streamlined internal transfer request setup for send, read, and write operations.

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -237,9 +237,8 @@ nixlUcxEp::sendAm(nixl::ucx::am_cb_op_t msg_id,
         return status;
     }
 
-    ucp_request_param_t param = {0};
-
-    param.op_attr_mask |= UCP_OP_ATTR_FIELD_FLAGS;
+    ucp_request_param_t param;
+    param.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS;
     param.flags = flags;
 
     nixl_ucx_am_cb_ctx_ptr_t ctx;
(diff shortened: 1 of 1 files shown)


ColinNV (approved):

tomerg-nvidia (approved):

michal-shalev:
Nice catch.

One comment about the wording in the **Why** section: from my understanding, this is a pre-existing NIXL initialization overhead that became more noticeable after `openucx/ucx#11258` increased the size of `ucp_request_param_t`.

Something like the following would be more accurate:

> The default NIXL path zero-initializes the entire `ucp_request_param_t` on every operation. After `openucx/ucx#11258` increased the size of the structure, the cost of this existing initialization became more noticeable.

I think this better reflects the root cause.

iyastreb:
> Nice catch.
> 
> One comment about the wording in the **Why** section: from my understanding, this is a pre-existing NIXL initialization overhead that became more noticeable after `openucx/ucx#11258` increased the size of `ucp_request_param_t`.
> 
> Something like the following would be more accurate:
> 
> > The default NIXL path zero-initializes the entire `ucp_request_param_t` on every operation. After `openucx/ucx#11258` increased the size of the structure, the cost of this existing initialization became more noticeable.
> 
> I think this better reflects the root cause.

Ok, updated

PR #2018, opened 2026-07-30
telemetry: optimize sharedRingBuffer and eliminate false sharing

pooriaPoorsarvi:
## What?

Currently sharedRingBuffer is an SPSC buffer, and already optimized for that with two atomics instead of locks. However the two atomics sit on the same cache line, so writing one invalidates the line on the other core. Also every push and pop reads the other side's position from shared memory, which pulls that core's cache line on every call. This is unnecessary in an SPSC buffer: since the producer and consumer each advance their position monotonically, a cached copy of the other side's position is safe to use. A stale value can only make the full/empty check conservative, never wrong.

Because the header layout changes, this also bumps TELEMETRY_VERSION from 4 to 5 and updates the tests and Python reader's struct to match. 

## Why?

The two indices sharing a cache line means every index update by one side invalidates the other side's line, and every push/pop re-reads the opposing index, adding a cross-core cache miss even when nothing changed. In NIXL as shipped the buffer is written only by the periodic flush task, so this cost is a negligible fraction of end-to-end time at current rates. The change removes the per-call coherence traffic at no cost to the current path, and the benefit grows if the buffer moves onto a hotter path (e.g., a higher flush rate or more frequent use).

## How?

We use alignas to make sure the two atomics are not on the same cache line. The alignment is conservatively hardcoded to 256, the value GCC uses for hardware_destructive_interference_size on ARM, instead of using the std constant directly, so the shared memory layout stays identical across compilers and build flags. The repeated reads of the opposing index are addressed with cached values that are only refreshed (with a load acquire) when the buffer appears full/empty. The caches are seeded on attach paths, the same change is applied to the Python example reader, and a new gtest covers a consumer re-attaching mid-stream.

The header fields are also ordered so version keeps byte offset 16 across layout versions, so a reader on a mismatched file always reads a real version field and fails with a clean mismatch.

## Results

To test the throughput improvement I send 200,000,000 16-byte events (same size as nixlTelemetryEvent) on a ring with 65536 entries from a pinned producer core to a pinned consumer core through the real sharedRingBuffer over /dev/shm. Latency is a two-ring ping-pong RTT, averaged over 1,000,000 round trips. All metrics are averaged over 3 runs on a multi-socket ARM server:

| Core pairing | Baseline M events/s | Optimized M events/s | Speedup | Baseline RTT (ns) | Optimized RTT (ns) | Speedup |
|---|---|---|---|---|---|---|
| same-socket | 11.9 | 51.9 | **4.4×** | 1022 | 555 | **1.8×** |
| cross-socket | 2.5 | 9.2 | **3.7×** | 4448 | 3353 | **1.3×** |

Note that this is a synthetic saturation microbenchmark studying the buffer's standalone behavior under pressure; it measures the buffer's ceiling, not an expected end-to-end win. I can include the benchmark and full logs as part of this PR if you'd like.

## New Features
- Updated the telemetry shared-memory format to version 5.
- Improved ring-buffer alignment and stream-reading efficiency.
- Enabled readers to resume correctly when reconnecting to an active stream.

## Bug Fixes
- Fixed event handling when a reader reconnects while unread events remain.

## Documentation
- Documented the updated telemetry format, buffer layout, versioning, and field offsets.
- Clarified compatibility validation and handling for mismatched telemetry versions.


iyastreb (dismissed):

docs/telemetry.md
```diff
@@ -165,7 +165,7 @@ Following sections applied specifically for configuration and usage of cyclic bu
...
+`TELEMETRY_VERSION` is the binary layout version of `nixlTelemetryEvent` as serialized in the shared-memory cyclic buffer. Bump it whenever the struct's size, field offsets, or enum widths change, or when a new event type is added to the stream (readers must know the full event-type set). Readers exact-match this value; the C++ shared-ring-buffer path unlinks the file on mismatch, while the Python example raises a `RuntimeError`. There is no backward-compatibility path. The current version is `5`: it was bumped from `4` when the buffer header was padded to two 256-byte-aligned blocks (512 bytes) to eliminate false sharing between `write_pos` and `read_pos`, which changed all header field offsets and the data-array offset. Version `4` was bumped from `3` when the `agent_telemetry_events_dropped` event type (`AGENT_TELEMETRY_EVENTS_DROPPED`) was added.
```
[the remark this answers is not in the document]
pooriaPoorsarvi: Makes sense, the opening sentence predates this PR, but worth updating the whole paragraph to be consistent. We can do something like:

```suggestion
`TELEMETRY_VERSION` is the binary layout version of the shared-memory cyclic buffer, covering its header, data placement, and the `nixlTelemetryEvent` entries serialized in it. Bump it whenever the buffer header layout or the struct's size, field offsets, or enum widths change, or when a new event type is added to the stream (readers must know the full event-type set). Readers exact-match this value; the C++ shared-ring-buffer path unlinks the file on mismatch, while the Python example raises a `RuntimeError`. There is no backward-compatibility path. The current version is `5`: it was bumped from `4` when the buffer header was padded to two 256-byte-aligned blocks (512 bytes) to eliminate false sharing between `write_pos` and `read_pos`, which changed all header field offsets and the data-array offset. Version `4` was bumped from `3` when the `agent_telemetry_events_dropped` event type (`AGENT_TELEMETRY_EVENTS_DROPPED`) was added.
```

I’d be more than happy to push my suggestion as a follow-up commit myself, go with the bot suggestion, or leave as is for this PR, feel free to let me know which one you’d like.

src/utils/common/cyclic_buffer.h
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
pooriaPoorsarvi: I think this is consistent with other SPDX headers in the repo, including the unchanged header of telemetry_test.cpp in this same PR, which already has the identical line. But feel free to let me know if this needs a change.

pooriaPoorsarvi:
Thanks for the triage, but I think there are some inconsistencies in @svc-nixl's conclusion.

The code before this PR was already lock-free and CAS-free on the head and tail indices, using release/acquire memory ordering with atomics because it is an SPSC buffer. This PR does not change that concurrency model; it only pads the two indices and caches the opposing index and the mask.

I also went through the tests the bot flagged. The suites are all instantiations of the `TestTransfer` fixture, and that fixture explicitly disables telemetry before creating any agents: `TestTransfer::SetUp()` sets `NIXL_TELEMETRY_ENABLE=n` ([test_transfer.cpp:154-155](https://github.com/ai-dynamo/nixl/blob/f18d9fa43f8d0e2f0a1d9111b98ea858a4ac779a/test/gtest/test_transfer.cpp#L154-L155)). With telemetry off, the agent never constructs a telemetry object, so no `sharedRingBuffer` is ever created and none of this PR's code runs in those tests. The one transfer fixture that does not do this, `TestTransferTelemetry`, creates two agents, but they cannot share a buffer: the shared-memory file is keyed by agent name in `getFilePath()` ([buffer_exporter.cpp:26](https://github.com/ai-dynamo/nixl/blob/f18d9fa43f8d0e2f0a1d9111b98ea858a4ac779a/src/core/telemetry/buffer_exporter.cpp#L26)), so each agent maps its own file. And within an agent there is still only one writer: the buffer is private to the exporter, and the only call path to `buffer_.push`, in `nixlTelemetryBufferExporter::exportEvent()` ([buffer_exporter.cpp:41](https://github.com/ai-dynamo/nixl/blob/f18d9fa43f8d0e2f0a1d9111b98ea858a4ac779a/src/core/telemetry/buffer_exporter.cpp#L41)), runs on the agent's single telemetry drain thread, constructed as `pool_(1)` in the `nixlTelemetry` constructor ([telemetry.cpp:117](https://github.com/ai-dynamo/nixl/blob/f18d9fa43f8d0e2f0a1d9111b98ea858a4ac779a/src/core/telemetry/telemetry.cpp#L117)). To make sure I didn't misunderstand the stack, I also logged pid, tid, and file path on every push and pop and ran the suites: within every test, no file is pushed or popped by more than one tid in my runs.

So I think "multi-writer contention on the telemetry shared ring buffer" cannot be the cause. As a next step I'll try and make sure my environment is close to the CI.

Also, if it's alright with you, I can merge the latest main so the next CI run picks it up, in case the slowdown is unrelated and already resolved.

src/utils/common/cyclic_buffer.h
```diff
@@ -50,13 +50,18 @@ template<typename T> class sharedRingBuffer {
     capacity() const;
 
 private:
+    // Hardcoded value instead of std::hardware_destructive_interference_size due to ABI;
+    // needs to be consistent across build and compilers.
+    // Conservatively set to 256 similar to GCC std::hardware_destructive_interference_size for ARM.
+    static constexpr size_t DESTRUCTIVE_INTERFERENCE_SIZE = 256;
+
     struct bufferHeader {
```
e-eygin, line 59: `version` moves from offset 16 to 264. A v5 reader on a leftover v4 file passes the `st_size` guard and reads offset 264, which is event 14's `eventType_` in the v4 layout — if that byte is `5` the version check passes and `capacity` comes from that event's `value_`.

src/utils/common/cyclic_buffer.tpp
```diff
@@ -120,6 +133,8 @@ sharedRingBuffer<T>::bufferHeader::bufferHeader(size_t size) : capacity(size), m
 
     static_assert(std::is_trivially_copyable<T>::value,
                   "T must be trivially copyable for shared memory");
+    static_assert(sizeof(bufferHeader) == 2 * DESTRUCTIVE_INTERFERENCE_SIZE,
```
e-eygin, line 136: This catches a size change but not a reordering: swapping any two fields keeps `sizeof` at 512 while silently desynchronizing the Python reader. `bufferHeader` is standard-layout, so `offsetof` asserts work here and would cover the case the assert is really guarding against.

e-eygin:
One request on the description rather than the code. The "Why?" says the false-sharing cost "on the producer side is on the transfer data path", but that isn't where this buffer sits today. The transfer path calls `stagingQueue_.tryPush` (mutex + vector append); the ring buffer is written only by the 100 ms periodic flush task, and the reference readers poll at 500 ms. So in NIXL as shipped, the line ping-pongs at roughly 10 Hz between a batch producer and a 2 Hz consumer, and the coherence traffic being removed is close to zero.

That doesn't make the change wrong — it's cheap, correct, and it's the right shape for when this buffer does move onto a hotter path. I'd just ask that the claim be softened and the table labeled as a synthetic saturation benchmark, so the 4.4× isn't read as an expected end-to-end win.

src/utils/common/cyclic_buffer.tpp
```diff
@@ -170,6 +185,8 @@ sharedRingBuffer<T>::createCyclicBuffer(const std::string &name, int version) {
     new (header_) bufferHeader(bufferSize_);
     header_->version.store(version, std::memory_order_release);
     header_->expected_version = version;
+
+    cachedMask_ = header_->mask;
```
e-eygin, line 199: Only `cachedMask_` is seeded here; the other two are correct only because their in-class `= 0` defaults happen to match a freshly constructed header. Seeding all three on both paths would make the invariant explicit.

pooriaPoorsarvi:
> One request on the description rather than the code. The "Why?" says the false-sharing cost "on the producer side is on the transfer data path", but that isn't where this buffer sits today. The transfer path calls `stagingQueue_.tryPush` (mutex + vector append); the ring buffer is written only by the 100 ms periodic flush task, and the reference readers poll at 500 ms. So in NIXL as shipped, the line ping-pongs at roughly 10 Hz between a batch producer and a 2 Hz consumer, and the coherence traffic being removed is close to zero.
> 
> That doesn't make the change wrong — it's cheap, correct, and it's the right shape for when this buffer does move onto a hotter path. I'd just ask that the claim be softened and the table labeled as a synthetic saturation benchmark, so the 4.4× isn't read as an expected end-to-end win.

Absolutely, that makes sense. I appreciate the context on where the buffer actually sits today. As you described, the `sharedRingBuffer` is written only by the periodic flush task, so it accounts for a negligible fraction of end-to-end time at current rates, and by Amdahl's law the end-to-end speedup from this change is correspondingly negligible; it matters more if the buffer moves onto a hotter path (e.g., a higher flush rate). I've updated the PR description accordingly: the "Why?" section now explains that the change removes the per-call coherence traffic at no cost to the current path, and that the benefit grows if the buffer sees more frequent use, and the Results table is updated with more detail, explicitly stating that the benchmark I ran is a synthetic saturation microbenchmark studying the buffer's standalone behavior under pressure, not an end-to-end benchmark.

src/utils/common/cyclic_buffer.tpp
pooriaPoorsarvi: Makes sense, all cached values are now seeded from the header on both attach paths (bff1a5e9).

src/utils/common/cyclic_buffer.tpp
pooriaPoorsarvi: Sure, added that: pinned all six field offsets next to the sizeof assert (4d046c39).

src/utils/common/cyclic_buffer.h
pooriaPoorsarvi: You're right, this was a real gap. I fixed it by reordering the fields so `version` keeps byte offset 16 across layout versions (7a2ba6d8): a reader on a leftover v4 file now reads the real version field and fails with a clean `Version mismatch: expected 5, got 4`. Happy to add a gtest for this if you'd like.

examples/python/telemetry_reader.py
```diff
@@ -86,12 +87,22 @@ class BufferHeader(ctypes.Structure):
...
-        ("mask", ctypes.c_size_t),
+        ("write_pos", ctypes.c_size_t),  # [0, 8)
+        ("capacity", ctypes.c_size_t),  # [8, 16)
+        # version stays at offset 16 across layout versions (see cyclic_buffer.h)
+        ("version", ctypes.c_uint32),  # [16, 20)
+        ("expected_version", ctypes.c_uint32),  # [20, 24)
+        ("mask", ctypes.c_size_t),  # [24, 32)
+        (
+            "_pad_write",
+            ctypes.c_char * (DESTRUCTIVE_INTERFERENCE_SIZE - 32),
```
e-eygin, line 98: `_pad_write` hardcodes `32` for the preceding fields; nothing enforces that `sizeof(BufferHeader)` is still 512. CI won't catch a drift that leaves version readable either — the reader just prints garbage events until `.gitlab/test_python.sh` `SIGINTs` it.

pooriaPoorsarvi: You're right, the Python side had no enforcement. Added import-time checks mirroring the C++ static_asserts, header size and all eight field offsets including the hand-maintained pads, raising on any drift (60276843).

Longer term, it might be worth adding Python bindings for a reader (pybind11 wrapping `sharedRingBuffer<nixlTelemetryEvent>`): it would remove the hand-maintained layout mirror and hide the complexity from users. The standalone script still has value as a dependency-free example, so this would complement it rather than replace it. Happy to take that as a follow-up PR if there's interest.

src/utils/common/cyclic_buffer.h
```diff
@@ -50,13 +51,21 @@ template<typename T> class sharedRingBuffer {
...
+        // on a mismatched file always reads a real version field instead of
+        // arbitrary bytes.
+        alignas(DESTRUCTIVE_INTERFERENCE_SIZE) std::atomic<size_t> write_pos{0};
+        const size_t capacity;
         std::atomic<int> version{0};
         int expected_version{0};
-        const size_t capacity;
-        size_t mask;
+        const size_t mask;
+        alignas(DESTRUCTIVE_INTERFERENCE_SIZE) std::atomic<size_t> read_pos{0};
```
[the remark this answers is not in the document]
pooriaPoorsarvi: Both checks are valid requirements: a file must match the current version and be large enough for its layout, so failing either one is correct grounds to reject, and nothing is read either way. A file smaller than the v5 header cannot be a valid v5 file at all, so everything in that window is an older or invalid layout and is still rejected; in fact such a file fails both requirements at once, and the only difference is that a version mismatch would be the more precise error message. The map-before-version ordering is also pre-existing, this PR only grew the header size. With `version` still pinned at offset 16, probing it early before the size check is straightforward but it would be a new early probe that unlinks; I'd be happy to do that here or as a follow-up PR if the maintainers prefer.

examples/python/telemetry_reader.py
```diff
@@ -86,15 +87,44 @@ class BufferHeader(ctypes.Structure):
...
+            * (DESTRUCTIVE_INTERFERENCE_SIZE - ctypes.sizeof(ctypes.c_size_t)),
+        ),  # match C++ compiler's tail padding: [264, 512)
     ]
 
 
+# Mirror the static_asserts in cyclic_buffer.tpp so a layout drift fails.
+# The pads are checked too since they are maintained by hand here, while
+# the C++ compiler derives them from alignas.
+if ctypes.sizeof(BufferHeader) != 2 * DESTRUCTIVE_INTERFERENCE_SIZE:
+    raise RuntimeError("BufferHeader layout drift: unexpected size")
```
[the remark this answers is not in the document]
pooriaPoorsarvi: Following pre-existing convention, also present in this same file, for example the existing raise RuntimeError for the version mismatch.

pooriaPoorsarvi:
Updated the branch with the latest main (62b500e). I also followed up on the "Run CPP tests" failure the CI triage bot attributed to this PR: I re-ran `.gitlab/test_cpp.sh`, unmodified except for two deviations: the Azure section (not available in my environment) and one socket test (depends on network behavior I can't change due to not having root on the compute node). Neither is implicated in the reported failure.

I ran it on both this branch (62b500e) and current main (2301000) as a control: all tests passed on both (209/209 and 208/208, with the extra case being the test this PR adds), with total runtime 5m59s vs 6m02s (within noise). The reported slowdown does not reproduce here, and the branch is now up to date with main so a fresh CI run will reflect the current state. Happy to make further changes if anything remains open on the earlier review items.

pooriaPoorsarvi:
Glad to see “Run CPP tests” passing on the latest run, matching my local tests. The newly failing test was added to the main branch after my recent merge. Looking at the triage it traces to another PR, and the same `test_prep_mem_view` failure appears on other open PRs (e.g. #2118 which is a version update and can't be the culprit) with the same triage conclusion. Happy to look into it further if you’d like, or to wait until the fix lands, then merge main again and re-test.

pooriaPoorsarvi:
@ColinNV I think the latest merge commit being unsigned is what blocked the auto-trigger, the pipelines never started. Happy to push a signed merge of latest main on top of your commit, or wait for you to re-trigger.

ColinNV:
> Happy to push a signed merge of latest main on top of your commit, or wait for you to re-trigger.

OK, go ahead.

pooriaPoorsarvi:
@ColinNV Sure, pushed the signed merge of latest main.

pooriaPoorsarvi:
Still declined even though the head commit shows verified on GitHub. Unfortunately it probably needs a /build from your side.

pooriaPoorsarvi:
I checked the CI logs yesterday, there were two separate failures. My push declined on the signature check even though the last commit has both DCO and an SSH signature that shows verified on GitHub. The /build run failed with "Security check failed, LDAP account is not active/enabled", which I don't think I can follow up on from my side. I see the blossom CI has been updated recently since then, so I'll wait from your side for retrying the build once the CI is fixed. Feel free to let me know if anything is needed from my end.

PR #2023, opened 2026-07-31
PLUGIN/UCX: Init only needed fields in UCP (#2017)

ovidiusm:
## What?
There is a visible 2% perf degradation in nixlbench (512Bx64000 8 threads) on EOS and DFW

## Why?
UCP SGL PR https://github.com/openucx/ucx/pull/11258 increased the size of request param, which is zero-initialized in NIXL on each operation. This extra initialization costs 2% of performance.

## How?
We need to explicitly set only fields that we populate, no need to zero-init the rest


ovidiusm:
Duplicate of #2022

PR #2024, opened 2026-07-31
PLUGINS/UCX: Build notification right after post for overlapping

iyastreb:
## What?
Micro-optimization that decreases Tx latency by building notification message right after posting, while requests are in-flight, and therefore benefit from overlapping.

## Why?
Before we were building notification (around 6 string allocations) after payload completion, therefore consuming CPU cycles before we can send the notification.

## How?
Serialize the notification once at post time and keep it in the handle as unique_ptr.

#### Testing performance on 4KB x 1 nixlbench WRITE
I see stable **3.5% BW** and latency improvement on low dimensions.
```
4096                1              0.935155       4.4            10.0           10.0           1.1            2.0            3.2            4.0
4096                1              0.968380       4.2            13.0           13.0           1.2            2.0            2.9            4.0
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of UCX notification delivery by keeping message data available until transmission completes.
  * Enhanced asynchronous notification handling for safer endpoint usage and more dependable cleanup.
  * Improved consistency when sending notifications through communication endpoints.

* **Performance**
  * Reduced temporary memory reallocations when constructing serialized messages.
  * Streamlined notification preparation and transmission for more efficient processing.

### DIFF
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1400,7 +1400,7 @@ nixlUcxEngine::sendNotif(std::unique_ptr<std::string> &&msg, const nixlUcxEp &ep
                      buffer->size(),
                      UCP_AM_SEND_FLAG_EAGER,
                      req,
-                     deleter);
+                     std::move(deleter));
 }
 
 nixl_status_t
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1385,7 +1385,7 @@ nixlUcxEngine::buildNotif(const std::string &msg) const {
 nixl_status_t
 nixlUcxEngine::sendNotif(std::unique_ptr<std::string> &&msg, const nixlUcxEp &ep, nixlUcxReq *req) {
     std::string *buffer = msg.release();
-    auto deleter = [buffer, req](void *completed_request, void *ptr) {
+    auto cleanup = [buffer, req](void *completed_request, void *ptr) {
         delete buffer;
         if ((req == nullptr) && (completed_request != nullptr)) {
             /* Caller is not interested in the request, free it */
@@ -1400,7 +1400,7 @@ nixlUcxEngine::sendNotif(std::unique_ptr<std::string> &&msg, const nixlUcxEp &ep
                      buffer->size(),
                      UCP_AM_SEND_FLAG_EAGER,
                      req,
-                     std::move(deleter));
+                     std::move(cleanup));
 }
 
 nixl_status_t
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -1383,7 +1383,7 @@ nixlUcxEngine::buildNotif(const std::string &msg) const {
 }
 
 nixl_status_t
-nixlUcxEngine::sendNotif(std::unique_ptr<std::string> msg, const nixlUcxEp &ep, nixlUcxReq *req) {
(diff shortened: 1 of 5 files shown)


src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1359,13 +1361,21 @@ nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
...
                       buffer->size(),
                       UCP_AM_SEND_FLAG_EAGER,
                       req,
                       deleter);
 }
 
+nixl_status_t
+nixlUcxEngine::notifSendPriv(const std::string &remote_agent,
+                             const std::string &msg,
+                             const std::unique_ptr<nixlUcxEp> &ep,
```
rakhmets: `ep` can be passed as reference (const ref is possible only with adding `const` to  `nixlUcxEp::sendAm`), since `notifSendPriv` and `sendNotif` do not need object ownership.

iyastreb: ok, done

rakhmets (dismissed):

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1335,19 +1333,19 @@ nixlUcxEngine::progressLoop() {
...
 
     ser_des.addStr("name", localAgent);
     ser_des.addStr("msg", msg);
     // TODO: replace with mpool for performance
+    return std::make_unique<std::string>(ser_des.exportStr());
+}
 
-    std::string *buffer = new std::string(ser_des.exportStr());
+nixl_status_t
+nixlUcxEngine::sendNotif(std::unique_ptr<std::string> msg, const nixlUcxEp &ep, nixlUcxReq *req) {
```
ColinNV: Can accept `msg` as r-value reference.

src/plugins/ucx/ucx_backend.cpp
```diff
@@ -1335,19 +1333,19 @@ nixlUcxEngine::progressLoop() {
...
+    std::string *buffer = msg.release();
```
ColinNV, line 1387: At this point, why not directly put the `unique_ptr` into the closure and get rid of the manual `delete`?

iyastreb: A lambda capturing a unique_ptr is move-only, so it won't compile. I don't see an easy way to have it

src/plugins/ucx/ucx_backend.cpp

src/plugins/ucx/ucx_backend.cpp
iyastreb: ok, done

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -225,7 +225,7 @@ nixlUcxEp::sendAm(nixl::ucx::am_cb_op_t msg_id,
                   size_t len,
                   uint32_t flags,
                   nixlUcxReq *req,
-                  const am_deleter_t &deleter) {
+                  am_deleter_t &&deleter) const {
```
ColinNV: Perhaps this is a good opportunity to rename "deleter" to "cleanup".

In the context of unique and shared pointers a "deleter" has a clearly defined role, but this "deleter" has a different role making the naming slightly confusing.

ColinNV (dismissed):
The "deleter" rename can also be done separately.

src/plugins/ucx/ucx_utils.cpp
iyastreb: renamed

ColinNV: Thanks.

ColinNV (approved):

PR #2027, opened 2026-07-31
Guard postXferReq/estimateXferCost against stale-generation xfer handles

jjthomas:
Follow-up to #1811 / #1987.

`postXferReq` guards the remote section by name only, so after a disconnect invalidates (frees) gen-N metadata and the peer is re-registered (gen-N+1), posting a gen-N xfer handle dereferences the freed `nixlUcxPublicMetadata` pinned in `targetDescs` — segfault in `sendXferRangeBatch`/`ucp_put_nbx`.                                                                                                                                  
                                                                                                                                                                                                                                                  
Reject stale-generation handles with `NIXL_ERR_NOT_FOUND` (same guard in `estimateXferCost`, which dereferences the same metadata). Verified with a fault-injection reproducer: without the fix the crash reproduces reliably; with it, none.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented remote transfer requests from using outdated metadata after the associated remote resource is re-registered.
  * Added validation to reject invalid or stale handles during cost estimation and transfer submission.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -1073,14 +1073,19 @@ nixlAgent::estimateXferCost(const nixlXferReqH *req_hndl,
     nixl_status_t ret;
     NIXL_SHARED_LOCK_GUARD(data->lock);
 
-    // Check if the remote agent connection info is still valid
-    // (assuming cost estimation requires connection info like transfers)
-    if (!req_hndl->remoteAgent.empty() &&
-        (data->remoteSections_.count(req_hndl->remoteAgent) == 0)) {
-        NIXL_ERROR_FUNC << "invalid request handle, remote agent was invalidated "
-                           "after transfer request creation";
-        data->addErrorTelemetry(NIXL_ERR_NOT_FOUND);
-        return NIXL_ERR_NOT_FOUND;
+    // Check if the remote agent connection info is still valid, matching the
+    // generation guard in postXferReq: the handle's targetDescs pin raw
+    // metadata pointers from the section generation it was created against,
+    // which may have been invalidated (and re-registered) since.
+    if (!req_hndl->remoteAgent.empty()) {
+        const auto sec_it = data->remoteSections_.find(req_hndl->remoteAgent);
+        if (sec_it == data->remoteSections_.end() ||
+            sec_it->second.getGeneration() != req_hndl->remoteGeneration_) {
+            NIXL_ERROR_FUNC << "invalid request handle, remote agent was invalidated "
+                               "after transfer request creation";
+            data->addErrorTelemetry(NIXL_ERR_NOT_FOUND);
+            return NIXL_ERR_NOT_FOUND;
+        }
     }
 
     if (!req_hndl->engine) {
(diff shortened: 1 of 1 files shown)


jjthomas:
@iyastreb @chaunceyjiang mind taking a look at this?

src/core/nixl_agent.cpp
```diff
@@ -1136,13 +1141,28 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
     }
 
     std::shared_lock<nixlLock> read_lock(data->lock);
-    // Check if the remote was invalidated before post/repost
-    if (data->remoteSections_.count(req_hndl->remoteAgent) == 0) {
+    // Check if the remote was invalidated before post/repost. Guard on the
```
iyastreb: Minor: personally I find this comment very excessive and repeating a trivial code below it, which already contains precise error messages. Maybe reduce/remove it?

src/core/nixl_agent.cpp
```diff
@@ -1073,14 +1073,19 @@ nixlAgent::estimateXferCost(const nixlXferReqH *req_hndl,
...
-    // Check if the remote agent connection info is still valid
-    // (assuming cost estimation requires connection info like transfers)
-    if (!req_hndl->remoteAgent.empty() &&
-        (data->remoteSections_.count(req_hndl->remoteAgent) == 0)) {
-        NIXL_ERROR_FUNC << "invalid request handle, remote agent was invalidated "
-                           "after transfer request creation";
-        data->addErrorTelemetry(NIXL_ERR_NOT_FOUND);
-        return NIXL_ERR_NOT_FOUND;
+    // Check if the remote agent connection info is still valid, matching the
+    // generation guard in postXferReq: the handle's targetDescs pin raw
```
iyastreb: imo no need to change the comment

src/core/nixl_agent.cpp
```diff
@@ -1073,14 +1073,19 @@ nixlAgent::estimateXferCost(const nixlXferReqH *req_hndl,
...
+    // metadata pointers from the section generation it was created against,
+    // which may have been invalidated (and re-registered) since.
+    if (!req_hndl->remoteAgent.empty()) {
+        const auto sec_it = data->remoteSections_.find(req_hndl->remoteAgent);
+        if (sec_it == data->remoteSections_.end() ||
```
ColinNV, line 1080: If each generation had an owning `std::shared_ptr` in `remoteSections_` we could use an (aliasing) `std::weak_ptr` to quickly check whether things are outdated. @iyastreb WDYT?

jjthomas: would this be much faster than the current check? seems you still need to do a `remoteSections_.find` in any case and the generation value fetches are basically free. seems like a more complex refactor better suited for a separate PR as well.

iyastreb: Do I get it right that we keep weak_ptr in the handle and therefore avoid the need to hash lookup?
To me it seems like a good idea!

jjthomas: should we do it in another PR since we need to add the `shared_ptr` wrapper to `remoteSections_` and change all users of it?

ColinNV: @iyastreb Yes, that's the direction I was thinking.

@jjthomas Might be a good idea to do separately. And perhaps the same principle can be applied in other places too.

ColinNV (approved):

src/core/nixl_agent.cpp
brminich: sounds good for another PR, @jjthomas do you have cycles for that? Would be good to have a unit test in that other PR also

brminich (approved):

src/core/nixl_agent.cpp
jjthomas: happy to take that up in the next week or so

ColinNV:
This PR fixes `estimateXferCost()` and `postXferReq()`, wouldn't `makeXferReq()` and `getXferStatus()` need the same generation check? Not sure about `releaseXferReq()`.

jjthomas:
@ColinNV I believe `getXferStatus` and `releaseXferReq` don't look at any agent-owned data that could be invalid. This problem does exist for `makeXferReq` but is a little more complicated because the object in question is a `nixlDlistH` --which doesn't have `remoteGeneration_` -- instead of a `nixlXferReqH`. So let me handle this case in the follow-up PR where I can add the `weak_ptr` to both `nixlDlistH` and `nixlXferReqH`.

PR #1912, opened 2026-07-08
CI: automate release wheel publishing (poller + nightly job reuse)

Alexey-Rivkin:
## What?

- `nixl-build-wheel-release-poller` (new): every 4 hours, finds release-branch commits with no published wheels in Artifactory and triggers `nixl-ci-build-wheel-nightly` per missing CUDA variant, with `PUBLISH_DIR=release/<ver>`.
- The nightly job itself is untouched here - its side of this work merged separately in #2151 and #2166.

Published wheels: https://artifactory.nvidia.com/artifactory/sw-nbu-swx-nixl-pypi-local/release/

## Why?

Release wheels are published manually today. With this, every merge to a release branch publishes automatically (HPCINFRA-4573).

## How?

- Fully dynamic: scans all `release/*` branches with version >= 1.4.0, auto-detected, no per-release config.
- Triggers with `NIXL_VERSION`, `CUDA_MAJOR`, `PUBLISH_DIR` and `ci_refspec` only. UCX and the bundled plugins are not passed - each release ships what its own `contrib/build-container.sh` pins.
- Skips a release whose `build-container.sh` predates `--build-options-file`, since the nightly always passes it and older scripts reject unknown options. Release branches cut after #2166 pass on their own.
- No marker files: wheel presence per CUDA variant decides. A failed build retries next cycle; delete the folder in Artifactory to force a rebuild.
- Pre-merge e2e testing: run the poller with `ci_refspec=refs/pull/<n>/head` - it forwards the ref to the builds it triggers.

## Summary by CodeRabbit

* **New Features**
  * Added a scheduled release poller that scans release branches for missing CUDA wheel artifacts and triggers builds only for the specific missing variant sets.
  * Nightly wheel verification now supports a separate source reference (`ci_refspec`) and a configurable publish subdirectory (`PUBLISH_DIR`, default `verification`), with updated Artifactory upload paths for release publishing.
  * Optional UCX SP-CX plugin support can be enabled for wheel builds.
* **Documentation**
  * Updated CI job inventory and wheel-build documentation to reflect the new release polling, gating, retry/partial-upload behavior, and publishing-folder rules.

PR #2030, opened 2026-08-02
ci: fix CUDA version in the nightly wheel build

Alexey-Rivkin:
## What?
- Nightly wheel build: derive `CUDA_VERSION` from `BASE_TAG` and pass `--cuda-version`.
- Move the nightly job to the same nvcr.io ubi8 CUDA base as the per-PR job (#2009).

## Why?
Since #2009 `CUDA_VERSION` is a Dockerfile ARG defaulting to 13.0, and the nightly never passes it. The cu12 nightly silently builds as cu13 since 2026-07-30: no `nixl` meta wheel published, torch `+cu130` in a 12.9 build.

## How?
`BASE_TAG` is the single source of truth - parse its leading `MAJOR.MINOR`, fail loudly otherwise. No new job param to drift.

Verification (throwaway Devops job, one cu12 cell) in progress.


Alexey-Rivkin:
Folded into #1912 - both PRs touch the same nightly wheel files, and the fix needs #2009 which #1912 has now merged in.

PR #2036, opened 2026-08-03
ci: build python bindings without pip build isolation

NirWolfer:
## Summary

pip provisions `[build-system].requires` into a fresh PEP 517 build environment on every run. The pinned `torch==2.11.*` can never be satisfied by the torch already in the CI images (an NVIDIA nightly, e.g. `2.13.0a0+8145d630e8.nv26.6`), so pip re-downloaded torch plus its CUDA dependency closure - several GB - on each invocation.

That step normally completes in ~2 min, but stalls for 10-30 min when the index or network degrades, tripping the stage timeout and aborting with exit 143. It caused 10 CI failures between Jul 1 and Aug 3 across `nixl-ci-gpu`, `nixl-ci-non-gpu` and `nixl-ci-dl-gpu`.

CI does not need an isolated build environment, so this builds with `--no-build-isolation` and provisions the build dependencies explicitly. **`[build-system].requires` is left untouched**, so the isolated-build contract for consumers building from source - including `nixl_ep`'s torch requirement - is unchanged.

## Changes

`.gitlab/test_python.sh` only:

- add `--no-build-isolation` to the `pip install .` invocation
- install the build dependencies with `--upgrade`, in a single branch

## Why `--upgrade` matters

Without it pip treats an already-present version as satisfying the requirement and leaves it in place. The images ship **meson 1.3.2**, and that is what the failing runs actually used:

```
Version: 1.3.2
meson setup ... --native-file=.../meson-python-native-file.ini
meson-python: error: Could not map installation path to an equivalent wheel directory: '{prefix}/lib/libcore.so'
```

`contrib/Dockerfile` already installs these same packages with `--upgrade`, so this brings the CI script in line with it.

The two branches previously diverged - the venv branch installed the full set while the system-python branch installed only `tomlkit` and relied on the image for the rest. Both now install the same set, so the build no longer depends on image contents.

## Verification

`bash -n` passes. The stall is environment-dependent and not locally reproducible, so confirmation is a green CI run - specifically that the isolated build environment is no longer provisioned, and that with an upgraded meson the wheel builds without the `{prefix}/lib` mapping error.

## Summary by CodeRabbit

* **Build and Installation**
  * Build dependencies are now fully installed and upgraded consistently.
  * Package builds use preinstalled build dependencies for more predictable results.

### DIFF
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -44,14 +44,9 @@ then
     # Raise exceptions for logging errors
     export NIXL_DEBUG_LOGGING=yes
 
-    # Install build dependencies
-    if [ -n "$VIRTUAL_ENV" ] ; then
-        # Install full build dependencies in venv
-        $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
-    else
-        # Install minimal build dependencies in system python
-        $pip3 install --break-system-packages tomlkit
-    fi
+    # The install below uses --no-build-isolation, so pip does not provision
+    # [build-system].requires itself.
+    $pip3 install --break-system-packages --upgrade meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
     # Set the correct wheel name based on the CUDA version
     cuda_major=$(nvcc --version | grep -oP 'release \K[0-9]+')
     case "$cuda_major" in
@@ -60,7 +55,7 @@ then
     esac
     ./contrib/tomlutil.py --wheel-name "nixl-cu${cuda_major}" pyproject.toml
     # Control ninja parallelism during pip build to prevent OOM (NPROC from common.sh)
-    $pip3 install --break-system-packages --config-settings=compile-args="-j${NPROC}" .
+    $pip3 install --break-system-packages --no-build-isolation --config-settings=compile-args="-j${NPROC}" .
     $pip3 install --break-system-packages dist/nixl-*none-any.whl
 fi
 
--- a/.gitlab/test_python.sh
+++ b/.gitlab/test_python.sh
@@ -46,7 +46,8 @@ then
 
     # The install below uses --no-build-isolation, so pip does not provision
     # [build-system].requires itself.
-    $pip3 install --break-system-packages --upgrade meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
+    # --ignore-installed: the distro-packaged copies have no pip RECORD and cannot be uninstalled.
+    $pip3 install --break-system-packages --upgrade --ignore-installed meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
     # Set the correct wheel name based on the CUDA version
     cuda_major=$(nvcc --version | grep -oP 'release \K[0-9]+')
     case "$cuda_major" in


pyproject.toml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 [build-system]
-requires = ["meson-python", "pybind11", "patchelf", "pyyaml", "types-PyYAML", "pytest", "build", "setuptools>=80.9.0", "torch==2.11.*"]
+requires = ["meson-python", "pybind11", "patchelf", "pyyaml", "types-PyYAML", "pytest", "build", "setuptools>=80.9.0"]
```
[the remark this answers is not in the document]
NirWolfer: Correct, and acknowledged as a known limitation: after this change an isolated `-Dbuild_nixl_ep=true` build finds no torch and skips the EP extension.

**No CI impact.** Every in-tree path that builds `nixl_ep` provides torch itself and does not rely on the PEP 517 contract:

| Path | How torch is provided |
|---|---|
| `contrib/build-wheel.sh` | per-matrix torch installed into a venv, built with `--no-build-isolation` |
| `contrib/Dockerfile` | container's own torch, also `--no-build-isolation` |
| `contrib/Dockerfile.manylinux` | delegates to `contrib/build-wheel.sh` |
| `.gitlab/build.sh` | plain `meson setup`, not a pip build |

On the first suggested remedy — keeping a torch build requirement, e.g. relaxing the pin to `>=2.11` — that does not work, and I verified it rather than assuming. pip's build isolation never reuses the outer environment, even when the requirement is already satisfied there:

```
$ python3 -c "import packaging; print(packaging.__version__)"
25.0
$ python3 -m pip wheel . --no-deps -w out -v      # requires = ["setuptools", "packaging>=20"]
  Collecting packaging>=20
    Using cached packaging-26.2-py3-none-any.whl
  Installing collected packages: setuptools, packaging
  Successfully installed packaging-26.2
```

`25.0` satisfies `>=20` and was ignored. So any torch specifier still triggers a multi-GB download, and — more importantly for EP — it installs a torch that is not the one present at runtime. `nixl_ep` compiles a torch-ABI-versioned extension (`nixl_ep_cpp_torch<ver>`), so building it against an isolated torch would produce an extension mismatched with the user's actual torch. The previous `torch==2.11.*` pin was not delivering a correct ABI either.

That leaves the second suggestion — failing loudly instead of skipping silently. `examples/device/meson.build` only enters `ep/meson.build` when `build_nixl_ep` is explicitly true, so the current `warning()` + `subdir_done()` silently drops something the user explicitly requested. Turning it into an `error()` is the right behaviour, but it is a breaking change for external isolated EP builds, so I am getting sign-off from the maintainer who owns the EP build before making it. Tracking separately from this PR.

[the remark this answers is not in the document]
ovidiusm: @coderabbitai This learning is incorrect and should be removed. The wheel build is using isolated build env on purpose, and removing torch as a requirement would result in potentially broken builds published to the customers.

CI does not need an isolated env, I suggest investigating why the build is broken with that flag

ovidiusm (changes requested):
This is incorrect, please do not merge

ovidiusm:
```
# Install build dependencies
    if [ -n "$VIRTUAL_ENV" ] ; then
        # Install full build dependencies in venv
        $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
    else
        # Install minimal build dependencies in system python
        $pip3 install --break-system-packages tomlkit
    fi
```

To make the build work without build isolation in CI, you need to install manually the build deps so modify `$pip3 install --break-system-packages tomlkit`

I don't remember exactly why this command is different from the one above, but keep in mind that we need to make both CI pass this and the verification team scripts pass the run @Bohatchuk

pyproject.toml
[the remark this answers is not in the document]
NirWolfer: Agreed — reverted. `[build-system].requires` is untouched, so the isolated-build contract for source consumers, including torch for `nixl_ep`, is exactly as before. Back to `--no-build-isolation` in CI with the underlying blocker fixed.

**Why the flag broke the build.** Not a missing build dependency — without isolation meson-python builds the wheel in-tree and aborts on any installed file it cannot place in the wheel layout:

```
meson-python: error: Could not map installation path to an equivalent wheel directory: '{prefix}/lib/libcore.so'
```

The prometheus-cpp CMake subproject's libraries carry install rules resolving to a bare `{prefix}/lib`, which has no wheel equivalent. Correlation was exact: 4/4 configs that built prometheus failed, 2/2 that did not passed, and the C++ build completed all 156 targets — purely a packaging error.

Those libraries are only ever linked into `libtelemetry_exporter_prometheus.so`, so they do not need to be installed separately. `cmake_opts.set_install(false)` overrides the install attribute meson derives from the CMake targets:

```python
# mesonbuild/cmake/interpreter.py
install_tgt = options.get_install(tgt.cmake_name, tgt.install)
```

so nothing from the subproject reaches the install plan. Static is required alongside it — with shared libs suppressed from the install plan the plugin would link a `.so` that never gets installed; static links the symbols in and leaves the plugin self-contained. `OFF` is prometheus-cpp's upstream default and `CMAKE_POSITION_INDEPENDENT_CODE` was already `ON`. `set_install` exists since meson 0.55.0, below our `>= 0.64.0` floor.

**On the build deps** — good catch, done. The system-python branch installed only `tomlkit` and relied on the image for meson/meson-python/pybind11/patchelf/pyYAML/setuptools; both branches now install the same set, so the build no longer depends on image contents. That should also cover the verification team scripts, since nothing is assumed about the environment beyond pip itself. @Bohatchuk worth a look.

One thing I have not been able to verify locally (no CUDA/UCX here): that `prometheus_sub.dependency('core')` resolves against the static targets and that bundled civetweb links cleanly. CI will confirm.

.gitlab/test_python.sh
```diff
@@ -44,14 +44,10 @@ then
...
-        # Install full build dependencies in venv
-        $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
-    else
-        # Install minimal build dependencies in system python
-        $pip3 install --break-system-packages tomlkit
-    fi
+    # Install build dependencies. The pip install below runs with
+    # --no-build-isolation, so pip does not provision [build-system].requires
+    # itself and everything the build backend needs must already be present.
+    $pip3 install --break-system-packages meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit 'setuptools>=80.9.0'
```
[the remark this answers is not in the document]
NirWolfer: Checked each of the three; none are needed to build, so I am leaving the list as is.

- **`types-PyYAML`** is a stub-only distribution. It ships `.pyi` files for type checkers and is never imported at build time.
- **`pytest`** is a test framework, not a build input. The script installs it after the build, alongside `pytest-timeout` and `zmq`, which is the right place for it.
- **`build`** is a PEP 517 *frontend*. The backend here is `mesonpy`, invoked by pip directly, so it is never used.

The repo already demonstrates this exact set is sufficient. `contrib/Dockerfile` installs

```
meson meson-python pybind11 patchelf pyYAML click tabulate auditwheel tomlkit   # :270
setuptools>=80.9.0                                                              # :280
```

and then builds this same source with `--no-build-isolation` at `:323`. No `types-PyYAML`, no `pytest`, no `build` — and that path has been green. Grepping the build tree for `import yaml`, `import pytest` or `import build` in `meson.build` and the `contrib/` helpers also returns nothing.

On **torch**: it is not required here because `examples/device/meson.build` gates `subdir('ep')` on `build_nixl_ep`, which defaults to false and is not set by this script, so the torch check in `examples/device/ep/meson.build` is never reached. The test images do ship torch, but the build does not consume it. `[build-system].requires` keeps `torch==2.11.*` for isolated source builds, which this PR deliberately leaves untouched.

The underlying observation is fair though — hand-provisioning build deps can drift from `[build-system].requires`. The mismatch here predates this PR: `pytest` and `build` are listed as build requirements without being build inputs. Worth tidying, but as its own change rather than by copying test-only and stub-only packages into the CI install.

src/plugins/telemetry/prometheus/meson.build
```diff
@@ -26,13 +26,18 @@ cmake_opts.add_cmake_defines({
     'ENABLE_COMPRESSION': 'OFF',
     'ENABLE_TESTING': 'OFF',
     'USE_THIRDPARTY_LIBRARIES': 'ON',
-    'BUILD_SHARED_LIBS': 'ON',
```
ovidiusm, line 29: Is this needed?

src/plugins/telemetry/prometheus/meson.build
```diff
@@ -26,13 +26,18 @@ cmake_opts.add_cmake_defines({
...
+    'BUILD_SHARED_LIBS': 'OFF',
     'CMAKE_POSITION_INDEPENDENT_CODE': 'ON',
     'CMAKE_LINK_DEPENDS_USE_LINKER': 'OFF',
     'CMAKE_C_FLAGS': '-Wno-unused-but-set-variable',
     'CMAKE_CXX_FLAGS': '-Wno-unused-but-set-variable'
 })
 
+# The subproject's libraries are linked into the plugin, so they must not be
+# installed themselves: meson would place them in a bare {prefix}/lib, which
+# meson-python cannot map to a wheel directory.
```
ovidiusm: You are missing the --upgrade flag in the pip install command, then the changes in this file can be reverted

.gitlab/test_python.sh
ovidiusm: install --upgrade ?

src/plugins/telemetry/prometheus/meson.build
NirWolfer: No, reverted. You were right that this was a stale toolchain rather than the subproject's install rules — `src/plugins/telemetry/prometheus/meson.build` is now untouched by this PR.

Details in the `test_python.sh` thread.

src/plugins/telemetry/prometheus/meson.build
NirWolfer: Confirmed, you were right. The images ship meson 1.3.2 and that is what the failing runs actually used:

```
Version: 1.3.2
meson setup ... --native-file=.../meson-python-native-file.ini
meson-python: error: Could not map installation path to an equivalent wheel directory: '{prefix}/lib/libcore.so'
```

Without `--upgrade`, pip treated the already-present meson as satisfying the requirement and left it in place, so the build kept running against 1.3.2. With the upgrade in place the mapping error is gone and CI is green, so this file is fully reverted — the PR now touches only `.gitlab/test_python.sh`.

One additional flag was needed to make `--upgrade` work on the Python 3.12 images; see the `test_python.sh` thread.

.gitlab/test_python.sh
NirWolfer: Added, and it needed `--ignore-installed` alongside it to work.

`--upgrade` on its own failed on the Python 3.12 images: they carry a Debian-packaged PyYAML 6.0.1 with no pip RECORD, and the newer pip there treats the uninstall as fatal rather than warning:

```
ERROR: Cannot uninstall PyYAML 6.0.1, RECORD file not found.
error: uninstall-no-record-file
```

Only the 3.12 images (cuda12.9-ubuntu24.04, pytorch26.06, both arches) hit it; the ubuntu22.04 / Python 3.10 ones passed. `--ignore-installed` skips the uninstall path entirely and installs into `dist-packages`, which precedes the distro location on `sys.path`. Restricting the flag to PyYAML would only have moved the failure one package along — meson is exposed the same way, since Ubuntu 24.04 ships exactly the 1.3.2 we were picking up.

I also folded the two branches into a single install. The system-python branch previously installed only `tomlkit` and relied on the image for meson, meson-python, pybind11, patchelf, pyYAML and setuptools, which is what let the stale meson through in the first place. Both paths now install the same set, so the build no longer depends on image contents — this should also cover the verification team scripts you mentioned, since nothing is assumed about the environment beyond pip itself.

All CI pipelines pass. The PR is now a single file: `--no-build-isolation` on the build, plus `--upgrade --ignore-installed` on the dependency install. `[build-system].requires` is untouched, so the isolated-build contract for source consumers is unchanged.

ovidiusm (approved):

PR #2038, opened 2026-08-04
libfabric: add tracing via nvtx backend

fengjica:
## What?
This PR tries to add tracing in libfabric backend plugin via the new nixl tracing API and nvtx tracing backend.

## How?
Add tracing instrumentation in code.
Changed nixl core code to make tracing facade available to a backend to use.

## Summary by CodeRabbit

* **New Features**
  * Added tracing across Libfabric transfers, communication, completions, retries, and deferred operations.
  * Added correlated transfer spans with device, size, rail, striping, and request metadata.
  * Added on-demand tracing backend loading with graceful handling of unavailable or failed backends.
  * Added NVTX events for Libfabric transfer lifecycle operations.

* **Bug Fixes**
  * Request device identifiers are now preserved and reset appropriately during request processing.

* **Documentation**
  * Documented Libfabric tracing spans, markers, attributes, and transfer correlation.

### DIFF
--- a/src/core/nixl_agent.cpp
+++ b/src/core/nixl_agent.cpp
@@ -380,6 +380,7 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     init_params.pthrDelay = data->config_.pthrDelay;
     init_params.syncMode = data->config_.syncMode;
     init_params.enableTelemetry_ = (data->telemetry_ != nullptr);
+    init_params.tracer_ = data->tracer_.get();
 
     // First, try to load the backend as a plugin
     auto& plugin_manager = nixlPluginManager::getInstance();
--- a/src/plugins/libfabric/meson.build
+++ b/src/plugins/libfabric/meson.build
@@ -28,6 +28,12 @@ libfabric_plugin_deps = [
 
 compile_flags = ['-DHAVE_LIBFABRIC']
 
+# The nixl::trace facade method bodies (Span/Tracer) live in nixl core, which this
+# plugin cannot link (core loads plugins at runtime).
+# Here, compile those method bodies straight into the plugin so the symbols resolve
+# under -Wl,--no-undefined, when it is not staticaly built.
+nixl_trace_facade_src = files('../../core/tracing/tracer.cpp')
+
 # Add CUDA support if available
 if cuda_dep.found()
     libfabric_plugin_deps += [cuda_dep]
@@ -59,6 +65,7 @@ else
         'libfabric_connection.h',
         'libfabric_handshake.cpp',
         'libfabric_plugin.cpp',
+        nixl_trace_facade_src,
         dependencies: libfabric_plugin_deps,
         include_directories: [nixl_inc_dirs, utils_inc_dirs],
         install: true,
--- a/src/plugins/tracing/nvtx/nvtx_events.cpp
+++ b/src/plugins/tracing/nvtx/nvtx_events.cpp
@@ -38,6 +38,16 @@ namespace {
         "nixl::fetchRemoteMD",
         "nixl::prepMemView",
         "nixl::releaseMemView",
+        // libfabric backend (src/utils/libfabric)
+        "nixl::libfabric.transfer.write",
+        "nixl::libfabric.transfer.read",
+        "nixl::libfabric.post_write",
+        "nixl::libfabric.post_read",
+        "nixl::libfabric.post_send",
+        "nixl::libfabric.local_completion.write",
+        "nixl::libfabric.local_completion.read",
+        "nixl::libfabric.recv_completion",
+        "nixl::libfabric.remote_write_completion",
     };
 
     [[nodiscard]] constexpr std::uint32_t
--- a/src/plugins/tracing/nvtx/nvtx_events.cpp
+++ b/src/plugins/tracing/nvtx/nvtx_events.cpp
@@ -44,6 +44,8 @@ namespace {
         "nixl::libfabric.post_write",
         "nixl::libfabric.post_read",
         "nixl::libfabric.post_send",
+        "nixl::libfabric.post_deferred.write",
+        "nixl::libfabric.post_deferred.read",
         "nixl::libfabric.local_completion.write",
         "nixl::libfabric.local_completion.read",
         "nixl::libfabric.recv_completion",
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -20,6 +20,8 @@
 #include "common/nixl_log.h"
 #include "serdes/serdes.h"
 #include "libfabric_common.h"
+#include "tracing/trace.h"
+#include "tracing/trace_macros.h"
 
 #ifdef HAVE_CUDA
 #include <cuda_runtime.h>
@@ -75,6 +77,7 @@ RequestPool::release(nixlLibfabricReq *req) const {
 
     req->in_use = false;
     req->xfer_id = 0;
+    req->device_id = -1;
     req->chunk_offset = 0;
     req->chunk_size = 0;
     req->completion_callback = nullptr;
@@ -391,10 +394,12 @@ DataRequestPool::allocate(nixlLibfabricReq::OpType op_type, uint32_t req_id) {
 
 nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
                                      const std::string &provider,
-                                     uint16_t id)
+                                     uint16_t id,
+                                     nixl::trace::Tracer *tracer)
     : rail_id(id),
       device_name(device),
       provider_name(provider),
+      tracer_(tracer),
       control_request_pool_(NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL, id),
       data_request_pool_(NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL, id),
       provider_supports_hmem_(false) {
@@ -910,6 +915,13 @@ nixlLibfabricRail::processLocalTransferCompletion(struct fi_cq_data_entry *comp,
     // Find the request from context to access the completion callback
     nixlLibfabricReq *req = findRequestFromContext(comp->op_context);
     if (req && req->in_use) { // Only process if request is still valid and in use
+        NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(req->xfer_id));
+        NIXL_TRACE_MARK(tracer_,
+                        req->operation_type == nixlLibfabricReq::WRITE ?
+                            "nixl::libfabric.local_completion.write" :
+                            "nixl::libfabric.local_completion.read",
+                        nixl::trace::Kind::Metadata);
+
         // Call completion callback if it exists
         if (req->completion_callback) {
             NIXL_TRACE << "Calling completion callback for " << operation_type << " request "
@@ -943,6 +955,9 @@ nixlLibfabricRail::processRecvCompletion(struct fi_cq_data_entry *comp) const {
     uint64_t msg_type = NIXL_GET_MSG_TYPE_FROM_IMM(comp->data);
     uint16_t agent_idx = NIXL_GET_AGENT_INDEX_FROM_IMM(comp->data);
     uint32_t xfer_id = NIXL_GET_XFER_ID_FROM_IMM(comp->data);
+
+    NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(xfer_id));
+    NIXL_TRACE_MARK(tracer_, "nixl::libfabric.recv_completion", nixl::trace::Kind::CommRecv);
     NIXL_TRACE << "Received control message type " << msg_type << " agent_idx=" << agent_idx
                << " XFER_ID=" << xfer_id << " imm_data=" << std::hex << comp->data << std::dec;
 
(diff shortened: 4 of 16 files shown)


src/core/nixl_agent.cpp
```diff
@@ -380,6 +380,7 @@ nixlAgent::createBackend(const nixl_backend_t &type,
     init_params.pthrDelay = data->config_.pthrDelay;
     init_params.syncMode = data->config_.syncMode;
     init_params.enableTelemetry_ = (data->telemetry_ != nullptr);
+    init_params.tracer_ = data->tracer_.get();
```
e-eygin, line 383: **Use-after-free at shutdown.**

`nixlAgentData` declares `tracer_` *after* `backendEngines_`, inside the block marked "The order of the following data members is crucial for destruction". Members destruct in reverse order, so `tracer_` is freed first — while the libfabric engine still holds this raw pointer and its progress thread (only stopped and joined in `~nixlLibfabricEngine`, i.e. during `backendEngines_` destruction) is still calling `NIXL_TRACE_MARK(tracer_, ...)` from `processLocalTransferCompletion()` and `processRecvCompletion()`.

The ordering predates this PR but was harmless: every prior use of `tracer_` was on a user-facing agent method, and `commThread` is joined in the body of `~nixlAgent` before `data` is destroyed. This PR is the first consumer that outlives it.

Please move `tracer_` above `backendHandles_`/`backendEngines_` in `agent_data.h`, and extend that ordering comment to mention that backends now hold a borrowed pointer.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1087,6 +1107,15 @@ nixlLibfabricRail::postSend(uint64_t immediate_data, fi_addr_t dest_addr, nixlLi
                << " XFER_ID=" << NIXL_GET_XFER_ID_FROM_IMM(immediate_data)
                << " dest_addr=" << dest_addr << std::dec << " context=" << &req->ctx;
 
+    // Tracing: span covers the (possibly retried) control-message send. xfer_id
+    // is carried in the immediate data on this path; correlate on it.
+    const std::uint64_t send_xfer_id = NIXL_GET_XFER_ID_FROM_IMM(immediate_data);
```
e-eygin: **This breaks `-Dwith_trace=false`.**

`send_xfer_id` is consumed only by `NIXL_TRACE_CORRELATION_SCOPE` and `NIXL_TRACE_ATTR`, both of which expand to `do {} while (0)` when `NIXL_TRACE_ENABLED` is undefined. The project sets `werror=true`, so this is `-Werror=unused-variable` (verified with the repo's gcc settings).

Inline the expression into the macro arguments, or mark it `[[maybe_unused]]`.

Since `with_trace` defaults to `true`, CI won't catch this — it only bites whoever turns the facade off. Worth adding a `-Dwith_trace=false` configure to CI. Related: a clang build of that configuration will also trip `-Wunused-private-field` on `nixlLibfabricRail::tracer_` (L602), since every use of it compiles out.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -403,6 +407,23 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
 
     // Determine striping strategy
     bool use_striping = usesStriping(transfer_size, selected_rails.size());
+
+    // Tracing: one span covers the prepare+submit for this descriptor. Correlate
+    // on xfer_id so it links to the per-rail post spans and the completion marker
+    // (which may fire on the progress thread). The RAII span ends on any return.
+    NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(xfer_id));
+    NIXL_TRACE_SCOPE(trace_span,
```
e-eygin, line 415: **These spans don't fire in progress-thread mode.**

`postWrite`/`postRead` are reachable only from the `!isProgressThreadEnabled()` branches (L450 and L547). With PT on, submission goes to `deferTransferRequest()` and the actual post happens in `nixlLibfabricRail::drainPostQueue()` via `fi_writemsg`/`fi_readmsg`, which is uninstrumented.

So in PT mode `nixl::libfabric.post_write` / `post_read` never appear at all, and this enclosing `transfer.write` span measures ring-buffer enqueue latency rather than post latency — which inverts what the trace appears to say, in exactly the mode you'd most want to profile.

`drainPostQueue` is a rail method and already has `tracer_` in scope; a span there correlated on `pr.req->xfer_id`, plus a mark at enqueue time, would close the gap. If you'd rather defer that, please state the limitation in `docs/tracing.md`.

src/plugins/libfabric/meson.build
```diff
@@ -28,6 +28,12 @@ libfabric_plugin_deps = [
 
 compile_flags = ['-DHAVE_LIBFABRIC']
 
+# The nixl::trace facade method bodies (Span/Tracer) live in nixl core, which this
+# plugin cannot link (core loads plugins at runtime).
+# Here, compile those method bodies straight into the plugin so the symbols resolve
+# under -Wl,--no-undefined, when it is not staticaly built.
+nixl_trace_facade_src = files('../../core/tracing/tracer.cpp')
```
e-eygin, line 35: **The facade duplication is the wrong shape of fix for a real problem.**

Compiling `tracer.cpp` into the plugin gives it a second machine-code copy of the facade that reaches directly into `Tracer`'s private `std::vector` member. Plugins are `dlopen`ed with `RTLD_LOCAL` (sometimes `RTLD_DEEPBIND`), so the plugin always binds its own copy. That is correct today *only* because `tracer.cpp` happens to have zero static state — the first person to add a function-local `static`, a `thread_local`, or a cache there gets two silently independent copies with no diagnostic.

`Span` and `Tracer` methods are all short loops over a vector. Marking them `inline` in `trace.h` and leaving only `makeTracer()` in libnixl removes the link problem outright: plugins need no symbols, no meson change, no duplication (inline functions are ODR-mergeable by construction, unlike whole-TU duplication), and it drops a cross-DSO call from the data path.

Alternatives, in descending preference: give `Tracer` pure-virtual methods so plugins only need the header; or, if duplication is really wanted, express it once as `static_library('nixl_trace_facade', pic: true)` + `declare_dependency()` exported from `src/core/tracing/meson.build`.

Either way this feels like it belongs in the tracing infrastructure PR rather than a backend PR — UCX will hit the exact same wall next.

test/unit/utils/libfabric/meson.build
```diff
@@ -22,6 +22,8 @@ if cuda_dep.found()
     libfabric_test_cpp_args += [ '-DCUDA_FOUND' ]
 endif
 
+nixl_trace_facade_src = files('../../../../src/core/tracing/tracer.cpp')
```
e-eygin, line 25: Second copy of the same source, with a second hand-written relative path (`../../../../`). The plugin has `../../` at `src/plugins/libfabric/meson.build:35`; a third consumer will copy it again, and any file move breaks all of them silently. See the note there — exporting the facade once as a dependency (or making it header-only) avoids this entirely.

src/plugins/tracing/nvtx/nvtx_events.cpp
```diff
@@ -38,6 +38,16 @@ namespace {
...
+        // libfabric backend (src/utils/libfabric)
+        "nixl::libfabric.transfer.write",
+        "nixl::libfabric.transfer.read",
+        "nixl::libfabric.post_write",
+        "nixl::libfabric.post_read",
+        "nixl::libfabric.post_send",
+        "nixl::libfabric.local_completion.write",
+        "nixl::libfabric.local_completion.read",
+        "nixl::libfabric.recv_completion",
+        "nixl::libfabric.remote_write_completion",
```
e-eygin, line 52: **Layering:** a transport-agnostic tracing backend now hardcodes libfabric span names, which means every future instrumented backend has to edit this plugin and keep it in sync.

**Cost:** `lookupRegistered()` is a linear `string_view` comparison run on every span *and* every mark — the table is now 23 entries deep, walked on the data path.

`eventForName()` already registers unknown names on demand, so this addition isn't needed for correctness. Either drop it, or replace the array with a lookup cache so the list can grow without a per-event cost.

src/utils/libfabric/libfabric_rail_manager.cpp
```diff
@@ -403,6 +407,23 @@ nixlLibfabricRailManager::prepareAndSubmitTransfer(
 
     // Determine striping strategy
     bool use_striping = usesStriping(transfer_size, selected_rails.size());
+
+    // Tracing: one span covers the prepare+submit for this descriptor. Correlate
+    // on xfer_id so it links to the per-rail post spans and the completion marker
+    // (which may fire on the progress thread). The RAII span ends on any return.
+    NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(xfer_id));
```
e-eygin, line 414: `xfer_id` is `uint16_t` and recycled, so over any real profiling run correlation ids alias and a viewer will visually join unrelated transfers. Consider a monotonically increasing 64-bit id, or at minimum document the caveat.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1305,6 +1335,16 @@ nixlLibfabricRail::postWrite(const void *local_buffer,
...
 
+    // Tracing: span covers the (possibly retried) post; correlate on xfer_id to
+    // link with the enclosing transfer span and the completion marker.
+    NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(req->xfer_id));
+    NIXL_TRACE_SCOPE(
+        trace_span, tracer_, "nixl::libfabric.post_write", nixl::trace::Kind::CommSend);
+    NIXL_TRACE_ATTR(trace_span, "device_id", static_cast<std::int64_t>(req->device_id));
+    NIXL_TRACE_ATTR(trace_span, "rail_id", static_cast<std::int64_t>(rail_id));
+    NIXL_TRACE_ATTR(trace_span, "length", static_cast<std::int64_t>(length));
+    NIXL_TRACE_ATTR(trace_span, "xfer_id", static_cast<std::int64_t>(req->xfer_id));
```
e-eygin, line 1373: Redundant push — `prepareAndSubmitTransfer` already pushed this same `xfer_id` before calling in, so the stack carries it twice for no benefit.

More generally, worth a note on cost: `prepareAndSubmitTransfer` runs once per descriptor and each span heap-allocates a vector plus an `NvtxSpan`, so a 1000-descriptor batch becomes thousands of allocations plus correlation push/pop pairs. Some measured overhead with NVTX active would make this much easier to sign off on.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -943,6 +955,9 @@ nixlLibfabricRail::processRecvCompletion(struct fi_cq_data_entry *comp) const {
     uint64_t msg_type = NIXL_GET_MSG_TYPE_FROM_IMM(comp->data);
     uint16_t agent_idx = NIXL_GET_AGENT_INDEX_FROM_IMM(comp->data);
     uint32_t xfer_id = NIXL_GET_XFER_ID_FROM_IMM(comp->data);
+
+    NIXL_TRACE_CORRELATION_SCOPE(tracer_, static_cast<std::uint64_t>(xfer_id));
+    NIXL_TRACE_MARK(tracer_, "nixl::libfabric.recv_completion", nixl::trace::Kind::CommRecv);
```
e-eygin, line 960: The correlation scope covers the whole function body, including the handshake and connection callbacks, so unrelated nested spans inherit this id — and for non-transfer control messages the id isn't meaningful in the first place.

`processRemoteWriteCompletion` gets this right (L1033) by pushing only inside the `msg_type == NIXL_LIBFABRIC_MSG_TRANSFER` branch. Same treatment here, or scope the guard tightly around the mark.

src/utils/libfabric/libfabric_rail.h
```diff
@@ -44,6 +44,7 @@ struct nixlLibfabricReq {
     size_t rail_id; ///< Rail ID that owns this request
     size_t pool_index; ///< Index in the pool for deque compatibility
     uint32_t xfer_id; ///< Pre-assigned globally unique transfer ID
+    int device_id; ///< Source device (GPU) index, for tracing (-1 when unknown)
```
e-eygin, line 47: This field exists solely to carry a trace attribute, but it grows a pooled hot-path struct and adds an unconditional store in `RequestPool::release()` (`libfabric_rail.cpp:80`) even when `with_trace=false`.

`nixlLibfabricPostRequest` already carries `device_id` (L391). Passing it as a `postWrite`/`postRead` parameter would avoid both costs.

src/utils/libfabric/libfabric_rail.h
```diff
@@ -400,7 +402,10 @@ class nixlLibfabricRail {
     struct fid_ep *endpoint; ///< Libfabric endpoint handle
 
     /** Initialize libfabric rail with all resources */
-    nixlLibfabricRail(const std::string &device, const std::string &provider, uint16_t id);
+    nixlLibfabricRail(const std::string &device,
+                      const std::string &provider,
+                      uint16_t id,
+                      nixl::trace::Tracer *tracer = nullptr);
```
e-eygin, line 408: This header names `nixl::trace::Tracer` here and at L602, but compiles only because `backend/backend_aux.h` transitively forward-declares it. Please declare it locally. Same in `libfabric_rail_manager.h:78`.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -1109,6 +1138,7 @@ nixlLibfabricRail::postSend(uint64_t immediate_data, fi_addr_t dest_addr, nixlLi
             NIXL_TRACE << "Send posted successfully"
                        << (attempt > 0 ? " after " + std::to_string(attempt + 1) + " attempts" :
                                          "");
+            NIXL_TRACE_ATTR(trace_span, "retries", static_cast<std::int64_t>(attempt));
```
e-eygin, line 1139: `retries` is recorded only on the success path, so a post that gives up after many `EAGAIN`s — the interesting case — carries no retry count. Same at L1384 and L1477.

e-eygin (commented):
Reviewed against the tracing facade, the libfabric submission paths, and the build wiring. The instrumentation choices are sensible and the correlation-id approach is right, but three things look blocking to me:

1. Use-after-free at shutdown (`src/core/nixl_agent.cpp:383`) — the agent's `Tracer` is destroyed before the backends that hold it.
2. `-Dwith_trace=false` no longer compiles (`src/utils/libfabric/libfabric_rail.cpp:1112`) — unused variable under `werror=true`.
3. The post spans never fire in progress-thread mode (`src/utils/libfabric/libfabric_rail_manager.cpp:415`) — `postWrite`/`postRead` are unreachable when PT is on.

Details inline, plus some design notes on the facade-duplication approach in meson and on the NVTX name table.

Two things that don't map to a line:

- `docs/tracing.md` keeps a table of span names/kinds/attributes (L127-128). The nine new spans should be added there, especially since the PR already touches the registered-name list.
- No tests. `test/gtest/unit/tracing/tracing_test.cpp` already provides `MockBackend`/`MockSpan`, and `test/unit/utils/libfabric/rail_active_refcount_test.cpp` already wraps the `fi_*` calls, so asserting that a submit path emits the expected span names and attributes into a mock tracer looks reachable without hardware.

Minor: the description is missing the `## Why?` section required by `.github/pull_request_template.md`, and there's no linked issue.

Issue #1983, opened 2026-07-22
UCX SRD backend: NIXL_ERR_REMOTE_DISCONNECT


jjthomas:
When using through sglang, seeing NIXL_ERR_REMOTE_DISCONNECT when a prefill container tries to connect with a newly started decode container via UCX SRD backend (Amazon EFA) on B200 instances. This error causes the remote agent to be deregistered and all subsequent transfers to that decode container to fail (even though the container is eventually reachable by this and other prefill containers).

Prefill log:

```
[2026-07-21 08:04:41] Processing ChatCompletion request a04107cc-ff53-49ba-b428-03eb77df5cd5 from workspace 6051921418418893, priority=REQUEST_PRIORITY_TYPE_DEFAULT, no images, has_json_schema=False, has_tools=False
[2026-07-21 08:04:41 TP0 EP0] Prefill batch, #new-seq: 1, #new-token: 64, #cached-token: 10048, token usage: 0.00, #running-req: 0, #queue-req: 0, #pending-token: 0, #bootstrap-req: 0, #inflight-req: 1, cuda graph: False, input throughput (token/s): 178.18, est. prefill TFLOPS/s (per GPU): 9.90, fwd occupancy: nan%
[1784621081.896649] [kimi-k2-7-disagg-on-sp-v1-prefill-92c6hc-07210529-x04q3qgn-pod:537  :1]       srd_iface.c:221  UCX  ERROR send completion[0] with error on rdmap132s0/0x55cf4425cfb0: Work Request Flushed Error, vendor_err 0x1 wr_id 0x7f8b20022480
[1784621081.896667] [kimi-k2-7-disagg-on-sp-v1-prefill-92c6hc-07210529-x04q3qgn-pod:532  :0]       srd_iface.c:221  UCX  ERROR send completion[0] with error on rdmap96s0/0x55f989790220: Work Request Flushed Error, vendor_err 0x1 wr_id 0x7f676c7db240
[1784621081.896648] [kimi-k2-7-disagg-on-sp-v1-prefill-92c6hc-07210529-x04q3qgn-pod:537  :0]       srd_iface.c:221  UCX  ERROR send completion[0] with error on rdmap133s0/0x55cf44182010: Work Request Flushed Error, vendor_err 0x1 wr_id 0x7f8b1c0f6b00
[...]

PR #2051, opened 2026-08-05
CORE: Own remote sections via shared_ptr; reject stale-generation prepped dlists in makeXferReq

jjthomas:
Follow-up to #2027 per the review discussion with @ColinNV / @brminich.

- `remoteSections_` now owns one `shared_ptr` per registration generation; `nixlXferReqH` and `nixlDlistH` carry weak refs to the section they were created from, so staleness is an O(1) `expired()` check instead of a name lookup + generation compare.
- `makeXferReq` rejects prepped remote dlists whose registration generation is gone (closes the `prepXferDlist` → `makeXferReq` window where stale remote keys would be bound into a request). `postXferReq`/`estimateXferCost` guards rewritten onto the weak refs; `getXferStatus` now also detects re-registration, not just plain invalidation. `getXferStatus`/`releaseXferReq` otherwise stay metadata-free by design: cached completion and teardown of stale handles keep working.
- New gtest suite `StaleGenerationTest` (2 agents, UCX loopback): fails before, passes after; rest of suite unchanged vs base.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of remote memory registrations across metadata reloads.
  * Stale transfer handles and descriptor lists are now detected and safely rejected.
  * Reloading unchanged metadata preserves existing handles and registrations.
  * Changed metadata invalidates outdated registrations and enables recovery with fresh handles.

* **Tests**
  * Added regression coverage for stale handles, descriptor lists, unchanged metadata reloads, and changed-metadata re-registration.


src/core/transfer_request.h
```diff
@@ -73,6 +76,9 @@ class nixlXferReqH {
     const std::string remoteAgent;
     // Generation of the remote-connection
     const uint64_t remoteGeneration_;
```
iyastreb: is remoteGeneration still needed?

src/core/transfer_request.h
```diff
@@ -42,7 +44,8 @@ class nixlXferReqH {
                  const nixl_mem_t local_type,
                  const nixl_mem_t remote_type,
                  const uint64_t remote_generation,
-                 const size_t desc_count);
+                 const size_t desc_count,
+                 const std::weak_ptr<nixlRemoteSection> &remote_section_ref);
```
iyastreb: minor maybe add `using` alias for this type?

src/core/agent_data.h
```diff
@@ -100,7 +100,10 @@ class nixlAgentData {
         std::unordered_map<nixl_backend_t, std::unique_ptr<nixlBackendH>> backendHandles_;
         std::unordered_map<nixl_backend_t, nixl_blob_t> connMd_;
         backend_map_t backendEngines_;
-        std::unordered_map<std::string, nixlRemoteSection> remoteSections_;
+        // Owning shared_ptr per registered remote generation: invalidating or re-registering
+        // an agent destroys the old section object, so weak pointers from transfer-request and
+        // prepped-dlist handles to it expire, flagging them as outdated without a map lookup.
```
iyastreb: maybe reduce this excessive comment

test/gtest/stale_generation.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "nixl.h"
+#include "common.h"
+
+namespace gtest {
+namespace stale_generation {
```
iyastreb: Maybe don't create a new test suite, but just extend existing error_handling one?
I think it covers most of these cases

src/core/transfer_request.h
ColinNV: I'm not sure that's a good idea, it obfuscates the code when it's full of `something_handle_t` and similar and it's not clear whether it's a shared pointer, unique pointer, plain pointer, or something else.

src/core/transfer_request.h
ColinNV: I would have also expected the generation to be superfluous.

src/core/transfer_request.h
jjthomas: see the approach i took where `weak` is in the name

src/core/transfer_request.h
jjthomas: removed

src/core/nixl_agent.cpp
```diff
@@ -100,9 +100,12 @@ nixlXferReqH::updateRequestStats(nixlTelemetry *telemetry_pub,
                << duration.count() << "us.";
 }
 
-nixlDlistH::nixlDlistH(const std::string &remote_agent, descs_t &&descs)
+nixlDlistH::nixlDlistH(const std::string &remote_agent,
+                       descs_t &&descs,
+                       const std::weak_ptr<nixlRemoteSection> &remote_section_ref)
```
iyastreb:
```suggestion
                       const nixl_remote_section_weak_t &remote_section_ref)
```

src/core/nixl_agent.cpp
```diff
@@ -1169,7 +1165,9 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
         if (req_hndl->status == NIXL_ERR_REMOTE_DISCONNECT) {
             read_lock.unlock();
             NIXL_LOCK_GUARD(data->lock);
-            data->invalidateRemoteData(req_hndl->remoteAgent, req_hndl->remoteGeneration_);
+            if (auto section = req_hndl->remoteSectionRef.lock()) {
+                data->invalidateRemoteData(req_hndl->remoteAgent, section->getGeneration());
```
iyastreb: What's the role of section->getGeneration() now?
Why it's needed?

src/core/agent_data.h
```diff
@@ -120,7 +122,7 @@ class nixlAgentData {
         nixl_status_t
         loadRemoteSections(const std::string &remote_name, nixlSerDes &sd);
         nixl_status_t
-        invalidateRemoteData(const std::string &remote_name, uint64_t generation);
+        invalidateRemoteData(const std::string &remote_name);
```
ColinNV: Should this receive the weak pointer instead of the generation so that something keyed to an earlier version doesn't invalidate a later one? @iyastreb

iyastreb: Right, this is a key point
I think we need to distinguish between 2 cases:
1) When user asks to invalidate - then we can remove by name (this is API)
2) When NIXL detects disconnected agent in status/post, we can either:
- (I think is preferred) in getStatus and post currently we remove it, but there is an open ticket to avoid doing that. So we can remove that drop from these 2 methods, but need to check the consequences
- If we want to keep dropping stale agents, then of course we can't do it by name

jjthomas: i'd add a +1 to not dropping stale agents, it would solve this issue for us: https://github.com/ai-dynamo/nixl/issues/1983

iyastreb: So let's do it

jjthomas: done

jjthomas: by the way, agents that have experienced a `NIXL_ERR_REMOTE_DISCONNECT` should still be usable later if the underlying connectivity issue has gone away, right? now that we've eliminated the drop

src/core/nixl_listener.cpp
```diff
@@ -779,8 +779,9 @@ nixlAgentData::loadConnInfo(const std::string &remote_name,
 
 nixl_status_t
 nixlAgentData::loadRemoteSections(const std::string &remote_name, nixlSerDes &sd) {
-    const auto [it, inserted] = remoteSections_.try_emplace(remote_name, remote_name);
-    const nixl_status_t ret = it->second.loadRemoteData(&sd, backendEngines_);
+    const auto [it, inserted] =
+        remoteSections_.try_emplace(remote_name, std::make_shared<nixlRemoteSection>(remote_name));
```
ColinNV: This should omit the `make_shared` and call `try_emplace( remote_name )` to prevent building a new `nixlRemoteSection` in case of re-use, then only create the new object when `inserted == true`.

More importantly, are we sure we are _not_ re-using a previous shared object in cases where we shouldn't?

jjthomas: how would it be possible to reuse a shared object in a bad case? it should have been removed from the map and therefore unreachable other than through the weak ref

src/core/agent_data.h
iyastreb: Currently there is no recovery mechanism, so the agent needs to be reconnected by application (remove_remote_agent + load_remote)

src/core/nixl_listener.cpp
iyastreb: I think the problem is: reloading metadata for an existing name keeps the same shared_ptr. Weak refs stay valid.

src/core/agent_data.h
jjthomas: i see, but if the underlying transport has recovered on its own (as is the case in the EFA issue we encountered) an explicit reconnect shouldn't be required right? basically NIXL is taking no action this point on a `NIXL_ERR_REMOTE_DISCONNECT` and does not change the underlying transport state

test/gtest/error_handling.cpp
```diff
@@ -484,6 +501,143 @@ TEST_P(TestErrorHandling, XferPostThenFail) {
...
+    }
+
+    // Release the stale dlist before replacing it, then re-prep must recover
+    EXPECT_EQ(initiator.releasedDlistH(remote_side), NIXL_SUCCESS);
+    remote_side = nullptr;
+    ASSERT_EQ(initiator.prepXferDlist(target_name, r_descs, remote_side), NIXL_SUCCESS);
+    EXPECT_EQ(make_xfer(), NIXL_SUCCESS);
+    EXPECT_EQ(initiator.releasedDlistH(local_side), NIXL_SUCCESS);
+    EXPECT_EQ(initiator.releasedDlistH(remote_side), NIXL_SUCCESS);
+}
```
iyastreb, line 565: I think we need to destroy both peers:
m_Target.destroy();
m_Initiator.destroy();

test/gtest/error_handling.cpp
```diff
@@ -484,6 +501,143 @@ TEST_P(TestErrorHandling, XferPostThenFail) {
...
+    EXPECT_EQ(initiator.releaseXferReq(req), NIXL_SUCCESS);
+
+    // A fresh request against the new registration posts and completes
+    ASSERT_EQ(initiator.createXferReq(NIXL_WRITE, l_descs, r_descs, target_name, req),
+              NIXL_SUCCESS);
+    const nixl_status_t status = initiator.postXferReq(req);
+    ASSERT_TRUE(status == NIXL_SUCCESS || status == NIXL_IN_PROG);
+    nixl_notifs_t target_notifs;
+    EXPECT_EQ(m_Initiator.waitForCompletion(req, m_Target, target_notifs), NIXL_SUCCESS);
+}
```
iyastreb, line 608: same

test/gtest/error_handling.cpp
```diff
@@ -484,6 +501,143 @@ TEST_P(TestErrorHandling, XferPostThenFail) {
...
+    // Unchanged re-broadcast of the same metadata
+    m_Initiator.loadRemoteMD(m_Target.getLocalMD());
+
+    const nixl_status_t status = initiator.postXferReq(req);
+    ASSERT_TRUE(status == NIXL_SUCCESS || status == NIXL_IN_PROG);
+    nixl_notifs_t target_notifs;
+    EXPECT_EQ(m_Initiator.waitForCompletion(req, m_Target, target_notifs), NIXL_SUCCESS);
+    EXPECT_EQ(initiator.releasedDlistH(local_side), NIXL_SUCCESS);
+    EXPECT_EQ(initiator.releasedDlistH(remote_side), NIXL_SUCCESS);
+}
```
iyastreb, line 645: same

iyastreb:
Test failure seems to be relevant

jjthomas:
I believe this is a flake — test seems to show pass now?On Aug 21, 2026, at 2:44 AM, Ilia Yastrebov ***@***.***> wrote:﻿iyastreb left a comment (ai-dynamo/nixl#2051)
Test failure seems to be relevant

iyastreb (dismissed):
Test failure seems to be relevant

jjthomas:
@iyastreb are you sure? it seemed the test had passed again the CI

iyastreb:
> @iyastreb are you sure? it seemed the test had passed again the CI

@jjthomas Sorry, somehow I reposted old comment unintentionally
This PR LGTM, @ColinNV wdyt?

src/core/nixl_agent.cpp
```diff
@@ -1107,20 +1111,12 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
     }
 
     std::shared_lock<nixlLock> read_lock(data->lock);
-    // Check if the remote was invalidated before post/repost
-    const auto sec_it = data->remoteSections_.find(req_hndl->remoteAgent);
-    if (sec_it == data->remoteSections_.end()) {
+    // The request was created against a specific remote registration generation: refuse to
```
ColinNV, line 1114: Would "connection" be more appropriate now that we don't have a generation counter?

src/core/nixl_agent.cpp
```diff
@@ -1107,20 +1111,12 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
...
-        return NIXL_ERR_NOT_FOUND;
-    }
-    if (sec_it->second.getGeneration() != req_hndl->remoteGeneration_) {
-        NIXL_ERROR_FUNC << "remote agent '" << req_hndl->remoteAgent
-                        << "' was re-registered after transfer request creation; "
-                           "refusing to post a stale-generation handle (created gen "
-                        << req_hndl->remoteGeneration_ << ", live gen "
-                        << sec_it->second.getGeneration() << ")";
+                        << "' was invalidated or re-registered after transfer request creation; "
+                           "refusing to post a stale-generation handle";
```
ColinNV:
```suggestion
                           "not posting stale handle";
```

src/core/nixl_agent.cpp
```diff
@@ -1185,11 +1178,8 @@ nixlAgent::postXferReq(nixlXferReqH *req_hndl,
 
     if (req_hndl->status < 0) {
         if (req_hndl->status == NIXL_ERR_REMOTE_DISCONNECT) {
-            read_lock.unlock();
```
ColinNV, line 1188: Getting rid of this re-locking is really good.

@rakhmets

src/core/nixl_agent.cpp
```diff
@@ -1786,8 +1773,13 @@ nixlAgentData::loadConnInfo(const std::string &remote_name,
 
 nixl_status_t
 nixlAgentData::loadRemoteSections(const std::string &remote_name, nixlSerDes &sd) {
-    const auto [it, inserted] = remoteSections_.try_emplace(remote_name, remote_name);
-    const nixl_status_t ret = it->second.loadRemoteData(&sd, backendEngines_);
+    // Reloads merge into the existing section so that partial metadata updates
+    // accumulate and handles weakly bound to the registration stay valid across
```
ColinNV:
```suggestion
    // accumulate. The handles weakly bound to the registration stay valid across
```

On first reading I parsed `handles` as verb.

src/core/nixl_agent.cpp
```diff
@@ -1786,8 +1773,13 @@ nixlAgentData::loadConnInfo(const std::string &remote_name,
...
+    // refreshes; handles retire only when the registration is explicitly
+    // invalidated and this entry is erased.
+    const auto [it, inserted] =
+        remoteSections_.try_emplace(remote_name, std::make_shared<nixlRemoteSection>(remote_name));
```
ColinNV, line 1781: Might be worth calling `try_emplace( remote_name )` and only constructing a `nixlRemoteSection` when it's needed, rather than creating and discarding when `inserted == false` as it is now.

src/core/transfer_request.h
```diff
@@ -33,6 +33,12 @@ enum nixl_telemetry_stat_status_t {
     NIXL_TELEMETRY_FINISH = 2
 };
 
+class nixlRemoteSection;
+
+// Weak ref to the owning remote section: expires when that registration generation is
+// invalidated or replaced, which is the staleness signal for handles created against it.
+using nixl_remote_section_weak_t = std::weak_ptr<nixlRemoteSection>;
```
ColinNV, line 40: I would remove this type alias, writing out the weak pointer is more readable as it spells out what it's pointing at.

src/core/transfer_request.h
```diff
@@ -71,8 +77,7 @@ class nixlXferReqH {
     nixl_meta_dlist_t targetDescs;
 
     const std::string remoteAgent;
-    // Generation of the remote-connection
-    const uint64_t remoteGeneration_;
+    const nixl_remote_section_weak_t remoteSectionRef;
```
ColinNV:
```suggestion
    const nixl_remote_section_weak_t remoteSection;
```
It's a weak pointer, not a reference, and it probably shouldn't be part of the name anyhow.

ColinNV (commented):
LGTM, just a few (mostly) cosmetic comments.

src/core/transfer_request.h
jjthomas: i think you and @iyastreb disagreed on this so let's just keep it as is

src/core/nixl_agent.cpp
jjthomas: the term generation is still used throughout the code so let's leave as is, i think it's understandable enough

src/core/nixl_agent.cpp
jjthomas: doubt the performance benefit really matters here right? and the code becomes a bit more verbose

iyastreb (dismissed):

iyastreb:
test failures are relevant

jjthomas:
fixed compilation issue

iyastreb (approved):

PR #2056, opened 2026-08-06
nixl_ep: round max tokens up to a multiple of 4 for TMA in elastic.py 

lishapira:
Low-latency dispatch aborts with "TMA requires the number of tokens to be
multiple of 4" whenever active_rank_bound * num_max_dispatch_tokens_per_rank is not a multiple of 4. For example: two ranks with an odd max_tokens could not
run. Four nightly cases fail on this.
 
Round the maximum tokens up in elastic.py instead of hitting the assert. Note
this rounds the declared capacity only, the actual token count is untouched.
 
Applied where elastic.py hands the value to NIXL EP: get_rdma_size_hint() and
the two dispatch() calls. 
 
The constraint comes from the FP8 scale tensor. Scales are stored channel-major: all M token slots for one channel block, then the next, where M = num_ranks * max_tokens. Each slot is 4 bytes in both quantization modes, so the gap between blocks is 4M bytes, and TMA needs that to be a multiple of 16 -- i.e. M a multiple of 4.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token-capacity handling for TMA alignment by rounding capacities up to multiples of four.
  * Ensured aligned capacities are used for buffer allocation and dispatch while preserving the actual workload token count.

* **Documentation**
  * Documented the requirement for low-latency buffer sizes and dispatch token capacities to be multiples of four.

### DIFF
--- a/examples/device/ep/nixl_ep/buffer.py
+++ b/examples/device/ep/nixl_ep/buffer.py
@@ -147,6 +147,7 @@ class Buffer:
 
         Arguments:
             num_max_dispatch_tokens_per_rank: the maximum number of tokens to dispatch, all the ranks must hold the same value.
+                `num_ranks * num_max_dispatch_tokens_per_rank` must be a multiple of 4 to match `low_latency_dispatch()`.
             hidden: the hidden dimension of each token.
             num_ranks: rank capacity used to size the low-latency buffers.
             num_experts: expert capacity, normally num_ranks * num_experts_per_rank.
@@ -353,6 +354,7 @@ class Buffer:
             topk_idx: `torch.Tensor` with `nixl_ep.topk_idx_t`, shaped as `[num_tokens, num_topk]`, only several top-k shapes
                 are supported. `-1` indices (not selecting any expert) are supported.
             num_max_dispatch_tokens_per_rank: the maximum number of tokens to dispatch, all the ranks must hold the same value.
+                `num_ranks * num_max_dispatch_tokens_per_rank` must be a multiple of 4 for TMA alignment.
             num_experts: deprecated optional parameter. This value is ignored by low-latency
                 dispatch; the number of experts is determined from the `num_experts_per_rank`
                 passed to `update_memory_buffers()` and the currently active ranks.
--- a/examples/device/ep/tests/elastic/elastic.py
+++ b/examples/device/ep/tests/elastic/elastic.py
@@ -49,6 +49,17 @@ from utils import (  # noqa: E402
 TCP_STORE_PORT = 9999
 RANK_SERVER_PORT = 10000
 
+# NIXL EP stores FP8 scales channel-major, so the gap between channel blocks is
+# 4 * num_ranks * max_tokens_per_rank bytes and TMA needs it 16-byte aligned,
+# i.e. num_ranks * max_tokens_per_rank must be a multiple of 4.
+TMA_TOKEN_ALIGNMENT = 4
+
+
+def tma_aligned_max_tokens(num_tokens: int) -> int:
+    return (
+        (num_tokens + TMA_TOKEN_ALIGNMENT - 1) // TMA_TOKEN_ALIGNMENT
+    ) * TMA_TOKEN_ALIGNMENT
+
 
 def non_negative_int(value: str) -> int:
     try:
@@ -90,6 +101,7 @@ def self_kill():
 
 def test_main(
     num_tokens: int,
+    max_tokens_per_rank: int,
     hidden: int,
     num_experts: int,
     num_topk: int,
@@ -208,7 +220,7 @@ def test_main(
                                 buffer.dispatch(
                                     current_x,
                                     topk_idx,
-                                    num_tokens,
+                                    max_tokens_per_rank,
                                     num_experts,
                                     use_fp8=dispatch_use_fp8,
                                     round_scale=round_scale,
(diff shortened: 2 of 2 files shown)


examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1247,7 +1256,10 @@ Buffer::combine(const torch::Tensor& x, const torch::Tensor& topk_idx, const tor
 }
 
 torch::Tensor
-Buffer::get_next_combine_buffer(int num_max_dispatch_tokens_per_rank, int hidden) const {
+Buffer::get_next_combine_buffer(int requested_max_tokens_per_rank, int hidden) const {
+    // Physical capacity: must match the rounding dispatch() applies (TMA alignment).
```
rakhmets: Code duplication and extra comments can be avoided.
Rounding should be done by EPlayout. The ctor of EPlayout should round the value, and then the value should be obtained from the class.

examples/device/ep/csrc/config.hpp
```diff
@@ -37,6 +37,13 @@ dtype_t align_up(dtype_t a, dtype_t b) {
     return ceil_div<dtype_t>(a, b) * b;
 }
 
+// Round the per-rank max tokens up for TMA alignment.
+constexpr int NUM_TMA_SCALE_TOKEN_ALIGNMENT = 4;
```
rakhmets:
```suggestion
constexpr int num_tma_scale_token_alignment = 4;
```
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es32-use-all_caps-for-all-macro-names

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -147,6 +147,7 @@ def get_low_latency_buffer_size_hint(
 
         Arguments:
             num_max_dispatch_tokens_per_rank: the maximum number of tokens to dispatch, all the ranks must hold the same value.
+                Rounded up to a multiple of 4 internally (TMA alignment).
```
rakhmets: Not sure if this addition is needed here.

examples/device/ep/csrc/config.hpp
ColinNV: Since it's in a header it should also be `inline`.

itayalroy (changes requested):
We cannot silently change `num_max_dispatch_tokens_per_rank` inside NIXL EP, it changes the output buffer layout, users use `num_max_dispatch_tokens_per_rank` to calculate offsets inside this output buffer and they will get wrong offsets.

To fix the nightly test issues, the round-up should happen in elastic.py (the NIXL EP user in our case), not in NIXL EP itself

lishapira:
> We cannot silently change `num_max_dispatch_tokens_per_rank`, it changes the output buffer layout, users use `num_max_dispatch_tokens_per_rank` to calculate offsets inside this output buffer and they will get wrong offsets.
> 
> To fix the nightly test issues, the round-up should happen in elastic.py (the NIXL EP user in our case), not in NIXL EP itself

But other users, like vLLM, can also fail without rounding num_max_dispatch_tokens_per_rank.

itayalroy:
> > We cannot silently change `num_max_dispatch_tokens_per_rank`, it changes the output buffer layout, users use `num_max_dispatch_tokens_per_rank` to calculate offsets inside this output buffer and they will get wrong offsets.
> > To fix the nightly test issues, the round-up should happen in elastic.py (the NIXL EP user in our case), not in NIXL EP itself
> 
> But other users, like vLLM, can also fail without rounding num_max_dispatch_tokens_per_rank.

We should probably just document that we need to choose max tokens % 4 == 0, default values in all existing integrations (vLLM, SGLang) already comply, and we already assert and fail gracefully when it is not the case

examples/device/ep/nixl_ep/buffer.py
lishapira: Do you still think that after the latest changes?

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -49,6 +49,17 @@
 TCP_STORE_PORT = 9999
 RANK_SERVER_PORT = 10000
 
+# NIXL EP stores FP8 scales channel-major, so the gap between channel blocks is
```
itayalroy: this is not a good place for this comment, we shouldn't explain nixl ep implementation details in the benchmark script, also we already documented this limitation in `buffer.py`, I don't think we need any other comment here

examples/device/ep/tests/elastic/elastic.py
```diff
@@ -491,8 +503,15 @@ def worker(torch_rank: int, args: argparse.Namespace):
     )
 
     # Initialize nixl_ep buffer
+    max_tokens_per_rank = tma_aligned_max_tokens(args.num_tokens)
+    if local_rank == 0 and max_tokens_per_rank != args.num_tokens:
+        print(
```
itayalroy: This isn't very interesting to the user imo and might also be misleading, the user really shouldn't care that we reserved memory for <= 3 additional tokens, let's keep the benchmark log clean and remove this print

lishapira: Removed

examples/device/ep/tests/elastic/elastic.py
lishapira: Removed

itayalroy (approved):

PR #2062, opened 2026-08-10
Fix/2032 gds mt interior pointer

### LINKED ISSUE #2032: [gds_mt] Preserve cuFile registered base and descriptor offset for interior GPU pointers in gds_mt
## Problem

The current `gds_mt` backend appears to pass the transfer descriptor pointer directly to `cuFileRead` / `cuFileWrite` as the buffer base, with a buffer offset of `0`.

Current shape:

```cpp
cuFileRead(req->fh, req->addr, req->size, req->file_offset, 0);
cuFileWrite(req->fh, req->addr, req->size, req->file_offset, 0);
```

This works when the transfer descriptor pointer is identical to the pointer registered with cuFileBufRegister. However, some KV cache layouts can produce transfer descriptors that point inside a larger registered allocation.

For example:
```cpp
cuFileBufRegister(registered_base, registered_size, flags);
descriptor_addr = registered_base + descriptor_offset;
```

In this case, the transfer descriptor is still inside registered memory, but the pointer passed to the cuFile API differs from the registered base pointer.

### [Why this matters]

For layouts such as layer-separated KV cache storage, the buffer registered with cuFile can be a larger contiguous allocation, while individual transfer descriptors can refer to interior regions.

In that case, `gds_mt` should preserve:
```cpp
devPtr_base   = registered_base;
devPtr_offset = descriptor_addr - registered_base;
```
and call cuFile APIs using the registered base plus descriptor-relative offset.

This follows the [GDS Best Practices Guide](https://docs.nvidia.com/gpudirect-storage/best-practices-guide/index.html) recommendation for accessing subranges of a parent-registered GPU buffer: use the registered buffer base as devPtr_base and specify the sub-buffer location through devPtr_offset.

### [Current behavior]

`gds_mt` currently stores a single pointer in the transfer request:
```cpp
struct GdsMtTransferRequestH {
    GdsMtTransferRequestH (void *a,
                           size_t s,
                           size_t offset,
                           CUfileHandle_t handle,
                           CUfileOpcode_t operation)
        : addr{a},
          size{s},
          file_offset{offset},
          fh{handle},
          op{operation} {}

    void *addr;
    size_t size;
    size_t file_offset;
    CUfileHandle_t fh;
    CUfileOpcode_t op;
};
```
and submits I/O using:
```cpp
nbytes = cuFileRead (req->fh, req->addr, req->size, req->file_offset, 0);
nbytes = cuFileWrite (req->fh, req->addr, req->size, req->file_offset, 0);
```
This does not distinguish between:
- the pointer originally registered with `cuFileBufRegister`
- the descriptor pointer used for this specific transfer

### [Proposed direction]
Store the registered memory base and descriptor-relative offset separately in the gds_mt transfer request.

Conceptually:
```cpp
struct GdsMtTransferRequestH {
    void *dev_ptr_base;
    size_t dev_ptr_offset;
    size_t size;
    size_t file_offset;
    CUfileHandle_t fh;
    CUfileOpcode_t op;
};
```
```cpp
cuFileRead(
    req->fh,
    req->dev_ptr_base,
    req->size,
    req->file_offset,
    req->dev_ptr_offset
);
```
and similarly for `cuFileWrite`.

### [Expected benefit]
This would make gds_mt handle interior GPU pointers consistently with cuFile registered-buffer semantics and avoid unintended fallback behavior when transfer descriptors refer to subranges of a registered allocation.

### [Validation plan]

I can follow up with a PR that adds unit coverage for registered-range resolution:

- descriptor exactly equal to registered base: offset is `0`
- descriptor inside registered range: offset is computed correctly
- descriptor outside registered range: request is rejected
- descriptor range partially exceeds registered range: request is rejected

If useful, I can also add benchmark data in a follow-up comment showing TTFT/QPS and peak bandwidth differences for a layer-separated KV cache workload.

Chije:
## What?
Fix GDS_MT handling of interior GPU pointers.

The GDS_MT backend currently treats the transfer descriptor pointer as the
cuFile registered base pointer. This is incorrect when a transfer descriptor
refers to an interior region of a larger GPU allocation.

This PR:

- Preserves the GPU buffer base registered with cuFile.
- Computes the descriptor-relative offset.
- Passes the registered base and descriptor offset to the GDS_MT transfer path.
- Rejects descriptors outside the registered allocation.
- Adds unit and integration test coverage.

Fixes #2032

## Why?

A GPU allocation may be registered once with cuFile while individual transfer
descriptors refer to subranges inside that allocation.

Without preserving the registered base and descriptor-relative offset:

- An interior descriptor may be passed to cuFile as the buffer base with
  `devPtr_offset = 0`, even though the larger allocation was registered at a
  different base address.
- A descriptor outside the registered allocation is not rejected during NIXL
  request preparation and may reach cuFile, where it can fail later as a
  backend I/O error.

This change makes GDS_MT follow cuFile registered-buffer semantics for
interior GPU pointers.

## How?

The GDS_MT buffer-resolution path distinguishes between:

- The base pointer registered with cuFile.
- The descriptor offset relative to the registered base.

The request preparation path validates that the complete transfer descriptor is
contained within the registered allocation before creating a transfer request.

Unit tests cover:

- Interior descriptor offset calculation.
- Descriptor equal to the registered base.
- Descriptor before the registered base.
- Descriptor extending beyond the registered allocation.

A separate GDS_MT integration test:

- Allocates a 64 MiB GPU buffer.
- Registers the full allocation with cuFile.
- Uses an 8 MiB transfer descriptor at a 4 MiB interior offset.
- Performs a GDS_MT file write.
- Clears the GPU buffer.
- Performs a GDS_MT file read using the same interior descriptor.
- Verifies payload integrity.
- Verifies that the regions before and after the descriptor remain unchanged.
- Records write/read latency through GTest properties.

## Relationship to #1856

PR #1856 proposes consolidating the GDS and GDS_MT implementations under the
`cuda_gds` source tree and sharing request preparation logic.

This PR is based on the current pre-consolidation GDS_MT source layout and does
not include the changes from #1856. If #1856 is merged first, this change will
need to be rebased and ported to the shared `cuda_gds` request-preparation
layer.

The intended follow-up is to preserve the same registered-base and
descriptor-offset semantics for both GDS and GDS_MT after the consolidation,
while ensuring that the offset is calculated and applied exactly once.

## Testing

Focused validation passed:

```text
6/6 tests passed

Passed tests:
- nixl:posix_plugin_test
- gds_mt_integration: nixl:gds_mt_interior_pointer_integration
- nixl:unit
- sanitizer: nixl:gtest
- nixl:telemetry_benchmark
- nixl:tracing_nsys
```

### Build environment

Validation was performed with:

- CUDA 13.1
- GDS enabled
- Custom UCX, etcd-cpp-api, and Abseil prefixes
- Local etcd endpoint: `http://127.0.0.1:2379`
- DOCA telemetry disabled because the available exporter library was
  ABI-incompatible

The dependency paths below are represented by environment variables because
they are machine-specific.

### Build configuration
```bash
$ meson setup build-gds-mt-pr-focused \
  -Dbuild_tests=true \
  -Dbuild_examples=false \
  -Dbuildtype=debug \
  -Ddisable_plugins=TELEMETRY_DOCA \
  -Ddisable_gds_backend=false \
  -Dgds_path="$CUDA_HOME" \
  -Detcd_inc_path="$ETCD_PREFIX/include" \
  -Detcd_lib_path="$ETCD_PREFIX/lib"
```

### Build
```bash
$ ninja -C build-gds-mt-pr-focused
```

### Test execution
```bash
$ meson test -C build-gds-mt-pr-focused \
  --no-rebuild \
  --timeout-multiplier 12 \
  --print-errorlogs \
  --logbase=pr-focused-testlog \
  posix_plugin_test \
  unit \
  gtest \
  telemetry_benchmark \
  tracing_nsys \
  gds_mt_interior_pointer_integration
```
The four GDS_MT offset unit tests passed as part of `nixl:unit`.

The two existing GDS path-mode smoke tests were excluded from this focused
validation because their runtime-directory setup issue is unrelated to #2032.
No smoke-test infrastructure changes are included in this PR.

`TELEMETRY_DOCA` was disabled because the available DOCA telemetry exporter
library was ABI-incompatible with the required symbols. The GDS and GDS_MT
validation was executed with GDS enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GDS multi-threaded transfers for interior pointers within registered GPU memory.
  * Added validation to reject missing, invalid, or out-of-bounds transfer metadata.
  * Ensured file transfers use the correct registered buffer location and offset.

* **Tests**
  * Added coverage for valid offsets, exact-base transfers, and invalid buffer ranges.
  * Added GPU-backed integration testing to verify interior-region transfers preserve surrounding data.


mbapatu-spec:
@Chije did you get a chance to run nixl bench on these changes?

Chije:
@mbapatu-spec I ran an additional NIXLBench comparison between mainline and
this PR after changing the NVMe device.

Configuration:

- Backend: `GDS_MT`
- Initiator memory: `VRAM`
- Total GPU allocation: 64 MiB
- Block size: 4 MiB
- Batch size: 16
- Iterations: 100
- Warmup iterations: 20
- GDS_MT threads: 16
- Direct storage I/O enabled
- CUDA 13.1 with GDS enabled
- The same benchmark configuration and NVMe device were used for both runs

Results:

| Operation | Mainline BW (GB/s) | PR BW (GB/s) | BW delta | Mainline Avg Lat. (us) | PR Avg Lat. (us) |
|---|---:|---:|---:|---:|---:|
| WRITE | 6.303 | 6.266 | -0.59% | 665.4 | 669.4 |
| READ | 12.648 | 12.855 | +1.64% | 331.6 | 326.3 |

The results show comparable WRITE performance and slightly higher READ
throughput for the PR. No meaningful throughput or latency regression was
observed in this run.

The benchmark uses a 64 MiB GPU allocation with multiple 4 MiB transfer
descriptors, including descriptors at interior offsets within the allocation.

This benchmark is supplementary to the unit and GDS_MT integration tests,
which validate the registered-base and descriptor-offset behavior.

The full raw logs are attached as : [nixlbench-pr2062-logs.tar.gz](https://github.com/user-attachments/files/30922660/nixlbench-pr2062-logs.tar.gz)

tell-rebanta:
If you are doing I/O using multiple threads simultaneously with the same base_ptr, then cufile library may fallback to the  slower I/O path due to the fact that base_ptr is busy with another I/O. Given that, you may not see any gain in performance defeating the purpose of your change.

Chije:
@tell-rebanta Thanks for pointing this out. 
We tested mainline and PR #2062 with
144 concurrent GDS_MT threads, a 256 MiB registered GPU buffer, and
1 MiB transfers.

Both versions issued the same number of GDS operations, with no POSIX
fallback events observed. The internal bounce-buffer completion
tracepoint was observed in both versions, with Registered=1 and
Unaligned=0.

Performance was effectively unchanged:

  READ:  13.181 GB/s (mainline) vs 13.248 GB/s (PR)
  WRITE: 6.197 GB/s (mainline) vs 6.199 GB/s (PR)

Therefore, we did not observe a meaningful performance regression, and
the bounce-buffer behavior was not introduced by this PR.

The goal of PR #2062 is to correctly preserve the cuFile-registered
base pointer and apply the descriptor-relative offset for interior GPU
pointers, rather than to change cuFile's internal bounce-buffer
selection policy.

The current tracepoint does not expose the exact reason for selecting
the bounce-buffer path, so this experiment cannot conclusively confirm
that a busy base_ptr is the cause. We can investigate this further with
a concurrency sweep if needed.

mbapatu-spec:
LGTM

tell-rebanta:
While I agree that the performance has not been regressed, it did not improve either with this change if that is the goal. FYI,  If you enable cufile logging in TRACE mode, cufile.log file can show if it has taken registered/unregistered path or not (specifically, I would search "unregistered work item" in the log).

Chije:
### Additional cuFile route-trace validation

I added cuFile route-trace results from a Dynamo KVBM + vLLM workload using
`Qwen3-30B-A3B-Instruct-2507`.

Test configuration:

- 48 layers, TP=1, PP=1
- Block size: 128
- KV path: `/mnt/kvbm/Qwen3-30B`
- Duration: 500 seconds
- 3 repeats per configuration
- GDS/cuFile: `1.15.1.6`
- Pre-PR runtime image: `vllm-runtime:1.3.0`

The route trace shows a clear difference in READ handling:

| Configuration | READ/BB | READ/DIRECT |
|---|---:|---:|
| Pre-PR vanilla | 100% | 0% |
| PR-applied | 65.32–68.26% | 31.74–34.68% |

In the pre-PR configuration, all READ operations used the bounce-buffer
route even though the trace reported zero unaligned and zero unregistered
operations. With the PR applied, approximately one third of READ operations
used the DIRECT route.

This is consistent with the intended behavior of preserving the cuFile
registered base pointer and applying the descriptor-relative offset for
interior GPU pointers. No read/write return errors were observed.

The attached route-specific latency histograms show the distributions for
`READ/BB` and `READ/DIRECT`. The average per-repeat percentiles were:

| Route | p50 | p95 | p99 |
|---|---:|---:|---:|
| Pre-PR `READ/BB` | 2.70 ms | 14.96 ms | 31.97 ms |
| PR `READ/BB` | 1.54 ms | 6.62 ms | 12.47 ms |
| PR `READ/DIRECT` | 0.78 ms | 1.82 ms | 1.98 ms |

The histogram is intentionally split by route because an aggregate READ
latency would combine BB and DIRECT operations and hide the route-specific
tail behavior.

PR #2063, opened 2026-08-10
ci: raise the registry push retry budget

NirWolfer:
## Summary

Transient `HTTP 503` responses from Artifactory were failing otherwise-green builds at the push step, after the image had already built successfully:

```
Error: trying to reuse blob sha256:3cbead0c... at destination: checking whether a blob
exists in artifactory.nvidia.com/...: received unexpected HTTP status: 503 Service Unavailable
Compiling NIXL Docker Image failed with exit code=125
```

Seen on `nixl-ci-gpu` #3107 (503, Aug 9) and #2651 (502 Bad Gateway, Jul 9). In both cases the build, install and image commit all succeeded and the stage failed only on the push, with steady blob-copy progress right up to the error - registry availability, not a hang or a code defect.

## Changes

Every push now retries. The mechanism differs by job because the two runner images ship different podman versions.

### Native flag - `build-wheel-matrix.yaml`, `build-container-matrix.yaml`

These run on `quay.io/podman/stable:v5.7.1`, so they use `--retry 5`.

podman already retries this class of error: `go.podman.io/common/pkg/retry` treats HTTP 502-504 and network errors as retryable, while returning `false` for unauthorized, denied, unknown name and unknown manifest - so a larger budget never delays a genuine failure. The problem is the budget, not the mechanism: the default of 3 attempts with a `2^attempt` second backoff spans only 1+2+4 = **7 seconds**, which both outages outlasted. `--retry 5` extends it to 1+2+4+8+16 = **31 seconds**.

`--retry-delay` is left unset deliberately - passing it replaces the exponential backoff with a fixed delay.

Both files invoke `docker push`, but `docker` is symlinked to `podman` in those jobs (`ln -sfT $(type -p podman) /usr/bin/docker`), so the flag applies.

### Shell loop - `test-matrix.yaml`, `test-dl-matrix.yaml`, `test-dl-ep-matrix.yaml`

These push from the `build_helper` container, which installs podman from Ubuntu 24.04 apt (4.9.x). `--retry` was added in **podman 5.0**, so the flag aborts the step outright:

```
+ podman push --retry 5 artifactory.nvidia.com/.../nixl-ci-gpu-test:3116
Error: unknown flag: --retry
```

Upgrading is not a practical option on this base - podman 5.x is not in `noble`, `noble-updates` or `noble-backports` (it first appears in Ubuntu 25.10), and `Dockerfile.build_helper` deliberately pins apt to the internal mirror with public archives removed. So these use a retry loop with the same 2, 4, 8, 16 second backoff.

If `build_helper` ever moves to podman 5.0+, the loop can be replaced by the flag - noted in the docs.

## Verification

All five matrix files parse as YAML, and the loop was checked for both the recovery path and a non-zero exit on exhaustion. The retry path only exercises on a registry failure, so a green run confirms the commands are accepted, not that the backoff helps; the real test is the next transient 503.

## Summary by CodeRabbit

* **Improvements**
  * Container, wheel, and GPU test image publishing now retries up to five times for transient registry or network failures.
  * GPU test workflows use increasing delays between attempts and stop immediately after a successful push.
  * Authentication and manifest errors remain non-retriable, while temporary service interruptions are handled automatically.

* **Documentation**
  * Added guidance on retry behavior, retryable failures, non-retriable errors, retry windows, and container selection across publishing workflows.

### DIFF
--- a/.ci/docs/ci-overview.md
+++ b/.ci/docs/ci-overview.md
@@ -264,6 +264,39 @@ new tag and triggers a rebuild automatically.
 **You never need to manually update `CI_IMAGE_TAG`.** The `CI_MANAGED`
 placeholder signals this clearly.
 
+## Registry push retries
+
+Transient `HTTP 503 Service Unavailable` responses from Artifactory were failing
+otherwise-green builds at the push step, after the image had already built
+successfully. Every image push now retries, but the mechanism differs because
+`--retry` was only added in podman 5.0 and most push steps run on an older one.
+
+**`build-container-matrix.yaml`** runs on `quay.io/podman/stable:v5.7.1` and uses
+the native `--retry 5`. podman already retries this class of error —
+`go.podman.io/common/pkg/retry` treats HTTP 502-504 and network errors as
+retryable, while failing fast on unauthorized, denied and unknown
+name/manifest, so a larger budget never delays a genuine failure. `--retry N`
+counts retries *after* the initial push, and the backoff is `2^attempt` seconds
+plus 10% jitter. The default of 3 retries therefore spans 1+2+4 = 7 seconds,
+which the outages above outlasted; `--retry 5` extends it to 1+2+4+8+16 = 31
+seconds over 6 attempts. `--retry-delay` is left unset on purpose: passing it
+replaces the exponential backoff with a fixed delay.
+
+**`test-matrix.yaml`, `test-dl-matrix.yaml`, `test-dl-ep-matrix.yaml` and
+`build-wheel-matrix.yaml`** push from a `Dockerfile.build_helper` container,
+which installs podman from Ubuntu 24.04 apt (4.9.x). podman 5 is not available
+for 24.04 — not in `noble`, `noble-updates` or `noble-backports` — so these use
+a shell retry loop deliberately matched to the flag: 6 attempts with 1, 2, 4, 8
+and 16 second delays. The one behavioural difference is that the loop retries on
+any failure, so an auth error waits out the backoff instead of failing
+immediately.
+
+`build-wheel-matrix.yaml` is the easy one to get wrong: its `Prepare` step
+symlinks `docker` to `podman` in two different containers, and the push in
+`Build sanity image` runs in the `build_helper_(vllm|sglang)` one, not the
+`manylinux` runner. Check the step's `containerSelector` against
+`runs_on_dockers`, not just whether `docker` is podman.
+
 ## Related docs
 
 - [Build Wheel Matrix CI Job Documentation](build-wheel-matrix-ci.md) — deep dive into `nixl-ci-build-wheel`.
--- a/.ci/jenkins/lib/build-container-matrix.yaml
+++ b/.ci/jenkins/lib/build-container-matrix.yaml
@@ -182,7 +182,7 @@ steps:
         local target_tag="$1"
         echo "Creating tag: ${target_tag}"
         docker tag "${LOCAL_TAG_BASE}${arch}" "${ARTIFACTORY_REGISTRY}:${target_tag}"
-        docker push "${ARTIFACTORY_REGISTRY}:${target_tag}"
+        docker push --retry 5 "${ARTIFACTORY_REGISTRY}:${target_tag}"
         curl --fail -H "Authorization: Bearer ${ARTIFACTORY_PASSWORD}" -X PUT \
           "${ARTIFACTORY_API}/${target_tag}?properties=${IMAGE_PROPERTIES}"
       }
--- a/.ci/jenkins/lib/build-wheel-matrix.yaml
+++ b/.ci/jenkins/lib/build-wheel-matrix.yaml
@@ -331,7 +331,11 @@ steps:
       # 2. Copy the whole workspace in (test script + helpers), like the other pipelines.
       printf 'FROM %s\nCOPY . /workspace/nixl\nWORKDIR /workspace/nixl\n' "${IMG}" \
         | docker build --platform linux/${arch} -t "${IMG}" -f - .
-      docker push "${IMG}"
+      for i in {1..6}; do
+          docker push "${IMG}" && break
+          [ "$i" = 6 ] && exit 1
+          sleep $((2 ** (i - 1)))
+      done
 
   - name: Allocate Environment
     containerSelector: "{name: 'build_helper_(vllm|sglang)'}"
--- a/.ci/jenkins/lib/test-dl-ep-matrix.yaml
+++ b/.ci/jenkins/lib/test-dl-ep-matrix.yaml
@@ -121,7 +121,11 @@ steps:
       --build-arg BASE_IMAGE=${registry_host}${registry_path}/${arch}/nixl-ci-dl-gpu-ep-base-pytorch26.06-cuda13.3-ubuntu24.04:${CI_IMAGE_TAG} \
       --tag ${PR_IMAGE} \
       -f .ci/dockerfiles/Dockerfile.gpu-test .
-      podman push --creds ${REPO_USER}:${REPO_PASS} ${PR_IMAGE}
+      for i in {1..6}; do
+          podman push --creds ${REPO_USER}:${REPO_PASS} ${PR_IMAGE} && break
+          [ "$i" = 6 ] && exit 1
+          sleep $((2 ** (i - 1)))
+      done
 
   - name: Allocate DL EP Environment
     containerSelector: "{name: 'build_helper_dl_ep'}"
--- a/.ci/jenkins/lib/test-dl-matrix.yaml
+++ b/.ci/jenkins/lib/test-dl-matrix.yaml
@@ -112,7 +112,11 @@ steps:
       --build-arg BASE_IMAGE=${registry_host}${registry_path}/${arch}/nixl-ci-dl-gpu-base-pytorch26.06-cuda13.3-ubuntu24.04:${CI_IMAGE_TAG} \
       --tag ${PR_IMAGE} \
       -f .ci/dockerfiles/Dockerfile.gpu-test .
-      podman push --creds ${REPO_USER}:${REPO_PASS} ${PR_IMAGE}
+      for i in {1..6}; do
+          podman push --creds ${REPO_USER}:${REPO_PASS} ${PR_IMAGE} && break
+          [ "$i" = 6 ] && exit 1
+          sleep $((2 ** (i - 1)))
+      done
 
   - name: Allocate DL Environment
     containerSelector: "{name: 'build_helper_dl'}"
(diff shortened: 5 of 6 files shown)


.ci/docs/ci-overview.md
```diff
@@ -264,6 +264,22 @@ new tag and triggers a rebuild automatically.
...
+## Registry push retries
+
+Every image push to Artifactory runs with `--retry 5`. Transient `HTTP 503
+Service Unavailable` responses from the registry were failing otherwise-green
+builds at the push step, after the image had already built successfully.
+
+`podman push` retries 3 times by default with exponential backoff starting at
+two seconds, which only covers a few seconds of registry downtime. `--retry 5`
+extends that to roughly 30 seconds without disabling the backoff (passing
+`--retry-delay` would replace the exponential backoff with a fixed delay).
```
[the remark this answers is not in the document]
NirWolfer: The jitter is present in v5.7.1, so I am keeping the claim.

Checked the copy actually vendored at the pinned tag — note the module path moved to `go.podman.io`, so it is `vendor/go.podman.io/common/pkg/retry/retry.go`, not the old `vendor/github.com/containers/common/...`:

https://github.com/containers/podman/blob/v5.7.1/vendor/go.podman.io/common/pkg/retry/retry.go

```go
for attempt := 0; err != nil && isRetryable(err) && attempt < options.MaxRetry; attempt++ {
    delay := time.Duration(int(math.Pow(2, float64(attempt)))) * time.Second
    if options.Delay != 0 {
        delay = options.Delay
    }
    logrus.Warnf("Failed, retrying in %s ... (%d/%d). Error: %v", delay, attempt+1, options.MaxRetry, err)
    delay += rand.N(delay / 10) // 10 % jitter so that a failure blip doesn't cause a deterministic stampede
```

So the base delays are 1, 2, 4, 8, 16 seconds and each has up to 10% added on top — the 31 s figure is the floor, with the ceiling around 34 s. The docs already describe it that way ("`2^attempt` seconds plus 10% jitter"), so no change needed.

The same file also carries the 502-504 branch this PR relies on, which is the other thing worth confirming at the pinned version:

```go
case docker.UnexpectedHTTPStatusError:
    if e.StatusCode >= http.StatusBadGateway && e.StatusCode <= http.StatusGatewayTimeout {
        return true
    }
```

.ci/docs/ci-overview.md
```diff
@@ -264,6 +264,28 @@ new tag and triggers a rebuild automatically.
...
+Service Unavailable` responses from the registry were failing otherwise-green
+builds at the push step, after the image had already built successfully.
+
+`podman push` already retries 3 times by default. The backoff is `2^attempt`
+seconds plus 10% jitter, so the default budget spans 1+2+4 = 7 seconds of
+registry downtime — the failures above happened despite it. `--retry 5` extends
+that to 1+2+4+8+16 = 31 seconds. `--retry-delay` is left unset on purpose:
+passing it replaces the exponential backoff with a fixed delay.
+
+Retries are limited to what is worth retrying (`containers/common/pkg/retry`):
```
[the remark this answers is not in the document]
NirWolfer: Good catch, fixed in 98192688.

`containers/common/pkg/retry` is where the package used to live and it does not exist at the pinned tag — that path 404s against v5.7.1, which is what sent the earlier finding astray. The docs now cite `go.podman.io/common/pkg/retry`, matching `vendor/go.podman.io/common/pkg/retry/retry.go` in podman v5.7.1.

.ci/jenkins/lib/build-wheel-matrix.yaml
```diff
@@ -331,7 +331,11 @@ steps:
       # 2. Copy the whole workspace in (test script + helpers), like the other pipelines.
       printf 'FROM %s\nCOPY . /workspace/nixl\nWORKDIR /workspace/nixl\n' "${IMG}" \
         | docker build --platform linux/${arch} -t "${IMG}" -f - .
-      docker push "${IMG}"
+      for i in 1 2 3 4 5; do
+        if docker push "${IMG}"; then break; fi
+        if [ "$i" = 5 ]; then exit 1; fi
+        sleep $((2 ** i))
+      done
```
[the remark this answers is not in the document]
NirWolfer: Fixed in 138fd8ad — the loops now run 6 attempts with 1, 2, 4, 8, 16 second delays, matching `--retry 5` exactly.

You were right that they were not equivalent. The old loop was 5 attempts with 2+4+8+16 = 30 s, because it slept `2**i` starting at i=1 rather than `2**(i-1)` starting at 0, and gave up one attempt early. Against the native path:

```go
err := operation()                                                  // attempt 1
for attempt := 0; err != nil && ... && attempt < options.MaxRetry; attempt++ {
    delay := time.Duration(int(math.Pow(2, float64(attempt)))) * time.Second   // 1,2,4,8,16
    ...
    err = operation()
}
```

`--retry 5` is 1 initial push plus 5 retries = 6 attempts, 31 s of backoff. The loop now matches. Applied to all four loop sites (`test-matrix`, `test-dl-matrix`, `test-dl-ep-matrix`, `build-wheel-matrix`) and the docs updated accordingly.

.ci/docs/ci-overview.md
```diff
@@ -264,6 +264,37 @@ new tag and triggers a rebuild automatically.
...
+`--retry` was only added in podman 5.0 and most push steps run on an older one.
+
+**`build-container-matrix.yaml`** runs on `quay.io/podman/stable:v5.7.1` and uses
+the native `--retry 5`. podman already retries this class of error —
+`go.podman.io/common/pkg/retry` treats HTTP 502-504 and network errors as
+retryable, while failing fast on unauthorized, denied and unknown
+name/manifest, so a larger budget never delays a genuine failure. The default of
+3 attempts with a `2^attempt` second backoff (plus 10% jitter) only spans
+1+2+4 = 7 seconds, which the outages above outlasted; `--retry 5` extends it to
+1+2+4+8+16 = 31 seconds. `--retry-delay` is left unset on purpose: passing it
```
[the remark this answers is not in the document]
NirWolfer: Corrected in 138fd8ad. The docs now say `--retry N` counts retries *after* the initial push, so the default of 3 retries spans 1+2+4 = 7 s, and `--retry 5` gives 1+2+4+8+16 = 31 s over 6 attempts. The 7 s and 31 s figures were already right; only the attempt/retry wording was off.

.ci/jenkins/lib/build-wheel-matrix.yaml
```diff
@@ -331,7 +331,11 @@ steps:
       # 2. Copy the whole workspace in (test script + helpers), like the other pipelines.
       printf 'FROM %s\nCOPY . /workspace/nixl\nWORKDIR /workspace/nixl\n' "${IMG}" \
         | docker build --platform linux/${arch} -t "${IMG}" -f - .
-      docker push "${IMG}"
+      for i in 1 2 3 4 5 6; do
```
dpressle: better bash writing easy to read:
```bash
for i in {1..6}; do
    docker push "${IMG}" && break
    [ "$i" = 6 ] && exit 1
    sleep $((2 ** (i - 1)))
done
```

.ci/jenkins/lib/build-container-matrix.yaml
```diff
@@ -182,7 +182,7 @@ steps:
         local target_tag="$1"
         echo "Creating tag: ${target_tag}"
         docker tag "${LOCAL_TAG_BASE}${arch}" "${ARTIFACTORY_REGISTRY}:${target_tag}"
-        docker push "${ARTIFACTORY_REGISTRY}:${target_tag}"
+        docker push --retry 5 "${ARTIFACTORY_REGISTRY}:${target_tag}"
```
dpressle, line 185: why retry works here but not on all other piepilines

dpressle (changes requested):

.ci/jenkins/lib/build-wheel-matrix.yaml
NirWolfer: Adopted, applied to all four loop sites in c9ca6880.

I checked it behaves correctly under the `set -eux` these steps run with, since `&&` and `set -e` interact awkwardly. It is fine — a failing command on the left of `&&` is exempt from `set -e`, so the loop retries rather than aborting on the first failure:

```
$ bash -c 'set -eu; n=0; push(){ n=$((n+1)); return 1; }
  for i in {1..6}; do push && break; [ "$i" = 6 ] && exit 1; echo "slept $((2 ** (i - 1)))s after attempt $i"; done'
slept 1s after attempt 1
slept 2s after attempt 2
slept 4s after attempt 3
slept 8s after attempt 4
slept 16s after attempt 5
rc=1
```

6 attempts, 1+2+4+8+16 = 31 s, non-zero exit on exhaustion, and it breaks out correctly when a later attempt succeeds.

.ci/jenkins/lib/build-container-matrix.yaml
NirWolfer: Because this is the only push step that runs on a podman new enough to have the flag.

`--retry` was added in **podman 5.0** (it is absent from the `podman-push` man page at v4.9.3 and present at v5.0.0). The runners differ:

| Pipeline | Push step runs in | podman |
|---|---|---|
| `build-container-matrix` | `quay.io/podman/stable:v5.7.1` (`runs_on_dockers`) | 5.7.1 — has `--retry` |
| `test-matrix` / `test-dl-matrix` / `test-dl-ep-matrix` | `build_helper` | 4.9.x |
| `build-wheel-matrix` | `build_helper_(vllm\|sglang)` | 4.9.x |

Those `build_helper*` containers are all built from `.ci/dockerfiles/Dockerfile.build_helper`, which does `apt install -y sudo podman` on `dockerhub.nvidia.com/ubuntu:24.04`, and Ubuntu 24.04 ships podman 4.9.3. I confirmed this the hard way — the first revision of this PR used `--retry 5` everywhere and those jobs failed with:

```
+ podman push --retry 5 artifactory.nvidia.com/.../nixl-ci-gpu-test:3116
Error: unknown flag: --retry
```

Upgrading is not really an option on that base: podman 5.x is not in `noble`, `noble-updates` or `noble-backports` (it first appears in Ubuntu 25.10), and `Dockerfile.build_helper` deliberately pins apt to the internal mirror with the public archives removed.

`build-wheel-matrix` is worth calling out because it is genuinely confusing: its `Prepare` step symlinks `docker` to `podman` in *two* different containers, and the push in `Build sanity image` runs in the `build_helper_(vllm|sglang)` one, not the `manylinux` runner. I got that wrong initially and it failed a second time. The docs change in this PR records that, so the next person checks `containerSelector` against `runs_on_dockers` rather than assuming.

Happy to drop the flag and use the loop in `build-container-matrix` too if you would rather have one mechanism. I kept the native flag there because it is better behaved: podman only retries 502-504 and network errors and fails fast on auth, whereas the shell loop retries any failure, so a bad credential waits out the full 31 s instead of failing immediately.

dpressle: ok so can you change the command to be podman instead of docker because docker does not have --rtry and its confusing.

Reminder: we only have podman in the images and we link docker to podman because the contrib scripts only runs docker commands

dpressle (approved):

PR #2069, opened 2026-08-11
cuda_gds: include cuFile error details in error messages

changhyeonnam:
## What

Every cuFile error log in the GDS plugin now names the cuFile API that failed and includes both the error string from `CUFILE_ERRSTR` and the numeric error code.

This also fixes a misleading message in `submitBatch()`: a `cuFileBatchIOSubmit` failure logged "Error in setting up Batch", the same string used by the actual setup path in the `nixlGdsIOBatch` constructor, so the two stages were indistinguishable in the logs.

## Why

While benchmarking GDS over NVMe in P2PDMA mode, batch submission failed with nothing but this:

```
gds_utils.cpp:163] Error in setting up Batch
postXferReq: backend 'GDS' failed to post the transfer request with status NIXL_ERR_BACKEND
```

The real cause, a batch entry exceeding the cuFile P2PDMA per-entry size limit, was only visible in `cufile.log`. The "setting up" wording cost additional time, since setup had in fact succeeded.

The GDS_MT plugin already appends the numeric cuFile error code to its logs, so this brings the GDS plugin in line with it.

## How

Each message now names the failing cuFile API and appends the `CUFILE_ERRSTR` string along with the numeric code.

Verified on a B200 system with a Gen5 NVMe drive by rebuilding the plugin and reproducing each failure.

**Batch submission failure** (P2PDMA per-entry size limit exceeded)

Before:
```
gds_utils.cpp:163] Error in setting up Batch
```
After:
```
gds_utils.cpp:169] cuFileBatchIOSubmit failed: batch submit internal error - retry later (err=5040)
```

**Batch setup failure** (`--gds_batch_limit 256`, above the cuFile limit of 128), which previously printed the exact same message as the submission failure above

Before:
```
gds_utils.cpp:103] Error in setting up Batch
```
After:
```
gds_utils.cpp:107] cuFileBatchIOSetUp failed: invalid arguments (err=5022)
```

**File handle registration on a filesystem without GDS support** (overlayfs)

Before:
```
gds_utils.cpp:34] file register error:
```
After:
```
gds_utils.cpp:34] cuFileHandleRegister failed: GPUDirect Storage not supported on current file (err=5008)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA GDS error messages with the failing API, descriptive error text, and numeric error code.
  * Updated the reported GDS initialization warning to accurately identify driver-opening failures.


maheshrbapatu:
Looking 👀

brminich (approved):

brminich (dismissed):

changhyeonnam:

src/plugins/cuda_gds/gds_utils.cpp
```diff
@@ -31,7 +31,8 @@ nixl_status_t gdsUtil::registerFileHandle(int fd,
 
     status = cuFileHandleRegister(&handle, &descr);
     if (status.err != CU_FILE_SUCCESS) {
-        NIXL_ERROR << "file register error:";
+        NIXL_ERROR << "cuFileHandleRegister failed: " << CUFILE_ERRSTR(status.err)
```
vvenkates27: Optional suggestion: the same logging pattern is repeated in several places. A small helper would reduce duplication, for example:

void
logCuFileError(std::string_view api, CUfileOpError err) {
    NIXL_ERROR << api << " failed: " << CUFILE_ERRSTR(err) << " (err=" << err << ")";
}

void
logCuFileWarn(std::string_view api, CUfileOpError err) {
    NIXL_WARN << api << " failed - will use compat mode: " << CUFILE_ERRSTR(err)
              << " (err=" << err << ")";
}

vvenkates27 (dismissed):
Thanks for this PR — better cuFile error logs and updating the matching LogIgnoreGuard is appreciated. LGTM.

src/plugins/cuda_gds/gds_utils.cpp
changhyeonnam: @vvenkates27 thanks for the suggestion, added both helpers and switched all eight call sites over.

src/plugins/cuda_gds/gds_utils.cpp
```diff
@@ -14,9 +14,26 @@
...
+
+void
+logCuFileError(std::string_view api, CUfileOpError err) {
+    NIXL_ERROR << api << " failed: " << CUFILE_ERRSTR(err) << " (err=" << err << ")";
+}
+
+void
+logCuFileWarn(std::string_view api, CUfileOpError err) {
+    NIXL_WARN << api << " failed - will use compat mode: " << CUFILE_ERRSTR(err) << " (err=" << err
+              << ")";
```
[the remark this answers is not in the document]
changhyeonnam: Applied the suggestion in cd5a261d but it broke the clang-format check — clang-format-19 reflows these statements to fill the 100-column limit, which is the opposite of the suggested wrapping. Restored the formatter's output in the follow-up commit.

vvenkates27 (approved):

PR #2070, opened 2026-08-11
nixlbench: document GDS compat mode hang with large batch sizes

changhyeonnam:
## What?

Add a troubleshooting entry to the nixlbench README for GDS backend hangs in cuFile compat mode with large batch sizes.

## Why?

I was benchmarking HBM to NVMe transfers, sweeping batch sizes up to 128 to compare GDS against compat mode. The GDS (P2PDMA) runs completed the full sweep, but the compat runs kept hanging at large batch sizes: no result row, no error on screen, nothing in cufile.log at the default log level. The process had to be killed every time.

Finding the cause took a while: thread state inspection showed the main thread blocked inside cuFileBatchIOSubmit, and cufile.log at DEBUG level eventually revealed it was waiting forever for a free bounce buffer, because compat mode takes one POSIX pool buffer per in-flight batch entry and the default pool for the 1MiB slab class is 64 buffers. Raising posix_pool_slab_count resolved it.

The GDS plugin README already recommends a larger pool for compat mode, but nothing connects the hang symptom to that configuration. If this had been in the nixlbench troubleshooting section, this would have been a five minute fix instead of a debugging session, so adding it for the next person.

## How?

Documentation only. Follows the existing troubleshooting entry format and links to the GDS plugin README for the full recommended compat mode configuration.

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting guidance for hangs in GDS cuFile compatibility mode.
  * Documented batch-size thresholds, cuFile POSIX pool exhaustion, relevant DEBUG log indicators, and `cufile.json` slab-count adjustments.


maheshrbapatu:
Looking 👀

changhyeonnam:

PR #2075, opened 2026-08-11
CORE: Fix plugin related race.

ColinNV:
## What?
Prevents plugins to be unloaded when a call to `nixlAgent::getPluginParams()` runs concurrently to another operation that loads (and doesn't want to unload) a plugin.

Also improves code readability regarding the mutex, clarifies role of the unload function and removes the unused function for telemetry plugins.

## Summary by CodeRabbit

* **Improvements**
  * Improved backend plugin loading and discovery for more consistent behavior.
  * Enhanced retrieval of backend configuration and memory-related settings.
  * Streamlined plugin lifecycle handling to improve reliability and consistency.

* **Tests**
  * Updated plugin lifecycle tests to reflect the revised backend cleanup process.
  * Improved validation of backend loading, parameter retrieval, and cleanup behavior.


src/core/nixl_plugin_manager.cpp
```diff
@@ -334,7 +332,7 @@ shouldDeepBindPlugin(const std::string &plugin_name) {
 std::shared_ptr<const nixlPluginHandle>
 nixlPluginManager::loadPluginFromPath(const std::string &plugin_path,
                                       nixlPluginLoaderFunc loader,
-                                      bool deepbind) {
+                                      bool deepbind) const {
```
iyastreb: can it be static?

src/core/nixl_plugin_manager.cpp
```diff
@@ -462,7 +460,7 @@ nixlPluginManager::addPluginDirectory(const std::string &directory) {
 std::string
 nixlPluginManager::composePluginPath(const std::string &dir,
                                      const std::string &plugin_prefix,
-                                     const std::string &plugin_name) {
+                                     const std::string &plugin_name) const {
```
iyastreb: static?

src/core/nixl_plugin_manager.cpp
```diff
@@ -476,7 +474,17 @@ nixlPluginManager::composePluginPath(const std::string &dir,
...
+    const std::lock_guard lock(mutex_);
+
+    const auto plugin_handle = loadBackendPluginImpl(plugin_name);
+    if (plugin_handle) {
+        loaded_backend_plugins_.try_emplace(plugin_name, plugin_handle);
+    }
+    return plugin_handle;
+}
+
+std::shared_ptr<const nixlBackendPluginHandle>
```
iyastreb, line 486: Maybe better approach is to store the plugin shared_ptr in nixlEngineDeleter?

ColinNV: Yes, that's probably better, but I wanted to keep it separate and at the very least check carefully whether it is 100% safe.

iyastreb (approved):

Issue #1958, opened 2026-07-16
[BUG] agent::releaseXferReq() cannot support asynchronous request cleanup


lluki:
`nixlAgent::releaseXferReq()` assumes that `releaseReqH()` either completes the release synchronously or fails. A backend cannot report that cancellation has started but still needs polling, because `NIXL_IN_PROG` is treated like successful release rather than as a distinct result.

Problematic code: [`nixlAgent::releaseXferReq()` lines 1314-1330](https://github.com/ai-dynamo/nixl/blob/a7f2232318903bb37ea8428f25e620c07e25461e/src/core/nixl_agent.cpp#L1314-L1330).

## Current control flow

Assume a transfer is active and the backend needs several progress calls to cancel its work and drain its resources:

1. `releaseXferReq()` sees `req_hndl->status == NIXL_IN_PROG` and calls   `checkXfer()` once.
2. If `checkXfer()` still returns `NIXL_IN_PROG`, `releaseXferReq()` calls the   backend's `releaseReqH()`.
3. The result is tested only with `status < 0`. Every non-negative result,   including `NIXL_IN_PROG`, follows the success path: `backendHandle` is set to   `nullptr`, the outer request is deleted, and `releaseXferReq()` returns   `NIXL_SUCCESS`.

## What the backend can do today

If `releaseReqH()` returns `NIXL_SUCCESS`, the request is destroyed immediately.This is correct only when the backend has already stopped all work and released every resource synchronously. Returning success merely after *starting* asynchronous cancellation loses the request and backend handle, so the caller cannot poll the cleanup to completion.

If `releaseReqH()` returns a negative error, `releaseXferReq()` returns `NIXL_ERR_REPOST_ACTIVE` without deleting the request. However, the negative backend result is stored in `req_hndl->status`; `getXferStatus()` only calls `checkXfer()` while that status is `NIXL_IN_PROG`, so subsequent status checks do not progress cleanup. An error can therefore report that release failed, but cannot represent cleanup that is proceeding normally and needs another poll.

If `releaseReqH()` returns `NIXL_IN_PROG`, the current `< 0` check treats it as success. `releaseXferReq()` clears `backendHandle`, deletes the request, and returns `NIXL_SUCCESS`, even though the backend explicitly reported that the release is unfinished. This is the missing state needed for asynchronous cleanup.

## Suggested fix

[...]

PR #2077, opened 2026-08-12
posix: fix use-after-free when releasing an in-progress transfer

### LINKED ISSUE #1955: [BUG] POSIX releaseXferReq() causes use-after-free
The POSIX backend deletes an active request without removing its pending queue entries. A later queue poll can invoke a callback through the freed request handle leading to use-after-free.

This code path is possible:

1. `postXferReq()` queues POSIX I/O entries. Each entry stores the backend request handle as its callback context.
2. The user calls `releaseXferReq()` while the transfer is still `NIXL_IN_PROG`.
3. The agent calls the backend's `releaseReqH()` ([call site](https://github.com/ai-dynamo/nixl/blob/main/src/core/nixl_agent.cpp#L1305-L1328)).
4. POSIX unconditionally deletes the handle and returns `NIXL_SUCCESS` ([implementation](https://github.com/ai-dynamo/nixl/blob/main/src/plugins/posix/posix_backend.cpp#L341-L345)).
5. The queued/in-flight I/O entries remain in the shared backend queue. A later `poll()`, possibly for another transfer, reaps one and calls `ioDoneClb()` with the freed context.

This can cause a use-after-free. Returning success is also unsafe because the kernel may still be using buffers or file descriptors that the caller is now allowed to release.

### Expected behavior

If POSIX cannot abort an active transfer, `releaseXferReq()` should return an error and keep the request alive and pollable. Release should succeed only after all entries referencing the request have completed or been canceled.

### Additional potential problem with THREAD_SYNC_RW

With `NIXL_THREAD_SYNC_RW`, two threads may concurrently release request A and poll request B. Both requests share the same I/O queue. If A remains in progress, `releaseReqH()` deletes its handle without taking `io_queue_lock_` or draining/canceling its I/Os. A subsequent poll through B can reap A’s completion and invoke `ioDoneClb()` using the freed handle stored in `ctx_`, causing memory corruption or a crash.

### Fix

After some internal discussion we reached an agreement that `releaseReqH` should follow the following pseudocode:
```
releaseReqH() {
	if(no outstanding requests){
	   delete handle;
	   return SUCCESS;
	} else {
	   // Optional: Cancel (synchronous) outstanding operations
	   // Optional: Initiate asynchronous cancellation, correct clients keep polling in this case on checkXfer. Reap cancel completions there.
	   return ERR; 
	}
}
```

-> quarantined collection should not be necessary anymore as the backend can be written assuming that the handle stays alive.
-> Incorrect clients that do not check the return value will leak memory, but that's ok, better than uncoditional delete and use-after-free
-> This pattern is not yet very prevalent in the codebase, but we will work towards it.

This is separate from the error-recovery PR (#1942 and following): it concerns user-requested release of a normal active transfer.

rishabhsinha17:
## What?

Fix the use-after-free from #1955: `nixlPosixEngine::releaseReqH()` deleted the backend request unconditionally, even while its I/O queue entries were still queued or in flight. Each entry stores the request handle as its completion callback context, so a later `poll()`, possibly for a different transfer on the shared queue, invoked `ioDoneClb()` through the freed pointer.

Revised per the review discussion: releasing an in-flight transfer is now accepted and the free is deferred until the request's outstanding I/Os complete, instead of refused as in the first revision.

## Why?

Fixes #1955. Implements the direction from the maintainer discussion in this PR: reference counting on the handle, flag flavor, so the handle is freed once its outstanding requests return.

The agent-side half is the negative-status branch of `nixlAgent::releaseXferReq()`, which #1958 describes as well: a negative `releaseReqH()` result was stored in `req_hndl->status`, and `getXferStatus()` only calls `checkXfer()` while that status is `NIXL_IN_PROG`, so after a failed release the request could never be driven to completion. POSIX no longer fails a release, but backends without cancellation or deferral support may, and this branch keeps them recoverable. #1959 changed the adjacent `NIXL_IN_PROG` branch of the same function and is closed unmerged; the two changes are complementary and do not overlap.

Distinct from the io_uring error-recovery series (#1942), which covers internal I/O failures and whose scope note leaves user-invoked cancellation (#1955) out. Once #1942's `cancel(void *ctx)` lands, `releaseReqH()` can additionally cancel the outstanding I/Os so the deferred free happens sooner; parking remains the fallback for queues without cancellation support.

## How?

- `nixlPosixBackendReqH::allConfirmed()` (new): true once every submitted I/O of this request has been confirmed, i.e. no queue entry references the request anymore. `num_confirmed_ios_` only advances inside the completion callback, the callback only runs inside `poll()`, and every prep/post/poll site holds `io_queue_lock_`, so evaluating the predicate under that lock is exact.
- `nixlPosixEngine::releaseReqH()`: under `io_queue_lock_`, a request that is `allConfirmed()` is freed on the spot; otherwise it is parked on the engine's `released_reqs_` list, which is the "user dropped his reference, free after completion" flag. Both paths return `NIXL_SUCCESS`. Previously `releaseReqH()` took no lock at all.
- `nixlPosixEngine::reapReleasedReqs()` (new): runs after the poll in `checkXfer()`, under the same `io_queue_lock_`, and frees every parked request whose last completion has been confirmed. Deletion therefore never happens inside a queue callback and never concurrently with a poll. The engine destructor frees whatever is still parked, so a released request cannot outlive the engine.
- `~nixlPosixIOQueueAIO()`: the POSIX AIO destructor cancelled every pool entry through possibly stale fds and never waited for operations `aio_cancel()` reports as `AIO_NOTCANCELED`, which keep referencing their `aiocb` after `ios_` is destroyed. It now cancels only in-flight operations, waits for non-cancelable ones with `aio_suspend()`, and reaps each with `aio_return()`, without invoking completion callbacks. The libaio and io_uring queues were already safe: `io_destroy()` blocks until non-canceled AIOs finish, and `io_uring_queue_exit()` tears the ring down without delivering CQEs.
- `nixlAgent::releaseXferReq()`: stop overwriting `req_hndl->status` with the backend release result, and document the release contract on `releaseXferReq()`: backends that cannot cancel in-flight operations either defer the free internally and return success, or refuse with an error, in which case the request stays `NIXL_IN_PROG` and pollable so the caller can complete it and retry.

Semantics note: a deferred release returns `NIXL_SUCCESS` while the kernel may still be reading or writing the transfer's buffers and files. That matches the pre-existing `releaseXferReq()` contract on `main`, where release of an active transfer already returned success, minus the crash; callers that need the data settled must poll to completion before releasing, which the updated `releaseXferReq()` documentation spells out.

Known limitation: on paths where a queue stops delivering completion callbacks, `num_confirmed_ios_` never reaches `queue_depth_` and a released request stays parked until engine teardown. Two reachable cases: an `enqueue()` failure part-way through `postXfer()`, and Linux AIO's `doCheckCompleted()` returning `NIXL_ERR_BACKEND` on `events[i].res < 0` before invoking `io->clb_`. The request is still freed in the engine destructor, so it no longer leaks; draining those error paths is what #1942 and follow-ups are for.

Tests: `test_posix_release_active` and `test_posix_release_teardown` in `test/unit/plugins/posix/nixl_posix_test.cpp`. The agent runs without a progress thread and every POSIX queue caps submission and reaping at 64 entries per poll, so with 256 descriptors the single poll inside `releaseXferReq()` cannot complete the transfer and the release deterministically hits an active transfer. Asserts: the in-flight release returns `NIXL_SUCCESS`; a second write on the same queue completes and is released, its polls reaping the released transfer's completions, which a `-Dsanitizer=address,undefined` build reports as heap-use-after-free in `ioDoneClb()` against the unfixed backend; and a read-back validates file contents. The teardown test instead destroys the agent right after the release, with buffers and files declared to outlive it, covering the destructor path that frees a parked request whose completions were never reaped. Leak checking at exit verifies the parked requests were actually freed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented active transfer requests from being released prematurely.
  * Requests that cannot yet be released remain available for polling and can be released after completion.
  * Improved error reporting while preserving request handles when release is refused.
  * Ensured in-progress POSIX operations are safely completed and cleaned up during shutdown.

* **Documentation**
  * Clarified backend-dependent behavior when releasing active transfers.

* **Tests**
  * Added coverage for active request release, deferred cleanup, subsequent transfers, and data integrity.


src/plugins/posix/posix_backend.cpp
```diff
@@ -339,7 +357,19 @@ nixlPosixEngine::checkXfer(nixlBackendReqH *handle) const {
...
+            return status;
+        }
+    }
+    catch (const nixlPosixBackendReqH::exception &e) {
+        NIXL_ERROR << e.what();
+        return e.code();
+    }
+
     delete handle;
     return NIXL_SUCCESS;
```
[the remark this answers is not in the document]
rishabhsinha17: The race this describes cannot occur once releaseXfer() has returned NIXL_SUCCESS. num_confirmed_ios_ is incremented only from ioDoneClb, ioDoneClb only runs inside io_queue_->poll(), and every site that polls, posts, or preps holds io_queue_lock_ (nixlPosixEngine::checkXfer, postXfer, prepXfer). A successful check under the lock therefore means every queue entry referencing this request has already been reaped, so a later poll() from another thread has nothing pointing at this request left to dereference.

The interleaving where guard placement would matter is checkXfer() or postXfer() on the same handle racing its own release. That is a caller-side use-after-release, and moving the delete inside the guard does not close it: the racing thread blocks on the lock, then dereferences the freed handle after unlock. Since ~nixlPosixBackendReqH is empty and io_queue_ is a reference to the engine-owned queue, the delete touches no shared state; keeping it outside the guard keeps the critical section minimal.

src/plugins/posix/posix_backend.h
```diff
@@ -59,6 +59,8 @@ class nixlPosixBackendReqH : public nixlBackendReqH {
     prepXfer();
     nixl_status_t
     checkXfer();
+    nixl_status_t
+    releaseXfer() const;
```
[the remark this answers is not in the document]
rishabhsinha17: Added in 7e4cb3f.

src/core/nixl_agent.cpp
```diff
@@ -1326,13 +1326,15 @@ nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
 
         if(req_hndl->status == NIXL_IN_PROG) {
 
-            req_hndl->status = req_hndl->engine->releaseReqH(
-                                         req_hndl->backendHandle);
+            const nixl_status_t rel_status = req_hndl->engine->releaseReqH(req_hndl->backendHandle);
```
iyastreb, line 1301: Currently many NIXL backends do not support invalidation of in-flight handles, so imo this is the right change.
I propose to document this behavior in the releaseXferReq description

lluki:
So in essence, we are replacing a use-after-free with a memory leak: If client code drops a request handle without completing, we will call into `releaseReqH` which can't free the handle, keeps it alive and the client code loses all it's references to it. The PR does improve the situation when someone is explicitly calling it (by returning an error).

I am unsure what is the correct behavior, but i think leaking memory is not great either, I think we should do either
a) reference counting on the handle (and freeing the handle after the requests return)
b) panic in the case we are leaking.

I am trying to push for a cancellation design  that would separate the "cancel" concern from the releaseReq and have releaseReq panic in case of outstanding requests. However this is a change client code might observe and we need to be careful with it to maintain compatibility.

Overall i think this PR improves the situation but we should discuss what is the actually correct approach.

lluki:
Hi @rishabhsinha17 we had some more discussions internally, we prefer to implement reference counting on the handle. Such that eventually, when the requests return, the handle is being freed.

I personally dont mind how we implement the ref counting, either with a smart pointer from C++ or also a simple flag we check on completion would be ok (where the flag means: "user dropped his reference, free after all outstanding reqs complete")

rishabhsinha17:
@lluki Implemented in 2e13966 with the flag flavor.

`releaseReqH()` now always accepts the release. A request with no outstanding I/Os is freed on the spot. Otherwise it is parked on an engine-side `released_reqs_` list, which is the flag, and `reapReleasedReqs()`, running after every poll under the same `io_queue_lock_` that already serializes all prep/post/poll sites, frees it once its last completion is confirmed. Deletion never happens inside a queue callback and never concurrently with a poll. The engine destructor frees anything still parked, which also covers the error paths where a queue stops delivering callbacks (partial `enqueue()` failure, AIO's early return on `events[i].res < 0`), so those degrade to a request held until teardown rather than a leak.

User-visible behavior is back to release-always-succeeds. The agent-side `NIXL_ERR_REPOST_ACTIVE` branch stays for backends that can neither cancel nor defer, and the contract is now documented on `releaseXferReq()` per @iyastreb. The regression test asserts the in-flight release succeeds, drives the queue with a second write so the orphaned completions are reaped (ASAN reports the old delete-on-release behavior as heap-use-after-free there), and leak checking at exit verifies the parked request is actually freed. PR description updated to match.

src/core/nixl_agent.cpp
rishabhsinha17: Documented in 2e13966: releaseXferReq() Doxygen now spells out the defer-or-refuse contract for backends that cannot cancel in-flight operations, including that on refusal the handle stays valid and pollable for retry. With this revision the POSIX backend takes the defer path, so this branch is exercised only by backends without cancellation or deferral support.

src/plugins/posix/posix_backend.cpp
```diff
@@ -224,6 +224,14 @@ nixlPosixEngine::nixlPosixEngine(const nixlBackendInitParams *init_params)
                                  io_queue_type_);
 }
 
+// Released requests whose completions were never fully reaped are still parked here;
+// nothing can poll the queue once the engine is going away, so free them now.
+nixlPosixEngine::~nixlPosixEngine() {
+    for (auto *req : released_reqs_) {
+        delete req;
+    }
+}
```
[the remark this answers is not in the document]
rishabhsinha17: Fixed in 9df02b5. `~nixlPosixIOQueueAIO` now iterates `ios_in_flight_` instead of every pool entry (the old loop also cancelled recycled entries through stale fds), waits for operations `aio_cancel()` reports as `AIO_NOTCANCELED` via `aio_suspend()`, and reaps each with `aio_return()` before `ios_` is destroyed, without invoking completion callbacks, whose context may already be gone at teardown. The other two queues are safe as-is: `io_destroy()` blocks until non-canceled AIOs finish, and `io_uring_queue_exit()` tears the ring down without delivering CQEs.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -779,6 +779,224 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+        }
+    } while (status == NIXL_IN_PROG);
+
+    status = agent.releaseXferReq(treq_drain);
+    if (status != NIXL_SUCCESS) {
+        std::cerr << "Failed to release completed drain request - status: "
+                  << nixlEnumStrings::statusStr(status) << std::endl;
+        return 1;
+    }
+    std::cout << "Second write completed; orphaned completions reaped" << std::endl;
```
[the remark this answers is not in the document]
rishabhsinha17: Added in 9df02b5: test_posix_release_teardown posts the same 256-descriptor write, releases it while active, and destroys the agent with no further transfer. Buffers and files are declared outside the agent's scope since teardown may still be draining kernel I/O that targets them. ASAN covers the destructor path freeing the never-confirmed parked request; leak checking at exit covers the free itself.

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -57,10 +57,18 @@ class nixlPosixIOQueueAIO : public nixlPosixIOQueueImpl<nixlPosixAioIO> {
...
+    // before ios_ is destroyed, without invoking completion callbacks: at teardown their
+    // context may already be gone.
+    for (auto *io : ios_in_flight_) {
+        if (aio_cancel(io->aio_.aio_fildes, &io->aio_) == AIO_NOTCANCELED) {
+            const struct aiocb *cbs[] = {&io->aio_};
+            while (aio_error(&io->aio_) == EINPROGRESS) {
+                aio_suspend(cbs, 1, nullptr);
+            }
         }
+        aio_return(&io->aio_);
```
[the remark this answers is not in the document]
rishabhsinha17: Fixed in f5f9f5c. The destructor now drains unconditionally: for every in-flight entry it calls aio_cancel() as a hint, then waits while aio_error() returns EINPROGRESS via aio_suspend(), then reaps with aio_return(), regardless of what aio_cancel() returned. Covers AIO_NOTCANCELED, a failed cancel (EBADF and friends), AIO_CANCELED (aio_error goes to ECANCELED, the wait loop is skipped) and AIO_ALLDONE alike, with exactly one aio_return() per entry.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -779,6 +779,346 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+    std::vector<std::unique_ptr<void, PosixMemalignDeleter>> dram_addr;
+    dram_addr.reserve(num_transfers);
+    std::vector<tempFile> fd;
+    fd.reserve(num_transfers);
+    std::unique_ptr<nixlBlobDesc[]> dram_buf(new nixlBlobDesc[num_transfers]);
+    std::unique_ptr<nixlBlobDesc[]> ftrans(new nixlBlobDesc[num_transfers]);
+    nixl_reg_dlist_t dram_for_posix(DRAM_SEG);
+    nixl_reg_dlist_t file_for_posix(FILE_SEG);
+    nixl_xfer_dlist_t dram_for_posix_xfer(DRAM_SEG);
+    nixl_xfer_dlist_t file_for_posix_xfer(FILE_SEG);
```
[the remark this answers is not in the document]
rishabhsinha17: Fixed in f5f9f5c. Both tests now build their buffers, files and descriptor lists through a shared setup declared before the agent, so destruction order matches test_posix_release_teardown's stated principle in both: an incompletely drained queue can no longer target freed buffers or closed descriptors.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -779,6 +779,346 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+        dram_buf[i].devId = 0;
+        dram_for_posix.addDesc(dram_buf[i]);
+        dram_for_posix_xfer.addDesc(dram_buf[i]);
+
+        ftrans[i].addr = 0;
+        ftrans[i].len = transfer_size;
+        ftrans[i].devId = fd[i].fd;
+        file_for_posix.addDesc(ftrans[i]);
+        file_for_posix_xfer.addDesc(ftrans[i]);
+    }
```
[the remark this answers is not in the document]
rishabhsinha17: Done in f5f9f5c: the duplicated loop moved into setup_release_test_buffers(), with a releaseTestBuffers holder owning the buffers, files and the four descriptor lists; the tests differ only in the infix ("_ra_"/"_rt_") they pass.

test/unit/plugins/posix/nixl_posix_test.cpp
```diff
@@ -779,6 +779,346 @@ test_posix_repost (std::string test_files_dir_path_abs_path, bool use_uring) {
...
+
+    do {
+        status = agent.getXferStatus(treq_drain);
+        if (status < 0) {
+            std::cerr << "Error during drain transfer - status: "
+                      << nixlEnumStrings::statusStr(status) << std::endl;
+            agent.releaseXferReq(treq_drain);
+            return 1;
+        }
+    } while (status == NIXL_IN_PROG);
```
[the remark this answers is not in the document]
rishabhsinha17: Leaving this one as is: every poll loop in this file (write, read, repost tests) is an unbounded do/while on getXferStatus(), so bounding only these two would be inconsistent, and the CI job timeout already converts a hang into a failure. Happy to add a wall-clock bound across the file in a follow-up if the maintainers want one.

rishabhsinha17:
@lluki circling back: the flag-flavor reference counting you asked for has been in since 2e13966 (current head ab01410, rebased onto main, all pre-CI checks green, CodeRabbit threads resolved). Anything else you want changed before someone kicks off the NVIDIA CI? The sanitizer job should show the before/after directly on the two regression tests.

rishabhsinha17:
@iyastreb the reference-counting rework lluki asked for has been in for a week (flag flavor, parked-list sweep under io_queue_lock_), rebased on main, all pre-CI checks green and CodeRabbit threads resolved. Could you or another maintainer authorize the copy-pr-bot run? nixl-ci-test-sanitizers is the job that shows the before/after directly on the release-active regression tests.

src/plugins/posix/posix_backend.cpp
```diff
@@ -337,11 +347,47 @@ nixlPosixEngine::checkXfer(nixlBackendReqH *handle) const {
...
+        if (posix_handle.allConfirmed()) {
+            delete handle;
+        } else {
+            NIXL_DEBUG << "POSIX request released while in progress; deferring free until its "
+                          "outstanding I/Os complete";
+            released_reqs_.push_back(&posix_handle);
+        }
+        return NIXL_SUCCESS;
+    }
+    catch (const nixlPosixBackendReqH::exception &e) {
```
lluki: How can this happen?

src/plugins/posix/posix_backend.cpp
```diff
@@ -224,6 +224,14 @@ nixlPosixEngine::nixlPosixEngine(const nixlBackendInitParams *init_params)
                                  io_queue_type_);
 }
 
+// Released requests whose completions were never fully reaped are still parked here;
+// nothing can poll the queue once the engine is going away, so free them now.
+nixlPosixEngine::~nixlPosixEngine() {
+    for (auto *req : released_reqs_) {
+        delete req;
```
lluki, line 296: Print error message in this case - it should never occur. But it's ok in the sense of defensive coding

src/plugins/posix/posix_aio_io_queue.cpp
```diff
@@ -57,10 +57,18 @@ class nixlPosixIOQueueAIO : public nixlPosixIOQueueImpl<nixlPosixAioIO> {
...
+    // aio_cancel() is only a hint and can itself fail (e.g. EBADF): any operation it does
+    // not report as done stays active and keeps referencing its aiocb. Whatever the
+    // cancel result, wait until the operation leaves EINPROGRESS and reap it before ios_
+    // is destroyed, without invoking completion callbacks: at teardown their context may
+    // already be gone.
+    for (auto *io : ios_in_flight_) {
+        aio_cancel(io->aio_.aio_fildes, &io->aio_);
+        const struct aiocb *cbs[] = {&io->aio_};
+        while (aio_error(&io->aio_) == EINPROGRESS) {
+            aio_suspend(cbs, 1, nullptr);
```
lluki: I think the old version of the destructor was ok. We dont want to delay the destructor either so i think the hint only behavior was good. We should print an error however

lluki (changes requested):
Thanks for your contribution @rishabhsinha17  

* We will merge your PR - I think the explicit tracking of the orphaned IOs is better than my PR
* rebase, use the new doCompleted() that also accounts for in-flight cancellations instead of your allCompleted()
* Please compact the comments to 1 lines (unless it is exceptionally complicated) - doxygen docstring is OK
* More comments inline

rishabhsinha17:
Reworked per the review, thanks @lluki.

* Rebased onto main as a single commit; the branch now includes #1942.
* The release/park gate and the reap use main's new `isComplete()` (the accessor that also accounts for in-flight cancellations), so a parked request is not freed while cancel completions are still expected. `allConfirmed()` is gone.
* Comments compacted to 1-2 lines each; the `releaseXferReq()` Doxygen contract from the earlier review round stays.
* `releaseReqH()` lost the try/catch: a null handle is now rejected explicitly with an error log (details in the inline reply).
* `~nixlPosixEngine()` logs an error when requests are still parked at teardown, then frees them.
* `~nixlPosixIOQueueAIO()` is back to main's hint-only `aio_cancel()`, with an error log when operations are still in flight.

src/plugins/posix/posix_backend.cpp
rishabhsinha17: Only for a null handle: `castPosixHandle()` throws just on nullptr, the cast itself cannot fail. Made it self-evident in the rework: the try/catch is gone and `releaseReqH()` rejects a null handle explicitly with an error log and `NIXL_ERR_INVALID_PARAM`.

src/plugins/posix/posix_backend.cpp
rishabhsinha17: Done: the destructor now logs an error with the count of still-parked requests before freeing them.

src/plugins/posix/posix_aio_io_queue.cpp
rishabhsinha17: Restored main's hint-only `aio_cancel()` over `ios_` and added an error log when `ios_in_flight_` is non-empty. For the record, the drain was there because AIO_NOTCANCELED operations are completed by glibc worker threads that write status into the aiocbs owned by `ios_`, so a teardown with stragglers can still see such writes after the pool is freed. That is pre-existing main behavior, bounded by process teardown; the error log now surfaces it.

lluki (dismissed):

src/api/cpp/nixl.h
```diff
@@ -378,8 +378,13 @@ class nixlAgent {
                           nixlBackendH* &backend) const;
 
         /**
-         * @brief  Release the transfer request `req_hndl`. If the transfer is active,
-         *         it will be canceled, or return an error if the transfer cannot be aborted.
+         * @brief  Release the transfer request `req_hndl`. If the transfer is active, the
```
iyastreb: There is ongoing discussion about adding API for request cancellation
I would propose no not touch a common NIXL API in this PR.
Can we create PR just for NIXL API changes and review it with architects?
This affects overall NIXL design, better make it a separate small focused PR

aranadive (dismissed):

src/api/cpp/nixl.h
aranadive: This is only a comment update, no? There is no API update here. Is the concern the behavior that req_hndl could now remain valid after the calling releaseXferReq?

ovidiusm: The change in nixl_agent.cpp is trivial, it does not need this comment. If we want to merge this for 1.5.0, I suggest we drop the public API comment change since it is not necessary, and it would require additional approvals

vvenkates27: +1 on the comment update. cc: @mkhazraee 

The new comment mixes two different objects and two different outcomes, so a caller can easily get the lifetime wrong.

On NIXL_SUCCESS, the agent handle is destroyed immediately (delete req_hndl). Only backend cleanup may continue asynchronously. A reader of “freeing the request internally once its outstanding operations complete” can think req_hndl is still valid and that they can still call getXferStatus() after a successful release which they cannot.

Probably something like
Release req_hndl. On success the handle is destroyed; if I/O is still
active, backend cleanup may continue asynchronously. On error the handle
stays valid — poll with getXferStatus() and retry.

Not a functional blocker — please tighten the public API contract so success vs error is unambiguous.

vvenkates27 (commented):
Thanks for the PR, Overall this PR LGTM, just one comment about the description modfication.

src/api/cpp/nixl.h
rishabhsinha17: Adopted @vvenkates27's wording verbatim in e6d8bd4f, thanks.

@aranadive correct, comment-only. The nixl.h delta vs main is now just that one doxygen block (+3/-2 lines); with comments stripped the file is identical to main. No API or behavior change.

@ovidiusm if the public API comment still complicates the 1.5.0 approval chain, happy to drop the nixl.h hunk entirely and re-propose it as a docs-only follow-up after the release. Maintainers' call.

New head e6d8bd4f may need a fresh /ok to test.

lluki (changes requested):
Alright, our internal cancellation discussion has found an end: We will stick with the approach that `releaseReqH` does cancel and release handle if possible after all. Hence, the approach of this PR is now considered wrong. We should simply not delete the backend handle if there are outstanding requests. This gives the Northbound Client a chance to continue polling. In pseudocode `releaseReqH` should work as follows:

```
releaseReqH() {
  if(requests outstanding) {
    return error - make sure northbound API returns err 
  }
  // no outstanding requests
  delete handle
  return success
}
```

Also, let's not change *anything* in nixl_agent.cpp - We agree the comment has to be changed but lets do it in a separate PR. 

I think at this point it would be best to close this PR and open another one. Sorry for the long review process.

src/api/cpp/nixl.h
```diff
@@ -378,8 +378,9 @@ class nixlAgent {
                           nixlBackendH* &backend) const;
 
         /**
-         * @brief  Release the transfer request `req_hndl`. If the transfer is active,
-         *         it will be canceled, or return an error if the transfer cannot be aborted.
+         * @brief  Release req_hndl. On success the handle is destroyed; if I/O is still
+         *         active, backend cleanup may continue asynchronously. On error the handle
+         *         stays valid — poll with getXferStatus() and retry.
```
brminich, line 383: imo it makes it more confusing that it used to be

brminich (changes requested):

rishabhsinha17:
Thanks for closing out the internal discussion. Opened #2252 on top of current main with the semantics from your review, releaseReqH refuses the release while I/Os are outstanding, and deletes the handle only once the request is complete. The regression test from here carries over with flipped expectations. nixl_agent.cpp and the nixl.h docs are unchanged here, the comment fix can be its own PR as you suggested. Closing this one.

PR #2081, opened 2026-08-12
Mooncake plugin: fix checkXfer use-after-free, postXfer null deref, and an uninitialized test length

xiaodouzi666:
Three independent fixes in the Mooncake plugin and its unit test, found while
working on TENT-mode support (RFC: #2083).

1. **checkXfer crashes when polled again after completion.** The batch is
   freed on completion but `request_count` is not reset, so a later poll
   dereferences the stale batch id as a descriptor pointer. Clear both fields
   and guard the entry.
2. **postXfer dereferences `opt_args` without a null check.** The SB API
   declares the parameter nullable.
3. **`allocateWrongGPUTest` reads `desc.len` uninitialized** before
   `allocateBuffer()` uses it as the allocation size (found via a random-size
   registration failure on a multi-GPU machine).

Verified on an 8×H20 / 5×RoCE-HCA server (Ubuntu 22.04, CUDA 13): upstream
`mooncake_backend_test` passes its DRAM and VRAM transfer sections (599 data
verifications, 300 notification checks). Note: the final `allocateWrongGPUTest`
assertion itself expects `NIXL_ERR_NOT_SUPPORTED`, which the plugin has never
returned — that pre-existing test/plugin mismatch aborts the run on any
multi-GPU machine and is unrelated to these fixes; happy to address it in a
follow-up once the intended semantics are clarified.

## Bug Fixes

* Improved reliability when checking completed transfer batches.
* Notifications now handle missing optional arguments safely.
* Request and batch state is initialized and reset consistently.

## Tests

* Updated GPU buffer allocation coverage with an explicit buffer size.

### DIFF
--- a/src/plugins/mooncake/mooncake_backend.cpp
+++ b/src/plugins/mooncake/mooncake_backend.cpp
@@ -281,7 +281,7 @@ nixlMooncakeEngine::postXfer(const nixl_xfer_op_t &operation,
         request[index].target_id = segment_id;
     }
     int rc = 0;
-    if (opt_args->hasNotif) {
+    if (opt_args && opt_args->hasNotif) {
         notify_msg_t notify_msg;
         notify_msg.name = const_cast<char *>(local_agent_name_.c_str());
         notify_msg.msg = const_cast<char *>(opt_args->notifMsg.c_str());
--- a/src/plugins/mooncake/mooncake_backend.cpp
+++ b/src/plugins/mooncake/mooncake_backend.cpp
@@ -226,8 +226,8 @@ struct nixlMooncakeBackendReqH : public nixlBackendReqH {
 
     virtual ~nixlMooncakeBackendReqH() {}
 
-    uint64_t batch_id;
-    size_t request_count;
+    uint64_t batch_id = INVALID_BATCH;
+    size_t request_count = 0;
 };
 
 nixl_status_t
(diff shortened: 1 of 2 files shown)


alogfans (commented):
LGTM. A comment: checkXfer() on a never-posted handle now returns NIXL_SUCCESS; this seems intentional, please confirm that matches the backend contract.

xiaodouzi666:
Confirmed, it is intentional — and thanks for pinning it down, the never-posted case deserves an explicit answer.

The guard covers two states that both mean "no outstanding work on this handle": the batch was reclaimed after every request completed (the crash this PR fixes — a stale batch id was handed back to the engine, which casts it to a descriptor pointer and dereferences it), and a handle that was created but never posted. Reporting a terminal state for both keeps checkXfer() idempotent, which is what a polling caller expects.

BackendGuide.md doesn't specify a return value for a never-posted handle, so I went with NIXL_SUCCESS for consistency with the completed-batch case. If you'd rather have the two distinguished — e.g. NIXL_ERR_INVALID_PARAM for a handle that was never posted, keeping NIXL_SUCCESS only for a reclaimed batch — that's a small change and I'm happy to make it; it would need a separate "was posted" flag on the handle since both states currently share the invalid batch id. Let me know which you prefer.

ovidiusm (approved):

PR #2015, opened 2026-07-30
[nixlbench] add worker part for device api

fteng-NV:
## What?
Integrate Device API into nixlbench so transfers can be initiated from a CUDA kernel instead of the host path.
Adds --use_device_api to enable GPU-kernel-initiated PUT transfers. When enabled, --num_threads is repurposed as the CUDA kernel block size, and the internal CPU thread count is forced to 1.
This builds on the kernel launch support landed in #1761.

## Why?
nixlbench could not exercise Device API end-to-end now. Wire the existing device kernels into the worker so we can benchmark and validate GPU-initiated WRITE transfers under a supported configuration.

## How?
- Add --use_device_api and validate required options at config parse time (UCX, VRAM↔VRAM, --enable_pt, pairwise, SG, WRITE, pipeline_depth=1, and a build with HAVE_UCX_GPU_DEVICE_API).

- In the NIXL worker, prepare local/remote nixlMemViews, allocate/exchange a VRAM completion counter, and launch nixlbenchLaunchDevicePut for the transfer path.

- On the target, wait on the device completion/error counters instead of host-side request completion.

- Keep the existing host transfer path unchanged when --use_device_api is not set.

## Summary by CodeRabbit

* **New Features**
  * Added an optional NIXL Device API mode for GPU-based transfer benchmarking.
  * Added GPU-resident completion tracking and automatic monitoring for warmup and timed transfers.
  * Added GPU memory preparation and cleanup for local and remote transfer views.
  * Added validation and status reporting for device API settings, including CUDA block size and supported transfer configurations.

* **Bug Fixes**
  * Device transfers no longer require remote completion signaling to be explicitly enabled.
  * Completion and error signaling now consistently use the remote counter region.

PR #2084, opened 2026-08-13
[nixlbench] device api improvements

fteng-NV:
## What?

Improve NIXLBench Device API execution by:

- Assigning each GPU execution group to a dedicated UCX device channel.
- Configuring enough UCX device channels for the number of GPU execution groups.
- Moving the benchmark iteration loop into the GPU kernel, avoiding one kernel launch per iteration.
- Collecting per-iteration timing directly on the GPU and reporting it through the existing benchmark statistics.
- Avoiding status polling when posting a transfer has already failed.

## Why?

The previous implementation [https://github.com/ai-dynamo/nixl/pull/2015](url) used a single UCX device channel and launched the GPU kernel separately for every benchmark iteration. This limited channel-level parallelism and added host-side kernel launch and synchronization overhead to the measured execution.

These changes allow independent GPU execution groups to use separate UCX channels and keep the complete iteration loop on the device, providing better concurrency and more representative Device API performance measurements.

## How?

- Use the CUDA thread/warp group ID as the `channel_id` passed to `nixlPut`.
- Set `ucx_num_device_channels` to at least the number of GPU execution groups.
- Pass the iteration count and a device timing buffer to the kernel.
- Execute all transfer iterations within a single kernel launch and record each iteration using the GPU global timer.
- Copy timing samples back to the host only when per-iteration statistics are requested.
- Signal the total number of completed iterations through the remote completion counter after all iterations finish.

## Result
Two ranks on two nodes, each use One GPU, --num-iter 4096, --num-threads 4
- device api --num-threads=4
image

- CPU --num-threads=4
image

## New Features
- Device transfer benchmarks support multiple iterations per operation.
- Configurable device channels and execution groups improve transfer scaling.
- Measured runs collect per-iteration posting and transfer durations.
- Completion and error reporting reflects all configured iterations.

## Bug Fixes
- Added validation for incompatible channel settings and launch configurations.
- Warmup runs no longer collect unnecessary iteration statistics.
- Improved validation for worker counts, iteration settings, buffer alignment, and execution-group inputs.
- Device duration measurements are validated and reported consistently in microseconds.


hbadihi:
@fteng-NV 

I think we should aim to have Device API timing closer to the CPU path semantically :

- CPU Post = time inside postXferReq()
- CPU Tx = time from postXferReq() returning until completion
- Device Tx currently combines time inside nixlPut() and polling, while Post is always zero.

Maybe for our Device API measure and report separately:
```
Post = time inside nixlPut()
Tx   = time from nixlPut() returning until nixlGpuGetXferStatus() succeeds
```
For batches/multiple groups, each sample should cover the whole iteration. Prep can remain zero because there is no separate device equivalent of createXferReq().

*This is all assuming the measurment cost is not that high please correct me if its not the case and the global timer is costly and might hurt performance

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -62,13 +70,15 @@ nixlbenchSignalCounter(const nixlbenchDeviceXferParams &params,
 }
 
 __device__ nixl_status_t
-nixlbenchSignalCompletion(nixlbenchDeviceXferParams params) {
-    return nixlbenchSignalCounter(params, params.completionCounterOffsetBytes, 1ull, "completion");
+nixlbenchSignalCompletion(nixlbenchDeviceXferParams params, uint64_t num_iterations) {
+    return nixlbenchSignalCounter<nixl_gpu_level_t::THREAD>(
```
ofirfarjun7: IMO level should be test param. And we should call all APIs with same level

fteng-NV: I updated nixlbenchSignalCompletion and nixlbenchSignalError to apply template Level in [b4d433d](https://github.com/ai-dynamo/nixl/pull/2084/commits/b4d433dc76708e99d34bb74f4e20d8c279ce33be). 
But in current scenario we only need one thread to nixlAtomicAdd comp/ error counter for target side notify. I think using Thread only fits our needs now.

hbadihi:
@fteng-NV 
Device `Avg Lat.` being smaller in your example than Post/Tx doesn't make sense to me.
I think the problem requires 2 fixes :

(1) Work distribution mismatch we have between CPU and Device

`num_iters` is pretty misleading naming by nixlbench, it behave more as the total number of transfers then iterations.


For Example:

With 1,000 iterations, 4 workers, and batch size 8:

```text
CPU thread: 250 iterations × 8 regions
GPU group:  1000 iterations × 2 regions
```

CPU divides iterations among threads, while Device divides each batch among groups. 


(2) `printStats` function is used in Device aPI but does not consider `block_threads` and uses the `num_threads` that you forced to 1, leads to wrong report

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -87,37 +98,72 @@ nixlbenchPutKernel(nixlbenchDeviceXferParams params) {
...
         num_groups = (blockDim.x + warpSize - 1) / warpSize;
     }
-
     nixlGpuXferStatusH &xfer_status = xfer_statuses[group_id];
-    nixl_status_t put_status = NIXL_SUCCESS;
-    for (size_t region_idx = group_id; region_idx < params.numRegions; region_idx += num_groups) {
-        put_status = nixlbenchPostPut<Level>(params, region_idx, xfer_status);
-        if (put_status != NIXL_IN_PROG) {
-            break;
+    const bool active_group = group_id < params.activeGroupNum;
```
hbadihi: @fteng-NV I think the change in `num_iters` mentioned in different comment should make this whole active group thing redundant as all groups should be active afterwards because they are dependent on number of iters. if num_groups > num items we should throw an error before the kernel starts. it will simplify the kernel

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -87,37 +98,72 @@ nixlbenchPutKernel(nixlbenchDeviceXferParams params) {
...
+            printf("[nixlbenchPutKernel] transfer did not complete: "
+                   "threadIdx.x=%u blockIdx.x=%u blockDim.x=%u final_status=%d\n",
+                   threadIdx.x,
+                   blockIdx.x,
+                   blockDim.x,
+                   static_cast<int>(put_status));
+        }
+        const bool any_put_failed = __syncthreads_or(put_status != NIXL_SUCCESS);
         if (any_put_failed) {
-            (void)nixlbenchSignalError(params);
```
hbadihi: @fteng-NV This synchronization across all threads should happen only once after all iterations are over not every iteration. I think every iteration after a batch is complete each thread / group should check for failure and when all iterations are done do `__syncthreads_..`

hbadihi (commented):
@fteng-NV Please see comments

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: Updates device api behavior to align with CPU side on [1cd28f4](https://github.com/ai-dynamo/nixl/pull/2084/commits/1cd28f4c5e0e310f42b313b47ef9b6a34cd1f8b5). So now nixlbenchSignalCompletion and nixlbenchSignalError all called using Level and channel_id

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -137,16 +183,25 @@ nixlbenchLaunchDevicePut(const nixlbenchDeviceXferParams &params, unsigned block
...
+    }
+
+    const size_t group_num = block_threads <= kWarpSize ? block_threads : block_threads / kWarpSize;
+    const size_t expected_active_group_num =
+        group_num < params.numRegions ? group_num : params.numRegions;
+    if (params.activeGroupNum != expected_active_group_num) {
+        std::cerr << "nixlbench: nixlbenchLaunchDevicePut: activeGroupNum=" << params.activeGroupNum
+                  << " does not match expected value " << expected_active_group_num << '\n';
+        return NIXL_ERR_INVALID_PARAM;
+    }
```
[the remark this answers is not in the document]
fteng-NV: updated in [1cd28f4](https://github.com/ai-dynamo/nixl/pull/2084/commits/1cd28f4c5e0e310f42b313b47ef9b6a34cd1f8b5)

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: Updated in [1cd28f4](https://github.com/ai-dynamo/nixl/pull/2084/commits/1cd28f4c5e0e310f42b313b47ef9b6a34cd1f8b5) and no active_group related contents now.

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: Updated in [1cd28f4](https://github.com/ai-dynamo/nixl/pull/2084/commits/1cd28f4c5e0e310f42b313b47ef9b6a34cd1f8b5), now each group posts and polls and signals its own share of the completion counter independently, no need synchronization at all.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
```diff
@@ -2025,12 +2109,15 @@ xferBenchNixlWorker::transfer(size_t block_size,
         releaseMemView(local_mvh);
     });
 
+    // Regions owned by a single execution group. The memory views flatten one IOV list per group
```
hbadihi: @fteng-NV Please omit source code comment to reduce bloat.

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -80,6 +80,9 @@
 #define XFERBENCH_INITIATOR_BUFFER_ELEMENT 0xbb
 #define XFERBENCH_TARGET_BUFFER_ELEMENT 0xaa
 
+// CUDA warp size, to derive Device API group number from block_threads
```
hbadihi: @fteng-NV Please omit source code comment to reduce bloat.

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -230,6 +233,14 @@ class xferBenchConfig {
     static bool use_device_api;
     static int block_threads;
 
+    /* Number of independent groups Device API kernel runs with. */
```
hbadihi: @fteng-NV Please omit source code comment to reduce bloat.

benchmark/nixlbench/src/utils/utils.h
```diff
@@ -230,6 +233,14 @@ class xferBenchConfig {
...
+    static int
+    deviceGroupNum();
+
+    /* Parallel workers split iterations across both CPU and Device API. */
```
hbadihi: @fteng-NV Please omit source code comment to reduce bloat.

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -61,19 +70,37 @@ nixlbenchSignalCounter(const nixlbenchDeviceXferParams &params,
...
+                     unsigned channel_id,
+                     nixlGpuXferStatusH &xfer_status) {
+    return nixlbenchSignalCounter<Level>(
+        params, params.errorCounterOffsetBytes, 1ull, channel_id, xfer_status, "error");
 }
 
 /**
  * Performs device-initiated NIXL PUT transfers and reports completion or errors
  * through remote counters.
+ *
```
hbadihi: @fteng-NV Please reduce code source comment to reduce bloat

hbadihi:
@fteng-NV  Looks good to me

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -87,37 +98,72 @@ nixlbenchPutKernel(nixlbenchDeviceXferParams params) {
...
+        }
+        if (active_group) {
+            if constexpr (Level == nixl_gpu_level_t::WARP) {
+                __syncwarp();
+            }
+            xfer_end_ns = nixlbenchGetTimeNs();
+            if (group_leader) {
+                const size_t sample_idx = iter * params.activeGroupNum + group_id;
+                params.postDurationNs[sample_idx] = post_end_ns - post_start_ns;
+                params.xferDurationNs[sample_idx] = xfer_end_ns - post_end_ns;
```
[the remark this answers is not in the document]
fteng-NV: With collapsed-CQ semantics, the last post covers all previous posts for one channel. If posting fails, the benchmark can simply report the failure and exit.

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -87,38 +114,63 @@ nixlbenchPutKernel(nixlbenchDeviceXferParams params) {
...
-        put_status = nixlbenchPostPut<Level>(params, region_idx, xfer_status);
-        if (put_status != NIXL_IN_PROG) {
-            break;
+    const unsigned channel_id = group_id % params.channelNum;
+    const bool group_leader = Level == nixl_gpu_level_t::THREAD || threadIdx.x % warpSize == 0;
+    const size_t region_base = group_id * params.numRegions;
+    bool group_failed = false;
+
+    for (uint64_t iter = 0; iter < params.numIterations && !group_failed; ++iter) {
+        if constexpr (Level == nixl_gpu_level_t::WARP) {
```
ofirfarjun7: why we need syncwarp here?

benchmark/nixlbench/src/utils/utils.cpp
ofirfarjun7: Maybe export device param check to helper func

fteng-NV: updated in [9bddd59](https://github.com/ai-dynamo/nixl/pull/2084/commits/9bddd595dc2164bcbb5feee386f165066e62377c)

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: This is to ensure all threads in warp start to put the batch at the same time. post_start_ns is recorded after it.

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -363,13 +370,66 @@ validateDeviceAPIConfig() {
...
+    const int group_num = xferBenchConfig::deviceGroupNum();
+
+    if (xferBenchConfig::device_channel_num == 0) {
+        xferBenchConfig::device_channel_num = group_num;
+    } else if (xferBenchConfig::device_channel_num > group_num) {
+        std::cout << "WARNING: Adjusting --device_channel_num from "
+                  << xferBenchConfig::device_channel_num << " to Device API group number "
+                  << group_num << std::endl;
+        xferBenchConfig::device_channel_num = group_num;
+    }
```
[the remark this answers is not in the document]
fteng-NV: Kernel is using channel_id = group_id % params.channelNum;

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
ofirfarjun7: But only leader report results, so maybe it is enough to ensure they all finish before capture post_end_ns

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -248,6 +249,11 @@ NB_ARG_BOOL(use_device_api,
             "When enabled, --num_threads is repurposed as the CUDA kernel "
             "block size (num_threads <= 32 -> THREAD level; > 32 -> WARP level, must be a "
             "multiple of 32), and the internal CPU thread count is forced to 1.");
+NB_ARG_INT32(device_channel_num,
+             0,
```
ofirfarjun7: why not set default to 1?

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -363,13 +370,66 @@ validateDeviceAPIConfig() {
...
     return reject("UCX GPU Device API support is not enabled in this build. "
                   "Set -Ducx_path=<path> with UCX GPU device headers available");
 #endif
 }
 
+static bool
+setupDeviceAPIConfig() {
+    const int num_threads = xferBenchConfig::num_threads;
+
+    if (num_threads < 1 || num_threads > 1024) {
```
ofirfarjun7: parentheses

benchmark/nixlbench/src/utils/utils.cpp
```diff
@@ -363,13 +370,66 @@ validateDeviceAPIConfig() {
...
+        std::cerr << "Invalid value for --num_threads: " << num_threads
+                  << ". Device API requires a GPU kernel block thread count in [1, 1024]"
+                  << std::endl;
+        return false;
+    }
+    if (num_threads > XFERBENCH_DEVICE_WARP_SIZE && num_threads % XFERBENCH_DEVICE_WARP_SIZE != 0) {
```
ofirfarjun7: parentheses

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: Only keep one syncwarp for xfer_end_ns in [030f383](https://github.com/ai-dynamo/nixl/pull/2084/commits/030f38331e3bbc88260d505b694e55d37011a78f)

benchmark/nixlbench/src/utils/utils.cpp
fteng-NV: change default to 1 in [030f383](https://github.com/ai-dynamo/nixl/pull/2084/commits/030f38331e3bbc88260d505b694e55d37011a78f)

benchmark/nixlbench/src/utils/utils.cpp
fteng-NV: Updated in [030f383](https://github.com/ai-dynamo/nixl/pull/2084/commits/030f38331e3bbc88260d505b694e55d37011a78f)

benchmark/nixlbench/src/utils/utils.cpp
fteng-NV: Updated in [030f383](https://github.com/ai-dynamo/nixl/pull/2084/commits/030f38331e3bbc88260d505b694e55d37011a78f)

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
```diff
@@ -87,38 +114,57 @@ nixlbenchPutKernel(nixlbenchDeviceXferParams params) {
...
-    }
+        const uint64_t post_end_ns = nixlbenchGetTimeNs();
 
-    const bool any_put_failed = __syncthreads_or(put_status != NIXL_SUCCESS);
-    if (threadIdx.x == 0) {
-        if (any_put_failed) {
-            (void)nixlbenchSignalError(params);
-            return;
+        if (put_status == NIXL_IN_PROG) {
+            put_status = nixlbenchPollXferStatus<Level>(put_status, xfer_status);
```
ofirfarjun7, line 137: for next PR need to consider if waiting for last put is good enough.

benchmark/nixlbench/src/worker/nixl/nixl_worker.cpp
ofirfarjun7, line 2239: We wait for total iter but AFAIU each group run `xferBenchConfig::num_iter / workers` iter.
Can it be a problem if the value is truncated?
Maybe it is also relevant for the `skip` value.

fteng-NV: It is the same as CPU side logic now. During config parsing in xferBenchConfig::loadParams, both num_iter and warmup_iter are rounded up to a multiple of workers * large_blk_iter_ftr. Therefore, the per-worker divisions are exact, no truncation happened. The same applies to skip.

benchmark/nixlbench/src/kernels/nixlbench_device_launch.cu
fteng-NV: Yes, agreed

ofirfarjun7 (approved):

ofirfarjun7:
/ok to test https://github.com/ai-dynamo/nixl/commit/bdf05a06bf649038569944fe9511ebfafe955add

Issue #2067, opened 2026-08-11
releaseXferReq use-after-free (SegFault) + nixlBasicDesc match key excludes metaInfo


hekhong-png:
## Summary

Two related defects in the NIXL core that together can cause use-after-free (SIGSEGV) and EBADF errors under concurrent FILE backend transfers. Discovered while debugging production EBADF errors (`aio_error failed: Unknown error -9`) in a GLM-5.2 deployment using the POSIX backend for KV cache offloading.

---

### Location

`src/core/nixl_agent.cpp:1319-1345` (`nixlAgent::releaseXferReq`)
`src/plugins/posix/posix_backend.cpp:341-345` (`nixlPosixEngine::releaseReqH`)

### Description

When `releaseXferReq` is called on a transfer with `status == NIXL_IN_PROG`:

1. It calls `checkXfer` (poll) — if still IN_PROG:
2. It calls `releaseReqH` (cancel in-flight IO) — for the POSIX backend, this is simply `delete handle` (`posix_backend.cpp:343`)
3. It then `delete req_hndl` (line 1343)

But in-flight IO callbacks (`ioDoneClb`) hold a pointer to `req_hndl`. After `delete req_hndl`, when a cancelled IO completes, its callback accesses the deleted `req_hndl` → **use-after-free → SIGSEGV**.

```cpp
// nixl_agent.cpp:1319
nixl_status_t
nixlAgent::releaseXferReq(nixlXferReqH *req_hndl) const {
    NIXL_SHARED_LOCK_GUARD(data->lock);
    if(req_hndl->status == NIXL_IN_PROG) {
        req_hndl->status = req_hndl->engine->checkXfer(req_hndl->backendHandle);
        if(req_hndl->status == NIXL_IN_PROG) {
            req_hndl->status = req_hndl->engine->releaseReqH(req_hndl->backendHandle);
            // ... error handling ...
            req_hndl->backendHandle = nullptr;
        }
    }
    delete req_hndl;   // ← callbacks for cancelled in-flight IO may still fire after this
    return NIXL_SUCCESS;
}
```

```cpp
// posix_backend.cpp:341
nixl_status_t
nixlPosixEngine::releaseReqH(nixlBackendReqH *handle) const {
    NIXL_ASSERT(handle != nullptr);
    delete handle;   // ← cancels IO but callbacks may still be pending
    return NIXL_SUCCESS;
}
```

### Reproduction

1. Start a POSIX backend transfer with multiple pages
[...]

PR #2085, opened 2026-08-13
fix(posix): recover AIO after I/O failures

### LINKED ISSUE #1696: POSIX backend (libaio & io_uring): ENOSPC short writes silently reported as DONE (data loss)

lluki:
Fix AIO handling of I/O errors. Right now we fail fast with leaving stale IOs, this PR fixes that by correctly cancelling and delaying completion until all IOs have completed (succeeded or cancelled).

Together with #1942 fixes #1696 . Most likely also related to `Defect A` of https://github.com/ai-dynamo/nixl/issues/2067

### Summary of the newly introduced Linux AIO error handling

Error handling mirrors the io_uring error paths. AIO is a bit simpler, because it does not support asynchronous cancel. So we always return 0 from cancel, which signals to the posix backend that it has no async cancel.

NIXL calls the `libaio` wrappers, so errors are returned directly as negative values such as `-EAGAIN`, rather than as `-1` with `errno` set.

Note that `io_submit` can signal both, errors which can be attributed to wrong I/Os as well as errors that can be attributed to the AIO library.

| Handling class | Function and return value | Current treatment |
|---|---|---|
| Accepted / success | `io_submit(): ret == nr` | The entire batch becomes in flight. |
| Accepted / success | `io_event.res == requested length` | Complete the IOCB successfully and invoke its callback with the transferred byte count. |
| Accepted / success | `io_cancel(): 0` | Cancellation succeeded synchronously. Process the returned `io_event` through the normal IOCB completion path. The event is not also placed on the completion queue. |
| Cancel refused; drain normally | `io_cancel(): any nonzero result` | Leave the IOCB in flight and rely on its ordinary completion callback. The code does not distinguish `-EAGAIN`, `-EFAULT`, `-EINVAL`, `-ENOSYS`, or another nonzero result. No asynchronous-cancellation completion count is returned. |
| Completion progress | `io_getevents(): rc > 0` | Process each returned `io_event` independently. The individual `io_event.res` value determines whether its IOCB succeeded or failed. |
| Fail IOCB | `io_submit(): -EINVAL` and the nonblocking `io_getevents()` health probe succeeds | Treat the error as a malformed or unsupported first IOCB. Fail the first IOCB and requeue the rest of the batch. |
| Fail IOCB | `io_submit(): any other negative result` except `-EAGAIN`, `-EINTR`, `-ENOSYS`, and terminal `-EINVAL` | Fail the first IOCB and requeue the rest. This includes documented `-EBADF`, `-EFAULT`, and `-EPERM`. The AIO queue remains usable. |
| Fail IOCB | `io_event.res < 0` | Fail that IOCB. An error delivered through a completion event is always operation-local and never poisons the AIO context. |
| Fail IOCB | `io_event.res >= 0`, but shorter than the requested length | Treat the short read or write as an IOCB failure. |
| No progress | `io_submit(): ret == 0` | Requeue the whole attempted batch and retry later. |
| No progress | `io_getevents(): 0` | No completion was available. Return `NIXL_IN_PROG`, unless every queue entry is already free. |
| Partial progress | `io_submit(): 0 < ret < nr` | Mark the exact accepted prefix `[0, ret)` in flight and requeue the unaccepted suffix while preserving order. |
| Terminal AIO failure | `io_submit(): -ENOSYS` | Enter terminal teardown because Linux AIO is unavailable. |
| Terminal AIO failure | `io_submit(): -EINVAL` and the nonblocking `io_getevents()` health probe fails | Treat `-EINVAL` as an invalid AIO context and enter terminal teardown. |
| Terminal AIO failure | `io_getevents(): any negative result except -EINTR` | Enter terminal teardown. Documented cases include `-EFAULT`, `-EINVAL`, and `-ENOSYS`. |
| Transient failure | `io_submit(): -EAGAIN` or `-EINTR` | Requeue the entire attempted batch without failing an IOCB and retry during a later poll. |
| Transient failure | `io_getevents(): -EINTR` | Return `NIXL_IN_PROG` and retry during a later poll. |

## Related manpages

- [io_submit(2)](https://man7.org/linux/man-pages/man2/io_submit.2.html)
- [io_getevents(2)](https://man7.org/linux/man-pages/man2/io_getevents.2.html)
- [io_cancel(2)](https://man7.org/linux/man-pages/man2/io_cancel.2.html)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved POSIX asynchronous I/O reliability for partially submitted, interrupted, cancelled, or failed requests.
  * Improved handling of queued and in-flight operations when the I/O backend becomes unavailable.
  * Improved detection and reporting of incomplete or short I/O operations.

* **Tests**
  * Added comprehensive Linux AIO coverage for recovery, cancellation, partial submissions, completion failures, and backend errors.
  * Expanded POSIX short-write testing across supported asynchronous I/O modes.
  * Linux AIO tests now run as part of the standard C++ test suite.


src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,184 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
         }
+
+        if (isTerminalSubmitError(ret)) {
+            NIXL_ERROR << "io_submit terminal error: " << nixl_strerror(-ret);
+            requeueFrom(to_submit, 0, num_ios);
+            enterTerminalError();
+            return NIXL_IN_PROG;
+        }
+
+        // A negative return rejects the first IOCB without accepting any of the batch.
```
lluki, line 163: This code path is new, if ret > 0 but < ios_to_submit,  we actively fail the IO (which then cascades into fast failing all the other IOs of that batch)

lluki:
@aranadive @vvenkates27

.gitlab/test_cpp.sh
```diff
@@ -93,8 +93,10 @@ telePID=$!
 sleep 15
 kill -s INT $telePID
 
-# Run the default POSIX queue, then add the io_uring mode when requested by CI.
+# Run the default POSIX queue and its focused Linux AIO tests, then add the io_uring mode
+# when requested by CI.
 ./bin/nixl_posix_test -n 128 -s 1048576
+./bin/nixl_posix_aio_test
```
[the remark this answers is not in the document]
lluki: But we do the same with `nixl_posix_test`. I prefer not to have conditional calls in CI as they might hide errors.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,185 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
-    rc = io_getevents(io_ctx_, 0, MAX_IO_CHECK_COMPLETED_BATCH_SIZE, events, &timeout);
+    int rc = io_getevents(io_ctx_, 0, MAX_IO_CHECK_COMPLETED_BATCH_SIZE, events, &timeout);
     if (rc < 0) {
-        NIXL_ERROR << "io_getevents error: " << rc;
+        if (rc == -EINTR) {
+            return NIXL_IN_PROG;
+        }
+        NIXL_ERROR << "io_getevents error: " << nixl_strerror(-rc);
         return NIXL_ERR_BACKEND;
     }
```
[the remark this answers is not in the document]
lluki: - The kernel_queue_size_ change is not valid. The local events array has 64 entries, while kernel_queue_size_ ranges from 16 to 1024. Passing kernel_queue_size_ directly as io_getevents(..., nr, events, ...) could let the kernel write beyond the 64-entry buffer.

    The current fixed 64 is a safe batching cap. Passing 64 when the AIO context capacity is smaller is fine: only as many completions as exist can be returned. If desired for clarity, the safe formulation would be min(kernel_queue_size_, 64), but it does not fix a
    demonstrated correctness issue.

  - The retryable-error distinction is already mostly present. doCheckCompleted() returns NIXL_IN_PROG for -EINTR; all other negative io_getevents results return NIXL_ERR_BACKEND.

    Since the call uses min_nr = 0 and a zero timeout, -EAGAIN is not an expected “no completion yet” result—the normal outcome is 0. Treating an unexpected -EAGAIN as retryable would be speculative unless backed by the libaio/kernel contract or a reproducer.

  - The concern about isTerminalSubmitError() misclassifying actual I/O completion errors is inaccurate. Completion errors are in events[i].res and go to completeIO(). isTerminalSubmitError() is only invoked after io_submit() returns -EINVAL; it calls doCheckCompleted()
    solely to determine whether the AIO context itself is invalid.

  So: keep the fixed completion batch size; the retry behavior is already correct for EINTR; no change is justified from this finding alone.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -47,16 +50,31 @@ class nixlPosixIOQueueLinuxAIO : public nixlPosixIOQueueImpl<nixlPosixLinuxAioIO
             void *ctx) override;
     virtual nixl_status_t
     poll(void) override;
+    virtual unsigned
+    cancel(void *ctx, nixlPosixIOQueueCancelDoneCb clb) override;
```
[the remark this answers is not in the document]
lluki: This is an override, the parent method is documented, this just follows it.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,185 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
+        if (io_cancel(io_ctx_, &io.io_, &event) == 0) {
+            // io_cancel returns the canceled operation's completion synchronously.
+            completeIO(&io, event.res);
+        }
     }
 
-    return NIXL_IN_PROG; // Some blocks are in flight, need to check again
+    // No asynchronous cancellations were requested.
+    return 0;
 }
```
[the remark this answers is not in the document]
lluki: it is bollocks. It confuses Linux AIO’s synchronous io_cancel() model with io_uring’s asynchronous cancel-CQE model.

  For Linux AIO:

  - If io_cancel() succeeds, it synchronously supplies the cancelled operation’s io_event; cancel() processes it immediately through completeIO().
  - If io_cancel() fails, no asynchronous cancellation was accepted or is pending. The original I/O remains in flight and will later be reaped normally by doCheckCompleted().
  - That normal reaping invokes the original I/O callback (io->clb_), which increments num_confirmed_ios_. The request cannot become complete until that count reaches queue_depth_.

  So the cancellation callback must not be retained/called for an unsuccessful io_cancel(): the base API defines it as a callback for asynchronous cancellations that were actually requested. Linux AIO requested none, hence its correct return is zero.

  The lifetime claim is also unsupported by the current control flow. The only poll terminal path first calls enterTerminalError(), which fails and confirms all queued/in-flight I/O before a later poll reports the backend error. Completion errors similarly remain
  NIXL_IN_PROG until remaining I/O is reaped.

  This distinction is why io_uring does retain the callback: it submits a separate asynchronous cancellation SQE and therefore must wait for its cancellation CQE. Linux AIO does not have that pending cancellation operation after io_cancel() fails.

  There is already a Linux-AIO unit test for this shape: it forces io_cancel() to return -EAGAIN, expects cancel() to return zero, and then verifies the original request completes through the normal I/O completion path.

test/unit/plugins/posix/nixl_posix_aio_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+        AIO_CHECK(test.queue->post() == NIXL_IN_PROG);
+
+        unsigned waits = test.queue->cancel(&cancelled, cancelCompletionCallback);
+        AIO_CHECK(waits == 0);
+        AIO_CHECK(test.enqueue(unrelated, scoped_request_count, 1));
+        AIO_CHECK(test.queue->post() == NIXL_IN_PROG);
+        AIO_CHECK(test.drain() == NIXL_SUCCESS);
+        AIO_CHECK(cancelled.completions == scoped_request_count && cancelled.errors > 0);
+        AIO_CHECK(cancelled.cancel_completions == static_cast<int>(waits));
+        AIO_CHECK(unrelated.completions == 1 && unrelated.errors == 0);
```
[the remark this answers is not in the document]
lluki: For Linux AIO:

  - Queued I/O is failed synchronously via the normal I/O callback.
  - A successful io_cancel() supplies its completion synchronously and is processed through the normal I/O callback.
  - A failed io_cancel() leaves the original I/O in flight; its eventual normal completion likewise uses the normal I/O callback.
  - Therefore waits == 0 and cancel_completions == 0 are the intended result, including when there were in-flight I/Os.

  The test does not test only queued I/O: it queues 65 operations, while post() submits up to 64, so it deliberately has queued work and submitted work when cancellation begins. Its important assertion is:

  cancelled.completions == scoped_request_count

  That confirms every one of the 65 original I/Os is accounted for by its normal completion callback. errors > 0 confirms the cancellation/error path occurred. The unrelated I/O assertion checks that this cleanup remains scoped to the cancelled request.

test/unit/plugins/posix/nixl_posix_aio_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+#include <cerrno>
+#include <chrono>
+#include <cstring>
+#include <dlfcn.h>
+#include <iostream>
+#include <libaio.h>
+#include <stdexcept>
+#include <thread>
+#include <unistd.h>
+#include <vector>
```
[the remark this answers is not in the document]
lluki: added

test/unit/plugins/posix/nixl_posix_aio_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+} // namespace
+
+extern "C" int
+io_submit(io_context_t ctx, long nr, struct iocb **iocbpp) {
+    using submit_fn = int (*)(io_context_t, long, struct iocb **);
+    static auto real_submit =
+        reinterpret_cast<submit_fn>(dlvsym(RTLD_NEXT, "io_submit", "LIBAIO_0.1"));
+    if (!real_submit) {
+        return -EINVAL;
+    }
```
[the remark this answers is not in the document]
lluki: added

lluki:
resolved CodeRabbit nits

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,185 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
+    if (error == -ENOSYS) {
+        return true;
+    }
+    if (error != -EINVAL) {
+        return false;
+    }
+
+    // EINVAL is ambiguous: it can describe either the AIO context or the first IOCB.
+    // Probe the context with a nonblocking completion check before declaring it terminal.
+    return doCheckCompleted() < 0;
```
vvenkates27: isTerminalSubmitError() treats EINVAL as terminal by calling doCheckCompleted(). That is not a side-effect-free probe: it can complete live I/Os, and io_getevents -EINTR returns NIXL_IN_PROG (>= 0), so a dead AIO context is classified as a per-IOCB reject. post() then failIO()s only the first IOCB and retries the rest of the batch on an invalid ctx. 

How about Probe the context with a completion-free io_getevents (e.g. min_nr=0, nr=0, timeout=0). Retry the probe on -EINTR. Only then call enterTerminalError(). Keep the per-IOCB failIO(to_submit[0]) + requeue path for non-terminal EINVAL.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,185 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
 
-        if (io->clb_) {
-            io->clb_(io->ctx_, events[i].res, 0);
+        it = ios_to_submit_.erase(it);
+        failIO(io);
+    }
+}
+
+void
+nixlPosixIOQueueLinuxAIO::enterTerminalError() {
```
vvenkates27, line 277: enterTerminalError() calls io_queue_release() and then failIO()s every in_flight_ slot back onto free_ios_. io_destroy does not guarantee those kernel ops are gone, so the pool can hold IOCBs the kernel may still own.?
Do not recycle in-flight slots after destroy (leave them off free_ios_, or drain/cancel first). maybe Skip io_cancel when !io_ctx_active_. Safe to defer: terminal_error_ already blocks enqueue.

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,185 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
+void
+nixlPosixIOQueueLinuxAIO::requeueFrom(nixlPosixLinuxAioIO *const *ios, int first, int count) {
+    for (int i = count - 1; i >= first; i--) {
+        ios_to_submit_.push_front(ios[i]);
+    }
+}
+
+bool
+nixlPosixIOQueueLinuxAIO::completeIO(nixlPosixLinuxAioIO *io, int64_t result) {
+    NIXL_ASSERT(io->in_flight_);
```
vvenkates27, line 187: completeIO() / failIO() always push onto free_ios_ and invoke clb_. NIXL_ASSERT(in_flight_) is a no-op in release builds. If io_cancel succeeds and a CQE still arrives, the same slot is freed twice and the callback runs twice.

vvenkates27 (commented):
Just some comments of flows which felt odd.

src/plugins/posix/linux_aio_io_queue.cpp
lluki: Yes, I think you're right, this sequence can happen. I'm not sure it is wrong, but agree, the probe is better. Will do it.

src/plugins/posix/linux_aio_io_queue.cpp
lluki: I think the `io_destroy` syscall guarantees the completion/cancellation of outstanding reqs: https://github.com/torvalds/linux/blob/master/fs/aio.c#L1521-L1526

Should be ok, no?

src/plugins/posix/linux_aio_io_queue.cpp
lluki: This can't happen, a successful `io_cancel`  (ret value 0, which we check) guarantees that the completion will not occur. If io_cancel cant guarantee this, it would report an error. So we are safe here. The assertion is correct imho,  this is only to make clear to potential future modifiers of this code what is the expectation.

src/plugins/posix/linux_aio_io_queue.cpp
lluki: Added

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,191 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
+        return false;
+    }
+
+    // EINVAL is ambiguous: it can describe either the AIO context or the first IOCB.
+    // Probe the context without reaping completions before declaring it terminal.
+    int rc;
+    do {
+        struct timespec timeout = {0, 0};
+        rc = io_getevents(io_ctx_, 0, 0, nullptr, &timeout);
+    } while (rc == -EINTR);
```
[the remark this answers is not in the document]
lluki: We can force the syscall by passing a nullptr timeout - implemented that

src/plugins/posix/linux_aio_io_queue.cpp
```diff
@@ -122,75 +148,191 @@ nixlPosixIOQueueLinuxAIO::post(void) {
...
+nixlPosixIOQueueLinuxAIO::requeueFrom(nixlPosixLinuxAioIO *const *ios, int first, int count) {
+    for (int i = count - 1; i >= first; i--) {
+        ios_to_submit_.push_front(ios[i]);
+    }
+}
+
+bool
+nixlPosixIOQueueLinuxAIO::completeIO(nixlPosixLinuxAioIO *io, int64_t result) {
+    NIXL_ASSERT(io->in_flight_);
+    int error = result < 0 || static_cast<size_t>(result) != io->len_;
```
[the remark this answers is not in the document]
lluki: done

src/plugins/posix/linux_aio_io_queue.cpp
vvenkates27: Fair. A 0 return from io_cancel means that completion is yours and will not also show up in io_getevents. For file AIO this path is also effectively unused. The in_flight_ assert is a good invariant. No change needed.

src/plugins/posix/linux_aio_io_queue.cpp
vvenkates27: You’re right. Successful io_destroy waits for in-flight requests (kill_ioctx + ctx->reqs), so synthesizing failIO() and putting those slots back on free_ios_ is safe. Dropping this.

vvenkates27 (approved):

lluki:
I give this one last test integrated in SGLang, then i will hit the merge button

lluki:
verified in SGL too

PR #2089, opened 2026-08-14
EXAMPLES/DEVICE/EP/CSRC: Disabled UCX CUDA IPC cache.

rakhmets:
## What?
Disabled UCX CUDA IPC cache.

## Why?
The previous fix of memory leak (https://github.com/ai-dynamo/nixl/pull/1846) was reverted (https://github.com/ai-dynamo/nixl/pull/1996).

These changes (https://github.com/openucx/ucx/pull/11288, https://github.com/openucx/ucx/pull/11730) help eliminate the memory leak by disabling CUDA IPC cache.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -1412,6 +1412,7 @@ void Buffer::_nixl_agent_init() {
     init_params["ucx_num_device_channels"] = num_channels_env ? num_channels_env : "4";
     init_params["ucx_error_handling_mode"] = "none";
     init_params["num_workers"] = std::to_string(1);
+    init_params["engine_config"] = "CUDA_IPC_CACHE=n";
 
     nixlBackendH* ucx_backend = nullptr;
     status = agent->createBackend("UCX", init_params, ucx_backend);


itayalroy (changes requested):
IIUC, `connect_ranks` relies on the UCX CUDA IPC cache, it destroys the old mview and creates a new one assuming the NVLink ptrs exported during the creation of the old mview are still valid. We cannot disable the cache until we get rid of this assumption. I'm already working on it so should have a PR up soon, so I think we should wait with merging #2089 until then

rakhmets:
> IIUC, `connect_ranks` relies on the UCX CUDA IPC cache, it destroys the old mview and creates a new one assuming the NVLink ptrs exported during the creation of the old mview are still valid. We cannot disable the cache until we get rid of this assumption. I'm already working on it so should have a PR up soon, so I think we should wait with merging #2089 until then

Do you mean that there will be an issue if `connect_ranks` is called, and NIXL EP still runs a CUDA kernel, which uses  inter process memory handles opened on previous iteration/stage?

itayalroy:
@linoygeva and I are seeing intermittent NIXL EP timeouts with vLLM Elastic EP when `UCX_CUDA_IPC_CACHE=n` which do not reproduce with `UCX_CUDA_IPC_CACHE=y`.

vLLM also uses the NIXL host API for EPLB weight transfers, so this environment variable disables the IPC cache for the separate EPLB agent as well. We suspect the timeouts are related to that. We cannot merge this PR until we fix this issue.

I think it will be easier to fix the memory leak with `UCX_CUDA_IPC_CACHE=y` first as we discussed @yosefe @ofirfarjun7 @rakhmets

Issue #2025, opened 2026-07-31
VLLM: [Bug] `remove_remote_agent` does not clean up UCX QP resources


tlrmchlsmth:
Situation: 
A large long-running WideEP vLLM deployment with several replicas running at low-concurrency. Since there are many ranks and fairly light load, the remote agents tend to time out due to sparse connectivity due to this 3600s TTL:

https://github.com/vllm-project/vllm/blob/60b3d39cd36c53a698040edbf51406d3febc97a7/vllm/distributed/kv_transfer/kv_connector/v1/nixl/base_worker.py#L2340-L2363

Eventually, decode nodes begin to crash with this when handshaking, since fresh handshakes are frequent in this scenario:
```
2026-07-26T23:22:41.639678536Z [1785108161.639653] [wide-ep-lws-nvidia-gpu-vllm-glm-5-2-v2-decode-ll-4-2:1901 :0]      ib_mlx5_dv.c:452  UCX  ERROR   mlx5dv_devx_obj_create(QP) failed on mlx5_8, syndrome 0x19c6d6: Remote I/O error
```

Diagnosis:
1. NIXL calls ucp_ep_close_nbx(..., FORCE).
  2. UCX logically closes the UCP endpoint and schedules asynchronous lane discard.
  3. UCX returns NULL, so NIXL has no completion request to wait on.
  4. Destroying the UCT endpoints, mlx5 DevX QPs, and pinned buffers still requires progress on that endpoint's owning UCX worker.
  5. NIXL's dedicated workers are idle after removal, so that progress does not occur.
  6. Repeated reconnects allocate new transport resources while prior teardown remains queued.
  
Simple repro code here: https://github.com/tlrmchlsmth/nixl-agent-reclaim-repro

PR #2090, opened 2026-08-14
PLUGINS/UCX: Deleted unused backend parameter.

rakhmets:
## What?
Deleted unused UCX backend parameter.

## Why?
The parameter is no longer used after these changes https://github.com/ai-dynamo/nixl/pull/1996.

### DIFF
--- a/src/plugins/ucx/ucx_utils.cpp
+++ b/src/plugins/ucx/ucx_utils.cpp
@@ -115,12 +115,17 @@ nixlUcxEp::setState(nixl::ucx::ep_state_t new_state) {
     state_ = new_state;
 }
 
+namespace {
+[[nodiscard]] ucs_status_ptr_t
+ucpEpClose(ucp_ep_h ep, uint32_t flags) {
+    const ucp_request_param_t req_param = {.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS, .flags = flags};
+    return ucp_ep_close_nbx(ep, &req_param);
+}
+} // namespace
+
 nixl_status_t
 nixlUcxEp::closeImpl() {
-    ucs_status_ptr_t request = nullptr;
     const nixl::ucx::ep_state_t current_state = state_;
-    const ucp_request_param_t req_param = {.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS,
-                                           .flags = closeFlags_};
 
     switch (current_state) {
     case nixl::ucx::ep_state_t::UNINITIALIZED:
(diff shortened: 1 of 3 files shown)


src/plugins/ucx/ucx_utils.cpp
```diff
@@ -115,12 +115,17 @@ nixlUcxEp::setState(nixl::ucx::ep_state_t new_state) {
     state_ = new_state;
 }
 
+namespace {
+[[nodiscard]] ucs_status_ptr_t
+ucpEpClose(ucp_ep_h ep, uint32_t flags) {
+    const ucp_request_param_t req_param = {.op_attr_mask = UCP_OP_ATTR_FIELD_FLAGS, .flags = flags};
+    return ucp_ep_close_nbx(ep, &req_param);
```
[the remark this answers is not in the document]
rakhmets: We have already moved to C++ 20. https://github.com/ai-dynamo/nixl/pull/1571.
The `.coderabbit.yaml` has not been updated.

src/plugins/ucx/ucx_utils.cpp
```diff
@@ -129,19 +134,15 @@ nixlUcxEp::closeImpl() {
...
         if (UCS_PTR_IS_PTR(request)) {
             ucp_request_free(request);
         }
         eph = nullptr;
         return NIXL_ERR_REMOTE_DISCONNECT;
     }
-    case nixl::ucx::ep_state_t::CONNECTED:
-        request = ucp_ep_close_nbx(eph, &req_param);
+    case nixl::ucx::ep_state_t::CONNECTED: {
+        ucs_status_ptr_t request = ucpEpClose(eph, 0);
```
iyastreb, line 145: There is ongoing discussion on that: should we close EPs forcefully
Motivation is resource leak: https://github.com/ai-dynamo/nixl/issues/2025
EPs that are closed without FORCE are never destroyed by UCX (when connection is considered as alive). UCX only destroys EPs when connection is actually broken

rakhmets: I'm not sure this is something that is related to this changes.

I added this flag to be able to forcibly close UCT endpoints along with `ucp_ep_close_nbx` during normal destruction flow. And it's not used anymore, 

NIXL always called `ucp_ep_close_nbx` without force flag in normal destruction flow, and with force flag in case of remote peer failure.

rakhmets: Your issue is that when peer failure mode is enabled, `ucp_ep_close_nbx` returns `NULL`. So, there is no way do progress/wait for the actual completion of `ucp_ep_close_nbx`. I do not remove `UCP_EP_CLOSE_FLAG_FORCE` from `ucp_ep_close_nbx` from error handling flow. I just want to remove unused option.

rakhmets: I checked the code of the small reproducer in the opened ticket. It uses this option. This option is not safe to use, as device API related UCX transports do not support the feature in the safe way.
The option `ucx_ep_close_force` was not in previous release, and shouldn't be in the next one.

rakhmets: The issue described in the comment should be fixed by changing closing endpoint logic, including changes in UCX code.

@iyastreb please resolve the comment, if you agree that this is not an issue.

iyastreb: ok

iyastreb (approved):

PR #2094, opened 2026-08-14
libfabric: disable shm provider on Neuron hosts

rongbingzhou:
## What?
Pass the detected system runtime (`fi_hmem_iface`) into `nixlLibfabricRail` and, when it is `FI_HMEM_NEURON`, call `fi_setopt` with `FI_OPT_SHARED_MEMORY_PERMITTED = false` on each rail's endpoint so intra-node transfers stay on the EFA path.

## Why?
The libfabric shm provider has no Neuron HMEM support. On Neuron hosts, libfabric can pick the shm provider for intra-node transfers, which cannot handle Neuron device memory.

## How?
- `nixlLibfabricRail` ctor takes an extra `enum fi_hmem_iface runtime` argument, stored as `runtime_`.
- `nixlLibfabricRailManager::createRails()` forwards its existing `runtime_` member when constructing each rail.
- After `fi_ep_bind` of the AV, Neuron rails call `fi_setopt(FI_OPT_SHARED_MEMORY_PERMITTED, false)` to disable shm provider.
- Rail creation is never failed by this: providers that do not implement the option return `-FI_ENOSYS`/`-FI_ENOPROTOOPT`, which we warn about and continue past.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accelerator-aware libfabric initialization by disabling shared-memory transfers when unsupported by the detected runtime.
  * Initialization failures related to shared-memory configuration are now logged clearly and stop setup instead of proceeding with an invalid configuration.

### DIFF
--- a/src/utils/libfabric/libfabric_rail.cpp
+++ b/src/utils/libfabric/libfabric_rail.cpp
@@ -29,6 +29,21 @@
 #include <stdexcept>
 #include <stack>
 
+namespace {
+
+// Whether libfabric's shared memory (shm) provider can move data for an HMEM runtime.
+//
+// shm has no Neuron HMEM support. Every other runtime keeps shm, so that intra-node
+// transfers retain its acceleration.
+//
+// Update this when a runtime's shm support changes.
+bool
+shmProviderSupportsRuntime(enum fi_hmem_iface runtime) {
+    return runtime != FI_HMEM_NEURON;
+}
+
+} // namespace
+
 // RequestPool Base Class Implementation
 
 RequestPool::RequestPool(size_t pool_size, size_t rail_id)
@@ -391,13 +406,15 @@ DataRequestPool::allocate(nixlLibfabricReq::OpType op_type, uint32_t req_id) {
 
 nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
                                      const std::string &provider,
-                                     uint16_t id)
+                                     uint16_t id,
+                                     enum fi_hmem_iface runtime)
     : rail_id(id),
       device_name(device),
       provider_name(provider),
       control_request_pool_(NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL, id),
       data_request_pool_(NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL, id),
-      provider_supports_hmem_(false) {
+      provider_supports_hmem_(false),
+      runtime_(runtime) {
     // Initialize all pointers to nullptr
     info = nullptr;
     fabric = nullptr;
@@ -527,6 +544,26 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
         if (ret) {
             NIXL_ERROR << "fi_ep_bind av failed for rail " << rail_id << ": " << fi_strerror(-ret);
             throw std::runtime_error("fi_ep_bind av failed for rail " + std::to_string(rail_id));
+        }
+
+        // On runtimes the shm provider cannot serve, disable it so that intra-node
+        // transfers stay on the EFA path.
+        if (!shmProviderSupportsRuntime(runtime_)) {
+            const bool shared_memory_permitted = false;
+            ret = fi_setopt(&endpoint->fid,
+                            FI_OPT_ENDPOINT,
+                            FI_OPT_SHARED_MEMORY_PERMITTED,
+                            &shared_memory_permitted,
+                            sizeof(shared_memory_permitted));
+            if (ret) {
+                NIXL_ERROR << "fi_setopt FI_OPT_SHARED_MEMORY_PERMITTED failed for rail " << rail_id
+                           << " (provider: " << provider_name << "): " << fi_strerror(-ret);
+                throw std::runtime_error(
+                    "fi_setopt FI_OPT_SHARED_MEMORY_PERMITTED failed for rail " +
+                    std::to_string(rail_id));
+            }
+            NIXL_INFO << "Disabled shared memory provider for rail " << rail_id
+                      << " (runtime has no shm support)";
         }
 
         if (provider_name == "efa") {
--- a/src/utils/libfabric/libfabric_rail.h
+++ b/src/utils/libfabric/libfabric_rail.h
@@ -400,7 +400,10 @@ public:
     struct fid_ep *endpoint; ///< Libfabric endpoint handle
 
     /** Initialize libfabric rail with all resources */
-    nixlLibfabricRail(const std::string &device, const std::string &provider, uint16_t id);
+    nixlLibfabricRail(const std::string &device,
+                      const std::string &provider,
+                      uint16_t id,
+                      enum fi_hmem_iface runtime);
 
     /** Destroy rail and cleanup all libfabric resources */
     ~nixlLibfabricRail();
@@ -600,6 +603,9 @@ private:
 
     // Provider capability flags
     bool provider_supports_hmem_;
+
+    // System runtime type (CUDA, NEURON, or SYSTEM) this rail was created for
+    enum fi_hmem_iface runtime_;
 
     void
     pollForCompletions();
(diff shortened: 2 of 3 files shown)


src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -529,6 +531,30 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
             throw std::runtime_error("fi_ep_bind av failed for rail " + std::to_string(rail_id));
         }
 
+        // The libfabric shm provider has no Neuron HMEM support. Thus, on Neuron hosts,
+        // we disable libfabric's shared memory provider so that intra-node transfers
+        // stay on the EFA path.
+        if (runtime_ == FI_HMEM_NEURON) {
```
bwbarrett: should this be a blocked list or an allow list?

amitrad-aws: I think it should be a blocked list but don't have a strong preference

rongbingzhou: Discussed offline. We'll make this a blocked list.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -529,6 +548,29 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
...
 
+        // On runtimes the shm provider cannot serve, disable it so that intra-node
+        // transfers stay on the EFA path.
+        if (!shmProviderSupportsRuntime(runtime_)) {
+            const bool shared_memory_permitted = false;
+            ret = fi_setopt(&endpoint->fid,
+                            FI_OPT_ENDPOINT,
+                            FI_OPT_SHARED_MEMORY_PERMITTED,
+                            &shared_memory_permitted,
+                            sizeof(shared_memory_permitted));
```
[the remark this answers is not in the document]
rongbingzhou: Skipping this one — the option is available in every configuration this plugin supports.

`FI_OPT_SHARED_MEMORY_PERMITTED` was added in libfabric v1.21.0 (`include/rdma/fi_endpoint.h:72`, present as of v1.21.0rc1), and v1.21.0 is exactly the documented floor: `src/plugins/libfabric/README.md:21` states "Minimum required version: `v1.21.0`", and every build path in the repo pins it — `.gitlab/build.sh:41`, `.gitlab/build-rocm.sh:229`, `contrib/Dockerfile:38`, `contrib/Dockerfile.manylinux:88`, `.ci/dockerfiles/Dockerfile.rocm:352`, `benchmark/nixlbench/contrib/Dockerfile:137`. So there is no supported configuration in which this fails to compile.

You're right that Meson accepts any installed libfabric, but that is a pre-existing property of `dependency('libfabric', required: false)` at `meson.build:402` and is not specific to this option. If we want to enforce the documented floor, the fix is a `version: '>=1.21.0'` constraint on that dependency, which would cover the whole plugin — not a per-symbol `HAVE_` gate here. But that's out-of-scope of this PR.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -29,6 +29,23 @@
...
+namespace {
+
+/**
+ * Whether libfabric's shared memory (shm) provider can move data for an HMEM runtime.
+ *
+ * shm has no Neuron HMEM support. Every other runtime keeps shm, so that intra-node
+ * transfers retain its acceleration.
+ *
+ * Update this when a runtime's shm support changes.
+ */
```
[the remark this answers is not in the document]
rongbingzhou: Will fix

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -529,6 +546,29 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
...
+                            &shared_memory_permitted,
+                            sizeof(shared_memory_permitted));
+            if (ret && ret != -FI_ENOSYS && ret != -FI_ENOPROTOOPT) {
+                NIXL_WARN << "fi_setopt FI_OPT_SHARED_MEMORY_PERMITTED failed for rail " << rail_id
+                          << ": " << fi_strerror(-ret) << " - continuing anyway";
+            } else if (ret == 0) {
+                NIXL_INFO << "Disabled shared memory provider for rail " << rail_id
+                          << " (runtime has no shm support)";
+            } else if (ret == -FI_ENOSYS || ret == -FI_ENOPROTOOPT) {
+                NIXL_WARN << "FI_OPT_SHARED_MEMORY_PERMITTED not supported for rail " << rail_id
```
amitrad-aws: Why are we continuing in case of error and not throwing an exception?

amitrad-aws (changes requested):

src/utils/libfabric/libfabric_rail.cpp
rongbingzhou: Make sense. Fixed.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -529,6 +546,26 @@ nixlLibfabricRail::nixlLibfabricRail(const std::string &device,
...
+                            FI_OPT_SHARED_MEMORY_PERMITTED,
+                            &shared_memory_permitted,
+                            sizeof(shared_memory_permitted));
+            if (ret) {
+                NIXL_ERROR << "fi_setopt FI_OPT_SHARED_MEMORY_PERMITTED failed for rail " << rail_id
+                           << " (provider: " << provider_name << "): " << fi_strerror(-ret);
+                throw std::runtime_error(
+                    "fi_setopt FI_OPT_SHARED_MEMORY_PERMITTED failed for rail " +
+                    std::to_string(rail_id));
+            }
```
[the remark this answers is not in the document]
rongbingzhou: This was changed at @amitrad-aws's request earlier in this PR — the previous version warned and continued, and the question was why we don't throw.

`FI_OPT_EFA_USE_UNSOLICITED_WRITE_RECV` isn't a comparable precedent: it is performance tuning, so an endpoint without it is still correct. Disabling shm on Neuron is a correctness requirement — shm cannot copy Neuron device memory. If we cannot turn it off, warning and continuing ships exactly the broken configuration
this change exists to prevent, and defers it into an intra-node hang instead of a clear failure at init.

src/utils/libfabric/libfabric_rail.cpp
```diff
@@ -391,13 +406,15 @@ DataRequestPool::allocate(nixlLibfabricReq::OpType op_type, uint32_t req_id) {
...
+                                     uint16_t id,
+                                     enum fi_hmem_iface runtime)
     : rail_id(id),
       device_name(device),
       provider_name(provider),
       control_request_pool_(NIXL_LIBFABRIC_CONTROL_REQUESTS_PER_RAIL, id),
       data_request_pool_(NIXL_LIBFABRIC_DATA_REQUESTS_PER_RAIL, id),
-      provider_supports_hmem_(false) {
+      provider_supports_hmem_(false),
+      runtime_(runtime) {
```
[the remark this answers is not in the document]
rongbingzhou: The constructor's brief at `src/utils/libfabric/libfabric_rail.h:402` documents none of its parameters — not `device`, `provider`, or `id` either — and that predates this change. Adding `@param runtime` alone would leave one of four parameters documented, which reads more inconsistent than the current brief.

That is also the file's convention rather than an oversight: `libfabric_rail.h` has 47 one-line `/** ... */` briefs and 2 `@param` uses in total, both on multi-line blocks for single-parameter methods. `docs/CodeStyle.md` asks for `@brief`, `@param` and friends "as appropriate", which I read as leaving room for a brief on a constructor whose parameters are self-describing. Giving this constructor full `@param` coverage would be a reasonable cleanup, but it applies to all four parameters and is separable from this change.

On the shm behaviour specifically, it is already documented at the two places that implement it — `shmProviderSupportsRuntime()` states which runtimes shm cannot serve and why, and the endpoint-setup block explains why failing to disable it is fatal. `runtime_` is described where it is declared (`libfabric_rail.h:607`). Restating it in the constructor brief would add a third copy to keep in sync with the exclusion list, which is exactly what the helper exists to centralise.

fengjica (approved):

amitrad-aws (approved):

PR #2095, opened 2026-08-16
nixl_ep: Allow CUDA graph reuse across rank changes

itayalroy:
PR https://github.com/ai-dynamo/nixl/pull/1693 deprecated num_experts because NIXL EP could derive the dispatch layout from active_rank_bound, which is maintained internally from the active rank mask.

However, masking or unmasking ranks changes active_rank_bound, which changes tensor shapes, buffer offsets, and kernel launch parameters captured in CUDA graphs. This prevents callers from reusing those graphs after rank changes.

Instead, honor num_experts when it is explicitly provided, while continuing to derive the layout from active_rank_bound by default. Callers can pass the maximum number of physical experts they expect, keeping the dispatch and combine layouts fixed while ranks are masked or unmasked.

Reusing CUDA graphs keeps kernel launch configurations fixed across rank changes, which may be suboptimal in some cases. We plan to address this in the next NIXL release.

## Summary by CodeRabbit

* **New Features**
  * Low-latency dispatch now supports an optional expert capacity, defaulting to the active expert count when unspecified.
  * Combine operations dynamically adapt to the configured rank or expert range.
  * Buffer sizing and validation now reflect the requested runtime configuration.

* **Bug Fixes**
  * Improved validation prevents incompatible tensor dimensions and bounds during dispatch and combine operations.

### DIFF
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -107,7 +107,7 @@ void Buffer::_refresh_active_rank_bound() {
     set_active_rank_bound(bound);
 }
 
-int Buffer::get_rank_bound(const std::optional<int>& num_experts) const {
+int Buffer::get_rank_bound(std::optional<int> num_experts) const {
     if (!num_experts)
         return active_rank_bound;
 
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -107,6 +107,17 @@ void Buffer::_refresh_active_rank_bound() {
     set_active_rank_bound(bound);
 }
 
+int Buffer::get_rank_bound(const std::optional<int>& num_experts) const {
+    if (!num_experts)
+        return active_rank_bound;
+
+    EP_HOST_ASSERT(*num_experts % num_experts_per_rank == 0);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    const int max_num_experts = max_num_ranks * num_experts_per_rank;
+    EP_HOST_ASSERT(*num_experts >= active_expert_bound and *num_experts <= max_num_experts);
+    return *num_experts / num_experts_per_rank;
+}
+
 void Buffer::init(int num_ranks, int num_experts_per_rank, int64_t num_nvl_bytes, int64_t num_rdma_bytes)
 {
     EP_HOST_ASSERT(num_ranks > 0);
@@ -1046,11 +1057,7 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
     EP_HOST_ASSERT(topk_idx.dim() == 2 and topk_idx.is_contiguous());
     EP_HOST_ASSERT(x.size(0) == topk_idx.size(0) and x.size(0) <= num_max_dispatch_tokens_per_rank);
     EP_HOST_ASSERT(topk_idx.scalar_type() == c10::CppTypeToScalarType<topk_idx_t>::value);
-    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
-    const int expert_bound = num_experts.value_or(active_expert_bound);
-    EP_HOST_ASSERT(expert_bound >= active_expert_bound and expert_bound <= max_num_ranks * num_experts_per_rank);
-    EP_HOST_ASSERT(expert_bound % num_experts_per_rank == 0);
-    const int rank_bound = expert_bound / num_experts_per_rank;
+    const int rank_bound = get_rank_bound(num_experts);
     const int num_max_recv_tokens = rank_bound * num_max_dispatch_tokens_per_rank;
     EP_HOST_ASSERT(num_max_recv_tokens % 4 == 0 and "TMA requires the number of tokens to be multiple of 4");
 
--- a/examples/device/ep/csrc/nixl_ep.cpp
+++ b/examples/device/ep/csrc/nixl_ep.cpp
@@ -1051,6 +1051,8 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
     EP_HOST_ASSERT(expert_bound >= active_expert_bound and expert_bound <= max_num_ranks * num_experts_per_rank);
     EP_HOST_ASSERT(expert_bound % num_experts_per_rank == 0);
     const int rank_bound = expert_bound / num_experts_per_rank;
+    const int num_max_recv_tokens = rank_bound * num_max_dispatch_tokens_per_rank;
+    EP_HOST_ASSERT(num_max_recv_tokens % 4 == 0 and "TMA requires the number of tokens to be multiple of 4");
 
     // Diagnosis tensors
     if (cumulative_local_expert_recv_stats.has_value()) {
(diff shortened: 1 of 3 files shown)


examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1046,6 +1046,11 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
     EP_HOST_ASSERT(topk_idx.dim() == 2 and topk_idx.is_contiguous());
     EP_HOST_ASSERT(x.size(0) == topk_idx.size(0) and x.size(0) <= num_max_dispatch_tokens_per_rank);
     EP_HOST_ASSERT(topk_idx.scalar_type() == c10::CppTypeToScalarType<topk_idx_t>::value);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    const int expert_bound = num_experts.value_or(active_expert_bound);
+    EP_HOST_ASSERT(expert_bound >= active_expert_bound and expert_bound <= max_num_ranks * num_experts_per_rank);
+    EP_HOST_ASSERT(expert_bound % num_experts_per_rank == 0);
+    const int rank_bound = expert_bound / num_experts_per_rank;
```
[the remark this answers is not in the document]
itayalroy: Unrelated to this PR

examples/device/ep/nixl_ep/buffer.py
rakhmets: It seems this comment is outdated.

examples/device/ep/nixl_ep/buffer.py
```diff
@@ -355,9 +355,9 @@ def low_latency_dispatch(
                 are supported. `-1` indices (not selecting any expert) are supported.
             num_max_dispatch_tokens_per_rank: the maximum number of tokens to dispatch, all the ranks must hold the same value.
                 `num_ranks * num_max_dispatch_tokens_per_rank` must be a multiple of 4 for TMA alignment.
-            num_experts: deprecated optional parameter. This value is ignored by low-latency
-                dispatch; the number of experts is determined from the `num_experts_per_rank`
-                passed to `update_memory_buffers()` and the currently active ranks.
+            num_experts: total physical expert capacity used for the dispatch
+                layout. It must cover all active experts and defaults to the
+                active expert count.
```
rakhmets: Please fix the line breaks.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1046,6 +1046,11 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
     EP_HOST_ASSERT(topk_idx.dim() == 2 and topk_idx.is_contiguous());
     EP_HOST_ASSERT(x.size(0) == topk_idx.size(0) and x.size(0) <= num_max_dispatch_tokens_per_rank);
     EP_HOST_ASSERT(topk_idx.scalar_type() == c10::CppTypeToScalarType<topk_idx_t>::value);
+    const int active_expert_bound = active_rank_bound * num_experts_per_rank;
+    const int expert_bound = num_experts.value_or(active_expert_bound);
+    EP_HOST_ASSERT(expert_bound >= active_expert_bound and expert_bound <= max_num_ranks * num_experts_per_rank);
+    EP_HOST_ASSERT(expert_bound % num_experts_per_rank == 0);
```
rakhmets: It seems we don’t need to check the value if `num_experts` is not passed.
Maybe it would be better to have a small helper function:
```C++
int
get_rank_bound(const std::optional<int> num_experts) const {
    if (!num_experts) {
        return active_rank_bound;
    }

    EP_HOST_ASSERT(*num_experts % num_experts_per_rank == 0);
    const int rank_bound = *num_experts / num_experts_per_rank;
    EP_HOST_ASSERT(rank_bound >= active_rank_bound && rank_bound<= max_num_ranks);
    return rank_bound;
}
```

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -1078,26 +1083,26 @@ Buffer::dispatch(const torch::Tensor& x, const torch::Tensor& topk_idx,
         stream_wait(launch_stream, compute_stream);
 
     // Allocate packed tensors
-    auto packed_recv_x = torch::empty({num_experts_per_rank, active_rank_bound * num_max_dispatch_tokens_per_rank, hidden},
+    auto packed_recv_x = torch::empty({num_experts_per_rank, rank_bound * num_max_dispatch_tokens_per_rank, hidden},
```
rakhmets: You can add a constant, and replace `rank_bound * num_max_dispatch_tokens_per_rank` here and in other 4 places.

examples/device/ep/nixl_ep/buffer.py
itayalroy: 1. I changed the description to refer to the output of dispatch(), which should be clearer. I agree this description is not ideal, as the layout is determined by a rank bound derived from num_experts. Accepting rank_bound on dispatch API directly would be cleaner, but it would require an API change, while num_experts is already used correctly by both vLLM and SGLang.
2. get_next_combine_buffer() is currently unused by both integrations, so we may consider removing it in a future release.
3. Longer term, we hope to have stable CUDA graphs without passing `num_experts`/`rank_bound`, then this argument can be deprecated again.

rakhmets: Thanks for the explanation.

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: We only run these asserts once and they are not harmful when `num_experts` is not passed so I prefer to keep it inline for smaller diff unless you are strongly against it

rakhmets: The main reason I left a comment is that it is not the passed parameter that is checked, but the derived value. I would probably even slightly correct my originally proposed code. To make it clear what exactly we want the client code to pass to the function.
```C++
EP_HOST_ASSERT(*num_experts % num_experts_per_rank == 0);
const int active_expert_bound = active_rank_bound * num_experts_per_rank;
const int max_num_experts = max_num_ranks * num_experts_per_rank;
EP_HOST_ASSERT(*num_experts >= active_rank_bound && *num_experts <= max_num_experts);
```
And I propose creating a separate helper function for this, since it’s easily separable logic, and the function already consists of hundreds of lines.

examples/device/ep/csrc/nixl_ep.cpp
```diff
@@ -107,6 +107,17 @@ void Buffer::_refresh_active_rank_bound() {
     set_active_rank_bound(bound);
 }
 
+int Buffer::get_rank_bound(const std::optional<int>& num_experts) const {
+    if (!num_experts)
+        return active_rank_bound;
+
+    EP_HOST_ASSERT(*num_experts % num_experts_per_rank == 0);
```
ofirfarjun7, line 114: Can reach this line with num_experts_per_rank == 0?

examples/device/ep/csrc/nixl_ep.hpp
```diff
@@ -175,6 +175,7 @@ struct Buffer {
     bool _is_rank_connected(int rank_id) const;
     void set_active_rank_bound(int bound);
     void _refresh_active_rank_bound();
+    int get_rank_bound(const std::optional<int>& num_experts) const;
```
rakhmets: It's ok (and even better) to pass std::optional<int> by value, as it's small.
```suggestion
    int get_rank_bound(std::optional<int> num_experts) const;
```

rakhmets (dismissed):

examples/device/ep/csrc/nixl_ep.cpp
itayalroy: Can't, we assert num_experts_per_rank > 0 in the beginning of `Buffer::init` which must be called before dispatch

rakhmets (approved):

PR #2097, opened 2026-08-17
Disable cmake install skip based on mtime

ovidiusm:
## What?
Disable CMake optimization to skip install step based on file modification timestamp

## Why?
CMake optimization to skip install step based on mtime uses 1 second resolution timestamps.

Complex CMake scripts sometimes have directives with the pattern:
- install file
- patch file
- install same file again

The 2nd install step may be skipped if the build machine is very fast and is executed less than 1 second after the first install step. This results into incorrect installation.

[Seen in CI](https://nbuprod.blsm.nvidia.com/nbu-swx-nixl-main/blue/organizations/jenkins/nixl-ci-test-sanitizers/detail/nixl-ci-test-sanitizers/760/pipeline/242/) in Azure SDK installation from source, that sometimes causes inconsistent headers to be installed, with different values of RTTI macro, resulting in memory corruption in the Azure storage plugin init/destroy.

## How?
Set [CMAKE_INSTALL_ALWAYS=1](https://github.com/Kitware/CMake/blob/f0952e573f5557b61847f8ee60641d51b36ef352/Source/cmFileInstaller.cxx#L32) which turns off the mtime-based install skip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability by ensuring installation steps consistently copy generated files, even when timestamps have not changed.
  * Reduced the risk of incomplete or stale artifacts in standard, ROCm, Docker, and manylinux builds.

* **Chores**
  * Applied consistent installation behavior across supported container and package build environments.

### DIFF
--- a/.ci/dockerfiles/Dockerfile.rocm
+++ b/.ci/dockerfiles/Dockerfile.rocm
@@ -78,6 +78,14 @@ ENV PIP_CERT=${SYSTEM_CA_BUNDLE} \
 # libgrpc_plugin_support.so from /opt/nixl/lib. Every stage inherits this ENV.
 ENV LD_LIBRARY_PATH="${NIXL_INSTALL_DIR}/lib:${NIXL_INSTALL_DIR}/lib64:/usr/local/lib:/usr/local/lib64"
 
+# Force CMake to always copy files in install directives, rather than skip based on file modification timestamp.
+# File modification timestamp check in CMake uses 1 second resolution.
+# This causes problems for fast builds that install, patch then reinstall the same file, as the final install step
+# may be incorrectly skipped.
+# Seen in CI as flaky ASAN failure due to inconsistent Azure SDK headers causing memory corruption.
+# Inherited by every *-build stage and `final`.
+ENV CMAKE_INSTALL_ALWAYS=1
+
 # Some ROCm base images ship broken cmake configs for these.
 RUN rm -rf /usr/lib/cmake/grpc /usr/lib/cmake/protobuf
 
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -108,6 +108,13 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
 
 WORKDIR /workspace
 
+# Force CMake to always copy files in install directives, rather than skip based on file modification timestamp.
+# File modification timestamp check in CMake uses 1 second resolution.
+# This causes problems for fast builds that install, patch then reinstall the same file, as the final install step
+# may be incorrectly skipped.
+# Seen in CI as flaky ASAN failure due to inconsistent Azure SDK headers causing memory corruption.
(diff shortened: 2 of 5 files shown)


.ci/dockerfiles/Dockerfile.rocm
```diff
@@ -78,6 +78,14 @@ ENV PIP_CERT=${SYSTEM_CA_BUNDLE} \
 # libgrpc_plugin_support.so from /opt/nixl/lib. Every stage inherits this ENV.
 ENV LD_LIBRARY_PATH="${NIXL_INSTALL_DIR}/lib:${NIXL_INSTALL_DIR}/lib64:/usr/local/lib:/usr/local/lib64"
 
+# Force CMake to always copy files in install directives, rather than skip based on file modification timestamp.
+# File modification timestamp check in CMake uses 1 second resolution.
+# This causes problems for fast builds that install, patch then reinstall the same file, as the final install step
+# may be incorrectly skipped.
+# Seen in CI as flaky ASAN failure due to inconsistent Azure SDK headers causing memory corruption.
+# Inherited by every *-build stage and `final`.
+ENV CMAKE_INSTALL_ALWAYS=1
```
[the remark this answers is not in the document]
ovidiusm: The manylinux case is already handled: contrib/Dockerfile.manylinux is in CI_FILES, so this change bumps CI_IMAGE_TAG and the wheel_base image is rebuilt. It is also moot there, since the wheel stage runs no cmake --install (only autotools make install and ninja install).

For Dockerfile.rocm, azure-build is the only stage invoking cmake --install, and the BUILDER_BASE_IMAGE override is an opt-in path for consuming prebuilt dep images, where the dependency is built in that image's own build and the setting belongs there. Duplicating the ENV across eleven *-build stages is not worth it. Not planning to change this.

dpressle (approved):

contrib/Dockerfile
```diff
@@ -108,6 +108,13 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
 
 WORKDIR /workspace
 
+# Force CMake to always copy files in install directives, rather than skip based on file modification timestamp.
+# File modification timestamp check in CMake uses 1 second resolution.
+# This causes problems for fast builds that install, patch then reinstall the same file, as the final install step
+# may be incorrectly skipped.
+# Seen in CI as flaky ASAN failure due to inconsistent Azure SDK headers causing memory corruption.
+ENV CMAKE_INSTALL_ALWAYS=1
```
tvegas1, line 116: are we sure we want ENV (build and image) and not ARG (build only)? question for other cases too

ovidiusm: We want this to persist, so yes

tvegas1 (approved):

michal-shalev (approved):

brminich (approved):

Issue #1157, opened 2025-12-27
CUDA context crash in libfabric backend progress thread when using GPU Direct RDMA


dmvevents:
## Description

When using the libfabric backend with GPU Direct RDMA (`FI_EFA_USE_DEVICE_RDMA=1`), the progress thread crashes with a CUDA context error because `pthrCudaCtx_` is NULL when the thread starts.

## Environment

- NIXL 0.8.0
- AWS P5.48xlarge (H100 GPUs + EFA)
- TRT-LLM disaggregated inference
- `FI_EFA_USE_DEVICE_RDMA=1`

## Symptoms

```
CUDA error: invalid device context
Segmentation fault during fi_read with GPU memory
```

## Root Cause

In `src/plugins/libfabric/libfabric_backend.cpp`, the progress thread starts in the constructor BEFORE `registerMem()` is called. When the thread tries to access GPU memory via `fi_read()`, the CUDA context (`pthrCudaCtx_`) is NULL.

The UCX backend handles this correctly by restarting the thread when the context changes, but the libfabric backend does not.

## Proposed Fix

Apply CUDA context INSIDE the progress loop on every iteration, not just at thread start:

```cpp
void LibfabricBackend::progressThread() {
    while (!progress_thread_stop_.load()) {
#ifdef HAVE_CUDA
        if (cuda_addr_wa_) {
            vramApplyCtx();  // Apply context on EVERY iteration
        }
#endif
        // ... rest of progress loop
    }
}
```

## Workaround

None - requires source patch.

## Impact

This blocks GPU Direct RDMA for all libfabric users with CUDA memory.

amitrad-aws:
I don't undertsand the issue you are describing.
fi_read is called from the postXfer API, it is not related to the progress thread.
Please share a failing scenario if there is one

dmvevents:
## Concrete Reproduction Steps

### Prerequisites
- AWS P5.48xlarge instance (8x H100, 32 EFA devices)
- Kubernetes cluster with EFA plugin
- Docker image with NIXL + TensorRT-LLM

### Step 1: Deploy with this ConfigMap

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nixl-test-config
data:
  NIXL_BACKEND: "LIBFABRIC"
  NIXL_LIBFABRIC_MAX_RAILS: "32"
  FI_PROVIDER: "efa"
  FI_EFA_USE_DEVICE_RDMA: "1"
[...]

PR #2103, opened 2026-08-17
PLUGINS/UCX: apply a CUDA context on the progress threads

### LINKED ISSUE #2102: [UCX] nixlUcxSharedThread::run() arms the worker with no CUDA context -> SIGSEGV in cuEventQuery (same defect class as #1157)
## Summary

`nixlUcxSharedThread::run()` calls `worker->arm()` without ever establishing a CUDA context on that
thread. When the UCX `cuda_copy` transport implements the arm, it calls `cuEventQuery`, which
segfaults in the driver. This is the same defect class as #1157 (libfabric progress thread), in the
backend that issue assumed was already safe.

We hit this **9 times in 43 hours** on a production 8×B200 PD-disaggregated deployment. It is
almost certainly the unexplained root cause behind sgl-project/sglang#23489 and
sgl-project/sglang#23499, both of which were closed without one.

## Stack (identical on all 9 crashes)

```
Fatal Python error: Segmentation fault

  File "<unknown>", line 0, in pthread_kill
  File "<unknown>", line 0, in gsignal
  File "<unknown>", line 0, in cuEventQuery
  File "<unknown>", line 0, in uct_cuda_base_iface_event_fd_arm
  File "<unknown>", line 0, in ucp_worker_arm
  File "<unknown>", line 0, in nixlUcxWorker::arm() const
  File "<unknown>", line 0, in nixlUcxSharedThread::run()
  File "<unknown>", line 0, in __clone
  File "<unknown>", line 0, in 0xffffffffffffffff
```

A corroborating UCX error appears on the same box shortly before some crashes:

```
cuda_copy_ep.c:217  UCX  ERROR failed to get context id of device GPU0 (0)
```

## Environment

- `nixl` / `nixl-cu13` **1.3.2** (UCX bundled in the wheel: `nixl_cu13.libs/libucp-*.so`)
- CUDA 13.0, driver-side B200 (sm_100), x86_64, 8 GPUs per host
- SGLang v0.5.17, `--disaggregation-transfer-backend nixl`, prefill and decode roles co-located on
  one host (prefill TP4/DP4 on GPUs 0-3, decode TP4/DP4 on GPUs 4-7)
- `UCX_TLS=cuda_copy,cuda_ipc,tcp`, `UCX_MEMTYPE_CACHE=n`, `UCX_RCACHE_MAX_UNRELEASED=1024`
- Each role runs with **all 8 GPUs visible** (`CUDA_VISIBLE_DEVICES` covers 0-7, own devices first)
  so that same-host CUDA IPC can map the peer role's KV memory — but each role only creates CUDA
  contexts on its own 4 devices.

## Analysis

`src/plugins/ucx/ucx_backend.cpp`, the shared thread's entire hot loop:

```cpp
// nixlUcxSharedThread::run(), ~line 339
do {
    worker->progressLoop();
} while (worker->arm() == NIXL_IN_PROG);
```

`arm()` → `ucp_worker_arm` → `uct_cuda_base_iface_event_fd_arm` → `cuEventQuery`. That last call
needs a valid CUDA context current on the calling thread. Nothing in the UCX plugin ever sets one.

Grepping the repo for context management returns hits in exactly one plugin:

```console
$ grep -rn "vramApplyCtx\|pthrCudaCtx\|cuCtxSetCurrent\|cuCtxPushCurrent" --include=*.cpp --include=*.h src/
src/plugins/libfabric/libfabric_backend.h:54:    CUcontext pthrCudaCtx_;
src/plugins/libfabric/libfabric_backend.h:356:    vramApplyCtx();
src/plugins/libfabric/libfabric_backend.cpp:302:    result = cuCtxSetCurrent(pthrCudaCtx_);
src/plugins/libfabric/libfabric_backend.cpp:971:                    vramApplyCtx();
src/plugins/libfabric/libfabric_backend.cpp:1023:        vramApplyCtx();
...
```

Zero hits under `src/plugins/ucx/`.

This is exactly the failure described in #1157 — a backend progress thread touching GPU state with
`pthrCudaCtx_` effectively NULL — and the fix there was to apply the context inside the progress
loop rather than once at thread start. Notably, #1157's own text states:

> The UCX backend handles this correctly by restarting the thread when the context changes, but the
> libfabric backend does not.

The current source does not support that claim. There is no context capture, no `cuCtxSetCurrent`,
and no restart-on-context-change anywhere in the UCX plugin.

## Why this configuration reaches it reliably

The thread only exists when the backend receives `num_threads > 0`:

```cpp
// ucx_backend.cpp:763
const size_t num_threads =
    nixl::getBackendParamDefaulted(init_params.customParams, "num_threads", 0u);
if (num_threads > 0) {
    engine = new nixlUcxThreadPoolEngine(init_params, num_threads);
```

SGLang passes `num_threads=8` for the **prefill** role and `0` for **decode**
(`python/sglang/srt/disaggregation/nixl/conn.py`). In our deployment the prefill role has crashed
9 times and the decode role — same host, same image, same UCX env, same traffic — has **never**
crashed. That asymmetry maps exactly onto the presence or absence of `nixlUcxSharedThread`.

Having 8 visible devices with contexts on only 4 appears to make the bad path much easier to hit,
which fits the `failed to get context id of device GPU0 (0)` error above.

## Reproduction rate

Nine crashes in ~43 hours of steady production traffic (~one per 4.8h), spread evenly across the
clock — not correlated with peak load, deploys, or any particular request shape. Each one kills the
role's scheduler process; peers then fail with gloo/NCCL "Connection reset by peer", and in 3 of 9
cases the survivors wedged for ~5 minutes before aborting with **Xid 43**. GPUs are healthy (zero
uncorrected ECC), no OOM.

## Workaround

Force the UCX backend to zero threads, which removes `nixlUcxSharedThread` entirely:

```python
create_backend("UCX", {"num_threads": "0"})
```

Under SGLang: `SGLANG_DISAGGREGATION_NIXL_BACKEND_PARAMS={"num_threads":"0"}`.

Cost is losing the dedicated progress thread, so progress falls to the caller's threads.

## Suggested fix

Mirror #1157 in the UCX plugin: capture the CUDA context at memory-registration time (or first
VRAM descriptor seen), store it on the engine, and apply it inside `nixlUcxSharedThread::run()`
before progressing/arming — on every iteration, not once at thread start, since the context can
change after the thread has already been created. The dedicated-thread path needs the same
treatment.

Happy to put up a PR if the maintainers agree on the shape.

## Related

- #1157 — CUDA context crash in libfabric backend progress thread (same defect class, fixed there)
- sgl-project/sglang#23489 — identical stack, closed same day as "Resolved via rebase"
- sgl-project/sglang#23499 — identical stack, closed with no linked fix; re-reported 2026-07-07 on
  a different GPU/CUDA/SGLang combination and unanswered since

skeptrunedev:
Fixes #2102.

## Problem

The UCX progress threads are created by NIXL, so no CUDA context is ever made current on them. The
UCX transports they drive call into the CUDA driver anyway: with `cuda_copy` in `UCX_TLS`,
`ucp_worker_arm()` reaches `uct_cuda_base_iface_event_fd_arm()`, which calls `cuEventQuery()` and
segfaults when there is no current context.

```
cuEventQuery
uct_cuda_base_iface_event_fd_arm
ucp_worker_arm
nixlUcxWorker::arm() const
nixlUcxSharedThread::run()
```

This is the same defect class as #1157 in the libfabric plugin, which was fixed there by applying
the context inside the progress loop. That issue's description states the UCX backend "handles this
correctly by restarting the thread when the context changes" — it does not. A repo-wide grep for
`vramApplyCtx|pthrCudaCtx|cuCtxSetCurrent|cuCtxPushCurrent` returns hits only under
`src/plugins/libfabric/`; the UCX plugin has no CUDA context handling at all.

Every agent with a progress thread is exposed, not only ones that request a thread pool:
`nixlUcxThreadEngine` constructs `nixlUcxSharedThread` whenever `enableProgTh` is set (default
true), and `num_threads` only adds the dedicated pool on top. In practice the initiator side is
what crashes, since it is what drives `cuda_copy` activity through the worker the progress thread
arms.

Observed at roughly one crash per 4.8 hours of steady traffic on an 8xB200 SGLang PD deployment
(9 crashes in 43 hours). It is very likely the unexplained cause behind sgl-project/sglang#23489
and sgl-project/sglang#23499, both closed without a root cause.

## Change

- `vramCaptureCurrentCtx()` — provisional capture via `cuCtxGetCurrent()` in the engine
  constructor, so the progress threads have a context from their first iteration rather than
  waiting for a registration that has not happened yet.
- `vramUpdateCtx(addr)` — authoritative capture from the first VRAM `registerMem()` via
  `cuPointerGetAttributes(CU_POINTER_ATTRIBUTE_CONTEXT)`, on the caller's thread.
- `vramApplyCtx()` — called at the top of both `nixlUcxSharedThread::run()` and
  `nixlUcxDedicatedThread::run()`.

The apply has to stay inside the loop rather than run once at thread start, because the threads are
created in the engine constructor, before any memory is registered, so the correct context is not
known at that point. Steady-state cost is one `cuCtxGetCurrent()`, a thread-local read;
`cuCtxSetCurrent()` is only called when the context actually differs.

The context is stored as `std::atomic<void *>` so that `cuda.h` stays out of `ucx_backend.h` and
the class layout is identical between CUDA and non-CUDA builds. All three helpers compile out
without CUDA.

`src/plugins/ucx/meson.build` now picks up `cuda_dep` and `-DHAVE_CUDA` the same way
`src/plugins/libfabric/meson.build` does. This is consistent with `src/core/meson.build`, which
already adds `cuda_dep` alongside a statically linked UCX plugin, and it makes the existing
"CUDA not found. UCX backend will be built without CUDA support" warning in the top-level
`meson.build` accurate for the first time.

## Notes and open questions for reviewers

- **Multiple contexts are not handled.** A registration from a second context replaces the stored
  one. One valid current context is enough to stop the driver from faulting, but a strictly correct
  fix would key the context per worker, which needs a worker-to-device mapping the plugin does not
  currently have. Happy to take direction here if you would prefer that shape.
- **No unit test.** The natural test needs a constructed `nixlUcxEngine`, so it needs UCX and a GPU.
  Pointers to where this would fit in the existing test layout are welcome.
- I have verified the new code compiles cleanly with `-Wall -Wextra -Werror` against CUDA 12.9
  headers in both the CUDA and non-CUDA configurations, that all four driver symbols used
  (`cuCtxGetCurrent`, `cuCtxSetCurrent`, `cuPointerGetAttributes`, `cuGetErrorString`) resolve
  against `libcuda`, and that the added lines are clean under the repo's `.clang-format`. A full
  build and GPU soak are still pending on my side.

## Summary by CodeRabbit

* **New Features**
  * Added CUDA-aware UCX support when CUDA is available.
  * Improved GPU memory registration and context handling across UCX progress threads.
  * Applied CUDA contexts consistently for shared and dedicated progress operations.

* **Bug Fixes**
  * Improved reliability of UCX progress processing involving VRAM and CUDA contexts.
  * Added error reporting when CUDA context application fails.

### DIFF
--- a/src/plugins/ucx/meson.build
+++ b/src/plugins/ucx/meson.build
@@ -38,12 +38,22 @@ ucx_backend_dependencies = [asio_dep,
 ucx_backend_include_directories = [nixl_inc_dirs,
                                    utils_inc_dirs]
 
+ucx_backend_compile_flags = []
+
+# The progress threads must re-apply a CUDA context before driving UCX transports that call into
+# the CUDA driver (see nixlUcxEngine::vramApplyCtx).
+if cuda_dep.found()
+    ucx_backend_dependencies += [cuda_dep]
+    ucx_backend_compile_flags += ['-DHAVE_CUDA']
+endif
+
 if 'UCX' in static_plugins
     ucx_backend_lib = static_library('UCX',
                ucx_backend_sources,
                dependencies: ucx_backend_dependencies,
                include_directories: ucx_backend_include_directories,
                install: false,
+               cpp_args : ucx_backend_compile_flags,
                name_prefix: 'libplugin_')  # Custom prefix for plugin libraries
 else
     ucx_backend_lib = shared_library('UCX',
@@ -51,7 +61,7 @@ else
                dependencies: ucx_backend_dependencies,
                include_directories: ucx_backend_include_directories,
                install: true,
-               cpp_args : ['-fPIC'],
+               cpp_args : ['-fPIC'] + ucx_backend_compile_flags,
                name_prefix: 'libplugin_',  # Custom prefix for plugin libraries
                install_dir: plugin_install_dir,
                build_rpath: get_option('ucx_path') != '' ? get_option('ucx_path') + '/lib' : '',
--- a/src/plugins/ucx/ucx_backend.cpp
+++ b/src/plugins/ucx/ucx_backend.cpp
@@ -33,6 +33,10 @@
 #include "absl/strings/str_split.h"
 #include <asio.hpp>
 
+#ifdef HAVE_CUDA
+#include <cuda.h>
+#endif
+
 namespace {
 [[nodiscard]] uint32_t
 epCloseFlags(const nixl_b_params_t *custom_params) {
@@ -276,6 +280,11 @@ protected:
     virtual void
     run() = 0;
 
+    const nixlUcxEngine *
+    getEngine() const noexcept {
+        return engine_;
+    }
+
 private:
     const nixlUcxEngine *engine_;
     std::vector<nixlUcxWorker *> workers_;
@@ -330,6 +339,10 @@ protected:
         bool timeout = true;
         bool pthr_stop = false;
         while (!pthr_stop) {
+            /* progressLoop() and arm() reach the CUDA driver through UCX's cuda_copy transport,
+             * which needs a current context on this thread. */
+            getEngine()->vramApplyCtx();
+
             for (size_t i = 0; i < pollFds_.size() - 1; i++) {
                 if (!(pollFds_[i].revents & POLLIN) && !timeout) continue;
                 pollFds_[i].revents = 0;
@@ -608,6 +621,10 @@ protected:
         NIXL_DEBUG << "dedicated " << *this << " running";
 
         while (!io_.stopped()) {
+            /* status() below drives progressLoop(), which reaches the CUDA driver through UCX's
+             * cuda_copy transport and needs a current context on this thread. */
+            getEngine()->vramApplyCtx();
+
             if (!requests_.empty()) {
                 io_.poll_one();
             } else {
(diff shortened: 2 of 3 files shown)


iyastreb:
We had CUDA context set before in the past, but then it was removed by https://github.com/ai-dynamo/nixl/pull/946
I think it's not needed with last UCX version.
@rakhmets

src/plugins/ucx/meson.build
```diff
@@ -38,20 +38,30 @@ ucx_backend_dependencies = [asio_dep,
 ucx_backend_include_directories = [nixl_inc_dirs,
                                    utils_inc_dirs]
 
+ucx_backend_compile_flags = []
+
+# The progress threads must re-apply a CUDA context before driving UCX transports that call into
+# the CUDA driver (see nixlUcxEngine::vramApplyCtx).
+if cuda_dep.found()
+    ucx_backend_dependencies += [cuda_dep]
```
rakhmets, line 46: NIXL UCX plugin doesn't use CUDA directly, and shouldn't depend on CUDA libraries. The only libraries that depend on CUDA in this stack are UCX internal libraries (e.g. libuct_cuda).

rakhmets (changes requested):
CUDA context management is handled at UCX level.

brminich (changes requested):
should be fixed in UCX, not NIXL

